PR #47, opened 2021-02-16
DTFS2-3517: Initial GEF environment for Frontend

monkemedia:


gef-ui/server/controllers/mandatory-criteria/index.js
```diff
@@ -0,0 +1,45 @@
+import _isEmpty from 'lodash/isEmpty'
+import * as Api from '../../services/api'
+import { parseBool, errorHandler } from '../../utils/helpers'
+
+const getMandatoryCriteria = async function (req, res) {
+  try {
+    const criteria = await Api.getMandatoryCriteria()
```
pricea1, line 7: I think it would be worth keeping with the Portal & TFM code style of lower case for api just so we don't have to think about it when we have to switch between different projects.

pricea1 (commented):
Just one small suggestion but otherwise looks good to me.
Might be worth @ttbarnes looking over it as he's more familiar with Frontend setup.

e2e-tests/gef/cypress/integration/mandatory-criteria.spec.js
```diff
@@ -0,0 +1,54 @@
...
+    it('redirects the user to the Name Application page when they select True', () => {
+      cy.get('[data-cy="mandatory-criteria-true"]').click()
+      cy.get('[data-cy="form"]').submit()
+      cy.url().should('eq', relative('/gef/name-application'))
+    })
+  })
+
+  describe('Clicking on Cancel', () => {
+    it('keeps the user on the same page FOR NOW', () => {
+      cy.get('[data-cy="cancel-button"]').click()
```
ttbarnes, line 50: would be nice to use a 'page' file that has all the selectors for a given page, but not a biggie. eg https://github.com/foundry4/dtfs2/blob/master/e2e-tests/portal/cypress/integration/pages/eligibilityCriteria.js

gef-ui/views/templates/name-application.njk
```diff
@@ -0,0 +1,32 @@
...
+{% import '_macros/success-message.njk' as success %}
+{% from "govuk/components/radios/macro.njk" import govukRadios %}
+{% from "govuk/components/button/macro.njk" import govukButton %}
+
+{% block pageTitle %}Confirm eligiblity (mandatory criteria){% endblock %}
+
+{% block content %}
+
+  {% if successMessage %}
+    {{ success.message(
```
ttbarnes, line 11: Could we use `render` so it reads

```
successMessage.render( ...
```

It's a good pattern imo, especially with other component names. E.g if we have "table" component it would read `table.render`

gef-ui/views/templates/name-application.njk
```diff
@@ -0,0 +1,32 @@
...
+      message = successMessage
+    )}}
+  {% endif %}
+    <h1 class="govuk-heading-l">
+      Name application
+    </h1>
+
+    {{ criteria | safe }}
+
+    <form action="name-application" method="post">
```
ttbarnes, line 21: I don't think `action="name-application"` does anything? as the server would pick up the form via the post method/route

gef-ui/server/routes/mandatory-criteria.js
```diff
@@ -0,0 +1,13 @@
...
+import { 
+  getMandatoryCriteria,
+  validateMandatoryCriteria
+} from '../controllers/mandatory-criteria'
+import validateToken from './middleware/validate-token'
+
+const router = express.Router()
+
+router.get('/mandatory-criteria', validateToken, (req, res) => getMandatoryCriteria(req, res))
+router.post('/mandatory-criteria', validateToken, (req, res) => validateMandatoryCriteria(req, res))
```
ttbarnes, line 11: just need to add some tests for the routes e.g:
- https://github.com/foundry4/dtfs2/blob/master/trade-finance-manager-ui/server/routes/index.test.js
- https://github.com/foundry4/dtfs2/blob/master/trade-finance-manager-ui/server/routes/case/index.test.js

ttbarnes (commented):
Looking great, just some minor comments

gef-ui/views/templates/name-application.njk
monkemedia: Going to remove this as its not needed :)

gef-ui/server/controllers/mandatory-criteria/index.js
monkemedia: Done

e2e-tests/gef/cypress/integration/mandatory-criteria.spec.js
monkemedia: Done

gef-ui/server/routes/mandatory-criteria.js
monkemedia: Done. Couldn't mock the anonymous function and its controller, so ended up using 'expect.any(Function)'. If you have a solution please let me know

ttbarnes: I think if it's changed to 

```
router.get('/mandatory-criteria', validateToken, getMandatoryCriteria);
```

Then the test can assert the passed controller:

```
expect(get).toHaveBeenCalledWith('/mandatory-criteria', validateToken, getMandatoryCriteria);
```

?

monkemedia: doesn't seem to work as i need to pass both the req and res parameters to the controllers

ttbarnes: OK cool would be nice to do, not a blocker for me.

The req/res should get passed to the controller automagically, I suspect the mock `validateToken` might not be passing this in a `next()` call, in the test

Can improve later...

ttbarnes (approved):

PR #55, opened 2021-02-22
DTFS2-3468—fe

monkemedia:


portal-api/api-tests/v1/gef/application.api-test.js
```diff
@@ -0,0 +1,271 @@
...
+  //     await as(anEditor).post(newMandatoryCriteria).to('/v1/gef/mandatory-criteria-versioned');
+  //     await as(anEditor).remove('/v1/gef/mandatory-criteria-versioned/1');
+
+  //     const { status, body } = await as(anEditor).get('/v1/gef/mandatory-criteria-versioned/1');
+
+  //     expect(status).toEqual(200);
+  //     expect(body).toEqual({});
+  //   });
+  // });
+});
```
ttbarnes, line 271: should this all be commented out?

gef-ui/templates/partials/name-application.njk
```diff
@@ -1,25 +1,67 @@
...
+        text: "Continue",
+        attributes: {
+        'data-cy': 'continue-button'
+      }
       }) }}
 
-      <a class="govuk-link" href="#">Cancel</a>
+      <a class="govuk-link" href="/gef/name-application" data-cy="cancel">Abandon application</a>
     </div>
   </form>
```
ttbarnes, line 66: ideally would have component tests as well for njk templates

portal-api/api-tests/v1/gef/application.api-test.js
monkemedia: For me to work on the FE I had to merge matts work into mine, so make sure you leave this comment on his PR, unless of course he has already resolved this issue.

gef-ui/templates/partials/name-application.njk
monkemedia: Shouldn't components be tested via E2E tests rather than unit tests? Usually I would do unit tests for components, but thats when you have javascript methods etc within the component

ttbarnes: Either approach works - in Portal v2 we initially started asserting "X page renders Y" in the e2e tests, but after say 100 e2e tests, the e2e tests start to run slowly. We then took the approach of creating component tests to check that a component renders XYZ when X params/props are passed to the component. Then the e2e tests could be more "click on X, go to Y and something happens" and less checking that things are rendered on a page (to a degree)

I think as long as it's tested somewhere, it's ok. Component tests vs e2e tests is more of an optimisation/faster feedback loop I guess

PR #57, opened 2021-03-02
GEF - Backend API - Application View 

mwyld:
Updates includes:

- Application Details View
- Exporter Information
- Facilities
- Mandatory Criteria

- Updated Mocks

Some Unit tests - More to follow


portal-api/src/v1/gef/controllers/exporter.controller.js
```diff
@@ -0,0 +1,30 @@
...
+
+exports.getById = async (req, res) => {
+  const collection = await db.getCollection(collectionName);
+  const doc = await collection.findOne({ _id: ObjectId(String(req.params.id)) });
+  const response = {
+    status: exporterStatus(doc),
+    details: doc,
+    validation: exporterValidation(doc),
+  };
+  res.status(200).send(response);
```
pricea1, line 17: Can we have a 404 returned if the exporter isn't found

portal-api/src/v1/gef/controllers/facilities.controller.js
```diff
@@ -0,0 +1,70 @@
...
+
+exports.getById = async (req, res) => {
+  const collection = await db.getCollection(collectionName);
+  const doc = await collection.findOne({ _id: ObjectId(String(req.params.id)) });
+  const response = {
+    status: facilitiesStatus(doc),
+    details: doc,
+    validation: facilitiesValidation(doc),
+  };
+  res.status(200).send(response);
```
pricea1, line 38: return 404 if not found

portal-api/src/v1/gef/routes.js
```diff
@@ -1,31 +1,50 @@
...
+const facilities = require('./controllers/facilities.controller');
+const mandatoryCriteriaVersioned = require('./controllers/mandatoryCriteriaVersioned.controller');
 
 const router = express.Router();
 
 // Application
 router.route('/application')
-  .get(validate({ role: ['maker', 'checker'] }), application.getAll)
-  .post(validate({ role: ['maker'] }), application.create);
+  .get(validate({ role: ['checker', 'data-admin'] }), application.getAll)
```
pricea1, line 13: should this also include 'maker'? As this stands, a maker cannot list all applications. Is this whats required?

utils/mock-data-loader/gef/api.js
```diff
@@ -41,12 +43,104 @@ const listApplication = async (token) => {
       Accepts: 'application/json',
       Authorization: token || '',
     },
-    url: `${urlRoot}/v1/gef/application?page=null&pageSize=null`,
+    url: `${urlRoot}/v1/gef/application`, // ?page=null&pageSize=null
```
pricea1, line 46: This needs the applicationID to identify which application to delete

pricea1 (commented):
A few small suggestions but otherwise LGTM

portal-api/src/v1/gef/models/facilities.js
```diff
@@ -0,0 +1,45 @@
...
+      this.currency = req.currency ? String(req.currency) : null;
+      this.value = req.value !== null ? Number(req.value) : null;
+      this.coverPercentage = req.coverPercentage !== null ? Number(req.coverPercentage) : null;
+      this.interestPercentage = req.interestPercentage !== null ? Number(req.interestPercentage) : null;
+      // 0 - In advance (quarterly) | 1 - In Arrears (quarterly)
+      this.paymentType = req.paymentType !== null ? Number(req.paymentType) : null;
+      this.updatedAt = Date.now();
+    }
+  }
+}
```
ttbarnes, line 41: Thinking ahead for mapping GEF deals / facilities - in a GEF deal/application, do we have a dealType ? e.g:

```
{
   ...application/deal,
   dealType: 'gef'
}
```

For facilities - in Portal we have `facilityType` as 'bond' or 'loan'. I can see there is `facility.type` of 0 or 1 for GEF facility.

Is there are benefit to having 0 or 1 values / what's the thinking behind it? Maybe coud just have it as 'Cash' or 'Contingent'?  

i'm just thinking of future TFM integration and mappings, particularly around reports/filtering.

pricea1: Good point. And if we need Bonds, Loans & Gef deals in a single list on a dashboard, won't we need them all in the same collection so mongo can do the sorting & filtering? So using deals collection instead on gef-application? Which also means aligning the fields used in GraphQL. 
Or maybe we need a new collection which contains index of deals & gef-applications and only contains fields required by dashboard?

ttbarnes: Seems that one collection of deals/applications, for GEF and Portal - and similarily a single collection for all facilities, would be good way to go. I think splitting out would have more complications - e.g if dashboard wants to add or change a filter in the future, could be a pain.

If all in the same collection - the data models would be different, but I think that's okay as we can do mappings in the graphql queries/filtering, based on `dealType` or `facilityType`. In TFM there's already some differnet mappings happening depending on the facilityType of bond/loan.

But I think either way it'd be good to have `dealType` and `facilityType` aligned; otherwise would be confusing having some facilities as "type 0" or "type 1" if the other facilities are not numbered like this

pricea1: @mwyld We'll need to store the GEF application in the same collection. As there'll be additional work around graphql to get this working it may be better to raise a separate ticket to do this so it doesn't hold Rich up.

@mwyld Can you align facilityType with Portal values in this ticket before submitting? i.e. have the value as gef-loan & gef-bond or whatever the gef equivalent is.

We'll need to raise the additional work for the collections with Gavin.

portal-api/src/v1/gef/controllers/facilities.controller.js
mwyld: updated but returned 204 No Content

portal-api/src/v1/gef/routes.js
mwyld: updated

utils/mock-data-loader/gef/api.js
mwyld: I believe your looking at the wrong route here

portal-api/src/v1/gef/models/facilities.js
mwyld: I agree lets raise a separate ticket on this and have a chat about amending it, so it doesn't block Richy

mwyld:
unit tests added

utils/mock-data-loader/gef/api.js
pricea1: As far as I can see, it's the /application/:id that has a delete associated with it? Is it referring to somewhere else?

**portal-api/src/v1/gef/routes.js:**
router.route('/application')
  .get(validate({ role: ['maker', 'checker', 'data-admin'] }), application.getAll)
  .post(validate({ role: ['maker', 'data-admin'] }), application.create);

router.route('/application/:id')
  .get(validate({ role: ['maker', 'checker', 'data-admin'] }), application.getById)
  .put(validate({ role: ['maker', 'data-admin'] }), application.update)
  .delete(validate({ role: ['maker', 'data-admin'] }), application.delete);

portal-api/src/v1/gef/models/facilities.js
ttbarnes: Just realised we also need to have an `associatedDealId` in each facility

PR #58, opened 2021-03-03
DTFS2-3470—fe

monkemedia:


gef-ui/server/controllers/name-application/index.js
```diff
@@ -24,16 +24,16 @@ const createApplication = async (req, res) => {
       userId,
     });
 
+    // eslint-disable-next-line no-underscore-dangle
+    return res.redirect(`application-details/${application._id}`);
+  } catch (err) {
     // Show validation errors from server
-    if (application.response && application.response.status === 422) {
+    if (err.status === 422) {
```
pricea1, line 31: What are the requirements for dealing with GEF validation errors? In Portal, a user can save an invalid deal (e.g. missing a mandatory field) and continue working on the deal.
This doesn't look like it will render the application that was saved.

gef-ui/server/utils/helpers.js
```diff
@@ -52,10 +58,101 @@ const validationErrorHandler = (errs, href = '') => {
...
+          list.push(`<li>${value}</li>`);
+        }
+      });
+
+      return { html: `<ul class="is-unstyled">${list.join('')}</ul>` };
+    }
+
+    if (options.isCurrency) {
+      return {
+        text: `${commaNumber(val)} ${currency}`,
```
pricea1, line 88: What's the requirement for currency formatting in GEF. On Portal, the requirement was for the format to be updated as the user type. As GEF is the same service I suspect the same functionality will be required.
see:
[https://github.com/foundry4/dtfs2/blob/1bc68911de1b7f5a189aeb434dc1293a83214c11/portal/templates/bond/bond-financial-details.njk#L35](https://github.com/foundry4/dtfs2/blob/1bc68911de1b7f5a189aeb434dc1293a83214c11/portal/templates/bond/bond-financial-details.njk#L35)

It uses iMask, which needs to be included in each page script. On page load it looks for the data-mask attribute
[https://github.com/foundry4/dtfs2/blob/1bc68911de1b7f5a189aeb434dc1293a83214c11/portal/scripts/masked-inputs.js#L20](https://github.com/foundry4/dtfs2/blob/1bc68911de1b7f5a189aeb434dc1293a83214c11/portal/scripts/masked-inputs.js#L20)

gef-ui/styles/styles.scss
```diff
@@ -1 +1,19 @@
...
+@import "./helpers.scss";
+
+.govuk-section-break {
+  &--bold {
+    border: 0;
+    border-bottom: 3px solid $govuk-text-colour;
+  }
+}
+
+@media (min-width: 40.0625em) {
```
pricea1, line 11: What does this min-width refer too? I wasn't aware there were any responsive designs?

portal-api/src/v1/gef/routes.js
```diff
@@ -1,31 +1,50 @@
...
+const facilities = require('./controllers/facilities.controller');
+const mandatoryCriteriaVersioned = require('./controllers/mandatoryCriteriaVersioned.controller');
 
 const router = express.Router();
 
 // Application
 router.route('/application')
-  .get(validate({ role: ['maker', 'checker'] }), application.getAll)
-  .post(validate({ role: ['maker'] }), application.create);
+  .get(validate({ role: ['checker', 'data-admin'] }), application.getAll)
```
pricea1, line 13: I've added a comment on Rich's PR about this too. Should a maker also be able to list applications?

pricea1 (commented):
A few comment that are worth addressing. 
Does this PR also include the PR Rich submitted? There seems to be identical changes in both.

gef-ui/server/controllers/name-application/index.js
monkemedia: Same here Andrew. Its just here where they MUST enter a Reference name

gef-ui/server/utils/helpers.js
monkemedia: The examples you have given are for inputs. Im just showing data (price with currency) in a summary list. When I start creating the inputs, I shall look at this.

gef-ui/styles/styles.scss
monkemedia: So the prototype kit already has break points for mobile devices. I just stole one of the break points (40.0625em) and used it here.

portal-api/src/v1/gef/routes.js
monkemedia: For me to work on this ticket, I had to merge Matts work into mine, this is why you are seeing PORT API changes. Once Matt has successfully merged his changes into MASTER, this will soon go.

gef-ui/server/controllers/name-application/index.js
pricea1: Ah - this is the start of the creation? Same with Portal - must have a deal name to proceed.

gef-ui/server/utils/helpers.js
pricea1: It also applies to text
[https://github.com/foundry4/dtfs2/blob/1bc68911de1b7f5a189aeb434dc1293a83214c11/portal/templates/bond/_macros/bond-submission-details.njk#L121](https://github.com/foundry4/dtfs2/blob/1bc68911de1b7f5a189aeb434dc1293a83214c11/portal/templates/bond/_macros/bond-submission-details.njk#L121)

gef-ui/styles/styles.scss
pricea1: If the GEF design doesn't require breakpoints then best remove this. We're not designing for mobile or tablets afaik

portal-api/src/v1/gef/routes.js
pricea1: I think we'll need to keep them separate in future if possible so you're both working off master. Matts PR has a couple of changes needed and there's danger they'll be overwritten if this PR is merged after Matts. Once Matt has merged, can you rebase your branch before merging into master?

gef-ui/styles/styles.scss
monkemedia: Even though we are not designing for mobile or tablet, Gov prototype is responsive, so when using all their patterns, the  application will naturally fit on a mobile device. By removing this breakpoint, this will be the only page which will break on smaller devices. If you happy for the page to go from this:
Screenshot 2021-03-04 at 09 49 18

to this:
Screenshot 2021-03-04 at 09 50 36

then i shall remove

gef-ui/server/utils/helpers.js
monkemedia: Ok.. will need to speak to you about this as Im not using any html as its generated by javascript.

portal-api/src/v1/gef/routes.js
monkemedia: Yeah thats fair enough. Merging matts work into mine is not the best solution, and there is a risk of overwriting changes etc. Im just worried that all our tasks will be delayed, as mocking data can only take you so far. If there is a better solution please let me know.

gef-ui/styles/styles.scss
monkemedia: Just to let you know, I have removed the breakpoint. We can always re introduce it at a later stage

portal-api/src/v1/gef/routes.js
pricea1: It's worth brining up with Gavin. I guess it's the danger of splitting the same function between frontend & backend. Maybe we could have a gef branch that you both branch from and then submit gef branch for PR once you're both ready? 
Given the limited timeframe we could probably live with what we have now and rely on e2e & api tests to highlight merge issues

gef-ui/styles/styles.scss
pricea1: I'd prefer any unnecessary code to be removed for clarity. Best confirm with Misha or Gavin that we're not concerned with mobile/tablet layouts.

portal-api/src/v1/gef/routes.js
monkemedia: Yeah that could be a good idea. Let me speak to Matt

gef-ui/server/utils/helpers.js
pricea1: yeah, lets speak this afternoon. What happens when JS is switched off? For accessibility, it should still be rendered and working when JS is off.
[https://www.gov.uk/service-manual/technology/accessibility-for-developers-an-introduction](https://www.gov.uk/service-manual/technology/accessibility-for-developers-an-introduction)
[https://www.gov.uk/service-manual/technology/using-progressive-enhancement](https://www.gov.uk/service-manual/technology/using-progressive-enhancement)

monkemedia: All the logic is server side, then to show it on the front end I use nunjucks for loop

`{% for item in facilities.data %}
    <h3 class="govuk-heading-m" data-cy="facility-type-heading">
      {{item.heading}} facility
    </h3>
    {{ govukSummaryList({
      rows: item.rows,
      attributes: {
        'data-cy': 'facility-summary-list'
      }
    }) }}
  {% endfor %}`

pricea1: Thats fine. I misunderstood and thought you were referring to client side JS.
I think you can add attributes to each item by populating the row.value.html property rather than row.value.text. 
[https://design-system.service.gov.uk/components/summary-list/#options-summary-list-example--rows](https://design-system.service.gov.uk/components/summary-list/#options-summary-list-example--rows)

So something like
``` 
html: `<span data-mask="currency">${commaNumber(val)} ${currency}</span>`,
```
It would ensure consistency of currency formatting across the site

pricea1 (approved):
LGTM

PR #66, opened 2021-03-15
DTFS2-3476—fe Company Address page

monkemedia:


gef-ui/server/controllers/exporters-address/index.js
```diff
@@ -0,0 +1,88 @@
...
+
+    // User has selected No so take them to about exporters page
+    if (correspondence === 'false') {
+      return res.redirect('about-exporter');
+    }
+
+    if (correspondence === 'true' && _isEmpty(postcode)) {
+      correspondenceError = {
+        errRef: 'postcode',
+        errMsg: 'Please enter a postcode',
```
ttbarnes, line 60: More design/UX, but GDS says to avoid using "please" in error messages; should be "Enter a postcode" https://design-system.service.gov.uk/components/error-message/

gef-ui/templates/includes/exporters-address/seperate-correspondence.njk
```diff
@@ -0,0 +1,127 @@
...
+              onClick: "correspondenceAddress.toggle(false)"
+            }
+          },
+          {
+            value: false,
+            text: "No",
+            checked: correspondence === "false",
+            attributes: {
+              'data-cy': 'correspondence-no',
+              onClick: "correspondenceAddress.toggle(true)"
```
ttbarnes, line 38: client side JS isn't good for accessibility - i'm not sure on the GEF design here, but in Portal we used a pattern where depending on what radio button you click, a different NJK file can be rendered - like a show/hide. Avoids having client side js, better for accessibility: https://design-system.service.gov.uk/components/radios/conditional-reveal/index.html

Then each radio button can be like this with the `conditional` option

```
{
  value: "test",
  text: "Test",
  conditional: {
    html: someOtherFormFields.render(user)
  }
}
```

gef-ui/server/controllers/application-details/index.js
```diff
@@ -1,24 +1,26 @@
...
 import {
-  mapSummaryList, status, facilityType, parseBool,
+  mapSummaryList, status, facilityType,
 } from '../../utils/helpers';
 import { exporterItems, facilityItems } from '../../utils/displayItems';
 
 const applicationDetails = async (req, res) => {
   try {
+    const NOT_STARTED = 'NOT_STARTED';
+    const COMPLETED = 'COMPLETED';
```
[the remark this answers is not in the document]
monkemedia: Done

gef-ui/server/controllers/exporters-address/index.js
monkemedia: Done

gef-ui/server/controllers/exporters-address/index.js
```diff
@@ -0,0 +1,88 @@
...
+import { validationErrorHandler } from '../../utils/helpers';
+
+function registeredAddressHandler(registeredAddress) {
+  const items = [];
+
+  if (registeredAddress) {
+    Object.values(registeredAddress).forEach((address) => {
+      items.push(`<p class="govuk-body">${address}</p>`);
+    });
+  }
```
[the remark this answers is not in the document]
monkemedia: Done - I think i was over complicating it

gef-ui/templates/includes/exporters-address/seperate-correspondence.njk
monkemedia: Hey Tony.. So i looked into the conditional reveal before implementing my code. I noticed that there was no way i could update it so that it would match our designs (ie. horizontal layout instead of vertical). So I ended up creating a small method to cater for this. Also, I catered for accessibility when coding it, so that all hidden inputs are revealed when JS is turned off in the browser. let me know what you think?

ttbarnes: OK as long as it works - in Portal, we had same scenario - did a similar approach with client side JS originally; all seemed well, but it would act differently in IE and screen readers. We ended up changing the design to use the radio reveal component. So would just be aware of that ...happy to get it merged for now

ttbarnes (approved):

gef-ui/templates/includes/exporters-address/seperate-correspondence.njk
monkemedia: Cheers Tony.. Will test this further later on

PR #68, opened 2021-03-17
GEF: Companies House and OS Places integration

mwyld:


reference-data-proxy/package.json
```diff
@@ -9,7 +9,7 @@
     "start:dev": "npx nodemon src/index.js",
     "unit-test": "jest --notify --coverage --verbose --config=unit.jest.config.js ",
     "unit-test-quick": "jest --config=unit.jest.config.js --testTimeout=10000",
-    "api-test": "jest --notify --coverage --verbose --runInBand --config=api-test.jest.config.js ",
+    "api-test": "jest --notify --coverage --verbose --runInBand --config=api-test.jest.config.js --watch",
```
pricea1, line 12: Can you remove --watch. The api-test is run as part of the pipeline which I don't think will finish & close with --watch flag

pricea1 (commented):
Just a jest flag to remove I think otherwise LGTM

reference-data-proxy/package.json
mwyld: removed

pricea1 (approved):
Good to go

PR #75, opened 2021-03-18
GEF: ENUMS, updates and SIC Codes

mwyld:


pricea1 (commented):
Can we have them defined in a constants file so they are all defined in one place like
https://github.com/foundry4/dtfs2/blob/master/portal-api/src/constants/deal.js

then refer to them with
status: CONSTANTS.FACILITIES.STATUS.NOT_STARTED

It will make is easier to maintain.

pricea1 (approved):
LGTM

mwyld:
Hi @pricea1, I've checked in a larger update. Could you look over and let me know if things look okay? I can separate out to different branches if you think it's worth it

PR #76, opened 2021-03-23
DTFS2 - 3479 postcode search

monkemedia:


gef-ui/server/controllers/companies-house/index.js
```diff
@@ -24,36 +24,37 @@ const validateCompaniesHouse = async (req, res) => {
...
-        errors: validationErrorHandler(regNumberError),
-        regNumber,
-        applicationId,
-      });
+    if (companiesHouseErrors.length === 0) {
+      companiesHouseDetails = await api.getCompaniesHouseDetails(regNumber, exporterId);
     }
 
-    await api.getCompaniesHouseDetails(regNumber, exporterId);
+    if (companiesHouseDetails && companiesHouseDetails.status === 422) {
```
pricea1, line 44: Is it only status 422 that gets returned on an error from CH? Should we check for a 200 instead and assume anything else is an error?

gef-ui/server/controllers/enter-exporters-correspondence-address/index.js
```diff
@@ -0,0 +1,77 @@
+import * as api from '../../services/api';
+import { validationErrorHandler } from '../../utils/helpers';
+
+const enterExportersCorrespondenceAddress = async (req, res) => {
+  const { params, session } = req;
+  const { applicationId } = params;
+  const { address } = session;
+  const parseAddress = address ? JSON.parse(address) : null;
```
pricea1, line 8: Why are we storing the exporter address in a session? Won't this populate a second application they create in the same session with the exporter from the first application?

gef-ui/server/controllers/exporters-address/index.js
```diff
@@ -43,29 +44,39 @@ const validateExportersAddress = async (req, res) => {
...
+    }
 
-    // return res.redirect('exporters-address');
-  } catch (err) {
-    // Validation errors
-    if (err.status === 422) {
-      return res.render('partials/companies-house.njk', {
-        errors: validationErrorHandler(err.data),
+    // Check for validation errors
+    if (addresses && addresses.status === 422) {
```
pricea1, line 59: Same question as above about whether only a 422 status gets returned on error. Should we assume anything other than 200 is error?

gef-ui/server/controllers/name-application/index.js
```diff
@@ -5,35 +5,24 @@ const nameApplication = async (req, res) => res.render('partials/name-applicatio
...
       userId,
     });
 
-    // eslint-disable-next-line no-underscore-dangle
-    return res.redirect(`application-details/${application._id}`);
-  } catch (err) {
-    // Show validation errors from server
-    if (err.status === 422) {
+    // Validation errors
+    if (application.status === 422) {
```
pricea1, line 17: Same as above. Should we check its !== 200 instead?

gef-ui/server/utils/displayItems.js
```diff
@@ -9,7 +9,7 @@ const exporterItems = (exporterUrl) => [
   {
```
pricea1, line 9: Can the filename be changed to display-items.js? Camel case can work ok on local but cause issues on Azure if its not referenced correctly

gef-ui/server/utils/helpers.js
```diff
@@ -18,7 +18,7 @@ const apiErrorHandler = ({ code, response }) => {
   }
   // Is validation error
   if (response.status === 422) {
```
pricea1, line 20: This might explain why you're only checking to 422 errors in other files. If this catches all other errors then ignore the comments above.

pricea1 (commented):
Mostly questions about http status checks but I think you've got that covered in gef-ui/server/utils/helpers.js. Didn't get to that until after I'd added the other comments. If helpers.js is catching all other http errors then ignore the other comments.

Also a question about storing exporter in session

Otherwise LGTM

gef-ui/server/utils/helpers.js
monkemedia: Yeah, all errors are THROWN, except for 422 errors. Basically all errors will be caught in the catch method except for 422. Hope that makes sense

pricea1: Great

gef-ui/server/controllers/enter-exporters-correspondence-address/index.js
monkemedia: Hope this was made clear in the stand up

PR #81, opened 2021-03-29
DTFS-3846—fe industry

monkemedia:


gef-ui/server/controllers/about-exporter/index.js
```diff
@@ -1,6 +1,15 @@
 import * as api from '../../services/api';
 import { validationErrorHandler, isTrueSet } from '../../utils/helpers';
 
+const mappedIndustries = (industries, selectedIndustry) => {
+  if (!industries) { return null; }
+  return industries.map((industry) => ({
+    value: JSON.stringify(industry),
+    html: `${industry.name}<br />${industry.class.name}`,
+    checked: selectedIndustry ? JSON.stringify(industry) === selectedIndustry : null,
+  }));
```
ttbarnes, line 10: i'd have any html in the template itself, not a blocker to merge

gef-ui/server/controllers/about-exporter/index.js
```diff
@@ -29,47 +40,58 @@ const validateAboutExporter = async (req, res) => {
...
+        aboutExporterErrors.push({
+          errRef: 'probabilityOfDefault',
+          errMsg: 'Enter the probability of default',
+        });
+      }
+
+      if (!body.isFinanceIncreasing) {
+        aboutExporterErrors.push({
+          errRef: 'isFinanceIncreasing',
+          errMsg: 'Enter the elect whether your financing to the exporter is increasing as a result of this new GEF facility ',
```
ttbarnes, line 73: Should this be "Slect whether ..." ?

gef-ui/server/utils/helpers.js
```diff
@@ -63,19 +64,15 @@ const mapSummaryList = (data, itemsToShow) => {
       return { html: '<span class="has-text-danger" data-cy="required">Required</span>' };
     }
 
-    if (val === null) {
+    /* Clean-Deep removes any properties with Null value from an Object. Therefore if all
+    properties are Null, this leaves us with an Empty Object. isEmpty checks to see if the
+    Object is empty or not. */
+    if (val === null || _isEmpty(cleanDeep(val))) {
```
ttbarnes, line 70: Could just have a nunjuck filter to display dash if there is no value?
```<p>{{ someValue | dashIfEmpty }}</p>```

https://github.com/foundry4/dtfs2/blob/master/gef-ui/server/nunjucks-configuration/filter-dashIfEmpty.js

I think either works though. Probably more of a personal preference. What do you think?

File: gef-ui/server/utils/helpers.js [hunk 4]
```diff
@@ -123,7 +119,10 @@ const mapSummaryList = (data, itemsToShow) => {
         items: [
           ...(href ? [{
             href,
-            text: `${value ? 'Change' : 'Add'}`,
+            /* Clean-Deep removes any properties with Null value from an Object. Therefore if all
+            properties are Null, this leaves us with an Empty Object. isEmpty checks to see if the
+            Object is empty or not. */
+            text: `${!_isEmpty(cleanDeep(value)) ? 'Change' : 'Add'}`,
```
ttbarnes, line 125: probably worth putting into it's on function as it's duplicated, maybe something like

```
const isEmpty = (value) =>
  _isEmpty(cleanDeep(value)
```

ttbarnes (commented):
LGTM, couple of minor comments

gef-ui/server/controllers/about-exporter/index.js
monkemedia: The radio buttons are dynamic, so wont be able to hard code them in. Is there a way to use a for loop for items in the njk file?

gef-ui/server/utils/helpers.js
monkemedia: The whole summary list is generated via JS, so there is html in the template.

gef-ui/server/controllers/about-exporter/index.js
ttbarnes:
```
{% for industry in industries %}
  <p>{{ industry.name }}</p>
{% endfor %}
```

gef-ui/server/utils/helpers.js
monkemedia: Done. Also added unit tests for it too

gef-ui/server/controllers/about-exporter/index.js
monkemedia: so this is my template code for the radio buttons:

`{{ govukRadios({
     idPrefix: "industry",
     name: "selectedIndustry",
     items: industries, // Mapped in the controller
     attributes: {
       'data-cy': 'industries'
     }
}) }}`

So are you saying to wrap this macro into a for loop?

ttbarnes: Sorry, not familiar with design /  didn't realise there were radios involved. I think what you have is ok - I don't think it's possible to populate the govukRadios items attribute dynamically, I think, can only pass in an array and that's it 👍

ttbarnes (approved):

PR #86, opened 2021-04-01
DTFS2-3574—fe choose facility

monkemedia:


gef-ui/server/controllers/facilities/index.js
```diff
@@ -0,0 +1,87 @@
...
+import { validationErrorHandler, isTrueSet } from '../../utils/helpers';
+import { FACILITY_TYPE } from '../../../constants';
+
+const facilities = async (req, res) => {
+  const { params, query } = req;
+  const { applicationId, facilityId } = params;
+  const { status } = query;
+  let { facilityType } = query;
+
+  facilityType = facilityType || 'CASH';
```
ttbarnes, line 11: can we use a constant for CASH

gef-ui/templates/partials/facilities.njk
```diff
@@ -0,0 +1,94 @@
...
+          'data-cy': 'back-link'
+        }
+      }) }}
+    {% endif %}
+
+  <header data-cy="application-details-page">
+    <span class="govuk-caption-xl" data-cy="heading-caption">
+      Add a facility
+    </span>
+  </header>
```
ttbarnes, line 34: I don't know the design, but it looks like this should be a heading tag?

monkemedia: I have been following https://design-system.service.gov.uk/styles/colour/ You can see that when they use a heading Caption they use a span tag, while the main title is in a heading tag. Let me know your thoughts

ttbarnes: I'm not following, if it's top level heading it should be h1/h2 tag https://design-system.service.gov.uk/styles/typography/

easy to change, happy to merge either way

ttbarnes (approved):

PR #87, opened 2021-04-01
GEF - multiple minor updates

mwyld:


mwyld:
More unit tests to follow, this is to aid Richards check in

pricea1 (approved):
LGTM

PR #94, opened 2021-04-16
Feature/delete facility

monkemedia:


gef-ui/templates/partials/facility-confirm-deletion.njk
```diff
@@ -0,0 +1,37 @@
+{% extends "index.njk" %}
+{% from "govuk/components/button/macro.njk" import govukButton %}
+
+{% block pageTitle %}Confirm that you want to delete {{ heading }} facility 1?{% endblock %}
```
ttbarnes, line 4: Should this be
```
delete {{ heading }} facility {{ facilityId}} ?
```

instead of "facility 1"?

ttbarnes (approved):
Beautiful

gef-ui/templates/partials/facility-confirm-deletion.njk
monkemedia: Good catch

PR #115, opened 2021-06-21
fix: update probability percentage validation

davehlewis:


gef-ui/server/controllers/about-exporter/index.js
```diff
@@ -78,12 +78,12 @@ const validateAboutExporter = async (req, res) => {
       }
     }
 
-    // Regex tests to see if value is a number between 0 and 100
-    const zeroToOneHundredRegex = /^([0-9]|[1-9][0-9]|100)$/;
-    if (body.probabilityOfDefault && !zeroToOneHundredRegex.test(body.probabilityOfDefault)) {
+    const maxPercentage = 14.1;
+    const probabilityOfDefault = Number(body.probabilityOfDefault);
+    if (Number.isNaN(probabilityOfDefault) || probabilityOfDefault > maxPercentage) {
```
pbradley-f4, line 83: Not one for the first day but extract this to a utils method of isNumberGreaterThan(maxPercentage) or the like

pricea1 (commented):
Agree with Pauls comments otherwise code lgtm.

In future, can you add the ticket reference at the start of the commit message
e.g. dtfs2-3969 update probability

pricea1 (approved):

gef-ui/server/controllers/about-exporter/index.js
davehlewis: This is a specific use case for validation so doesn't really suit being a util. Returns true if it is not a number for example.

gef-ui/server/controllers/about-exporter/index.js
```diff
@@ -78,12 +78,17 @@ const validateAboutExporter = async (req, res) => {
...
+    const probabilityOfDefault = Number(body.probabilityOfDefault);
+    if (
+      Number.isNaN(probabilityOfDefault)
+      || !precisionTwoRegex.test(body.probabilityOfDefault)
+      || probabilityOfDefault > maxPercentage
+    ) {
       aboutExporterErrors.push({
         errRef: 'probabilityOfDefault',
-        errMsg: 'You must enter a percentage between 0% to 100%',
+        errMsg: `You must enter a percentage between 0% to ${maxPercentage}}%`,
```
pbradley-f4, line 91: too many `}`

davehlewis: Sorry should have caught that

PR #124, opened 2021-06-22
DTFS2-3066 - tfm - sort deals by id

ttbarnes:


pricea1 (commented):
Would it be better to do the sort on the mongo collection rather than in the reducer?
i.e  deals = await dealsCollection.find(query).sort(sortParams).toArray();

ttbarnes:
> Would it be better to do the sort on the mongo collection rather than in the reducer?
> i.e deals = await dealsCollection.find(query).sort(sortParams).toArray();

That makes a lot more sense moving forward. I just hooked into the existing sorting in placE. Eventually it'll need to be done from the central-api GET/query so i'll get this in place now.

ttbarnes:
@pricea1 updated

PR #131, opened 2021-06-28
DTFS2-3969: remove decimal point limit from default probabilitiy

davehlewis:
Removes decimal point requirement from probability of default question


ttbarnes (approved):
LGTM

PR #132, opened 2021-06-28
TFM e2e tests calling number generator

ttbarnes:
This fixes an issue with estore integration - we were previously using mock ukefDealId's and ukefFacilityId's. 

Using mock UKEF ids causes the estore API calls to fail because the same id is being used multiple times and estore folders are not created - which then fails the "submit deal" api call.

In the E2E tests, the `supplier-name` is now dynamically generated with the deal id we get back from number generator. `supplier-name` is used for the estore site name, so this should also not be duplicated.


e2e-tests/trade-finance-manager/cypress/support/reference-data-api/api.js
```diff
@@ -0,0 +1,18 @@
+const api = () => {
+  const url = `${Cypress.config('referenceDataApiProtocol')}${Cypress.config('referenceDataApiHost')}:${Cypress.config('referenceDataApiPort')}`;
+  return url;
+};
+
+module.exports.getIdFromNumberGenerator = (entityType) => {
+  console.log('getIdFromNumberGenerator::');
```
pricea1, line 7: we can remove the console.log

pricea1 (approved):
LGTM apart from stray console.log

PR #141, opened 2021-07-01
feat: adds retry error state for address api errors

pbradley-f4:


gef-ui/server/controllers/exporters-address/index.test.js
```diff
@@ -175,20 +184,62 @@ describe('Validate Exporters Address', () => {
...
+    api.getApplication = () => Promise.resolve(mockApplicationResponse);
+    api.getExporter = () => Promise.resolve(mockExporterResponse);
+    api.getAddressesByPostcode = () => Promise.reject(mockedRejection);
+
+    await validateExportersAddress(mockRequest, mockResponse);
+    expect(mockResponse.render).toHaveBeenCalledWith(
+      'partials/exporters-address.njk',
+      {
+        errors: {
+          errorSummary: [
```
davehlewis, line 223: I'd consider mocking the response of this personally but not important

davehlewis (approved):
LGTM

eddieleeper (approved):

gef-ui/server/controllers/exporters-address/index.test.js
pbradley-f4: fair point, but for now I think its ok to just have a static object to test against

davehlewis: yeah happy with that - just a minor thing really. Can just keep some of these objects simpler

PR #149, opened 2021-07-09
DTFS2-4240 & 4241 - update tfm task email templates

ttbarnes:
2 TFM task emails were using a unique template - updated to use the same generic template and variables as all other tasks.

Spit out `generateTaskUrl` and `generateTaskEmailVariables` functions.


pricea1 (approved):
LFTM

PR #160, opened 2021-07-19
DTFS2-4338: add draft as default status

davehlewis:
- added draft status
- changed application model to make draft the default status


eddieleeper (approved):
LGTM!

PR #163, opened 2021-07-20
DTFS-4120: temporary disable GEF aggregation

davehlewis:
Temporarily removes GEF aggregation from DB query as not supported by cosmos


pricea1 (approved):
👍

PR #167, opened 2021-07-21
Removed e2e workflow tests

pricea1:
Workflow is no longer required so removing e2e tests specifically testing the workflow interface


ttbarnes:
This is amazing, best PR ever

PR #171, opened 2021-07-22
fix: replace node-sass with dart-sass

davehlewis:
Replaces Ruby based node-sass with dart-sass as recommended by SASS themselves. Reduces need for node-gyp


ttbarnes (approved):
LGTM. Assuming dart-sass doesn't need to be in package.json?

davehlewis:
> LGTM. Assuming dart-sass doesn't need to be in package.json?

That Sass package runs on Dart Sass so is not needed directly

PR #178, opened 2021-07-23
DTFS2-4524 - add submission details/fields to GEF deal

ttbarnes:
- dealType
- submissionCount
- submissionType

Submission type is added to the deal when cover terms is decided.


eddieleeper (approved):
Nice 👍

PR #179, opened 2021-07-23
Fix/DTFS2-3724 - TFM - assign underwriter or underwriter managers as lead underwriter

ttbarnes:
- add underwriters as "assign to" options in "assign lead underwriter" form
- sort options alphabetically


eddieleeper (approved):
LGTM!

PR #182, opened 2021-07-26
fixes to dev environment to play nicely with windows

davehlewis:
- moves env variables to .env files
- creates safe .env.sample files to get people started (no secrets)
- updates .gitignore for .env files
- adds .env files to docker compose
- automatic windows linting fixes


gef-ui/docker-compose.yml
```diff
@@ -15,6 +15,7 @@ services:
       - ../portal/styles:/app/styles:ro
       - ../portal/templates:/app/templates:ro
       - ../portal/server:/app/server:ro
+      - ./portal/.env:/app/.env:ro
```
pricea1, line 18: Should this be ../portal ?

gef-ui/docker-compose.yml
```diff
@@ -66,6 +67,7 @@ services:
       - "5001:5001"
     volumes:
       - ../portal-api/src:/app/src:ro
+      - ./portal-api/.env:/app/.env:ro
```
pricea1, line 70: Should this be ../portal-api ?

docker-compose.yml
```diff
@@ -1,258 +1,261 @@
...
-      - AZURE_PORTAL_STORAGE_ACCOUNT
-      - AZURE_PORTAL_STORAGE_ACCESS_KEY
-      - AZURE_PORTAL_FILESHARE_NAME
-      - AZURE_PORTAL_EXPORT_FOLDER
-      - DTFS_PORTAL_SCHEDULER
-      - GOV_NOTIFY_API_KEY
-      - GOV_NOTIFY_EMAIL_RECIPIENT
-
-    entrypoint: npx nodemon src/index.js
+version: "3"
```
pricea1, line 1: Just wondering why the whole file has been marked as a diff

pricea1 (commented):
I think the paths in docker-compose.yml need addressing but otherwise looks good to me.

gef-ui/docker-compose.yml
davehlewis: Thanks, have resolved these

docker-compose.yml
davehlewis: This was the wonderful world of windows line endings. Despite setting up my environment properly it converted these to CRLF. Resolved now. Will watch for it

pbradley-f4 (dismissed):

portal-api/Dockerfile
```diff
@@ -2,39 +2,17 @@ FROM node:lts AS image
 
 WORKDIR /app
 
-ADD package.json .
+ADD . .
```
eddieleeper, line 5: This doesn't look right to me.

reference-data-proxy/Dockerfile
```diff
@@ -2,38 +2,17 @@ FROM node:lts AS image
 
 WORKDIR /app
 
-ADD package.json .
+ADD . .
```
eddieleeper, line 5: As per previous comment - doesn't seem right.

portal-api/Dockerfile
davehlewis: how so? This just adds all the files from the package into the dockerfile. Previously only the package file was being copied, then other files selectively. Is there something better we should be doing?

pricea1: I think this should be ok as node_modules and dist folders that the docker image would generate are in .gitignore so the local ones won't conflict with the docker ones. There will be some files in docker image that docker doesn't need but I don't think that matters for local dev environment.

pricea1 (commented):
Shouldn't we also change the dockerfiles for tfm-ui, tfm-api, dtfs-central-api?

ttbarnes:
For the moved .env file - `portal-api/.env` to `portal-api/.env.sample`, should we wipe the values or make them dummy? This is what is has: https://github.com/foundry4/dtfs2/blob/a2bd84aa3e025371c37ce9d91627d469887103c0/portal-api/.env.sample

davehlewis:
> Shouldn't we also change the dockerfiles for tfm-ui, tfm-api, dtfs-central-a

> For the moved .env file - `portal-api/.env` to `portal-api/.env.sample`, should we wipe the values or make them dummy? This is what is has: https://github.com/foundry4/dtfs2/blob/a2bd84aa3e025371c37ce9d91627d469887103c0/portal-api/.env.sample

I think these are fine because no API keys or anything. Not the full data you need for it to work.

pbradley-f4 (approved):

ttbarnes (approved):

ttbarnes (approved):

PR #184, opened 2021-07-26
fix - better handle Mulesoft exchange rate api fail

pricea1:
Fix to prevent hanging on Mulesoft currency exchange API fail.

-- Ignore the branch name, forgot to change it before pushing. It's nothing to do with Companies House API


ttbarnes (approved):

davehlewis (approved):
LGTM

PR #191, opened 2021-07-28
fix: map GEF V3 status to V2 UI labels for Portal

pbradley-f4:


portal/templates/_macros/status-label.njk
```diff
@@ -0,0 +1,19 @@
...
+    UKEF_IN_PROGRESS: "In progress by UKEF",
+    UKEF_ACCEPTED_CONDITIONAL: "Accepted by UKEF (with conditions)",
+    UKEF_ACCEPTED_UNCONDITIONAL: "Accepted by UKEF (without conditions)",
+    UKEF_DECLINED: "Rejected by UKEF",
+    EXPIRED: "Expired",
+    WITHDRAWN: "Withdrawn"
+  }
+%}
+
+{% macro getLabel(status) %}
```
davehlewis, line 17: Would this not be better as a filter than a macro? much easier to use via pipe function I think

pbradley-f4: I can change it if you feel strongly enough ... just say the word

davehlewis: Given the reuse elsewhere definitely feels the best way - sorry to be annoying about it

pbradley-f4: its cool .... will make it so

davehlewis (approved):
LGTM

PR #193, opened 2021-07-28
DTFS2-4412

eddieleeper:
Remove api calls.

### DIFF
--- a/portal-api/src/v1/gef/controllers/application-submit.js
+++ b/portal-api/src/v1/gef/controllers/application-submit.js
@@ -3,7 +3,6 @@ const {
   getAllFacilitiesByApplicationId,
   update: updateFacility,
 } = require('./facilities.controller');
-const refDataApi = require('../../../reference-data/api');
 
 const generateSubmissionData = async (existingApplication) => {
   const result = {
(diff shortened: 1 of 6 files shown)


ttbarnes (dismissed):

portal-api/src/v1/gef/controllers/application-submit.js
```diff
@@ -28,7 +28,12 @@ const generateUkefDealId = async (application) => {
   };
 };
 
-const generateUkefFacilityId = (facility) => (facility.ukefFacilityId !== null ? facility.ukefFacilityId : generateId('facility'));
+const generateUkefFacilityId = async (facility) => {
+  const ukefFacilityId = facility.ukefFacilityId !== null ? facility.ukefFacilityId : await generateId('facility');
```
davehlewis, line 32: Could this not be written as `facility.ukeffacilityId || await generateId('facility')` or is there a chance of a legitimate falsey value?

eddieleeper: Thanks @davehlewis - Yes, it probably could - when testing like that I got a failure, so added the explicit check.  I'll have a look at refactoring on the next PR.  This branch is a work in progress for the various steps on DTFS-4412

davehlewis: No problem, can sometimes be a bit odd. If we could amend our eslint I would say nullish coalescing would work move cleanly for it.

eddieleeper: Updated.

ttbarnes (approved):

portal-api/src/v1/gef/controllers/application-submit.js
```diff
@@ -52,7 +50,7 @@ const addUkefFacilityIdToFacilities = async (applicationId) => {
   const facilities = await getAllFacilitiesByApplicationId(applicationId);
 
   facilities.forEach(async (facility) => {
-    const ukefFacilityId = await generateUkefFacilityId(facility);
+    const ukefFacilityId = generateUkefFacilityId(facility);
```
davehlewis, line 53: why has the async been removed from these?

PR #198, opened 2021-07-30
Feature/dtfs2 3968 return maker

davehlewis:
https://ukef-dtfs.atlassian.net/browse/DTFS2-3968
- Adds return to maker page
- Adds changes required status to GEF-UI and Portal API


gef-ui/server/controllers/return-to-maker/index.js
```diff
@@ -0,0 +1,66 @@
...
+export const MAX_COMMENT_LENGTH = 400;
+
+export const getReturnToMaker = async (req, res) => {
+  const { params } = req;
+  const { applicationId } = params;
+  const { status } = await getApplication(applicationId);
+
+  if (status !== PROGRESS.BANK_CHECK) {
+    console.log('Incorrect status or permissions, redirecting to dashboard');
+    return res.redirect('/dashboard/gef/');
```
ttbarnes, line 16: Should this maybe go to the deal/application route instead?

ttbarnes (approved):
Beautiful

gef-ui/server/controllers/return-to-maker/index.js
davehlewis: Can do - nothing specified in the ticket and was trying to avoid issue with too many redirects (this to application to dashboard or wherever)

ttbarnes: No biggie, just an observation - suppose this is a bigger piece of UX to do in the future e.g maybe nice set of error pages like "you don't have access", "unable to do this" etc

ttbarnes (approved):

PR #199, opened 2021-07-30
DTFS2-4545 - disable fetchWorkflowTypeB in portal scheduler jobs

ttbarnes:


pricea1 (commented):
As this is the only job in scheduler we can switch it off by setting the env var DTFS_PORTAL_SCHEDULER='false'

or we could remove the initScheduler() from createApp.js

davehlewis (approved):
LGTM

ttbarnes:
@pricea1 i'll update the environment variable. I think it's good to keep the workflow schedular task commented out though, just incase it needs to be renabled.

ttbarnes:
Environment variable that enables the schedular now has false flag, so this PR is no longer needed.

pricea1:
Odd its missing. If it's missing then the scheduler doesn't run. It only runs if DTFS_PORTAL_SCHEDULER=true.
The Azure dev portal-api has it set to false

PR #203, opened 2021-08-03
dtfs2-4239 Sending acknowledgement email using updated deal rather th…

pricea1:
…an sent deal.

Notify formatting changes


ttbarnes (approved):
👍

PR #205, opened 2021-08-04
feat: adds validation to cover start date

pbradley-f4:


gef-ui/server/controllers/about-facility/index.js
```diff
@@ -55,30 +57,53 @@ const validateAboutFacility = async (req, res) => {
...
+          });
+        }
+      }
+
+      if (!coverEndDateDay || !coverEndDateMonth || !coverEndDateYear) {
+        aboutFacilityErrors.push({
+          errRef: 'coverEndDate',
+          errMsg: 'Enter a cover end date',
+        });
+      }
```
ttbarnes, line 106: I think need to add some validation for day/month/year missing, for GDS. i.e "X must include a year": https://design-system.service.gov.uk/components/date-input/#dates-with-errors-example

e2e-tests/gef/cypress/integration/about-facility.spec.js
```diff
@@ -1,12 +1,15 @@
...
 /* eslint-disable no-undef */
+import moment from 'moment';
 import relative from './relativeURL';
 import aboutFacility from './pages/about-facility';
 import CREDENTIALS from '../fixtures/credentials.json';
 
 const applications = [];
 let token;
 
+const now = moment();
```
ttbarnes, line 11: should we be using date-fns for this?

gef-ui/server/controllers/about-facility/index.js
pbradley-f4: Ive added this

e2e-tests/gef/cypress/integration/about-facility.spec.js
pbradley-f4: Im leaving the cypress piece for now as it uses moment everywhere

ttbarnes (approved):

PR #206, opened 2021-08-04
DTFS2-4361 - TFM submission mapping BSS/EWCS deals into generic structure

ttbarnes:
This allows BSS/EWCS and GEF deals to use the same data structure when calling apis from TFM submit controller.

E.g instead of this:
```js
deal.dealSnapshot.submissionDetails.supplyContractCurrency
deal.dealSnapshot.details.submissionType
deal.dealSnapshot.bondTransactions.items[0]
deal.dealSnapshot.loanTransactions.items[0]
```

deal.submit has a simplified, generic mapping like this:
```js
deal.dealCurrency
deal.submissionType
deal.facilities[]
```

This removes the need to do dealType / field checking and multiple facility iterations in each api call.

The mappings are in `mappings/map-submitted-deal/` directory.

This is also a step towards a unified, generic data model that should be implemented in the future.

### Other necessities
- move default tfm object into default constants
- add `dealType` to all BSS/EWCS deals used in submit-to-tfm e2e tests
- add `dealType` to BSS/EWCS deals on creation
- rename tfm's `findOneDeal` controller to `findOneTfmDeal`
- remove `sendAinMinIssuedFacilitiesAcknowledgementByDealId`, can use `sendAinMinIssuedFacilitiesAcknowledgement` now without need to find a deal by id

### Next steps (for context)
- seperate PR to update facility fields into generic structure
- seperate PR to map GEF fields into the generic structure
- ensure submission api calls work with GEF data


trade-finance-manager-api/api-tests/v1/deals/deals-submit.facilities.api-test.js
```diff
@@ -46,14 +46,14 @@ describe('/v1/deals', () => {
 
         expect(status).toEqual(200);
 
-        const bond = body.dealSnapshot.bondTransactions.items[0];
+        const bond = body.facilities.find(({ facilityType }) => facilityType === 'bond');
```
pricea1, line 49: Should we use the CONSTANT for facilityType instead?

trade-finance-manager-api/src/v1/__mocks__/mock-deal-MIA-second-submit-facilities-unissued-to-issued.js
```diff
@@ -194,6 +194,7 @@ const MOCK_DEAL_AIN_SUBMITTED_FACILITIES_UNISSUED_TO_ISSUED = {
         'coverEndDate-year': '2020',
         status: 'Submitted',
         ukefFacilityID: '65431',
+        requestedCoverStartDate: 1610369832226.0,
```
pricea1, line 197: We shouldn't hard code the timestamp. This timestamp is for 4/08/2021. Once this date passes all the tests that require requestedCoverStartDate to not be in the past will fail

Maybe change to `Date.now()`

trade-finance-manager-api/src/v1/controllers/update-facilities.js
```diff
@@ -11,66 +11,47 @@ const updateFacilities = async (deal) => {
...
+      ...facilityCurrencyConversion,
+      ...facilityExposurePeriod,
+      facilityGuaranteeDates,
+      riskProfile: DEFAULTS.FACILITY_RISK_PROFILE,
+      premiumSchedule: facilityPremiumSchedule,
+    };
+
+    const updateFacilityResponse = await api.updateFacility(facilityId, facilityUpdate);
+
+    const updatedFacility = {
```
pricea1, line 43: Does updateFacilityResponse return the complete facility or just whats been updated?
Is there need to reconstruct it again?

trade-finance-manager-api/src/v1/controllers/update-issued-facilities.js
```diff
@@ -11,115 +11,97 @@ const updatedIssuedFacilities = async (deal) => {
...
+      facilityStage,
+      previousFacilityStage,
+      hasBeenAcknowledged,
+    } = modifiedFacility;
+
+    // we only need to update issued facilities if:
+    // - a facility has changed from `Unissued` to `Issued` (`Conditional` to `Unconditional` for a loan)
+    // - and the facility changing from `Unissued` to `Issued` has NOT been previously acknowledged
+
+    const bondIsIssued = ((previousFacilityStage === CONSTANTS.FACILITIES.FACILITY_STAGE_PORTAL.UNISSUED)
```
pricea1, line 32: There's a helper for determining if a facility is issued
https://github.com/foundry4/dtfs2/blob/master/trade-finance-manager-api/src/v1/helpers/issued-facilities.js

pricea1 (commented):
I've added a few comments

trade-finance-manager-api/src/v1/__mocks__/mock-deal-MIA-second-submit-facilities-unissued-to-issued.js
ttbarnes: I should have learnt from before 😂 i'll update

trade-finance-manager-api/src/v1/controllers/update-facilities.js
ttbarnes: I'll look into this, can probably be simplified

trade-finance-manager-api/src/v1/controllers/update-issued-facilities.js
ttbarnes: Could even add to the facility mappings `isIssued` - GEF facilities already have this. I'll update this function to use the `issued-facilities` function for now - next PR will be for facilities

ttbarnes: The logic in update-issued-facilities only applies to facilities that have been previously unissued and then changed to issued, so existing helper would return false positives. Created a new helper: 'was-previously-unissued' and used with 'is-issued', a lot cleaner 👍

trade-finance-manager-api/src/v1/controllers/update-facilities.js
ttbarnes: Simplified this a lot here: https://github.com/foundry4/dtfs2/pull/206/commits/0e3614fa177cdb4bfe50e8a184ebcfd495f698c2

pbradley-f4 (dismissed):

pbradley-f4 (approved):

PR #218, opened 2021-08-09
DTFS2-4361 - central API generating TFM facility snapshots for a GEF deal

ttbarnes:
- central API deal submission using `dealId` and `dealType` in the request body instead of params
- central API endpoints to get and create GEF deals
- central API endpoints to get and create GEF facilities 
- generate TFM facility snapshots from a GEF deal
- add test coverage for central API submit endpoint
- strip facility value commas in mappings instead of currency conversion api call


pricea1 (approved):
What about GEF deal/facility updating? Is that part of the ticket?
Otherwise LGTM

ttbarnes:
@pricea1 Not yet - the ticket is a massive job so i'm breaking it down into different PRs to avoid huge PRs. This is the 4th one I think 😄

The next PR will be for getting the submission API calls to work with GEF data. One of the calls is updating Gef deal status so will tackle then.

PR #219, opened 2021-08-10
dtfs2-3793 Azure infrastructure changes for Number Generator Function

pricea1:
Added Azure function services ahead of Number Generator azure function PR


eddieleeper (approved):
LGTM!

PR #222, opened 2021-08-10
DTFS2-4579 - mock BSS deals & facilities for user research

ttbarnes:
The most interesting changes are here:
- `utils/mock-data-loader/bss/dates.js` (used to create dates in the mock facilities)
- `utils/mock-data-loader/insert-mocks.js`

Also changed the mocks directory structure so that BSS & GEF are in their own directories.


pricea1 (dismissed):
LGTM

PR #226, opened 2021-08-11
Regenerate mock-data-loader package-lock.json

ttbarnes:
Was previously generated with a newer version of node.


eddieleeper (approved):
Works for me without warnings now.

PR #228, opened 2021-08-12
DTFS2-4361 - wait for all facility api calls on GEF submission

ttbarnes:
Without this, TFM submit can receive facilities without the ukef ids.


eddieleeper (approved):
Nice! 👍

PR #231, opened 2021-08-13
feat: DTFS2-3959 add validation for save and return

pbradley-f4:


gef-ui/server/controllers/about-facility/index.js
```diff
@@ -50,28 +50,45 @@ const validateAboutFacility = async (req, res) => {
   const coverStartDateDay = body['cover-start-date-day'];
   const coverStartDateMonth = body['cover-start-date-month'];
   const coverStartDateYear = body['cover-start-date-year'];
+  const coverStartDateIsFullyComplete = coverStartDateDay && coverStartDateMonth && coverStartDateYear;
```
pricea1, line 53: Is coverStartDateMonth zero indexed i.e. Does Jan = 0? Will coverStartDateIsFullyComplete fail if coverStartDateMonth = 0?

pricea1 (commented):
LGTM but have one query.

gef-ui/server/controllers/about-facility/index.js
pricea1: Actually, scrap that. I can see it isn't as you use coverEndDateMonth - 1 to set a date object further down.

pricea1 (approved):
👍

PR #233, opened 2021-08-16
Portal UI server logs for missing session secret fix small TFM mapping bug

ttbarnes:
- add logs for when session secret env var is missing
- remove `SESSION_SECRET` crypto fallback as it doesn't work
- Also fixed TFM facility currency mapping so it only maps currencyCode if currency object exists.

### DIFF
--- a/gef-ui/server/index.js
+++ b/gef-ui/server/index.js
@@ -5,7 +5,6 @@ import redis from 'redis';
 
 import flash from 'connect-flash';
 import path from 'path';
-import crypto from 'crypto';
 import json2csv from 'express-json2csv';
 import './azure-env';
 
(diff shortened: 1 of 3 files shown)


gef-ui/server/index.js
```diff
@@ -20,49 +19,54 @@ const app = express();
...
-const SESSION_SECRET = process.env.SESSION_SECRET || crypto.randomBytes(256 / 8).toString('hex');
+if (!process.env.SESSION_SECRET) {
+  console.error('GEF UI server - SESSION_SECRET missing');
+}
+
+if (!process.env.REDIS_HOSTNAME) {
+  console.error('GEF UI server - REDIS_HOSTNAME missing');
+}
+
+if (!process.env.REDIS_KEY) {
```
pricea1, line 30: I think REDIS_KEY is optional an only needed for Azure envs. Works fine locally if not configured.

gef-ui/server/index.js
```diff
@@ -20,49 +19,54 @@ const app = express();
 
 const PORT = process.env.PORT || 5006;
 
-// Fail-safe fallback to a 256-bit random value:
-const SESSION_SECRET = process.env.SESSION_SECRET || crypto.randomBytes(256 / 8).toString('hex');
+if (!process.env.SESSION_SECRET) {
+  console.error('GEF UI server - SESSION_SECRET missing');
+}
+
+if (!process.env.REDIS_HOSTNAME) {
```
pricea1, line 26: We also get a connection error on line 53 is REDIS_HOSTNAME is invalid. Do we need this too?

portal/server/index.js
```diff
@@ -21,47 +20,52 @@ const app = express();
 
 const PORT = process.env.PORT || 5000;
 
-// Fail-safe fallback to a 256-bit random value:
-const SESSION_SECRET = process.env.SESSION_SECRET || crypto.randomBytes(256 / 8).toString('hex');
+if (!process.env.SESSION_SECRET) {
```
pricea1, line 23: Same comments as gef-ui

pricea1 (commented):
A couple of queries.

ttbarnes:
Good points @pricea1 i've removed the console logs for missing Redis env vars. So now the server changes is just logging out if the session secret is missing, and simplifies some conditions.

pricea1 (approved):
LGTM

PR #235, opened 2021-08-16
DTFS2-4239 - fix tfm submission email formatting issues with facilities

ttbarnes:
- some facility fields used in the email template were missing after deal submission field mappings were introduced
- only populate "unissued facilities" email section if there are unissued facilities

### DIFF
--- a/trade-finance-manager-api/src/v1/helpers/notify-template-formatters.js
+++ b/trade-finance-manager-api/src/v1/helpers/notify-template-formatters.js
@@ -14,51 +14,47 @@ const generateFacilitiesListString = (facilities) => facilities.reduce((acc, fac
 }, '');
 
 const generateBSSListString = (facilities) => {
-  if (facilities.length) {
-    const bssList = facilities.reduce((acc, facility) => {
-      const {
-        bondType, ukefFacilityID, bankReferenceNumber, uniqueIdentificationNumber,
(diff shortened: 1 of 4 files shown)


trade-finance-manager-api/src/v1/helpers/notify-template-formatters.js
```diff
@@ -14,37 +14,51 @@ const generateFacilitiesListString = (facilities) => facilities.reduce((acc, fac
...
-    const bankReference = uniqueIdentificationNumber || bankReferenceNumber;
-    const bankRefString = bankReference
-      ? `*Your bank ref: ${bankReference}\n`
-      : '';
-
-    return `${acc}*${bondType}\n${bankRefString}*UKEF facility ID: ${ukefFacilityID} \n\n`;
-  }, '');
-  const bssHeading = bssList ? '#Bond Support Scheme\n\n' : '';
-  return `${bssHeading}${bssList}`;
+  if (facilities.length) {
```
pricea1, line 17: I'm not sure why we need a facilities.length check. If its an empty array then the facilties.reduce would return ''

ttbarnes:
Thanks @pricea1 a lot of those changes were not needed. Should be good now

eddieleeper (approved):
LGTM!

PR #236, opened 2021-08-16
Feature/dtfs2 3793 number generator function

pricea1:


pricea1:
Cloned into https://github.com/foundry4/dtfs2/pull/266

PR #237, opened 2021-08-16
dtfs2-3793 Number Generator function - infrastructure

pricea1:
Create infrastructure across all environments for number generator Azure function

### DIFF
--- a/.github/workflows/env_demo.yml
+++ b/.github/workflows/env_demo.yml
@@ -785,7 +785,7 @@ jobs:
   # END Redis setup
 
   # START Function setup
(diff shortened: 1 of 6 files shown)


ttbarnes:
I don't understand most of this but it looks good to me. I think it's a copy of other azure functions as well. Happy to approve when conflict is resolved

ttbarnes (commented):
I don't understand most of this... I think it's a copy of other azure functions. Happy to approve when conflict is resolved

ttbarnes (approved):

PR #241, opened 2021-08-17
DTFS2-4380 - central api bank endpoints, TFM mapping bank in GEF deal

ttbarnes:
- central API get & create bank
- fetch and return bank by id when getting a GEF deal
- map bank name in UI graphql reducer

### DIFF
--- a/dtfs-central-api/src/v1/controllers/portal/deal/get-deal.controller.js
+++ b/dtfs-central-api/src/v1/controllers/portal/deal/get-deal.controller.js
@@ -3,6 +3,7 @@ const db = require('../../../../drivers/db-client');
 const CONSTANTS = require('../../../../constants');
 const { findAllGefFacilitiesByDealId } = require('../gef-facility/get-facilities.controller');
 const { findOneExporter } = require('../gef-exporter/get-gef-exporter.controller');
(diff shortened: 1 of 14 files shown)


dtfs-central-api/src/v1/controllers/portal/deal/get-deal.controller.js
```diff
@@ -95,10 +96,15 @@ const findOneGefDeal = async (_id, callback) => {
   if (deal) {
     deal.facilities = await findAllGefFacilitiesByDealId(_id);
```
eddieleeper, line 97: Low priority - These calls could be done in parallel

eddieleeper (approved):
Just an observation that the apis in this service don't have any authentication requirements.

dtfs-central-api/src/v1/controllers/portal/deal/get-deal.controller.js
ttbarnes: Good point, they'll be another call we need to add here so i'll do this in another PR

ttbarnes:
RE Authentication - I suspect the central API might be locked down in Azure @pricea1 ?

PR #242, opened 2021-08-18
DTFS2-4382 - TFM mapping GEF facility product & type

ttbarnes:
- renders 'GEF', 'X facility', and 'General Export Facility' in the UI, for GEF facilities
- clean up api tests to use constants
- remove duplicate component tests for facility.njk - tests already exist for facility overview


pricea1 (approved):
LGTM

PR #247, opened 2021-08-18
Docs: update env samples and readme for usage

davehlewis:
- Adds all known keys for each application to sample file. Sensitive data is left blank
- Updates readme to use `.env` files rather than secrets scripts
- Secret scripts not removed to avoid possible issues with workflows for now


ttbarnes (approved):
LGTM

PR #248, opened 2021-08-18
feature/DTFS2-4187: adds ability to Abandon a deal

pbradley-f4:


gef-ui/templates/application-abandon.njk
```diff
@@ -0,0 +1,61 @@
...
+        You’ll still be able to find this {% if application.isAutomaticCover %} notice {% else %} application {% endif %} in the abandoned folder. 
+      </p>
+    </div>
+
+    <div class="govuk-grid-column-one-half-from-desktop">
+      <form
+        method="post"
+        data-cy="form">
+
+        <p class="govuk-button-group govuk-!-margin-top-7" style="vertical-align: middle">
```
ttbarnes, line 40: is the inline style needed?

gef-ui/server/controllers/application-abandon/index.test.js
```diff
@@ -0,0 +1,177 @@
...
+  ],
+});
+
+const MockFacilityResponse = () => {
+  const res = {};
+  res.status = 'IN_PROGRESS';
+  res.data = [];
+  return res;
+};
+
```
ttbarnes, line 75: Not for this PR, seems to be a lot of duplication of the mock responses, would be nice to move into one shared file

ttbarnes (dismissed):
non blocking comments otherwise LGTM

gef-ui/server/controllers/application-abandon/index.test.js
pbradley-f4: yeah I will refactor that later when we have time

gef-ui/templates/application-abandon.njk
pbradley-f4: good catch ..  I think that was leftover from when it was a link. button-group resolves this so will remove

ttbarnes (approved):

PR #255, opened 2021-08-20
Update README.md

ttbarnes:


README.md
```diff
@@ -22,31 +22,36 @@ Prerequisites
  * GovUK and MOJ design systems
  * Nunjucks (UI templates)
 
+### Getting started
+
+* Clone this repo
+* Run `nvm install` to ensure you are using the correct version of node
+* Create `.env` files for each service. You can use `.env.sample` as a base. Some sensitive variables need to be shared from the team
+* Generate JWT keypairs with `secrets/set_jwt_keypair.sh`
```
IgnatG, line 30: on Windows you may need to run it as `bash secrets/set_jwt_keypair.sh`

ttbarnes: updated

README.md
```diff
@@ -22,31 +22,36 @@ Prerequisites
...
 
+### Getting started
+
+* Clone this repo
+* Run `nvm install` to ensure you are using the correct version of node
+* Create `.env` files for each service. You can use `.env.sample` as a base. Some sensitive variables need to be shared from the team
+* Generate JWT keypairs with `secrets/set_jwt_keypair.sh` (`bash secrets/set_jwt_keypair.sh` for Windows)
+* Base64 encode the generated public and private keys and declare as environment variables in your terminal. E.g:
+  * `export JWT_SIGNING_KEY=1234`
+  * `export JWT_VALIDATING_KEY=1234`
```
IgnatG, line 33: this should work. However I'm not sure if you need to run the same commands again if you close the current terminal. 

So as an alternative, one can add the `JWT_SIGNING_KEY` and `JWT_VALIDATING_KEY` to the `.env ` file and then populate them with the encoded (using [base64encode](https://www.base64encode.org/))  `private.pem` and `public.pem`.

ttbarnes: I think will need to run them again if the terminal is closed. I thought we tried adding to the `.env` file, but it didn't work on your machine - could you try again and confirm?

IgnatG: it worked from `.env` file. the only one that was problematic was the `SESSION_SECRET`

ttbarnes: OK great i'll update this PR. I just had the issue with `SESSION_SECRET` after closing my terminal. Has to be set as env var in the terminal rather than .env

ttbarnes: updated and marking as resolved as it's out of date now

IgnatG (approved):

PR #260, opened 2021-08-24
DTFS2-4519 Facility value page save and return, add page for repayment frequency

eddieleeper:


gef-ui/server/controllers/facility-value/facility-value.js
```diff
@@ -0,0 +1,34 @@
...
+function validateFacilityValue({ interestPercentage, coverPercentage }, saveAndReturn = false) {
+  const facilityValueErrors = [];
+  // Regex tests to see if value between 1 and 80
+  const oneToEightyRegex = /^(?:[1-9]|[1-7][0-9]|80)$/;
+  // Regex tests to see if value is between 0 and 100. Also allows for decimal places ie. 20.1
+  const zeroToOneHundredRegex = /^(\d{1,2}(\.\d{1,2})?|100(\.00?)?)$/;
+
+  const coverPercentageError = {
+    errRef: 'coverPercentage',
+    errMsg: 'You can only only enter a number between 1 and 80',
```
IgnatG, line 11: not a issue, but the word 'only' appears twice

gef-ui/server/controllers/facility-value/facility-value.test.js
```diff
@@ -0,0 +1,114 @@
+import validateFacilityValue from './facility-value';
+
+const interestErrorExpected = {
+  errMsg: 'You can only enter a number between 0 and 100',
+  errRef: 'interestPercentage',
+};
+const coverErrorExpected = {
+  errMsg: 'You can only only enter a number between 1 and 80',
```
IgnatG, line 8: not a issue, but the word 'only' appears twice

IgnatG (approved):

PR #302, opened 2021-09-02
DTFS2-4633 - enable estore API calls for GEF. Fix TFM mapping bug

ttbarnes:
Also fixed a bug where TFM was mapping non existent correspondence address, which causes graphQL query to fail.


davehlewis (approved):
LGTM

PR #306, opened 2021-09-06
DTFS2-4727 - only call estore if BSS

ttbarnes:
- remove incomplete exporters from GEF data. Can cause issues with estore
- in GEF submit-to-ukef e2e test, use a deal with only 1 facility
- disable estore API call until it doesn't error (pending Mulesoft IT team)

### DIFF
--- a/e2e-tests/gef/cypress/integration/submit-to-ukef.spec.js
+++ b/e2e-tests/gef/cypress/integration/submit-to-ukef.spec.js
@@ -5,7 +5,7 @@ import submitToUkef from './pages/submit-to-ukef';
 import submitToUkefConfirmation from './pages/submit-to-ukef-confirmation';
 import CREDENTIALS from '../fixtures/credentials.json';
 
-const applicationIds = [];
+let applicationId;
 
 context('Submit to UKEF', () => {
   before(() => {
@@ -16,10 +16,9 @@ context('Submit to UKEF', () => {
         cy.apiFetchAllApplications(token);
       })
       .then(({ body }) => {
-        body.items.forEach((item) => {
-          applicationIds.push(item._id);
-        });
+        applicationId = body.items[0]._id;
       });
+
     cy.login(CREDENTIALS.CHECKER);
 
     cy.on('uncaught:exception', () => false);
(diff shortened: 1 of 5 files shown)


IgnatG (approved):

pbradley-f4 (approved):
:+1:

e2e-tests/gef/cypress/integration/submit-to-ukef.spec.js
```diff
@@ -16,18 +16,17 @@ context('Submit to UKEF', () => {
         cy.apiFetchAllApplications(token);
       })
       .then(({ body }) => {
-        body.items.forEach((item) => {
-          applicationIds.push(item._id);
-        });
+        applicationId = body.items[0]._id;
```
davehlewis, line 19: The previous test uses the third item in the array - is this a problem?

trade-finance-manager-api/src/v1/api.js
```diff
@@ -499,7 +499,7 @@ const createEstoreFolders = async (eStoreFolderInfo) => {
     });
     return response.data;
   } catch (err) {
-    return err;
+    return false;
```
davehlewis, line 502: needs some degree of error handling here it feels like - at least a log

ttbarnes: There is a log but not in this API. This endpoint calls reference data - which then uses estore controller. The Estore controller logs out errors

davehlewis: ok makes sense

e2e-tests/gef/cypress/integration/submit-to-ukef.spec.js
ttbarnes: I don't think so. The mock deals are almost exactly the same. The indexing in mock data loader is weird.... Probably don't need this change now.. Either way it doesn't seem to make a difference to the test now along with the other changes. I'd vote to not have this as a blocker and get this merged so we can progress other PRs

davehlewis: Sounds good to me

davehlewis (approved):

PR #308, opened 2021-09-06
DTFS2-4409 - add userId and submissionType to mock GEF deals. Fix mapping bug in tasks 

ttbarnes:
- add userId and submissionType to mock GEF deals (allows for quicker dev work when submitting to TFM)
- TFM tasks: fix submissionType mapping for BSS/GEF


eddieleeper (approved):
LGTM!

PR #317, opened 2021-09-08
Fix/DTFS2 4726

pbradley-f4:
Fixes 
- Cover type validation
- Save and return on currency page
- Block ability to add cover value without currency


davehlewis (approved):
LGTM

PR #326, opened 2021-09-10
DTFS2-4792 - Update MongoDB

ttbarnes:
### Package updates

- Update mongoDB to 4.0
- Update jest
- Remove mongoose
- Add node-notifier that's a required dep from another package (...)
- Add bson package

### Mongo/code updates

- Remove `numberOfRetries` from db config, no longer supported
- Import ObjectID from `bson` package instead of mongodb, no longer exported
- Update instances of `findOneAndUpdate`. `{ returnOriginal: false }` no longer returns the updated document. Replaced with `{ returnDocument: 'after', returnOriginal: false },`
- Update instances of `insertOne` where the returned document would be used. Mongo no longer returns the document, instead returning a simple object with the created id. Therefore replaced all instances of `response.ops[0]` with `response.insertedId`, returning the `insertedId` as `_id` in responses.
- Portal API (BSS & GEF) Deals and facilities - when creating deals and facilities, do a `findOne`to return the created documents.

### API tests
- Many API tests that had "insert X document" and then asserting that the reponse has the created document, have been updated. Now, inserting a document does not actually return the document, only the ID. A lot of api tests have been changed from e.g this:
```js
const { body } = await api.post('/criteria', documentToInsert)
exepect(body).toEqual(documentToInsert);
```

to this:

```js
const { body: createdDocument, status } = await api.post('/criteria', documentToInsert)
expect(status).toEqual(200);

const { body: document } = await api.get`/criteria/${createdDocument._id}`)
exepect(document).toEqual({
  _id: createdDocument._id},
  documentToInsert,
});
```

### E2E tests
For Portal and "submit to TFM" e2e test, it uses a cypress command `insertManyDeals`. It previously used to just insert a deal with one endpoint and then use the response data in the E2E specs.

This has been updated so that `insertManyDeals` now creates a deal and then gets the deal, so the deal details are returned in the response.

### Miscellaneous
- Removed log controller in portal-api (was used for Workflow integration)
- Mock `node-cron` in TFM API tests (was causing issues after Jest update)
- Set portal UI jest environment to `jsdom`, it was complaining after updating jest
- Remove `--detectOpenHandles` option from jest scripts; was causing issues. Also causes "_significant performance penalty and should only be used for debugging_". Only in some of the watch scripts now.
- fixed flaky api test in central APi


portal-api/api-tests/v1/deals/deal-clone.api-test.js
```diff
@@ -92,14 +92,15 @@ describe('/v1/deals/:id/clone', () => {
...
+        const { body: createdDeal } = await as(anHSBCMaker).post(dealToClone).to('/v1/deals');
+        const originalDealId = createdDeal._id;
         createdFacilities = await createFacilities(anHSBCMaker, originalDealId, completedDeal.mockFacilities);
+
+        const { body: dealAfterCreation } = await as(anHSBCMaker).get(`/v1/deals/${originalDealId}`);
+        originalDeal = dealAfterCreation.deal;
       });
 
-      it('clones a deal with only specific properties in `details`, wipes `comments`, `editedBy` `ukefComments`, `specialConditions`, changes `maker` to the user making the request, marks status `Draft`', async () => {
+      it.only('clones a deal with only specific properties in `details`, wipes `comments`, `editedBy` `ukefComments`, `specialConditions`, changes `maker` to the user making the request, marks status `Draft`', async () => {
```
pbradley-f4, line 103: .only

portal-api/src/v1/controllers/feedback.controller.js
```diff
@@ -89,7 +87,7 @@ exports.create = async (req, res) => {
     emailVariables,
   );
 
-  return res.status(200).send(createdFeedback);
+  return res.status(200).send({ _id: createdFeedback.insertedId });
```
pbradley-f4, line 90: `200` feels wrong here `201` more suitable since were changing the response data IMO

pbradley-f4 (changes requested):
Looks good. Just the `.only` in the test suite to remove and maybe one for another time to look at the status codes we use

portal-api/api-tests/v1/deals/deal-clone.api-test.js
ttbarnes: Good spot! Was from debugging. I've removed this and updated/corrected the other tests here: https://github.com/UK-Export-Finance/dtfs2/pull/326/commits/d309ae92e51b360b5ab14bc6aca061b6587be38d

IgnatG (approved):

davehlewis (approved):

pbradley-f4 (approved):
:+1:

PR #339, opened 2021-09-15
feat: adds ability to edit deal references plus new validation DTFS2-4789

pbradley-f4:


portal-api/api-tests/v1/gef/application.api-test.js
```diff
@@ -209,7 +197,7 @@ describe(baseUrl, () => {
       expect(body).toEqual([{
         errCode: 'MANDATORY_FIELD',
         errRef: 'bankInternalRefName',
-        errMsg: 'Application Reference Name is Mandatory',
+        errMsg: 'bankInternalRefName is Mandatory',
```
IgnatG, line 200: is this `bankInternalRefName` displayed to users? if so, do they know what it means?
I'm thinking that the previous text was more readable (i.e. 'Application Reference Name')

IgnatG (approved):

davehlewis:
I would recommend squashing the commits before merging

portal-api/api-tests/v1/gef/application.api-test.js
pbradley-f4: this msg is just from the api ie a technical message. The UI msg is mapped from the error code returned by the api.

PR #348, opened 2021-09-20
fix/DTFS2-4362 - TFF deal submit email only rendering issued facilities text if there are issued facilities

ttbarnes:
Also added root .env.sample


trade-finance-manager-api/src/v1/helpers/notify-template-formatters.js
```diff
@@ -93,7 +93,10 @@ const generateGefFacilityLists = (facilities) => {
   const unissuedCashList = generateFacilitiesListString(unissuedCash);
   const unissuedContingentList = generateFacilitiesListString(unissuedContingent);
 
-  const issued = `${issuedCashList}\n${issuedContingentList}`;
```
ttbarnes, line 96: This was the issue - it was rendering a string with an empty space, so the rest of the checks thought that there was an issued facilities list because the "length" is 1.

IgnatG (approved):

PR #382, opened 2021-10-01
DTFS2-4029: add additional application get to help with race condition

davehlewis:
- adds call to get application after upload to avoid race condition when uploading multiple files

### DIFF
--- a/gef-ui/server/controllers/supporting-information/manual-inclusion-questionnaire/index.js
+++ b/gef-ui/server/controllers/supporting-information/manual-inclusion-questionnaire/index.js
@@ -89,8 +89,9 @@ const postUploadManualInclusion = async (req, res) => {
     const uploadedFiles = validFiles.length ? await uploadAndSaveToDeal(
       validFiles,
       FIELD_NAME,
-      application,
+      applicationId,
       userToken,
+      user,
       MAX_FILE_SIZE,
     ) : [];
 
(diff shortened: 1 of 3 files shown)


gef-ui/server/controllers/supporting-information/manual-inclusion-questionnaire/index.js
```diff
@@ -172,6 +173,7 @@ const uploadManualInclusion = async (req, res) => {
         [file],
         FIELD_NAME,
         application,
```
pbradley-f4, line 175: Havent updated implemtation to pass id rather than deal?

gef-ui/server/controllers/supporting-information/manual-inclusion-questionnaire/index.js
```diff
@@ -90,6 +90,7 @@ const postUploadManualInclusion = async (req, res) => {
       validFiles,
       FIELD_NAME,
       application,
```
pbradley-f4, line 92: Havent updated implemtation to pass id rather than deal?

pbradley-f4 (changes requested):

gef-ui/server/controllers/supporting-information/manual-inclusion-questionnaire/index.js
davehlewis: Good catch, thanks. Sorry juggling this and errors in numgen

pbradley-f4 (approved):

pbradley-f4 (approved):

PR #383, opened 2021-10-04
DTFS2-4383 - TFM mapping GEF facility "provided on basis" ids and values

ttbarnes:
This was previously just taking e.g `RESOLVING` and changing to title case to render in the UI.

Now it maps with the text so e.g `Amazing resolving basis` will render instead.

### DIFF
--- a/trade-finance-manager-api/src/graphql/reducers/mappings/gef-facilities/mapGefFacilityProvidedOn.js
+++ b/trade-finance-manager-api/src/graphql/reducers/mappings/gef-facilities/mapGefFacilityProvidedOn.js
@@ -1,14 +1,14 @@
 const CONSTANTS = require('../../../../constants');
 
 const mapGefFacilityProvidedOn = (details) =>
-  details.map((basis) => {
+  details.reduce((previousValue, currentValue) => {
     // 'Other basis' should not display in TFM UI.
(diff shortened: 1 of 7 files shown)


trade-finance-manager-api/src/graphql/reducers/mappings/gef-facilities/mapGefFacilityProvidedOn.js
```diff
@@ -1,7 +1,14 @@
-const { titleCaseString } = require('../../../../utils/string');
+const CONSTANTS = require('../../../../constants');
 
 const mapGefFacilityProvidedOn = (details) =>
-  details.map((d) => titleCaseString(d));
+  details.map((basis) => {
```
davehlewis, line 4: Are you sure you want map here and not reduce? Will keep array same length just empty. Might cause unexptected results?

davehlewis (approved):
Single comment otherwise LGTM

trade-finance-manager-api/src/graphql/reducers/mappings/gef-facilities/mapGefFacilityProvidedOn.js
ttbarnes: Good shout, i've updated to use reduce

davehlewis (approved):

PR #386, opened 2021-10-04
Fix/dtfs2 4786 tfm UI webpack refactor

davehlewis:
https://ukef-dtfs.atlassian.net/browse/DTFS2-4786
- Refactors TFM UI in same manner as other UI's
- Allows hot reloading
- Simplifies docker file
- Adds linting and unit tests to PR pipeline


trade-finance-manager-ui/.dockerignore copy
```diff
@@ -0,0 +1,8 @@
+buildspec.yml
```
ttbarnes, line 1: `.dockerignore copy`

ttbarnes (commented):
Nice! LGTM just minor comment

trade-finance-manager-ui/.dockerignore copy
davehlewis: thanks, have fixed

ttbarnes (approved):

PR #387, opened 2021-10-05
Number generator tweaks

ttbarnes:
- Remove number generator error response if env vars are missing. This is sometimes returned locally even if they do exist.
- update some GEF mongo calls used by number generator
- add/remove some console logs


davehlewis (approved):
LGTM

PR #388, opened 2021-10-05
DTFS2-4966: replace express-json2csv

davehlewis:
https://ukef-dtfs.atlassian.net/browse/DTFS2-4966
- Removes express-json2csv (with it's vulnerabilities)
- Replaces with util that performs the same function


ttbarnes (approved):
Beautiful

PR #404, opened 2021-10-12
DTFS-4586 : ACBS facility update when a facility stage code has been …

abhi-markan:
This branch deals with ACBS update when a facility (which is unissued on ACBS and the deal has been submitted to ACBS and acknowledged by the TFM) stage code has been issued on the portal.

This commit also includes various linting fixes, argument validation and removal of **Number()** on a date operation causing date to be invalid.

### DIFF
--- a/reference-data-proxy/src/v1/controllers/acbs.controller.js
+++ b/reference-data-proxy/src/v1/controllers/acbs.controller.js
@@ -75,7 +75,7 @@ exports.findOne = async (req, res) => {
     return res.status(facilityIdStatus).send();
   }
 
-  return res.status(404).send();
+  return res.status(500).send();
 };
 
 const issueAcbsFacility = async (id, facility, supplierName) => {


azure-functions/acbs-function/mappings/facility/facility-update.js
```diff
@@ -40,18 +40,14 @@
...
-    obligorName,
-  };
-};
+const facilityUpdate = (facility, acbsFacility, obligorName) => ({
+  ...acbsFacility,
+  capitalConversionFactorCode: helpers.getCapitalConversionFactorCode(facility),
+  issueDate: helpers.getIssueDate(facility, acbsFacility.effectiveDate),
+  facilityStageCode: CONSTANTS.FACILITY.STAGE_CODE.ISSUED,
+  foreCastPercentage: CONSTANTS.FACILITY.FORECAST_PERCENTAGE.ISSUED,
+  productTypeName: facility.facilityType || facility.facilitySnapshot.type,
```
ttbarnes, line 49: I think need to add an extra `facilityType` condition here for it to work in both scenarios, with GEF and BSS?

```js
(facility.facilityType
  || facility.facilitySnapshot.type
  || facility.facilitySnapshot.facilityType)
```

When this is called with submission mapping, all facilities will have `facility.facilityType`. Otherwise - only GEF has `.type`, BSS has `.facilityType`.

trade-finance-manager-api/src/v1/api.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-console */
```
ttbarnes, line 1: can remove this and in the file above

reference-data-proxy/src/v1/controllers/acbs.controller.js
```diff
@@ -78,41 +79,51 @@ exports.findOne = async (req, res) => {
...
+    return response;
+  }
+  return {};
 };
 
 exports.createAcbsRecordPOST = async (req, res) => {
-  const { deal, bank } = req.body;
-  const { status, data } = await createAcbsRecord(deal, bank);
-  return res.status(status).send(data);
+  if (req) {
```
ttbarnes, line 123: Do we need to add the conditions to each function? Wherever the function is called from, could the check be from there instead? e.g

```js
// other-file.js

if (deal) {
  await createAcbsRecord(deal);
}
```

This way, ACBS controller is cleaner and less logic, doing pure API calls with a little destructuring?

trade-finance-manager-api/src/v1/api.js
```diff
@@ -424,42 +425,50 @@ const getPremiumSchedule = async (premiumScheduleParameters) => {
...
-      },
-    });
-    return response.data;
-  } catch (err) {
-    return err;
+  if (!!facility && !!supplierName) {
+    try {
+      const response = await axios({
+        method: 'post',
+        // eslint-disable-next-line no-underscore-dangle
```
ttbarnes, line 455: no need for this 👍

ttbarnes (commented):
Few minor comments otherwise LGTM!

ttbarnes (approved):

ttbarnes (approved):

PR #406, opened 2021-10-13
feat/DTFS2-4525 - Ability to query TFM deals by specific field name and value

ttbarnes:
### Central API 
- TFM get deals endpoint: add `byField` to do field/value equality checks in MongoDB find() call.
- split up TFM get deals API tests

### TFM API
- Update TFM deals schema with `byField` query param to accept an array of objects containing field name and values. E.g now deals can be queried via GraphQL like this:
```js
const queryParams  = {
  sortBy: { field: 'deal.submissionDate', order: 'ascending' },
  searchString: 'Amazing bank',
  byField: [
    {
      name: 'tfm.dateReceived',
      value: '11-10-2021'
    },
    {
      name: 'dealSnapshot.someOtherField',
      value: true
    }
  ]
};
```

### TFM UI
- Update "deals" GraphQL query to include `byField`. This is just an example, consumed by other UKEF systems. TFM UI uses the "deals light" query.


dtfs-central-api/api-tests/v1/tfm/deal/tfm-deals-get-sorting.api-test.js
```diff
@@ -0,0 +1,240 @@
+const moment = require('moment');
```
IgnatG, line 1: same as above

IgnatG (approved):

PR #412, opened 2021-10-15
DTFS2-4698 - prevent same user from submitting their own deal

eddieleeper:
Work in progress - cypress test currently failing


abhi-markan (approved):

gef-ui/server/controllers/application-submission/index.test.js
```diff
@@ -174,6 +176,7 @@ describe('controllers/application-submission', () => {
           comments: [{
             role: 'maker', userName: 'maker', createdAt: expect.any(Number), comment: mockRequest.body.comment,
           }],
+          editorId: expect.any(Number),
```
ttbarnes, line 179: I think outside of tests this will be a string as it uses the userId? Might want to update this and the mock `res.editorId` to be a string

IgnatG (approved):

PR #417, opened 2021-10-18
DTFS2-4672 : TFM sort functionality

abhi-markan:
TFM should now be able to sort the deals under **All Deals** page by clicking on the respective column.
This commit ensure both BSS / ECWS & GEF varied data models are taken into account while performing the sorting operation.

This commit has also addressed code simplification for an existing utility function.

**PR Review 1**
1. Code simplification
2. All deal njk template now adheres to GEF data model instead of BSS/ECWS data model paths.
3. Cross data model mapping now returns BSS/ECWS data model path.

**PR Review 2**
1. Moved empty check function to utils
2. Renamed argument

**PR Review 3**
1. Added Jest test cases covering all the functions.
2. Minor amends in the functions.

### DIFF
--- a/dtfs-central-api/src/utils/setEmptyIfNull.js
+++ b/dtfs-central-api/src/utils/setEmptyIfNull.js
@@ -1,5 +1,5 @@
 const setEmptyIfNull = (value) => {
-  if (value === null || value === undefined) return '';
+  if (!value) return '';
   return value;
 };
 


dtfs-central-api/src/utils/getCrossDataModelProperty.js
```diff
@@ -0,0 +1,23 @@
...
+};
+
+const getCrossDataModelProperty = (deal, propertyPath) => {
+  if (deal && propertyPath) {
+    if (deal.dealSnapshot.dealType === CONSTANTS.DEALS.DEAL_TYPE.GEF) return getGEFProperty(propertyPath);
+  }
+  return propertyPath;
+};
+
+module.exports = getCrossDataModelProperty;
```
ttbarnes, line 23: Currently `deals-table.njk` has button values based on BSS, e.g:

```js
buttonValue: 'dealSnapshot.submissionDetails.supplier-name',
```

I was thinking it could be changed to be like the GEF model which we want to get towards:

```js
buttonValue: 'dealSnapshot.exporter.companyName',
```

Then in `getCrossDataModelProperty.js`, could do the mapping so if it's BSS, use `submissionDetails.supplier-name` instead

This way when the data models are aligned, it can just be removed from this one function. Also wouldn't need the `getProperty` function. What do you think?

Also because this is doing mapping I think it should go into a mappings directory/file instead of utils?

dtfs-central-api/src/v1/controllers/tfm/deal/tfm-get-deals.controller.js
```diff
@@ -2,11 +2,17 @@ const moment = require('moment');
...
+const isEmpty = (value) => {
+  if (value === null) return '';
+  return value;
+};
 
 const sortDeals = (deals, sortBy) =>
   deals.sort((xDeal, yDeal) => {
-    const xField = getObjectPropertyValueFromStringPath(xDeal, sortBy.field);
-    const yField = getObjectPropertyValueFromStringPath(yDeal, sortBy.field);
+    const xField = isEmpty(getObjectPropertyValueFromStringPath(xDeal, getCrossDataModelProperty(xDeal, sortBy.field)));
```
ttbarnes, line 14: Why do we need `isEmpty`? It's a bit hard to read this line

abhi-markan: To find values which are null, if they are then give a unicode value which improves the sorting. I can break into multiple lines no problem.

dtfs-central-api/src/utils/getCrossDataModelProperty.js
ttbarnes: 👍 no worries, GEF is the newer cleaner version - very flat, whereas BSS has lots of nesting and some strange/bad field names

dtfs-central-api/src/utils/getObjectPropertyValueFromStringPath.js
```diff
@@ -1,9 +1,7 @@
...
-  str.split('.').reduce((p, propertyName) => {
-    if (p && p[propertyName]) {
-      return p[propertyName];
-    }
+const getObjectPropertyValueFromStringPath = (deal, field) =>
+  field.split('.').reduce((o, p) => {
+    if (o && o[p]) return o[p];
     return null;
-  }, obj);
+  }, deal);
```
ttbarnes, line 5: I'd keep this as obj because then it's generic, could be used on things that aren't deals 😉

dtfs-central-api/src/v1/controllers/tfm/deal/tfm-get-deals.controller.js
```diff
@@ -2,11 +2,27 @@ const moment = require('moment');
...
+        xDeal,
+        getCrossDataModelProperty(xDeal, sortBy.field),
+      ),
+    );
+    const yField = isEmpty(
+      getObjectPropertyValueFromStringPath(
+        yDeal,
+        getCrossDataModelProperty(yDeal, sortBy.field),
+      ),
+    );
```
ttbarnes, line 25: `isEmpty` feels strange to me because it returns a value, i'd expect it to return true or false - and could be extracted into utils/helper directory with tests, so can be used anywhere.

I'd do it like this:

```js
const xField = !isEmpty && (
  getObjectPropertyValueFromStringPath(
    xDeal,
    mapFieldName(xDeal, sortBy.field),
  ),
);
```

Also think `mapFieldName` is cleaner name than `getCrossDataModelProperty`.

What do you think?

dtfs-central-api/src/utils/setEmptyIfNull.js
```diff
@@ -0,0 +1,6 @@
+const setEmptyIfNull = (value) => {
+  if (value === null || value === undefined) return '';
```
ttbarnes, line 2: Could probably just be:

```js
if (!value) return '';
```

PR #464, opened 2021-11-08
DTFS2-4525 - Fix/improve TFM API deals query for external teams. Query timestamp fields by date string

ttbarnes:
### Why?

- Without this change, only deals with specific timestamps would get returned if they are passed into the `byField` request body. 
- Now, a date format can be passed in, e.g '01-12-2021' and any timestamp that's within this day will be returned in the response.
- How? We generate start-of-day and end-of-day timestamps with `date-fns` and then use these in the MongoDB query, with `$gte`/`$lte` operators.

### Central API

- Create `get-deals-date-helpers` function for API call:
  - keeps date input format and supported timestamp fields in one place.
  - main function: `dayStartAndEndTimestamps`: given a date in '01-12-2021' format, this will return 2 timestamps of the start and end of this day.
  - all of these functions are then used in the "get deals" endpoint/MongoDB query, if an appropriate field is passed into `byField` request body.

### TFM API

- Add `bank.emails` and `bank.partyUrn` to GEF GraphQL reducer mapping.
- Update GraphQL schema with missing fields and incorrect timestamp types (Float instead of Int).
- Fix issue where api.queryDeals function was not correctly passing `byField` param.
- Update example `deals-query` that external teams use.

### Other

- Various lint fixes.

### DIFF
--- a/dtfs-central-api/api-tests/v1/tfm/deal/tfm-deals-get-query-field.api-test.js
+++ b/dtfs-central-api/api-tests/v1/tfm/deal/tfm-deals-get-query-field.api-test.js
@@ -1,3 +1,4 @@
+const { getTime, format } = require('date-fns');
 const wipeDB = require('../../../wipeDB');
 const app = require('../../../../src/createApp');
 const api = require('../../../api')(app);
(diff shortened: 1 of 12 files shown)


IgnatG (approved):

ttbarnes:
Closing this - moved to another branch/PR because of branch name. https://github.com/UK-Export-Finance/dtfs2/pull/468

PR #471, opened 2021-11-10
DTFS-4443 added RISK_MANAGERS to credit risk edit group

Zainzzkk:


trade-finance-manager-ui/server/controllers/case/underwriting/pricing-and-risk/index.test.js
```diff
@@ -8,12 +8,13 @@ import {
 
 const res = mockRes();
 
+// changed to RISK_MANAGERS TO CHECK WORKS
```
IgnatG, line 11: the Underwriters are not supposed to be in this list anymore?

trade-finance-manager-ui/server/controllers/case/underwriting/pricing-and-risk/helpers.js
```diff
@@ -2,6 +2,6 @@ const CONSTANTS = require('../../../../constants');
 const { userIsInTeam } = require('../../../../helpers/user');
 
 const userCanEditGeneral = (user) =>
-  userIsInTeam(user, [CONSTANTS.TEAMS.UNDERWRITERS, CONSTANTS.TEAMS.UNDERWRITER_MANAGERS]);
+  userIsInTeam(user, [CONSTANTS.TEAMS.UNDERWRITERS, CONSTANTS.TEAMS.UNDERWRITER_MANAGERS, CONSTANTS.TEAMS.RISK_MANAGERS]);
```
ttbarnes, line 5: Can we update the test to include the new user? `pricing-and-risk/helpers.test.js `

ttbarnes (changes requested):
Could we add an E2E test for this in `pricing-and-risk.spec.js`?

Something like

```js
it('a user in the RISK_MANAGERS team can edit Credit Rating', () => {
  cy.login(riskManagerUser);
  
  // fill out form and submit
  // all should be good
});
```

At the moment if I were to remove `BUSINESS_SUPPORT` from the editable list, all tests would still pass.

trade-finance-manager-ui/server/controllers/case/underwriting/pricing-and-risk/helpers.js
Zainzzkk: done

e2e-tests/trade-finance-manager/cypress/integration/journeys/case/underwriting/pricing-and-risk-risk-managers.spec.js
```diff
@@ -0,0 +1,160 @@
...
+      cy.url().should('eq', relative(`/case/${dealId}/underwriting/pricing-and-risk`));
+    });
+
+    it('edit button is visible as compared to add', () => {
+      pages.underwritingPricingAndRiskPage.addRatingLink().should('not.exist');
+      pages.underwritingPricingAndRiskPage.exporterTableChangeCreditRatingLink().should('exist');
+      pages.underwritingPricingAndRiskPage.exporterTableChangeProbabilityOfDefaultLink().should('exist');
+    });
+  });
+});
```
ttbarnes, line 160: I think we can simplify this file a bit - most of this is already tested in the other pricing-and-risk E2E test.

Should we have just the last 3 tests in here, so it's only testing the submission and after submission?

I think we should remove the tests before this one:
```submitting a rating displays the rating in table on `pricing and risk` page and does not render `add credit rating` link```

Just thinking of E2E test time, the more tests we have the longer it takes

What do you think?

ttbarnes (approved):
Great stuff - just minor comment about simplifying the new E2E test file. Happy to merge either way 👍

e2e-tests/trade-finance-manager/cypress/integration/journeys/case/underwriting/pricing-and-risk-risk-managers.spec.js
Zainzzkk: think good idea, have done it

ttbarnes (approved):

IgnatG (approved):

PR #472, opened 2021-11-10
Company house non UK (British isles only) country codes

abhi-markan:
1. UK countries from company house API will be converted to GBR.
2. When GBR is pre-defined ACBS country code call is skipped.

### DIFF
--- a/azure-functions/acbs-function/constants/deal.js
+++ b/azure-functions/acbs-function/constants/deal.js
@@ -24,10 +24,24 @@ const CURRENCY = {
   DEFAULT: 'GBP',
 };
 
+const COUNTRY = {
+  DEFAULT: 'GBR',
+};
+
+const UNITED_KINGDOM = [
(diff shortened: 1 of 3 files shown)


azure-functions/acbs-function/constants/deal.js
```diff
@@ -24,10 +24,24 @@ const CURRENCY = {
   DEFAULT: 'GBP',
 };
 
+const COUNTRY = {
+  DEFAULT: 'GBR',
+};
+
+const UNITED_KINGDOM = [
+  'united kingdom',
```
IgnatG, line 32: are they supposed to be lowercased?

abhi-markan: I have converted it to lower case for comparison, I also convert the country to lower case for true like comparison.

IgnatG (approved):

PR #474, opened 2021-11-10
feature(DTFS-4180): Portal - Ability to review a UKEF declined manual  inclusion application

IgnatG:
- general linting
- updated the packages
- core functionality for reviewing a declined application


trade-finance-manager-api/src/v1/api.js
```diff
@@ -583,6 +583,23 @@ const updatePortalGefDealStatus = async (dealId, status) => {
...
+      url: `${centralApiUrl}/v1/portal/gef/deals/${dealId}/comment`,
+      headers: {
+        'Content-Type': 'application/json',
+      },
+      data: { dealId, commentType, comment },
+    });
+
+    return response.data;
+  } catch ({ response }) {
+    return false;
```
abhi-markan, line 599: Can we please add console.error({ response }) for logging purposes.

IgnatG: Good spot. I've fixed this now

abhi-markan (approved):

PR #478, opened 2021-11-10
Effective date format change

abhi-markan:
1. formatTimestamp only works well for EPOCH or Date format not for Date time format, function change resolves the bug.


ttbarnes (approved):
LGTM

PR #481, opened 2021-11-15
DTFS-4535 Added e2e test for returning MIN to maker

Zainzzkk:


ttbarnes (commented):
LGTM, couple of v.minor comments.

What are the test PDF files for? Doesn't look like they're used

e2e-tests/gef/cypress/integration/return-to-maker-min.spec.js
```diff
@@ -0,0 +1,110 @@
...
+      })
+      .then(({ body }) => {
+        applicationId = body.items[2]._id;
+
+        cy.login(CREDENTIALS.MAKER);
+      });
+  });
+
+  describe('creates and submits min', () => {
+    beforeEach(() => {
```
IgnatG, line 29: I'm not sure if you need `beforeEach` here because there is only one test inside this `describe` block

e2e-tests/gef/cypress/integration/return-to-maker-min.spec.js
```diff
@@ -0,0 +1,110 @@
...
+
+      applicationDetails.submitButton().click();
+
+      applicationSubmission.submitButton().click();
+      applicationSubmission.confirmation();
+    });
+  });
+
+  describe('return to maker as checker', () => {
+    beforeEach(() => {
```
IgnatG, line 67: same as above

e2e-tests/gef/cypress/integration/return-to-maker-min.spec.js
```diff
@@ -0,0 +1,110 @@
...
+
+    it('returns to maker', () => {
+      applicationPreview.returnButton().click();
+      returnToMaker.comment().type('comment1');
+      returnToMaker.submitButton().click();
+    });
+  });
+
+  describe('return to maker', () => {
+    beforeEach(() => {
```
IgnatG, line 81: same as above

e2e-tests/gef/cypress/integration/return-to-maker-min.spec.js
ttbarnes: I think it's needed - we were debugging earlier - the user flow is
- maker logs in, does stuff
- checker logs in, does stuff
- maker logs in, does stuff

Each beforeEach logs in as the relevant user. Without this or with a single `before(() ...)`, the E2E file just errors in cypress. It also only seems to like "preserve cookies" in a beforeEach

If there is better way that'd be great :D

IgnatG: oh! interesting behaviour. no idea what the alternative is then :) so leave it as it is

e2e-tests/gef/cypress/integration/return-to-maker-min.spec.js
```diff
@@ -0,0 +1,110 @@
...
+      Cypress.Cookies.preserveOnce('connect.sid');
+      cy.login(CREDENTIALS.CHECKER);
+      cy.visit(relative(`/gef/application-details/${applicationId}`));
+    });
+
+    it('returns to maker', () => {
+      applicationPreview.returnButton().click();
+      returnToMaker.comment().type('comment1');
+      returnToMaker.submitButton().click();
```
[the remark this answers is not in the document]
Zainzzkk: sorted by checking goes back to dashboard

IgnatG (approved):

PR #492, opened 2021-11-18
DTFS2-5080 - ability to pass excludedTasks to TFM tasks generation

ttbarnes:
## Summary 
- Allow an array of task titles to be passed to `create-tasks.js`.
- `create-tasks` will not use any tasks in the provided `excludedTasks` array
- Use a manual incremental taskId counter with `forEach`, instead of using index with `.map`. Otherwise will have array with null's and mismatched task IDs.
- When a deal is submitted and it has an exporter `partyURN`, exclude the "match parties" task.

## Other
- Move and rename `tasks-creation.js`/functions into `helpers/create-tasks.js` Because now it is not simple defaults.

## Note on what's happening with tasks creation:

Tasks structure is like this after creation - an array of "groups". Each group can have as many tasks as they like. E.G:

```js
deal.tfm.tasks = [
  {
    groupTitle: 'Group A',
    id: 1,
    groupTasks: [
      {
        title: 'Task 1',
        id: '1',
        groupId: 1,
        status: 'To do',
        canEdit: true,
      },
      {
        title: 'Task 2',
        id: '2',
        groupId: 1,
        status: 'Cannot start yet',
      },
    ]
  },
  {
    groupTitle: 'Group B',
    id: 1,
    groupTasks: [
      {
        title: 'Task 1',
        id: '1',
        groupId: 2,
        status: 'Cannot start yet',
      },
      {
        title: 'Task 2',
        id: '2',
        groupId: 2,
        status: 'Cannot start yet',
      },
    ],
  },
];
```

MIA has many tasks, AIN only has a few.

This PR allows an array to be passed like this and will not add/create any tasks with a given title.

```js
const excludedTasks = [
  'Complete something on Salesforce',
];

const tasks = createTasks(
  'Automatic Inclusion Notice',
  excludedTasks
);
```

For the ticket/business requirement, we exclude one party matching task if a party URN already exists. This solution allows for as many exclusions as required :)


IgnatG (approved):
LGTM

PR #504, opened 2021-11-24
DTFS2-3549 - TFM tasks creation based on list of additional tasks (depending on deal data)

ttbarnes:
## Business requirement
- Add a task creation rule for a BSS MIA deal that has eligibility criteria answer 11 as false (additional task gets added)

## How
- Add `isConditional` flag to tasks that can be included or not included (on deal submission), depending on deal data
- Remove ability to pass a list of 'excluded tasks' to task creation function

## How / clean up / scalability
- Instead of having a list of 'exluded tasks', we now only specifically define a list of 'additional tasks' that will be added, depending on the deal data
- Make `deal.tasks.js` more functional - each task rule in it's own function
- Add logic to 'create group tasks' function so that a task will only be added to a group if:
  - the task is *not*  a conditional task (therefore always applies)
  - or the task is a conditional task and is included in the list of additional tasks to add

## Why 
- This approach saves logic/conditions and potentially hard to read or confusing code
- Specifically creating 'additional tasks' to add (depending on the deal data), is clearer and more scalable than having a list of 'excluded tasks'
- Splitting out each task creation condition ("if X in deal, add this task") ...into their own functions is nice and clean; can be easily removed, expanded and tested

### DIFF
--- a/trade-finance-manager-api/src/v1/helpers/create-tasks.api-test.js
+++ b/trade-finance-manager-api/src/v1/helpers/create-tasks.api-test.js
@@ -24,7 +24,7 @@ describe('defaults - tasks creation', () => {
   });
 
   describe('createGroupTasks', () => {
-    const mockTasks = [
+    const mockGroupTasks = [
       {
         title: 'Title A',
         team: 'Team A',
@@ -36,7 +36,7 @@ describe('defaults - tasks creation', () => {
     ];
 
     it('should return array of tasks with incremented task id', () => {
-      const result = createGroupTasks(mockTasks, 2);
+      const result = createGroupTasks(mockGroupTasks, 2);
 
       const expected = [
         {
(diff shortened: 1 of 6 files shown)


trade-finance-manager-api/src/v1/helpers/create-tasks.api-test.js
```diff
@@ -85,41 +85,69 @@ describe('defaults - tasks creation', () => {
...
+        expect(conditionalTask1InResult).toBeUndefined();
+
+        const conditionalTask2InResult = result.find((task) => task.title === conditionalTask2.title);
+        expect(conditionalTask2InResult).toBeUndefined();
+      });
+    });
+
+    describe('when additionalTasks array of strings is passed', () => {
+      it('should return any additional tasks (with isConditional flag) in a group that match a title in additionalTasks, retaining incremental task ids', () => {
+        const mockGroupTasks = [
```
IgnatG, line 119: could we rename this to something else? it looks like it was declared already on line 27

IgnatG:
LGTM
I left a small comment. not a show stopper though

trade-finance-manager-api/src/v1/helpers/create-tasks.api-test.js
ttbarnes: Done here: https://github.com/UK-Export-Finance/dtfs2/pull/504/commits/8b46da370dbdc077d6e2441434661ffa529cf55c

IgnatG (approved):

PR #505, opened 2021-11-25
DTFS2-4056 Reversed order activities are added to DB 

Zainzzkk:
* Reversed order activities are added to DB so timeline shows most recent activity first
* Fixed tfm-deal-put-tfm.api-test.js to reflect this order change


ttbarnes (approved):
✨

PR #510, opened 2021-11-26
DTFS2-4124 : Task section

abhi-markan:
Task section on the application details page is only available when either of the condition is true:

1. Comments are available (more than zero)
2. Decision from UKEF is avialable

### DIFF
--- a/gef-ui/templates/includes/application-preview/task-comments.njk
+++ b/gef-ui/templates/includes/application-preview/task-comments.njk
@@ -1,29 +1,31 @@
 {% from "moj/components/banner/macro.njk" import mojBanner %}
 
-<header data-cy="application-preview-page">
+{% if isUkefReviewAvailable or comments.length > 0 %}
+  <header data-cy="application-preview-page">
 
-  {% set bannerHtml %}
+    {% set bannerHtml %}
 
-    {% if isUkefReviewAvailable %}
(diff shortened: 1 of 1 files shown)


gef-ui/templates/includes/application-preview/task-comments.njk
```diff
@@ -1,29 +1,31 @@
...
-          {{ comment.createdAt | localiseTimestamp('DD MMM YYYY h:mma', timezone) }}</p>
-        <p>{{ comment.comment }}</p>
-      {% endfor %}
-    </div>
+      <div data-cy="latest-comment">
+        {% for comment in comments %}
+          <p><strong>Comments from {{ comment.role }} ({{ comment.userName }})</strong>
+            <br>
+            {{ comment.createdAt | localiseTimestamp('DD MMM YYYY h:mma', timezone) }}</p>
+          <p>{{ comment.comment }}</p>
```
IgnatG, line 19: what are they allowed to type in the comment box? Is it anything? & do we need to ` | safe` pipe? or do we need to include the `white-space: pre-wrap;` styling?

abhi-markan: They can type anything, have added Safe filter.
Pre-wrap is already available in the style.

IgnatG (approved):

PR #513, opened 2021-11-29
Feature/DTFS2-4174 - Task dependency logic for underwriting group tasks

ttbarnes:
WIP

### DIFF
--- a/trade-finance-manager-api/src/constants/tasks.js
+++ b/trade-finance-manager-api/src/constants/tasks.js
@@ -15,6 +15,12 @@ const AIN_AND_MIA = {
   },
 };
 
+const AIN_AND_MIA_ID = {
+  GROUP_1: {
+    CREATE_OR_LINK_SALESFORCE: '1',
+  },
+};
+
 const AIN = {
   GROUP_1: {
     GROUP_TITLE: AIN_AND_MIA.GROUP_1.GROUP_TITLE,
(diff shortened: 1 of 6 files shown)


trade-finance-manager-api/src/constants/tasks.js
```diff
@@ -66,18 +98,26 @@ const MIA = {
         isConditional: true,
       },
       {
+        id: AIN_AND_MIA_ID.GROUP_1.CREATE_OR_LINK_SALESFORCE,
```
ttbarnes, line 101: task ids here need to be removed, they're generated dynamically. Can be changed - depending on the deal data, some tasks will or will not be added, so the task IDs need to be dynamic. Group IDs don't need to be dynamic right now

trade-finance-manager-api/src/v1/controllers/tasks.controller.js
```diff
@@ -77,26 +77,38 @@ const updateTasksCanEdit = async (allTaskGroups, groupId, taskIdToUpdate, deal,
...
+            if (task.groupId === CONSTANTS.TASKS.MIA.GROUP_3.id
+               && task.title === CONSTANTS.TASKS.MIA_GROUP_3_TASKS.CHECK_EXPOSURE
+               && task.status === CONSTANTS.TASKS.STATUS.CANNOT_START) {
+              for (let i = index + 1; i <= index + 2; i += 1) {
+                const nextTask = array[i];
+                nextTask.canEdit = true;
+                nextTask.status = CONSTANTS.TASKS.STATUS.TO_DO;
+                // Send task notification emails
+                sendUpdatedEmailRequests.push(sendUpdatedTaskEmail(nextTask, deal, urlOrigin));
+              }
```
ttbarnes, line 102: I think this could be cleaner with no need for a for loop and index based, leave with me 😄

ttbarnes:
Closing as in a new branch: #517

PR #520, opened 2021-12-07
DTFS2-4973 - TFM - GEF facility/fee record fixes

ttbarnes:
- do not calculate GEF facility fee record if the deal type is MIA
- do not use `submittedAsIssuedDate` in facility `coverStartDate` mapping, just use the `coverStartDate`.
- add missing facility feeType mapping for "At Maturity" value
- lint fixes


abhi-markan (approved):
Look good to me

PR #523, opened 2021-12-08
Updated GHA AZ CLI version

abhi-markan:


.github/workflows/promote_to_test.yml
```diff
@@ -592,7 +592,7 @@ jobs:
     - name: Delete slot
       # Clean up the slot regardless of errors
       if: ${{ always() }}
-      uses: Azure/cli@v1.0.0
+      uses: Azure/cli@v1.0.0.0.0
```
IgnatG, line 595: do we need these `0.0` at the end?
there are a few items above as well that have the same numbers

IgnatG (approved):

PR #526, opened 2021-12-09
DTFS2-5184 - GEF deals using deals collection. GEF exporter in the deal

ttbarnes:
All of these changes are required to:
- get GEF deals into deals collection (so BSS and GEF deals in same collection)
- get GEF exporter into the GEF deal, instead of using `exporterId` and constantly calling `gef-exporter` collection.

## GEF deals collection
- Remove `gef-applications` collection usage
- All GEF deals now using same collection as BSS: `deals`

## GEF exporter
- Remove `gef-exporter` collection usage and endpoints
- A deal's exporter is now inside the GEF deal. No more `deal.exporterId` or `deal.exporter.details`
- Update all code that consumed exporterId/api calls. Less calls, logic, mocks, yay

## Mock data loader
- Update mock deals so that BSS and GEF deals can be added and removed from the same collection; also with GEF exporter inside the deal
- Add more logs and dealType to existing BSS deals
- Improve exports/objects in mock GEF exporters

## Portal API
- Add graphQL reducer/mapping for "query all deals". Without this, GEF deals will not get returned in this response and things won't work.
- Remove GEF exporter and address models. No longer used.

## Other
- GEF UI: fix/update/improve some server tests relating to exporter/address. Not sure how these were passing before...
- GEF UI: move constants into server directory. Things were failing where a `DEAL_STATUS` constant could not be found. 🤷 
- GEF E2E: enter exporter-corr-address - small tweaks to update deal instead of exporter; clean up test description as it was incorrect and not doing what it says
- BSS E2E: 'fix'/simplify a test relating to deleting a facility. Not sure how this was passing before...
- Portal API: extract companies house response mapping into it's own function.

## Note
In Portal API, we have several graphQL queries for "get all deals" that are currently used to display BSS or GEF deals in individual pages. One GraphQL call is used for both BSS and GEF but in the UI, is only used for one or the other. I've made the current functionality/UI design work with these existing calls; but after this PR, this will all be simplified so we only have 1 "deals" page, and one deals GraphQL query.


gef-ui/server/utils/helpers.js
```diff
@@ -144,7 +143,7 @@ const mapSummaryList = (data, itemsToShow, preview = false) => {
     // If value is a number, convert to String as 0 can also become falsey
     const value = typeof details[item.id] === 'number' || typeof details[item.id] === 'boolean' ? details[item.id].toString() : details[item.id];
     const { currency, detailsOther } = details;
-    const isRequired = required.includes(item.id);
+    const isRequired = (validation && validation.required && validation.required.includes(item.id));
```
IgnatG, line 146: can we do `validation?.required?.includes(item.id)` - or something like that?

portal-api/src/v1/gef/mappings/map-companies-house-data.js
```diff
@@ -0,0 +1,30 @@
+const mapCompaniesHouseData = (data, industries) => {
+  const address = data.registered_office_address;
+
+  let selectedIndustry = null;
+  if (industries && industries.length === 1) {
+    // eslint-disable-next-line prefer-destructuring
+    selectedIndustry = industries[0];
```
IgnatG, line 7: can we do `[selectedIndustry] = industries;`? - and we can remove the `//eslint-disable...` comment from above

gef-ui/server/controllers/about-exporter/index.js
```diff
@@ -16,18 +16,17 @@ const aboutExporter = async (req, res) => {
...
+    const probabilityOfDefault = JSON.stringify(exporter.probabilityOfDefault);
 
     return res.render('partials/about-exporter.njk', {
-      smeType: details.smeType,
-      probabilityOfDefault: probabilityOfDefault !== 'null' ? probabilityOfDefault : null,
-      isFinanceIncreasing: isFinanceIncreasing !== 'null' ? isFinanceIncreasing : null,
-      selectedIndustry: details.selectedIndustry,
       applicationId,
+      smeType: exporter?.smeType,
+      probabilityOfDefault: probabilityOfDefault !== 'null' ? exporter.probabilityOfDefault : null,
```
IgnatG, line 27: do we expect the `probabilityOfDefault` to be `'null'` - meaning, as a string? or can we do it without quotes?

gef-ui/server/controllers/about-exporter/index.js
```diff
@@ -43,11 +42,10 @@ const validateAboutExporter = async (req, res) => {
   const { saveAndReturn, status } = query;
 
   try {
-    const { exporterId } = await api.getApplication(applicationId);
-    const { details } = await api.getExporter(exporterId);
+    const { exporter } = await api.getApplication(applicationId);
 
     // Only validate industries if it contains more than 1 SIC code
-    if (!saveAndReturn && !body.selectedIndustry && (details.industries && details.industries.length > 1)) {
+    if (!saveAndReturn && !body.selectedIndustry && (exporter.industries && exporter.industries.length > 1)) {
```
IgnatG, line 48: can we do `if (... && (exporter?.industries?.length > 1)){...}` ?

gef-ui/server/controllers/application-details/index.js
```diff
@@ -55,15 +55,19 @@ function buildBody(app, previewMode) {
...
     isAutomaticCover: app.submissionType === DEAL_SUBMISSION_TYPE.AIN,
     exporter: {
       status: app.exporterStatus,
-      rows: mapSummaryList(app.exporter, exporterItems(exporterUrl, {
-        showIndustryChangeLink: app.exporter.details.industries && app.exporter.details.industries.length > 1,
-      }), previewMode),
+      rows: mapSummaryList(
+        { details: app.exporter }, // wrap in details because mapSummaryList relies this.
+        exporterItems(exporterUrl, {
+          showIndustryChangeLink: app.exporter?.industries && app.exporter?.industries.length > 1,
```
IgnatG, line 67: wouldn't this work? `showIndustryChangeLink:  app.exporter?.industries?.length > 1,`

gef-ui/server/controllers/about-exporter/index.js
ttbarnes: Good point I think this can be simplified a lot as nothing in exporter defaults to null now.

gef-ui/server/controllers/about-exporter/index.js
```diff
@@ -16,18 +16,17 @@ const aboutExporter = async (req, res) => {
...
 
     return res.render('partials/about-exporter.njk', {
-      smeType: details.smeType,
-      probabilityOfDefault: probabilityOfDefault !== 'null' ? probabilityOfDefault : null,
-      isFinanceIncreasing: isFinanceIncreasing !== 'null' ? isFinanceIncreasing : null,
-      selectedIndustry: details.selectedIndustry,
       applicationId,
+      smeType: exporter?.smeType,
+      probabilityOfDefault: probabilityOfDefault !== 'null' ? exporter.probabilityOfDefault : null,
+      isFinanceIncreasing: isFinanceIncreasing !== 'null' ? isFinanceIncreasing : null,
```
abhi-markan, line 28: I suppose same goes for `isFinanceIncreasing`

gef-ui/server/controllers/exporters-address/index.js
```diff
@@ -25,9 +25,10 @@ const validateExportersAddress = async (req, res) => {
     const { params, body } = req;
     const { correspondence, postcode } = body;
     const { applicationId } = params;
-    const { exporterId } = await api.getApplication(applicationId);
-    const { details } = await api.getExporter(exporterId);
-    const { registeredAddress } = details;
+
+    const { exporter } = await api.getApplication(applicationId);
+    const { companyName = '' } = exporter;
```
abhi-markan, line 30: Is this statement required?

gef-ui/server/services/api.js
```diff
@@ -153,9 +135,9 @@ const getEligibilityCriteria = async () => {
...
 };
 
-const getCompaniesHouseDetails = async (companyRegNumber, exporterId) => {
+const getCompaniesHouseDetails = async (companyRegNumber) => {
   try {
-    const { data } = await Axios.get(`/gef/company/${companyRegNumber}`, { params: { exporterId } });
+    const { data } = await Axios.get(`/gef/company/${companyRegNumber}`);
     return data;
   } catch (err) {
     return apiErrorHandler(err);
```
abhi-markan, line 143: Please add `console.error('Unable to get company house details', { err });`

abhi-markan (changes requested):
Have left comments. Thanks

gef-ui/server/controllers/about-exporter/index.js
ttbarnes: Made it loads simpler

gef-ui/server/controllers/about-exporter/index.js
ttbarnes: Made loads simpler

gef-ui/server/controllers/exporters-address/index.js
ttbarnes: Yes, it's used in the template - if it's undefined it will throw error during this destructuring - so defaulting it to an empty string.

gef-ui/server/services/api.js
ttbarnes: Done - there's loads of other functions in this file that could do with the same treatement, but i've added just for this

abhi-markan (approved):

IgnatG (approved):
Lgtm

PR #528, opened 2021-12-13
DTFS2-5053 : BSS/EWCS ACBS deal creation condition

abhi-markan:
**BSS/ECWS & GEF Deals**
Deals are only created in ACBS when the status of the application is one of the following:
1. AIN
2. MIN


ttbarnes (changes requested):
Need tests

trade-finance-manager-api/src/v1/controllers/deal.controller-test.js
```diff
@@ -0,0 +1,9 @@
+const MOCK_DEAL_MIA_SUBMITTED = require('../__mocks__/mock-deal-MIA-submitted');
+
+const dealController = require('./deal.controller');
+
+describe('dealCanBeSubmittedToACBS()', () => {
+  it('Should ascertain whether the deal can be submitted to the ACBS', async () => {
+    expect(dealController.dealCanBeSubmittedToACBS(MOCK_DEAL_MIA_SUBMITTED)).toEqual(false);
+  });
```
ttbarnes, line 8: Can we add a couple of tests for the true values as well - should be true when AIN or MIN?

trade-finance-manager-api/src/v1/controllers/deal.submit.controller.js
```diff
@@ -95,8 +95,7 @@ const submitDealAfterUkefIds = async (dealId, dealType, checker) => {
        * Current requirement only allows AIN & MIN deals to be send to ACBS
        */
       const updatedDeal = await api.updateDeal(dealId, updatedDealWithTasks);
-      if (mappedDeal.submissionType === CONSTANTS.DEALS.SUBMISSION_TYPE.AIN
-      || mappedDeal.submissionType === CONSTANTS.DEALS.SUBMISSION_TYPE.MIN) {
+      if (dealController.dealCanBeSubmittedToACBS(dealId)) {
```
ttbarnes, line 98: Could we just pass in `mappedDeal.submissionType` here, save a DB call in this function?

ttbarnes: Also would be great to add a test in the actual submit endpoint `deals-submit.api-test.js` to check that `submitACBSIfAllPartiesHaveUrn` is called on submission - it's already stubbed in the test, but nothing checking if it's called

Not a blocker for this PR as it's already missing coverage, nice to have

ttbarnes (dismissed):

ttbarnes (dismissed):

ttbarnes (approved):
😍

PR #529, opened 2021-12-14
DTFS2-5078 - Maker from same bank can re-submit

abhi-markan:
This patch allows maker from the same bank to re-submit the application to the checker provided the user role is `Maker`.

### DIFF
--- a/gef-ui/server/utils/helpers.js
+++ b/gef-ui/server/utils/helpers.js
@@ -308,7 +308,7 @@ const makerCanReSubmit = (maker, application) => {
     CONSTANTS.DEAL_STATUS.UKEF_APPROVED_WITH_CONDITIONS,
   ];
   const coverDateConfirmed = coverDatesConfirmed(application.facilities);
-  const makerAuthorised = (maker._id === application.maker._id);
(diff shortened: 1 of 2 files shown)


gef-ui/server/utils/helpers.js
```diff
@@ -308,7 +308,7 @@ const makerCanReSubmit = (maker, application) => {
     CONSTANTS.DEAL_STATUS.UKEF_APPROVED_WITH_CONDITIONS,
   ];
   const coverDateConfirmed = coverDatesConfirmed(application.facilities);
-  const makerAuthorised = (maker._id === application.maker._id);
+  const makerAuthorised = (maker.roles.includes('maker') && maker.bank.id === application.bankId);
```
ttbarnes, line 311: where's the test 😝

ttbarnes: Can we rename `maker` to user since it could not be a maker

abhi-markan: Updated :-)

ttbarnes (approved):

PR #531, opened 2021-12-15
Fix GEF exporter status/validation issue. Add test coverage

ttbarnes:
## GEF UI server
- POST endpoint: consume existing exporter data before adding submitted data
- POST endpoint: add unit test to check that api.updateApplication is called with correct data

## GEF API
- Exporter status/validation: cleaning, fix issue where a (valid) false value, could cause incomplete status. Add test coverage

## Other
- Add E2E tests to check exporter status depending on the data shape

### DIFF
--- a/e2e-tests/gef/cypress/integration/about-exporter.spec.js
+++ b/e2e-tests/gef/cypress/integration/about-exporter.spec.js
@@ -1,11 +1,13 @@
 import relative from './relativeURL';
 import aboutExporter from './pages/about-exporter';
+import applicationDetails from './pages/application-details';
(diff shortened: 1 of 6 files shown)


abhi-markan (changes requested):
Do we need `answers.isFinanceIncreasing !== false` for all required fields?

ttbarnes:
@abhi-markan that check is only required for this field as it's the only field that is a boolean 👍

abhi-markan (approved):

abhi-markan (approved):

IgnatG (approved):
LGTM

PR #532, opened 2021-12-15
Show only issued facilities

abhi-markan:
1. UI server function to only return issued facilities
2. Function name updated
3. Unit test case update

### DIFF
--- a/trade-finance-manager-api/src/v1/controllers/deal.controller.js
+++ b/trade-finance-manager-api/src/v1/controllers/deal.controller.js
@@ -90,6 +90,7 @@ const submitACBSIfAllPartiesHaveUrn = async (dealId) => {
   const allRequiredPartiesHaveUrn = allPartiesHaveUrn(deal);
 
   if (allRequiredPartiesHaveUrn) {
+    console.log('*******ACBS*************');
     await acbsController.createACBS(deal);
   }
 };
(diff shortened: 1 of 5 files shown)


trade-finance-manager-api/src/v1/controllers/deal.controller.js
```diff
@@ -90,6 +90,7 @@ const submitACBSIfAllPartiesHaveUrn = async (dealId) => {
   const allRequiredPartiesHaveUrn = allPartiesHaveUrn(deal);
 
   if (allRequiredPartiesHaveUrn) {
+    console.log('*******ACBS*************');
```
IgnatG, line 93: do we need this console.log?

trade-finance-manager-api/src/v1/controllers/deal.submit.controller.js
```diff
@@ -96,6 +96,7 @@ const submitDealAfterUkefIds = async (dealId, dealType, checker) => {
        */
       const updatedDeal = await api.updateDeal(dealId, updatedDealWithTasks);
       if (dealController.canDealBeSubmittedToACBS(mappedDeal.submissionType)) {
+        console.log('*************YES***************');
```
IgnatG, line 99: do we need this console.log?

trade-finance-manager-api/src/v1/controllers/deal.controller.js
abhi-markan: Sorry just removed.

trade-finance-manager-api/src/v1/controllers/deal.submit.controller.js
abhi-markan: Removed

IgnatG (approved):

PR #538, opened 2021-12-16
Removed ukefDecisionAccepted and updated the coverDateConfirmed property

IgnatG:


abhi-markan (approved):

portal-api/src/v1/gef/controllers/validation/exporter.js
```diff
@@ -70,13 +68,8 @@ const exporterCheckEnums = (doc) => {
   return enumErrors.length === 0 ? null : enumErrors;
 };
 
-const exporterValidation = (doc) => ({
-  required: hasRequiredItems(doc),
-});
-
```
ttbarnes, line 76: I think this is incorrect and was fixed in other PR that's been merged, do a pull?

portal-api/src/v1/gef/controllers/clone-application.controller.js
```diff
@@ -78,6 +78,8 @@ const cloneFacilities = async (currentDealId, newDealId) => {
       allFacilities[val].updatedAt = Date.now();
       // reset the ukefFacilityId
       allFacilities[val].ukefFacilityId = null;
+      // reset the coverDateConfirmed
+      allFacilities[val].coverDateConfirmed = null;
```
ttbarnes, line 82: Can we test this?

portal-api/src/v1/gef/controllers/validation/exporter.js
IgnatG: It's in a different PR, yes. 
I don't think the function is being used though as I couldn't find a reference to it anywhere else

portal-api/src/v1/gef/controllers/clone-application.controller.js
IgnatG: this is going to be tricky because the deal has to be accepted by ukef (with or without conditions). 
The problem is that when you submit the deal from GEF to TFM, you have to wait for the number generator to finish and for the statuses to get updated. Otherwise if I log in immediately (which is what the e2e-tests will do), the status will get overwritten

portal-api/src/v1/gef/controllers/validation/exporter.js
ttbarnes: You're right! Sorry, I confused myself.

portal-api/src/v1/gef/controllers/clone-application.controller.js
ttbarnes: I was thinking unit test at least?

PR #540, opened 2021-12-16
BSS deal submissionType on the top level

ttbarnes:
This PR updates all references of `submissionType` so that it's not added or read from `deal.details`, in both BSS and TFM.


IgnatG (approved):
LGTM

PR #542, opened 2021-12-16
DTFS2-4153 : MIA - Checker send back to the maker

abhi-markan:
This commit deals with the Checker's ability to send back an **Approved UKEF** application back to the maker for further input or clarification, following updates have been added to this commit.

1. `isUkefReviewAvailable` and `isUkefReviewPositive` helper functions now accept the application as an argument to ascertain whether `UkefDecision` object exists.
2. `isUkefReviewAvailable` condition added on various **njk** templates for stronger compound conditional branching.
3. Added CSS rule
4. Variable name updated from `newPortalStatus` to `mappedPortalStatus` for clarity.
5. Variable is passed to `UkefDecision` array for correct decision text for GEF deal only.


gef-ui/server/utils/helpers.js
```diff
@@ -228,21 +228,27 @@ const stringToBoolean = (str) => (str === 'false' ? false : !!str);
...
-  return acceptable.includes(applicationStatus);
+const isUkefReviewPositive = (application) => {
+  if (application.ukefDecision?.length > 0) {
+    const acceptable = [
+      CONSTANTS.DEAL_STATUS.UKEF_APPROVED_WITHOUT_CONDITIONS,
+      CONSTANTS.DEAL_STATUS.UKEF_APPROVED_WITH_CONDITIONS,
+    ];
+    return acceptable.includes(application.status) || acceptable.includes(application.ukefDecision[0].decision);
+  }
+  return false;
```
ttbarnes, line 251: Do you think we simplify the params passed to these functions, so it only get what it needs? i.e

```js
const isUkefReviewPositive = (ukefDecision, dealStatus) => ....
```

gef-ui/templates/includes/application-preview/task-comments.njk
```diff
@@ -5,17 +5,21 @@
 
     {% set bannerHtml %}
 
-      {% if isUkefReviewAvailable and not coverDatesConfirmed %}
+      {% if isUkefReviewAvailable and isUkefReviewPositive and not coverDatesConfirmed %}
```
ttbarnes, line 8: Since this is getting a bit complex now in the template, should we move these 3 conditions do the server, so the template here could be something like

```js
{% if renderReviewDecisionLink %}
```

Thoughts?

gef-ui/server/controllers/application-details/index.js
```diff
@@ -93,8 +93,8 @@ function buildBody(app, previewMode) {
     checkerCanSubmit: app.checkerCanSubmit,
     makerCanReSubmit: makerCanReSubmit(userSession, app),
     ukefDecision: app.ukefDecision,
-    isUkefReviewAvailable: isUkefReviewAvailable(app.status),
-    isUkefReviewPositive: isUkefReviewPositive(app.status),
+    isUkefReviewAvailable: isUkefReviewAvailable(app),
+    isUkefReviewPositive: isUkefReviewPositive(app),
```
ttbarnes, line 97: There should be either an updated test for this or a failing test 🤔

abhi-markan: Yep have updated the tests below, they were all failing should be passing now.

gef-ui/server/utils/helpers.test.js
```diff
@@ -600,26 +600,56 @@ describe('isNotice()', () => {
...
 
-  it('Should return FALSE for application with UKEF_REFUSED status', () => {
-    expect(isUkefReviewPositive(CONSTANTS.DEAL_STATUS.UKEF_REFUSED)).toEqual(false);
+  it('Should return TRUE for application with UkefDecision set to `Accepted (with conditions)` and status set to UKEF_APPROVED_WITHOUT_CONDITIONS status', () => {
+    const UKEF_APPPROVED_DEAL = MOCK_DEAL;
+    UKEF_APPPROVED_DEAL.status = CONSTANTS.DEAL_STATUS.UKEF_APPROVED_WITHOUT_CONDITIONS;
+    UKEF_APPPROVED_DEAL.ukefDecision = [{
+      text: '1. Condition 1\r\n2. Condition 2\r\n3. Condition 3',
+      decision: 'Accepted (with conditions)',
+      timestamp: 1639657163,
```
ttbarnes, line 650: Could create one variable for `ukefDecision` in the test file so it's not repeated and just pass a custom `decision` string to each test scenario. Not a blocker though

ttbarnes (approved):
LGTM, minor comment. Great work! 🚀

gef-ui/server/utils/helpers.test.js
abhi-markan: Good suggestion.

IgnatG (approved):
LGTM - well done!

PR #551, opened 2021-12-22
DTFS2-4149 : submit MIA to UKEF

abhi-markan:
Allows checker to submit an MIA (previously approved with or without conditions) to UKEF with issued facilities, which converts this into an MIN.

1. Addition of `MIN` in constants
2.  `ukefDecisionAccepted` is now available on `submit-to-ukef-confirmation.njk` for conditional branching confirmation message.
3.  `generateUkefDealId` and `generateUkefFacilityId` functions will only invoke number generator if respective IDs are not already available.
4. `issueAcbsFacilities` ACBS interaction is only invoked the application is either an **MIN** or **AIN**
5. Introduction of `dealHasAllUkefId` function which evaluates whether deal and its associated facilities have IDs already generated or not, if they are then `submitDealAfterUkefIds` function is called else `submitDealBeforeUkefIds`.
6. Simplification of facilities record creation and update.


portal-api/src/v1/gef/controllers/application-submit.js
```diff
@@ -34,18 +34,32 @@ const generateId = async ({
...
 });
 
-const generateUkefDealId = async (application) => application.ukefDealId || generateId({
-  entityId: application._id,
-  entityType: 'deal',
-  dealId: application._id,
-  user: application.checkerId,
-});
+const generateUkefDealId = async (application) => {
+  if (application.ukefDealId) {
```
ttbarnes, line 38: I think this condition (same for `generateUkefFacilityId` function) should be outside of this because otherwise it's not actually generating one. I'd vote to do it like this, what do you think?

```js
// application-submit.js
const addSubmissionData = async (dealId, existingApplication) =>  {
  // ...
  await addSubmissionDateToIssuedFacilities(dealId);
  // ...

  const submissionData = {
    submissionCount: count,
    submissionDate: date,
    // ...
  };

  if (!existingApplication.ukefId) {
    const { ukefId } = await generateUkefDealId(existingApplication);
    submissionData.ukefId = ukefId;
  }

  return submissionData;
};

```

trade-finance-manager-api/src/v1/controllers/deal.submit.controller.js
```diff
@@ -144,14 +149,33 @@ const submitDealAfterUkefIds = async (dealId, dealType, checker) => {
 
 exports.submitDealAfterUkefIds = submitDealAfterUkefIds;
 
+const dealHasAllUkefId = async (dealId) => {
```
ttbarnes, line 152: small niggle, can we add an S please `dealHasAllUkefIds`

ttbarnes: Can we also move this to the helpers directory, and add unit tests?

trade-finance-manager-api/src/v1/controllers/deal.submit.controller.js
```diff
@@ -144,14 +149,33 @@ const submitDealAfterUkefIds = async (dealId, dealType, checker) => {
...
   const {
     dealId,
     dealType,
     checker,
   } = req.body;
+  let deal;
 
-  const deal = await submitDealBeforeUkefIds(dealId, dealType, checker);
+  if (await dealHasAllUkefId(dealId)) {
+    deal = await submitDealAfterUkefIds(dealId, dealType, checker);
```
ttbarnes, line 175: Do you think it'd be easier on the eye like this?

```js
const canSubmitDealAfterUkefIds = await dealHasAllUkefId(dealId);

if (canSubmitDealAfterUkefIds) {
 // ...
} else {
 // ...
}
```

Just my personal preference though 🤷

trade-finance-manager-api/src/v1/controllers/update-facilities.js
```diff
@@ -19,39 +19,41 @@ const updateFacilities = async (deal) => {
...
         facility,
         facilityExposurePeriod,
         facilityGuaranteeDates,
       );
+      facilityUpdate = {
+        premiumSchedule: facilityPremiumSchedule,
+      };
+    } else if (dealType === CONSTANTS.DEALS.DEAL_TYPE.GEF) {
+      // Fee record is only valid for GEF facilities
+      if (submissionType !== CONSTANTS.DEALS.SUBMISSION_TYPE.MIA) {
```
ttbarnes, line 43: Can we combine the `else if` and `if` ?


```
else if (dealType === CONSTANTS.DEALS.DEAL_TYPE.GEF
         && submissionType !== CONSTANTS.DEALS.SUBMISSION_TYPE.MIA) { ....
```

trade-finance-manager-api/src/v1/controllers/update-issued-facilities.js
```diff
@@ -30,54 +30,60 @@ const updatedIssuedFacilities = async (deal) => {
...
       const facilityStatusUpdate = CONSTANTS.FACILITIES.FACILITY_STATUS_PORTAL.ACKNOWLEDGED;
-      await api.updatePortalFacilityStatus(facilityId, facilityStatusUpdate);
-
-      // add acknowledged flag to Portal facility
-      const portalFacilityUpdate = {
-        hasBeenAcknowledged: true,
-      };
 
-      const updatedPortalFacility = await api.updatePortalFacility(facilityId, portalFacilityUpdate);
+      // Add `hasBeenAcknowledged` flag to non-GEF facility
```
ttbarnes, line 40: I'd say this instead just so it's totally clear

`
// add hasBeenAcknowledged flag and status to BSS/EWCS facilities
`

trade-finance-manager-api/src/v1/controllers/update-issued-facilities.js
```diff
@@ -30,54 +30,60 @@ const updatedIssuedFacilities = async (deal) => {
...
         ...facilityExposurePeriod,
         facilityGuaranteeDates,
-        premiumSchedule: facilityPremiumSchedule,
-        feeRecord,
       };
 
       const updateFacilityResponse = await api.updateFacility(facilityId, facilityUpdate);
-
-      // add the updated properties to the returned facility
-      // to retain flat, generic facility mapping used in deal submission calls.
```
ttbarnes, line 78: I think we should keep this comment in there (but in the new place), incase someone wonders why we do this

ttbarnes (changes requested):
Need more tests 😝

trade-finance-manager-api/src/v1/controllers/update-facilities.js
abhi-markan: My intention was to make this future proof in case we have any other records or computation to be done purely for GEF.

trade-finance-manager-api/src/v1/controllers/deal.submit.controller.js
abhi-markan: Good shout, have updated.

portal-api/src/v1/gef/controllers/application-submit.js
abhi-markan: This would mean I would also need to add an else for returning `existingApplication.ukefId`?

ttbarnes: I don't think so - after this function is called, the GEF mongo query does a "find one and update" so it will only add/update and fields you pass to it. So if the ukefId already exists and you don't pass it in, it won't get deleted it'll just remain the same

trade-finance-manager-api/src/v1/controllers/update-facilities.js
ttbarnes: makes sense 👍

trade-finance-manager-api/src/v1/controllers/deal.submit.controller.js
abhi-markan: Have added.

trade-finance-manager-api/src/v1/helpers/dealHasAllUkefIds.api-test.js
```diff
@@ -0,0 +1,36 @@
...
+    const result = await dealHasAllUkefIds(mockDeal._id);
+    expect(result).toEqual(false);
+  });
+
+  it('Should return TRUE when a deal and facilities have UKEF IDs', async () => {
+    const mockDeal = MOCK_DEAL;
+
+    const result = await dealHasAllUkefIds(mockDeal._id);
+    expect(result).toEqual(true);
+  });
```
ttbarnes, line 35: Can we add a test for if GEF ... if BSS please? It's the only condition not covered

trade-finance-manager-api/src/v1/helpers/dealHasAllUkefIds.js
```diff
@@ -0,0 +1,25 @@
+const {
+  findOneTfmDeal,
+} = require('../controllers/deal.controller');
+const CONSTANTS = require('../../constants');
+
+/**
+ * Verified whether the deal and all its facilities have a UKEF ID
+ * @param {Integer} dealId UKEF deal ID
```
ttbarnes, line 8: it'll be a string 😛

ttbarnes (commented):
LGTM just minor comment and some failing e2e tests

dtfs-central-api/src/v1/controllers/portal/gef-deal/update-deal.controller.js
```diff
@@ -0,0 +1,47 @@
...
+      const updatedDeal = await updateDeal(
+        dealId,
+        dealUpdate,
+      );
+      return res.status(200).json(updatedDeal);
+    }
+
+    return res.status(404).send();
+  });
+};
```
ttbarnes, line 47: can we add some tests?

gef-ui/server/controllers/submit-to-ukef/index.js
```diff
@@ -59,14 +63,23 @@ const createSubmissionToUkef = async (req, res) => {
...
     // Always update with the latest checkers details.
     application.checkerId = user._id;
     await api.updateApplication(dealId, application);
     await api.setApplicationStatus(dealId, CONSTANTS.DEAL_STATUS.SUBMITTED_TO_UKEF);
     // TODO: DTFS2-4706 - add a route and redirect instead of rendering?
     return res.render('partials/submit-to-ukef-confirmation.njk', {
       submissionType: application.submissionType,
-      isNotice: isNotice(application.submissionType),
+      isNotice: isDealNotice,
+      ukefDecisionAccepted,
```
ttbarnes, line 82: Can we update the test for new property and condition? `isDealNotice` could also be extracted and have it's own test

gef-ui/server/controllers/submit-to-ukef/index.js
```diff
@@ -59,14 +63,23 @@ const createSubmissionToUkef = async (req, res) => {
       comments.push(commentObj);
       application.comments = comments;
     }
+
+    /**
+     * Change the application submission type to MIN from MIA,
+     * upon second submission. This is to ensure correct confirmation
+     * message is shown on the NJK template else isNotice() will return false
+     */
+    const isDealNotice = application.ukefDecisionAccepted ? true : isNotice(application.submissionType);
```
ttbarnes, line 72: CouldisDealNotice be extracted and have it's own test?

gef-ui/server/controllers/about-facility/index.js
```diff
@@ -242,13 +242,16 @@ const validateAboutFacility = async (req, res) => {
...
     await api.updateFacility(facilityId, {
       name: body.facilityName,
       shouldCoverStartOnSubmission: isTrueSet(body.shouldCoverStartOnSubmission),
       monthsOfCover: body.monthsOfCover || null,
       coverStartDate: coverStartDate ? format(coverStartDate, DATE_FORMAT.COVER) : null,
       coverEndDate: coverEndDate ? format(coverEndDate, DATE_FORMAT.COVER) : null,
-      coverDateConfirmed: DEAL_SUBMISSION_TYPE.AIN ? true : null,
+      coverDateConfirmed: deal.submissionType === DEAL_SUBMISSION_TYPE.AIN
+        ? true
+        : null,
```
ttbarnes, line 254: i'd add a test for this, existing test is either missing or invalid 🙃

gef-ui/server/controllers/submit-to-ukef/index.js
abhi-markan: Extracted to helper with its own unit test cases.

gef-ui/server/controllers/about-facility/index.js
abhi-markan: Done

portal-api/src/v1/gef/controllers/application-submit.js
```diff
@@ -85,7 +85,9 @@ const addUkefFacilityIdToFacilities = async (dealId) => {
   const facilities = await getAllFacilitiesByDealId(dealId);
 
   await Promise.all(facilities.map(async (facility) => {
-    const { ukefId } = await generateUkefFacilityId(facility);
+    const { ukefId } = !facility.ukefFacilityId
+      ? await generateUkefFacilityId(facility)
+      : facility.ukefFacilityId;
```
ttbarnes, line 90: I found this a bit hard to read and we can avoid unnecessary update here:

```js
await Promise.all(facilities.map(async (facility) => {
  if (!facility.ukefFacilityId) {
    const ukefId = await generateUkefFacilityId(facility);

    const update = {
      ukefFacilityId: ukefId,
    };

    await updateFacility(facility._id, update);
  }
}));
```

trade-finance-manager-api/src/v1/controllers/deal.submit.controller.js
```diff
@@ -111,25 +113,34 @@ const submitDealAfterUkefIds = async (dealId, dealType, checker) => {
...
       // NOTE: this is the one and only time that TFM updates a snapshot.
       // Without this, it would involve additional API calls going around in circles.
       const { dealSnapshot } = await api.updateDealSnapshot(dealId, portalMINUpdate);
 
       updatedDeal.submissionType = dealSnapshot.submissionType;
-      updatedDeal.manualInclusionNoticeSubmissionDate = dealSnapshot.details.manualInclusionNoticeSubmissionDate;
-      updatedDeal.checkerMIN = dealSnapshot.details.checkerMIN;
+      if (dealType === CONSTANTS.DEALS.DEAL_TYPE.GEF) {
+        updatedDeal.manualInclusionNoticeSubmissionDate = dealSnapshot.manualInclusionNoticeSubmissionDate;
+        updatedDeal.checkerMIN = dealSnapshot.checkerMIN;
```
ttbarnes, line 135: Are we able to add a test for this? We have one for BSS https://github.com/UK-Export-Finance/dtfs2/blob/main/trade-finance-manager-api/api-tests/v1/deals/deals-submit.second-submission.api-test.js#L303

trade-finance-manager-api/src/v1/controllers/update-portal-deal-from-MIA-to-MIN.js
```diff
@@ -2,21 +2,36 @@ const CONSTANTS = require('../../constants');
...
   console.log('Updating Portal deal from MIA to MIN');
+  let update;
+  let dealUpdate;
 
-  const dealUpdate = {
-    submissionType: CONSTANTS.DEALS.SUBMISSION_TYPE.MIN,
-    details: {
+  if (dealType === CONSTANTS.DEALS.DEAL_TYPE.GEF) {
+    dealUpdate = {
+      submissionType: CONSTANTS.DEALS.SUBMISSION_TYPE.MIN,
```
ttbarnes, line 12: Same as above here - if we add a test in that second submit API test like for BSS, this new condition will have test coverage 👍

ttbarnes (approved):

ttbarnes (approved):
Great work mate!

PR #554, opened 2021-12-29
GEF deal creation with bank object instead of bankId

ttbarnes:
Instead of GEF deals having: `deal.bankId` string, it now stores the whole bank object in the deal: `deal.bank`.

This is now the same as BSS - although BSS will be updated in another PR so the bank object is on the top level and has the same object name.

This change also simplifies a Central API 'get deal' endpoint, where it used to make an API call to get the bank from `bankId`.

## Changes to Portal "all deals" filters

Currently we have an "all deals" query made from Portal  for 2 views - 1 for BSS, 1 for GEF. Filters are passed to the query which check that the bank id matches the users's bank id. This was being made inside the portal-api by default regardless of the `dealType`. This has been moved to the UI side, so it's just an additional filter passed from the UI and now handles the current differences between BSS and GEF data structure.

In the next PR this can be reverted because the data structure will be the same.

Also, there are 2 failing Portal APi tests (99.9% sure these are flaky) that have been disabled (Agreed with the team). They assert BSS deals sorting order by updated timestamp. This works in the real world, but fails a lot in the API tests. This will all change in 2 PRs time so feels like wasted time trying to fix (already spent time on it with no success).

### DIFF
--- a/gef-ui/server/controllers/application-details/index.test.js
+++ b/gef-ui/server/controllers/application-details/index.test.js
@@ -13,7 +13,6 @@ describe('controllers/application-detaills', () => {
   let mockResponse;
   let mockRequest;
   let mockApplicationResponse;
-  let mockFacilityResponse;
   let mockFacilitiesResponse;
   let mockUserResponse;
   let mockEligibilityCriteriaResponse;
@@ -22,13 +21,12 @@ describe('controllers/application-detaills', () => {
     mockResponse = MOCKS.MockResponse();
     mockRequest = MOCKS.MockRequest();
     mockApplicationResponse = MOCKS.MockApplicationResponseDraft();
-    mockFacilityResponse = MOCKS.MockFacilityResponse();
     mockFacilitiesResponse = MOCKS.MockFacilitiesResponse();
     mockUserResponse = MOCKS.MockUserResponse();
     mockEligibilityCriteriaResponse = MOCKS.MockEligibilityCriteriaResponse();
 
     api.getApplication.mockResolvedValue(mockApplicationResponse);
-    api.getFacilities.mockResolvedValue(mockFacilityResponse);
+    api.getFacilities.mockResolvedValue(mockFacilitiesResponse);
     api.getEligibilityCriteria.mockResolvedValue(mockEligibilityCriteriaResponse);
     api.getUserDetails.mockResolvedValue(mockUserResponse);
   });
--- a/gef-ui/server/controllers/application-details/index.test.js
+++ b/gef-ui/server/controllers/application-details/index.test.js
@@ -4,7 +4,7 @@ import {
 } from '.';
 import api from '../../services/api';
 
-import mocks from '../mocks';
+import MOCKS from '../mocks';
 import CONSTANTS from '../../constants';
 
 jest.mock('../../services/api');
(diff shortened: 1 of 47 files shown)


gef-ui/server/controllers/application-details/index.test.js
```diff
@@ -4,7 +4,7 @@ import {
 } from '.';
 import api from '../../services/api';
 
-import mocks from '../mocks';
+import MOCKS from '../mocks';
```
Zainzzkk, line 7: this lowercase may have been me sorry 😂

Zainzzkk (commented):
LGTM but 2 failing tests

gef-ui/server/controllers/application-details/index.test.js
ttbarnes: np :D

gef-ui/server/controllers/mocks/index.js
```diff
@@ -85,45 +86,50 @@ const MockApplicationResponseDraft = () => ({
...
-      lastName: 'Bloggs',
-      id: 1235,
+const MockApplicationResponseSubmitted = () => {
+  const now = new Date();
+  const yesterday = sub(now, { days: 1 });
+
+  return {
+    _id: '1234',
+    exporter: {},
+    bank: { id: 'BANKID' },
```
abhi-markan, line 98: Should we not also have `bankInternalRefName` inside `bank` object?

gef-ui/server/utils/helpers.js
```diff
@@ -413,11 +413,15 @@ const areUnissuedFacilitiesPresent = (application) => {
    from date of submission
 */
 const facilityIssueDeadline = (submissionDate) => {
-  // converts to timestamp from epoch - '+' to convert from str to int
-  const date = new Date(+submissionDate);
-  const deadlineDate = add(new Date(date), { months: 3 });
+  if (submissionDate) {
+    // converts to timestamp from epoch - '+' to convert from str to int
+    const date = new Date(+submissionDate);
```
abhi-markan, line 418: Instead of using unary plus, shall we use native `Number()` or `parseInt()`?

trade-finance-manager-api/src/v1/controllers/acbs.controller.js
```diff
@@ -36,8 +36,8 @@ const clearACBSLog = async () => {
 
 const createACBS = async (deal) => {
   // Add bank's full details so we can reference partyUrn in function
-  const bankId = deal.dealSnapshot.bankId
-    ? deal.dealSnapshot.bankId
+  const bankId = deal.dealSnapshot.bank
+    ? deal.dealSnapshot.bank.id
     : deal.dealSnapshot.details.owningBank.id;
```
abhi-markan, line 41: `owingBank.id` will it be a separate PR where it will appear under `bank` object?

abhi-markan (changes requested):

gef-ui/server/controllers/mocks/index.js
ttbarnes: `bankInternalRefName` is like a "name of the deal" for the bank. Could move inside bank object, but this is big change

gef-ui/server/utils/helpers.js
ttbarnes: @Zainzzkk ? I did not add this, just moved the comment

trade-finance-manager-api/src/v1/controllers/acbs.controller.js
ttbarnes: Yes will be seperate PR

gef-ui/server/controllers/mocks/index.js
abhi-markan: Okay cool can leave it for another PR

gef-ui/server/utils/helpers.js
abhi-markan: @Zainzzkk happy for me to do this in my current PR?

abhi-markan (approved):

gef-ui/server/utils/helpers.js
Zainzzkk: Yeah that is fine, easier for others to understand whats happening if parseInt actually

PR #558, opened 2021-12-31
GEF deal with maker object instead of userId

ttbarnes:
Also disabled 2x e2e test assertion for BSS and GEF deals table Tried to fix but it's going to get removed very shortly so pointless investing more time into it.


dtfs-central-api/src/v1/controllers/portal/deal/get-deal.controller.js
```diff
@@ -110,21 +109,9 @@ const findOneGefDeal = async (_id, callback) => {
...
-
-    const [
-      facilities,
-      maker,
-    ] = [...promises];
+    const facilities = await findAllGefFacilitiesByDealId(_id);
 
     deal.facilities = facilities;
-
-    if (maker) {
```
abhi-markan, line 125: Should this be removed?

abhi-markan (changes requested):

dtfs-central-api/src/v1/controllers/portal/deal/get-deal.controller.js
ttbarnes: Yes, at this point, with these changes - the deal will already have the maker, so no need to get user from DB collection and manually add to the deal here like it was before 😄

Zainzzkk (approved):
LGTM

abhi-markan (approved):

PR #560, opened 2022-01-04
DTFS2-4812 : Create GEF Customer

abhi-markan:
* `United Kingdom` being set as default country when no country is returned back from companies house API call, usually the case for UK based registered office address.
* Removal redudant of constant file inclusion.


ttbarnes (commented):
test?

abhi-markan:
> unit test?

Have added, thanks for spotting it out.

ttbarnes (approved):

PR #562, opened 2022-01-04
DTFS2-4391 : GEF deal record - multiple currencies

abhi-markan:
* Use GBP converted value for the facility when a deal has multiple currencies in a facility.

### DIFF
--- a/azure-functions/acbs-function/mappings/facility/helpers/get-facility-value.js
+++ b/azure-functions/acbs-function/mappings/facility/helpers/get-facility-value.js
@@ -1,9 +1,9 @@
 const getFacilityValue = (facility, currency) => {
-  if (facility.facilitySnapshot.conversionRate && currency) {
+  if (facility.facilitySnapshot.conversionRate && currency !== '') {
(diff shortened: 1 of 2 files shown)


azure-functions/acbs-function/mappings/facility/helpers/get-facility-value.js
```diff
@@ -1,9 +1,9 @@
 const getFacilityValue = (facility, currency) => {
-  if (facility.facilitySnapshot.conversionRate && currency) {
+  if (facility.facilitySnapshot.conversionRate && currency !== '') {
     return Number(facility.facilitySnapshot.value) / Number(facility.facilitySnapshot.conversionRate);
   }
 
-  if (facility.tfm.facilityValueInGBP && currency) {
+  if (facility.tfm.facilityValueInGBP && currency !== '') {
```
ttbarnes, line 6: Could do `currency.length` instead?

abhi-markan: Actually have removed currency as its no longer needed.

ttbarnes (approved):

PR #571, opened 2022-01-06
DTFS2-4812 : Create ACBS GEF customer record

abhi-markan:
This is a bug patch for DTFS-4812 ticket, which rectifies following:

1. Ability to submit deal to the ACBS when parties URN are added from TFM.
2. Jest unit test case have also been added.


IgnatG (commented):
LGTM

IgnatG (approved):

ttbarnes (approved):
Beautiful

PR #575, opened 2022-01-07
DTFS 4320 Ability to search on the "All facilities" page

IgnatG:


e2e-tests/trade-finance-manager/cypress/integration/journeys/facility/facilities.spec.js
```diff
@@ -1,26 +1,40 @@
...
+  let dealOne;
+  let dealTwo;
+  const dealOneFacilities = [];
+  const dealTwoFacilities = [];
 
   before(() => {
-    cy.deleteTfmDeals(dealId);
-
+    cy.deleteTfmDeals(dealOne?._id);
+    cy.deleteTfmDeals(dealTwo?._id);
```
ttbarnes, line 17: seems there's no need to pass anything here, just a single `cy.deleteTfmDeals();` should work?

ttbarnes (approved):

PR #581, opened 2022-01-11
DTFS2-4341 DTFS2-4342 - Portal "All deals" filters, keyword search

ttbarnes:
## Summary
- MOJ filters in Portal "All Deals" for deal type, submission type, status 
- Keyword search/filter in Portal "All Deals"

#### Rendering all possible filters
- Declare a list of filters (via constants) that should be rendered
- Create helper function `submittedFiltersArray`. This turns the list of filters into an array of objects. Each object can have multiple values; each object represents a single filter/field in DB. Each filter/field value is an array as they can have multiple values - e.g `dealType` can have BSS and GEF.

#### Submitting filters
- Build graphQL query variables and send to the API (Portal API > Central API). The API creates a MongoDB friendly query format and returns data based on the filters.
- UI server adds the submitted filter to the express session.

#### Consuming the filters after submission
- A new filters object is created from the array of filters we created: `submittedFiltersObject`. This allows for form submission of single or multiple values and easy consumption in the UI server as we will have an object like this:
  - `{ submissionType: ['Automatic', 'Manual'], dealType: ['GEF'] }`

#### Rendering MOJ filters components with submitted filters
- A helper function (`dashboardFilters`) consumes the submitted filters and is then consumed by MOJ filters component; adding `checked: true` if it's been submitted.
- Submitted/selected/active filters: Create a simple array of objects with helper function `selectedDashboardFilters`. This has a slightly different format required for MOJ filters component. This data is then consumed in two places:
  - MOJ filters component ("selected flters" sub-component)
  - Custom "selected filters" component displayed in the main page

#### Removing filters
- When selected filters are rendered in the UI, each selected filter generates a HREF with `fieldName` and `fieldValue`. When the user clicks on this, UI endpoint will be called and the filter is removed from the session filters. The main "get deals" is then called again with the updated session.

## Portal API
- Add "db query helper" function to build a mongoDB `$or` query for keyword search. This will create several fields for the query that will return any deals that have a matching value against the user provided "keyword" search.

## Other notes
- Add `FIELD_NAME` constants for fields used in the filters.
- Add `CONTENT_STRINGS` constants for filter titles and any custom values that are displayed in the page, but don't actually exist in the data ("All statuses")
- E2E tests: Added many scenarios for filtering. Every possible scenario is _not_ added, I don't think it's needed right now. For example, one test, checks for applying 2 filters: "MIA deals with draft status". I think it's overkill to test every possible status.
- Updated schema/filters input used for BSS reports. This is now differently formatted and queried than All Deals. Further tech improvements (making things clear) to be done later.

## Tech improvements
- Created a single place to store "default dashboard filters" session. This is used/declared on login and in the "remove all filters" route.
- Extract Portal UI - dashboard/deals logic for filter queries into it's own file
- Extract Portal UI `getUserRoles` function into it's own helper file, add tests
- Re-order Portal UI helpers function to be in alphabetical order
- Remove old Portal UI filter templates that are not used (`deal-filters.njk`, `transaction-advanced-filters.njk`)


portal/templates/dashboard/deals.njk
```diff
@@ -113,53 +81,102 @@
...
+      </div>
+    </div>
+  </div>
+
+{% endblock %}
 
-  {{ pagination.pagination(totalPages = pages.totalPages, currentPage = pages.currentPage, totalItems = pages.totalItems, paginationRoot = '/dashboard/deals' ) }}
 
+{% block scripts %}
+  <script>
```
IgnatG, line 162: can we add this as an external script instead? so we load it like `<script src="path-to-script-in-public-folder"></script>`

portal/server/controllers/dashboard/filters/ui-selected-filters.test.js
```diff
@@ -0,0 +1,94 @@
...
+        CONSTANTS.PRODUCT.BSS_EWCS,
+      ];
+
+      const result = generateSelectedFiltersObject(
+        mockHeading,
+        mockFieldName,
+        mockSubmittedFieldFilters,
+      );
+
+      const expectedFormattedFieldValue = (fieldValue) => fieldValue.replace(/[\s+/]/g, '-').replace('\'', '');
```
IgnatG, line 24: what type of fields do we expect here?

portal/server/controllers/dashboard/filters/ui-selected-filters.js
```diff
@@ -0,0 +1,75 @@
...
+  heading,
+  fieldName,
+  submittedFieldFilters,
+) => ({
+  heading: {
+    text: heading,
+  },
+  items: submittedFieldFilters.map((fieldValue) => {
+    // const formattedFieldValue = fieldValue.replace('/', '-');
+    const formattedFieldValue = fieldValue.replace(/[\s+/]/g, '-').replace('\'', '');
```
IgnatG, line 18: same as the other file

portal/server/controllers/dashboard/filters/query.js
```diff
@@ -0,0 +1,63 @@
+const CONSTANTS = require('../../../constants');
+const CONTENT_STRINGS = require('../../../content-strings');
+const {
+  getUserRoles,
+  isSuperUser,
+} = require('../../../helpers');
+
+const dashboardFiltersQuery = (
```
IgnatG, line 8: could you add in a comment above what the query would look like? some sort of example of what to expect?

portal/templates/index.njk
```diff
@@ -58,5 +58,7 @@
     <script src="/assets/js/govukFrontend.js" type="text/javascript"></script>
     <script src="/assets/js/mojFrontend.js" type="text/javascript"></script>
     <script src="/assets/js/maskedInputs.js" type="text/javascript"></script>
+
+    {% block scripts %}{% endblock %}
```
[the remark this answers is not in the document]
ttbarnes: I followed same aproach as GEF UI - might not need this if the script is moved to a .js file, i'll try this out

portal/server/controllers/dashboard/filters/ui-selected-filters.test.js
ttbarnes: Good spot - i'll move this to a seperate function and add some comments/docs - it's used in a couple of places; one has a comment, others don't

It will format a field/filter value into a 'friendly selector', e.g these:
```
Further Maker's input required
Approved by UKEF (with conditions)
BSS/EWCS
```

will become:
```
Further-Makers-input-required
Approved-by-UKEF-with-conditions
BSS-EWCS
```
which are then used for data-cy attributes

The one in this file I think might be different to others, but they can all use the same one. Will improve 👍

portal/server/controllers/dashboard/filters/query.js
ttbarnes: Sure thing - just realised there's some funcitons that don't have comments/documentation at the top, will add

ttbarnes: Added comments to the missing functions 👍

portal/server/controllers/dashboard/filters/ui-selected-filters.test.js
ttbarnes: Moved this function into a helper file, added comments/docs. Both places using the same function

portal/templates/dashboard/deals.njk
ttbarnes: Done :) Had to add it to index.njk, couldn't get it to work in deals.njk (any ideas?)

portal/templates/index.njk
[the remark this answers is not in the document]
ttbarnes: Removed this as it's loaded via it's own .js now

IgnatG (commented):
LGTM - well done!
I'll approve it once the tests finished

IgnatG (approved):

PR #584, opened 2022-01-11
DTFS2-4054 : 'async' keyword for all ACBS calls

abhi-markan:
`async` keyword has been added for all ACBS API functions


ttbarnes (commented):
Does this not work without `async`? Looks like it should because the returned function is async

abhi-markan:
> Does this not work without `async`? Looks like it should because the returned function is async

It works but for some reason postToACBS() function is not behaving as expected especially when calling them simultaneously.

IgnatG (approved):

PR #586, opened 2022-01-12
Title case for Cash, Loan, Contingent & Bond facility types

IgnatG:
Currently, GEF and BSS have the `facilityType` property very much inconsistent - see below.  And we have to write additional code when they get displayed on the page to change the text to Title format. This will be particularly useful when we combine GEF and BSS facilities under 1 collection, but also when we display all facilities in TFM.

What's changed: 
- `CASH` changed to `Cash`
- `loan` changed to `Loan`
- `CONTINGENT` changed to `Contingent`,
- `bond` changed to `Bond`


gef-ui/server/controllers/facility-confirm-deletion/index.js
```diff
@@ -8,7 +8,7 @@ const facilityConfirmDeletion = async (req, res) => {
 
   try {
     const { details } = await api.getFacility(facilityId);
-    const heading = startCase(FACILITY_TYPE[details.type].toLowerCase());
+    const heading = startCase(FACILITY_TYPE[details.type.toUpperCase()].toLowerCase());
```
ttbarnes, line 11: Do we need to do this now, could it just be `FACILITY_TYPE[details.type]`?

gef-ui/server/controllers/facility-currency/index.js
```diff
@@ -30,7 +30,7 @@ const updateFacilityCurrency = async (req, res) => {
   const { dealId, facilityId } = params;
   const { currency, facilityType } = body;
   const { returnToApplication, status, saveAndReturn } = query;
-  const facilityTypeConst = FACILITY_TYPE[facilityType];
+  const facilityTypeConst = FACILITY_TYPE[facilityType?.toUpperCase()];
```
ttbarnes, line 33: Same as above but here

gef-ui/server/controllers/facility-confirm-deletion/index.js
IgnatG: unfortunately, no. the facilityType has to be upper-cased because it matches the `key` from the constants.
i.e.
`CASH: 'Cash'`
and this `[details.type.toUpperCase()]` matches the `CASH` part

ttbarnes (commented):
Great stuff @IgnatG !

gef-ui/server/controllers/facility-confirm-deletion/index.js
ttbarnes: Ahh I see, makes sense

ttbarnes (approved):

PR #588, opened 2022-01-13
Added `hasBeenIsueed` property to BSS facilities to align them with GEF

IgnatG:
Added the `hasBeenIssued` property to BSS facilities. 
Currently, they have 4 formats - see below. And we need to bring them in-line with GEF. We need this in Portal to filter facilities based on `facilityStage`, but also in TFM `/facilities` page

- `Conditional` facilityStage = `hasBeenIssued: false`
- `Unconditional` facilityStage = `hasBeenIssued: true`
- `Issued` facilityStage = `hasBeenIssued: true`
- `Unissued` facilityStage = `hasBeenIssued: false`


utils/mock-data-loader/bss/facilities/deal-9-facilities.js
```diff
@@ -47,6 +48,7 @@ module.exports = [
 		"facilityType": "Bond",
 		"ukefGuaranteeInMonths": null,
 		"facilityStage": "Issued",
+    "hasBeenIssued": true,
```
ttbarnes, line 51: Indentation here and in above files

portal-api/src/v1/controllers/bonds.controller.js
```diff
@@ -89,6 +90,7 @@ const facilityStageFields = (bond) => {
     modifiedBond['coverEndDate-month'] = null;
     modifiedBond['coverEndDate-year'] = null;
     modifiedBond.uniqueIdentificationNumber = null;
+    modifiedBond.hasBeenIssued = false;
```
ttbarnes, line 93: Great work,a looking great! Can we update or add API test for this (and loan), it's in:

```
portal-api/api-tests/v1/bonds/bonds.api-test.js
portal-api/api-tests/v1/loans/loans.api-test.js
```

Or I can do this if you'd like?

There's a couple of describe blocks, first one is this:

```
describe('when a bond has req.body.facilityStage as `Issued`', () => { ...
```

Probably need to update this and the one below, for bonds and loans 
PS sorry for the API tests on this, they're a bit long/complicated.


Side note: we could simplify some TFM graphQL mappings for facility with this in place, no need to be done right now

IgnatG: thanks @ttbarnes 
I think you have more experience with these tests. if it's not too much work, can I pass it over to you?

utils/mock-data-loader/bss/facilities/deal-9-facilities.js
IgnatG: funny enough... in my editor they are aligned properly :|
I'll format the document though - that should sort it :)

portal-api/src/v1/controllers/bonds.controller.js
IgnatG: @ttbarnes I've added the tests now - let me know if this is what you had in mind

ttbarnes: I was just pulled the branch to do this and yes, perfect :)

ttbarnes (approved):
LGTM great work!

PR #594, opened 2022-01-16
Combined GEF & BSS facilities under one collection

IgnatG:
- removed the `gef-facilities` collection and combined all GEF facilities with BSS

### DIFF
--- a/dtfs-central-api/src/v1/controllers/portal/facility/get-facilities.controller.js
+++ b/dtfs-central-api/src/v1/controllers/portal/facility/get-facilities.controller.js
@@ -14,7 +14,8 @@ exports.findAll = findAll;
 
 const findAllFacilitiesByDealId = async (dealId) => {
   const collection = await db.getCollection('facilities');
-  const facilities = await collection.find({ dealId }).toArray();
+  // BSS facilities
+  const facilities = await collection.find({ dealId, $or: [{ facilityType: 'Bond' }, { facilityType: 'Loan' }] }).toArray();
   return facilities;
 };
 exports.findAllFacilitiesByDealId = findAllFacilitiesByDealId;
--- a/portal-api/src/v1/gef/controllers/facilities.controller.js
+++ b/portal-api/src/v1/gef/controllers/facilities.controller.js
@@ -12,7 +12,7 @@ const {
   calculateGuaranteeFee,
 } = require('../calculations/facility-calculations');
 
-const collectionName = 'facilities';
+const facilitiesCollectionName = 'facilities';
 const dealsCollectionName = 'deals';
 
 exports.create = async (req, res) => {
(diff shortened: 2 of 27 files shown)


abhi-markan:
> * removed the `gef-facilities` collection and combined all GEF facilities with BSS

Just to confirm all the facilities now will be stored in `facilities` collection?

dtfs-central-api/src/v1/controllers/portal/facility/get-facilities.controller.js
```diff
@@ -14,7 +14,8 @@ exports.findAll = findAll;
 
 const findAllFacilitiesByDealId = async (dealId) => {
   const collection = await db.getCollection('facilities');
-  const facilities = await collection.find({ dealId }).toArray();
+  // BSS facilities
+  const facilities = await collection.find({ dealId, $or: [{ facilityType: 'Bond' }, { facilityType: 'Loan' }] }).toArray();
```
abhi-markan, line 18: Should this function also return GEF facilities?

portal-api/src/v1/gef/controllers/facilities.controller.js
```diff
@@ -180,6 +180,11 @@ const facilitiesFilters = (user, filters = []) => {
     };
   }
 
+  // GEF facilities only - TODO: for @Tony
```
abhi-markan, line 183: Should @ttbarnes add his work here for it is merged?

abhi-markan (changes requested):

IgnatG:
> > * removed the `gef-facilities` collection and combined all GEF facilities with BSS
> 
> Just to confirm all the facilities now will be stored in `facilities` collection?

yes, that's correct. we did the same thing for bss & gef deals a few weeks ago. now we have it for facilities as well

dtfs-central-api/src/v1/controllers/portal/facility/get-facilities.controller.js
IgnatG: that call is only for BSS facilities -  and only `Bond` & `Loan` are part of bss facilities

portal-api/src/v1/gef/controllers/facilities.controller.js
IgnatG: not yet. this will be needed for a future ticket that @ttbarnes will be working on. for the time being, I've had to add that condition to avoid breaking the existing workflow - this until we merge the gef and bss facilities in the UI as well (currently it's only in the database where they are merged)

dtfs-central-api/src/v1/controllers/portal/facility/get-facilities.controller.js
abhi-markan: Should we consider renaming the function if it is for BSS onlye `findAllBssFacilitesByDealId`?

IgnatG: I'm going to do that, yes. but not in this PR because there will be a few things moving and I wanted to test it separately.
I was planning to combine some of the files together as well in central-api since now we have both deals and facilities in the same collection and there is no need to have separate folders/files for bss/gef

portal-api/src/v1/gef/controllers/facilities.controller.js
ttbarnes: Nice, yes after my current ticket i'll be merging the collections

abhi-markan (approved):

PR #595, opened 2022-01-16
Simplify the All facilities query for TFM

IgnatG:
- last week we've added `hasBeenIssued` property for BSS facilities. As a result, we can simplify the query for All facilities page
- fixed a typo

### DIFF
--- a/dtfs-central-api/src/v1/controllers/tfm/facility/tfm-get-facilities.controller.js
+++ b/dtfs-central-api/src/v1/controllers/tfm/facility/tfm-get-facilities.controller.js
@@ -94,59 +94,7 @@ exports.getAllFacilities = async (req, res) => {
               // create the `companyName` property - this is inside the `dealSnapshot.exporter` property, NOT `facilities` array
               companyName: '$dealSnapshot.exporter.companyName',
               // create the `hasBeenIssued` property
-              // GEF has this property already, but BSS doesn't - so we need to create them using a switch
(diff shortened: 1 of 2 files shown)


dtfs-central-api/src/v1/controllers/tfm/facility/tfm-get-facilities.controller.js
```diff
@@ -94,59 +94,7 @@ exports.getAllFacilities = async (req, res) => {
...
-                        ],
-                      },
-                      then: true,
-                    },
-
-                  ],
-                  default: false,
-                },
-              },
+              hasBeenIssued: '$$facilities.hasBeenIssued',
```
abhi-markan, line 97: Should it be double $$?

abhi-markan (changes requested):

dtfs-central-api/src/v1/controllers/tfm/facility/tfm-get-facilities.controller.js
IgnatG: yes, because it's referring to the alias/variable defined above. we use single `$` when we refer to the root document field

abhi-markan: Understood thank you for the clarification.

abhi-markan (approved):

PR #603, opened 2022-01-19
Remove original BSS reports and related queries/endpoints/controllers

ttbarnes:
Also disabled GEF and BSS facilities e2e tests in portal - they're not working after changing and removing things in the schema.

Tried to fix, but these will be removed and combined into one in the next ticket, so is a bit pointless spending more time trying to fix.


dtfs-central-api/src/v1/controllers/portal/deal/get-deal.controller.js
```diff
@@ -44,39 +44,6 @@ const extendDealWithFacilities = async (deal) => {
   return mappedDeal;
 };
 
-const queryDeals = async (query, start = 0, pagesize = 0) => {
```
IgnatG, line 47: was this used only for reports? :|

ttbarnes: Honestly i'm not sure anymore 😂 I think some used this, others used "transactions".

e2e-tests/portal/cypress/support/portal-api/api.js
```diff
@@ -45,7 +45,7 @@ module.exports.listAllDeals = (token) => {
     body: JSON.stringify(body),
   }).then((resp) => {
     expect(resp.status).to.equal(200);
-    return resp.body.data.deals.deals;
+    return resp.body.data.allDeals.deals;
```
Zainzzkk, line 48: for the future, is this not counter productive? 'allDeals.deals'

Zainzzkk (approved):
this deletion PR LGTM 😝

e2e-tests/portal/cypress/integration/pages/facilitiesDashboard.js
```diff
@@ -1,5 +1,5 @@
 const page = {
-  visit: () => cy.visit('/dashboard/facilities'),
+  visit: () => cy.visit('/dashboard/facilities/gef'),
```
abhi-markan, line 2: Is this GEF specific test, if so should the file be renamed with GEF?

e2e-tests/portal/cypress/support/portal-api/deleteDeals.js
```diff
@@ -4,11 +4,11 @@ const deleteAllDeals = (token, deals) => {
   if (!deals || !deals.length) return;
 
   deals.forEach(async (deal) => {
-    if (deal.dealType === 'BSS/EWCS') {
+    if (deal.product === 'BSS/EWCS') {
```
abhi-markan, line 7: Is the dealType property being changed to product?

utils/mock-data-loader/clean-all-tables.js
```diff
@@ -36,11 +36,11 @@ const cleanDeals = async (token) => {
       // to use the correct .find({ _id ... }) with or without ObjectId.
       // When BSS and GEF have the same _id generation,
       // they can use the same endpoint.
-      if (deal.dealType === 'BSS/EWCS') {
+      if (deal.product === 'BSS/EWCS') {
```
abhi-markan, line 39: Same as above?

abhi-markan (changes requested):

e2e-tests/portal/cypress/support/portal-api/api.js
ttbarnes: I agree it is a bit counter intuitive - it's handling the response from graphQL.

`response.body.data`  = response object, but from here we get this because of the current schema.

```js
response.body.data = {
  allDeals: {
    count: 100,
    deals: [ ... ],
  },
};
```

`allDeals` is the query name in graphql schema

e2e-tests/portal/cypress/integration/pages/facilitiesDashboard.js
ttbarnes: good point - this is page selectors file rather than an actual test. Could move this around but very soon it will be generic "facilities", feels pointless to move it to be GEF specific or BSS specific right now as it's only going to change

e2e-tests/portal/cypress/support/portal-api/deleteDeals.js
ttbarnes: No in the DB it's still dealType. The "all deals" graphQL query (that gets mapped with MongoDB query in central API), returns dealType as product field. 

I had to update it here because before, it was relying on old "deals" query rather than "all deals" query.

If we want to change this to be dealType instead of product (I think either is good), should be seperate PR

abhi-markan: Agree for consistency should be `dealType` can be future.

abhi-markan (approved):

PR #604, opened 2022-01-19
Fix/DTFS2-4700

ttbarnes:


gef-ui/server/controllers/facility-guarantee/facility-guarantee.js
```diff
@@ -23,15 +25,19 @@ function validateFacilityGuarantee({
     errMsg: 'Select how often your bank will pay the fee to UKEF',
   };
 
-  if (!['in advance', 'in arrears', 'at maturity'].includes(feeType)) {
+  if (![
+    CONSTANTS.FACILITY_PAYMENT_TYPE.IN_ADVANCE,
+    CONSTANTS.FACILITY_PAYMENT_TYPE.IN_ARREARS,
+    CONSTANTS.FACILITY_PAYMENT_TYPE.AT_MATURITY,
+  ].includes(feeType)) {
```
Zainzzkk, line 32: for readability, may it be easier to do the same as in line 35? where array defined as variable and checking if variable includes?

ttbarnes: Great idea, updated 👍

gef-ui/templates/partials/facility-guarantee.njk
```diff
@@ -154,25 +154,25 @@
           },
           items: [
             {
-              value: "in advance",
+              value: "In advance",
```
Zainzzkk, line 157: not a deal-breaker at all, (not sure how this component works) but might it be easier making this array in JS and passing it as a variable to items: ? Can use constants then (if that is what it is using)

ttbarnes: Perhaps, pros and cons to this approach IMO - i'm just updating the values here though think should be done later

IgnatG (approved):

PR #608, opened 2022-01-20
fix - GEF status update emails

ttbarnes:
Sending an email when the deal status changes previously relied on references to uppercase constants and matching uppercase value in the deal. We changed the uppercase deal status to be non-uppercase and this stopped the emails from being sent (no status found, therefore Notify would reject the email because of missing email variables).

There was also no test coverage for this.

This is now fixed, added test coverage.


abhi-markan (dismissed):
Good work on the coverage.

IgnatG (approved):

PR #610, opened 2022-01-21
Align updatedAt timestamp

ttbarnes:
- BSS/EWCS facilities now use `updatedAt` timestamp instead of `lastEdited`
- Various places: use `Date.now()`instead of legacy `now()` function (bonus: including TFM task timestamps)

This is required for better/simpler "All facilities" DB query.


dtfs-central-api/src/v1/swagger-definitions/portal/facility-bss.js
```diff
@@ -16,8 +16,8 @@
  *       createdDate:
  *         type: string
  *         example: '1632392646415'
- *       lastEdited:
- *         type: string
+ *       updatedAt:
+ *         type: integer
  *         example: '1632392778918'
```
IgnatG, line 21: minor thing:
if type is integer, then we can remove the quotes on this one, right?

ttbarnes: Great spot, fixed

utils/mock-data-loader/bss/facilities/deal-9-facilities.js
```diff
@@ -68,7 +68,7 @@ module.exports = [
     'requestedCoverStartDate-year': twoMonths.year,
     bondBeneficiary: '',
     ukefExposure: '360,000.00',
-    lastEdited: nowTimestamp,
+    updatedAt: nowTimestamp,
```
IgnatG, line 71: is this `nowTimestamp` being built using date-fns? or `Date.now()` and does it matter here which one do we use?
(this is for the other files listed above)

ttbarnes: Yes it's using `Date.now()` from date-fns, just renamed it here

IgnatG (approved):
LGTM - nice job!

PR #626, opened 2022-01-26
Simplified the deal status in TFM-API

IgnatG:
- combined `DEAL_STATUS_PORTAL_BSS` and `DEAL_STATUS_PORTAL_GEF` into one constant called `PORTAL_DEAL_STATUS`
- simplified the map-tfm-deal-stage-to-portal-status since GEF and BSS have similar statuses


trade-finance-manager-api/src/constants/tasks.js
```diff
@@ -157,7 +157,7 @@ const MIA = {
 
 const STATUS = {
   TO_DO: 'To do',
-  IN_PROGRESS: 'In progress',
+  IN_PROGRESS_BY_UKEF: 'In progress',
```
ttbarnes, line 160: Not a fan of this one as all task statuses will be "by ukef" if you see what I mean

ttbarnes (commented):
Great stuff! Just a minor comment/point about TFM tasks constants - up to you. Happy to approve

trade-finance-manager-api/src/constants/tasks.js
IgnatG: makes sense. I've updated this now

ttbarnes (approved):

PR #629, opened 2022-01-26
Replaced `specialConditions` with `ukefDecision`

IgnatG:
- replaced  `specialConditions` property with `ukefDecision`


ttbarnes (approved):
Great stuff!

PR #630, opened 2022-01-26
DTFS2-4709 | DTFS2-4710 UKEF decision days remaining to review/send MIN to UKEF report

IgnatG:
- UKEF decision days remaining to review/send MIN to UKEF report - 20 days
- UKEF decision days remaining to review/send MIN to UKEF report - 10 days


portal/server/api.js
```diff
@@ -624,15 +624,27 @@ const gefFacilities = async (start, pagesize, filters, token) => {
   return response.data.gefFacilities || { count: 0, facilities: [] };
 };
 
-const getUnissuedFacilitiesReports = async (token) => {
+const getUnissuedFacilitiesReport = async (token) => {
```
abhi-markan, line 627: Wrap it in a try-catch block please.

portal/server/api.js
```diff
@@ -624,15 +624,27 @@ const gefFacilities = async (start, pagesize, filters, token) => {
...
     url: `${portalApi}/v1/reports/unissued-facilities`,
     headers: {
       Authorization: token,
       'Content-Type': 'application/json',
     },
+  });
+  return response.data;
+};
 
+const getUkefDecisionReport = async (token, payload) => {
```
abhi-markan, line 639: Wrap it in a try-catch block please.

abhi-markan (changes requested):

portal-api/src/v1/gef/models/application.js
```diff
@@ -68,6 +68,7 @@ class Application {
         'facilitiesUpdated',
         'ukefDecisionAccepted',
         'portalActivities',
+        'ukefDecision'
```
ttbarnes, line 71: Is this decision, only added by TFM? I.e, GEF should not be able to change or modify this itself right?

I know it's the api model and TFM will end up calling this - maybe ideally "fields that GEF shouldn't change itself" should be a seperate endpoint/model .... food for thought.

Could we add a comment in the code to mention "gef shouldn't update this itself" kind of thing?

portal/styles/components/separator.scss
```diff
@@ -0,0 +1,7 @@
+.separator-line {
+  position: relative;
+  background-color: #0b0c0c;
```
ttbarnes, line 3: Could use `$govuk-text-colour` here

ttbarnes (commented):
Great stuff - few minor comments

portal/server/api.js
IgnatG: sorted

portal/server/api.js
IgnatG: sorted

portal-api/src/v1/gef/models/application.js
IgnatG: yes, it's only generated from tfm-api. this would/should only be used for api-tests/e2e
I've added a comment in the code.

we can look into separate endpoints as well. it wouldn't be part of this PR though

abhi-markan (approved):

PR #633, opened 2022-01-28
Update TFM deals query and mappings

ttbarnes:
This is for Cedar integration.

- add `hasBeenIssued` to graphQL schema and mappings
- add GEF mappings for deal `exporter.companyName` and `submissionType`
- update query example


abhi-markan (dismissed):

IgnatG (approved):
LGTM

PR #634, opened 2022-01-29
Added self-hosted runner

IgnatG:
- used self-hosted runner instead of Github hosted
- added docker-compose.gha.yml to run the tests on self hosted runners


abhi-markan (changes requested):
PR description please.

ttbarnes (approved):

abhi-markan (approved):

PR #641, opened 2022-02-01
Converted Ref Data Proxy to TypeScript

IgnatG:
- Converted the `reference-data-proxy` folder to TypeScript
- added models/interfaces 
- removed unused code
- updated the folders structure
- updated the `/healthcheck` endpoint


reference-data-proxy/api-tests/v1/api.api-test.ts
```diff
@@ -0,0 +1,21 @@
+// import { findACBSIndustrySector } from '../../src/v1/controllers/industry-sectors.controller';
+
+describe('findACBSIndustrySector lookup', () => {
+  it('should return status code from industry sector lookup', async () => {
+    // const expectedBody = [
```
IgnatG, line 5: I will re-enable this as part of a separate PR

portal-api/src/v1/controllers/reports/unissued-facilities.controller.js
```diff
@@ -98,13 +98,13 @@ exports.findUnissuedFacilitiesReports = async (req, res) => {
         const defaultDate = item.submissionDate || '';
         const setDateToMidnight = (new Date(parseInt(defaultDate, 10))).setHours(0, 0, 1, 0);
         // add 3 months to the submission date - as per ticket
-        const deadlineForIssuing = add(setDateToMidnight, { months: 3 });
+        const deadlineForIssuing = add(setDateToMidnight, { days: 90 });
```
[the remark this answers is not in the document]
IgnatG: yup. it needs to be 90 days in the e2e test as well

reference-data-proxy/.eslintrc.js
```diff
@@ -1,24 +1,65 @@
...
   env: {
     jest: true,
     browser: true,
   },
-  root: true,
+  settings: {
+    // 'import/parsers': {
+    //   '@typescript-eslint/parser': ['.ts', '.tsx'],
+    //   alwaysTryTypes: true,
+    // },
```
ttbarnes, line 20: should we remove this?

reference-data-proxy/src/v1/routes/healthcheck.route.ts
```diff
@@ -0,0 +1,31 @@
...
+const GITHUB_SHA = process.env.GITHUB_SHA || 'undefined';
+
+export const pingMulesoft = async () => 'Not configured';
+
+healthcheck.get('/healthcheck', (req: Request, res: Response) => {
+  const mulesoft = pingMulesoft();
+
+  const data = {
+    uptime: process.uptime(),
+    message: 'Ok',
```
Zainzzkk, line 15: make and use constants instead for strings?

IgnatG: It may not be necessary here because it's only being used once

reference-data-proxy/api-tests/v1/industry-sectors.api-test.ts
```diff
@@ -1,15 +1,13 @@
...
 
 describe('/industry-sectors', () => {
   describe('GET /industry-sectors', () => {
     it('returns a list of industry-sectors', async () => {
       const { status, body } = await get('/industry-sectors');
-
       expect(status).toEqual(200);
       expect(body.industrySectors.length).toBeGreaterThan(1);
-      body.industrySectors.forEach((industrySector) => {
+      body.industrySectors.forEach((industrySector: any) => {
```
ttbarnes, line 10: Would be nice to create custom type definition for industrySector rather than using any

ttbarnes (approved):
Great stuff. There's some "any"'s in places, but not a blocker. We can improve as go along

reference-data-proxy/api-tests/v1/industry-sectors.api-test.ts
IgnatG: yup, I agree.
However we can change it further down the line :)

PR #645, opened 2022-02-01
DTFS2 - Anonymisation of PII

abhi-markan:
Removal of any PII across the platform whilst not hindering the performance of the system.
Any personally identifiable information of an individual or an organisation should be removed, unless explicit written consent is place to do.

If one needs to be mentioned, then please use their initials only.


e2e-tests/portal/cypress/fixtures/templates/aDealWithOneBond.json
```diff
@@ -18,18 +18,18 @@
...
     "indemnifier-address-line-3" : "",
-    "indemnifier-address-line-1" : "27a",
-    "indemnifier-address-line-2" : "Maxwell Road",
-    "indemnifier-address-postcode" : "HA6 2XY",
-    "indemnifier-address-town" : "Northwood",
+    "indemnifier-address-line-1" : "1A",
+    "indemnifier-address-line-2" : "Test Road",
+    "indemnifier-address-postcode" : "SW1A 2HQ",
+    "indemnifier-address-town" : "London",
     "indemnifier-companies-house-registration-number" : "08547313",
```
ttbarnes, line 25: Do we need to remove companies house registration numbers as well?

IgnatG (approved):

PR #662, opened 2022-02-04
Removed the PR limits for Central-API and 3rd Party APIs in GHA

IgnatG:


ttbarnes (commented):
Great stuff! Why are we disabling some api tests? Happy to approve otherwise

abhi-markan (approved):

PR #664, opened 2022-02-04
Align supporting documents for GEF and BSS deals

IgnatG:
- aligned the `supportingInformation` for both GEF and BSS deals - previously GEF was using `supportingInformation` but BSS was using `dealFiles`


ttbarnes (approved):
Beautiful! Just need to remember to tell Rohit about the schema change when it's deployed to dev 👍

PR #669, opened 2022-02-07
Cache Dockerfiles

IgnatG:
- removed the memory restrictions for tfm-api and UI
- cached the Dockerfiles so it takes less time to load on start up


.github/workflows/pr_portal-api.yml
```diff
@@ -178,8 +178,8 @@ jobs:
       # https://github.com/cypress-io/github-action/issues/48
       fail-fast: false
       matrix:
-        # run 3 copies of the job in parallel
-        machines: [ 1, 2, 3 ]
+        # run 1 copy of the job
+        machines: [ 1 ]
```
IgnatG, line 182: in PR, some of these tests are flaky and it looks like this is because we are running the tests in parallel - I've noticed this because locally everything passes.
So I've changed it in GHA as well to run everything on 1 machine

ttbarnes (approved):

abhi-markan (approved):
Well done

PR #672, opened 2022-02-07
DTFS2 - Rectified console statements

abhi-markan:
In this tech-debt ticket I have aim to resolve ambiguous console statements whilst also manage to add / update lint configurations.

##Console
* `console.info` have been updated across the myriads of files to be used for general purpose logging.
* `console.error` have been update to use purely for errors and / or in `catch` block upon an erroneous execution of a statement.
* `console.log` should purely be used for debugging and development purposes going forward, currently only Gov UK frontend files contains `console.log`.

##Lint
* ES Lint configuration has been update to turn off any `console` object usage.
* Lint fixes

### DIFF
--- a/dtfs-central-api/.eslintrc.js
+++ b/dtfs-central-api/.eslintrc.js
@@ -1,22 +1,21 @@
 module.exports = {
-    extends: "airbnb-base",
-    env: {
-      jest: true,
-      browser: true,
-    },
-    root: true,
-    rules: {
-      'max-len': ['error', 160, 2, {
-        ignoreUrls: true,
-        ignoreComments: false,
-        ignoreRegExpLiterals: true,
-        ignoreStrings: true,
-        ignoreTemplateLiterals: true,
-      }],
-      "no-underscore-dangle": ['error', { allow: ['_id'] }],
-      "import/no-named-as-default": 0,
-      "implicit-arrow-linebreak": 0,
-      'no-console': 'off',
-    }
-  };
-  
+  extends: 'airbnb-base',
+  env: {
+    jest: true,
+    browser: true,
+  },
+  root: true,
+  rules: {
+    'max-len': ['error', 160, 2, {
+      ignoreUrls: true,
+      ignoreComments: false,
+      ignoreRegExpLiterals: true,
+      ignoreStrings: true,
+      ignoreTemplateLiterals: true,
+    }],
+    'no-underscore-dangle': ['error', { allow: ['_id'] }],
+    'import/no-named-as-default': 0,
+    'implicit-arrow-linebreak': 0,
+    'no-console': 'off',
+  },
+};
--- a/reference-data-proxy/src/v1/controllers/estore/eStoreApi.controller.ts
+++ b/reference-data-proxy/src/v1/controllers/estore/eStoreApi.controller.ts
@@ -11,7 +11,7 @@ const postToEstore = async (apiEndpoint: string, apiData: any) => {
     return false;
   }
 
-  console.log(`Calling eStore API (/${apiEndpoint})`);
+  console.info(`Calling eStore API (/${apiEndpoint})`);
 
   const response = await axios({
     method: 'post',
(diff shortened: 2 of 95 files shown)


utils/mock-data-loader/temporary-token-handler.js
```diff
@@ -13,11 +13,11 @@ module.exports = async (user) => {
     },
     data: user,
   }).catch((err) => {
-    console.log('failed to create temp user');
-    console.log(`${JSON.stringify(err)}`);
+    console.info('failed to create temp user');
+    console.info(`${JSON.stringify(err)}`);
```
IgnatG, line 17: should these 2 get merged into 1 console? we may need to use `console.error` here?

utils/mock-data-loader/.eslintrc.js
```diff
@@ -1,4 +1,4 @@
-module.exports = {
+eslintrc.jsmodule.exports = {
```
IgnatG, line 1: I'm not sure if this one is working as expected?

abhi-markan: Ah! typo, good spot have fixed.

IgnatG (approved):

utils/mock-data-loader/temporary-token-handler.js
abhi-markan: Good shout, will do.

dtfs-central-api/.eslintrc.js
```diff
@@ -1,22 +1,21 @@
...
+      ignoreUrls: true,
+      ignoreComments: false,
+      ignoreRegExpLiterals: true,
+      ignoreStrings: true,
+      ignoreTemplateLiterals: true,
+    }],
+    'no-underscore-dangle': ['error', { allow: ['_id'] }],
+    'import/no-named-as-default': 0,
+    'implicit-arrow-linebreak': 0,
+    'no-console': 'off',
```
Zainzzkk, line 19: should no-console be on so it trips up in PRs if unexpected console.log?

ttbarnes, line 19: Could we do `no-console`, but allow only info and error? https://eslint.org/docs/rules/no-console#options

reference-data-proxy/src/v1/controllers/estore/eStoreApi.controller.ts
```diff
@@ -11,7 +11,7 @@ const postToEstore = async (apiEndpoint: string, apiData: any) => {
     return false;
   }
 
-  console.log(`Calling eStore API (/${apiEndpoint})`);
+  console.info(`Calling eStore API (/${apiEndpoint})`);
```
Zainzzkk, line 14: does printing the endpoint show the full url? and isn't the url under an .env so kept 'secret'?

trade-finance-manager-ui/server/session-configuration/index.js
```diff
@@ -16,7 +16,7 @@ const sessionConfig = () => {
   };
 
   if (process.env.REDIS_HOSTNAME) {
-    console.log(`Connecting to redis server: redis://${process.env.REDIS_HOSTNAME} `);
+    console.info(`Connecting to redis server: redis://${process.env.REDIS_HOSTNAME} `);
```
Zainzzkk, line 19: do we want to show the hostname here?

utils/mock-data-loader/temporary-token-handler.js
abhi-markan: Updated.

reference-data-proxy/src/v1/controllers/estore/eStoreApi.controller.ts
abhi-markan: Nope just the endpoint, not the URL. Also they are in logs which can only be accessed by us.

dtfs-central-api/.eslintrc.js
abhi-markan: ES lint configuration has been updated.

abhi-markan: ES lint configuration has been updated.

PR #683, opened 2022-02-09
lint fixes

ttbarnes:
Now we have 0 lint issues! 🎉


abhi-markan (dismissed):
Good work

reference-data-proxy/src/database/mongo-client.ts
```diff
@@ -1,4 +1,4 @@
-const { MongoClient } = require('mongodb');
+import { MongoClient } from 'mongodb';
```
IgnatG, line 1: I'm not sure if this works. locally it's throwing an error when we try to connect to the mongo client - on line 9

abhi-markan (approved):

IgnatG (approved):

PR #687, opened 2022-02-10
DTFS2-4615 - Dashbord facility filters

ttbarnes:
## Summary
- Adds the ability to filter facilities by:
  - Facility product (Cash, Contingent, Bond, Loan)
  - Submission/Notice type (AIN, MIN, MIA)
  - Facility stage (Issued, Unissued)

## Changes - Portal UI
- Add facility filters query that picks up req.body values and sends to API (just like deals filtering)
- Add 'selected filters' functionality so that submitted filters are rendered in the UI (just like deals). This includes additional functions to handle `hasBeenIssued` field so that string boolean values (submitted from UI), are converted to boolean values. See `selectedHasBeenIssuedFilters`.
- Differentiate submissionType UI field name filters for deals and facilities:
  - Deal filters query by field `submissionType`
  - Facility filters query by field `deal.submissionType`
  - Both deals and facilities use the same function to generate the submissionType filters, but a custom field name is passed

## Fixes (deal and facility filters)
- Portal API:
  - GraphQL db-helpers: ensure that if a string boolean value is passed, it converts to a boolean
- Portal UI:
  - In the facilities table, only render currency table cell value if both currency ID and value exists
  - Add a default empty `submittedFilters` object to a function that could fail on checking `createdByYou`, if it doesn't exist or no filters are passed

## Tech improvements
- Portal UI:
  - Remove Facilities 'deal filters', keep in 'facility filters' (any deal field queries like `submissionType` get's handled by 
  - Various renamings for filter functions (e.g `dashboardDealsFilters` is now `dealsTemplateFilters`)
- Portal E2E tests:
  - move shared dashboard selectors into shared dashboard page file and dashboardFilters partial.
  
## Note
- Discovered a bug as part of this where if you select multiple fields to filter by - the UI works, but the DB query is not built correctly and so the expected filtered results are not returned. There's a commented out E2E test that will be re-enabled in the next PR when this is fixed.


portal/server/controllers/dashboard/filters/generate-selected-filters.js
```diff
@@ -35,18 +34,59 @@ const generateSelectedFiltersObject = (
...
 
+/**
+ * Map true/false string to Issued/Unissued string.
+ *
+ * @param {string} the submitted filter value
+ * @example ( 'true' )
+ * @returns 'Issued'
+ */
+const mapIssuedValueToText = (hasBeenIssued) => {
+  if (hasBeenIssued === 'true') {
```
IgnatG, line 57: should this be a boolean value?

ttbarnes: Not with how it works at the moment - I know it seems strange; what happens is:

1) Generate filters to display in the page (with expected boolean format) e.g:

```js
{
  type: [
    { text: 'Loan', value: 'Loan' },
    { text: 'Cash' value: 'Cash' }
  ],
{
  hasBeenIssued [
    { text: 'Issued', value: true},
    { text: 'Unissued' value: false }
  ],
}
```
2) Nunjucks renders elements with this - Nunjucks converts boolean values to strings
3) User submits some filters, UI server receives these in `req.body`
4) Submitted filters are added to `req.session`- at this point `hasBeenIssued` values are strings.
5) Query to API (as part of this, string booleans are converted to booleans for 'correct' DB query)
6) Re-render the page with selected filters - the selected filters are sourced from the `req.session`, rather than re-mapping with the sent DB query that has actual booleans


Hope it makes sense! In next PR I need to change and improve how the filters are sent to the API, to fix a bug apart from anything else. Maybe as part of this I can also look at immediately converting any string booleans in the submitted filters, before it's added to the session. Or could do as part of this PR if you'd like?

e2e-tests/portal/cypress/integration/journeys/dashboard/dashboard-deals/dashboard-deals-panel-filter-remove.spec.js
```diff
@@ -34,7 +34,7 @@ context('Dashboard Deals - panel selected filters - remove a filter', () => {
     dashboardFilters.showHideButton().click();
 
     // apply filter
-    dashboardFilters.panel.form.submissionType.MIA.checkbox().click();
+    dashboardDeals.filters.panel.form.submissionType.MIA.checkbox().click();
     dashboardFilters.panel.form.applyFiltersButton().click();
```
Zainzzkk, line 38: should this one be changed?

e2e-tests/portal/cypress/integration/journeys/dashboard/dashboard-deals/dashboard-deals-panel-filter-remove.spec.js
```diff
@@ -61,7 +61,7 @@ context('Dashboard Deals - panel selected filters - remove a filter', () => {
     dashboardFilters.panel.selectedFilters.list().should('not.exist');
```
Zainzzkk, line 61: or this? ignore me if I am being dumb :P

e2e-tests/portal/cypress/integration/journeys/dashboard/dashboard-deals/dashboard-deals-panel-filter-remove.spec.js
ttbarnes: `dashboardDeals`are things specific to deals, `dashboardFilters`is a cypress partial that has generic dashboard filter elements that aren't deals/facilities specific. Do you think it could be made clearer somehow?

Zainzzkk: could it just be filters? dashboardFilters and dashboardDeals.filters looks confusing

portal/server/controllers/dashboard/filters/generate-selected-filters.js
IgnatG: next PR is fine, no problemo :)

e2e-tests/portal/cypress/integration/journeys/dashboard/dashboard-deals/dashboard-deals-panel-filter-remove.spec.js
ttbarnes: Thing is `filters` is a generic name and it is specific to dashboard. How about these instead?:

`dashboard.filters`
`dashboardDeals.filters`
`dashboardFacilities.filters`

But then it's pretty much as the same as what you're saying isn't clear. Not sure what else to suggest  😕

ttbarnes: ^ I've done this in the latest commit, I think it's clearer now. The file names remain the same, but the variable name/import is just `filters` so when e.g `deals.filters` and generic filter elements are used it's clearer imo

IgnatG (approved):
LGTM

PR #691, opened 2022-02-11
Fix the mock-data-loader for BSS deals/facilities

IgnatG:
- the mock data loader was assigning 7 facilities to one BSS deal, while the others had none 
- cleaning up the mock bss deals


abhi-markan (approved):
Nice work

PR #696, opened 2022-02-14
Dashboard filters - use node endpoint instead of GraphQL. Fix multiple field values issue

ttbarnes:
## Summary

- Instead of building graphQL a filter query in the UI, sending to graphQL API, and then API re-mapping the filters, just construct the query as required in the UI, and send to the API.
- Remove GraphQL from portal, portal-api, mock-data-loader. It was only used for "get deals" and "get facilities" and (because of above), was causing complex solutions for filtering

## Other things
- Remove db query helper functions
- Add query deals and query facility endpoints to portal API
- Use more constants for field names constructed in filter queries.
- When facility filters are submitted, sanitise req.body straight away (transform string booleans into booleans). This saves handling this further down the line.
- Ensure that if only 1 filter is passed ("status: All Statuses"), an empty $or filter is _not_ added to the query (which will cause DB query to error)
- This fixes a bug where if multiple filter values were applied (i.e one field but with multiple values), no results would be returned as the MongoDB query was not built correctly.
- Added e2e tests to cover filtering with multiple fields with multiple values.
- Fixed Portal E2E test where date validation fails (validation was updated in main branch and this test did not run for some reason)


utils/mock-data-loader/package-lock.json
```diff
@@ -1,437 +1,9 @@
 {
   "name": "mock-data-loader",
   "version": "1.0.0",
-  "lockfileVersion": 2,
+  "lockfileVersion": 1,
```
IgnatG, line 4: this needs to be `2` . You may need to update your npm version to use the latest one

portal/server/controllers/dashboard/filters/helpers.js
```diff
@@ -22,7 +22,8 @@ const submittedFiltersArray = (allSubmittedFilters = {}) => {
       const submittedValue = submittedFilters[field];
       const fieldHasMultipleValues = Array.isArray(submittedValue);
 
-      if (submittedValue) {
+      // NOTE: some filter values can have a value of false.
```
IgnatG, line 25: what other values do we expect here?

portal/server/controllers/dashboard/facilities/facilities-filters-query.test.js
```diff
@@ -15,50 +16,36 @@ describe('controllers/dashboard/facilities - filters query', () => {
...
+    };
 
     expect(result).toEqual(expected);
   });
 
   it('should add multiple custom filters to the query', () => {
     const mockFilters = [
-      { type: ['Cash', 'Bond'] },
-      { hasBeenIssued: ['true'] },
+      { [CONSTANTS.FIELD_NAMES.FACILITY.TYPE]: ['Cash', 'Bond'] },
```
IgnatG, line 30: can we add constants for `Cash` and `Bond` as well?

portal/package-lock.json
```diff
@@ -1,12563 +1,8 @@
 {
   "name": "dtfs2",
   "version": "1.0.0",
-  "lockfileVersion": 2,
+  "lockfileVersion": 1,
```
IgnatG: this needs to be `2` - you may need to update npm to the latest version

portal-api/package-lock.json
```diff
@@ -1,9225 +1,8 @@
 {
   "name": "dtfs-portal-api",
   "version": "1.0.0",
-  "lockfileVersion": 2,
+  "lockfileVersion": 1,
```
IgnatG: same as the other package lock files

portal/server/controllers/dashboard/filters/helpers.js
ttbarnes: It will either be a string (could be any string value), true or false. If it's false then `if(submittedValue)` fails because it's javascript, so have to add an "or if value is false".

Added the comment so it's clear, does it make sense?

portal/package-lock.json
ttbarnes: Updated 👍

PR #697, opened 2022-02-15
DTFS 5405 5404 5403 5402 5401: Fix bugs on Reports page

IgnatG:
- DTFS2-5405 Portal - Unissued facilities report - figures after decimal place are missing
- DTFS2-5404 Portal - Unissued facilities report download - facility ID's missing 00 at start
- DTFS2-5403 Portal - Unissued facilities report - Overdue numbers do not match
- DTFS2-5402 Portal reports - when the MIN is accepted the records are not removed from the report
- DTFS2-5401 Portal reports - Number of decisions text has not been updated


abhi-markan (approved):
Nice work

PR #698, opened 2022-02-15
DTFS2-5300: TFM - "Value (export currency)" field missing the value after decimal place

IgnatG:
- TFM - "Value (export currency)" field missing the value after decimal place - I've used the `comma-number` library to generate this instead


abhi-markan (dismissed):
Approved

PR #702, opened 2022-02-15
DTFS2-4585 : Facility issue - Loan & Fee records creations

abhi-markan:
### Introduction
Upon issuance of a facility, its `loan` and `fee` records must be created. This ticket aims to deliver respective ACBS records creations, while keeping the mapping functions consistent for both creation and update. 

* Reduction of arguments when issuing facility in `acbs.controller`.
* Optimised `deal` object for required properties only.
* `formatDate` now evaluates EPOCH date in string or integer, mitigates returning Invalid date.
* Removal of unwanted argument from `getIncomeClassCode` helper function.
* Improved documentation.

### DIFF
--- a/trade-finance-manager-api/src/v1/controllers/acbs.controller.js
+++ b/trade-finance-manager-api/src/v1/controllers/acbs.controller.js
@@ -122,7 +122,16 @@ const issueAcbsFacilities = async (deal) => {
     // Only concerned with issued facilities on Portal that aren't issued on ACBS
     const facilityStageInAcbs = facility.tfm.acbs && facility.tfm.acbs.facilityStage;
     return !isIssued(facilityStageInAcbs) && !facility.tfm.acbs.issuedFacilityMaster && isIssued(facility.facilityStage);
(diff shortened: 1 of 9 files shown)


trade-finance-manager-api/src/v1/controllers/acbs.controller.js
```diff
@@ -122,7 +122,16 @@ const issueAcbsFacilities = async (deal) => {
...
-  }).map((facility) => api.updateACBSfacility(facility, deal.dealType, deal.exporter.companyName));
+  }).map((facility) => api.updateACBSfacility(facility, {
+    dealSnapshot: {
+      dealType: deal.dealType,
+      submissionType: deal.submissionType,
+      submissionDate: deal.submissionDate,
+    },
+    exporter: {
+      ...deal.exporter,
+    },
```
ttbarnes, line 133: could just be 


```js
exporter: deal.exporter
```

ttbarnes (approved):
Great stuff!

PR #704, opened 2022-02-15
DTFS2-4615 - Filter dashboard facilities by keyword. Reusable keyword query functions

ttbarnes:
This adds the ability to filter by keyword in Dashboard facilities, just like deals, but with different fields.

Also as part of this - created a resuable `generateKeywordQuery` function(s) that is consumed by both deals and facilities. Just need to pass in an array of fields and the keyword value, and it'll generate an array of objects with correct format/regex that'll be sent to the API for MongoDB query. 🚀


IgnatG (approved):
Beautiful!

PR #706, opened 2022-02-16
DTFS2-4615 - ability to remove dashboard facility filters

ttbarnes:
## Summary
- Remove a single facility filter via filters panel, panel form, or main container 'selected filters' list.
- Remove all facility filters.

## How
- Add remove filter endpoints to facility UI routes.
- Create `removeSessionFilter` function to use in both deals and facilities when "remove a filter" endpoint is called. Sanitise any `req.params` that have string booleans (received from Nunjucks)

## Other necessities
- Move `selectedHasBeenIssuedFilters` and text/value mapping into facilities directory.
- Create `generateSelectedFiltersObjectWithMappedValues` so that a selected filter object will display the intended text value, but a different value to be sent to API. E.g, `hasBeenIssued` with `true` boolean, will have `Issued` text/value displayed, but the href/checkbox value will be `true`
  - Without this, any filters that have a true/false value, will not be removed when sent to the server, because the value will not match the actual value applied. In other words, if a value of "Issued" is passed to the "remove filter" endpoint, it won't work without this mapping - because the actual value (same as the actual data), is `true`, not `Issued`. So we display "Issued" to the user, but the actual value is `true`.

## Bonus
- Use constants for active tab, primary nav and page size for dashboard deals and facilities.

### DIFF
--- a/e2e-tests/portal/cypress/integration/journeys/dashboard/dashboard-facilities/dashboard-facilities-filter-remove-all-filters.spec.js
+++ b/e2e-tests/portal/cypress/integration/journeys/dashboard/dashboard-facilities/dashboard-facilities-filter-remove-all-filters.spec.js
@@ -1,0 +1,91 @@
+const relative = require('../../../relativeURL');
+const MOCK_USERS = require('../../../../fixtures/users');
+const { dashboardFacilities } = require('../../../pages');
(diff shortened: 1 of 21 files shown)


IgnatG (approved):
LGTM - well done!

IgnatG (approved):

PR #709, opened 2022-02-16
DTFS2-2948 : Password reset

abhi-markan:
### Introduction
Upon reset a password user was not required to supply their current password this accounts to a vulnerability.
The bug has been patched alongside with recommendations from pen-test.

*  `sendPasswordUpdateEmail` dispatches a confirmation email to the user whose password has been updated with the timestamp.
* `requireCurrentPassword` supplied with a true value upon reset password.
* Formatted email timestamp.
* `passwordUpdatedAt` field hold timestamp of password update.
* `currentPassword` is now reset upon password update.
* Improved documentation
* Lint fixes


portal-api/src/v1/users/reset-password.controller.js
```diff
@@ -14,6 +14,23 @@ const sendResetEmail = async (emailAddress, resetToken) => {
   );
 };
 
+/**
+ * Send a password update confirmation email with update timestamp.
+ * @param {String} emailAddress User email address
+ * @param {String} timestamp Password update timestamp
+ */
+exports.sendPasswordUpdateEmail = async (emailAddress, timestamp) => {
+  const EMAIL_TEMPLATE_ID = '41235821-7e52-4d63-a773-fa147362c5f0';
```
IgnatG, line 23: just a thought: it looks like in portal-api we have a few template IDs scattered around in multiple files. Can we combine all of them under 1 file at import them where needed? I'm thinking it will be easier to manage in the long term

IgnatG: it won't be necessary for this PR - but maybe you can create a separate PR for that?

abhi-markan: Yes of!course good suggestion I can do it for this PR.

IgnatG (dismissed):

IgnatG (approved):

PR #715, opened 2022-02-17
DTFS2-4783 - update BSS/EWCS data migration mappings. Remove old workflow/type-b files

ttbarnes:
## Summary
- When "import deal into our deals collection" is called, add each facility to the facilities collection.
- Update deal and facility property mappings to match the data model
  - Create `map-deal-root` for all top level properties, move and update mappings
  - Add mapping for `exporter.companyName`
  - Remove usage of existing portal _id as the main deal/facility _id (keeping a reference to it under `dataMigrationInfo.v1_ID`). Use new mongo ID instead.
  - Add facility types (bond/cash etc)

## Other things
- Update mandatory criteria import
- Add extra logs
- Add `npm install typescript` to reference-data Dockerfile
- Remove unused type-b integration functions/mappings
- Move k2-mappings into data-migration directory

### DIFF
--- a/e2e-tests/gef/cypress/integration/page-header.spec.js
+++ b/e2e-tests/gef/cypress/integration/page-header.spec.js
@@ -48,5 +48,15 @@ context('Check GEF GOVUK header displays correctly', () => {
         expect(href).to.equal('/logout');
       });
     });
+
+    it('displays the beta banner correctly', () => {
+      pageBanner.userName().contains(CREDENTIALS.MAKER.username);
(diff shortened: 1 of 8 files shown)


IgnatG (approved):

abhi-markan (changes requested):
Failing test.

abhi-markan (approved):
`log.addError` where does that saves the data?

IgnatG (approved):

PR #719, opened 2022-02-21
DTFS2 - Git Hub Runners

abhi-markan:
* Git Hub Runners changed to `self-hosted` runners.
* Docker compose file specification to `docker-compose.gha.yml`.


Zainzzkk (approved):
LGMTM

PR #738, opened 2022-02-24
DTFS2-5416 - TFM Bug - comment added multiple times

Zainzzkk:
# Problem : 

* multiple duplicate objects added to tfm activities when resubmitting deal from portal -> tfm
* was due to full tfm object being sent to controller -> thinking it was an extra activity to add and adding it

# Changes made : 

* Added check for duplicate entries by checking label and timestamp
* Added to test

### DIFF
--- a/dtfs-central-api/api-tests/v1/tfm/deal/tfm-deal-put-tfm.api-test.js
+++ b/dtfs-central-api/api-tests/v1/tfm/deal/tfm-deal-put-tfm.api-test.js
@@ -419,5 +419,98 @@ describe('/v1/tfm/deal/:id', () => {
       expect(secondCommentUpdate.deal.tfm.activities).toEqual([commentObj]);
       expect(secondCommentUpdate.deal.tfm.activities.length).toEqual(1);
     });
+
+    it('should not add duplicated activity objects when adding multiple duplicate activities with different labels', async () => {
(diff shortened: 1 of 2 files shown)


dtfs-central-api/api-tests/v1/tfm/deal/tfm-deal-put-tfm.api-test.js
```diff
@@ -289,5 +289,135 @@ describe('/v1/tfm/deal/:id', () => {
...
+      const { body: firstCommentUpdate } = await api.get(`/v1/tfm/deals/${dealId}`);
+
+      expect(firstCommentUpdate.deal.tfm.activities).toEqual([commentObj]);
+
+      const secondCommentObj = {
+        type: 'COMMENT',
+        timestamp: 13345665,
+        text: 'test1',
+        author: MOCK_AUTHOR,
+        label: 'Comment added',
```
IgnatG, line 406: can we add a test for a different label as well? essentially, multiple labels, from which at least 2 of them are the same, but  you also have different ones

Zainzzkk: done

IgnatG (approved):

PR #748, opened 2022-03-01
DTFS2-5428 : Password reset verification

abhi-markan:
### Introduction
Upon requesting password reset, neither password nullable nor match verification were in place. This PR aims to mitigate these issues through following updates:

* Current password, new password and confirm new password nullable check.
* New password and confirm new password match check.
* `400` status retuned upon an error message.
* Lint fixes

### DIFF
--- a/portal-api/src/v1/users/routes.js
+++ b/portal-api/src/v1/users/routes.js
@@ -91,7 +91,7 @@ module.exports.create = (req, res, next) => {
 };
 
 module.exports.findById = (req, res, next) => {
-  findOne(req.params._id, (err, user) => { // eslint-disable-line no-underscore-dangle
+  findOne(req.params._id, (err, user) => {
     if (err) {
       next(err);
     } else if (user) {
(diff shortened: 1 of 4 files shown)


IgnatG (approved):

portal-api/src/v1/users/routes.js
```diff
@@ -193,6 +193,66 @@ module.exports.resetPassword = async (req, res) => {
...
+          password: {
+            text: 'Password do not match',
+          },
+          passwordConfirm: {
+            text: 'Password do not match',
+          },
+        },
+      },
+    });
+  }
```
Zainzzkk, line 255: could these be a switch statement so a bit neater? or would it not work?

portal-api/src/v1/users/routes.js
```diff
@@ -193,6 +193,66 @@ module.exports.resetPassword = async (req, res) => {
...
+
+  if (password.trim() !== passwordConfirm.trim()) {
+    return res.status(200).json({
+      success: false,
+      errors: {
```
ttbarnes, line 243: Should we be returning 400 status code for all of these added responses instead of 200? Because errors are returned.

PR #750, opened 2022-03-02
DTFS2-5419 - TFM Bug - cover end date displaying as Invalid date

Zainzzkk:
# Issue : 

* cover end date on tfm deals page was showing as invalid date
* function for formatting date was taking in invalid date and outputting invalid date rather than undefined

# Changes made : 

* in mapGefFacilityDetails, if coverEndDate is null, then it is not set (and is undefined)
* Added test
* Added e2e tests to test if - or not

### DIFF
--- a/trade-finance-manager-api/src/graphql/reducers/mappings/gef-facilities/mapGefFacilityDates.js
+++ b/trade-finance-manager-api/src/graphql/reducers/mappings/gef-facilities/mapGefFacilityDates.js
@@ -26,6 +26,7 @@ const mapGefFacilityDates = (facilitySnapsot, facilityTfm, dealSnapshot) => {
       exposurePeriodInMonths,
     ),
   };
+
   // only set coverEndDate if not null, else is undefined
   if (coverEndDate) {
     mapped.coverEndDate = mapCoverEndDate(


e2e-tests/trade-finance-manager/cypress/integration/journeys/case/deal-page-GEF.spec.js
```diff
@@ -144,6 +151,13 @@ context('User can view a GEF AIN case deal', () => {
   });
 
   describe('facilities table', () => {
+    it('facility coverEndDate should be a dash', () => {
+      const facilityId = dealFacilities._id;
+      const facilityRow = pages.caseDealPage.dealFacilitiesTable.row(facilityId);
+
+      facilityRow.facilityEndDate().contains('-');
+    });
+
```
ttbarnes, line 160: For these 'should be a dash' tests, can we make it clear that "when date doesn't exist ...it should render a dash"

trade-finance-manager-api/src/graphql/reducers/mappings/facilities/mapCoverEndDate.api-test.js
```diff
@@ -50,4 +50,20 @@ describe('mapCoverEndDate', () => {
       expect(yearResult).toEqual('2021');
     });
   });
+
+  describe('when cover end date is invalid', () => {
+    it('should return 4 digit year', () => {
```
ttbarnes, line 55: it's returning undefined, not a 4 digit year?

e2e-tests/trade-finance-manager/cypress/integration/journeys/case/deal-page-GEF.spec.js
Zainzzkk: done

trade-finance-manager-api/src/graphql/reducers/mappings/facilities/mapCoverEndDate.api-test.js
Zainzzkk: done

trade-finance-manager-api/src/graphql/reducers/mappings/facilities/mapCoverEndDate.js
```diff
@@ -3,6 +3,10 @@ const { hasValue } = require('../../../../utils/string');
 const { formatYear } = require('../../../../utils/date');
 
 const mapCoverEndDate = (day, month, year) => {
+  // checks if valid date - not valid if coverEndDate is not set
+  if (Number.isNaN(Number(day)) || Number.isNaN(Number(month)) || Number.isNaN(Number(year))) {
+    return undefined;
+  }
```
ttbarnes, line 9: I think this should be combined into the `hasCoverEndDate` function and should use `Number.isInteger()` instead of `isNaN`.

In fact this approach makes it confusing to me - the `hasValue` checks below check if it's a string. But here we're making sure it's an integer? And then when the `coverEndDate`is set, it transforms day/month/year into a number

What are the expected inputs here for GEF and BSS? Maybe one is string, one is number?

trade-finance-manager-api/src/graphql/reducers/mappings/facilities/mapCoverEndDate.js
```diff
@@ -3,6 +3,10 @@ const { hasValue } = require('../../../../utils/string');
 const { formatYear } = require('../../../../utils/date');
 
 const mapCoverEndDate = (day, month, year) => {
+  // checks if valid date - not valid if coverEndDate is not set
+  if (Number.isNaN(Number(day)) || Number.isNaN(Number(month)) || Number.isNaN(Number(year))) {
```
abhi-markan, line 7: Can we use `Number.isInteger` or `Number.isSpecialInteger` please above looks like a gotcha.

abhi-markan (changes requested):
Good work

trade-finance-manager-api/src/graphql/reducers/mappings/facilities/mapCoverEndDate.js
Zainzzkk: done

trade-finance-manager-api/src/graphql/reducers/mappings/facilities/mapCoverEndDate.js
Zainzzkk: removed

abhi-markan (approved):

e2e-tests/trade-finance-manager/cypress/integration/journeys/case/deal-page-GEF.spec.js
```diff
@@ -144,6 +151,13 @@ context('User can view a GEF AIN case deal', () => {
   });
 
   describe('facilities table', () => {
+    it('facility coverEndDate should be a dash when date doesn\'t exist', () => {
```
ttbarnes, line 154: need to add "when date doesn\'t exist'" to the last modified test in this file. Can do after this PR

trade-finance-manager-api/src/graphql/reducers/mappings/gef-facilities/mapGefFacilityDates.api-test.js
```diff
@@ -71,4 +71,11 @@ describe('mapGefFacilityDates', () => {
 
     expect(result.tenor).toEqual(expected);
   });
+
+  it('should return object with undefined coverEndDate if coverEndDate is null', () => {
```
ttbarnes, line 75: could just say

```
should not map coverEndDate when date is null
```

ttbarnes (approved):
Great stuff

PR #753, opened 2022-03-04
DTFS2-5459 - TFM Bug - Incorrect sign in page labels

Zainzzkk:
# Changes made: 

* Changed headings for sign in fields on TFM
* Added data-cy tags
* Added tests to ensure tags are the same
* Added padding on top of sign in header


IgnatG (approved):
LGTM

PR #754, opened 2022-03-05
Updated the PR tests to use docker-compose.gha.yml and reduced the number of parallel tests

IgnatG:
- I've removed the parallel tests for APIs as these won't run
- Updated the PR tests to use `docker-compose.gha.yml`
- reduced the number of parallel tests for e2e - mainly because it would prevent other PRs from starting their tests


.github/workflows/pr_portal-api.yml
```diff
@@ -189,7 +181,7 @@ jobs:
       # https://github.com/cypress-io/github-action/issues/48
       fail-fast: false
       matrix:
-        machines: [ 1, 2, 3, 4, 5, 6, 7 ]
+        machines: [ 1, 2, 3, 4 ]
```
IgnatG, line 184: 7 machines would work fine if only 1 PR is executing the tests. But if there is more than 1, then those PRs would take longer to run their tests because it needs to wait for the other PRs to finish. 
This is because we have a limited number of ubuntu-latest machines we can connect to

abhi-markan: Cool

abhi-markan (approved):

PR #768, opened 2022-03-08
Add Eligibility Criteria version to GEF deals on creation

ttbarnes:
- Pass and add EC version when GEF deal is created
- lint fixes


IgnatG (approved):
LGTM

PR #770, opened 2022-03-08
Security: Load JS scripts from file

IgnatG:
- removed script tags and replaced them with script files


portal/webpack.common.config.js
```diff
@@ -11,6 +11,9 @@ module.exports = {
     jsEnabled: './scripts/js-enabled.js',
     fileUpload: './scripts/file-upload.js',
     correspondenceAddress: './scripts/correspondence-address.js',
+    ukefexposure: './scripts/ukef-exposure.js',
```
ttbarnes, line 14: could we camelCase it to ukefExposure just to be consistent

PR #771, opened 2022-03-09
Portal - BSS/EWCS - Mandatory Criteria versions/versioning

ttbarnes:
- add all Mandatory Criteria versions to mock data loader.
- align existing Mandatory Criteria version with real data (33 instead of 1).
- add endpoint to get the latest Mandatory Criteria.
- change DB/collection structure so that each version is one document; Previously each document was one criterion.
- ensure Mandatory Criteria ids and versions are an integer instead of string.
- add spreadsheet that could/would display in the UI for historic Mandatory Criteria versions.


IgnatG (approved):
Lgtm

PR #773, opened 2022-03-09
DTFS2-4783 - BSS/EWCS Eligibility Criteria versioning

ttbarnes:
- add all Eligibility Criteria versions to mock-data-loader, make sure it deletes based on version.
- change DB/collection structure so that each version is one document; Previously each document was one criterion.
- add endpoint to get the latest Eligibility Criteria
- update TFM eligibility criteria content strings mappings so this is done via the correct version (only one version atm)
- rename GEF `eligibility.terms` to `eligibility.criteria` (same as BSS)
- ensure that GEF `deal.eligibility` includes the version (same as BSS)
- update mock deals


reference-data-proxy/Dockerfile
```diff
@@ -1,11 +1,19 @@
 # Node Alpine Docker Image
-FROM node:17-alpine3.14
+FROM node:14.18.1-alpine3.13
```
IgnatG, line 2: this should work with v17. does it throw any errors?

ttbarnes: didn't mean to commit this, i'll revert.

IgnatG (approved):

PR #776, opened 2022-03-10
fix BSS deal creation error with mandatory criteria API response

ttbarnes:
BSS UI api call to Mandatory Criteria was not returning `response.data`. This caused BSS deal creation to break.


Zainzzkk (approved):
LGTM

PR #779, opened 2022-03-10
GEF smeType values as real values instead of uppercase values

ttbarnes:
Instead of mapping e.g "MICRO" to "Micro" in the GEF UI and (near future) data migration, just save the values in the DB as what they are.

Also for ACBS smeType mapping function - added another type because GEF and BSS have different values ("non sme", "not sme"). @abhi-markan


IgnatG (approved):

abhi-markan (approved):
Good work

PR #783, opened 2022-03-11
DTFS2-5501 : Global SEO headers

abhi-markan:
### Introduction
GDS requires non gov.uk application should not be allowed to indexed and followed on search engines. While this satisfies GDS requirements this also matches our BR ensuring no deals and its associated contents can be searched, linked and indexed across search engines.

Following measures have been taken to achieve the above requirements:

* Removal of `robots.txt` this would prevent crawlers from accessing the headers thus linked pages could potentially be served on search engines. 
* Removal of `humans.txt` this is frivolous for this project.
* Introduction of server header via global middleware across all the services - `noindex, nofollow, noarchive, noimageindex, nosnippet`
* Removal of code snippets text files.
* Lint fix


portal-api/src/createApp.js
```diff
@@ -23,12 +23,13 @@ const { authRouter, openRouter, authRouterAllowXss } = require('./v1/routes');
 configurePassport(passport);
 
 const app = express();
-// TODO: re-enable Helmet (Jira - 4998)
-// app.use(
```
IgnatG, line 27: we need to re-enable this. can we add it back, but leave it disabled for now?

abhi-markan: Reverted it back

gef-ui/server/azure-hack.txt
```diff
@@ -1,13 +0,0 @@
...
-  let jsonString = JSON.stringify({"associatedApplications": [
-    {"dealId": "131ec5af-72ab-42e9-9739-84438eea3371"}
-  ]});
-  // Using a Buffer is how you prevent Express from adding a charset.
-  // Azure's domain verification barfs if you include one:
-  let b = Buffer.from(jsonString)
-  res.setHeader('Content-Length', b.length.toString())
-  res.write(b)
-  res.end()
-});
```
ttbarnes, line 13: 🤦 I remember this.... good find

ttbarnes (approved):

PR #787, opened 2022-03-13
DTFS2-2949: Cross-Site Request Forgery

IgnatG:
- added csrf package to improve the security for csrf vulnerability
- updated the URL for supporting documents to differentiate between documents that are being uploaded and security details


portal/server/index.js
```diff
@@ -94,6 +102,15 @@ app.use(
 
 app.get('*', (req, res) => res.render('page-not-found.njk', { user: req.session.user }));
 
-console.info(`GITHUB_SHA: ${process.env.GITHUB_SHA}`);
+// error handler
+app.use((err, req, res, next) => {
+  if (err.code === 'EBADCSRFTOKEN') {
+    // handle CSRF token errors here
+    res.status(err.statusCode || 500);
+    res.redirect('/');
```
ttbarnes, line 110: Should this redirection be an error page rather than home page? Thinking of GDS pattern

ttbarnes (approved):
Very minor comment/question otherwise LGTM

portal/server/index.js
IgnatG: I was discussing this the other day with @abhi-markan . We agreed to redirect to the main page for security reasons :)

PR #789, opened 2022-03-14
DTFS2-5521 - portal BSS/EWCS - update facilityStage when manually issuing facility

ttbarnes:
## Summary

- When a BSS/EWCS facility is issued via "issue facility form", declare and render the facility/facility stage as issued straight away before deal submission.
- Previously, a maker would issue a facility in the same manner, but the facility would remain unissued until the deal has been submitted to checker. This was the original V1>V2 design.

## How
- When "issue facility" endpoint is called, add these additional properties to the facility:
  - `facilityStage` (issued)
  - `previousFacilityStage`
- When deal is submitted to checker, do _not_ change the `facilityStage`.
- Edit `canIssueOrEditIssueFacility` function checks so that the "edit issued facility" link will appear only if the issued facility has _not_ been submitted. This was previously only checking for previous and current `facilityStage` (which is now invalid)
- Enable the "proceed to review" button if the `dealHasIssuedFacilitiesToSubmit` flag is true.

## Also
- general lint fixes


abhi-markan (approved):
Good work and well done for lint fixes.

abhi-markan (changes requested):
Few failing tests I'm afraid mate.

abhi-markan (approved):

PR #799, opened 2022-03-15
DTFS2-5121: Supporting documentation - Security details & Exporter licence

IgnatG:
- The next supporting document should be the Security Details page, not deal details


gef-ui/server/controllers/supporting-information/security-details/index.js
```diff
@@ -109,9 +109,6 @@ const postSecurityDetails = async (req, res) => {
 
     await updateApplication(dealId, application);
 
-    if (application.supportingInformation?.requiredFields?.includes('exportLicence')) {
```
Zainzzkk, line 112: does the security-details index.test.js need updating?

IgnatG: it didn't have a test before. so removing it doesn't do anything to that test

Zainzzkk (approved):
LGTM

PR #809, opened 2022-03-17
DTFS2-5478 - Fix feedback emails

Zainzzkk:
# Issues: 

* emails were not sending with notify template
* email address variable was set even when left blank
* emails were not sending if not logged in
* TFM feedback object in DB was empty

# Changes made:

* added template for TFM feedback email
* fixed TFM feedback DB creation
* Changed variables so email sends
* added N/A string if username is null (not logged in) so email sends
* Renamed email variable so is not populated with submitters email address
* fixed some failing tfm-api tests


trade-finance-manager-api/src/v1/controllers/feedback-controller.js
```diff
@@ -40,16 +40,20 @@ exports.create = async (req, res) => {
     submittedBy,
   } = modifiedFeedback;
 
+  if (submittedBy.username === null) {
```
IgnatG, line 43: same as above

portal-api/src/v1/controllers/feedback.controller.js
```diff
@@ -58,6 +58,10 @@ exports.create = async (req, res) => {
     submittedBy
   } = modifiedFeedback;
 
+  if (submittedBy.username === null) {
+    submittedBy.username = 'N/A - not logged in';
```
IgnatG, line 62: is this what James asked for? `N\A - not logged in` ?

Zainzzkk: no, I added this because it breaks the template if passing null

IgnatG: Can we leave it as `N\A` only?

portal-api/src/v1/controllers/feedback.controller.js
```diff
@@ -58,6 +58,10 @@ exports.create = async (req, res) => {
     submittedBy
   } = modifiedFeedback;
 
+  if (submittedBy.username === null) {
```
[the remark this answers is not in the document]
Zainzzkk: done

IgnatG (approved):

PR #811, opened 2022-03-17
GEF facilities - save details as value instead of uppercase constant

ttbarnes:
Instead of saving to the DB:

```js
details: [
  'OTHER',
  'TERM'
]
```

Save the actual value instead and avoid mapping this everywhere.

```js
details: [
  'Other',
  'Term basis'
]
```

Also updated hard coded strings to use constants.


IgnatG (commented):
LGTM - happy to approve once the tests finish

IgnatG (approved):

PR #813, opened 2022-03-17
DTFS2-5552-Facility-Amount

abhi-markan:
## Introduction
When creating any facility record, the amount or the maximum liability should be set to the UKEF exposure rather than the total facility value. Following updates standardise this requirement.

* Improved `getMaximumLiability` helper function.
* Introduction of a new helper function `getLoanMaximumLiability`, responsible for calculating `GEF` facility loan amount (10% of UKEF exposure)

### DIFF
--- a/azure-functions/acbs-function/mappings/facility/facility-master.js
+++ b/azure-functions/acbs-function/mappings/facility/facility-master.js
@@ -56,7 +56,7 @@ const facilityMaster = (deal, facility, acbsData, acbsReference) => {
     facilityIdentifier: facility.facilitySnapshot.ukefFacilityId.padStart(10, 0),
     portfolioIdentifier: CONSTANTS.FACILITY.PORTFOLIO.E1,
     dealBorrowerIdentifier: acbsData.parties.exporter.partyIdentifier,
-    maximumLiability: helpers.getMaximumLiability(facility.facilitySnapshot),
+    maximumLiability: helpers.getMaximumLiability(facility),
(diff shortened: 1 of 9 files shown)


IgnatG (dismissed):

azure-functions/acbs-function/mappings/facility/facility-master.js
```diff
@@ -56,7 +56,7 @@ const facilityMaster = (deal, facility, acbsData, acbsReference) => {
     facilityIdentifier: facility.facilitySnapshot.ukefFacilityId.padStart(10, 0),
     portfolioIdentifier: CONSTANTS.FACILITY.PORTFOLIO.E1,
     dealBorrowerIdentifier: acbsData.parties.exporter.partyIdentifier,
-    maximumLiability: helpers.getMaximumLiability(facility.facilitySnapshot),
+    maximumLiability: helpers.getMaximumLiability(facility),
```
IgnatG, line 59: does the `facility` have the dealType in it? I've seen that in the function we are checking
```
const ukefExposure = dealType === CONSTANTS.PRODUCT.TYPE.GEF
    ? amount * 0.10
    : amount;
```

but if it doesn't have the dealType, then it will always be false

abhi-markan: Above snippet is from `getLoanMaximumLiability` helper function, which has second argument as `dealType`.
` getMaximumLiability` only has one argument as `facility` object.

PR #823, opened 2022-03-21
DTFS2-5414 - GEF data migration

ttbarnes:
- add environment variables for V1 storage account
- fix issue where `probabilityOfDefault` could be saved as string
- change saved BSS migration info object from `deal.dataMigrationInfo.v1_ID` to `deal.dataMigration.drupalDealId`
- create import GEF deal controller (also imports facilities and calls `cloneAzureFilesGef`)
- create `migrate-azure-files` for GEF deals/files
- change GEF eligibility controller route/controller to get EC by version instead of _id
- add a flag in GEF facility validation so that `name` is *not* required if the facility is a migrated (V1) facility
- update data-migration .gitignore so that data or dump directories are not added
- update README
- move various migration related files around, seperating BSS and GEF
- extract existing BSS migration files/functions so that GEF and BSS use the same functions, e.g migration logs, teardown function
- create migration-map so map V1 GEF values/field names to V2
- change migration `getUserByEmail` to actually get a user by email instead of username
- improve mock GEF mandatoryCriteria draft version


utils/data-migration/gef/migrate-deal-files.js
```diff
@@ -0,0 +1,67 @@
+const {
```
IgnatG, line 1: Do we need this file? We have in portal-api the migration script for files :)
(not a show stopper - we can do it after the merge)

IgnatG (approved):
LGTM - amazing work!

IgnatG (approved):

PR #824, opened 2022-03-21
DTFS2-5464 - Portal Bug - Portal facilities issued stage

Zainzzkk:
# Issue: 

* Facilities in dashboard facilities table were showing issued when not yet submitted to UKEF

# Changes made: 

* Added hasBeenIssuedAndAcknowledged flag to logic so when that is set to true, then issued shows
* Added e2e test
* Added mapping for this extra flag in portal-api query


portal/templates/dashboard/_macros/dashboard-facilities-table.njk
```diff
@@ -25,7 +25,7 @@
       {% endif %}
     {% endset %}
 
-    {% if facility.hasBeenIssued %}
+    {% if facility.hasBeenIssued and facility.hasBeenIssuedAndAcknowledged %}
```
ttbarnes, line 28: Are we sure we want this - last week had a requirement change where an Issued facility should show as Issued on the BSS deal page, before it submitted to UKEF?

Zainzzkk: @IgnatG

e2e-tests/portal/cypress/integration/journeys/dashboard/fixtures/MIA-deal-with-accepted-status-issued-facilities-cover-start-date-in-past.js
```diff
@@ -0,0 +1,462 @@
+const now = new Date().valueOf();
+const nowMinusDay = () => {
+  const date = new Date();
+  date.setDate(date.getDate() - 1);
+  return date.valueOf();
+};
+
+const deal = {
```
ttbarnes, line 8: is there no existing mock we can use for this and extend?

portal/templates/dashboard/_macros/dashboard-facilities-table.njk
IgnatG: interesting!
@Zainzzkk  can you clarify with Clara?

PR #827, opened 2022-03-21
DTFS2-5525 - add new TFM task - Check adverse history check, fix BSS EC issue

ttbarnes:
Add new task to:
- constants
- mock tasks
- task email template switch-case

Also: 
- bump `redirectionLimit` from 50 to 60 in TFM e2e tests. Required because extra tasks increases the amount of redirects in e2e test.
- fixed an issue in Portal API where creating a BSS deal with Eligibility Criteria passed in via `req.body` would cause incorrect question/answer mapping in the deal/DB.


e2e-tests/trade-finance-manager/cypress.json
```diff
@@ -16,7 +16,7 @@
   "testFiles": "**/*.spec.js",
   "responseTimeout": 100000,
   "pageLoadTimeout": 120000,
-  "redirectionLimit": 50,
+  "redirectionLimit": 60,
```
IgnatG, line 19: do we need this?

ttbarnes: Yes, because an additional task has been added. One test for example checks for completing each task in sequential order. This means that each task goes into a new route (form) and the back to the tasks page. For a manual deal in one test spec it'll redirect at least 20 times for example. 

Without this change cypress throws an error saying there's too many redirects 👍

IgnatG (approved):

PR #847, opened 2022-03-23
DTFS2-5576 : Facility records bugs

abhi-markan:
## Bugs
- Facility **master** record
Upon facility master record creation `premiumFrequencyCode` was not being set to the correct value when fee type is `At Maturity`, due to logic absence.

- Facility **loan** record
In light of recent ACBS UAT testing, previous BR provided was wrong, where the facility loan amount was being set to disbursement amount for `Loan`, this should be set to UKEF exposure of the facility.

## Resolutions
- Facility **master** record
  * If fee type is `At Maturity` then fee frequency for the facility will be set to `Annually`.
  * Introduction of relevant constant.
  * Function documentation.

- Facility **loan** record
Facility UKEF exposure will be set for both `Loan` and `Bond`, while `10%` logic remains for GEF.


Zainzzkk (commented):
are there any tests for these?

IgnatG (approved):

abhi-markan:
> are there any tests for these?

Unfortunately we can't run tests for durable functions effectively.

Zainzzkk (approved):
LGTM

IgnatG (approved):

PR #848, opened 2022-03-23
DTFS2-5502: Login functionality using Username & Password for TFM

IgnatG:
### Changelog:
- Updated the login functionality for TFM - now it requires a username and password
- updated the e2e tests to include the new functionality (this inside `submit-to-tfm` and `trade-finance-manager` folders)
- added passwords for tfm users in mock data loader

### Other changes
- updated the package-lock files to fix the security issues reported by dependabot 
- removed unused files
- fixed the gef_ui, portal_api and portal_ui PR files as they were not pulling the correct image for tfm-api
- created a folder called `e2e-fixures` at the root of `e2e-tests` folder. this will be updated in the future with more files. The general idea is to reduce the number of duplicated files/content.


trade-finance-manager-api/src/v1/controllers/user/user.controller.js
```diff
@@ -0,0 +1,103 @@
+const { ObjectId } = require('mongodb');
+const db = require('../../../drivers/db-client');
+const { BLOCKED, ACTIVE } = require('../../../constants/user.constant').DEAL_STATUS;
+const { mapUserData } = require('./helpers/mapUserData.helper');
+const utils = require('../../../utils/crypto.util');
+
+const businessRules = { loginFailureCount_Limit: 5 };
```
ttbarnes, line 7: `loginFailureCount_Limit` naming convention feels strange, not a blocker

trade-finance-manager-ui/server/helpers/generateErrorSummary.helper.js
```diff
@@ -0,0 +1,84 @@
...
+    };
+  });
+
+  return {
+    count: filteredValidationErrorListArray.length,
+    errorList: filteredValidationErrorList,
+    summary,
+  };
+};
+
```
ttbarnes, line 83: Could use `server/helpers/validation.js` as it seems to do the same thing, not a blocker though, can clean up after launch

IgnatG: good shout. I'll update it after the merge though - I have plenty of other things to update anyway :)

trade-finance-manager-ui/server/controllers/user/index.js
```diff
@@ -0,0 +1,39 @@
...
+      user: req.session.user,
+      validationErrors: formattedValidationErrors,
+      requireCurrentPassword: true,
+    });
+  }
+  const message = 'Unable to update the user details: The current user is not logged in';
+  console.error(message);
+  return { status: 400, message };
+};
+
```
ttbarnes, line 38: Could do with a unit test for this file, not a blocker (go live/priorities)

ttbarnes (approved):
Great stuff

abhi-markan (approved):

PR #849, opened 2022-03-23
fix/DTFS2-5056 - TFM tasks - add emailSent flag to a task when an email has been sent

ttbarnes:
## Issue
- When deal is submitted to TFM, the same email (for the first TFM task), can be triggered multiple times; in some cases over 10 times.
- I believe there maybe another reason for this outside of TFM tasks specifically; this has _sometimes_ happened with other emails so makes me think that perhaps TFM submit endpoint or functions are being called more than once. However the first Task email is the most problematic/bad experience - this PR should fix this.
- Note: Duplicate emails has so far been very tricky if not impossible to replicate locally.

## Solution for multiple task emails
- When deal is submitted to TFM and and the first Task email is sent out ("task ready to start" email), add an `emailSent` flag to that first task.
- When a TFM task is updated (outside of deal submission) and a Task email has been successfully sent, map over the tasks and add an `emailSent` to the appropriate task.
- When a task is updated, only send an email if the `emailSent` flag is false.
- When a deal is submitted and it tries to send the "first task ready to start" email, only send this if the first task's`emailSent` flag is false.

Also various tidying up/improve comments.

Hopefully, this should prevent the issue in dev/test/non-local environments.


trade-finance-manager-api/src/v1/content-strings/index.js
```diff
@@ -106,6 +106,57 @@ const DEAL = {
           text: 'Any upfront or arrangement fees, as well as interest rates and risk margin fees are in an approved currency.',
         },
       },
+      2: {
```
IgnatG, line 109: this version should be in main already

ttbarnes: Cheers - when I put the PR up it wasn't in main and was causing some issues - merged main ino there, it's gone now

IgnatG (approved):
LGTM

PR #851, opened 2022-03-24
DTFS2-5585: Remove the MGA links from Portal footer

IgnatG:
- remove all occurrences of MGA links


trade-finance-manager-ui/templates/case/deal/deal.njk
```diff
@@ -92,10 +92,6 @@
         </div>
       </div>
       <div class="govuk-grid-row govuk-!-margin-0 govuk-!-padding-top-6">
-        {# <div class="govuk-grid-column-one-fifth govuk-!-padding-bottom-4 govuk-!-padding-left-0">
```
abhi-markan, line 95: Looks good mate, could you please write a one liner on why it is commented. :-)

abhi-markan (changes requested):
One minor comment. Good work.

trade-finance-manager-ui/templates/case/deal/deal.njk
IgnatG: actually, this was removed already :)
so there is no need for a comment

PR #853, opened 2022-03-24
DTFS2-5571- GEF - Submit to UKEF page with checkbox

Zainzzkk:
# Changes made:

* aligned gef submit to ukef page with BSS
* Added checkbox and removed comments box
* Added validation 
* Updated e2e tests


gef-ui/server/controllers/submit-to-ukef/index.test.js
```diff
@@ -94,6 +94,7 @@ describe('controllers/submit-to-ukef', () => {
   describe('Ascertain correct deal submission type', () => {
     it('Should stay as an AIN', async () => {
       const submissionType = CONSTANTS.DEAL_SUBMISSION_TYPE.AIN;
+      mockRequest.body.confirmSubmitUkef = 'true';
```
IgnatG, line 97: is this because the html component is returning the value as a string?

Zainzzkk: yes, it won't return it as boolean

PR #861, opened 2022-03-27
Validate the `req.params._id` and `_id` property to ensure it's a valid ObjectId

IgnatG:
### Changelog:
- added a validation for `req.params.id` properties in central-api to check whether an `_id` is a valid Object Id. If it's not, then the server crashes because it tries to query the DB with an invalid _id

### DIFF
--- a/dtfs-central-api/src/v1/controllers/portal/deal/add-deal-comment.controller.js
+++ b/dtfs-central-api/src/v1/controllers/portal/deal/add-deal-comment.controller.js
@@ -5,41 +5,47 @@ const { findOneDeal } = require('./get-deal.controller');
 const addDealComment = async (_id, commentType, comment) => {
   const collection = await db.getCollection('deals');
 
-  const findAndUpdateResponse = await collection.findOneAndUpdate(
-    { _id: ObjectId(_id) },
-    {
-      $push: {
-        [commentType]: {
-          $each: [{
-            ...comment,
-            timestamp: Date.now(),
-          }],
-          $position: 0,
+  if (ObjectId.isValid(_id)) {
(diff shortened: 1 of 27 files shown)


dtfs-central-api/src/v1/controllers/portal/deal/add-deal-comment.controller.js
```diff
@@ -5,41 +5,47 @@ const { findOneDeal } = require('./get-deal.controller');
...
-    { _id: ObjectId(_id) },
-    {
-      $push: {
-        [commentType]: {
-          $each: [{
-            ...comment,
-            timestamp: Date.now(),
-          }],
-          $position: 0,
+  if (ObjectId.isValid(_id)) {
```
ttbarnes, line 8: Do we need this condition here? It's done in the POST function below

ttbarnes: Noticed this in all the functions - I imagine it's fail safe with some of the functions that will be called without the "crud endpoint"? Happy to approve.

ttbarnes (approved):

dtfs-central-api/src/v1/controllers/portal/deal/add-deal-comment.controller.js
IgnatG: exactly. it should be handled as soon as the endpoint is called. but sometimes we import those functions in other files and it skips that validation from "crud endpoint". So yes, it's a fail-safe :)

ttbarnes: Brilliant 😄

PR #869, opened 2022-03-28
DTFS2-5397 - DTFS2-5569 - rename Acknowledged by UKEF to Acknowledged

ttbarnes:
This ensures that BSS and GEF are consistent. Also should fix two bugs that seem to be caused by outdated/incorrect conditional statements that rely on "Acknowledged" status.


abhi-markan (approved):
Good work 👍🏼

PR #879, opened 2022-03-29
DTFS2-5508 - Security.txt

abhi-markan:
## Static files
`security.txt` and `thanks.txt` could not be served from the root due to existing route setup which serves all the public files from `assets` directory, thus a new router setup was required.

## Resolution
* Introduction of a new router `staticRoutes.js`.
* Removal of static files `security.txt` and `thanks.txt`.


trade-finance-manager-api/src/v1/controllers/deal.controller.js
```diff
@@ -88,6 +88,7 @@ const submitACBSIfAllPartiesHaveUrn = async (dealId) => {
   2. BSS/EWCS - Check all the parties have a URN
   */
   const allRequiredPartiesHaveUrn = allPartiesHaveUrn(deal);
+  console.log('========', allRequiredPartiesHaveUrn);
```
IgnatG, line 91: we don't need this :)

portal/server/routes/static.js
```diff
@@ -0,0 +1,37 @@
+const express = require('express');
+
+const router = express.Router();
+
+router.get('/security.txt', (req, res) => {
```
IgnatG, line 5: can we not use the `public` folder to server static files? we have other files in there as well

abhi-markan: We can't because public folder is configured to serve files from `/assets/` directory, which means our link will be `/assets/security.txt`.

IgnatG (approved):

PR #895, opened 2022-03-30
DTFS2-Cookie : Secure

abhi-markan:
## Secure
Introduction of global middleware which ascertain whether the `req` has been originated from either `localhost` or not. If localhost then secure attribute is set to `false` else `true`.


gef-ui/server/index.js
```diff
@@ -29,6 +29,17 @@ const cookie = {
   maxAge: 604800000, // 7 days
 };
 
+/**
+ * `secure` cookie for HTTPS environments, whilst `false` for localhost.
+ */
+app.use((req, res, next) => {
+  const { host } = req.headers;
+  if (host) {
+    cookie.secure = Boolean(host.indexOf('localhost'));
```
IgnatG, line 38: I know we discussed this earlier. I wonder though if we can use an environment variable instead? by default it would not be set. but if it's in Azure, then we set it to `true`. what do  you think?

abhi-markan: I looked into `env` variable, do I need to add that env variable in the docker-compose?

IgnatG: not necessary. because we can leave it out by default. and only add it in Azure.
we can say:
if it exists, then set the cookie to `secure` mode.
If it doesn't, then don't do anything - which means we can leave it out of docker compose

abhi-markan: Okay cool I'll update

abhi-markan: Updated to `HTTPS` environment variable.

IgnatG (approved):

PR #897, opened 2022-03-30
DTFS2-ACBS : Facility & Party

abhi-markan:
### Bug
When a `BSS/EWCS` deal is created alongside with bond `beneficiary` and `issuer` details, facility bond records were not being created and thus hampering the creating of other essential ACBS records.

### Resolution
* Property de-structure path correction.
* Void function name correction.

### Improvement
* Improved constant usage across all the parties.

### Bug
Facility issuance execution was not invoking ACBS endpoint, this was due to the absence of `acbs` object inside `facility.tfm`, which can be either due to failed ACBS execution or other intermittent issue.

### Resolution
* Removal of object existence verification (temporary), post-launch this will need to be assessed.
* Removal of respective test (temporary)


trade-finance-manager-api/src/v1/controllers/acbs.controller.js
```diff
@@ -112,11 +112,12 @@ const checkAzureAcbsFunction = async () => {
 };
 
 const issueAcbsFacilities = async (deal) => {
-  if (!deal.tfm || !deal.tfm.acbs) {
```
IgnatG, line 115: I think we still need this check?

IgnatG (dismissed):

IgnatG (approved):

PR #906, opened 2022-04-01
DTFS2-SCSS : Removed Inline CSS

abhi-markan:
## CSP
Due to no inline CSS permitted by `CSP`, all inline CSS have been removed and added to respective `SCSS` files.


Zainzzkk (approved):
LGTM

trade-finance-manager-ui/templates/case/facility/facility.njk
```diff
@@ -26,7 +26,7 @@
 
 
   <div>
-    <h2 class="ukef-heading-grey" style="padding-top:90px">Facility {{ facility.ukefFacilityId }}</h2>
+    <h2 class="ukef-heading-grey" id="facility-ukef-heading-grey">Facility {{ facility.ukefFacilityId }}</h2>
```
IgnatG, line 29: do we not need that `padding-top: 90` in the css file?

IgnatG: ah. never mind. it's there

IgnatG (changes requested):
You need to run `npm build` in every folder first

abhi-markan:
> You need to run `npm build` in every folder first

Apologies, have build and pushed.

PR #910, opened 2022-04-04
fix: fixes for data migration

IgnatG:
### Changelog:
- fixed the mandatory criteria versioning
- exclude `securityDetails` from being migrated with the azure files
- sanitised the filenames to be used for sharepoint


e2e-tests/portal/cypress/integration/journeys/login/reset-password.spec.js
```diff
@@ -20,19 +20,20 @@ context('Reset password', () => {
       resetPassword.resetPasswordError().should('exist');
     });
 
-    it('should redirect to login page on successful request for reset password', () => {
-      resetPassword.emailInput().type('test_no_notify@ukexportfinance.gov.uk');
-      resetPassword.submit().click();
+    // TODO: re-enable the tests once the user is added back in the mock data loader
```
Zainzzkk, line 23: is there a ticket number for this todo?

utils/data-migration/gef/map-facilities.js
```diff
@@ -68,7 +68,7 @@ const mapCoverEndDate = (coverStartDate, exposurePeriod) => {
     months: exposurePeriod,
   });
 
-  return getUnixTime(date);
+  return new Date(getUnixTime(date));
```
abhi-markan, line 71: We are not storing date as EPOCH?

abhi-markan (commented):
A minor clarification please.

utils/data-migration/gef/map-facilities.js
IgnatG: unfortunately not for cover end date.

abhi-markan: Ah okay, can we store it as EPOCH or is it going to be a big job to do so and out of scope for this PR?

abhi-markan (approved):

PR #915, opened 2022-04-05
DTFS2-5467 - Bug - Admin dashboard enable links

Zainzzkk:
# Issue:

* admins were not able to click on application links
* facility links on dashboard did not take user to the facility
* gef facility id required ukef id which did not work for unsubmitted deals

# Changes made:

* added id for bss facility rows
* all facilities dashboard links take you to facility for gef and bss
* changed facility html ids to _id as compared to ukef id
* added facility id to activity object
* updated tests
* added to e2e tests


gef-ui/server/models/application.js
```diff
@@ -62,7 +62,7 @@ class Application {
     try {
       const application = await getApplication(id);
 
-      if (application.bank.id !== user.bank.id) {
+      if (application.bank.id !== user.bank.id && !user.roles.includes('admin')) {
```
IgnatG, line 65: was it just this line that we needed to get admins to see the deals?

Zainzzkk: Yes. Admin doesn't have the matching bank id. So this line should stop admins returning null so they can see the deals and stop other banks from opening other banks deals

PR #919, opened 2022-04-06
Updated the `Change password` page. Validate passwords that are too short

IgnatG:
- ask for the current password only if the user is already logged in
- validate the new password and ensure that it's length is bigger than 8 chars
- updated the e2e-tests/

### DIFF
--- a/portal-api/src/v1/email/index.js
+++ b/portal-api/src/v1/email/index.js
@@ -13,7 +13,7 @@ const sendEmail = async (templateId, sendToEmailAddress, emailVariables) => {
     })
     .then((response) => response)
     .catch((err) => {
-      console.error(`Portal API - Failed to send email to address: ${sendToEmailAddress} with templateId ${templateId}`, err?.response?.data);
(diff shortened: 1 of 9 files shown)


portal-api/src/v1/email/index.js
```diff
@@ -13,7 +13,7 @@ const sendEmail = async (templateId, sendToEmailAddress, emailVariables) => {
     })
     .then((response) => response)
     .catch((err) => {
-      console.error(`Portal API - Failed to send email to address: ${sendToEmailAddress} with templateId ${templateId}`, { err });
+      console.error(`Portal API - Failed to send email to address: ${sendToEmailAddress} with templateId ${templateId}`, err?.response?.data);
```
abhi-markan, line 16: Might be good to `err?.response` in case we want to debug the status code.

abhi-markan (changes requested):
One minor comment.

abhi-markan (approved):

PR #920, opened 2022-04-06
DTFS2-Collation : Reduced strength

abhi-markan:
## Introduction
`users` and `tfm-users` were migrated to the database using lower-case username, however the default `findOne()` collation strength is set to `3` which does case based comparison. Due to this user will not be able to log-in when using case-sensitive email address.

## Resolution
* `findOne()` collation strength has been explicitly specified and reduced to `2` (which only performs base and diacritics comparison) thus ignore case-sensitive comparison.


trade-finance-manager-api/src/v1/controllers/user/user.controller.js
```diff
@@ -15,7 +15,7 @@ exports.findOne = async (_id, callback) => {
 exports.findByUsername = async (username, callback) => {
   const collection = await db.getCollection('tfm-users');
 
-  collection.findOne({ username }, callback);
+  collection.findOne({ username }, { collation: { locale: 'en', strength: 2 } }, callback);
```
IgnatG, line 18: e2e test :)

IgnatG (approved):

PR #927, opened 2022-04-08
Allow PIM users to receive email notifications

IgnatG:
- PIM users should receive a copy of AIN/MIN & facilities


abhi-markan (approved):
Nice work

PR #928, opened 2022-04-09
Prevent eStore API calls from creating multiple records in the database

IgnatG:
- added a validation in place to ensure that the `cron-job-logs` collection does not get inundated with records if the number generator gets called multiple times
- formatted the exporterName to allow blank spaces and `&` characters only.


abhi-markan (approved):
Nice work

PR #936, opened 2022-04-13
Prevent migrated deals from reaching TFM

IgnatG:
- added a validation that checks whether the deal has the `dataMigration` property
- if a deal has the `dataMigration` property, then it shouldn't create a record in TFM and it shouldn't trigger ACBS, Estore, etc

### DIFF
--- a/portal-api/src/v1/gef/controllers/application.controller.js
+++ b/portal-api/src/v1/gef/controllers/application.controller.js
@@ -227,17 +227,13 @@ exports.changeStatus = async (req, res) => {
   const enumValidationErr = validatorStatusCheckEnums(req.body);
 
   if (enumValidationErr) {
-    return res.status(422)
-      .send(enumValidationErr);
+    return res.status(422).send(enumValidationErr);
   }
 
   const collection = await db.getCollection(dealsCollection);
-  const existingApplication = await collection.findOne({
-    _id: ObjectId(String(dealId)),
-  });
+  const existingApplication = await collection.findOne({ _id: ObjectId(String(dealId)) });
   if (!existingApplication) {
-    return res.status(404)
-      .send();
+    return res.status(404).send();
   }
 
   const { status } = req.body;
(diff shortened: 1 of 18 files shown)


portal-api/src/v1/gef/controllers/application.controller.js
```diff
@@ -272,28 +263,17 @@ exports.changeStatus = async (req, res) => {
...
 
   // If status of correct type, send update email
-  if (
-    [
-      DEAL_STATUS.READY_FOR_APPROVAL,
-      DEAL_STATUS.CHANGES_REQUIRED,
-      DEAL_STATUS.SUBMITTED_TO_UKEF,
-    ].includes(status)
-  ) {
+  if ([DEAL_STATUS.READY_FOR_APPROVAL, DEAL_STATUS.CHANGES_REQUIRED, DEAL_STATUS.SUBMITTED_TO_UKEF].includes(status)) {
```
abhi-markan, line 271: Should we leave it in separate lines, as more than 3 should be in separate lines (if in future new statuses are added).

trade-finance-manager-api/src/v1/controllers/deal.submit.controller.js
```diff
@@ -163,20 +165,60 @@ const submitDealAfterUkefIds = async (dealId, dealType, checker) => {
...
+  const { dealId, dealType, checker } = req.body;
   let deal;
 
-  const canSubmitDealAfterUkefIds = await dealHasAllUkefIds(dealId);
-
-  if (canSubmitDealAfterUkefIds) {
-    deal = await submitDealAfterUkefIds(dealId, dealType, checker);
+  const { status: canSubmitDealAfterUkefIds, message } = await dealHasAllUkefIds(dealId);
+  // check if the deal has been migrated from Portal V1 to Portal v2
+  if (message === 'Migrated deal') {
```
abhi-markan, line 212: Constant?

abhi-markan (changes requested):
Minor comments

trade-finance-manager-api/src/v1/controllers/deal.submit.controller.js
IgnatG: I don't think we need it for this one. It's only used here and we will remove it once we finish amendments.

abhi-markan (approved):

PR #946, opened 2022-04-20
DTFS2-5671 : Reset password bugs

abhi-markan:
# Introduction
During UAT various user password reset bugs have been reported, this PR aims to resolve those bugs.

## Change password
User actioned change password was disregarding the existence of current password field.
* Route level validation of all empty fields.
* Enhanced error reporting and handling.
* E2E tests.

## Forgot password
User actioned forgot password was disregarding an empty email field input.
* Enhanced error reporting and handling.
* Updated NJK template to adhere to GDS standards.
* Improved E2E test.

## Miscellaneous
* Added `spellcheck`, `type` and `autocomplete` to email address field on user add / edit screen.
* `user` object existence verification.

### DIFF
--- a/e2e-tests/portal/cypress/integration/journeys/login/reset-password.spec.js
+++ b/e2e-tests/portal/cypress/integration/journeys/login/reset-password.spec.js
@@ -28,15 +28,16 @@ context('Password management screens', () => {
       resetPassword.submit().click();
 
       resetPassword.emailInputError().should('exist');
+      resetPassword.emailInputError().contains('Enter an email address in the correct format, for example, name@example.com');
     });
 
-    it('An non-existant email displays error message', () => {
+    it('A non-existant email displays error message', () => {
       resetPassword.emailInput().type('email_is_not_valid@ukexportfinance.gov.uk');
       resetPassword.submit().click();
 
       cy.url().should('eq', relative('/reset-password?passwordreseterror=1'));
-
       resetPassword.resetPasswordError().should('exist');
+      resetPassword.resetPasswordError().contains('There was a problem resetting the password. Please try again.');
     });
 
     it('should redirect to login page on successful request for reset password', () => {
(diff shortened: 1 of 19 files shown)


e2e-tests/portal/cypress/integration/journeys/login/reset-password.spec.js
```diff
@@ -6,18 +6,32 @@ const MOCK_USERS = require('../../../fixtures/users');
...
-    it('An incorrect email displays error message', () => {
-      resetPassword.emailInput().type('aaa');
+    it('Should have email address input, submit and cancel buttons on the page', () => {
+      cy.url().should('eq', relative('/reset-password'));
+      resetPassword.emailInput().should('exist');
+      resetPassword.submit().should('exist');
+      resetPassword.cancel().should('exist');
+    });
+
+    it('Enter an empty email address displays error message', () => {
```
IgnatG, line 26: We should check here what the error message says when the user submits an empty form

e2e-tests/portal/cypress/integration/journeys/login/reset-password.spec.js
```diff
@@ -6,18 +6,32 @@ const MOCK_USERS = require('../../../fixtures/users');
...
+      resetPassword.emailInput().should('be.empty');
+      resetPassword.submit().click();
+
+      resetPassword.emailInputError().should('exist');
+    });
+
+    it('An non-existant email displays error message', () => {
```
IgnatG, line 33: we should check what the message says

portal/templates/reset-password.njk
```diff
@@ -19,19 +21,48 @@
...
+            type: "email",
+            autocomplete: "email",
+            errorMessage: errors.fieldErrors.email and {
+              text: errors.fieldErrors.email.text,
+              attributes: {
+                "data-cy": "reset-password-email-error"
+              }
+            },
+            attributes: {
+              "data-cy": "reset-password-email",
```
IgnatG, line 40: the previous `data-cy` was set to `email`. do we need to change that anywhere else?

portal/templates/user/change-password.njk
```diff
@@ -35,13 +35,18 @@
 
         {% include "_partials/password-create.njk" %}
 
+        {% if user._id %}
+          {% set userId = user._id %}
+        {% else %}
+          {% set userId = '' %}
```
IgnatG, line 41: I assume we should not be able to change the password if the userId is set to `''`?

portal/server/routes/user/profile.js
```diff
@@ -46,18 +46,57 @@ router.get('/:_id/change-password', async (req, res) => {
 router.post('/:_id/change-password', async (req, res) => {
   // Ensure that the user is logged in
   if (req?.session?.user) {
+    let formattedValidationErrors;
+
     const { _id } = requestParams(req);
+    const { currentPassword, password, passwordConfirm } = req.body;
 
-    const { status, data } = await api.updateUser(_id, req.body, req.session.userToken);
+    if (!currentPassword || !password || !passwordConfirm) {
```
IgnatG, line 54: do we have a test for this?

IgnatG (commented):
Looks good. 
We need a few more tests though to ensure that those issues are not happening again. 

Do we have a fix for this btw?
> On the account below I’ve selected “Maker” and “Checker”, these are bank roles, so I should be able to select the bank these roles apply to, but the only option is “Any”.

portal/templates/reset-password.njk
abhi-markan: Nope.

portal/templates/user/change-password.njk
abhi-markan: This is just for the cancel button, since we are using this screen for both the scenarios:
1. User actioned reset password (when `_id` will have the user profile ID)
2. Admin creates an account, user sets the password (when `_id` will be empty)

Above is done simply to ensure `Cancel` button does not have `undefined` in the link.

e2e-tests/portal/cypress/integration/journeys/login/reset-password.spec.js
abhi-markan: Updated.

e2e-tests/portal/cypress/integration/journeys/login/reset-password.spec.js
abhi-markan: Updated.

abhi-markan:
> Looks good. We need a few more tests though to ensure that those issues are not happening again.
> 
> Do we have a fix for this btw?
> 
> > On the account below I’ve selected “Maker” and “Checker”, these are bank roles, so I should be able to select the bank these roles apply to, but the only option is “Any”.

Agree, have added few tests to cover updated scenarios.
No update was required as banks list is available on the prod server, issue can not be reproduced.

portal/server/routes/user/profile.js
abhi-markan: Have added extensive tests covering both user change their own password and password reset (upon forgot password email).

IgnatG (approved):

PR #948, opened 2022-04-21
DTFS2-Infra : Infrastructure updates

abhi-markan:
* Added `START_PAGE_REDIRECT` environment variable to `tfs-prod-portal-ui`.
* Reduced throughput on Cosmo DB throughput on `dev` and `staging` to `100`.
* Added `TZ` environment variable to `dev`, `staging` and `prod`
* `AZURE_DIGITAL_TEST` updated to `AZURE_DIGITAL_STAGING` for consistent naming convention.
* Added `UKEF_INTERNAL_NOTIFICATION` environment variable to `tfs-*-trade-finance-manager-api` across all three environemnts.


.github/workflows/env_prod.yml
```diff
@@ -1846,7 +1846,8 @@ jobs:
               REDIS_PORT="$redis_sslPort" \
               REDIS_KEY="$redis_key" \
               REFERENCE_DATA_PROXY_URL="http://$reference_data_proxy_url" \
-              HTTPS=1
+              HTTPS=1 \
+              START_PAGE_REDIRECT=1
```
IgnatG, line 1850: this would re-deploy the prod infrastructure if we merge it in. 
I'd like to avoid that unless it's really necessary. 
Can we change the triggers for this deployments though to a different branch? `infrastructure-prod` or something like that

PR #950, opened 2022-04-22
DTFS2-5630 : Loan disbursement amount

abhi-markan:
## Introduction
Loan maximum liability amount for facility type `Loan` should take into account UKEF cover percentage, ignoring this would lead to loan amount being higher than the facility amount.

`EWCS` formula has been updated to reflect this requirement.


IgnatG (approved):
LGTM

PR #952, opened 2022-04-22
DTFS2 : Dash if empty

abhi-markan:
## Introduction
Upon an application of a plugin or a function which will return `NaN` as an output `dashIfEmpty` plugin considers it as a valid text thus prevents it from displaying `-`.

* Filter function has been updated to consider `NaN` as an empty value.
* Jest tests cases updates.


Zainzzkk (approved):
LGTM

PR #976, opened 2022-05-04
DTFS2-5710 : Cover start date validation

abhi-markan:
## Bug
Upon facility issuance, if the facility cover start date was previously saved but not submitted an error message is thrown stating **cover start date cannot precede issue date** despite the fact that the cover start date is set to the issuance date.

This was due to cover start date selection was being ignored.

## Resolution
* `shouldCoverStartOnSubmission` is added to compound conditions.
* `coverStartDate` property in the return object returns issue date if `shouldCoverStartOnSubmission` is true otherwise user defined cover start date.
* Added Jest unit test.
* E2E test introducted.

### DIFF
--- a/gef-ui/server/controllers/unissued-facilities/index.test.js
+++ b/gef-ui/server/controllers/unissued-facilities/index.test.js
@@ -189,6 +189,8 @@ describe('postChangeUnissuedFacility()', () => {
     mockRequest.body.facilityType = CONSTANTS.FACILITY_TYPE.CASH;
     mockRequest.body.facilityName = 'UKEF123';
     mockRequest.query.saveAndReturn = 'true';
+    mockRequest.body.shouldCoverStartOnSubmission = 'false';
+
     mockRequest.body['issue-date-day'] = format(now, 'd');
     mockRequest.body['issue-date-month'] = format(now, 'M');
     mockRequest.body['issue-date-year'] = format(now, 'yyyy');
@@ -209,7 +211,7 @@ describe('postChangeUnissuedFacility()', () => {
         coverEndDate: format(tomorrow, 'MMMM d, yyyy'),
         coverStartDate: format(tomorrow, 'MMMM d, yyyy'),
         issueDate: format(now, 'MMMM d, yyyy'),
-        shouldCoverStartOnSubmission: null,
+        shouldCoverStartOnSubmission: false,
         monthsOfCover: 30,
         name: 'UKEF123',
         hasBeenIssued: true,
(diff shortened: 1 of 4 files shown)


gef-ui/server/controllers/unissued-facilities/validation.js
```diff
@@ -220,7 +222,10 @@ const facilityValidation = async (body, query, params) => {
   }
 
   if (coverStartDateIsFullyComplete) {
-    coverStartDate = set(new Date(), { year: coverStartDateYear, month: coverStartDateMonth - 1, date: coverStartDateDay });
+    const value = shouldCoverStartOnSubmission
```
IgnatG, line 225: can we use an `if/else` statement for this? the ternary operator gets harder to read if it's too complex

IgnatG (commented):
It looks good. 
You need an E2E test though - or update an existing one to include this

gef-ui/server/controllers/unissued-facilities/validation.js
abhi-markan: Have updated.

gef-ui/server/controllers/unissued-facilities/validation.js
```diff
@@ -220,7 +222,22 @@ const facilityValidation = async (body, query, params) => {
   }
 
   if (coverStartDateIsFullyComplete) {
-    coverStartDate = set(new Date(), { year: coverStartDateYear, month: coverStartDateMonth - 1, date: coverStartDateDay });
+    let value;
```
IgnatG, line 225: you can assign a default value to `value`. and override it inside the `if` statement. in this scenario you wouldn't need the `else` part

abhi-markan:
> It looks good. You need an E2E test though - or update an existing one to include this

That's all added.

gef-ui/server/controllers/unissued-facilities/index.test.js
```diff
@@ -209,7 +211,7 @@ describe('postChangeUnissuedFacility()', () => {
         coverEndDate: format(tomorrow, 'MMMM d, yyyy'),
         coverStartDate: format(tomorrow, 'MMMM d, yyyy'),
         issueDate: format(now, 'MMMM d, yyyy'),
-        shouldCoverStartOnSubmission: null,
+        shouldCoverStartOnSubmission: false,
```
IgnatG, line 214: should this be a string? i.e. `false`
I've seen in the validation file that you are using something like `shouldCoverStartOnSubmission === 'false'`

IgnatG, line 214: should this be a string? i.e. `false`
I've seen in the validation file that you are using something like `shouldCoverStartOnSubmission === 'false'`

IgnatG: ah. never mind. that's the payload that we send to the server

IgnatG: ah. never mind. that's the payload that we send to the server

IgnatG (approved):

PR #980, opened 2022-05-10
DTFS2-5759 : Migrated deal notification

abhi-markan:
## Introduction
Due to the migration of previous portal deals to latest portal various deals are still eligible for ACBS records creation, but due to the lack of TFM deals existence these deals are bound to fail because of various checks in place before ACBS interaction can be initiated.

## Resolution
* Introduction of `MIGRATED_DEAL_SUBMITTED` email template which will notify internal team upon a migrated deal submission.
* Console info logged with deal id.

## Notes
Despite the fact ACBS interaction function will be called as soon as the deal is either `AIN` or `MIN`, records will not be created due to following reasons:

* **AIN** : Draft AIN deals were all discarded, all new AIN deals will be subjected to Portal V2.
* **MIN** : Deal will not exists in `tfm-deals` collection.
* **Issue facility** : Facility will not exists in `tfm-facilities` and associated deal in `tfm-deals` collection.


trade-finance-manager-api/src/v1/controllers/deal.submit.controller.js
```diff
@@ -196,6 +200,14 @@ const submitMigratedDeal = async (dealId, dealType, checker) => {
       updatedDeal.manualInclusionNoticeSubmissionDate = portalMINUpdate.details.manualInclusionNoticeSubmissionDate;
       updatedDeal.checkerMIN = portalMINUpdate.details.checkerMIN;
     }
+
+    // ACBS interaction : AIN or MIN only
+    if (dealController.canDealBeSubmittedToACBS(portalMINUpdate.submissionType)) {
+      console.info('Migrated deal ACBS interaction initiated: ', dealId);
+      await sendMigratedDealEmail(dealId);
```
IgnatG, line 207: where is this email going to?

trade-finance-manager-api/src/v1/controllers/deal.submit.controller.js
```diff
@@ -196,6 +200,14 @@ const submitMigratedDeal = async (dealId, dealType, checker) => {
...
+      await dealController.submitACBSIfAllPartiesHaveUrn(dealId);
```
IgnatG, line 208: does this work if the deal doesn't exist in tfm?

trade-finance-manager-api/src/v1/controllers/deal.submit.controller.js
IgnatG: you need to specify the email address when you send a notification in Production env

abhi-markan: Me, you and James.

IgnatG: Ok. But you need specify the email address in production otherwise it won't work

trade-finance-manager-api/src/v1/controllers/send-deal-submit-emails.js
```diff
@@ -188,11 +188,29 @@ const sendDealSubmitEmails = async (deal) => {
   };
 };
 
+const sendMigratedDealEmail = async (dealId) => {
+  if (!dealId) {
+    return false;
+  }
+
+  const sendToEmailAddress = process.env.INTERNAL_EMAIL_GROUP;
```
IgnatG, line 196: Is this being triggered locally as well and/or Staging env as well?

IgnatG (approved):

PR #984, opened 2022-05-22
feat(DTFS2-4292 | DTFS2-4293): amendment details page

IgnatG:
### Changelog:
- Ability to review amendments in progress
- Create amendments details page

### Fixes: 
- Amendments - Cover end date - should error when date not completed


e2e-tests/trade-finance-manager/cypress/integration/journeys/case/amendments/show-amendment-button.spec.js
```diff
@@ -100,52 +100,7 @@ context('Amendments page', () => {
     });
   });
 
-  describe('MIN - not confirmed', () => {
```
IgnatG, line 103: note: I've removed this one because it's not relevant anymore. once a MIA changes to MIN, the status changes to `Confirmed` for a MIN deal. this is a new feature added directly from `main` branch

trade-finance-manager-ui/server/controllers/helpers/amendments.helper.js
```diff
@@ -31,15 +12,10 @@ const showAmendmentButton = (deal, userTeam) => {
   const acceptableUser = CONSTANTS.TEAMS.PIM;
   const acceptableStatus = [CONSTANTS.DEAL.DEAL_STAGE.CONFIRMED];
 
-  if (acceptableSubmissionType.includes(deal.dealSnapshot.submissionType) && userTeam.includes(acceptableUser)
-    && acceptableStatus.includes(deal.tfm.stage)) {
+  if (acceptableSubmissionType.includes(deal.dealSnapshot.submissionType) && userTeam.includes(acceptableUser) && acceptableStatus.includes(deal.tfm.stage)) {
```
abhi-markan, line 15: Can we have this compound condition as a helper function please?

trade-finance-manager-ui/styles/table.scss
```diff
@@ -5,3 +5,60 @@
...
+  text-align: center;
+  vertical-align: middle;
+}
+
+.gov-table_cell--border-bottom-transparent {
+  border-bottom: 1px solid transparent;
+}
+
+.facilities-table td {
+  text-align: right !important;
```
abhi-markan, line 23: Do we need `!important`, can we use `id` or ` class` instead please?

trade-finance-manager-ui/templates/case/facility/_macros/amendment-details.njk
```diff
@@ -0,0 +1,141 @@
...
+            </div>
+          </div>
+
+          <div class="govuk-grid-row">
+            <div class="govuk-grid-column-one-quarter">
+              <h4 class="govuk-heading-s govuk-body-s govuk-!-font-weight-bold">UKEF approval required</h4>
+            </div>
+
+            <div class="govuk-grid-column-three-quarters">
+              <p class="govuk-body-s" data-cy="amendment--details-{{amendment.version}}-require-approval">{{ amendment.requireUkefApproval | dashIfEmpty}}</p>
```
abhi-markan, line 41: XSS Filter please `safe`.

abhi-markan (changes requested):

trade-finance-manager-ui/styles/table.scss
IgnatG: it was like this before - I've just removed the other files and moved the code here. I'm not sure why it needs `!important` either

trade-finance-manager-ui/templates/case/facility/_macros/amendment-details.njk
IgnatG: I'm not sure that's needed here. the `requireUkefApproval` is created inside [here](https://github.com/UK-Export-Finance/dtfs2/blob/feature/amendment-details/trade-finance-manager-ui/server/controllers/case/index.js#L187)

the value can only be `Yes` or `No`

trade-finance-manager-ui/server/controllers/helpers/amendments.helper.js
IgnatG: The function is already inside a helper function/file though. and this validation is not being used anywhere else. it only checks for 3 things

trade-finance-manager-ui/styles/table.scss
IgnatG: I found out why - the table is generated using the govuk component. so adding an `id` or `class` to a `td` element is not possible

abhi-markan (approved):

PR #986, opened 2022-05-22
chore: updated nodejs to v.18.2-alpine3.14. & updated all npm packages

IgnatG:
### Changelog:
- updated NodeJS to v18.2-alpine3.14
- updated npm to v8.10
- updated all npm packages
- updated the dockerignore files
- install npm packages using `npm ci` to reduce the risk of vulnerabilities
- install only packages related to `production` for UI services (i.e. gef-ui, portal and tfm-ui). this reduces the size of the Docker image by more than 100mb each
- used `COPY` instead of `ADD` (this to follow docker best practices)


abhi-markan (approved):
Nice work.
All looks good to me especially replacement of `COPY` from `ADD`, though need to thoroughly test this especially with Azure SSH functionality.

PR #989, opened 2022-05-23
feat(DTFS2-5777) : Data migration

abhi-markan:
# Introduction
This data migration PR aims to migrate portal v1 deals into DTFS TFM, following product and deal types and amendments are covered:

## Products
* BSS
* EWCS
* GEF

## Deal Types
* AIN
* MIA
* MIN

## Amendments (Cross products)
* Automatic
* Manual


utils/data-migration/helpers/date.js
```diff
@@ -0,0 +1,38 @@
+/**
+ * Date helper functions
+ */
+
+/**
+ * Returns formatted date string in `DD-MM-YYYY` format from an EPOCH integer argument.
+ * @param {Integer} epoch EPOCH
+ * @returns {String} Formatted date string, `Invalid` if null provided.
+ */
+const getDDMMYYYY = (epoch) => {
```
IgnatG, line 10: just being nosy here, can we use date-fns `format` for this?

utils/data-migration/tfm/migrate-gef-deals.js
```diff
@@ -0,0 +1,180 @@
...
+} = require('../helpers/deal');
+const { ObjectId } = require('mongodb');
+const CONSTANTS = require('../constant');
+
+const version = '0.0.1';
+
+// ******************** DEALS *************************
+
+/**
+ * Return all the portal deals, without (default) or with filter specified.
```
IgnatG, line 26: something to note on this, we don't need all deals from portal. only the ones that have been migrated

utils/data-migration/tfm/migrate-gef-deals.js
```diff
@@ -0,0 +1,180 @@
...
+
+/**
+ * Construct deal's `dealSnapshot` object.
+ * @param {Object} deal Deal object
+ * @returns {Object} dealSnapshot Deal snapshot
+ */
+const buildDealsnapshot = (deal) => (
+  {
+    dealSnapshot: {
+      ...deal,
```
IgnatG, line 59: you probably have this in your list as well, but I'll leave it here as a reference:
you'll need the `_id` and `dealSnapshot._id` as a MongoDB object. `getCollections` will return it as a string only.
you'll also need inside the `dealSnapshot.facilities` object the entire facility - not just the `id` of it

PR #990, opened 2022-05-25
DTFS2-5821 : Unissued facility exposure period

abhi-markan:
# Bug
Upon submission of an `unissued` facility exposure period for the master facility record was not being set and therefore was being set to `0`. This is incorrect as facility cover month can be used in such cases.

# Resolution
##GEF
* `monthsOfCover` value will be used if the facility is unissued otherwise `exposurePeriodInMonths`.
* In any case desired property does not exists `0` value will be returned.

## BSS/EWCS
* `ukefGuaranteeInMonths` value will be referred to in such cases, otherwise value will be calculated based on cover start and end date.
* In any case desired property does not exists `0` value will be returned.
* Returned values are type casted to `Number` for function return data type consistency.
* `coverStartDate` moved to local scope.

## Extra
* Documentation

### DIFF
--- a/azure-functions/acbs-function/mappings/facility/helpers/get-exposure-period.js
+++ b/azure-functions/acbs-function/mappings/facility/helpers/get-exposure-period.js
@@ -3,19 +3,31 @@ const moment = require('moment');
 const CONSTANTS = require('../../../constants');
 const { formatYear, formatTimestamp } = require('../../../helpers/date');
 
+/**
+ * Evalutes facility's exposure period in months.
+ * If `Unissued` then guarantee month value is returned.
+ * If `Issued` then cover start and end date difference in months.
+ * @param {Object} facility Facility object
+ * @param {String} dealType Deal type `GEF`, `BSS`, `EWCS`
+ * @returns {Integer} Exposure period in months
+ */
 const getExposurePeriod = (facility, dealType) => {
-  let coverStartDate;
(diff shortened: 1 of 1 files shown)


azure-functions/acbs-function/mappings/facility/helpers/get-exposure-period.js
```diff
@@ -3,19 +3,31 @@ const moment = require('moment');
...
+ * @returns {Integer} Exposure period in months
+ */
 const getExposurePeriod = (facility, dealType) => {
-  let coverStartDate;
   const { facilitySnapshot } = facility;
 
+  // GEF
   if (dealType === CONSTANTS.PRODUCT.TYPE.GEF) {
-    return facility.tfm.exposurePeriodInMonths || 0;
+    return facilitySnapshot.hasBeenIssued
```
IgnatG, line 19: can we define the values separately for this? it will be easier to read
i.e. 
```
const exposurePeriodInMonths = facility.tfm?.exposurePeriodInMonths || 0;
const monthsOfCover = facilitySnapshot?.monthsOfCover || 0;
const { hasBeenIssued } = facilitySnapshot;
return hasBeenIssued ? exposurePeriodInMonths : monthsOfCover;
```

abhi-markan: Have updated.

PR #1012, opened 2022-06-13
fix(DTFS2-5816): Probability of default validation

Zainzzkk:
# Issue: 
* Probability of default did not accept decimals
* Allowed to add probability of default below 0.01 and above 14.09

# Changes made: 
* changed probability of default to Float from Int
* Added validation using JOI to check for decimal places and limits
* updated tests


IgnatG (approved):
LGTM

PR #1016, opened 2022-06-14
fix(DTFS2-5859): Amendments bug - change bank decision

Zainzzkk:
# Issue: 
* effective date still displayed after changing bank decision from proceed to withdraw

# Changes made:
* added check if effectiveDate exists, to change to null
* added unit test and e2e test for this scenario


IgnatG (approved):
LGTM

PR #1019, opened 2022-06-15
fix: updated the eStore validation for deals that already exist

IgnatG:
### Changelog:
- update the validation for existing deals - this to ensure that the same API call is not triggered multiple times


Zainzzkk (dismissed):
Description please.  Else LGTM

Zainzzkk (approved):
LGTM

PR #1031, opened 2022-06-23
build: updated Node.JS to `node:18.4-alpine3.16`

IgnatG:
Dockerfile base image upgraded to `node:18.4-alpine3.16`


abhi-markan (approved):
💯 Good

PR #1043, opened 2022-06-27
fix(DTFS2-5883-5884): Bug - previous declined/withdrawn amendment values used

Zainzzkk:
# Issue:
* When amendment withdrawn or declined, previous values from request were being used

# Changes made:
* checks if latest amendment has currentCoverEndDate/ currentValue and sets it.  If not, then uses from snapshot.
* checks if previous amendment declined or not.  If not declined, then uses value from latest amendment
* added unit tests for helper
* added e2e test


abhi-markan (dismissed):
Looks good.

trade-finance-manager-ui/server/controllers/helpers/amendments.helper.js
```diff
@@ -86,11 +86,53 @@ const hasAmendmentInProgressDealStage = async (dealId) => {
...
+  if (value && ukefDecisionApproved && bankProceed) {
+    return true;
+  }
+
+  return false;
+};
+
+const latestAmendmentCoverEndDateAccepted = (amendment) => {
+  const { ukefDecision, bankDecision, coverEndDate, requireUkefApproval } = amendment;
+  const { APPROVED_WITH_CONDITIONS: withConditions } = DECISIONS.UNDERWRITER_MANAGER_DECISIONS;
```
IgnatG, line 111: we don't need the `withConditions` value here. we can use directly  `APPROVED_WITH_CONDITIONS`

Zainzzkk: done

IgnatG (approved):

PR #1046, opened 2022-06-28
DTFS2-5582 : disbursementAmount undefined - FLR

abhi-markan:
## Facility Loan Record
`disbursementAmount` is not present as expected and due to datatype being `undefined` replace function is being executed which only exists for string datatype.

## Resolution
* Property null check
* Type of `string` comparison


Zainzzkk (approved):
LGTM

PR #1052, opened 2022-06-29
fix(DTFS2-5878/5879): BSS amendments email fix

Zainzzkk:
# Issue:
* BSS amendment emails were not sending

# Changes made:
* pointed to UKEFDealId in details collection if deal type is BSS
* added api-tests

### DIFF
--- a/trade-finance-manager-api/src/v1/emails/amendments/automatic-approval-email-variables.js
+++ b/trade-finance-manager-api/src/v1/emails/amendments/automatic-approval-email-variables.js
@@ -1,15 +1,23 @@
+const { DEAL_TYPE } = require('../../../constants/deals');
+
 const automaticAmendmentEmailVariables = (amendmentVariables) => {
   const { amendment, dealSnapshot, user } = amendmentVariables;
   const { ukefFacilityId } = amendment;
-  const { exporter, ukefDealId, bankInternalRefName } = dealSnapshot;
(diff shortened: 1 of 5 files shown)


trade-finance-manager-api/src/v1/emails/amendments/automatic-approval-email-variables.js
```diff
@@ -1,15 +1,23 @@
+const { DEAL_TYPE } = require('../../../constants/deals');
+
 const automaticAmendmentEmailVariables = (amendmentVariables) => {
   const { amendment, dealSnapshot, user } = amendmentVariables;
   const { ukefFacilityId } = amendment;
-  const { exporter, ukefDealId, bankInternalRefName } = dealSnapshot;
+  const { exporter, ukefDealId, bankInternalRefName, dealType, details } = dealSnapshot;
   const { firstname, surname } = user;
 
+  let ukefDealIdFromDeal = ukefDealId;
```
IgnatG, line 9: can we use something like this:
`const ukefDealId = dealSnapshot.ukefDealId ?? dealSnapshot.details.ukefDealId` - same thing applies to the other files
This means, if the value exists and is not `null` or `undefined`, then get the ukefDealId. otherwise, search for a value inside the details object

Zainzzkk: done

IgnatG (approved):

PR #1054, opened 2022-06-30
fix: updated the eStore timers to execute after 59s

IgnatG:
Changelog:
- updated the timers for CronJobs to execute 59s after the previous function finished.
(this because we need a deploy between different API calls)

### DIFF
--- a/reference-data-proxy/src/cronJobs/eStoreCronJobManager.ts
+++ b/reference-data-proxy/src/cronJobs/eStoreCronJobManager.ts
@@ -12,7 +12,7 @@ export const eStoreCronJobManager = new CronJobManager(
     start: true,
     onComplete: async () => {
       const collection = await getCollection('cron-job-logs');
-      await collection.insertOne({ status: 'eStore Cron Job Manager started successfully', timestamp: new Date() });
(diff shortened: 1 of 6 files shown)


reference-data-proxy/src/cronJobs/eStoreCronJobManager.ts
```diff
@@ -12,7 +13,7 @@ export const eStoreCronJobManager = new CronJobManager(
     onComplete: async () => {
       const collection = await getCollection('cron-job-logs');
       await collection.insertOne({ status: 'eStore Cron Job Manager started successfully', timestamp: new Date() });
-      console.info('eStore Cron Job Manager started successfully');
+      console.info('eStore Cron Job Manager started successfully at', new Date());
```
abhi-markan, line 16: Can we use `cronJobTimer` instead of `new Date()` to avoid instantiating another variable in the memory?

abhi-markan (changes requested):

reference-data-proxy/src/cronJobs/eStoreCronJobManager.ts
IgnatG: good shout - I've updated that now

abhi-markan (approved):
👍🏼

PR #1058, opened 2022-07-01
feat(DTFS2-4614): ACBS facility amendment

abhi-markan:
## Introduction
When a facility amendment is issued by the UKEF it can be either an **Automatic** or **Manual** amendment, once all the pre-requisites have been satisfied for either of the amendment scenarios it will trigger an ACBS facility record amendment via following API path:

```
TFM API -> RDP API -> ACBS DF -> Mulesoft API -> FIS API -> ACBS
```

As of now following two facility records are amended upon either of the attribute has been updated:

* Facility amount
* Facility cover end date

Following mandatory ACBS facility records are amended:

* Facility Master Record (FMR)
* Facility Loan Record (FLR)

## TFM
Upon a successful completion of the ACBS request via status query URI, ACBS response including payload sent is then fed back into `tfm-facility` `acbs` property.


trade-finance-manager-api/src/v1/controllers/acbs.controller.js
```diff
@@ -147,10 +147,19 @@ const issueAcbsFacilities = async (deal) => {
   return acbsIssueFacilityTasks;
 };
 
+const amendAcbsFacility = async (ukefFacilityId, amendments, deal) => {
+  api.amendACBSfacility(ukefFacilityId, amendments, deal)
+    .then((r) => console.log('RESPONSE===', r))
```
abhi-markan, line 152: Please ignore this `console.log` for time being, while I implement FLR (Facility Loan Record).

azure-functions/acbs-function/mappings/facility/facility-master-amend.js
```diff
@@ -0,0 +1,91 @@
...
+  try {
+    // Construct base record
+    let record = {
+      ...fmr,
+      productTypeName: deal.dealSnapshot.dealType,
+      obligorName: deal.exporter.companyName.substring(0, 35),
+    };
+
+    // Construct amendment record
+    if (amendments.amendment) {
```
Zainzzkk, line 63: deconstruct amendments.amendment at top?

azure-functions/acbs-function/mappings/facility/helpers/get-maximum-liability.js
```diff
@@ -5,14 +5,22 @@
  * @returns {Float} Facility UKEF exposure amount
  */
 const getMaximumLiability = (facility, overallAmount = false) => {
-  const value = overallAmount
+  let value = 0;
+
+  if (facility.facilitySnapshot) {
+  value = overallAmount
     ? facility.facilitySnapshot.value
```
Zainzzkk, line 12: deconstruct so bit neater?

trade-finance-manager-api/src/v1/helpers/amendment.helpers.js
```diff
@@ -205,6 +209,39 @@ const sendManualBankDecisionEmail = async (amendmentVariables) => {
...
+
+  // Manual amendment
+  if (manual) {
+    // Bank Decision
+    const { submitted, decision } = amendment.bankDecision;
+    // Bank has accepted the UW decision
+    const proceed = decision === AMENDMENT_BANK_DECISION.PROCEED;
+    // Ensure not all of the amendment requests are declined
+    const { value, coverEndDate } = amendment.ukefDecision;
+    const declined = value === AMENDMENT_UW_DECISION.DECLINED && coverEndDate === AMENDMENT_UW_DECISION.DECLINED;
```
Zainzzkk, line 236: will this work if the amendment only has one change? and will it not crash trying to match undefined to declined?

Zainzzkk (commented):
trade-finance-manager-api/src/constants/amendments.js is empty.  Should it be deleted?

azure-functions/acbs-function/acbs-amend-facility/index.js
```diff
@@ -0,0 +1,116 @@
...
+const mappings = require('../mappings');
+
+module.exports = df.orchestrator(function* amendACBSFacility(context) {
+  try {
+    const { amendment } = context.df.getInput();
+
+    // UKEF Facility ID exists in the payload
+    const hasFacilityId = Boolean(amendment.facilityId);
+    // At least one of the amendment exists in the payload
+    const hasAmendment = Boolean(amendment.amount) || Boolean(amendment.coverEndDate);
```
IgnatG, line 32: for facility value, we tend to use the `value` property from amendment. can we keep it consistent please?
if you want to check whether a facility has an amendment, then you can use the `changeFacilityValue` or `changeCoverEndDate` properties. For those we wouldn't need a typecast because the values would always be boolean

azure-functions/acbs-function/acbs-amend-facility/index.js
```diff
@@ -0,0 +1,116 @@
...
+      // 1. DAF : activity-get-facility-master: Retrieve ACBS `Facility Master Record` with eTag
+      const { acbsFacility: fmr, etag } = yield context.df.callActivityWithRetry(
+        'activity-get-facility-master',
+        retryOptions,
+        { facilityId },
+      );
+
+      if (!!fmr && !!etag) {
+        let facilityMasterRecordAmendments;
+        const amendments = {
```
IgnatG, line 49: isn't the `amendment` already an object? or you just need to create a copy of it?

azure-functions/acbs-function/mappings/facility/facility-master-amend.js
abhi-markan: Have updated.

trade-finance-manager-api/src/v1/helpers/amendment.helpers.js
abhi-markan: Yes this works, have tested a manual amendment.
`undefined` comparison should not cause the script to crash it will simply be `false`.

azure-functions/acbs-function/mappings/facility/helpers/get-maximum-liability.js
abhi-markan: Have updated.

abhi-markan:
> trade-finance-manager-api/src/constants/amendments.js is empty. Should it be deleted?
No, as I might require it for FLR. Thanks

trade-finance-manager-api/src/v1/helpers/amendment.helpers.js
IgnatG, line 236: @Zainzzkk  do we have a helper function when both amendments have been declined?

trade-finance-manager-api/src/v1/helpers/amendment.helpers.js
```diff
@@ -205,6 +209,39 @@ const sendManualBankDecisionEmail = async (amendmentVariables) => {
...
+ */
+const canSendToAcbs = (amendment) => {
+  // Ensure at least one of the attribute has been amended
+  const hasBeenAmended = amendment.changeCoverEndDate || amendment.changeFacilityValue;
+  // Amendment status is marked as `Completed`
+  const completed = amendment.status === AMENDMENT_STATUS.COMPLETED;
+  // Amendment has been submitted by PIM team
+  const PIM = Boolean(amendment.submittedByPim);
+  // Manual amendment verification
+  const manual = Boolean(amendment.requireUkefApproval) && Boolean(amendment.bankDecision);
```
IgnatG, line 226: we shouldn't need the typecast because `requireUkefApproval` is always boolean. If you need to know whether an amendment was submitted back to the bank, then I think there is a specific flag that you can use for that - which would be a boolean flag already

trade-finance-manager-api/src/v1/helpers/amendment.helpers.js
```diff
@@ -205,6 +209,39 @@ const sendManualBankDecisionEmail = async (amendmentVariables) => {
...
+ * This function evaluated across all amendment types.
+ * @param {Object} amendment Facility amendments object
+ */
+const canSendToAcbs = (amendment) => {
+  // Ensure at least one of the attribute has been amended
+  const hasBeenAmended = amendment.changeCoverEndDate || amendment.changeFacilityValue;
+  // Amendment status is marked as `Completed`
+  const completed = amendment.status === AMENDMENT_STATUS.COMPLETED;
+  // Amendment has been submitted by PIM team
+  const PIM = Boolean(amendment.submittedByPim);
```
IgnatG, line 224: same as the others, the `submittedByPim` is always a boolean value

trade-finance-manager-api/src/v1/helpers/amendment.helpers.js
```diff
@@ -205,6 +209,39 @@ const sendManualBankDecisionEmail = async (amendmentVariables) => {
...
+  if (manual) {
+    // Bank Decision
+    const { submitted, decision } = amendment.bankDecision;
+    // Bank has accepted the UW decision
+    const proceed = decision === AMENDMENT_BANK_DECISION.PROCEED;
+    // Ensure not all of the amendment requests are declined
+    const { value, coverEndDate } = amendment.ukefDecision;
+    const declined = value === AMENDMENT_UW_DECISION.DECLINED && coverEndDate === AMENDMENT_UW_DECISION.DECLINED;
+
+    return hasBeenAmended && completed && PIM && Boolean(submitted) && proceed && !declined;
```
IgnatG, line 238: same as the others `submitted` is always boolean. we wouldn't need a typecast for it

trade-finance-manager-api/src/v1/api.js
```diff
@@ -759,6 +759,35 @@ const updateACBSfacility = async (facility, deal) => {
...
+ * @param {String} ukefFacilityId UKEF Facility ID
+ * @param {Object} amendments Facility object comprising of amendments
+ * @returns {Object} updated FMR upon success otherwise error
+ */
+const amendACBSfacility = async (ukefFacilityId, amendments, deal) => {
+  if (amendments) {
+    try {
+      const response = await axios({
+        method: 'post',
+        url: `${refDataUrl}/acbs/facility/${ukefFacilityId}/amend`,
```
IgnatG, line 773: the general convention that we used for endpoints was to use the full name of it
i.e. `.../amendment` - if we are changing 1 amendment 
or
`../amendments` if we we are expecting more items

IgnatG: second item:
are we using the `ukefFacilityId` in the endpoint? shouldn't we use the `_id` (MongoDB) item instead?

trade-finance-manager-api/src/v1/helpers/amendment.helpers.js
Zainzzkk: ukefDecisionRejected => trade-finance-manager-ui\server\controllers\helpers\amendments.helper.js

trade-finance-manager-api/src/v1/helpers/amendment.helpers.js
```diff
@@ -239,10 +276,54 @@ const sendFirstTaskEmail = async (taskVariables) => {
...
+
+  return payload;
+};
+
+/**
+ * Converts non-ms epoch to ms epoch.
+ * @param {Object} payload Amendment payload
+ * @returns {Object} Computed payload with EPOCH sm compatible `coverEndDate`.
+ */
+const calculateCoverEndDate = (payload) => {
```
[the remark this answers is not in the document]
abhi-markan: Formatting, I suppose naming is not quite right I'll update the function name.

trade-finance-manager-api/src/v1/helpers/amendment.helpers.js
abhi-markan: Updated.

trade-finance-manager-api/src/v1/api.js
abhi-markan: 1. Good spot, have updated.
2. Nope `ukefFacilityId` to keep it consistent with Azure DF calls and also I need `ukefFacilityId` to extract FMR which can only be fetched (GET) using `ukefFacilityId`.

trade-finance-manager-api/src/v1/controllers/amendment.controller.js
```diff
@@ -172,6 +102,113 @@ const getAllAmendmentsInProgress = async (req, res) => {
...
+  await sendAmendmentEmail(amendmentId, facilityId);
+
+  // Fetch complete amendment object
+  const amendment = await api.getAmendmentById(facilityId, amendmentId);
+  // Fetch deal object from deal-tfm
+  const tfmDeal = await api.findOneDeal(amendment.dealId);
+  // Construct acceptable deal object
+  const deal = {
+    dealSnapshot: {
+      dealType: tfmDeal.dealSnapshot.dealType,
```
Zainzzkk, line 188: not urgent.  Simplify later please

abhi-markan: Will do, thanks.

trade-finance-manager-api/src/v1/helpers/amendment.helpers.js
IgnatG: @abhi-markan  it may be worth looking at this ^ helper function that we already have in place

azure-functions/acbs-function/acbs-amend-facility/index.js
```diff
@@ -0,0 +1,117 @@
...
+module.exports = df.orchestrator(function* amendACBSFacility(context) {
+  try {
+    const { amendment } = context.df.getInput();
+    if (amendment) {
+    // UKEF Facility ID exists in the payload
+      const hasFacilityId = Boolean(amendment.facilityId);
+      // At least one of the amendment exists in the payload
+      const hasAmendment = Boolean(amendment.amount) || Boolean(amendment.coverEndDate);
+      // Deal properties existence check
+      const hasDeal = amendment.deal && Object.prototype.hasOwnProperty.call(amendment.deal, 'dealSnapshot');
```
IgnatG, line 34: I've mentioned this in a previous comment but it got overridden because of another push:
> for facility value, we tend to use the `value` property from an amendment. can we keep it consistent please?
if you want to check whether a facility has an amendment, then you can use the `changeFacilityValue` or `changeCoverEndDate` properties. For those we wouldn't need a typecast because the values would always be boolean

abhi-markan: `amount` has been used in ACBS throughout to keep it consistent with Mulesoft payload field, outside ACBS I am using `value`.

azure-functions/acbs-function/activity-get-loan-id/index.js
```diff
@@ -0,0 +1,68 @@
...
+      // Validate returned data
+      if (data.length > 0) {
+        const loan = data[0];
+
+        if (loan.loanIdentifier) return loan.loanIdentifier;
+      }
+
+      // Throw an error upon data validation failure
+      throw new Error(
+        JSON.stringify({
```
IgnatG, line 53: do we need the `JSON.stringify` property? can we return an object instead?

azure-functions/acbs-function/activity-update-facility-loan/index.js
```diff
@@ -0,0 +1,70 @@
...
+    const {
+      loanId,
+      facilityId,
+      acbsFacilityLoanInput,
+    } = context.bindingData;
+
+    const missingMandatory = findMissingMandatory(acbsFacilityLoanInput, mandatoryFields);
+
+    if (missingMandatory.length) {
+      return Promise.resolve({ missingMandatory });
```
IgnatG, line 38: can we use `await` instead above?

azure-functions/acbs-function/activity-update-facility-master/index.js
```diff
@@ -44,38 +51,48 @@ const mandatoryFields = [
...
+    } = context.bindingData;
 
-  const missingMandatory = findMissingMandatory(acbsFacilityMasterInput, mandatoryFields);
+    const missingMandatory = findMissingMandatory(acbsFacilityMasterInput, mandatoryFields);
 
-  if (missingMandatory.length) {
-    return Promise.resolve({ missingMandatory });
-  }
+    if (missingMandatory.length) {
+      return Promise.resolve({ missingMandatory });
```
IgnatG, line 65: same as above - `await` on line 62?
otherwise we have a mix of Promise.resolve and async/await

azure-functions/acbs-function/api.js
```diff
@@ -24,23 +25,48 @@ const getACBS = async (apiRef) => {
       headers: {
         'Content-Type': 'application/json',
       },
-    }).catch((err) => ({
-      status: !!err.response
-        ? err.response.status
-        : err,
-    }));
-    return response;
+    }).catch((e) => {
```
IgnatG, line 28: should this be a `try-catch` block instead? I think we have a mix of `try-catch` and `axios.catch` throughout the code

azure-functions/acbs-function/mappings/facility/facility-loan-amend.js
```diff
@@ -0,0 +1,46 @@
...
+          ...record,
+          expiryDate: formatDate(coverEndDate),
+        };
+      }
+    }
+
+    return record;
+  } catch (e) {
+    console.error('Unable to map facility loan record: ', { e });
+    return e;
```
IgnatG, line 42: in other files we've used `throw new Error()`. Do we need to do the same here?

azure-functions/acbs-function/mappings/facility/facility-master-amend.js
```diff
@@ -0,0 +1,91 @@
...
+          guaranteeExpiryDate: formatDate(coverEndDate),
+        };
+      }
+    }
+
+    // Return amended FMR
+    return record;
+  } catch (e) {
+    console.error('Unable to map facility master amendment record: ', { e });
+    return e;
```
IgnatG, line 87: same as above

reference-data-proxy/src/v1/controllers/acbs.controller.ts
```diff
@@ -77,6 +77,34 @@ export const findOne = async (req: Request, res: Response) => {
...
+        bank,
+      },
+    }).catch((err: any) => err);
+    return response;
+  }
+  return {};
+};
+
+export const createAcbsRecordPOST = async (req: Request, res: Response) => {
+  if (req) {
```
IgnatG, line 96: do we need this check here? I assume that this function is called through an API call. in which case, we should always have a request

reference-data-proxy/src/v1/controllers/acbs.controller.ts
```diff
@@ -77,6 +77,34 @@ export const findOne = async (req: Request, res: Response) => {
...
+    try {
+      const { deal, bank } = req.body;
+      const { status, data } = await createAcbsRecord(deal, bank);
+      return res.status(status).send(data);
+    } catch (error) {
+      console.error('ACBS post failed ', { error });
```
IgnatG, line 102: do we need to return a value here? or a status?

reference-data-proxy/src/v1/controllers/acbs.controller.ts
```diff
@@ -103,30 +131,63 @@ export const issueAcbsFacilityPOST = async (req: Request, res: Response) => {
...
-export const createAcbsRecordPOST = async (req: Request, res: Response) => {
+/**
+ * ACBS facility amendment entry function.
+ * Constructs acceptable payload by DOF.
+ * @param {Object} req Request
+ * @param {Object} res Response
+ * @return {Object} Response object with HTTP code as `status` and response as `data`.
+ */
+export const amendAcbsFacilityPost = async (req: Request, res: Response) => {
   if (req) {
```
IgnatG, line 164: same as above - do we need this check here?

reference-data-proxy/src/v1/controllers/acbs.controller.ts
```diff
@@ -103,30 +131,63 @@ export const issueAcbsFacilityPOST = async (req: Request, res: Response) => {
...
 
-export const createAcbsRecordPOST = async (req: Request, res: Response) => {
+/**
+ * ACBS facility amendment entry function.
+ * Constructs acceptable payload by DOF.
+ * @param {Object} req Request
+ * @param {Object} res Response
+ * @return {Object} Response object with HTTP code as `status` and response as `data`.
+ */
+export const amendAcbsFacilityPost = async (req: Request, res: Response) => {
```
IgnatG, line 163: do we need this entire function inside a `try-catch`?

reference-data-proxy/src/v1/controllers/acbs.controller.ts
```diff
@@ -103,30 +131,63 @@ export const issueAcbsFacilityPOST = async (req: Request, res: Response) => {
...
+    // Refine payload
+
+    // Change requested
+    const { changeFacilityValue, changeCoverEndDate } = amendments;
+    // UW Decision
+    const { value, coverEndDate } = amendments.ukefDecision;
+    const valueDeclined = value === UNDERWRITER_MANAGER_DECISIONS.DECLINED;
+    const coverEndDateDeclined = coverEndDate === UNDERWRITER_MANAGER_DECISIONS.DECLINED;
+    // Delete frivolous property
+    if (!changeFacilityValue || valueDeclined) delete payload.amount;
```
IgnatG, line 184: can we use something like this instead?
```
if (!changeFacilityValue || value === UNDERWRITER_MANAGER_DECISIONS.DECLINED) {
    delete payload.amount;
}

```
we don't really write the `if` statements in online - we tend to be explicit even if it expands on multiple lines

reference-data-proxy/src/v1/controllers/acbs.controller.ts
```diff
@@ -103,30 +131,63 @@ export const issueAcbsFacilityPOST = async (req: Request, res: Response) => {
...
-  if (deal) {
+/**
+ * Invoked Azure DOF using HTTP `POST` method.
+ * @param {Object} amendment Amendment object comprising facility ID and amends. A amendment at a time is processed.
+ * @returns {Object} DOF Response
+ */
+const amendAcbsFacility = async (amendment: object) => {
+  const hasAcceptablePayload = Object.prototype.hasOwnProperty.call(amendment, 'coverEndDate') || Object.prototype.hasOwnProperty.call(amendment, 'amount');
+
+  if (amendment && hasAcceptablePayload) {
```
IgnatG, line 142: we can simplify this here by saying:
```
if (amendment.coverEndDate || amendment.amount){...}
```

trade-finance-manager-api/src/v1/helpers/amendment.helpers.js
IgnatG: if we just need to check whether an amendment is manual, then we only need to check the `requireUkefApproval` property.

trade-finance-manager-api/src/v1/helpers/amendment.helpers.js
```diff
@@ -205,6 +210,56 @@ const sendManualBankDecisionEmail = async (amendmentVariables) => {
...
+ * This function evaluated across all amendment types.
+ * @param {Object} amendment Facility amendments object
+ */
+const canSendToAcbs = (amendment) => {
+  // Ensure at least one of the attribute has been amended
+  const hasBeenAmended = amendment.changeCoverEndDate || amendment.changeFacilityValue;
+  // Amendment status is marked as `Completed`
+  const completed = amendment.status === AMENDMENT_STATUS.COMPLETED;
+  // Amendment has been submitted by PIM team
+  const pim = Boolean(amendment.submittedByPim);
```
IgnatG, line 245: I don't think this line is needed. `amendment.submittedByPim` is always boolean and we can use that directly in the conditions below

azure-functions/acbs-function/activity-get-loan-id/index.js
abhi-markan: We do to improve readability in Azure storage explorer.

azure-functions/acbs-function/mappings/facility/facility-fee.js
```diff
@@ -55,7 +55,7 @@ const facilityFee = (deal, facility) => {
   // Multiple EWCS/BSS Fee records as per premium schedule - Array
     feeRecord = [];
     if (facility.tfm.premiumSchedule) {
-      facility.tfm.premiumSchedule.forEach((premiumSchedule, index) => {
+      facility.tfm.premiumSchedule.forEach((_premiumSchedule, index) => {
```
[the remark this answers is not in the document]
abhi-markan: This variable is not being used and as per Sonarlint SCA it should start with an `_` to make it intention explicit.

trade-finance-manager-api/src/v1/helpers/amendment.helpers.js
abhi-markan: I have removed Boolean but to keep the code clean I have left the variable assignment.

trade-finance-manager-api/src/v1/helpers/amendment.helpers.js
```diff
@@ -205,6 +210,56 @@ const sendManualBankDecisionEmail = async (amendmentVariables) => {
   }
 };
 
+/**
+ * Ascertain whether the requested amendment
+ * have been declined or not.
+ * @param {Object} amendment Amendment object
+ * @returns {Boolean} Whether both the amendments decision has been declined by the underwriter.
+ */
+const amendmentDeclined = (amendment) => {
```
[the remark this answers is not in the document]
abhi-markan: `amendmentDeclined` simplifies the helper function `ukefDecisionRejected` which could do with code quality improvement.

reference-data-proxy/src/v1/controllers/acbs.controller.ts
abhi-markan: Have re-introduced `if` branch.
Have left the variable in place to keep it consistent and future proof.

reference-data-proxy/src/v1/controllers/acbs.controller.ts
abhi-markan: Good spot! have added.

reference-data-proxy/src/v1/controllers/acbs.controller.ts
abhi-markan: Have rectified.

azure-functions/acbs-function/mappings/facility/facility-master-amend.js
abhi-markan: This is a safety net in case of any network or firewall issue stripping the request.

azure-functions/acbs-function/mappings/facility/facility-loan-amend.js
abhi-markan: Nope, need to throw error for durable activity functions so they can be stored in Azure storage.
In this case (mapping function) if an error is thrown it will be caught through-out the call stack and displayed on console only.

azure-functions/acbs-function/api.js
abhi-markan: This will will suffice, as it will catch any error thrown.

azure-functions/acbs-function/activity-update-facility-loan/index.js
abhi-markan: This has always been `return` I have just formatted thus lines are being highlighted.

trade-finance-manager-api/src/v1/helpers/amendment.helpers.js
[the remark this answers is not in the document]
IgnatG: in that case, can we combine them in 1 function? otherwise we have to maintain both :)

reference-data-proxy/src/v1/controllers/acbs.controller.ts
```diff
@@ -79,78 +80,98 @@ export const findOne = async (req: Request, res: Response) => {
...
 
 /**
  * Invoked Azure DOF using HTTP `POST` method.
  * @param {Object} amendment Amendment object comprising facility ID and amends. A amendment at a time is processed.
  * @returns {Object} DOF Response
  */
-const amendAcbsFacility = async (amendment: object) => {
-  const hasAcceptablePayload = Object.prototype.hasOwnProperty.call(amendment, 'coverEndDate') || Object.prototype.hasOwnProperty.call(amendment, 'amount');
+const amendAcbsFacility = async (amendment: Amendment) => {
+  const hasAcceptablePayload = amendment.coverEndDate || amendment.amount;
```
IgnatG, line 159: if we leave it here, can we rename it into something like: `hasAmendment`?

abhi-markan: Updated.

azure-functions/acbs-function/acbs-amend-facility/index.js
IgnatG: ok cool. do we need the `hasOwnProperty.call` bit?
Or can we use something like `const hasDeal = amendment.deal && amendment.deal.dealSnapshot` ?

azure-functions/acbs-function/activity-get-loan-id/index.js
```diff
@@ -0,0 +1,68 @@
...
+            dataReceived: data,
+          }, null, 4),
+        );
+      }
+
+      // Validate returned data
+      if (data.length > 0) {
+        const loan = data[0];
+
+        if (loan.loanIdentifier) return loan.loanIdentifier;
```
IgnatG, line 48: we don't really use the `if` statement without the brackets. I believe it may be an eslint error as well.

azure-functions/acbs-function/activity-get-loan-id/index.js
```diff
@@ -0,0 +1,68 @@
...
+          JSON.stringify({
+            name: 'ACBS Loan ID fetch error',
+            facilityId,
+            dataReceived: data,
+          }, null, 4),
+        );
+      }
+
+      // Validate returned data
+      if (data.length > 0) {
```
IgnatG, line 45: we can simply use `if(data.length)`

azure-functions/acbs-function/activity-get-loan-id/index.js
```diff
@@ -0,0 +1,68 @@
...
+        const loan = data[0];
```
IgnatG, line 46: you can deconstruct this
`const [loan] = data;

azure-functions/acbs-function/api.js
```diff
@@ -24,23 +25,50 @@ const getACBS = async (apiRef) => {
...
+      console.error('Error calling GET to ACBS');
+      return {
+        status: e.response ? e.response.status : e,
```
IgnatG, line 31: This can be simplified into this
`status: e?.response?.status ?? e `

azure-functions/acbs-function/api.js
```diff
@@ -24,23 +25,50 @@ const getACBS = async (apiRef) => {
...
+        data: { error: e.response ? e.response.data : e },
```
IgnatG, line 32: Same as above

azure-functions/acbs-function/api.js
```diff
@@ -24,23 +25,50 @@ const getACBS = async (apiRef) => {
...
+const postToAcbs = async (apiRef, acbsInput) => {
   if (!!apiRef && !!acbsInput) {
-    const additionalHeader = etag ? {
-      'If-Match': etag,
-    } : null;
+    return axios({
+      method: 'post',
+      url: `${process.env.MULESOFT_API_UKEF_TF_EA_URL}/${apiRef}`,
+      auth: {
+        username: process.env.MULESOFT_API_KEY,
```
IgnatG, line 45: Can we assign the env variables to a const  variable at the top of the file? I see these values being referenced a few times in this file

azure-functions/acbs-function/api.js
```diff
@@ -24,23 +25,50 @@ const getACBS = async (apiRef) => {
...
+      data: [acbsInput],
+    }).catch((e) => {
+      console.error('Error calling POST to ACBS');
+      return {
+        status: e.response ? e.response.status : e,
+        data: { error: e.response ? e.response.data : e },
+      };
+    });
+  }
+  return {};
```
IgnatG, line 60: Shouldn't this return a status and data object?

azure-functions/acbs-function/api.js
```diff
@@ -24,23 +25,50 @@ const getACBS = async (apiRef) => {
...
+        password: process.env.MULESOFT_API_SECRET,
+      },
+      headers: {
+        'Content-Type': 'application/json',
+      },
+      data: [acbsInput],
+    }).catch((e) => {
+      console.error('Error calling POST to ACBS');
+      return {
+        status: e.response ? e.response.status : e,
```
IgnatG, line 55: Same as above

azure-functions/acbs-function/api.js
```diff
@@ -52,69 +80,69 @@ const putToACBS = async (apiRef, acbsInput, etag) => {
...
-          ? err.response.data
-          : err,
-      },
-    }));
-    return response;
+    }).catch((e) => {
+      console.error('Error calling PUT to ACBS');
+      return {
+        status: e.response ? e.response.status : e,
+        data: { error: e.response ? e.response.data : e },
```
IgnatG, line 87: Same as above

azure-functions/acbs-function/api.js
```diff
@@ -52,69 +80,69 @@ const putToACBS = async (apiRef, acbsInput, etag) => {
...
-          : err,
-      },
-    }));
-    return response;
+      data: acbsInput,
+    }).catch((e) => {
+      console.error('Error calling PATCH to ACBS');
+      return {
+        status: e.response ? e.response.status : e,
+        data: { error: e.response ? e.response.data : e },
```
IgnatG, line 118: Same as above

azure-functions/acbs-function/mappings/facility/facility-fee.js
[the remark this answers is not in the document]
IgnatG: Can we have it without the `_` please? We are not using this format anywhere else. 
With sonnarLint I assume it's a warning rather than  error

reference-data-proxy/src/v1/controllers/acbs.controller.ts
```diff
@@ -77,56 +78,141 @@ export const findOne = async (req: Request, res: Response) => {
...
 };
 
 export const issueAcbsFacilityPOST = async (req: Request, res: Response) => {
-  if (req) {
-    const { id } = req.params;
-    const { facility, deal } = req.body;
-    const { status, data } = await issueAcbsFacility(id, facility, deal);
-    return res.status(status).send(data);
+  try {
+    if (req) {
```
IgnatG, line 134: Do we need this if statement?

reference-data-proxy/src/v1/controllers/acbs.controller.ts
```diff
@@ -77,56 +78,141 @@ export const findOne = async (req: Request, res: Response) => {
...
+ * Invoked Azure DOF using HTTP `POST` method.
+ * @param {Object} amendment Amendment object comprising facility ID and amends. A amendment at a time is processed.
+ * @returns {Object} DOF Response
+ */
+const amendAcbsFacility = async (amendment: Amendment) => {
+  const hasAmendment = amendment.coverEndDate || amendment.amount;
 
-export const createAcbsRecordPOST = async (req: Request, res: Response) => {
-  if (req) {
+  if (amendment && hasAmendment) {
```
IgnatG, line 158: Technically you don't need the `amendment` part since that's part of `hasAmendment` check

trade-finance-manager-api/src/v1/controllers/acbs.controller.js
```diff
@@ -138,13 +156,36 @@ const issueAcbsFacilities = async (deal) => {
...
+ */
+const amendAcbsFacility = async (amendments, facility, deal) => {
+  let payload;
+
+  // TO-DO : EPOCH Convergence
+  if (amendments.coverEndDate) {
+    payload = formatCoverEndDate(amendments);
+  }
+
+  api.amendACBSfacility(payload, facility, deal)
```
IgnatG, line 180: Isn't this one supposed to have `await`?

trade-finance-manager-ui/server/controllers/case/amendments/managersConditionsAndComments.controller.js
```diff
@@ -180,7 +180,7 @@ const postManagersConditionsAndCommentsSummary = async (req, res) => {
     console.error('Unable to submit the underwriter managers decision');
     return res.redirect(`/case/${dealId}/facility/${facilityId}/amendment/${amendmentId}/managers-conditions/summary`);
   } catch (err) {
-    console.error("There was a problem submitting the manager's decision %O", { response: err?.response?.data });
+    console.error("There was a problem submitting the manager's decision %O", { response: err?.response?.data }, { err });
```
IgnatG, line 183: Do we need the `{err}` part?

trade-finance-manager-ui/server/controllers/helpers/amendments.helper.js
```diff
@@ -34,28 +34,23 @@ const userCanEditBankDecision = (amendment, user) => {
...
-    // else if only 1, checks either is declined
-    return true;
+  const { changeFacilityValue, changeCoverEndDate } = amendment;
+  const { value, coverEndDate } = amendment.ukefDecision;
+
+  // Ensure not all of the amendment requests are declined
+
+  // Dual amendment request
+  if (changeFacilityValue && changeCoverEndDate) {
+    return value === DECISIONS.UNDERWRITER_MANAGER_DECISIONS.DECLINED && coverEndDate === DECISIONS.UNDERWRITER_MANAGER_DECISIONS.DECLINED;
```
IgnatG, line 50: Can we deconstruct the `DECISIONS` variable?- similar to what we had originally in this function

trade-finance-manager-api/src/v1/controllers/acbs.controller.js
abhi-markan: No as I am resolving the promise with `then`.

trade-finance-manager-ui/server/controllers/case/amendments/managersConditionsAndComments.controller.js
abhi-markan: Nope, apologies that was an accidental push only used it for debugging.

trade-finance-manager-ui/server/controllers/helpers/amendments.helper.js
```diff
@@ -42,15 +42,16 @@ const userCanEditBankDecision = (amendment, user) => {
 const ukefDecisionRejected = (amendment) => {
   const { changeFacilityValue, changeCoverEndDate } = amendment;
   const { value, coverEndDate } = amendment.ukefDecision;
+  const { UNDERWRITER_MANAGER_DECISIONS } = DECISIONS;
```
IgnatG, line 45: can we destructure the the UW_MANAGER_DECISION as well?
so something like this:
```js
const { DECLINED } = DECISIONS.UNDERWRITER_MANAGER_DECISIONS;
```

and then below in the return statement we can have

```js
return value === DECLINED && coverEndDate === DECLINED
```

azure-functions/acbs-function/api.js
```diff
@@ -9,6 +9,9 @@
  * All the function have argument validation check and return object verification in
  * case err object does not have expected properties due to network connection, SSL verification or other issues.
  */
+const endpoint = process.env.MULESOFT_API_UKEF_TF_EA_URL;
```
IgnatG, line 12: can we call it something like `acbsUrl` - or something like this. just to be more descriptive as to what the endpoint is doing

trade-finance-manager-api/src/v1/helpers/amendment.helpers.js
IgnatG: I remember you were working on this change the other day. did you push it?

trade-finance-manager-api/src/v1/helpers/amendment.helpers.js
```diff
@@ -233,6 +238,56 @@ const sendManualBankDecisionEmail = async (amendmentVariables) => {
...
+  const { changeFacilityValue, changeCoverEndDate } = amendment;
+  const { value, coverEndDate } = amendment.ukefDecision;
+
+  // Ensure not all of the amendment requests are declined
+
+  // Dual amendment request
+  if (changeFacilityValue && changeCoverEndDate) {
+    return value === UNDERWRITER_MANAGER_DECISIONS.DECLINED && coverEndDate === UNDERWRITER_MANAGER_DECISIONS.DECLINED;
```
IgnatG, line 255: can we deconstruct this row and 258 the same way we did in the other file?

azure-functions/acbs-function/acbs-amend-facility/index.js
```diff
@@ -0,0 +1,157 @@
...
+          // 2.3 FLR update
+          // 2.3.1 Extract loan id for facility id
+          const loanId = yield context.df.callActivityWithRetry('activity-get-loan-id', retryOptions, {
+            facilityId,
+          });
+
+          if (loanId) {
+            // 2.3.2 - UKEF Exposure
+            // Loan amount - `Bond` facility types only
+            if (amendment.amount && facilitySnapshot.type === CONSTANTS.FACILITY.FACILITY_TYPE.BOND) {
```
Zainzzkk, line 110: is it possible to deconstruct { FACILITY } = require('___/constants); or even {FACILITY_TYPE} = require('___/constants/facility);

azure-functions/acbs-function/acbs-amend-facility/index.js
```diff
@@ -0,0 +1,157 @@
...
+        const { acbsFacility: fmr, etag } = yield context.df.callActivityWithRetry('activity-get-facility-master', retryOptions, { facilityId });
+
+        if (!!fmr && !!etag && !!facility.facilitySnapshot) {
+          let facilityMasterRecordAmendments;
+          let facilityLoanRecordAmendments;
+          const amendments = {
+            amendment,
+          };
+          const { facilitySnapshot } = facility;
+
```
Zainzzkk, line 53: can we remove some of the spaces? file looks super long

abhi-markan: This is because of safety net conditions, indentation is as to be expected.

azure-functions/acbs-function/activity-get-facility-master/index.js
```diff
@@ -12,28 +12,33 @@ const api = require('../api');
...
+        throw new Error(
+          JSON.stringify({
+            name: 'ACBS Facility fetch error',
+            facilityId,
+            dataReceived: data,
+          }, null, 4),
+        );
+      }
+
+      return {
```
Zainzzkk, line 31: is there a check to make sure that data exists?

azure-functions/acbs-function/activity-update-facility-loan-amount/index.js
```diff
@@ -0,0 +1,72 @@
...
+
+const updateFacilityLoanAmount = async (context) => {
+  try {
+    const {
+      loanId,
+      facilityId,
+      acbsFacilityLoanInput,
+    } = context.bindingData;
+
+    const missingMandatory = findMissingMandatory(acbsFacilityLoanInput, mandatoryFields);
```
Zainzzkk, line 37: what about using await findMissingMandatory() so don't need Promise.resolve? ignore if I am wrong

azure-functions/acbs-function/acbs-amend-facility/index.js
abhi-markan: Updated.

azure-functions/acbs-function/activity-update-facility-loan-amount/index.js
```diff
@@ -0,0 +1,72 @@
...
+        }, null, 4),
+      );
+    }
+
+    return {
+      status,
+      submittedToACBS,
+      receivedFromACBS: moment().format(),
+      dataSent: acbsFacilityLoanInput,
+      ...data,
```
Zainzzkk, line 64: same as above, any checks for if data is not null?

azure-functions/acbs-function/activity-get-facility-master/index.js
abhi-markan: If null or empty data is returned then the further checks on the calling function will not be further executed, so no check is necessary here.

azure-functions/acbs-function/activity-update-facility-loan-amount/index.js
abhi-markan: Nope, it is just a normal procedural function.

azure-functions/acbs-function/activity-update-facility-loan-amount/index.js
abhi-markan: Same as above :-)

azure-functions/acbs-function/mappings/facility/facility-loan-amend.js
```diff
@@ -0,0 +1,55 @@
+const CONSTANTS = require('../../constants');
+const { formatDate, now } = require('../../helpers/date');
+const helpers = require('./helpers');
+
+const facilityLoanAmend = (amendments, facility, facilityMasterRecord) => {
+  try {
+    // Default facility loan record
+    let record = { portfolioIdentifier: CONSTANTS.FACILITY.PORTFOLIO.E1 };
```
Zainzzkk, line 8: can we destructure constants please

azure-functions/acbs-function/mappings/facility/helpers/get-exposure-period.js
```diff
@@ -4,26 +4,45 @@ const CONSTANTS = require('../../../constants');
...
+    let exposurePeriod = moment(endDate).diff(moment(startDate), 'months') + 1;
+    // Month offset
+    const offset = moment(startDate).date() === moment(endDate).date() ? -1 : 0;
+    exposurePeriod += offset;
+
+    return String(exposurePeriod);
+  }
+
+  // New facility exposure calculation
   if (dealType === CONSTANTS.PRODUCT.TYPE.GEF) {
```
Zainzzkk, line 38: same as above

azure-functions/acbs-function/mappings/facility/helpers/get-next-due-date.js
```diff
@@ -13,15 +13,14 @@ const getNextDueDate = (facility) => {
   const { guaranteeCommencementDate, guaranteeExpiryDate } = facility.tfm.facilityGuaranteeDates
     ? facility.tfm.facilityGuaranteeDates
     : facility.update;
-
   const months = getFeeFrequencyMonths(facility);
   const feeType = mapFeeType(facility.facilitySnapshot);
 
   if (feeType === CONSTANTS.FACILITY.FEE_TYPE.AT_MATURITY) {
```
Zainzzkk, line 19: same as above please

azure-functions/acbs-function/mappings/facility/facility-loan-amend.js
abhi-markan: Done

azure-functions/acbs-function/mappings/facility/helpers/get-exposure-period.js
abhi-markan: Done

azure-functions/acbs-function/mappings/facility/helpers/get-next-due-date.js
abhi-markan: Done

trade-finance-manager-api/src/v1/api.js
```diff
@@ -783,6 +783,37 @@ const updateACBSfacility = async (facility, deal) => {
...
+      return response.data;
+    } catch (err) {
+      console.error('TFM-API Facility amend error', { err });
+      return err;
+    }
+  }
+  return {};
+};
+
 const getFunctionsAPI = async (type = CONSTANTS.DURABLE_FUNCTIONS.TYPE.ACBS, url = '') => {
```
Zainzzkk, line 817: same as above

azure-functions/acbs-function/api.js
```diff
@@ -1,120 +1,163 @@
 /* eslint-disable no-extra-boolean-cast */
```
Zainzzkk, line 1: do you need this?

abhi-markan: Good spot, have removed.

trade-finance-manager-api/src/v1/api.js
abhi-markan: Done

trade-finance-manager-api/src/v1/helpers/amendment.helpers.js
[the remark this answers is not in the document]
IgnatG: I can see in this file a function called `isDeclined`. And this one is called `amendmentDeclined`.
What is the difference between these 2?

.github/workflows/pr_e2e_tests.yml
```diff
@@ -69,7 +69,11 @@ jobs:
       name: ${{ needs.set-environment.outputs.environment }}
     runs-on: ubuntu-latest
     strategy:
-      fail-fast: true
+      # when one test fails, DO NOT cancel the other
```
IgnatG, line 72: We need to revert these changes back

IgnatG (dismissed):

IgnatG (approved):

PR #1063, opened 2022-07-04
DTFS2-bug: `underwriterManagersDecision` property existence check

abhi-markan:
* `underwriterManagersDecision` property existence check in `tfmDeal` object.
* Return false upon failure.


trade-finance-manager-api/src/v1/controllers/should-update-deal-from-MIA-to-MIN.js
```diff
@@ -3,12 +3,14 @@ const CONSTANTS = require('../../constants');
 const shouldUpdateDealFromMIAtoMIN = (deal, tfmDeal) => {
   const isMIA = deal.submissionType === CONSTANTS.DEALS.SUBMISSION_TYPE.MIA;
 
-  const hasApprovedDecision = (tfmDeal.underwriterManagersDecision
+  if (Object.prototype.hasOwnProperty.call(tfmDeal, 'underwriterManagersDecision')) {
```
IgnatG, line 6: can we run something like
`if (tfmDeal.underwriterManagersDecision) { ... }`

abhi-markan: Updated.

PR #1087, opened 2022-07-14
fix(DTFS2-5930): Facility fixed fee income class code

abhi-markan:
## Introduction
Facility fixed fee record / premium schedule has `incomeClassCode` field which was being set as per the deal type due to lack of clarity in the user story. Due to deal type for BSS and EWCS and set in unison i.e. `BSS/EWCS` rather individually this has caused the helper function to reset to the default case which was set to BSS, thus supplying `BPM` for EWCS facility.

This bug has caused issue with three production EWCS deals which have been set as `BPM` instead of `EPM` hence causing issues with reconciliation with CODA.

## Resolution
* Facility type evaluation for setting the correct income class code.
* Improved function documentation.


Zainzzkk (approved):
LGTM

PR #1099, opened 2022-07-27
feat(DTFS-5944): Add submissionType = Manual/Automatic to amendments

IgnatG:
#### Changelog:
- added `submissionType` to amendments payload to differentiate in plain text between Manual/Automatic amendments


trade-finance-manager-ui/server/controllers/case/amendments/amendmentRequestApproval.controller.js
```diff
@@ -31,6 +31,7 @@ const postAmendmentRequestApproval = async (req, res) => {
   const { facilityId, amendmentId } = req.params;
   const { requireUkefApproval } = req.body;
   const approval = requireUkefApproval === 'Yes';
```
abhi-markan, line 33: I think this value is `Boolean`?

IgnatG: this value comes from the form. It's string at that time

abhi-markan (approved):

PR #1102, opened 2022-07-31
fix(DTFS2-5943): sanitise the `reasons`, `conditions` and `comments` input for amendments

IgnatG:
#### Changelog:
- Sanitised the "reasons", "conditions" and "comments" input


abhi-markan (changes requested):
Can you please add a test for `<embed>` tag.

abhi-markan (approved):

PR #1112, opened 2022-08-03
feat(DTFS2-5965) : TFM API  field support

abhi-markan:
## Introduction
When querying `v1/tfm/deals` with below sample body, TFM API returns empty data set.
This was due to `lastUpdated` field was missed out from array of acceptable fields.

```{
    "queryParams": {
        "byField": [
            {
                "name": "tfm.lastUpdated",
                "value": "02-08-2022"
            }
        ]
    }
}
```

## Resolution
* `tfm.lastUpdated` has been added to `TIMESTAMP_FIELDS`.
* API Unit test case.
* Lint fix.


ttbarnes (dismissed):
Brilliant great stuff, thank you - i'd say this is a new feature rather than a bug, wasn't a requirement until now 😉

abhi-markan:
> Brilliant great stuff, thank you - i'd say this is a new feature rather than a bug, wasn't a requirement until now 😉

Brilliant thanks, have updated commit message and label.

IgnatG (approved):

PR #1113, opened 2022-08-04
fix(DTFS2-5962): amendment 2 is used twice and the latest one is not at the top

Zainzzkk:
# Issue:
* Amendment version did not increase when withdrawn meaning 2 were on the same version

# Changes made:
* Made new call to return version number of latest completed amendment (rather than latest proceed amendment).
* Added e2e test


dtfs-central-api/src/v1/controllers/tfm/amendments/tfm-get-amendments.controller.js
```diff
@@ -324,6 +324,31 @@ const findLatestCompletedAmendmentByFacilityId = async (facilityId) => {
...
+      const collection = await db.getCollection('tfm-facilities');
+      const amendment = await collection.aggregate([
+        { $match: { _id: ObjectId(facilityId) } },
+        { $unwind: '$amendments' },
+        { $match: { 'amendments.status': COMPLETED } },
+        { $sort: { 'amendments.updatedAt': -1, 'amendments.version': -1 } },
+        { $project: { _id: 0, amendments: 1 } },
+        { $limit: 1 }
+      ]).toArray();
+      return amendment[0]?.amendments?.version ?? null;
```
abhi-markan, line 342: Can we check whether `amendment` has an data inside an array using `amendment.length > 0` please?

dtfs-central-api/src/v1/controllers/tfm/amendments/tfm-post-amendments.controller.js
```diff
@@ -28,8 +29,8 @@ exports.postTfmAmendment = async (req, res) => {
           status: CONSTANTS.AMENDMENT.AMENDMENT_STATUS.NOT_STARTED,
           version: 1
         };
-        if (latestCompletedAmendment) {
-          amendment.version = latestCompletedAmendment.version + 1;
+        if (latestCompletedAmendmentVersion) {
+          amendment.version = latestCompletedAmendmentVersion + 1;
```
abhi-markan, line 33: Do we need to typecast it to `Number()`?

abhi-markan (changes requested):

dtfs-central-api/src/v1/controllers/tfm/amendments/tfm-post-amendments.controller.js
IgnatG: That won't be necessary as the version is always a number

dtfs-central-api/src/v1/controllers/tfm/amendments/tfm-get-amendments.controller.js
IgnatG: We shouldn't need that though. If `amendment[0]` fails, then it will return the value after `??` - in this case `null`

abhi-markan (approved):

PR #1116, opened 2022-08-08
fix(DTFS2-5966): Amendment facility time value fix

Zainzzkk:
# Issue: 
* TFM facility page showed maximum facility  exposure timestamp as January 1970 as value was mapped using date-fns in epoch seconds rather than milliseconds

# Changes made: 
* converted from date-fns timestamp to native epoch timestamp with ms
* Linked to amendment submission date
* added to existing e2e tests

### DIFF
--- a/trade-finance-manager-api/src/graphql/reducers/mappings/facilities/mapUkefExposure.js
+++ b/trade-finance-manager-api/src/graphql/reducers/mappings/facilities/mapUkefExposure.js
@@ -1,4 +1,3 @@
-const { fromUnixTime } = require('date-fns');
 const { formattedNumber } = require('../../../../utils/number');
 const api = require('../../../../v1/api');
 const { calculateNewFacilityValue, calculateUkefExposure, isValidCompletedValueAmendment } = require('../../../helpers/amendment.helpers');
@@ -26,7 +25,7 @@ const mapUkefExposure = async (facilityTfm, facility) => {
         const { coverPercentage, coveredPercentage } = facility.facilitySnapshot;
         const { requireUkefApproval, submittedAt, bankDecision } = latestCompletedAmendment;
         // value of ukefExposureCalculationTime from automatic amendment submission time or manual amendment bankDecision submission time
-        const ukefExposureCalculationTime = requireUkefApproval ? bankDecision.submittedAt : submittedAt;
+        const ukefExposureTimestamp = requireUkefApproval ? bankDecision.submittedAt : submittedAt;
 
         // BSS is coveredPercentage while GEF is coverPercentage
         const coverPercentageValue = coverPercentage || coveredPercentage;
@@ -37,7 +36,7 @@ const mapUkefExposure = async (facilityTfm, facility) => {
         // sets new exposure value based on amendment value
         formattedUkefExposure = formattedNumber(ukefExposureValue);
         // converts from seconds unix timestamp to one with milliseconds
-        ukefExposureCalculationTimestampValue = fromUnixTime(ukefExposureCalculationTime).valueOf();
+        ukefExposureCalculationTimestampValue = new Date(ukefExposureTimestamp * 1000).valueOf();
       }
     }
 
--- a/trade-finance-manager-api/src/graphql/reducers/mappings/facilities/mapUkefExposure.js
+++ b/trade-finance-manager-api/src/graphql/reducers/mappings/facilities/mapUkefExposure.js
@@ -1,4 +1,4 @@
-const { getUnixTime } = require('date-fns');
+const { fromUnixTime } = require('date-fns');
 const { formattedNumber } = require('../../../../utils/number');
 const api = require('../../../../v1/api');
 const { calculateNewFacilityValue, calculateUkefExposure, isValidCompletedValueAmendment } = require('../../../helpers/amendment.helpers');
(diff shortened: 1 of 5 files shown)


trade-finance-manager-api/src/graphql/reducers/mappings/facilities/mapUkefExposure.js
```diff
@@ -33,7 +36,8 @@ const mapUkefExposure = async (facilityTfm, facility) => {
 
         // sets new exposure value based on amendment value
         formattedUkefExposure = formattedNumber(ukefExposureValue);
-        ukefExposureCalculationTimestampValue = getUnixTime(new Date());
+        // converts from seconds unix timestamp to one with milliseconds
+        ukefExposureCalculationTimestampValue = fromUnixTime(ukefExposureCalculationTime).valueOf();
```
IgnatG, line 40: @abhi-markan  I need your input here: do we want to use `*1000` or `valueOf()`? alternatively, the `submitted` at value is generated using date-fns. Do we want to use `Date.now()` instead?

Zainzzkk: just to add, for the second option for changing the generated value may mean changing quite a few files to display ms properly

abhi-markan:
> # Issue:
> * TFM facility page showed maximum facility time as January 1970 as value was mapped using date-fns in epoch seconds rather than milliseconds
> 
> # Changes made:
> * converted from date-fns timestamp to native epoch timestamp with ms
> * Linked to amendment submission date
> * added to existing e2e tests

Sorry what do you mean by `maximum facility time`?

trade-finance-manager-api/src/graphql/reducers/mappings/facilities/mapUkefExposure.js
IgnatG: noted. 
@abhi-markan if you want this standardised, then it needs to be implemented before we go live with amendments

trade-finance-manager-api/src/graphql/reducers/mappings/facilities/mapUkefExposure.js
```diff
@@ -24,6 +24,9 @@ const mapUkefExposure = async (facilityTfm, facility) => {
 
       if (isValidCompletedValueAmendment(latestCompletedAmendment)) {
         const { coverPercentage, coveredPercentage } = facility.facilitySnapshot;
+        const { requireUkefApproval, submittedAt, bankDecision } = latestCompletedAmendment;
+        // value of ukefExposureCalculationTime from automatic amendment submission time or manual amendment bankDecision submission time
+        const ukefExposureCalculationTime = requireUkefApproval ? bankDecision.submittedAt : submittedAt;
```
abhi-markan, line 29: Are saving precise time of `UKEF Exposure` calculation time?

trade-finance-manager-api/src/graphql/reducers/mappings/facilities/mapUkefExposure.js
abhi-markan: 1. What format of value does the variable `ukefExposureCalculationTimestampValue` will have for both the conditions.
2. We need to ensure both the values of above variable in ternary operator both are consistent i.e. EPOCH without ms.

abhi-markan (changes requested):

trade-finance-manager-api/src/graphql/reducers/mappings/facilities/mapUkefExposure.js
Zainzzkk: no, because it is done every time we map once amendment completed.  if no amendment then it's saved as deal submission date in mongodb

trade-finance-manager-api/src/graphql/reducers/mappings/facilities/mapUkefExposure.js
Zainzzkk: mongodb value is in epoch milliseconds (none amendment based).  Amendment is in epoch seconds for both sides of the ternary operator

trade-finance-manager-api/src/graphql/reducers/mappings/facilities/mapUkefExposure.js
abhi-markan: Should we rename the variable to `submissionDate` or something similar (if I have understood the reasoning behind the variable).

trade-finance-manager-api/src/graphql/reducers/mappings/facilities/mapUkefExposure.js
abhi-markan: So just to confirm values being saved in DB are both saved from `Date` function?

trade-finance-manager-api/src/graphql/reducers/mappings/facilities/mapUkefExposure.js
Zainzzkk: So, in tfm facilities-> tfm.  Without amendments the value is stored as ukefExposureCalculationTimestamp which is generated on deal submission and is epoch milliseconds.  When an amendment is submitted, there is submittedAt value which is generated using getUnixTime() from date-fns.  May need @IgnatG input if we are changing the name

Zainzzkk: So, in tfm facilities-> tfm. Without amendments the value is stored as ukefExposureCalculationTimestamp which is generated on deal submission and is epoch milliseconds. When an amendment is submitted, there is submittedAt value which is generated using getUnixTime() from date-fns. May need @IgnatG input if we are changing the name

trade-finance-manager-api/src/graphql/reducers/mappings/facilities/mapUkefExposure.js
Zainzzkk: done - and to answer your question @abhi-markan ukefExposureCalculationTimestamp is saved in tfm-facilities from Date function whereas submittedAs is from getUnixTime from date-fns

trade-finance-manager-api/src/graphql/reducers/mappings/facilities/mapUkefExposure.js
Zainzzkk: done

abhi-markan (approved):

PR #1118, opened 2022-08-08
feat(DTFS2-5945): Amendment change leadUnderwriterId to leadUnderwriter object

Zainzzkk:
# Changes made:
* updated leadUnderwriterId so it is stored as object containing  _id and first and last name
* updated tests

### DIFF
--- a/trade-finance-manager-ui/server/controllers/case/amendments/leadUnderwriter.controller.js
+++ b/trade-finance-manager-ui/server/controllers/case/amendments/leadUnderwriter.controller.js
@@ -16,8 +16,8 @@ const getAmendmentLeadUnderwriter = async (amendment, user) => {
   let leadUnderwriterId;
 
   // checks if leadUnderwriter exists in amendments object and sets currentLeadUnderWriterUserId
-  if (amendment?.leadUnderwriterId) {
-    leadUnderwriterId = amendment.leadUnderwriterId;
+  if (amendment?.leadUnderwriter?._id) {
+    leadUnderwriterId = amendment.leadUnderwriter._id;
   }
 
   // checks if set and not unassigned and gets details
@@ -50,6 +50,8 @@ const getAssignAmendmentLeadUnderwriter = async (req, res) => {
     return res.redirect('/not-found');
   }
 
+  const { leadUnderwriter } = amendment;
+
   const { user } = req.session;
 
   // checks if can edit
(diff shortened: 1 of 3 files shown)


trade-finance-manager-ui/server/controllers/case/amendments/leadUnderwriter.controller.js
```diff
@@ -88,16 +90,30 @@ const getAssignAmendmentLeadUnderwriter = async (req, res) => {
...
+    const { assignedTo: assignedToValue } = req.body;
+    // assignedToValue is only the id so have to get the user's name from api call
+    const user = await api.getUser(assignedToValue);
 
-  const { assignedTo: assignedToValue } = req.body;
-  const update = { leadUnderwriterId: assignedToValue };
+    // if no user.firstName or lastName, then values set to null
+    const updateUser = {
+      _id: assignedToValue,
+      firstName: user.firstName ? user.firstName : null,
```
IgnatG, line 107: can we use something like:
`firstName: user.firstName ?? null`
?

trade-finance-manager-ui/server/controllers/case/amendments/leadUnderwriter.controller.js
```diff
@@ -88,16 +90,30 @@ const getAssignAmendmentLeadUnderwriter = async (req, res) => {
...
+      lastName: user.lastName ? user.lastName : null,
```
abhi-markan, line 108: I would suggest usage of nullish coalescing for the lastname as well please.

trade-finance-manager-ui/server/controllers/case/amendments/leadUnderwriter.controller.js
```diff
@@ -88,16 +90,30 @@ const getAssignAmendmentLeadUnderwriter = async (req, res) => {
...
+    }
+
+    const { assignedTo: assignedToValue } = req.body;
+    // assignedToValue is only the id so have to get the user's name from api call
+    const user = await api.getUser(assignedToValue);
 
-  const { assignedTo: assignedToValue } = req.body;
-  const update = { leadUnderwriterId: assignedToValue };
+    // if no user.firstName or lastName, then values set to null
+    const updateUser = {
```
abhi-markan, line 105: Shall we rename this variable to `leadUnderwriter` (if no conflicts)?

abhi-markan (changes requested):

trade-finance-manager-ui/server/controllers/case/amendments/leadUnderwriter.controller.js
Zainzzkk: done

trade-finance-manager-ui/server/controllers/case/amendments/leadUnderwriter.controller.js
Zainzzkk: done

trade-finance-manager-ui/server/controllers/case/amendments/leadUnderwriter.controller.js
Zainzzkk: done

abhi-markan (approved):
Nice work

PR #1126, opened 2022-08-16
fix: Remove ambiguous function

abhi-markan:
## Introduction
Function with similar logic was flagged.

## Resolution

* Removal of function with least amount of logic.
* `isDeclined ` has been replaced with `amendmentDeclined`


Zainzzkk (approved):
lgtm

PR #1130, opened 2022-08-17
fix(): Appropriate returns

abhi-markan:
## Introduction
When an ACBS container is not running locally, no response code is being send from `acbs.controller.ts` to the call stack further up the chain. Lack of appropriate return encourages a snowball effect of bugs.

## Resolution
* HTTP status code `400` send upon no response from ACBS orchestrated container.
* `addtoACBSLog` only called upon a valid response.
* `null` return where necessary.

### DIFF
--- a/trade-finance-manager-api/src/v1/controllers/deal.controller.js
+++ b/trade-finance-manager-api/src/v1/controllers/deal.controller.js
@@ -92,6 +92,8 @@ const submitACBSIfAllPartiesHaveUrn = async (dealId) => {
   if (allRequiredPartiesHaveUrn) {
     await acbsController.createACBS(deal);
   }
+
+  return null;
 };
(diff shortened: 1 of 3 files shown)


IgnatG (dismissed):

Zainzzkk (dismissed):
LGTM

trade-finance-manager-api/src/v1/controllers/deal.controller.js
```diff
@@ -92,6 +92,8 @@ const submitACBSIfAllPartiesHaveUrn = async (dealId) => {
   if (allRequiredPartiesHaveUrn) {
     await acbsController.createACBS(deal);
```
Zainzzkk, line 93: should this not be return here?

PR #1131, opened 2022-08-17
feat(DTFS2-5990): Amendment completion remapping

Zainzzkk:
# Issue:
* Amendment mapping was calling external and internal APIs a lot and was using resources

# Changes made:
* Changed mapping to take latest amendment tfm object with populated values
* Calculated relevant fields using on amendment completion and added to TFM object in latest amendment
* Added api-tests and modified failing tests


trade-finance-manager-api/src/graphql/helpers/amendment.helpers.js
```diff
@@ -64,80 +62,37 @@ const calculateUkefExposure = (facilityValueInGBP, coverPercentage) => {
   return null;
 };
 
-// checks for coverStartDate and returns based on deal-type as stored differently
-const dealTypeCoverStartDate = (facilitySnapshot) => {
-  const { coverStartDate } = facilitySnapshot;
+const findLatestCompletedAmendment = (amendments) => {
+  const amendmentsReversed = [...amendments].reverse();
```
IgnatG, line 66: why do we need the amendments in reverse order?

Zainzzkk: as we need the last item in the array with a tfm object for the most latest amendment.  findLast() doesn't work yet so this was best way to use find

PR #1135, opened 2022-08-22
fix(DTFS2-ACBS-Bugs): Backward compatible script

abhi-markan:
## Introduction
Azure Functions App do not support optional chaining `?` and nullish coalescing`??`.

## Resolution
* These have been removed and replaced with `ternary operator` to mitigate compatibility issue.


Zainzzkk (approved):
LGTM

PR #1137, opened 2022-08-24
fix(DTFS2-6013): Two decimal point format - Facility Records

abhi-markan:
## Introduction
Amid recent ACBS facility master records creation failure, a two decimal point formatting is required.
Despite this has never been an issue before (Deals submitted prior to 24/08/2022) but nevertheless a back-end level decimal point formatting is a necessity. 

## Resolution
* Introduction of `decimalPoint` function responsible for trailing two decimal point formatting.
* Default value set to `0`.


IgnatG (approved):
Lgtm

PR #1138, opened 2022-08-26
fix(ABCS-Bug): Payload initialisation and void check  

abhi-markan:
## Introduction
ACBS was not being executed upon a facility value amendment only.
This is due to `null` payload being send to calling function.

## Resolution
* Default payload initialisation
* Void check upon receiving null from the call stack.

### DIFF
--- a/trade-finance-manager-api/src/v1/controllers/acbs.controller.js
+++ b/trade-finance-manager-api/src/v1/controllers/acbs.controller.js
@@ -175,7 +175,7 @@ const issueAcbsFacilities = async (deal) => {
  * @param {Object} deal Bespoke deal object
  */
 const amendAcbsFacility = async (amendments, facility, deal) => {
-  let payload;
+  let payload = amendments;
 
   // TO-DO : EPOCH Convergence
   if (amendments.coverEndDate) {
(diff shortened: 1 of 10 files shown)


trade-finance-manager-api/src/v1/controllers/acbs.controller.js
```diff
@@ -184,7 +184,7 @@ const amendAcbsFacility = async (amendments, facility, deal) => {
 
   api.amendACBSfacility(payload, facility, deal)
     .then((acbsTaskLinks) => {
-      if (acbsTaskLinks.id) {
+      if (acbsTaskLinks && acbsTaskLinks.id) {
```
IgnatG, line 187: is the second validation needed? i.e. `acbsTaskLinks` ? 
I thought `acbsTaskLinks.id` would check both `acbsTaskLinks` and `acbsTaskLinks.id`

IgnatG (dismissed):

trade-finance-manager-api/src/v1/controllers/acbs.controller.js
abhi-markan: This is in case, `id` does not exists so the first condition would stop it there and prevent from crashing.

IgnatG: it wouldn't crash :) - it would simply return `undefined`
in this scenario it would crash `acbsTaskLinks.anotherValue.id` - if `anotherValue` is undefined, then it would crash

IgnatG (approved):

PR #1139, opened 2022-08-26
chore(git-ignore): Added data migration directories

abhi-markan:
## Introduction
Data migration source directories contain sensitive information which should not be added to any VCS.

## Resolution
* Added `tfm/amendments` and `tfm/actionsheets` to `.gitignore`.


IgnatG (approved):

Zainzzkk (approved):
LGTM

PR #1140, opened 2022-08-26
fix(DTFS2-5970): BSS exposure rounded off

Zainzzkk:
# Issue:
* Exposure values for BSS facilities were rounded to the nearest number so decimal places were always 00

# Changes made:
* Removed split which cut off decimal places in mapping function for BSS
* Added to tests


abhi-markan (approved):
LGTM 👍🏼

PR #1144, opened 2022-08-30
fix(DTFS2-5938): TFM Premium Schedule - Facility currency

abhi-markan:
## Introduction
Premium Schedule label in TFM was hardcoded with `GBP` as a default currency, despite PS API having no concept of currency default facility currency should be used instead of `GBP`. This has been confirmed with internal teams.

This PR aims to mitigate this UI issue.

## Resolution
* Facility currency usage.
* If property `currency` does not exists then `GBP` will be the default currency.


Zainzzkk (approved):
lgtm

PR #1151, opened 2022-08-31
fix(DTFS2-gitignore): Added sensitive directories

abhi-markan:
## Introduction
`utils` directory have various I/O scripts which could produce sensitive information.
This data should never make it to any VCS, in order to prevent such scenarios following paths have been added to utils `.gitignore`.

## Resolution
* exports/
* actionsheets/
* json/
* tfm/amendments


Zainzzkk (approved):
LGTM

PR #1152, opened 2022-08-31
fix: EC - Criterion existence check

abhi-markan:
## Introduction
Migrated deals especially `BSS/EWCS` have various criterions which do not exists in pre-defined ECs.
Upon accessing a void criterion GQL resolver does not gets the expected data thus returning `null` data back to `TFM UI`.

## Resolution
* Criterion `id` existence check in `versionContentStrings` (pre-defined ECs)


Zainzzkk (commented):
can a test be added for this?

Zainzzkk (approved):
LGTM

PR #1153, opened 2022-09-01
fix(DTFS2-6002): TFM all facilities graphql query fix

Zainzzkk:
# Issue:
* All facilities endpoint for tfm-api (graphql) crashing when extra '-' in coverEndDate

# Changes made:
* Added regex check for coverEndDate to ensure covers right format of 'YYYY-MM-DD'


abhi-markan (approved):

trade-finance-manager-api/src/graphql/resolvers/query-facilities.js
```diff
@@ -19,15 +19,23 @@ exports.queryAllFacilities = async (queryParams) => {
     // finds latest completed amendment tfm object with mapped values
     const latestCompletedAmendment = findLatestCompletedAmendment(item?.amendments);
 
+    let facilityCoverEndDate = '-';
+    let coverEndDateEpoch = '';
+    // regex to check date is correct format - YYYY-MM-DD
+    const dateRegex = /^(19|20)\d{2}-(0[1-9]|1[0-2])-(0[1-9]|1\d|2\d|3[01])$/;
```
abhi-markan, line 25: Can we simplify this please. Else looks good.

abhi-markan (changes requested):

trade-finance-manager-api/src/graphql/resolvers/query-facilities.js
Zainzzkk: done

trade-finance-manager-api/src/graphql/resolvers/query-facilities.js
```diff
@@ -19,15 +19,23 @@ exports.queryAllFacilities = async (queryParams) => {
     // finds latest completed amendment tfm object with mapped values
     const latestCompletedAmendment = findLatestCompletedAmendment(item?.amendments);
 
+    let facilityCoverEndDate = '-';
```
abhi-markan, line 22: Is there a reason why the variable `facilityCoverEndDate` has been instantiated with `-`?

abhi-markan (changes requested):
Do we need to add a test or is it already covered?

trade-finance-manager-api/src/graphql/resolvers/query-facilities.js
Zainzzkk: fixed

abhi-markan (approved):

PR #1155, opened 2022-09-01
fix(EC): Added EC Criterions version 1

abhi-markan:
## Introduction
Eligibility criteria for migrated deals especially `BSS/EWCS` predominantly work around EC version `1`.

## Resolution
* Introduction of version `1` for `BSS/EWCS` deals.
* All `25` criterions have been added.

### DIFF
--- a/trade-finance-manager-api/src/v1/content-strings/index.js
+++ b/trade-finance-manager-api/src/v1/content-strings/index.js
@@ -1,6 +1,133 @@
 const DEAL = {
   ELIGIBILITY_CRITERIA: {
     'BSS/EWCS': {
+      1: {
+        1: {
+          text: 'The Supplier has confirmed in its Supplier Supplementary Declaration that the Supply Contract does not involve agents and the Bank is not aware that any of the information contained within it is inaccurate.',
+          textList: [],
+          group: 'Supply Contract / Transaction criteria',
+        },
+        2: {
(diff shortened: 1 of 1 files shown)


Zainzzkk (approved):
LGTM

ttbarnes (approved):
LGTM - it's a tech debt item, but could change the structure so it's clear that it's a version like in [here](https://github.com/UK-Export-Finance/dtfs2/blob/main/utils/mock-data-loader/gef/eligibilityCriteria.js#L3), e.g:


```js
{
  version: 1.5,
  criteria: [ ... ] 
}
```

instead of:

```js
{
  1: {
    1: {},
    2: {},
    // ...
  }
}
```

trade-finance-manager-api/src/v1/content-strings/index.js
```diff
@@ -1,6 +1,133 @@
 const DEAL = {
   ELIGIBILITY_CRITERIA: {
     'BSS/EWCS': {
+      1: {
+        1: {
+          text: 'The Supplier has confirmed in its Supplier Supplementary Declaration that the Supply Contract does not involve agents and the Bank is not aware that any of the information contained within it is inaccurate.',
+          textList: [],
+          group: 'Supply Contract / Transaction criteria',
```
IgnatG, line 8: not a show-stopper, but you shouldn't need the `group` property

abhi-markan: I did not notice but because it is being used in old EC, I think might be best to leave it in case we need it in future?

PR #1160, opened 2022-09-06
feat(DTFS2-6022): Facility amendments updates TFM properties

abhi-markan:
## Introduction
Upon a facility amendment submission to the TFM, following TFM properties should have also been updated.
However during ACBS amendments QA neither of the properties are being updated.

* Premium Schedule (BSS/EWCS) or Fixed Fee (GEF)
* Guarantee dates
* Exposure period in months
* UKEF Exposure

## Resolution
With an introduction of a new controller responsible for TFM properties computation upon an amendment submission.
Controller is responsible for amendment evaluation, respective TFM property update, facility mapping and then eventually updating the TFM properties.

During facility update only the `facility.tfm` property is update whilst the base `facility` property is left untouched in order to adhere to `facilitySnapshot` principle.


trade-finance-manager-api/src/v1/controllers/amend-issued-facility.js
```diff
@@ -0,0 +1,172 @@
...
+        }
+
+        // Save TFM in `tfm.history`
+        history.push(tfm);
+        facilityTfmUpdate = {
+          ...facilityTfmUpdate,
+          history,
+        };
+
+        // Updated `facility.tfm` property
```
Zainzzkk, line 156: refactor in future please

Zainzzkk (approved):
LGTM - 
1) Please make future tickets to refactor and split trade-finance-manager-api/src/v1/controllers/amend-issued-facility.js into a few functions as is long
2) can we do facilityTfmUpdate.history.push in the future
3) Do we have anything to revert latest history object if amendment deleted?

abhi-markan:
> LGTM -
> 
> 1. Please make future tickets to refactor and split trade-finance-manager-api/src/v1/controllers/amend-issued-facility.js into a few functions as is long
> 2. can we do facilityTfmUpdate.history.push in the future
> 3. Do we have anything to revert latest history object if amendment deleted?

Hello will do.
Nothing in place to delete a tfm.history if the amendment is deleted can be added in.

PR #1162, opened 2022-09-08
fix(DTFS2-6037): Year format remove error message

Zainzzkk:
# Changes made:
* Only run isBefore or isAfter check on amendment request date if year is in correct format
* Updated tests


abhi-markan (approved):
LGTM

PR #1165, opened 2022-09-09
fix: GQL resolver exceptional handling and optional chaining

abhi-markan:
## Introduction
Upon an unsuccessful resolution of a GQL in `TFM-API` no `deal` object is returned.
TFM-UI crashes when accessing void property.

## Resolution
* Exceptional handling TFM-API GQL resolver.
* Optional chaining of `response?.data?.deal`


Zainzzkk (approved):
LGTM

PR #1170, opened 2022-09-13
fix(DTFS2-6035): TFM UI - Indemnifier address overlap

abhi-markan:
## Introduction

When an `indemnifier` is present with its correspondence address and country and when logged in as a non-editable user address overlaps with the heading on the left this was due to a child DOM element with class `govuk-grid-column-one-half` exists inside a parent DOM element with same class, thus reducing the width to `50%`.

## Resolution
* Removal of condition and `<div>` object as they do not serve any meaningful purpose.

### DIFF
--- a/trade-finance-manager-ui/templates/case/parties/_macros/parties-indemnifier-area.njk
+++ b/trade-finance-manager-ui/templates/case/parties/_macros/parties-indemnifier-area.njk
@@ -102,10 +102,7 @@
             key: 'Approval level required',
             value: ''
           }) }}
-        {% if not userCanEdit %}
-          </div>
-          <div class="govuk-grid-column-one-half">
-        {% endif %}
+
           {{ keyValueArrayGridRow.render({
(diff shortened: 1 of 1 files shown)


trade-finance-manager-ui/templates/case/parties/_macros/parties-indemnifier-area.njk
```diff
@@ -102,10 +102,7 @@
             key: 'Approval level required',
             value: ''
           }) }}
-        {% if not userCanEdit %}
```
Zainzzkk, line 105: Does it look fine if the window is made smaller?

abhi-markan: Yep

Zainzzkk (approved):
LGTM

IgnatG (approved):

PR #1174, opened 2022-09-15
fix(DTFS2-ACBS): Mandatory field verfication + Updated Package lock

abhi-markan:
## Introduction
Mandatory field computation does not take into account field with value `0` which is an acceptable value.

## Resolution
* `null`, `undefined` and `empty` string check.

## Miscellaneous
* Updated `package-lock.json`.


abhi-markan:
> is there a reason the package.lock changes?

Mainly `integrity` and `resolved` attribute.

Zainzzkk (approved):
LGTM

PR #1175, opened 2022-09-16
fix(DTFS2-6047): GEF validation issue facility 

Zainzzkk:
# Issue:
* Issuing facility page went to error page if special characters in date boxes

# Changes made:
* Added validation on issue facility page if errors 
* Added helper
* Added tests

### DIFF
--- a/gef-ui/server/controllers/unissued-facilities/validation.test.js
+++ b/gef-ui/server/controllers/unissued-facilities/validation.test.js
@@ -345,7 +345,7 @@ describe('validation()', () => {
     expect(result.aboutFacilityErrors).toEqual(expectedFacilityErrors);
   });
 
-  it('should return object with errors populated if issue date has errors', async () => {
+  it('should return object with errors populated if issue date has only symbols', async () => {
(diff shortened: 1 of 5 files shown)


gef-ui/server/controllers/unissued-facilities/validation.test.js
```diff
@@ -345,6 +345,296 @@ describe('validation()', () => {
     expect(result.aboutFacilityErrors).toEqual(expectedFacilityErrors);
   });
 
+  it('should return object with errors populated if issue date has errors', async () => {
```
abhi-markan, line 348: Can abstract the test cases to more illustrative blocks i.e. 
* should return error when date is blank...
* Should return error when date has symbol..

Can we please add a test case for either date/month/year to be blank please.

abhi-markan (changes requested):
Good work, just one minor comment to address please.

gef-ui/server/controllers/unissued-facilities/validation.test.js
Zainzzkk: done

abhi-markan (approved):
LGTM

PR #1178, opened 2022-09-20
fix: ACBS Exposure period - Null cover start and end date

abhi-markan:
## Introduction
When cover start date and end date is null due to facility not being issued, exposure period helper function returns `NaN` for `BSS/EWCS` deals.

## Resolution
* Returns `exposure` when both month duration and month off set are void.


Zainzzkk (approved):
LGTM

PR #1181, opened 2022-09-23
fix(DTFS2-6056): All facilities - fix dealId conversion to string

Zainzzkk:
# Issue: 
* MIN facilities do not show up on all facilities tab
* MIN tfm-deal dealSnapshot._id converted to string instead of ObjectId

# Changes made:
* returned dealUpdate from `updatePortalDealFromMIAtoMIN` as compared to returning full deal 
* removed full update body from update in `updateDealSnapshot` 
* Added extra tests

### DIFF
--- a/trade-finance-manager-api/api-tests/v1/deals/deals-submit.second-submission.api-test.js
+++ b/trade-finance-manager-api/api-tests/v1/deals/deals-submit.second-submission.api-test.js
@@ -654,7 +654,7 @@ describe('/v1/deals', () => {
 
         expect(status).toEqual(200);
         expect(body.submissionType).toEqual(CONSTANTS.DEALS.SUBMISSION_TYPE.MIN);
-        expect(typeof body.manualInclusionNoticeSubmissionDate).toEqual('number');
+        expect(typeof body.manualInclusionNoticeSubmissionDate).toEqual('string');
       });
     });
   });
--- a/trade-finance-manager-api/src/v1/controllers/update-portal-deal-from-MIA-to-MIN.api-test.js
+++ b/trade-finance-manager-api/src/v1/controllers/update-portal-deal-from-MIA-to-MIN.api-test.js
@@ -1,0 +1,58 @@
+const { updatePortalDealFromMIAtoMIN } = require('./update-portal-deal-from-MIA-to-MIN');
+const api = require('../api');
+
+const CONSTANTS = require('../../constants');
+
+describe('updatePortalDealFromMIAtoMIN()', () => {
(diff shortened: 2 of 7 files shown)


trade-finance-manager-api/api-tests/v1/deals/deals-submit.second-submission.api-test.js
```diff
@@ -654,7 +654,7 @@ describe('/v1/deals', () => {
 
         expect(status).toEqual(200);
         expect(body.submissionType).toEqual(CONSTANTS.DEALS.SUBMISSION_TYPE.MIN);
-        expect(typeof body.manualInclusionNoticeSubmissionDate).toEqual('number');
+        expect(typeof body.manualInclusionNoticeSubmissionDate).toEqual('string');
```
abhi-markan, line 657: Why is this changed to `string` please?

trade-finance-manager-api/src/v1/controllers/update-portal-deal-from-MIA-to-MIN.api-test.js
```diff
@@ -0,0 +1,58 @@
...
+  });
+
+  it('should return the dealUpdate object on BSS deal update', async () => {
+    const update = await updatePortalDealFromMIAtoMIN(dealId, CONSTANTS.DEALS.DEAL_TYPE.BSS_EWCS, checker);
+
+    const dealUpdate = {
+      submissionType: CONSTANTS.DEALS.SUBMISSION_TYPE.MIN,
+      details: {
+        checkerMIN: checker,
+        manualInclusionNoticeSubmissionDate: expect.any(String),
```
abhi-markan, line 46: LOC:33 and :46 should be `number`?

abhi-markan (changes requested):
Two minor comments to address please.

trade-finance-manager-api/src/v1/controllers/update-portal-deal-from-MIA-to-MIN.api-test.js
Zainzzkk: it's using toString() on the now function

trade-finance-manager-api/api-tests/v1/deals/deals-submit.second-submission.api-test.js
Zainzzkk: now does .toString() so not sure why it was a number in the first place

abhi-markan (approved):

PR #1182, opened 2022-09-28
fix(DTFS2-6065): Null ukefFacilityId fix

Zainzzkk:
# Issue:
* If ukefFacilityId is null, tfm-ui crashes as graphql replaces facility with null on all facilities page mapping

# Changes made:
* Replaced null with '-' on query-facilities resolver
* Added test


abhi-markan (approved):
LGMT

PR #1193, opened 2022-10-05
fix: `ObjectId` facility IDs instead of `String`

abhi-markan:
## Introduction
When opening a deal in TFM especially *migrated deal* and upon facility IDs fetch from `deal.dealSnapshot.facilities` these IDs should be stored as `ObjectId` rather than `String`. 

If stored as a `String` ID, GQL will thrown an exception due to MongoDB query fetch will result in void dataset return.

## Resolution
* Facility IDs to be stored as `ObjectId` in `facilityIds` array.


Zainzzkk (commented):
If we change the facilities to ObjectId instead of string, then this will not be needed? (future MIA->MIN will always be ObjectIds)

abhi-markan:
> If we change the facilities to ObjectId instead of string, then this will not be needed? (future MIA->MIN will always be ObjectIds)

They will be changed to `ObjectId` however we do need this explicit declaration in my opinion.

Zainzzkk (approved):
LGTM

PR #1196, opened 2022-10-10
feat(DTFS2-5977): Internal UKEF email notification

abhi-markan:
## Introduction
When an amendment is received an internal email notification is required so manual amendments processing of various facilities records in ACBS can be performed by a respective team.

## Resolution
* Addition of a bespoke `GovNotify` template.
* Introduction of `internalAmendmentEmail` function to send an email upon amendment submission (automatic) and approval (if manual).

## Pre-requisite
Internal email address will be read from micro-service environmental variable.
Please ensure `UKEF_INTERNAL_NOTIFICATION` variable exists in respective environment.

### DIFF
--- a/docker-compose.gha.yml
+++ b/docker-compose.gha.yml
@@ -78,6 +78,7 @@ services:
       - SENTRY_DSN=${SENTRY_DSN}
       - JWT_SIGNING_KEY=${JWT_SIGNING_KEY}
       - TZ=Europe/London
+      - UKEF_INTERNAL_NOTIFICATION=test
     entrypoint: npx nodemon src/index.js -L
 
   portal:


trade-finance-manager-api/src/v1/helpers/amendment.helpers.js
```diff
@@ -311,6 +311,30 @@ const sendFirstTaskEmail = async (taskVariables) => {
...
+const internalAmendmentEmail = async (ukefFacilityId) => {
+  try {
+    const templateId = EMAIL_TEMPLATE_IDS.INTERNAL_AMENDMENT_NOTIFICATION;
+    const sendToEmailAddress = process.env.UKEF_INTERNAL_NOTIFICATION;
+    const emailVariables = {
+      ukefFacilityId,
+    };
+
+    if (!templateId || !sendToEmailAddress || !ukefFacilityId) {
+      throw new Error('Void argument specified');
```
IgnatG, line 328: do we need this? it should be captured by the `try` block if the `sendTfmEmail` fails

trade-finance-manager-api/src/v1/helpers/amendment.helpers.js
```diff
@@ -311,6 +311,30 @@ const sendFirstTaskEmail = async (taskVariables) => {
...
+    }
+
+    return sendTfmEmail(templateId, sendToEmailAddress, emailVariables);
+  } catch (error) {
+    console.error('Error sending internal amendment email', { error });
+    return false;
```
IgnatG, line 334: can we return a status/message instead? `false` may not be descriptive

trade-finance-manager-api/src/v1/helpers/amendment.helpers.js
abhi-markan: I have replaced with return `false`.
GovNotify API will return appropriate response, but there is no check in place to verify whether the `emailVariable` object is valid or not. I think need a check in place to avoid invoking API if `ukefFacilityId` is null.

trade-finance-manager-api/src/v1/helpers/amendment.helpers.js
abhi-markan: Good shout, have updated accordingly.

trade-finance-manager-api/src/v1/helpers/amendment.helpers.js
```diff
@@ -311,10 +311,34 @@ const sendFirstTaskEmail = async (taskVariables) => {
...
+    const sendToEmailAddress = process.env.UKEF_INTERNAL_NOTIFICATION;
+    const emailVariables = {
+      ukefFacilityId,
+    };
+
+    if (!ukefFacilityId) {
+      return false;
+    }
+
+    return sendTfmEmail(templateId, sendToEmailAddress, emailVariables);
```
IgnatG, line 331: SonarCloud is flagging this as a potential bug because it needs the `await` part. is eslint saying something else? are you able to have a look?

abhi-markan: Yep `eslint` doesn't like `await` and woud like `return`.
I had `await` before.

PR #1201, opened 2022-10-14
feat(DTFS2-6023): Facility amendment SOFs

abhi-markan:
## Introduction
Facility amendment could consists of both value increment or value decrement.
Value increment works as expected, however decrement require facility records amendment to be swapped.

In a standard value increment records are amended in following sequence:

* FMR
* FLR

However, in case of decrement records are amended in reverse:

* FLR
* FMR

## Resolution
* Execution sequence are swapped upon operation type.
* GEF facility amendment bug resolution with regards to 10% of the amendment value and condition update for payload construction.

## Miscellaneous
* Documentation update

### DIFF
--- a/azure-functions/acbs-function/acbs-amend-facility-loan-record/index.js
+++ b/azure-functions/acbs-function/acbs-amend-facility-loan-record/index.js
@@ -1,0 +1,87 @@
+/**
+ * Facility Loan Record amendment SOF
+ * **********************************
+ * This is a sub-orchestrated function invoked from it's master orchestrator.
+ * `acbs-amend-facility` is the master orchestrator for this function.
+ *
+ * Sub-Orchestration Function (SOF)
+ * --------------------------------
+ * This function cannot be invoked directly and must be invoked by a DOF.
+ *
+ * Prerequisites
+ * -------------
+ * 0. 'npm install durable-functions'
+ * 1. Durable HTTP starter function (acbs-http)
+ * 2. DOF (acbs-amend-facility)
+ * 3. DAF (activity-get-loan-id, activity-update-facility-loan-amount, activity-update-facility-loan)
+ *
+ * ACBS
+ * ----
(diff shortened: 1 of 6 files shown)


azure-functions/acbs-function/acbs-amend-facility-loan-record/index.js
```diff
@@ -0,0 +1,87 @@
...
+ *
+ * ACBS
+ * ----
+ * This function is responsible for amending FLR.
+ * FLR is one of a child record of an issued facility.
+ */
+const df = require('durable-functions');
+const retryOptions = require('../helpers/retryOptions');
+const mappings = require('../mappings');
+const CONSTANTS = require('../constants');
```
IgnatG, line 26: If we only need the Facility, then we can destructure it here

azure-functions/acbs-function/acbs-amend-facility-master-record/index.js
```diff
@@ -0,0 +1,80 @@
...
+ * This is a sub-orchestrated function invoked from it's master orchestrator.
+ * `acbs-amend-facility` is the master orchestrator for this function.
+ *
+ * Sub-Orchestration Function (SOF)
+ * --------------------------------
+ * This function cannot be invoked directly and must be invoked by a DOF.
+ *
+ * Prerequisites
+ * -------------
+ * 0. 'npm install durable-functions'
```
IgnatG, line 13: Do we need this line?

abhi-markan: Yep, especially for a clear documentation.

azure-functions/acbs-function/acbs-amend-facility-loan-record/index.js
abhi-markan: Good spot, have updated.

IgnatG (approved):
LGTM

PR #1202, opened 2022-10-18
fix: Void user ID

abhi-markan:
## Introduction
Since migration deals will have user ID which either are invalid, do not exists or are null.
Given the above predicament, portal activities and comments controller should be able to handle a null response from DB `findOne` API call.

## Resolution
* Conditional check on `userProfile`.
* Instantiation  of `firsname` and `lastname` empty string variables.

### DIFF
--- a/portal-api/src/v1/gef/controllers/portal-activities.controller.js
+++ b/portal-api/src/v1/gef/controllers/portal-activities.controller.js
@@ -9,14 +9,17 @@ const CONSTANTS = require('../../../constants');
 // retrieves user information from database
 const getUserInfo = async (userId) => {
   const userCollectionName = 'users';
+  let firstname = '';
+  const surname = '';
 
   const userCollection = await db.getCollection(userCollectionName);
-  const {
-    firstname,
-    surname = '',
-  } = userId
+  const userProfile = userId
     ? await userCollection.findOne({ _id: ObjectId(String(userId)) })
     : {};
+
+  if (userProfile) {
+    firstname = userProfile.firstName;
+  }
 
   // creates user object which can be used
   const user = {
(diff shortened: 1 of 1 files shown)


portal-api/src/v1/gef/controllers/portal-activities.controller.js
```diff
@@ -9,21 +9,25 @@ const CONSTANTS = require('../../../constants');
...
-  const {
-    firstname,
-    surname = '',
-  } = userId
+  const userProfile = userId
     ? await userCollection.findOne({ _id: ObjectId(String(userId)) })
     : {};
 
+  if (userProfile) {
+    firstname = userProfile.firstName;
```
Zainzzkk, line 21: how is the surname added?

Zainzzkk (commented):
Can you add a test please?

Zainzzkk (approved):
LGTM

portal-api/api-tests/v1/gef/portal-activities.controller.api-test.js
```diff
@@ -291,4 +291,23 @@ describe('getUserInfo()', () => {
...
   });
+
+  it('Should return empty `firstname` and `lastname` when user does not exists', async () => {
+    // Wipe existing collections
+    await wipeDB.wipe([collectionName]);
+    await wipeDB.wipe([applicationCollectionName]);
+
+    // Fetch user from the collection
+    const userCollection = await db.getCollection('users');
+    const checker = await userCollection.findOne({ _id: '0' });
```
IgnatG, line 302: we need to avoid mongo queries in tests. In the long run, I'd like to remove it altogether and keep only API calls to various endpoints.

can we use an API call instead to get these details?

PR #1212, opened 2022-10-26
fix(DTFS2-6116): Facilities mapping on existence only

abhi-markan:
## Introduction
Certain `Abandoned` deals exists without any facilities, `map` method should only be applied on valid facilities array.

## Resolution
* Facilities array length check


trade-finance-manager-api/src/graphql/reducers/mappings/gef-facilities/mapGefFacilities.js
```diff
@@ -3,10 +3,15 @@ const mapGefFacility = require('./mapGefFacility');
 const mapGefFacilities = (dealSnapshot, dealTfm) => {
   const { facilities } = dealSnapshot;
 
-  const mappedFacilities = facilities.map((facility) =>
-    mapGefFacility(facility, dealSnapshot, dealTfm));
+  // Map facilities if only they exists
+  if (facilities?.length) {
```
Zainzzkk, line 7: do we need optional chaining here? or can it just be facilities.length

abhi-markan: Facilities can be `unidentified` in certain occasions.

Zainzzkk (approved):
Please merge main first

PR #1216, opened 2022-10-27
fix(DTFS2-6116): Object validation + DTFS Central API Resource upgrade

abhi-markan:
## Introduction
DTFS central API resource starvation has been an issue with returning GQL data.

## Resolution
* Express json limit has been increased to `10mb`
* Empty facility object check

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -32,7 +32,7 @@ services:
       resources:
         limits:
           cpus: "0.1"
-          memory: 400M
+          memory: 1024M
     environment:
       - PORT=5005
       - REFERENCE_DATA_PROXY_URL=http://reference-data-proxy:5002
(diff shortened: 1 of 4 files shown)


docker-compose.yml
```diff
@@ -32,7 +32,7 @@ services:
       resources:
         limits:
           cpus: "0.1"
-          memory: 400M
+          memory: 1024M
```
IgnatG, line 35: docker compose is not executed in Azure.
We need to increase the limit in `createApp.js` file in central-api
currently the limit is: `app.use(express.json({ limit: '500kb' }));`
Can we increase it to 20mb to see what happens? only for Dev env first

abhi-markan: Sorry, have pushed.

IgnatG (approved):

PR #1219, opened 2022-10-28
fix(DTFS2-6116): Payload size & void object check

abhi-markan:
## Introduction
Increment of payload size has no effect on the DTFS central API facility fetch calls (due to large number of facilities especially post data-migration) it is prudent to revert it back to its original intended size.

## Resolution
* Payload limit reverted back to `500kb`.
* Void object check to return `null` or empty array if premium schedule.
* Removed `console.log`.


IgnatG (approved):

e2e-tests/trade-finance-manager/cypress/plugins/index.js
```diff
@@ -0,0 +1,22 @@
...
+//
+// You can read more here:
+// https://on.cypress.io/plugins-guide
+// ***********************************************************
+
+// This function is called when a project is opened or re-opened (e.g. due to
+// the project's config changing)
+
+/**
+ * @type {Cypress.PluginConfig}
```
IgnatG, line 16: do we need this file?

abhi-markan: Cypress was failing without it, locally.
Could be due to Cypress v10?

IgnatG: in Cypress 10 it's not needed anymore
did you run `npm install` in the root folder?

IgnatG (approved):

PR #1237, opened 2022-11-11
feat(DTFS2-5960): Portal facilities created by you

Zainzzkk:
# Changes made:
* Added created by you logic to facilities page in filters box
* Added correct matching query as calling joined deals table
* Added e2e test and unit tests


abhi-markan (approved):
LGTM

PR #1250, opened 2022-12-07
feat(DTFS2-6224): Added Emirates NBD Bank

abhi-markan:
## Introduction
A new bank has board and signed MGA.

## Resolution
* Added bank name to the portal UI login screen NJK template.


Zainzzkk (approved):
LGTM - add e2e test later

PR #1251, opened 2022-12-09
feat(tfm-reporting): TFM reporting

abhi-markan:
## Introduction
A bespoke TFM report is generated upon a request, in order for a swift report generation a simple node script has been written based on @IgnatG initial work.

This script reads the data from `tfm-deals` collection and construct the rows as an `array` which then is passed over to `generateReport` function responsible for appending processed data array at the bottom of the bespoke columns. Once amalgamated and converted into CSV data is then passed to file write method.

### DIFF
--- a/utils/reporting/tfm/tfm-reporting.js
+++ b/utils/reporting/tfm/tfm-reporting.js
@@ -12,12 +12,14 @@ const { write } = require('../../data-migration/helpers/io');
 // ******************** DEALS *************************
 /**
    * Return all the TFM deals with `MIA/MIN` filter.
-   * @param {Object} filter Mongo filter
+   * @param {Integer} EPOCH Fetch records greater than.
+   * Defaulted to `1648684800` (31-03-2022). This argument
+   * accepts EPOCH with `ms`
    * @returns {Object} Collection object
    */
-const getTfmDeals = () => getCollection(
+const getTfmDeals = (epoch = 1648684800) => getCollection(
   CONSTANTS.DATABASE.TABLES.TFM_DEAL,
-  { 'tfm.stage': 'Confirmed', 'tfm.dateReceivedTimestamp': { $gte: 1648684800 } },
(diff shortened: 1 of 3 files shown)


utils/reporting/tfm/tfm-reporting.js
```diff
@@ -0,0 +1,180 @@
...
+
+// ******************** DEALS *************************
+/**
+   * Return all the TFM deals with `MIA/MIN` filter.
+   * @param {Object} filter Mongo filter
+   * @returns {Object} Collection object
+   */
+const getTfmDeals = () => getCollection(
+  CONSTANTS.DATABASE.TABLES.TFM_DEAL,
+  { 'tfm.stage': 'Confirmed', 'tfm.dateReceivedTimestamp': { $gte: 1648684800 } },
```
IgnatG, line 20: I wonder if `1648684800` can actually be a default value instead that can be overwritten if needed?
So the function can look something like this:
```
const getTfmDeals = (timestamp = 1648684800) => getCollection(...)
```

we can also add a comment that `1648684800` means  1 April 2022 in EPOCH (without milliseconds)

abhi-markan: Good suggestion, have amended.

utils/reporting/tfm/tfm-reporting.js
```diff
@@ -0,0 +1,182 @@
...
+// ******************** FORMATTING *************************
+
+/**
+    * Strips commas and return UKEF exposure
+    * as number.
+    * @param {String} string Raw string with commas
+    * @param {Boolean} Number Whether output should be a `Number`
+    * @returns {Integer} Formatted value
+    */
+const stripCommas = (string, number = false) => {
```
IgnatG, line 34: I've seen we have another file called `rad-reporting.js`. This function exists in that file as well. Is it worth creating a helper function and loading these in? to avoid duplications?

abhi-markan: Excellent point @IgnatG have re-factored code.

IgnatG (approved):
LGTM

PR #1261, opened 2022-12-28
feat(DTFS2-6237): TFM Non Delegated Banks amendment tasks

Zainzzkk:
# Changes made:
* Added tasks for non delegated banks amendments
* Added to controller to process these tasks
* Added facility call to check if non-delegated bank flag is set
* Added api-test

### DIFF
--- a/trade-finance-manager-api/src/v1/helpers/create-tasks-amendment.helper.api-test.js
+++ b/trade-finance-manager-api/src/v1/helpers/create-tasks-amendment.helper.api-test.js
@@ -1,0 +1,52 @@
+const { createAmendmentTasks, createTasksAutomaticAmendment, createTasksManualAmendment } = require('./create-tasks-amendment.helper');
+
+describe('createAmendmentTasks()', () => {
+  const mockFacility = {
+    requireUkefApproval: false,
+    tfm: {},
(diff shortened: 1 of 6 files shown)


trade-finance-manager-api/src/v1/helpers/create-tasks-amendment.helper.api-test.js
```diff
@@ -0,0 +1,52 @@
...
+
+  it('should create automatic amendment tasks if !requireUkefApproval and !nonDelegatedBank', () => {
+    const response = createAmendmentTasks(mockFacility.requireUkefApproval, mockFacility.tfm);
+
+    const expected = createTasksAutomaticAmendment();
+
+    expect(response).toEqual(expected);
+  });
+
+  it('should create manual amendment tasks if requireUkefApproval and !nonDelegatedBank', () => {
```
abhi-markan, line 17: Can we add a case which covers following please:
* When `not` a task should be created for a `NDB.
* When a manual deal tasks should be created for `NDB`.

abhi-markan (changes requested):
One minor comment please.

abhi-markan (approved):

PR #1264, opened 2022-12-29
fix(DTFS2-6239): Facility level default currency

abhi-markan:
## Introduction
Due to a know tech-debt issue, on certain occasion `facility.currency.id` does not exists, due to which perils are high when creating records in ACBS especially during facility level records.

## Resolution
* If the `currency.id` is absent then currency will be defaulted to `GBP`.


ttbarnes (approved):

Zainzzkk (approved):
LGTM

PR #1272, opened 2022-12-30
feat(DTFS2-6213): NDB EC mapping

abhi-markan:
## Introduction
Since Non-delegated banks do not have EC, this needs to be reflected across GQL mapping.

## Resolution
* Conditional branching.


Zainzzkk (approved):
LGTM - after failing API test

PR #1297, opened 2023-02-03
feat(DTFS2-6283): Bond - Facility loan record amendment

abhi-markan:
## Introduction
Facility loan record should only be amended for facility type `bond`.

## Resolution
* Ensure facility is type `bond`.
* Removed `bond` verification from `acbs-amend-facility-loan-record` DAF.


Zainzzkk (approved):
LGTM

PR #1318, opened 2023-02-15
fix(no-activities): Prevents TFM crash when no acitivity exist

abhi-markan:
## Introduction
TFM UI crashes when `tfm.activities` object does not exist.
Albeit this is an edge case and represent a corrupt deal submission in TFM, TFM UI should not crash.

## Resolution
* Object void check.

### DIFF
--- a/trade-finance-manager-ui/server/controllers/case/activity/index.js
+++ b/trade-finance-manager-ui/server/controllers/case/activity/index.js
@@ -8,19 +8,25 @@ const { DEAL } = CONSTANTS;
 
 const MAX_COMMENT_LENGTH = 1000;
 
-const mappedActivities = (activities) => activities.map((activity) => ({
-  label: {
-    text: activity.label,
-  },
-  text: activity.text,
-  datetime: {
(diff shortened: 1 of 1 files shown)


trade-finance-manager-ui/server/controllers/case/activity/index.js
```diff
@@ -8,19 +8,25 @@ const { DEAL } = CONSTANTS;
...
-    timestamp: fromUnixTime(new Date(activity.timestamp)),
-    type: 'datetime',
-  },
-  byline: {
-    text: `${activity.author.firstName} ${activity.author.lastName}`,
-  },
-}));
+const mappedActivities = (activities) => {
+  if (!activities) {
+    return false;
```
ttbarnes, line 13: Would an empty array be better here? Since if there is data it returns an array

abhi-markan: Would need to return false to keep it consistent with other function.

IgnatG (approved):

PR #1329, opened 2023-02-27
feat(avoid-duplicate-clicks): preventDoubleClicks set to false

abhi-markan:
## Introduction
A known bug in GDS button component, allows user to trigger an execution twice by double clicking a button.
This has caused an amendment, Party URN to be triggered twice.

## Resolution
* Added `preventDoubleClick` set to `true` via `govuk-frontend.js`.
* Lint fixes.
* Dependencies updates.


gef-ui/package.json
```diff
@@ -77,6 +77,7 @@
     "@babel/preset-env": "7.20.2",
     "@babel/runtime": "7.20.7",
     "@types/jest": "^29.2.5",
+    "ajv": "^8.12.0",
```
IgnatG, line 80: is this package used anywhere?

abhi-markan: Yes it is being used during `npm run build`.

IgnatG (approved):
LGTM

PR #1331, opened 2023-02-27
fix: updated the eStore intervals

IgnatG:
### Changelog
- updated the eStore intervals (to match the changes in its API)
- updated the stage when TFM deals are updated

### DIFF
--- a/reference-data-proxy/src/cronJobs/eStoreDealFolderCreationJob.cron.ts
+++ b/reference-data-proxy/src/cronJobs/eStoreDealFolderCreationJob.cron.ts
@@ -1,5 +1,3 @@
-// eslint-disable-next-line @typescript-eslint/no-var-requires
-const { ObjectId } = require('mongodb');
 import addSeconds from 'date-fns/addSeconds';
 import { getCollection } from '../database';
 import { Estore } from '../interfaces';
@@ -11,7 +9,6 @@ import { createDealFolder } from '../v1/controllers/estore/eStoreApi';
 export const eStoreDealFolderCreationJob = async (eStoreData: Estore) => {
   try {
     const cronJobLogsCollection = await getCollection('cron-job-logs');
-    const tfmDealsCollection = await getCollection('tfm-deals');
 
     // create the Deal folder
     console.info('API Call started: Create the Deal folder for ', eStoreData.dealIdentifier);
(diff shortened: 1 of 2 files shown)


reference-data-proxy/src/cronJobs/eStoreDealFolderCreationJob.cron.ts
```diff
@@ -28,9 +25,6 @@ export const eStoreDealFolderCreationJob = async (eStoreData: Estore) => {
       // stop and the delete the cron job to release the memory
       eStoreCronJobManager.deleteJob(`Deal${eStoreData.dealId}`);
 
-      // update the `tfm-deals` collection once the buyer and deal folders have been created
-      tfmDealsCollection.updateOne({ _id: ObjectId(eStoreData.dealId) }, { $set: { 'tfm.estore': { siteName: eStoreData.siteName } } });
```
abhi-markan, line 32: Are we not going to update the collection when deal folder is created?

abhi-markan (changes requested):

reference-data-proxy/src/cronJobs/eStoreTermStoreAndBuyerFolder.cron.ts
```diff
@@ -1,3 +1,5 @@
+// eslint-disable-next-line @typescript-eslint/no-var-requires
```
Zainzzkk, line 1: is there a way to fix this without disable?

reference-data-proxy/src/cronJobs/eStoreDealFolderCreationJob.cron.ts
IgnatG: I've moved it earlier on in the process. it's in this file:
`reference-data-proxy/src/cronJobs/eStoreTermStoreAndBuyerFolder.cron.ts`

reference-data-proxy/src/cronJobs/eStoreTermStoreAndBuyerFolder.cron.ts
IgnatG: nope - because import doesn't work for this version of mongo :(

reference-data-proxy/src/cronJobs/eStoreTermStoreAndBuyerFolder.cron.ts
```diff
@@ -48,8 +50,8 @@ export const eStoreTermStoreAndBuyerFolder = async (eStoreData: any) => {
     { returnDocument: 'after' },
   );
 
-  // ensure that there is only 1 retry for buyer folder creation
-  if (response?.value?.buyerFolderRetries === 1) {
+  // ensure that there is less than 3 retries for buyer folder creation
+  if (response?.value?.buyerFolderRetries <= 3) {
```
Zainzzkk, line 54: for the future, wondering if its best to put 3 and 99 in constants so easier to change?

Zainzzkk (approved):
LGTM

abhi-markan (approved):

PR #1340, opened 2023-03-24
fix(DTFS2-6352): TFM GEF EC 2.15 updated to reflect MGA

abhi-markan:
## Introduction
Following GEF ECs have been updated in line with the MGA:

* EC 2 15
* EC 1.5 15

## Resolution
* Hard coded content string updated as per the ticket.

## Miscellaneous
* Dependencies updates.

### DIFF
--- a/trade-finance-manager-api/src/v1/content-strings/index.js
+++ b/trade-finance-manager-api/src/v1/content-strings/index.js
@@ -379,7 +379,7 @@ const DEAL = {
             'human cloning',
             'pornography',
             'tobacco or gambling',
-            'coal or coal power related',
+            'coal, oil, gas or fossil fuel energy',
           ],
         },
         16: {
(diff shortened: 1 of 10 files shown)


trade-finance-manager-api/src/v1/content-strings/index.js
```diff
@@ -379,7 +379,7 @@ const DEAL = {
             'human cloning',
             'pornography',
             'tobacco or gambling',
-            'coal or coal power related',
+            'coal, oil, gas or fossil fuel energy',
```
ttbarnes, line 382: LGTM, but should this be a new version i.e `2.1`? 

Or if it's missing in this version, are there data updates and notifications to end users required, for deals that have already been submitted?

abhi-markan: Yes we did miss it in `2.0`.

ttbarnes: OK - RE previously submitted deals and changing the EC, I think this OK - just noticed this is for TFM, so internal only. I was thinking about the user facing EC's, didn't spot it was for TFM only

ttbarnes (approved):

IgnatG (approved):

PR #1577, opened 2023-04-26
fix: removed dependabot

IgnatG:
#### Changelog
- removed `dependabot.yml` since it's polluting the PR with requests
- Reverted various breaking dependencies.


abhi-markan:
Closing, a new PR #1578 has been introduced to tackle breaking deps.

PR #1632, opened 2023-05-10
fix(DTFS2-6419): `/currency/exchange` endpoint migration

abhi-markan:
## Introduction
DTFS APIM business testing for MDM endpoints.
Provided a high DTFS exposure to MDM endpoints, migration PR will be created on individual basis to ease PR reviews and aide any troubleshooting.

## Resolution
* Replaced endpoint specific environment variable and converted to APIM gateway generic environment variable, this will reduce number of variables and also acts as a future proofing catalyst.
* Removed `MULESOFT_API_CURRENCY_EXCHANGE_RATE_URL`, `MULESOFT_API_CURRENCY_EXCHANGE_RATE_KEY` and `MULESOFT_API_CURRENCY_EXCHANGE_RATE_SECRET` environment variables.
* Improved documentation


reference-data-proxy/src/v1/controllers/currency-exchange-rate.controller.ts
```diff
@@ -20,10 +61,14 @@ export const getExchangeRate = async (req: Request, res: Response) => {
     actualTarget = source;
   }
 
+  const url = exchangeRateDate
```
IgnatG, line 64: 80% of this is the same on either side. I wonder if this is easier to read:
```js
  let url = `${exchangeRateURL}currencies/exchange?source=${actualSource}&target=${actualTarget}`;
  if (exchangeRateDate) {
    url = `${url}&exchangeRateDate=${exchangeRateDate}`
  }
```

abhi-markan: Have interpolated above.

IgnatG (approved):
LGTM

PR #1634, opened 2023-05-11
feat(DTFS2-6419): MDM APIM migration

abhi-markan:
## Introduction
Migrate all MDM endpoints from Mulesoft to APIM.

## Resolution
* Removed various `MULE_` environment variables and replaced with `APIM_MDM_` respective variables.
* Added and updated `.env.sample` across various micro-services.
* Various variables and files have been renamed to reflect the migration.
* Code refactorisation
* Fixed mocked data.

## Miscellaneous
* Updated documentation
* Lint fixes
* Dependencies updates
* Removed Slack GHA integration.


azure-functions/acbs-function/apim-mdm.js
```diff
@@ -0,0 +1,36 @@
...
+  [String(process.env.APIM_MDM_KEY)]: process.env.APIM_MDM_VALUE,
+};
+
+/**
+ * APIM MDM `GET` generic endpoint invocation call
+ * @param {String} endpoint Endpoint with provided query parameters
+ * @returns Response object
+ */
+const getAPI = async (endpoint) => {
+  const response = await axios({
```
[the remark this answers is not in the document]
abhi-markan: Good spot.

reference-data-proxy/src/v1/controllers/premium-schedule.controller.ts
```diff
@@ -77,10 +88,10 @@ export const getPremiumSchedule = async (req: Request, res: Response) => {
     return res.status(400).send();
   }
 
-  if (postPremiumScheduleResponse === 200 || postPremiumScheduleResponse === 201) {
+  if (successStatus.includes(postPremiumScheduleResponse)) {
```
IgnatG, line 91: would this return a 400/404/etc status? I'm just looking at the old validation where we were checking for 200 or 201 statues only

abhi-markan: `successStatus` array only has `200` and `201` statues, actual implementation is still the same.

azure-functions/acbs-function/apim-mdm.js
```diff
@@ -0,0 +1,36 @@
...
+      error: err.response.data.error ?? err,
+    },
+  }));
+
+  return response;
+};
+
+const getCurrency = (currencyCode) => getAPI(`currencies/${currencyCode}`);
+const getACBSIndustrySector = async (industryId) => getAPI(`sector-industries?ukefIndustryId=${industryId}`);
+const getACBSCountryCode = async (country) => getAPI(`markets?searchtext=${country}`);
```
avaitonis, line 30: We decided to rename query parameter from "searchtext" to "search"

azure-functions/acbs-function/apim-mdm.js
```diff
@@ -0,0 +1,36 @@
...
+  const response = await axios({
+    method: 'get',
+    url: `${process.env.APIM_MDM_URL}${endpoint}`,
+    headers,
+  }).catch((err) => ({
+    status: err.response.status ?? err,
+    data: {
+      error: err.response.data.error ?? err,
+    },
+  }));
```
lmarrai-sw, line 23: The logic in the `catch` has changed slightly from before. It used to be

```
catch((err) => ({
    status: err.response
      ? err.response.status
      : err,
    data: {
      error: err.response
        ? err.response.data.error
        : err,
    },
  }))
```

which looks like it's built to expect that `err.response` can sometimes be `undefined`, but the new logic would throw a `TypeError` if `err.response` is `undefined`.

Has the `err` object changed so that we know that `err.response` is always defined? Or was it never necessary in the first place?

azure-functions/acbs-function/apim-mdm.js
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * APIM MDM `GET` generic endpoint invocation call
+ * @param {String} endpoint Endpoint with provided query parameters
+ * @returns Response object
+ */
+const getAPI = async (endpoint) => {
+  const response = await axios({
+    method: 'get',
+    url: `${process.env.APIM_MDM_URL}${endpoint}`,
```
lmarrai-sw, line 16: This used to be `${process.env.MULESOFT_API_UKEF_MDM_EA_URL}/${type}` which adds a `/` between the environment variable and the input arg, but the new code isn't adding the `/`. Have we added the slash to the end of the `APIM_MDM_URL` environment variable instead?

azure-functions/number-generator-function/api.js
```diff
@@ -20,47 +24,53 @@ const getAPI = async (type) => {
...
-  console.info('Azure functions - api.postToAPI');
-  if (!process.env.MULESOFT_API_UKEF_TF_EA_URL) {
+/**
+ * Post to MDM endpoints
+ * @param {String} endpoint MDM endpoints
+ * @param {Object} payload Payload
+ * @returns API endpoint invoke response
+ */
+const postToAPI = async (endpoint, payload) => {
+  console.info('Calling number generator');
```
lmarrai-sw, line 34: It's surprising that the function seems to allow any post to MDM, based on its name and docs, but the log message is specifically about the `/numbers` endpoint. Maybe we should move the log message to the `callNumberGenerator` function instead?

reference-data-proxy/src/v1/controllers/premium-schedule.controller.ts
```diff
@@ -29,11 +37,8 @@ const postPremiumSchedule = async (premiumSchedulePayload: any) => {
   try {
     const response = await axios({
       method: 'post',
-      url: `${mdmEAurl}/premium/schedule`,
-      auth: { username, password },
-      headers: {
-        'Content-Type': 'application/json',
-      },
+      url: `${mdm}/premium/schedule`,
```
lmarrai-sw, line 40: In a lot of other places we've used `process.env.APIM_MDM_URL` as if it is already going to have a slash at the end, but in this case and lower down in the file we've added a slash in the code. Is that deliberate?

azure-functions/acbs-function/apim-mdm.js
lmarrai-sw: There are a few other places in the code where we've used `process.env.APIM_MDM_URL` without adding our own slash at the end - I've not commented on those to save you from the repeated comment, but if we do need to change it then there are a few other places to change too

lmarrai-sw (commented):
Thanks for doing this @abhi-markan!

reference-data-proxy/src/v1/controllers/premium-schedule.controller.ts
abhi-markan: Good spot, have removed forward slash.

azure-functions/number-generator-function/api.js
abhi-markan: Good spot, have re-phrased the console.info.

azure-functions/acbs-function/apim-mdm.js
abhi-markan: We have indeed added `/` to the env variable `APIM_MDM_URL`.

azure-functions/acbs-function/apim-mdm.js
abhi-markan: Good point, have reverted back.

azure-functions/acbs-function/docker-compose.yml
```diff
@@ -7,13 +7,13 @@ services:
...
-      - MULESOFT_API_KEY
-      - MULESOFT_API_SECRET
+      - APIM_MDM_URL
+      - APIM_MDM_KEY
+      - APIM_MDM_VALUE
       - MULESOFT_API_UKEF_TF_EA_URL
-      - MULESOFT_API_UKEF_MDM_EA_URL
-      - MULESOFT_API_UKEF_MDM_EA_KEY
-      - MULESOFT_API_UKEF_MDM_EA_SECRET
-      - TZ=Europe/London
```
IgnatG, line 19: Do we not need the timezone Env?

azure-functions/number-generator-function/activity-get-number-from-generator/index.js
```diff
@@ -34,24 +33,22 @@ const getNumberFromGenerator = async (context) => {
...
-  // Set the maximum number of tries in case acbsCheck is unavailable and it gets stuck in an infinite loop
+  /**
+   * Maximum tries mitigates infinite loop execution
+   */
   while (!numberIsAvailable && loopCount < MAX_NUMBER_OF_TRIES) {
-    console.info('Azure functions - getNumberFromGenerator - trying. Count ', loopCount);
-
-    numberFromGenerator = await numberGeneratorController.callNumberGenerator(numberType);
-    console.info('Azure functions - numberFromGenerator ', numberFromGenerator);
+    console.info(`⚡️ Execution #${loopCount}`);
```
IgnatG, line 40: Should we add a description about what it's executing? I e. Executing number generator

azure-functions/number-generator-function/activity-get-number-from-generator/index.js
```diff
@@ -34,24 +33,22 @@ const getNumberFromGenerator = async (context) => {
...
 
-    if (numberFromGenerator.error) {
-      throw new Error(JSON.stringify(numberFromGenerator.error));
-    }
+    number = await numberGeneratorController.callNumberGenerator(numberType);
+    console.info(`✅ ${number} successfully received`);
```
IgnatG, line 43: Is the response always successful? Just looking at the original implementation

azure-functions/number-generator-function/controllers/number-generator.controller.js
```diff
@@ -8,21 +8,50 @@
...
+      throw new Error('Void response received');
+    }
 
-  if (response.error) {
-    console.error('Azure functions - callNumberGenerator error');
+    if (!response.data.length) {
+      throw new Error('Void response length');
+    }
 
+    if (!response.data[0].maskedId) {
```
IgnatG, line 44: Shouldn't these 3 `if` statements be combined? L36 to L44. Something like:
```

if (!response?.data[0]?.maskedId) {...}
```

azure-functions/number-generator-function/numbergenerator/index.js
```diff
@@ -14,20 +14,21 @@ const retryOptions = require('../helpers/retryOptions');
...
+  if (!context.df.getInput()) {
+    throw new Error('🚩 Void input');
   }
 
   const { entityType } = context.df.getInput();
 
   if (entityType !== CONSTANTS.NUMBER_GENERATOR.ENTITY_TYPE.DEAL
     && entityType !== CONSTANTS.NUMBER_GENERATOR.ENTITY_TYPE.FACILITY) {
-    return {
-      num: 'ERROR_NUM_GENERATOR',
```
IgnatG, line 28: Can we check if this was used anywhere?

reference-data-proxy/src/v1/controllers/premium-schedule.controller.ts
```diff
@@ -10,10 +10,18 @@ import { Request, Response } from 'express';
 import { objectIsEmpty } from '../../utils';
 dotenv.config();
 
-const mdmEAurl: any = process.env.MULESOFT_API_UKEF_MDM_EA_URL;
-const username: any = process.env.MULESOFT_API_UKEF_MDM_EA_KEY;
-const password: any = process.env.MULESOFT_API_UKEF_MDM_EA_SECRET;
+const mdm: any = process.env.APIM_MDM_URL;
+const headers: any = {
+  'Content-Type': 'application/json',
```
IgnatG, line 15: We may not need this since we are always returning JSON types. But double check

azure-functions/acbs-function/apim-mdm.js
```diff
@@ -0,0 +1,45 @@
...
+      method: 'get',
+      url: `${process.env.APIM_MDM_URL}${endpoint}`,
+      headers,
+    }).catch((err) => ({
+      status: err.response
+        ? err.response.status
+        : err,
+      data: {
+        error: err.response
+          ? err.response.data.error
```
IgnatG, line 25: Do we need to check for undefined? I e response?.data?.error

IgnatG (approved):
LGTM

avaitonis (approved):
LGTM

azure-functions/number-generator-function/api.js
```diff
@@ -20,47 +24,41 @@ const getAPI = async (type) => {
...
-      status,
-      statusText,
-      data,
-      requestConfig,
-      date: new Date().toISOString(),
-    },
-  }));
+
+  const response = await axios(request)
+    .catch((error) => error);
```
lmarrai-sw, line 49: I don't think this `catch` does anything, so you can remove it

lmarrai-sw (approved):
LGTM, I left a tiny comment about being able to remove an unnecessary `catch` that I just noticed but that can be a separate PR or we can re-approve that change ASAP

PR #1636, opened 2023-05-15
fix(portal-api): health check `/`

abhi-markan:
## Introduction
`Portal-api` healthcheck

## Resolution
* Added `/` express route to return `200` for Azure slot pre-swap healthcheck.

## Miscellaneous
* Added a eslint avoid rule.
* Updated npm services.
* Updated dependencies.


avaitonis (dismissed):
LGTM

avaitonis (approved):
LGTM 

GET / could use some standard approach

portal-api/src/createApp.js
```diff
@@ -32,6 +32,15 @@ app.use('/v1', openRouter);
 app.use('/v1', authRouterAllowXss);
 app.use('/v1', authRouter);
 
+// Return 200 on get to / to confirm to Azure that
+// the container has started successfully:
+const rootRouter = express.Router();
+rootRouter.get('/', async (req, res) => {
+  res.status(200).send();
+});
```
lmarrai-sw, line 40: You don't need the `async` here because you don't have `await`, unless you want to force the anonymous function to return a promise for some reason

lmarrai-sw (commented):
LGTM @abhi-markan, I left a minor style comment but it's up to you if you want to do anything about it

PR #1639, opened 2023-05-15
feat(DTFS-6424): add api input validation

ChristianMcc:
PR to add some input validation to TFM API. Note no sanitization has been added as the validation added covers not needing any sanitisation on the parameters affected. 

This should fix some of the security alerts for the TFM API that DTFS currently has.

Points of discussion:
- I have only added validation to API params that I could see getting passed through to API calls causing the github security alerts. I have not validated any of the bodies to the POST/PUT endpoints (happy to do that if we want to extend this)
- Some assumptions have been made on what is numeric/alphanumeric based on tests we have I've seen. Would be good for someone to double check `route-validators.js`. It's not overly strict just restricting to numeric or alphanumeric characters.
- Validation is done on a per route basis, leads to some repetition but easy to iterate on
- Error messages returned have all the information given from express-validator, AFAIK only the UIs call this API so the error messages will most likely be swallowed or need custom code in the frontends to be useful
- Haven't looked into validating GraphQL endpoints as they should be being deleted soon


dtfs-central-api/src/v1/controllers/tfm/amendments/tfm-get-amendments.controller.js
```diff
@@ -443,7 +443,7 @@ const findLatestCompletedAmendmentByFacilityIdVersion = async (facilityId) => {
 };
 exports.findLatestCompletedAmendmentByFacilityIdVersion = findLatestCompletedAmendmentByFacilityIdVersion;
 
-exports.getLatestCompletedValueAmendment = async (req, res) => {
```
ChristianMcc, line 446: I updated the names of a couple of functions here because it confused me and thought this was better - very trivial happy to undo if anyone wants me to

trade-finance-manager-api/src/v1/validation/route-validators/validation-handler.js
```diff
@@ -0,0 +1,18 @@
+const { validationResult } = require('express-validator');
+
+/**
+ * Validation middleware, used on a per route basis to handle when validations are run on the inputs to an API route.
+ * @param {Object} req Request object
+ * @param {Object} res Response object
+ * @param {String} next Callback function name
+ */
+// eslint-disable-next-line consistent-return
```
abhi-markan, line 9: Can we please remove inline rule and resolve the linting error?

abhi-markan (changes requested):

trade-finance-manager-api/src/v1/validation/route-validators/validation-handler.js
IgnatG: agreed
`return next();` should to the trick

trade-finance-manager-api/src/v1/validation/route-validators/route-validators.js
```diff
@@ -0,0 +1,35 @@
+const { param } = require('express-validator');
+
+const userParamNumericValidation = param('user').isAlphanumeric().withMessage('The User ID (user) provided should be numeric');
+const facilityIdNumericValidation = param('facilityId').isNumeric().withMessage('The Facility ID (facilityId) provided should be numeric');
```
IgnatG, line 4: isMongoId()

trade-finance-manager-api/src/v1/validation/route-validators/route-validators.js
```diff
@@ -0,0 +1,35 @@
...
+const amendmentIdNumericValidation = param('amendmentId').isNumeric().withMessage('The Amendment ID (amendmentId) provided should be numeric');
+const dealIdNumericValidation = param('dealId').isNumeric().withMessage('The Amendment ID (amendmentId) provided should be numeric');
```
IgnatG, line 6: we need to use `isMongoId()`

trade-finance-manager-api/src/v1/validation/route-validators/route-validators.js
```diff
@@ -0,0 +1,35 @@
...
+const partyURNNumericValidation = param('urn').isNumeric().withMessage('The party URN (urn) provided should be numeric');
```
IgnatG, line 7: the format for the `urn` will be: `"00122431"`. i.e. numbers, but in quotes/string.

trade-finance-manager-api/src/v1/validation/route-validators/route-validators.js
```diff
@@ -0,0 +1,35 @@
...
+
+exports.userEndpointValidation = [userParamNumericValidation];
+
```
IgnatG, line 10: some of these validations will repeat themselves (i.e. the facilityId will be validated on multiple endpoints).
I wonder if it's worth combining them into 1 statement, rather than having multiple statements. i.e. when it comes to validating the facilityId, we can use one statement:
```js
exports.facilityIdValidation = [facilityIdNumericValidation];
```

ChristianMcc: Yep wasn't sure on my preference between per endpoint or reducing repetition as much as possible - happy to refactor to above

trade-finance-manager-api/src/v1/validation/route-validators/route-validators.js
ChristianMcc: So does the frontend pass around the mongo Object IDs rather than say facility IDs that looks more like `0030605900` (with the first 3 digits depending on environment?)

trade-finance-manager-api/api-tests/v1/amendments/update-tfm-deal-amendment-completion.api-test.js
```diff
@@ -7,9 +7,9 @@ const MOCK_GEF_AIN_DEAL = require('../../../src/v1/__mocks__/mock-TFM-deal-AIN-s
 
 describe('update tfm-deals on amendment completion', () => {
   const mockAmendment = {
-    dealId: '123',
-    facilityId: '321',
-    amendmentId: '111',
+    dealId: '1234abcd1234abcd1234abcd',
```
IgnatG, line 10: is it worth using this -> https://observablehq.com/@hugodf/mongodb-objectid-generator to generate an id? to make sure it's valid :)

trade-finance-manager-api/src/v1/validation/route-validators/route-validators.js
```diff
@@ -0,0 +1,24 @@
+const { param } = require('express-validator');
+const { escape } = require('validator');
+
+const userParamEscapingSanitization = param('user').isString('User ID must be a string').escape();
```
ChristianMcc, line 4: As the get user endpoint can be done via username and the username doesn't have any validation added to it when it's created I've had to opt to just sanitize and escape the input rather than anything else

trade-finance-manager-api/src/v1/validation/route-validators/route-validators.js
```diff
@@ -0,0 +1,23 @@
+const { param } = require('express-validator');
+
+const userParamEscapingSanitization = param('user').isString('User ID must be a string').escape();
+const userParamMongoIdcValidation = param('user').isMongoId().withMessage('The User ID (user) provided should be a Mongo ID');
+const facilityIdMongoIdValidation = param('facilityId').isMongoId().withMessage('The Facility ID (facilityId) provided should be a Mongo ID');
```
IgnatG, line 5: we won't need the "mongoId" part. we can keep it as: `facilityIdValidation` - same for the other vars

IgnatG (approved):
LGTM - great work!

trade-finance-manager-api/src/v1/__mocks__/mock-TFM-deal-AIN-submitted.js
```diff
@@ -171,7 +171,7 @@ const MOCK_TFM_DEAL_AIN_SUBMITTED = {
     parties: {
       agent: { partyUrn: '', partyUrnRequired: false },
       buyer: { partyUrn: '', partyUrnRequired: false },
-      exporter: { partyUrn: '00307249', partyUrnRequired: true },
+      exporter: { partyUrn: '"00307249"', partyUrnRequired: true },
```
abhi-markan, line 174: Can this please be single quotes, to keep it consistent.

abhi-markan (changes requested):
All good bar one minor change to address please.

trade-finance-manager-api/src/v1/__mocks__/mock-TFM-deal-AIN-submitted.js
ChristianMcc: Ah good catch! Forgot to undo that one

IgnatG (approved):
LGTM

abhi-markan (approved):

PR #1640, opened 2023-05-15
fix: fixed the webpack build

JennyHumphries:
## Introduction
The webpack build was failing for `portal`, `trade-finance-manager-ui`, and `gef-ui`.

### Solution
* The failure was because the webpack version had been updated, which required a more recent version of `ajv`.
* I ran `npm i webpack` in each of the services to pick up the latest dependencies needed.
    * This generated the changes in the `package-lock` files
    * For `trade-finance-manager-ui` this was quite fiddly because there is a broken dependency between `apollo` and `graphql`. Which means you have to append `--legacy-peer-deps` to be able to run `npm i`. We seem to need `--legacy-peer-deps` in a few places. Is it worth raising a tech debt ticket to try and fix the dependencies, or one for this particular dependency? (Edit: Discussed on Slack, there are changes planned in TFM that will hopefully remove the need for using this flag, so there is no action needed for this PR)
* I then ran `npm run build` to verify that the webpack builds all compile.
    * This generated the changes in the `public` folders
* Updated `govuk-frontend` to `4.6.0`.

### Webpack Warnings
There are 36 warnings from webpack when I build portal. They're a combination of these two issues:
* `Deprecation $weight: Passing a number without unit % (20) is deprecated`
* `Deprecation Using / for division outside of calc() is deprecated and will be removed in Dart Sass 2.0.0.`
They are all in `node_modules\govuk-frontend\govuk\all.scss`
I get similar warnings when building `gef-ui` and `trade-finance-manager-ui`, again all in the govuk node modules.

## Miscellaneous
* Fixed various E2E tests with latest `govuk-frotend` error messages.
* Updated `cy.get` DOM path
* Updated `govuk-frontend` version
* Updated `es-lint` version
* Updated infrastructure GHA


JennyHumphries:
> Seems to be missing `package.json` files with latest version for `webpack` and `webpack-cli` deps?

The webpack version in `package.json` was already updated on `main` in this commit: 78f0d7bd00.
I've not done anything to the `webpack-cli` version. Would you like me to update that as well?

abhi-markan:
> > Seems to be missing `package.json` files with latest version for `webpack` and `webpack-cli` deps?
> 
> The webpack version in `package.json` was already updated on `main` in this commit: [78f0d7b](https://github.com/UK-Export-Finance/dtfs2/commit/78f0d7bd0035bc1e2829e2666f3914978a44b32b). I've not done anything to the `webpack-cli` version. Would you like me to update that as well?

Okay cool, I can see `main` already have all three UI latest version. Thanks

abhi-markan:
Few failing tests to address please.

JennyHumphries:
There are some linting errors when I run `npm run lint:all`. But they also appear to happen on main so I don't think are introduced in these changes.

I can only see the errors when I run that command locally, does the linter get run in the pipeline at all? I might be missing it in this PR.

The linter is also run in the pre-commit hook. It's very slow to run for me, which might just be a VM issue, so I generally turn it off when developing and manually run `npm run lint:all`. Do other people find the same problem? If no-one's really using it we could potentially remove it.

Also, since it's broken on main at the moment, it feels like the checks it runs might be inconsistent with the checks run in the pipeline. So we could either remove the lint check from the pre-commit hook, or change it to match whatever check happens in the pipeline. What do you think?

ChristianMcc:
@JennyHumphries don't worry about those lint:all failures, will fix in a new PR (I also spotted them during my last PR). I've done similar but I think it's a VM thing rather than an issue with the command itself - tempted to leave it for now and see how it performs when we get laptops and come back to it if it's an issue. Given it's tempting and easy to just turn off the pre-commit hook I agree might be worth duplicating the hook in the pipeline so we get to see it there too! (Don't think there is a linting step in the pipeline right now)

ChristianMcc:
Looks like some old copy is in the E2E tests, or new copy is in there and it's not updated properly somewhere else.

Abhi was the fix just upgrading govuk dependency and have they got anywhere in their readme/changelog that 4.5.0 isn't compatible with the current version of webpack? Seems like an annoying place to lose some time on!

reference-data-proxy/src/cronJobs/eStoreCronJobManager.ts
```diff
@@ -2,6 +2,7 @@ import CronJobManager from 'cron-job-manager';
 import { getCollection } from '../database';
 
 const cronJobTimer = new Date();
+
 if (process.env.NODE_ENV === 'dev') {
```
IgnatG, line 6: I think the value of NODE_ENV is `development` - based on docker-compose files

abhi-markan: Good spot.

docker-compose.yml
```diff
@@ -212,7 +212,7 @@ services:
       - REFERENCE_DATA_PROXY_URL=http://reference-data-proxy:5002
       - TZ=Europe/London
       - NODE_ENV=${NODE_ENV}
-    entrypoint: npx nodemon src/index.ts -L --max-old-space-size=350
```
JennyHumphries, line 215: Just want to double check this was meant to be included?

abhi-markan: Ah good shout, have reverted back.

e2e-tests/trade-finance-manager/cypress/e2e/journeys/facility/facilities.spec.js
```diff
@@ -55,7 +55,7 @@ context('Facility page', () => {
     });
   });
 
-  it('renders all facilities from TFM', () => {
+  it.only('renders all facilities from TFM', () => {
```
IgnatG, line 58: @abhi-markan  is this on purpose?

abhi-markan: Apologies, this has been removed.

IgnatG (approved):

abhi-markan (approved):

PR #1644, opened 2023-05-19
fix(currency-exchange-rate): Currency exchange rate calculation

abhi-markan:
## Introduction
Currency exchange for non GBP source currencies.

## Resolution
* Refactored currency inversion for non-GBP i.e. `USD -> GBP`
* Enhanced api tests.
* Improved documentation.

## Miscellaneous
Renamed PS title from source currency to GBP since they are calculated in GBP.

### DIFF
--- a/reference-data-proxy/api-tests/v1/currency-exchange-rate.api-test.ts
+++ b/reference-data-proxy/api-tests/v1/currency-exchange-rate.api-test.ts
@@ -7,18 +7,56 @@ const mockResponses = {
     status: 200,
     data: [
       {
-        id: 258671,
-        sourceCurrencyId: 37,
-        targetCurrencyId: 12,
-        currencyPair: 'USD-GBP X-RATE',
-        bidPrice: 0.729820464165815,
-        askPrice: 0.72976720426184,
-        lastPrice: 0.72976720426184,
-        midPrice: 0.729793833242109,
-        created: '2021-01-27T16:21:32',
-        updated: '2021-01-27T16:21:32',
-        effectiveTo: '9999-12-31T00:00:00',
-        effectiveFrom: '2021-01-27T00:00:00',
+        id: 344336,
+        sourceCurrencyId: 12,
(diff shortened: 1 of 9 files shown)


trade-finance-manager-ui/templates/case/facility/_macros/premium_schedule.njk
```diff
@@ -84,7 +84,7 @@
             <tr class="govuk-table__row">
               <th scope="col" class="govuk-table__header govuk-!-width-one-quarter govuk-table__header--numeric govuk-body-s" data-cy="facilities-table-heading-facility-id">Payment</th>
               <th scope="col" class="govuk-table__header govuk-!-width-one-quarter govuk-table__header--numeric govuk-body-s" data-cy="facilities-table-heading-cover-end-date">Due date</th>
-              <th scope="col" class="govuk-table__header govuk-!-width-one-quarter govuk-table__header--numeric govuk-body-s" data-cy="facilities-table-heading-value-gbp">Amount to be paid to UKEF ({{ currencySymbol }})</th>
+              <th scope="col" class="govuk-table__header govuk-!-width-one-quarter govuk-table__header--numeric govuk-body-s" data-cy="facilities-table-heading-value-gbp">Amount to be paid to UKEF (GBP)</th>
```
IgnatG, line 87: Are all currencies are in GBP here? I assume {{ currency symbol }} could return other currencies

reference-data-proxy/api-tests/v1/currency-exchange-rate.api-test.ts
IgnatG, line 1: Can we add a test for RON to GBP and vice versa? USD and euro were working previously by default. JPY is a new currency we introduced

abhi-markan: Have added some additional tests cases.

IgnatG (approved):
LGTM

PR #1648, opened 2023-05-23
fix(acbs): exposure period calculation

abhi-markan:
## Introduction
Exposure period calculation has not be correct especially for `BSS/EWCS` deals.

## Resolution
* Return `exposurePeriodInMonths` if already calculated by TFM.


IgnatG (approved):
LGTM

PR #1649, opened 2023-05-30
feat(DTFS2-6404): underwriter decision validation

JennyHumphries:
## Changes
* Increased the character limit for the underwriter manager's decision text areas from 1000 to 8000
* Remove the validation from these text areas that restricted it to specific characters
* Add XSS prevention sanitisation middleware to when the underwriter manager's decision is saved

## Testing
* I've verified that it is possible to save an 8000 character comment.
* I've verified that if you try to save a comment with `<script>` tags, they get stripped out before the value is saved.

## To-Do
* There's a failing static analysis check - All checks now fixed except the `cy is not defined` message which has been confirmed as safe to ignore.


trade-finance-manager-ui/server/controllers/helpers/validateSubmittedValues.helper.js
```diff
@@ -1,21 +1,15 @@
...
   let count = errorsCount;
 
-  if (!isAlphanumeric(value)) {
-    count = increment(count);
-
-    errors = generateValidationErrors(fieldId, `${fieldLabel} must only include letters a to z, numbers, hyphens, commas and spaces`, count, errors);
-  }
-
   // remove new lines from textarea input value
   const strippedValueArray = value.split(/[\r]/g);
```
JennyHumphries, line 14: Do you think I should remove this as well. 
It feels like something the user's would want, but I'm not sure why it was introduced originally. Perhaps we still need it for some reason?

trade-finance-manager-ui/server/routes/case/index.js
```diff
@@ -121,7 +122,7 @@ router.get('/:_id/underwriting/lead-underwriter/assign', underwritingController.
 router.post('/:_id/underwriting/lead-underwriter/assign', underwritingController.postAssignLeadUnderwriter);
 
 router.get('/:_id/underwriting/managers-decision/edit', underwritingController.getUnderwriterManagersDecisionEdit);
-router.post('/:_id/underwriting/managers-decision/edit', underwritingController.postUnderwriterManagersDecision);
+router.post('/:_id/underwriting/managers-decision/edit', xss(), underwritingController.postUnderwriterManagersDecision);
```
JennyHumphries, line 125: I've only added this sanitising to this specific endpoint rather than globally - I thought it's likely there would be unwanted side-effects if I used it everywhere.

trade-finance-manager-ui/server/controllers/helpers/validateSubmittedValues.helper.js
```diff
@@ -1,21 +1,15 @@
 const CONSTANTS = require('../../constants');
 
-const { hasValue, isAlphanumeric } = require('../../helpers/string');
+const { hasValue } = require('../../helpers/string');
 const { increment } = require('../../helpers/number');
 const generateValidationErrors = require('../../helpers/validation');
 
-const MAX_COMMENTS_LENGTH = 1000;
+const MAX_COMMENTS_LENGTH = 8000;
```
abhi-markan, line 7: Can this be saved and referred from micro-service constants please?

trade-finance-manager-ui/server/controllers/helpers/validateSubmittedValues.helper.js
abhi-markan: Happy for the above function (LOC:14) to be removed.

trade-finance-manager-ui/server/controllers/helpers/validateSubmittedValues.helper.test.js
```diff
@@ -5,41 +5,18 @@ import generateValidationErrors from '../../helpers/validation';
 
 describe('POST underwriting - managers decision - validate submitted values', () => {
   describe('validateCommentField', () => {
-    describe('when field is NOT alphanumeric', () => {
```
abhi-markan, line 8: Can we please add following test assertions:

1. Lorem ipsum with special characters previously not allowed.
2. An empty input with white spaces `   `
3. An input with various HTML tags

trade-finance-manager-ui/server/helpers/string.js
```diff
@@ -15,27 +15,8 @@ const hasValue = (str) => {
 const containsNumber = (str) =>
   /\d/.test(str);
 
-const isAlphanumeric = (str) => {
```
abhi-markan, line 18: Can we leave the function in please, might be useful in future.

trade-finance-manager-ui/server/helpers/string.test.js
```diff
@@ -42,33 +41,4 @@ describe('helpers - string', () => {
       expect(result).toEqual(false);
     });
   });
-
-  describe('isAlphanumeric', () => {
```
abhi-markan, line 46: Please revert, so the function can be used in future if and when required.

trade-finance-manager-ui/server/routes/case/index.js
abhi-markan: Can we check other `POST` endpoints where `xss()` might also be required please.

abhi-markan (changes requested):
Good work, can we also add/update Cypress testing with following assertions please:

1. Comments with characters over 8K.
2. Comments with XSS injection.
3. Comments with empty white spaces `  `

trade-finance-manager-ui/package.json
```diff
@@ -56,6 +56,7 @@
     "dotenv": "16.0.3",
     "express": "^4.18.2",
     "express-session": "1.17.3",
+    "express-xss-sanitizer": "^1.1.6",
```
JennyHumphries, line 59: This is a wrapper around `sanitize-html`: https://www.npmjs.com/package/sanitize-html?activeTab=readme

Which has a default list of allowed tags (in the Default Options) section of that link. @abhi-markan, do you know what tags if any we should allow for these inputs?

trade-finance-manager-ui/templates/case/underwriting/managers-decision/_macros/managers-decision-submitted.njk
```diff
@@ -45,7 +45,7 @@
 
             <h4 class="govuk-heading-m govuk-!-font-weight-bold" data-cy="comments-heading">{{commentsHeadingText }} </h4>
 
-            <p class="ukef-preserve-white-space" data-cy="conditions-value">{{ decision.comments | safe }}</p>
+            <p class="ukef-preserve-white-space" data-cy="conditions-value">{{ decision.comments }}</p>
```
JennyHumphries, line 48: We use `safe` in quite a lot of other places in DTFS. Is there anything in the backlog about removing them?

I get the impression it's only supposed to be used when you're absolutely sure that what you're displaying can't be malicious, which we can't be sure of when it's from a user input.

If there isn't already a ticket for it, I think we should raise one to remove all the instances of `safe`. What do you think?

trade-finance-manager-ui/package.json
abhi-markan: Hey @JennyHumphries no tags should be allowed. Thanks

trade-finance-manager-ui/templates/case/underwriting/managers-decision/_macros/managers-decision-submitted.njk
abhi-markan: `safe` should have only been used where we do not want to escape HTML tags, in which case ` | safe` NJK filter should be removed please.

trade-finance-manager-ui/server/routes/case/index.js
JennyHumphries: Please could we add a separate ticket to cover this and removing the other usages of the `| safe` filter? I'm happy to pick up that ticket next.

I think there'll be a lot of extra QA testing needed if we're changing lots of endpoints which I think will end up being quite confusing. Also it'll be quite time consuming going through all the endpoints to work out where this is needed so it would be good to not delay getting the rest of this ticket deployed.

Alternatively, we could add the xss filtering globally. It would take a lot less dev time than going through the endpoints individually but would need a lot of smoke testing. Again, if we go for this approach I think it should be as a separate ticket to make it easier to track.

Let me know what you think.

trade-finance-manager-ui/server/helpers/string.js
JennyHumphries: I'm a bit reluctant to leave this in.

It's quite an odd list of allowed characters that I don't think is likely to be reused.

The regex is also a little strange, it looks like some characters like `/\'` are only allowed at the start of a string but not at the end. And I think it looks like it allows all letters with the `A-Za-z` part, but then also explicitly allows `s` again later on. We'd probably be better off writing some new regex if we did need anything similar in future.

Finally it's not accurately named since it will return `true` for something that isn't alphanumeric like `.,-,.,`, which could cause confusion in future.

trade-finance-manager-ui/server/controllers/helpers/validateSubmittedValues.helper.test.js
JennyHumphries: What's the expected behaviour for the second case? I think previously whitespace was allowed.

It's not a required field, but it might be confusing if a user does save a whitespace response. So maybe it makes most sense to trim any leading/trailing whitespace, but not make a validation error for it. What do you think?

JennyHumphries: Oh actually, the validator doesn't change the value at all so it's not simple to trim any whitespace. I think if we do want to add that behaviour we should raise a separate ticket, but if not the behaviour is unchanged so it's probably not worth adding a test for it here.

For assertion 3, the HTML tags are stripped out by the middleware rather than this method so I won't add a test with them here.

trade-finance-manager-ui/server/routes/case/index.js
abhi-markan: I can create a separate ticket for `safe` filter removal.

trade-finance-manager-ui/server/controllers/helpers/validateSubmittedValues.helper.test.js
abhi-markan: 2. To ensure an empty input made of white spaces only is not saved and a validation error is thrown.
3. Understood, if we can cover this in an E2E test then that will suffice.

trade-finance-manager-ui/server/helpers/string.js
IgnatG: I'd suggest we remove it if it's not being used. We will forget about it and we could leave junk in the codebase.

trade-finance-manager-ui/server/routes/case/index.js
IgnatG: Let's create a separate ticket. Otherwise this PR can get quite big

trade-finance-manager-ui/templates/case/underwriting/managers-decision/_macros/managers-decision-submitted.njk
IgnatG: Agreed. We need to remove the safe pipe. A separate ticket will be needed for it

trade-finance-manager-ui/server/controllers/helpers/validateSubmittedValues.helper.test.js
JennyHumphries: Oh I see, I missed that there's already some validation to disallow whitespace. 

That validation doesn't happen in this file so I've not added a test here, but I've added a test for whitespace in `underwriter-managers-decision.spec.js`.

Same for the HTML tags, I've added a check for it in that file.

trade-finance-manager-ui/package.json
JennyHumphries: Great, thanks! The change to disallow all tags is in `middleware/xss-sanitizer.js`

e2e-tests/trade-finance-manager/cypress/e2e/journeys/case/underwriting/underwriter-managers-decision.spec.js
```diff
@@ -61,45 +61,146 @@ context('Case Underwriting - Underwriter Manager\'s decision', () => {
...
   });
 
-  it('after valid form submit, displays submitted values and updates deal stage', () => {
+  it('after valid form submit, cleans and displays submitted values and updates deal stage', () => {
     pages.underwritingPage.addUnderwriterManagerDecisionButton().click({ force: true });
 
-    const MOCK_COMMENTS = 'Testing';
-    const MOCK_INTERNAL_COMMENTS = 'Internal comment';
+    const MOCK_COMMENTS = 'Approval comment. <div>Div contents</div><script>Script contents</script>';
+    const MOCK_INTERNAL_COMMENTS = 'Internal comment. <div>Div contents</div><script>Script contents</script>';
```
abhi-markan, line 203: Minor adjustment to the `MOCK_COMMENTS` can we use this please to test the robustness of the `xss()`.

```
Approval comment. <div>Div contents</div><script>Script contents</script> &lt;img src = 'data:abc' /&gt;
```

abhi-markan (changes requested):
Good work, only few minor (repeating) comments to address please.

trade-finance-manager-ui/server/helpers/string.test.js
JennyHumphries: (See conversation in `../helpers/string.test.js`)

trade-finance-manager-ui/server/routes/case/index.test.js
```diff
@@ -1,3 +1,5 @@
+// eslint-disable-next-line import/no-extraneous-dependencies
```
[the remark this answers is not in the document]
JennyHumphries: I don't think we can, I get these errors if I do:
image

e2e-tests/trade-finance-manager/cypress/e2e/journeys/case/underwriting/underwriter-managers-decision.spec.js
JennyHumphries: I tested this out and the `&lt;img src = 'data:abc' /&gt` doesn't get stripped out.

Is it worth including it in this assertion to show that it doesn't get stripped out, or should I leave the test just showing what does get stripped out?

abhi-markan: Please leave it in the assertion and show that it doesn't get stripped out, I would expect it not to be stripped out and rendered as `<` or `>`

JennyHumphries: Will do. As confirmed on Slack I will make this change and merge this pull request.

abhi-markan (approved):

PR #1694, opened 2023-06-01
fix(lint): lint fixes

abhi-markan:
## Introduction
Linting errors

## Resolution
* Various lint fixes across all the micro-services.

## Miscellaneous
* Dependency updates


dtfs-central-api/api-tests/api.js
```diff
@@ -11,7 +12,9 @@ module.exports = (app) => ({
     to: async (url) => {
       const results = [];
 
+      // eslint-disable-next-line no-restricted-syntax
```
JennyHumphries, line 15: These two rules (Line 15 and 17) are disabled in some of the other services. Should we disable them for this service as well rather than opting out on individual lines? Or is it needed for the non-test files in `dtfs-central-api`?

dtfs-central-api/api-tests/wipeDB.js
```diff
@@ -1,7 +1,8 @@
 const db = require('../src/drivers/db-client');
 
 const wipe = async (collections) => {
-  const drop = async (collection) => new Promise(async (resolve, reject) => {
+  // eslint-disable-next-line no-async-promise-executor
+  const drop = async (collection) => new Promise(async (resolve) => {
```
JennyHumphries, line 5: I think we can fix this lint error rather than disabling it if we change it to:
```
  const drop = async (collection) => new Promise((resolve) => {
    db.getCollection(collection).then((collection) => {
      collection.drop(() => { resolve() });
    });
  });
```

portal-api/api-tests/wipeDB.js
```diff
@@ -1,10 +1,11 @@
 const db = require('../src/drivers/db-client');
 
 const wipe = async (collections) => {
-  const drop = async (collection) => new Promise(async (resolve, reject) => {
+  // eslint-disable-next-line no-async-promise-executor
```
JennyHumphries, line 4: I think this might be able to be fixed in the same way as the other `wipeDB` file

portal/.eslintrc.js
```diff
@@ -21,6 +21,7 @@ module.exports = {
     'object-curly-newline': ['error', {
       consistent: true,
     }],
+    'no-restricted-syntax': 0,
```
JennyHumphries, line 24: As in the other config file, please can we change `0` to `'off'`? And on line 18 and 19?

utils/data-migration/.eslintrc.js
```diff
@@ -19,6 +19,10 @@ module.exports = {
     'implicit-arrow-linebreak': 0,
     'comma-dangle': 'off',
     'no-loop-func': 'off',
+    'import/no-unresolved': 0,
+    'no-await-in-loop': 0,
+    'import/extensions': 0,
+    'no-restricted-syntax': 0,
```
JennyHumphries, line 25: As above, please can we make all the rules use `'off`' instead of 0?

utils/data-migration/.eslintrc.js
```diff
@@ -19,6 +19,10 @@ module.exports = {
     'implicit-arrow-linebreak': 0,
     'comma-dangle': 'off',
     'no-loop-func': 'off',
+    'import/no-unresolved': 0,
+    'no-await-in-loop': 0,
```
JennyHumphries, line 23: Now that these rules are disabled here, we should be able to remove all individual instances where it's disabled in this service. For example line 167 in `migrate-deals.js`.

I think there are quite a few places we can make similar changes.

JennyHumphries (changes requested):
Looks good, just a couple of small changes suggested.

dtfs-central-api/api-tests/api.js
abhi-markan: Have disabled across the micro-service and removed individual lines from across the files.

portal-api/.eslintrc.js
```diff
@@ -23,6 +23,8 @@ module.exports = {
     'object-curly-newline': ['error', {
       consistent: true,
     }],
+    'no-restricted-syntax': 0,
```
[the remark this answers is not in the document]
abhi-markan: Have replaced across all the micro-services.

dtfs-central-api/api-tests/wipeDB.js
abhi-markan: Good shout, have refactored.

portal-api/api-tests/wipeDB.js
abhi-markan: Have refactored,.

portal/.eslintrc.js
abhi-markan: Have updated.

utils/data-migration/.eslintrc.js
abhi-markan: Have updated.

utils/data-migration/.eslintrc.js
abhi-markan: Have removed individual instances.

abhi-markan:
> Looks good, just a couple of small changes suggested.

Thanks @JennyHumphries can you please re access the PR have addressed all your comments. Thanks

JennyHumphries (approved):
Looks good to me.

It looks like some of the cypress tests and code checks are failing, I'm not sure if that's something that normally needs to be fixed before merging or not.

IgnatG (approved):

PR #1696, opened 2023-06-06
feat(FN-70): bicep conversion - networking

OxonianCambion:
NOTE: this is to be merged to a feature branch. There will be further work required, but I need to break things down a little.
(See FN-419 to 430 here: https://ukef-dtfs.atlassian.net/browse/FN-12 but I'll need to add others too. e.g. refactoring.)

This PR enables a simple manual deploy of the networking infrastructure using Bicep.
Note that some code is commented out and will need including as other elements are added.
I've avoided commonising too early. e.g. There is duplication in Tags.

It does not cover automated deployment / linting etc.

I'm creating it as a draft, but I don't expect the final PR to differ too much. (Including having TODOs, commented out code and other suboptimal elements.)


infrastructure/arm_templates/gw-nsg/parameters.json
```diff
@@ -0,0 +1,5 @@
+{
```
OxonianCambion, line 1: Note that I'm preserving the ARM templates from the export for now. These are used to generate the Bicep files, though I am not committing the decompiled bicep files that I am using as inspiration for mine.

I don't think it is particularly worth reading through the ARM files Jump to `infrastructure/resource_group_level/main.bicep.`

infrastructure/resource_group_level/main.bicep
```diff
@@ -0,0 +1,138 @@
...
+@minLength(5)
+@maxLength(50)
+@description('Provide a globally unique name of your Azure Container Registry')
+param acrName string = 'tfsfeatureacr${uniqueString(resourceGroup().id)}'
+@description('Provide a tier of your Azure Container Registry.')
+// Dev uses 'Standard'
+param acrSku string = 'Basic'
+
+
+// Dev uses 172.16.4x.xx
```
OxonianCambion, line 17: I'll turn these into parameter sets, if appropriate later.

infrastructure/resource_group_level/main.bicep
```diff
@@ -0,0 +1,138 @@
...
+// Demo (legacy?) uses 172.16.6x.xx
```
OxonianCambion, line 18: Can I kill everything to do with Demo?
Not that Test & Staging seem to be overlapping too. Is this expected? @IgnatG

infrastructure/resource_group_level/modules/nat-gw_nat_ip.bicep
```diff
@@ -0,0 +1,56 @@
+param location string
+param appServicePlanName string
+
+var natGatewayName = 'tfs-${appServicePlanName}-nat-gw'
+var natGatewayIpAddressesName = 'tfs-${appServicePlanName}-nat-ip'
+
+// TODO:DTFS-6422 It looks like this is duplicated by the natGatewayIpAddresses below. Remove if all is ok.
```
OxonianCambion, line 7: It does seem to work, but I'll check a bit more carefully before wiping this.
(It seems the bicep decompile will wire up dependencies from both sides - this duplication needs to be fixed by removing one wiring.)

infrastructure/resource_group_level/modules/privatelink-azurewebsites-net.bicep
```diff
@@ -0,0 +1,521 @@
...
+  }
+  properties: {
+    registrationEnabled: false
+    virtualNetwork: {
+      id: vnetId
+    }
+  }
+}
+
+// TODO:DTFS-6422 Wire up A record IPs correctly, getting appropriate values.
```
OxonianCambion, line 44: I need to understand how the IPs are allocated before defining them. I certainly can't just use the values already there as they are from Dev (and "Demo"). I suspect I'll uncomment (or even remove if they are more appropriately created elsewhere) IPs as I add other modules.

infrastructure/resource_group_level/modules/route-tables.bicep
```diff
@@ -0,0 +1,41 @@
...
+param mulesoftSubnetCidr string
+param productionSubnetCidr string
+
+// TODO:DTFS-6422 consider resource naming convention
+var routeTableName = '${resourceGroup().name}-UDR'
+
+resource routeTable 'Microsoft.Network/routeTables@2022-11-01' = {
+  name: routeTableName
+  location: location
+  tags: {
```
OxonianCambion, line 12: I'll commmonise tags at the end.

infrastructure/subscription_level/main.bicep
```diff
@@ -0,0 +1,14 @@
+param location string = 'uksouth'
+param resourceGroupName string = 'Digital-Feature'
+
+
+// I probably don't need a resource group module and can probably do everything here...
```
OxonianCambion, line 5: Note that the `infrastrcuture.yml` infrastructure can also be done in Bicep. (I.e. the stuff that isn't scoped to the Resource Group.) Note that these are scoped at the subscription level.
I've kept it separated for now, but it may not be necessary.

infrastructure/subscription_level/main.bicep
```diff
@@ -0,0 +1,14 @@
+param location string = 'uksouth'
+param resourceGroupName string = 'Digital-Feature'
```
IgnatG, line 2: can we name it `rg-dtfs-dev-001`? something like that? otherwise we have to rename it again later on

OxonianCambion: My current plan is to replicate the current environments as accurately as possible, including naming, then update them when I'm confident they correspond. (Perhaps I'm just making more work for myself though!)

Obviously I'm not going to get confused with the resource group though!

Note that once deployed, we can't rename things! I'm not sure if it is worth redeploying resources in dev/test/staging/prod just to have consistent naming.

Some bloke here has found a bit of a hacky workaround: https://azureis.fun/posts/Adding-a-friendly-name-to-Azure-Resources/

IgnatG: We need to follow Azure's best practices in terms of naming conventions -> [https://learn.microsoft.com/en-us/azure/cloud-adoption-framework/ready/azure-best-practices/resource-naming](https://learn.microsoft.com/en-us/azure/cloud-adoption-framework/ready/azure-best-practices/resource-naming ).
So yes, we need to rename all resources

OxonianCambion:
Note that the static analysis isn't being run because the target branch isn't set to have it applied.
I'm happy to not run static analysis for now, while the code may be a little ropy...

https://app.codacy.com/gh/UK-Export-Finance/dtfs2/pullRequest?prid=11930169

infrastructure/subscription_level/main.bicep
OxonianCambion: We need to form a plan for renaming of production resources - Azure's weird decision to make name and id connected means a rename is a delete and create...
I'll look into that at the appropriate time.

If it's ok, I'll stick with my plan to follow best naming practice after I've produced a working deployment. (i.e. in a separate card near the end.)

IgnatG: Ok cool

OxonianCambion: I've raised https://ukef-dtfs.atlassian.net/browse/FN-462 to make this more explicit.

infrastructure/resource_group_level/modules/vnet.bicep
```diff
@@ -0,0 +1,218 @@
...
+            }
+          ]
+          delegations: []
+          privateEndpointNetworkPolicies: 'Disabled'
+          privateLinkServiceNetworkPolicies: 'Enabled'
+        }
+        type: 'Microsoft.Network/virtualNetworks/subnets'
+      }
+      // TODO:DTFS-6422 Work out what these demo resources are for.
+      // Gabriel says it is a very old legacy item.
```
IgnatG, line 100: yup, it needs to go. there was a 4th environment called `demo` up until 2022. That got decommissioned. We don't need to replicate it now so we should remove anything related to it

infrastructure/resource_group_level/modules/vnet.bicep
```diff
@@ -0,0 +1,218 @@
...
+param privateEndpointsCidr string
+param appServicePlanEgressPrefixCidr string
+param applicationGatewayCidr string
+param vmCidr string
+
+param appServicePlanName string
+
+param routeTableId string
+param networkSecurityGroupId string
+// TODO:DTFS-6422 Note that none of these parameters/values seem to appear in the old IaC scripts
```
IgnatG, line 13: we can remove this now - no need to address it in a different PR

PR #1697, opened 2023-06-08
feat: destructure feedback form on TFM API

ChristianMcc:
The feedback controller on the TFM amendments API was the only endpoint that was vulnerable of saving unwanted form fields from the forms in the frontend so I've moved the object destructuring that already existed to before the DB save call so that unwanted fields are filtering out from the request body.


IgnatG (approved):
LGTM

PR #1707, opened 2023-06-14
feat(DTFS2-6505): Fix vulnerable forms/endpoints on GEF UI

ChristianMcc:
## Introduction
Sanitise user input across form submissions in GEF UI that makes requests to the portal-api micro-service.

## Resolution
Added the constructPayload function from portal to GEF including tests. Used this as well as adding to already created custom payloads when sending data to the portal API from the body of requests from submitting forms in GEF


gef-ui/server/utils/constructPayload.test.js
```diff
@@ -0,0 +1,94 @@
...
+    const returned = constructPayload(mockBody, payloadProperties);
+
+    expect(expected).toEqual(returned);
+  });
+
+  it('Should return expected payload as per `properties` and without CSRF argument with data in standard body', () => {
+    const expected = {
+      currentPassword: 'AbC!2345',
+      password: 'AbC!23456',
+      passwordConfirm: 'AbC!23456',
```
abhi-markan, line 88: Can we add an assertion with following:

```javascript
 const expected = {
      currentPassword: 'AbC!2345',
      password: 'AbC!23456',
      passwordConfirm: 'AbC!23456',
      roles: {
      'admin',
      'maker',
      'checker'
      }
};
```

portal/server/helpers/constructPayload.js
```diff
@@ -25,13 +25,11 @@ const constructPayload = (body, properties, csrf = true) => {
   // Property insertion
   properties
     .filter((property) => property in body)
-    .map((property) => {
+    .forEach((property) => {
```
abhi-markan, line 28: Is `forEach` expensive on execution isn't?

abhi-markan (changes requested):

portal/server/helpers/constructPayload.js
IgnatG: happy for us to keep `.map`

ChristianMcc: From what I've read foreach is a bit faster than map in this kind of use case but happy to revert

gef-ui/server/utils/constructPayload.test.js
ChristianMcc: Sorry not quite understanding this one - do yo want an extra `payloadPropertiesWithRoles` array and to test that that now pulls through the roles from the mockExtraBody?

abhi-markan (approved):

PR #1753, opened 2023-06-15
feat(fn-418): redis

OxonianCambion:
## Introduction

*   produce a complete new deployment environment (feature) mirroring the current ones.
*   be in a position to take over the extant environments.

Note that there have been some manual changes to the environments. This work also functions as an audit of the current infrastructure.

This card covers wiring up Redis only. Note that there will be some tidying / refactoring when all the components are in place, (e.g. commonise tags, consider parameter sets), but the code should be of reasonable quality and clear.

Note also that this replicates the current setup. Enhancements, including adding a private link are covered in FN-504 

## Resolution

This code wires up Redis

### DIFF
--- a/infrastructure/arm_templates/redis/parameters.json
+++ b/infrastructure/arm_templates/redis/parameters.json
@@ -1,0 +1,9 @@
+{
+    "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentParameters.json#",
+    "contentVersion": "1.0.0.0",
+    "parameters": {
+        "Redis_tfs_dev_redis_name": {
+            "value": null
+        }
+    }
+}
--- a/infrastructure/arm_templates/redis/template.json
+++ b/infrastructure/arm_templates/redis/template.json
@@ -1,0 +1,73 @@
+{
+    "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#",
+    "contentVersion": "1.0.0.0",
+    "parameters": {
+        "Redis_tfs_dev_redis_name": {
+            "defaultValue": "tfs-dev-redis",
+            "type": "String"
+        }
+    },
+    "variables": {},
+    "resources": [
+        {
+            "type": "Microsoft.Cache/Redis",
+            "apiVersion": "2023-05-01-preview",
+            "name": "[parameters('Redis_tfs_dev_redis_name')]",
+            "location": "UK South",
+            "tags": {
+                "Environment": "Preproduction"
+            },
+            "properties": {
+                "redisVersion": "6.0",
+                "sku": {
+                    "name": "Basic",
+                    "family": "C",
+                    "capacity": 0
+                },
+                "enableNonSslPort": false,
+                "minimumTlsVersion": "1.2",
+                "publicNetworkAccess": "Enabled",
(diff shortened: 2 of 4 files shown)


infrastructure/arm_templates/redis/parameters.json
```diff
@@ -0,0 +1,9 @@
+{
```
OxonianCambion, line 1: Skip over the arm templates.

infrastructure/arm_templates/redis/template.json
```diff
@@ -0,0 +1,73 @@
...
+                "redisConfiguration": {
+                    "maxmemory-reserved": "0",
+                    "maxfragmentationmemory-reserved": "0",
+                    "maxmemory-policy": "volatile-lru",
+                    "maxmemory-delta": "0"
+                }
+            }
+        },
+        {
+            "type": "Microsoft.Cache/Redis/accessPolicies",
```
OxonianCambion, line 40: I'm not sure why it exports the three standard access policies. They aren't in the CLI instructions and I haven't tried to include them in my code.

infrastructure/resource_group_level/modules/redis.bicep
```diff
@@ -0,0 +1,42 @@
+param location string
+param environment string
+
+var redisName = 'tfs-${environment}-redis'
+
+// See https://learn.microsoft.com/en-us/azure/azure-cache-for-redis/cache-redis-cache-bicep-provision?tabs=CLI
```
OxonianCambion, line 6: I could add some of the extras in their example. but it doesn't seem especially necessary.

infrastructure/resource_group_level/modules/redis.bicep
```diff
@@ -0,0 +1,42 @@
...
+// for an example Bicep file
+
+// TODO:FN-504 Configure private link.
+// See line 718 in developmentinfrastructure.yml
+// and https://docs.microsoft.com/en-us/azure/azure-cache-for-redis/cache-private-link
+
+// TODO:FN-504 decide what sku to use.
```
IgnatG, line 13: this will be based on environment - prod can haver a higher SKU than dev/test

infrastructure/resource_group_level/modules/redis.bicep
```diff
@@ -0,0 +1,42 @@
...
+  name: redisName
+  location: location
+  tags: {
+    Environment: 'Preproduction'
+  }
+  properties: {
+    redisVersion: '6.0'
+    sku: sku
+    enableNonSslPort: false
+    minimumTlsVersion: '1.2'
```
IgnatG, line 32: small comment here: can we put some of these values in a constant/variable at the top? it should be easier to change from there if we need to in the future

OxonianCambion: As I've just added above - I'll make `sku` a parameter. I was going to just make it an object, but I note that the example splits them so it can put validation on them. If you prefer that approach, say and I'll use it. https://learn.microsoft.com/en-us/azure/azure-cache-for-redis/cache-redis-cache-bicep-provision?tabs=CLI

I don't plan on adding any of the other parameters here as I don't think they will be changed.

IgnatG (approved):

PR #1755, opened 2023-06-16
Fix(DTFS2): Add interface to premium Schedule

ChristianMcc:
## Introduction
The get premium schedule endpoint posted the request body without any sanitization meaning it was vulnerable to having extra fields than what it should accept in request body and sending all of those on.

## Resolution
Introduce new interface to cast the request body to so that extra fields aren't passed through.

### DIFF
--- a/external-api/src/interfaces/premiumSchedule.interface.ts
+++ b/external-api/src/interfaces/premiumSchedule.interface.ts
@@ -1,14 +1,14 @@
 export interface PremiumSchedule {
-  premiumTypeId: string;
-  premiumFrequencyId: string;
+  premiumTypeId: number;
+  premiumFrequencyId: number;
   productGroup: string;
-  facilityURN: string;
+  facilityURN: number;
   guaranteeCommencementDate: string;
   guaranteeExpiryDate: string;
-  guaranteeFeePercentage: string;
-  guaranteePercentage: string;
+  guaranteeFeePercentage: number;
+  guaranteePercentage: number;
   dayBasis: string;
-  exposurePeriod: string;
+  exposurePeriod: number;
   maximumLiability: number;
   cumulativeAmount: number;
 }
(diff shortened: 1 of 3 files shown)


external-api/src/interfaces/premiumSchedule.interface.ts
```diff
@@ -0,0 +1,14 @@
+export interface PremiumSchedule {
+  premiumTypeId: string;
+  premiumFrequencyId: string;
+  productGroup: string;
+  facilityURN: string;
+  guaranteeCommencementDate: string;
+  guaranteeExpiryDate: string;
+  guaranteeFeePercentage: string;
+  guaranteePercentage: string;
```
abhi-markan, line 9: This should be a `number` `guaranteePercentage`

external-api/src/interfaces/premiumSchedule.interface.ts
```diff
@@ -0,0 +1,14 @@
+export interface PremiumSchedule {
```
ChristianMcc, line 1: From what I read in the other file all of these are always provided and not nullable, but if it's safer I'm happy to make these all optional parameters in the interface

external-api/src/interfaces/premiumSchedule.interface.ts
```diff
@@ -0,0 +1,14 @@
...
+  premiumTypeId: string;
+  premiumFrequencyId: string;
+  productGroup: string;
+  facilityURN: string;
+  guaranteeCommencementDate: string;
+  guaranteeExpiryDate: string;
+  guaranteeFeePercentage: string;
+  guaranteePercentage: string;
+  dayBasis: string;
+  exposurePeriod: string;
```
abhi-markan, line 11: `exposurePeriod` as `number` please.

abhi-markan (changes requested):
Looks good, few comments please.

abhi-markan (approved):

PR #1758, opened 2023-06-22
feat(DTFS2-6499): Make amendment button (Portal)

JennyHumphries:
## Introduction
Adding links to the start amendment journey in Portal

## Resolution
* Added "Amend" links to each row in the Bond and Loan tables
* Links only appear when:
    * Deal is acknowledged
    * User is a maker
    * Facility has been Issued
* Clicking the link takes you to the "What do you need to change page" which is currently blank and will be filled in as part of future tickets.

## Other Notes
* The Codacy check is failing, but I think the errors are safe to ignore again. They are things like `cy` not being defined.


e2e-tests/portal/cypress/e2e/journeys/maker/amend-facility.spec.js
```diff
@@ -0,0 +1,139 @@
...
+                  setLoanCallback(createdFacilities.filter((f) => f.type === 'Loan')[0]);
+
+                  facilitiesToCleanUp.push(...createdFacilities);
+                });
+            });
+        });
+    };
+
+    before(() => {
+      cy.deleteDeals(ADMIN);
```
JennyHumphries, line 59: This seems to be done in the other portal cypress tests. It means that your local data gets cleaned out when you run these tests locally.

I'm not sure why it's used, and why other tests clear up the facilities after running but not the deals.

portal/server/routes/contract/dealWithFacilityFlags.js
```diff
@@ -0,0 +1,30 @@
+const canAmendFacility = require('./canAmendFacility');
```
JennyHumphries, line 1: This file is renamed from `dealWithCanIssueOrEditIssueFacilityFlags.js`,

I've also changed lines 10-25

portal/templates/contract/components/bond-transactions-table.njk
```diff
@@ -73,10 +73,17 @@
 
                   {% set requestedCoverDateCanEdit = userIsMaker and bond.status !== "Maker's input required" and dealStatusAllowsRequestedCoverDateChange and dealSubmissionType === 'Manual Inclusion Application' and bond.facilityStage === 'Issued' and bond.issueFacilityDetailsSubmitted %}
 
-                  {% if requestedCoverDateCanEdit %}
-                    {{ requestedCoverStartDateLink.render(componentData) }}
-
-                    {% else %}
+                  {% if requestedCoverDateCanEdit or bond.canAmendFacility %}
```
JennyHumphries, line 76: I feel like there should be a better way of doing this, I've ended up with a lot of ifs here which is a bit hard to follow.

I'm also not sure about the `<span>|</span>`. Is there a better way of adding the separator between links?

portal/templates/contract/contract-view.njk
```diff
@@ -68,118 +68,110 @@
...
-          <h2 class="govuk-heading-m govuk-!-margin-bottom-0">
-            <span class="govuk-grid-column-one-half govuk-!-padding-left-0">
-                  Loan transactions
-                </span>
-
-            <span class="govuk-!-margin-left-6">
-              {{ statusTag.render(deal.loanTransactions.status) }}
-            </span>
+      <div class="govuk-grid-column-three-quarters">
+        <h2 class="govuk-grid-row govuk-heading-m govuk-!-margin-bottom-0">
```
JennyHumphries, line 133: Added class `govuk-grid-row` so that the two spans are a subgrid of this class.

portal/templates/contract/contract-view.njk
```diff
@@ -68,118 +68,110 @@
...
+        {{ bondTransactionsTable.render(componentData) }}
+      {% endif %}
+
+    </div>
 
-          </h2>
-        </div>
+    <div class="govuk-grid-row govuk-!-padding-top-2 govuk-!-padding-bottom-2 box">
+      <div class="govuk-grid-column-three-quarters">
+        <h2 class="govuk-grid-row govuk-heading-m govuk-!-margin-bottom-0">
```
JennyHumphries, line 155: Added class `govuk-grid-row` so that the two spans are a subgrid inside this header.

portal/templates/contract/contract-view.njk
```diff
@@ -68,118 +68,110 @@
...
-            <span class="govuk-grid-column-one-half govuk-!-padding-left-0">
-                  Loan transactions
-                </span>
-
-            <span class="govuk-!-margin-left-6">
-              {{ statusTag.render(deal.loanTransactions.status) }}
-            </span>
+      <div class="govuk-grid-column-three-quarters">
+        <h2 class="govuk-grid-row govuk-heading-m govuk-!-margin-bottom-0">
+          <span class="govuk-grid-column-one-half">
```
JennyHumphries, line 134: Removed `govuk-!-padding-left-0`

portal/templates/contract/contract-view.njk
```diff
@@ -68,118 +68,110 @@
...
+      {% endif %}
+
+    </div>
 
-          </h2>
-        </div>
+    <div class="govuk-grid-row govuk-!-padding-top-2 govuk-!-padding-bottom-2 box">
+      <div class="govuk-grid-column-three-quarters">
+        <h2 class="govuk-grid-row govuk-heading-m govuk-!-margin-bottom-0">
+          <span class="govuk-grid-column-one-half">
```
JennyHumphries, line 156: Removed `govuk-!-padding-left-0`

portal/templates/contract/contract-view.njk
```diff
@@ -68,118 +68,110 @@
...
-
-            <span class="govuk-!-margin-left-6">
-              {{ statusTag.render(deal.loanTransactions.status) }}
-            </span>
+      <div class="govuk-grid-column-three-quarters">
+        <h2 class="govuk-grid-row govuk-heading-m govuk-!-margin-bottom-0">
+          <span class="govuk-grid-column-one-half">
+            Bond transactions
+          </span>
+          <span class="govuk-grid-column-one-half">
```
JennyHumphries, line 137: Replaced `govuk-!-margin-left-6` with `govuk-grid-column-one-half`

portal/templates/contract/contract-view.njk
```diff
@@ -68,118 +68,110 @@
...
+      <div class="govuk-grid-column-three-quarters">
+        <h2 class="govuk-grid-row govuk-heading-m govuk-!-margin-bottom-0">
+          <span class="govuk-grid-column-one-half">
+            Loan transactions
+          </span>
 
-        <div class="govuk-grid-column-one-quarter box__header-link">
-          {{ addALoanButton.render(componentData) }}
-        </div>
+          <span class="govuk-grid-column-one-half">
```
JennyHumphries, line 160: Replaced `govuk-!-margin-left-6` with `govuk-grid-column-one-half`

portal/templates/contract/components/bond-transactions-table.njk
JennyHumphries: I think this logic might be wrong, I was thinking that the `issueOrDeleteFacilityLink` couldn't appear at the same time as the amend link, but actually I think it might be able to, so it shouldn't be an `else`

e2e-tests/portal/cypress/e2e/journeys/maker/amend-facility.spec.js
```diff
@@ -0,0 +1,139 @@
+const MOCK_USERS = require('../../../fixtures/users');
```
ChristianMcc, line 1: At some point we should think about the case of not showing amendment buttons when an amendment has started, although I imagine that first happens when we start to save data (on the next screen) and change the state of the facility. Not something to test here but need to remember it - is it worth adding a TODO with a ticket number I'll find in a bit to make sure we add tests for when the buttons shouldn't show? I guess same for GEF!

e2e-tests/portal/cypress/fixtures/mockFacilities.js
```diff
@@ -0,0 +1,36 @@
...
+  ukefGuaranteeInMonths: '24',
+  name: '1234',
+  bondBeneficiary: 'test',
+  value: '123',
+  currencySameAsSupplyContractCurrency: 'true',
+  riskMarginFee: '1',
+  coveredPercentage: '2',
+  feeType: 'test',
+  feeFrequency: 'Quarterly',
+  dayCountBasis: 'test',
```
ChristianMcc, line 15: Are these just copy pasted from somewhere else? Wonder if we should put some legit values in here like 360 and feeType as `In arrear'

portal/server/routes/contract/canAmendFacility.js
```diff
@@ -0,0 +1,8 @@
+const { FACILITY_STAGE, STATUS, USER_ROLES } = require('../../constants');
+
+const canAmendFacility = (userRoles, deal, facility) =>
```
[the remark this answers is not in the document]
ChristianMcc: I think I'm leaning towards what you've said here, it's a bit late if they're already seeing this button if they aren't from the right bank. I'm a bit curious about e.g. the 'what do you need to change' page. Is our only defence that the ID of the facility is hard to guess?
How do we defend against it on other pages you've looked at? Do we at the start of the controller check the users bank + the facility's bank is the same?

ChristianMcc (commented):
Code's looking good - could you read through some of my comments and just give your thoughts please! Especially the viewing of other bank's facilities one

abhi-markan:
Closing for now, to be re-opened once portal amendment work has been re-scheduled.

PR #1766, opened 2023-06-26
fix(DTFS2): Fix get all amendments endpoint

ChristianMcc:
## Introduction
I found a bug when loading the homepage in TFM where getting all amendments in progress was failing.

## Resolution
Change req.query -> req.params as the status is in path not query


IgnatG (approved):
LGTM

abhi-markan (approved):

PR #1768, opened 2023-06-27
fix(SR-3): External API - API key middleware

ttbarnes:
## Introduction ✏️ 
This PR adds API key checks to all external-api endpoints.

## Resolution ✔️ 
* Create `checkApiKey` middleware. This checks if a provided key matches an environment variable and if not, the request is rejected.
* Update `supertest` constructor used in the API tests, so set a valid `x-api-key` header.
* Code refactorization.
* Update docker/GHA environment variables and .env samples.
* Update all instances of external API calls in portal and TFM API to send `x-api-key`.

## Miscellaneous 🗒️ 
* Fixed linting in README file.
* Updated all outdated dependencies except for mongodb.
* Fixed `npm start` command.
* Added CORS middleware.
* Update security headers middleware to include `no-store` `Cache-Control`, added missing test coverage.
* Created test mocks for express `req`, `res` and `next`.
* Fixed a unit test that was failing due to hard coded timestamps (estore controller).
* Fixed a linting issue with estore controller.
* Increase portal-api jest timeout due to flaky tests.


external-api/src/middleware/check-api-key/index.ts
```diff
@@ -0,0 +1,40 @@
...
+  const {
+    headers: { 'x-api-key': xApiKey },
+  } = req;
+
+  /**
+   * x-api-key is not provided.
+   * Reject the request.
+   */
+  if (!xApiKey) {
+    return res.status(400).send();
```
Zainzzkk, line 24: should it still not be 401 unauthorized since not provided?

ttbarnes: We discussed in slack, I think this makes sense because it's genuinely a bad request

IgnatG: Just jumping here, it needs to be 401. 400 is a bad request that will be applicable if the body is invalid. 
We need to keep it consistent with everything else

ttbarnes: Updated 👍

external-api/src/middleware/check-api-key/index.ts
```diff
@@ -0,0 +1,40 @@
...
+  const {
+    headers: { 'x-api-key': xApiKey },
+  } = req;
+
+  /**
+   * x-api-key is not provided.
+   * Reject the request.
+   */
+  if (!xApiKey) {
+    return res.status(401).send();
```
abhi-markan, line 24: Can we change to `400` please.

abhi-markan: Can we return a string message please with text `Bad request` please.

external-api/src/middleware/check-api-key/index.ts
```diff
@@ -0,0 +1,40 @@
...
+   */
+  if (xApiKey === API_KEY) {
+    return next();
+  }
+
+  /**
+   * x-api-key is invalid.
+   * Reject the reuqest
+   */
+  return res.status(401).send();
```
abhi-markan, line 39: Can we return a string message please stating `Unauthorised`

external-api/src/v1/controllers/estore/eStore.controller.ts
```diff
@@ -21,17 +21,8 @@ const validateEstoreInput = (eStoreData: any) => {
...
-    siteName,
-    facilityIdentifiers,
-    supportingInformation,
-    exporterName,
-    buyerName,
-    dealIdentifier,
-    destinationMarket,
-    riskMarket
-  } = req.body;
+  const { dealId, siteName, facilityIdentifiers, supportingInformation, exporterName, buyerName, dealIdentifier, destinationMarket, riskMarket } = req.body;
```
abhi-markan, line 24: Multiple lines please, more than three.

external-api/src/middleware/check-api-key/index.ts
ttbarnes: It was originally `400` as per conversations, but was asked to change to `401`: https://github.com/UK-Export-Finance/dtfs2/pull/1768#discussion_r1244095254

Have added messaging.

abhi-markan (dismissed):

abhi-markan (approved):

PR #1767, opened 2023-06-26
feat(apim): mulesoft to apim tfs migration

abhi-markan:
## Introduction ✏️ 
DTFS Mulesoft to APIM TFS migration

## Resolution ✔️ 
* Removal of `MULESOFT_**` variables.
* Introduced `APIM_TFS_**` env variables.
* Code refactorization.
* SSRF mitigation.
* Fixed various failing `api-tests`.

## Miscellaneous 🗒️ 
* Dependencies updates.
* `.env.sample` cleanup.
* Added `acbs` and `number-generator` api-tests.

## Notes ⚠️ 
Please note this code has not been tested with ACBS.

PR #1769, opened 2023-06-27
feat(FN-419): Add ACBS function to Bicep

OxonianCambion:
### Introduction

This card covers:

- Creating the Azure Function for the ACBS function, wired up to the correct docker image from CI

### Resolution

The code:
- creates the Azure Function
- wired up with dummy environment variables
- with Application Insights
Note:
- that there isn't currently an image to deploy!
- It does not wire up the A Records for the private link yet. I was planning on just adding them all in a batch for all resources. @IgnatG I'm happy to just do it per-resource / PR if you prefer.
- I don't currently send in the environment variables to the module. This will need fixing - either now or when @IgnatG adds the secrets themselves.

### DIFF
--- a/infrastructure/resource_group_level/main.bicep
+++ b/infrastructure/resource_group_level/main.bicep
@@ -1,4 +1,5 @@
 param location string  = resourceGroup().location
+// Expected values are 'feature', 'dev', 'test', 'staging' & 'prod'
 param environment string = 'feature'
 
 param appServicePlanName string = 'feature'
@@ -8,7 +9,8 @@ param appServicePlanKind string = 'linux'
 @minLength(5)
 @maxLength(50)
 @description('Provide a globally unique name of your Azure Container Registry')
-param acrName string = 'tfsfeatureacr${uniqueString(resourceGroup().id)}'
+// Note that the existing environments have: tfsdev, tfsstaging & tfsproduction
+param containerRegistryName string = 'tfsfeatureacr${uniqueString(resourceGroup().id)}'
 @description('Provide a tier of your Azure Container Registry.')
 // Dev uses 'Standard'
 param acrSku string = 'Basic'
@@ -72,10 +74,15 @@ resource appServicePlan 'Microsoft.Web/serverfarms@2022-09-01' = {
     name: appServicePlanSku
   }
   kind: appServicePlanKind
+  properties: {
+    // Linux ASPs need to have reserved as true:
+    // https://learn.microsoft.com/en-us/azure/templates/microsoft.web/serverfarms?pivots=deployment-language-bicep#appserviceplanproperties
+    reserved: appServicePlanKind == 'linux'
+  }
 }
 
 resource containerRegistry 'Microsoft.ContainerRegistry/registries@2023-01-01-preview' = {
-  name: acrName
+  name: containerRegistryName
   location: location
   sku: {
     name: acrSku
@@ -191,5 +198,18 @@ module redis 'modules/redis.bicep' = {
   params: {
     location: location
     environment: environment
+  }
+}
+
+module functionAcbs 'modules/function-acbs.bicep' = {
+  name: 'functionAcbs'
+  params: {
+    environment: environment
+    location: location
+    containerRegistryName: containerRegistryName
+    appServicePlanEgressSubnetId: vnet.outputs.appServicePlanEgressSubnetId
+    appServicePlanId: appServicePlan.id
+    privateEndpointsSubnetId: vnet.outputs.privateEndpointsSubnetId
+    storageAccountName: storage.outputs.storageAccountName
   }
 }
--- a/infrastructure/resource_group_level/main.bicep
+++ b/infrastructure/resource_group_level/main.bicep
@@ -207,7 +207,7 @@ module functionAcbs 'modules/function-acbs.bicep' = {
   params: {
     environment: environment
     location: location
-    containerRegistryName: containerRegistryName
+    containerRegistryName: containerRegistry.name
     appServicePlanEgressSubnetId: vnet.outputs.appServicePlanEgressSubnetId
     appServicePlanId: appServicePlan.id
     privateEndpointsSubnetId: vnet.outputs.privateEndpointsSubnetId
--- a/infrastructure/resource_group_level/main.bicep
+++ b/infrastructure/resource_group_level/main.bicep
@@ -1,5 +1,6 @@
 param location string  = resourceGroup().location
-// Expected values are 'feature', 'dev', 'test', 'staging' & 'prod'
+// Expected values are 'feature', 'dev', 'staging' & 'prod'
+// Note that legacy values of 'test' and 'qa' may be observed in some resources. These are equivalent to 'staging'.
 param environment string = 'feature'
 
 param appServicePlanName string = 'feature'
--- a/infrastructure/resource_group_level/modules/function-acbs.bicep
+++ b/infrastructure/resource_group_level/modules/function-acbs.bicep
@@ -50,7 +50,7 @@ var settings = {
 
 // These values are taken from an export of Configuration on Dev
 var additionalSettings = {
-  APPINSIGHTS_INSTRUMENTATIONKEY: 'TODO:FN-419 replace with appInsights.properties.InstrumentationKey'
+  APPINSIGHTS_INSTRUMENTATIONKEY: applicationInsights.properties.InstrumentationKey
   // TODO:FN-419 - Exported string had "EndpointSuffix=core.windows.net" in it. Check if needed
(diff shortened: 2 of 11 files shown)


infrastructure/resource_group_level/main.bicep
```diff
@@ -1,4 +1,5 @@
 param location string  = resourceGroup().location
+// Expected values are 'feature', 'dev', 'test', 'staging' & 'prod'
```
OxonianCambion, line 2: I think there is confusion between `qa`, `test` & `staging`. I think they are all the same, right?

infrastructure/resource_group_level/modules/function-acbs.bicep
```diff
@@ -0,0 +1,172 @@
...
+  MULESOFT_API_KEY: 'test-value'
+  MULESOFT_API_SECRET: 'test-value'
+  MULESOFT_API_UKEF_TF_EA_URL: 'test-value'
+  APIM_MDM_KEY: 'test-value'
+  APIM_MDM_URL: 'test-value'
+  APIM_MDM_VALUE: 'test-value' // different in staging and dev
+}
+
+// These values are taken from an export of Configuration on Dev
+// TODO:FN-419 Add to GitHub secrets?
```
OxonianCambion, line 22: I see that in https://github.com/UK-Export-Finance/dtfs2/pull/1767 these values will change. @abhi-markan 

I can either wait for your PR to be merged, merge that into my long-lived, branch, rebase this, and make the change now, or make the change in a future card / PR, with a TODO here. I'll do the latter by default!

infrastructure/resource_group_level/modules/function-acbs.bicep
```diff
@@ -0,0 +1,172 @@
...
+  name: storageAccountName
+}
+
+var storageAccountKey = storageAccount.listKeys().keys[0].value
+
+// These values are hardcoded in the CLI scripts
+var settings = {
+  FUNCTIONS_WORKER_RUNTIME: 'node'
+  // TODO:FN-419 consider making this a parameter
+  WEBSITE_DNS_SERVER: '168.63.129.16'
```
OxonianCambion, line 47: I can't see where this IP comes from. Any ideas?

infrastructure/resource_group_level/modules/function-acbs.bicep
```diff
@@ -0,0 +1,172 @@
...
+  WEBSITE_VNET_ROUTE_ALL: '1'
+}
+
+// These values are taken from an export of Configuration on Dev
+var additionalSettings = {
+  APPINSIGHTS_INSTRUMENTATIONKEY: applicationInsights.properties.InstrumentationKey
+  // TODO:FN-419 - Exported string had "EndpointSuffix=core.windows.net" in it. Check if needed
+  AzureWebJobsStorage: 'DefaultEndpointsProtocol=https;AccountName=${storageAccountName};AccountKey=${storageAccountKey}'
+  // TODO:FN-419 I think DOCKER_CUSTOM_IMAGE_NAME is overridden by linuxFxVersion
+  DOCKER_CUSTOM_IMAGE_NAME: dockerImageName
```
OxonianCambion, line 57: By default, I'll just leave this as is.

infrastructure/resource_group_level/modules/function-acbs.bicep
```diff
@@ -0,0 +1,172 @@
...
+  FUNCTIONS_WORKER_RUNTIME: 'node'
+  // TODO:FN-419 consider making this a parameter
+  WEBSITE_DNS_SERVER: '168.63.129.16'
+  WEBSITE_VNET_ROUTE_ALL: '1'
+}
+
+// These values are taken from an export of Configuration on Dev
+var additionalSettings = {
+  APPINSIGHTS_INSTRUMENTATIONKEY: applicationInsights.properties.InstrumentationKey
+  // TODO:FN-419 - Exported string had "EndpointSuffix=core.windows.net" in it. Check if needed
```
OxonianCambion, line 54: The EndpointSuffix isn't needed, right? https://learn.microsoft.com/en-us/azure/storage/common/storage-configure-connection-string#create-a-connection-string-with-an-endpoint-suffix

infrastructure/resource_group_level/modules/function-acbs.bicep
```diff
@@ -0,0 +1,172 @@
...
+  AzureWebJobsStorage: 'DefaultEndpointsProtocol=https;AccountName=${storageAccountName};AccountKey=${storageAccountKey}'
+  // TODO:FN-419 I think DOCKER_CUSTOM_IMAGE_NAME is overridden by linuxFxVersion
+  DOCKER_CUSTOM_IMAGE_NAME: dockerImageName
+  DOCKER_ENABLE_CI: 'true'
+  DOCKER_REGISTRY_SERVER_URL: '${containerRegistryName}.azurecr.io'
+  DOCKER_REGISTRY_SERVER_USERNAME: dockerRegistryServerUsername
+  FUNCTION_APP_EDIT_MODE: 'readOnly'
+  FUNCTIONS_EXTENSION_VERSION: '~3'
```
OxonianCambion, line 62: We can upgrade to V4 later. We should do it soon after though. V3 is EOL.

infrastructure/resource_group_level/modules/function-acbs.bicep
```diff
@@ -0,0 +1,172 @@
...
+  location: location
+  tags: {
+    Environment: 'Preproduction'
+  }
+  kind: 'functionapp,linux,container'
+  properties: {
+    httpsOnly: false
+    serverFarmId: appServicePlanId
+    siteConfig: {
+      // TODO:FN-419 These siteConfig values are included in the much more comprehensive config object, produced by the export.
```
OxonianCambion, line 92: Note that when I re-export this, I get the duplication, so I think all is well and I put the settings here, and just use the `functionAcbsAppSettings` resource below for environment variables.

infrastructure/resource_group_level/modules/function-acbs.bicep
```diff
@@ -0,0 +1,172 @@
...
+// These values are hardcoded in the CLI scripts
+var settings = {
+  FUNCTIONS_WORKER_RUNTIME: 'node'
+  // TODO:FN-419 consider making this a parameter
+  WEBSITE_DNS_SERVER: '168.63.129.16'
+  WEBSITE_VNET_ROUTE_ALL: '1'
+}
+
+// These values are taken from an export of Configuration on Dev
+var additionalSettings = {
```
OxonianCambion, line 52: I'm not totally happy with the Azure convention of `param`, `var`, `resource` ordering here as this var has a dependency on the applicationInsights resource below. I suppose it is ok.

infrastructure/resource_group_level/modules/function-acbs.bicep
```diff
@@ -0,0 +1,172 @@
...
+  name: applicationInsightsName
+  location: location
+  kind: 'web'
+  properties: {
+    Application_Type: 'web'
+  }
+}
+
+
+// TODO:FN-419 Add automatic A Record generation.
```
OxonianCambion, line 172: say if I should do this as part of the current PR, or later in a big batch.

infrastructure/resource_group_level/main.bicep
IgnatG: `staging` and `test` are the same. - they got merged into `staging` in the end

infrastructure/resource_group_level/modules/function-acbs.bicep
IgnatG: good point. no idea :(

infrastructure/resource_group_level/modules/function-acbs.bicep
IgnatG: double check if it's needed. if not, then we can remove it

infrastructure/resource_group_level/modules/function-acbs.bicep
IgnatG: agreed

infrastructure/resource_group_level/modules/function-acbs.bicep
```diff
@@ -0,0 +1,172 @@
...
+      minimumElasticInstanceCount: 1
+      // The following Fields have been added after comparing with dev
+      // TODO:FN-419 Check these values are ok.
+      vnetRouteAllEnabled: true
+      // Note that the following appear in the separate config object
+      ftpsState: 'Disabled'
+      scmMinTlsVersion: '1.0'
+      remoteDebuggingVersion: 'VS2019'
+      httpLoggingEnabled: true // false in staging
+      // TODO:FN-419 Note that the following appear in dev but not staging or prod
```
OxonianCambion, line 109: I've moved this to a "mopping up things that will be easy to see when we can deploy" card.

IgnatG (approved):
LGTM

PR #1770, opened 2023-06-28
fix(SR-1): Add JWT authentication to all endpoints in portal-api

JennyHumphries:
## Introduction
Fix for SR-1. Portal API already had JWT authentication in place for most endpoints. But there were some endpoints that incorrectly weren't covered.

## Resolution
* There were three categories the endpoints were previously split into: "open", "authAllowXss" and "auth". I've moved all of the requests that were in "open" into "authAllowXss" except for the login endpoint and the two reset password endpoints. And the post feedback endpoint that now uses the api key.
* This meant I also needed to fix `re-insert-mocks`, as this relies on using the users endpoint when not logged in. As discussed on Slack, the approach we've gone for is to add a new `user` endpoint that has api key authentication which will be used for creating the initial test user.

## TODO
* Ensure all tests are passing ✔


portal-api/src/createApp.js
```diff
@@ -35,7 +36,11 @@ app.use(cors({
 }));
 
 app.use('/v1', openRouter);
+
+app.use(passport.authenticate('jwt', { session: false }));
```
JennyHumphries, line 40: I've moved this and the cleanXss steps into this file, so that the routes file only contains the routes.

I think it's easier to follow the order that things are applied this way.

portal-api/src/v1/routes.js
```diff
@@ -28,25 +27,50 @@ const { ukefDecisionReport, unissuedFacilitiesReport } = require('./controllers/
...
+// 1. Request password reset
+openRouter.route('/users/reset-password').post(users.resetPassword);
+
+// 2. Password reset post request
+openRouter.route('/users/reset-password/:resetPwdToken').post(users.resetPasswordWithToken);
+
 const authRouterAllowXss = express.Router();
 
-authRouterAllowXss.use(passport.authenticate('jwt', { session: false }));
+authRouterAllowXss.route('/feedback').post(feedback.create);
```
JennyHumphries, line 47: This request and the ones on lines 49-51 previously used `openRouter` so didn't have the JWT check

portal-api/src/v1/routes.js
```diff
@@ -28,25 +27,50 @@ const { ukefDecisionReport, unissuedFacilitiesReport } = require('./controllers/
...
+const users = require('./users/routes');
 const gef = require('./gef/routes');
 
-const authRouter = express.Router();
 const openRouter = express.Router();
+
+openRouter.route('/login').post(users.login);
+
+// 1. Request password reset
+openRouter.route('/users/reset-password').post(users.resetPassword);
```
JennyHumphries, line 40: Should both this and the route on line 43 stay on the `openRouter`? I think these are the ones that would be used if you've forgotten your password and are resetting it, so shouldn't be locked behind being logged in.

portal-api/src/v1/routes.js
```diff
@@ -1,5 +1,4 @@
 const express = require('express');
```
JennyHumphries, line 1: I've rearranged this file so that the requests are grouped by what router they are using, and put them in the order that they are run from `createApp.js`

ChristianMcc:
Agreed with this approach for portal-API - unless it's already in the new endpoints that are covered you'll need to update the api.js file in portal-UI to pass through the JWT token in the authorization header for the new endpoints. A bunch of them already have it so it's just a case of making sure it's passed through everywhere :) (And hopefully that fixes a bunch of the tests apart from the re-insert mocks one!)

portal-api/src/v1/routes.js
ChristianMcc: I've got a similar one on TFM - this means the UI page also needs to be not accessible when not logged in and redirect when logged in. Haven't fixed this on my branch yet but just a heads up :)

portal-api/src/v1/routes.js
ChristianMcc: Agreed with you here! I'm assuming these either send an email to the registered email or similar so if that's the case it's fine for it to be open. 
For the password reset token have to look at that, not sure how and when that's called. I'll come take a look later unless someone beats me to it looking at this!

JennyHumphries: I think both these look okay to keep open.

`reset-password` is used in a "I've forgotten my password" link. So you shouldn't need to be logged in for that.

And `reset-password/:resetPwdToken` looks like it comes from the link in a reset password email. So again that feels like you shouldn't need to be logged in for it.

utils/mock-data-loader/.env.sample
```diff
@@ -1,3 +1,6 @@
 DEAL_API_URL=http://localhost:5001
 DTFS_CENTRAL_API=http://localhost:5005
-TFM_API=http://localhost:5004
\ No newline at end of file
+TFM_API=http://localhost:5004
+
+MONGO_INITDB_DATABASE=
```
JennyHumphries, line 5: Once this is merged devs will need to fill in these values. They should already have the values from other `.env` files. I've made a note to post on Slack when this is merged

portal-api/src/v1/middleware/headers/check-api-key.js
```diff
@@ -0,0 +1,42 @@
+const dotenv = require('dotenv');
+
+dotenv.config();
+
+const { API_KEY } = process.env;
```
JennyHumphries, line 5: This is using the same api key as `dtfs-central-api` and `external-api`. Just flagging to make sure that that's the correct approach

dtfs-central-api/api-tests/v1/routes/middleware/headers/checkApiKey.api-test.js
```diff
@@ -1,5 +1,5 @@
 const dotenv = require('dotenv');
-const checkApiKey = require('../../../../../src/v1/routes/middleware/headers/checkApiKey');
+const checkApiKey = require('../../../../../src/v1/routes/middleware/headers/check-api-key');
```
ChristianMcc, line 2: Nitpick as not yours but if this is check-api-key the file name should also be check-api-key.api-test.js

portal-api/api-tests/api-test-eligibilityCriteria.js
```diff
@@ -14,7 +14,7 @@ module.exports.initialise = async (app, user) => {
     const existingECs = currentECResponse.body.eligibilityCriteria;
 
     for (const existingEC of existingECs) {
-      await remove(`/v1/users/${existingEC._id}`);
+      await remove(`/v1/eligibility-criteria/${existingEC._id}`);
```
ChristianMcc, line 17: 🙈

portal-api/api-tests/api-test-users.js
```diff
@@ -199,22 +202,50 @@ const finder = () => {
...
+  roles: ['maker'],
+  bank: banks.any,
+};
+
+const setUpApiTestUser = async (as) => {
+  const { salt, hash } = genPassword(apiTestUser.password);
+
+  const userToCreate = {
+    'user-status': 'active',
+    timezone: apiTestUser.timezone || 'Europe/London',
```
ChristianMcc, line 221: As it's a const defined above here and not a parameter this line is redundant

portal-api/api-tests/api-test-users.js
```diff
@@ -199,22 +202,50 @@ const finder = () => {
...
+    salt,
+    hash,
+    ...apiTestUser,
+  };
+
+  userToCreate.password = '';
+  userToCreate.passwordConfirm = '';
+
+  const collection = await db.getCollection('users');
+  await collection.insertOne(userToCreate);
```
ChristianMcc, line 231: Do API tests already insert etc.? I've got lost in whether the new endpoint was for mock data loader only or if we are suppose to use it for these tests too. Personally I still prefer this!

portal-api/src/createApp.js
ChristianMcc: I'm happy to leave it this way but I find it slightly more intuitive to have the 3 separate routers with different levels of protection/middleware defined on the separate routers rather than looking at the ordering of the different routers but I suspect that's personal preference!

ChristianMcc (commented):
LGTM! Couple of nit-picks but nothing to do with code logic apart from one question on seeding the API test user!

portal-api/src/createApp.js
JennyHumphries: Hmm yes I see what you mean. I'll move them back in

portal-api/api-tests/api-test-users.js
JennyHumphries: It looks like there are some other api tests that access the db directly. But they just use it to find a user rather than add one.

I think I agree, since the tests already have direct access to the database, it's best to directly insert the new user rather than going via the `/user` endpoint.

dtfs-central-api/api-tests/v1/routes/middleware/headers/checkApiKey.api-test.js
Zainzzkk: fixed in my PR

portal-api/src/v1/routes.js
JennyHumphries: Resolving this thread as we're now keeping the page visible to non-logged in users and using an api key as authentication

portal-api/src/v1/routes.js
```diff
@@ -28,25 +27,50 @@
...
-const { cleanXss, fileUpload } = require('./middleware');
+const { fileUpload } = require('./middleware');
 
+const users = require('./users/routes');
 const gef = require('./gef/routes');
 
-const authRouter = express.Router();
 const openRouter = express.Router();
+
+openRouter.route('/login').post(users.login);
```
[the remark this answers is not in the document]
abhi-markan: There is another SR which will cover rate limiting. Thanks

portal-api/src/v1/routes.js
abhi-markan: That is correct @JennyHumphries

utils/mock-data-loader/api.js
```diff
@@ -4,6 +4,7 @@ require('dotenv').config();
 const { gef } = require('./gef/api');
 
 const portalApiUrl = process.env.DEAL_API_URL;
+const API_KEY = process.env.API_KEY;
```
abhi-markan, line 7: Can we re-write it as below please.

```javascript
const { API_KEY } = process.env
```

utils/mock-data-loader/api.js
```diff
@@ -124,12 +125,13 @@ const createEligibilityCriteria = async (eligibilityCriteria, token) => {
...
 };
 
-const createUser = async (user) => {
+const createUser = async (user, token) => {
   const response = await axios({
     method: 'post',
     headers: {
       'Content-Type': 'application/json',
       Accepts: 'application/json',
+      Authorization: token || '',
```
abhi-markan, line 134: Do we need to an empty string if `token` is falsey? Also applies to all of the below aswell.

utils/mock-data-loader/api.js
```diff
@@ -138,6 +140,21 @@ const createUser = async (user) => {
...
+  const response = await axios({
+    method: 'post',
+    headers: {
+      'Content-Type': 'application/json',
+      'x-api-key': API_KEY,
+      Accepts: 'application/json',
+    },
+    url: `${portalApiUrl}/v1/user`,
+    data: user,
+  }).catch((err) => { console.error(`err: ${err}`); });
```
abhi-markan, line 153: Can the console.error please be written as:

```javascript
console.error('Unable to create initial user %s', {$err});
```

utils/mock-data-loader/api.js
```diff
@@ -347,19 +365,33 @@ const listEligibilityCriteria = async (token) => {
...
 
+const login = async (user) => {
+  const response = await axios({
+    method: 'post',
+    url: `${portalApiUrl}/v1/login`,
+    headers: {
+      'Content-Type': 'application/json',
+    },
+    data: { username: user.username, password: user.password },
+  }).catch((err) => { console.error(`err: ${err}`); });
```
abhi-markan, line 390: Can the console.error please be written as:

```javascript
console.error('Unable to login %s', {$err});
```

utils/mock-data-loader/user-helper.js
```diff
@@ -0,0 +1,36 @@
...
+  password: 'AbC!2345',
+  firstname: 'Mock',
+  surname: 'DataLoader',
+  roles: ['editor', 'maker'],
+  email: 're-insert-mocks-data-loader@ukexportfinance.gov.uk',
+  bank: MOCK_BANKS.find((bank) => bank.id === '9'),
+};
+
+const createAndLogInAsInitialUser = async () => {
+  console.info(`try log in as "${mockDataLoaderUser.username}"`);
```
abhi-markan, line 15: We are writing console.info messages without double quotes inside the string, can they all please be removed.

abhi-markan (changes requested):
Good work, only few minor comments.

portal-api/src/v1/routes.js
```diff
@@ -28,25 +28,54 @@ const { ukefDecisionReport, unissuedFacilitiesReport } = require('./controllers/
...
+openRouter.route('/users/reset-password').post(users.resetPassword);
+
+// 2. Password reset post request
+openRouter.route('/users/reset-password/:resetPwdToken').post(users.resetPasswordWithToken);
+
+// API Key Routes
+openRouter.route('/feedback').post(checkApiKey, feedback.create);
+openRouter.route('/user').post(checkApiKey, users.create); // This endpoint is only used by mock-data-loader, for setting up an initial user
+
+const authRouterAllowXss = express.Router();
```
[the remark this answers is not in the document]
JennyHumphries: I think so, these routes come before `authRouter.use(cleanXss)` on line 74.

There were already some routes set up on this router before my changes. I'm not sure why they were set up to avoid the `cleanXss` check.

But I have moved the users routes to this one. Do you think they should they be moved to `authRouter` instead?

utils/mock-data-loader/api.js
JennyHumphries: I'm not sure, I was following the same pattern as in the rest of this file. But I think since this is only called from within `mock-data-loader` we wouldn't expect the token to be falsey so maybe shouldn't account for it.

Should I remove the `|| ''` from all of these endpoints?

utils/mock-data-loader/api.js
JennyHumphries: Yeah no problem.

Just out of interest, why is the preference to use `%s` rather than the other type of string interpolation?

utils/mock-data-loader/api.js
abhi-markan: Please remove.

portal-api/src/v1/routes.js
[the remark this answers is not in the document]
JennyHumphries: `mock-data-loader` still works after moving the user endpoints to `authRouter` so I will move them there.

abhi-markan (approved):

PR #1772, opened 2023-06-28
fix(DTFS2): Fix get amendments related central API tests

ChristianMcc:
## Introduction
API tests on main were failing in central API.

## Resolution
Fix one case for invalid amendment APIs and then fix remaining test suites. Majority of the fixes are test file related only.


JennyHumphries (approved):

ChristianMcc:
Have a couple of failing E2E tests which I think are flaky and not to do with changes, will rerun and merge once they are all passing

abhi-markan (approved):

PR #1777, opened 2023-06-29
fix(SR-6): Portal UI - consume and sanitise only specific query params

ttbarnes:
## Introduction ✏️ 
This PR adds some middleware to the portal UI so that only a single valid query parameter will be consumed. This prevents unwanted and "bad" query parameters from being consumed by the UI.

## Resolution ✔️ 
* Create new `sanitiseData` file with `replaceCharactersWithCharacterCode` helper function. This is also used in other other services.
* Create new `queryParams` middleware that checks for specific params, consumes and sanitises only the first param via `getSanitisedQueryValue`. Otherwise, return params as an empty object.

## Miscellaneous 🗒️ 
* Replace `responseMock` with `test-mocks` - now aligned with other services.


abhi-markan:
Closing for now, to be re-opened post pen-test report.

PR #1780, opened 2023-06-30
fix(SR-6): Portal UI - validate referer and origin request headers

ttbarnes:
⚠️ Need this PR merged first - there will be merge conflicts RE test mocks that will resolve the unit tests in this PR - there should be no `responseMock.js` and instead, replaced with `test-mocks` https://github.com/UK-Export-Finance/dtfs2/pull/1777

## Introduction ✏️ 
This PR adds some middleware to the portal UI so that the "referer" and "origin" request headers will be checked. If either header does not match the servers URL origin/hostname, or are invalid URLs, the headers will be changed to an empty string. This prevents unwanted manipulation and consumption of request headers" in the UI.

## Resolution ✔️ 
* Add `is-url` package to check that the said headers are valid URLs.
* Create new security middleware function `getUrlOrigin` to split a URL and return the origin.
* Create new security middleware function `isRequestHeaderOriginValid`, to check if a provided header URL matches the true origin.
* Update headers security middleware to check the said headers via `isRequestHeaderOriginValid` and if they are invalid, change the headers to an empty string.

## Miscellaneous 🗒️ 
* Update the mock request in test mocks.
* Update security middleware `Cache-Control` header, add missing test coverage. This is actually in another PR as well so maybe a merge conflict here.


abhi-markan:
Closing for now, to be re-opened post pen-test report.

PR #1815, opened 2023-07-05
fix(SR6): removed deal and facility import endpoints

IgnatG:


abhi-markan (approved):

abhi-markan:
Happy for the files to be deleted as we will not use the import service anymore.

PR #1817, opened 2023-07-05
fix(portal-api): Replaced `authRouterAllowXss` with `authRouter`

abhi-markan:
## Introduction
Removed `authRouterAllowXss` express router.

## Resolution
* Replaced with `authRouter`.
* Mandatory criteria `POST` and `PUT` HTTP methods allow insertion on non-production environment only.
* Mandatory criteria `GET` and `DELETE` HTTP methods are subjected to XSS.


portal-api/src/v1/routes.js
```diff
@@ -28,25 +28,49 @@ const { ukefDecisionReport, unissuedFacilitiesReport } = require('./controllers/
...
 
+const users = require('./users/routes');
 const gef = require('./gef/routes');
 
-const authRouter = express.Router();
 const openRouter = express.Router();
-const authRouterAllowXss = express.Router();
+const authRouter = express.Router();
+
+authRouter.use(cleanXss);
```
JennyHumphries, line 40: As mentioned on the call, we should potentially swap this and line 41, just so the checks are done in the same order as previously

JennyHumphries (approved):

portal-api/src/v1/routes.js
IgnatG, line 1: Is it worth deleting this file: clean-xss.js
Since we are not using it anymore

portal-api/src/v1/routes.js
```diff
@@ -44,32 +47,43 @@ openRouter.route('/users/reset-password/:resetPwdToken').post(users.resetPasswor
...
-authRouterAllowXss.use(passport.authenticate('jwt', { session: false }));
+authRouter
+  .route('/mandatory-criteria/:version')
+  .put(validate({ role: ['editor'] }), mandatoryCriteria.update);
 
-authRouterAllowXss.route('/users').get(users.list).post(users.create);
-authRouterAllowXss.route('/users/:_id').get(users.findById).put(users.updateById).delete(users.remove);
-authRouterAllowXss.route('/users/:_id/disable').delete(users.disable);
+// Enable XSS
+authRouter.use(cleanXss);
```
IgnatG, line 67: Do we know why we need this?

IgnatG: Previously we had it enabled only for a few endpoints. Now we have it enabled for all of them

IgnatG (dismissed):

portal-api/src/v1/routes.js
```diff
@@ -44,32 +47,43 @@ openRouter.route('/users/reset-password/:resetPwdToken').post(users.resetPasswor
...
-openRouter.route('/user').post(checkApiKey, users.create); // This endpoint is only used by mock-data-loader, for setting up an initial user
+// This endpoint is only used by mock-data-loader, for setting up an initial user
+openRouter.route('/user').post(checkApiKey, users.create);
+
+/**
+ * Mandatory Criteria routes
+ * Allow POST & PUT of MC HTML tags
+ * on non-production environments only
+ */
+authRouter
```
JennyHumphries, line 58: I think it would be better to have this and the routes on line 62 in a separate router. Because currently some routes on `authRouter` are subject to different checks than others which could be confusing to future developers.

We could call this router something like `nonProductionRouter` to make it clear that the endpoints on it should only be available to non-production environments. And maybe a comment to explain that they shouldn't be available on production because they are not subject to the xss check.

portal-api/src/v1/routes.js
```diff
@@ -33,6 +33,9 @@ const users = require('./users/routes');
 const gef = require('./gef/routes');
 
 const openRouter = express.Router();
+const authRouter = express.Router();
```
JennyHumphries, line 36: Can we move the create router and the `use` statements to be next to each other? For someone new to this file currently it would be easy for them to miss that `authRouter` has an xss check because it's so much further down the file.

portal-api/src/v1/controllers/mandatoryCriteria.controller.js
```diff
@@ -26,10 +26,15 @@ const findOneMandatoryCriteria = async (version, callback) => {
 };
 
 exports.create = async (req, res) => {
-  const collection = await db.getCollection('mandatoryCriteria');
-  const mandatoryCriteria = await collection.insertOne(req.body);
+  // MC insertion on non-production environments
+  if (process.env.NODE_ENV !== 'production') {
```
JennyHumphries, line 30: It might be safer to instead list the environments we're happy for this to run on.

If production was ever renamed, or if another production-like environment was added these endpoints would be available on it by default. It could be easy to miss locking it down in future.

JennyHumphries (commented):
Just a couple of small suggestions 😃

portal-api/src/v1/routes.js
JennyHumphries: Actually, if we do go for this option, we could potentially move the environment check up into `createApp` and only use the router for non-production environments. 
Then it wouldn't be possible for any endpoints using it to accidentally forget to add an environment check.

portal-api/src/v1/controllers/mandatoryCriteria.controller.js
abhi-markan: NODE_ENV is always declared, other envs list is large so decided to target prod environment only.

portal-api/src/v1/routes.js
abhi-markan: Not all the routes are used for non production environment.

JennyHumphries: The two routes I'm talking about are only used in non-production though right? Just the POST and the PUT?

I mean something like this:
```
const nonProductionOnlyRouter = express.Router();

nonProductionOnlyRouter
  .route('/mandatory-criteria')
  .post(passport.authenticate('jwt', { session: false }), validate({ role: ['editor'] }), mandatoryCriteria.create);

nonProductionOnlyRouter
  .route('/mandatory-criteria/:version')
  .put(passport.authenticate('jwt', { session: false }), validate({ role: ['editor'] }), mandatoryCriteria.update);

const authRouter = express.Router();
authRouter.use(passport.authenticate('jwt', { session: false }));
authRouter.use(cleanXss);

authRouter
  .route('/mandatory-criteria')
  .get(mandatoryCriteria.findAll);

authRouter
  .route('/mandatory-criteria/latest')
  .get(mandatoryCriteria.findLatest);

...
```

And then in `createApp.js` you'd have something like this:
```
app.use('/v1', openRouter);
if (process.env.NODE_ENV !== 'production') {
  app.use('/v1', nonProductionOnlyRouter)
}
app.use('/v1', authRouter);
```
So that the check for what environment we're on is brought right to the top in createApp rather than being needed to be implemented on each individual route.


But I'll leave it up to you to decide whichever way you prefer. 😄

JennyHumphries (approved):
Looks good, I've just added one comment but I'll leave it to you to decide if you want to change it or not so approving the PR.

portal-api/src/v1/routes.js
abhi-markan: Thanks for the comment ill cover this in a separate PR where I'll clean the code and introduce functions ascertaining environment type. Thanks

PR #1819, opened 2023-07-06
fix(SR13): fixed exception text reinterpreted as HTML

IgnatG:
### Changeling
- Fixed SR13: exception text being sent as part of the response


portal-api/src/v1/gef/controllers/files.controller.js
```diff
@@ -64,38 +67,40 @@ exports.create = async (req, res) => {
...
+      }),
+    );
 
     const status = processedFiles.every((file) => !!file.error) ? 200 : 201;
 
     return res.status(status).send(processedFiles);
   } catch (err) {
     console.error(`Error uploading file(s): ${err}`);
-    return res.status(500).send(err.message);
+    return res.status(500).send();
```
abhi-markan, line 103: Error message needs to be sanitized please

abhi-markan (changes requested):
We need to sanitize the error message

portal-api/src/v1/gef/controllers/files.controller.js
IgnatG: @abhi-markan updated

abhi-markan (approved):

abhi-markan:
Nice work. Failing API test to address please.

PR #1820, opened 2023-07-06
feat(FN-421): add external-api webapp to Bicep

OxonianCambion:
### Introduction

This card covers:
- Adding the external-api webapp (previously called reference data proxy)
### Resolution
- Adds the external-api webapp in a way that should make it easy to commonise later
- DOESN'T add Application Insights, but I've left the code to do so, but commented out.

### DIFF
--- a/infrastructure/resource_group_level/modules/external-api.bicep
+++ b/infrastructure/resource_group_level/modules/external-api.bicep
@@ -7,17 +7,35 @@ param privateEndpointsSubnetId string
 param cosmosDbAccountName string
 param cosmosDbDatabaseName string
 param logAnalyticsWorkspaceId string
+param acbsFunctionDefaultHostName string
+param numberGeneratorFunctionDefaultHostName string
 
-var dockerImageName = '${containerRegistryName}.azurecr.io/dtfs-central-api:${environment}'
+var dockerImageName = '${containerRegistryName}.azurecr.io/external-api:${environment}'
 var dockerRegistryServerUsername = 'tfs${environment}'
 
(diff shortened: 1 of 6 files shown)


infrastructure/resource_group_level/modules/external-api.bicep
OxonianCambion, line 1: I'm trying to use almost the exact same form as for `dtfs-central-api`. I hope to commonise all the WebApps if it is easy.
My plan is to create them all like this so I can spot the differences more easily, then commonise.

infrastructure/resource_group_level/modules/external-api.bicep
```diff
@@ -0,0 +1,185 @@
...
+    manualPrivateLinkServiceConnections: []
+    subnet: {
+      id: privateEndpointsSubnetId
+    }
+    ipConfigurations: []
+    // Note that the customDnsConfigs array gets created automatically and doesn't need setting here.
+  }
+}
+
+// Note that Application Insights is disabled for this resource.
```
OxonianCambion, line 163: Currently, I'm trying to replicate the current infrastructure exactly.

I'd imagine we probably want Application Insights enabled for all our webapps, though. Is it worth the effort of preserving the differences and then harmonising them later? (Possibly - I'm not sure of the right answer here!)

Anyway, I thought it reasonable to leave this commented out to make it easier to see how it could be commonised as per FN-712.

IgnatG: for now we can leave it as is it. the APIs will be refactored behind the scenes. But in the long run all apps would have App Insights enabled

PR #1821, opened 2023-07-07
fix(sr-5): Add SSRF protection to TFM microservices

ChristianMcc:
## Introduction
TFM needs protection against SSRF attacks

## Resolution
Add in validation of all params that are used in further API calls to make sure we are protected against SSRF attacks


trade-finance-manager-ui/server/helpers/validateIds.js
```diff
@@ -0,0 +1,19 @@
+const validator = require('validator');
+
+const isValidMongoId = (mongoId) => {
```
IgnatG, line 3: we can simplify this into
```js
return mongoId ? validator.isMongoId(mongoId) : false;
```

trade-finance-manager-api/src/v1/validation/validateIds.js
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  const regex = /^\d+$/;
+
+  return regex.test(numericId);
+};
+
+const isValidCurrencyCode = (currencyCode) => (currencyCode ? validator.isISO4217(currencyCode) : false);
+
+const sanitizeUsername = (username) => validator.escape(username);
```
ChristianMcc, line 38: We have one endpoint that uses username in a URL. This comes from a graphql `mutation-update-lead-underwriter` which could come from user input. However username's have no validation when they are saved and can be made up of any characters so it is hard to validate and narrow down what an allowed username is so right now all I do is escape the username. Would welcome any thoughts here

trade-finance-manager-api/api-tests/api-ssrf.api-test.js
```diff
@@ -0,0 +1,1349 @@
...
+      const urlTraversal = '../../../etc/stealpassword';
+      const expectedResponse = { status: 400, data: 'Invalid deal id' };
+
+      const response = await api.updatePortalDeal(urlTraversal, { deal: 'Mock deal' });
+
+      expect(response).toMatchObject(expectedResponse);
+    });
+
+    it('Returns an error when a local IP is supplied', async () => {
+      const localIp = '127.0.0.1';
```
abhi-markan, line 64: Repetitive values, can we declare them as global variables please?

trade-finance-manager-api/api-tests/api-ssrf.api-test.js
```diff
@@ -0,0 +1,1349 @@
...
+      const expectedResponse = { status: 400, data: 'Invalid deal id' };
+
+      const response = await api.findOnePortalDeal(localIp);
+
+      expect(response).toMatchObject(expectedResponse);
+    });
+
+    it('Makes an axios request when the deal id is valid', async () => {
+      const validDealId = '5ce819935e539c343f141ece';
```
abhi-markan, line 38: Can we add assertions where deal ID is in valid format but does not exists:

abhi-markan: Same applies to all of the below please.

trade-finance-manager-api/api-tests/v1/utils/validateIds.api-test.js
```diff
@@ -0,0 +1,130 @@
...
+  isValidPartyUrn,
+  isValidNumericId,
+  isValidCurrencyCode,
+  isValidTeamId,
+} = require('../../../src/v1/validation/validateIds');
+
+describe('validateIds', () => {
+  describe('isValidMongoId', () => {
+    it('should return false if an id is not provided', () => {
+      const result = isValidMongoId();
```
abhi-markan, line 13: Can we add assertion for all the test cases where input as:

* {}
* []

trade-finance-manager-api/src/v1/api.js
```diff
@@ -229,6 +293,12 @@ const findFacilitesByDealId = async (dealId) => {
 
 const updateFacility = async (facilityId, facilityUpdate) => {
   try {
+    const isValidFacilityId = isValidMongoId(facilityId);
+
+    if (!isValidFacilityId) {
+      return { status: 400, data: 'Invalid facility id' };
```
abhi-markan, line 299: Will all these returned be caught and consoled error at the calling side or should we add `console.error` for all the non happy branches?

trade-finance-manager-api/src/v1/api.js
```diff
@@ -346,13 +416,13 @@ const getLatestCompletedAmendmentValue = async (facilityId) => {
       return { status: 500, data: err?.response?.data };
     }
   } else {
-    console.error('Invalid facility Id');
-    return { status: 400, data: 'Invalid facility Id amendment Id provided' };
+    console.error(`Invalid facility Id: ${facilityId}`);
```
abhi-markan, line 419: Please change all console.error for consistency to below:

```javascript
console.error('Invalid facility id %s', facilityId);
```

trade-finance-manager-api/src/v1/api.js
```diff
@@ -956,20 +1108,32 @@ const getAllFacilities = async (searchString) => {
...
+
     const response = await axios({
       method: 'get',
       url: `${DTFS_CENTRAL_API_URL}/v1/bank/${bankId}`,
       headers: headers.central,
     });
     return response.data;
-  } catch (response) {
+  } catch ({ response }) {
     console.error('Unable to get bank by id', response?.data);
```
abhi-markan, line 1124: Console error with string substitution `%s` please.

trade-finance-manager-api/src/v1/api.js
```diff
@@ -956,20 +1108,32 @@ const getAllFacilities = async (searchString) => {
...
     return response?.data;
   }
 };
 
 const getGefMandatoryCriteriaByVersion = async (version) => {
   try {
+    const isValidVersion = isValidNumericId(version);
+
+    if (!isValidVersion) {
+      return { status: 400, data: 'Invalid mandatory criteria version provided' };
```
abhi-markan, line 1134: Console.error required here or will be printed by the calling function?

trade-finance-manager-api/src/v1/validation/validateIds.js
```diff
@@ -0,0 +1,60 @@
+const validator = require('validator');
+const teams = require('../../constants/teams');
+
+const isValidMongoId = (mongoId) => (mongoId ? validator.isMongoId(mongoId) : false);
+
+const isValidUkefNumericId = (ukefId) => {
```
abhi-markan, line 6: Missing function documentation for all the function here please.

trade-finance-manager-api/src/v1/validation/validateIds.js
```diff
@@ -0,0 +1,60 @@
...
+  const id = parseInt(ukefId, 10);
+
+  if (Number.isNaN(id)) {
+    return false;
+  }
+
+  const regex = /^\d{10}$/;
```
abhi-markan, line 13: Can all these REGEX please be defined and picked from CONSTANTS please.

abhi-markan: A single helper regex function will work where `isValidInput(REGEX_UKEFID, id)` and etc.

trade-finance-manager-api/src/v1/validation/validateIds.js
abhi-markan: Can we ensure following symbols are escapes please (ideally unit test to satisfy these scenarios):

* $
* ../
* \
* '
* <
* >

trade-finance-manager-ui/server/controllers/case/index.js
```diff
@@ -449,7 +449,7 @@ const confirmTfmFacility = async (req, res) => {
       api
         .getParty(urn, userToken)
         // Non-existent party urn
-        .then((company) => (!company || !company.data ? Promise.resolve(false) : Promise.resolve(true))));
+        .then((company) => (!company || !company.data || company.status !== 200 ? Promise.resolve(false) : Promise.resolve(true))));
```
abhi-markan, line 452: It should be `Promise.reject()` instead of `Promise.resolve(false)`, thoughts?

abhi-markan: Optional chaining would be good have.

trade-finance-manager-ui/server/api.js
```diff
@@ -221,11 +228,17 @@ const login = async (username, password) => {
   }
 };
 
-const updateUserPassword = async (id, update, token) => {
+const updateUserPassword = async (userId, update, token) => {
   try {
+    const isValidUserId = isValidMongoId(userId);
+
+    if (!isValidUserId) {
+      return { status: 400, data: 'Invalid user id' };
```
abhi-markan, line 236: Are console.errors required here and below non-happy paths?

trade-finance-manager-ui/server/controllers/case/parties/index.js
```diff
@@ -146,17 +144,15 @@ const getPartyUrnDetails = async (req, res) => {
...
         dealId,
         party,
         urn: partyUrn,
       });
     }
 
-    const name = company.data.length
-      ? company.data[0].name
-      : '';
+    const name = company.data.length ? company.data[0].name : '';
```
abhi-markan, line 155: Optional chaining please.

trade-finance-manager-ui/server/controllers/case/parties/index.js
```diff
@@ -211,9 +207,11 @@ const getBondUrnDetails = async (req, res) => {
...
     }
 
-    const companies = partyUrns.map((urn) => api.getParty(urn, userToken)
-      // Non-existent party urn
-      .then((company) => (!company?.data?.length ? Promise.resolve() : Promise.resolve(company.data[0].name))));
+    const companies = partyUrns.map((urn) =>
+      api
+        .getParty(urn, userToken)
+        // Non-existent party urn
+        .then((company) => (!company?.data?.length ? Promise.resolve() : Promise.resolve(company.data[0].name))));
```
abhi-markan, line 214: Promise.reject() for negative response please.

trade-finance-manager-ui/server/helpers/validateIds.js
```diff
@@ -0,0 +1,14 @@
+const validator = require('validator');
+
+const isValidMongoId = (mongoId) => (mongoId ? validator.isMongoId(mongoId) : false);
+
+const isValidPartyUrn = (partyUrn) => {
+  const regex = /^\d{8}$/;
```
abhi-markan, line 6: * Function documentation.
* Regex from constants please.

trade-finance-manager-ui/server/api-ssrf.test.js
```diff
@@ -0,0 +1,537 @@
...
+
+      expect(response.data).toEqual(mockResponse);
+    });
+  });
+
+  describe('createFacilityAmendment', () => {
+    const mockResponse = 'Mock amendment';
+    beforeAll(() => {
+      mockAxios.reset();
+      const url = /^.*\/v1\/facilities\/.*\/amendments$/;
```
abhi-markan, line 82: Global variables with repeating valies please.

trade-finance-manager-ui/server/api-ssrf.test.js
```diff
@@ -0,0 +1,537 @@
...
+      mockAxios.onPost(url).reply(200, mockResponse);
+    });
+
+    it('Returns an error when a url traversal is supplied', async () => {
+      const urlTraversal = '../../../etc/stealpassword';
+      const expectedResponse = { status: 400, data: 'Invalid facility id' };
+
+      const response = await api.createFacilityAmendment(urlTraversal, 'mock token');
```
abhi-markan, line 90: These hard coded credentials are test data only, happy for you to mark them all as `Uses in tests` @ChristianMcc

abhi-markan (changes requested):

trade-finance-manager-ui/server/controllers/case/index.js
ChristianMcc: Will add the chaining, I only didn't change to promise.reject as it's existing behaviour I didn't want to impact. Happy to move to promise.reject

trade-finance-manager-api/api-tests/api-ssrf.api-test.js
ChristianMcc: As I'm completely mocking responses here is there much we gain from having a valid deal ID but doesn't exist? I'd just be changing the mockResponse in `mockAxios.onGet(url).reply(200, mockResponse);` and then asserting that it equals the mock response later on so it feels like the same test as having a valid ID

ChristianMcc (commented):
Good comments - have replied to any that I have questions about will address the rest!

trade-finance-manager-api/src/v1/validation/validateIds.js
ChristianMcc: This function escapes `\` `/` `<` `"` `'` which I think should cover all attacks in SSRF apart from trying to insert localhost, although none of our endpoints using this username in an endpoint are going to be vulnerable to a `localhost` or local IP attack

ChristianMcc: (In my next commit will have a test for all these characters)

trade-finance-manager-api/src/v1/api.js
abhi-markan: Missing above. Thanks

trade-finance-manager-api/src/v1/api.js
abhi-markan: Missing above, thanks.

trade-finance-manager-ui/server/api.js
```diff
@@ -461,6 +578,13 @@ const getLatestCompletedAmendmentByDealId = async (dealId, token) => {
 
 const getParty = async (partyUrn, token) => {
   try {
+    const isValidUrn = isValidPartyUrn(partyUrn);
+
+    if (!isValidUrn) {
+      console.error('getParty: Invalid party urn provided: %s', dealId);
```
abhi-markan, line 584: `dealId` is not defined, did you mean `partyUrn`?

trade-finance-manager-ui/server/controllers/case/index.js
```diff
@@ -449,7 +449,7 @@ const confirmTfmFacility = async (req, res) => {
       api
         .getParty(urn, userToken)
         // Non-existent party urn
-        .then((company) => (!company || !company.data ? Promise.resolve(false) : Promise.resolve(true))));
+        .then((company) => (!company?.data || company?.status !== 200 ? Promise.reject(false) : Promise.resolve(true))));
```
abhi-markan, line 452: Promise.reject() cannot be false have to be an error.
Does the calling function rely on `false` response?

trade-finance-manager-ui/server/helpers/validateIds.js
```diff
@@ -0,0 +1,32 @@
+const validator = require('validator');
+const regexes = require('../../constants/regex');
```
abhi-markan, line 2: Few path resolution errors from GitHub.

abhi-markan (changes requested):
Looks good, few minor comments please.

trade-finance-manager-api/api-tests/api-ssrf.api-test.js
abhi-markan: True, happy to leave it as it is thank you for the explanation.

trade-finance-manager-ui/server/controllers/case/index.js
abhi-markan: In that case if the calling functions rely on `false` respone then happy to leave it as `Promie.resolve(false)`. Thanks

trade-finance-manager-ui/server/controllers/case/index.js
ChristianMcc: Yep there is a boolean check later on, thought I might be able to pass in false that would get wrapped by an error but you're right - can I leave as resolve(false) for now? This is used less as a rejected error and more as a if the company does or doesn't exist right nowu

abhi-markan (approved):

PR #1840, opened 2023-07-12
fix(tfm): array validation

abhi-markan:
## Introduction
An `Unauthorised` error message does satisfy `.length > 0` condition given `String` type in nature.

## Resolution
* Added `Array.isArray()` validation to compound condition for both `deals` and `facilities`.
* Added **temporary** console info for `dev` debug only.
* Replaced `%O` with `%S`


Zainzzkk (approved):

ChristianMcc:
LGTM, is it worth checking the API side that we have the correct user in pasport etc. and it's not there that's the issue before deploying?

ChristianMcc (approved):

PR #1843, opened 2023-07-12
feat(FN-424): add tfm-api to Bicep

OxonianCambion:
### Introduction

This card covers
* Represent the trade-finance-manager-api webapp infrastructure in Bicep
### Resolution
* trade-finance-manager-api webapp added


infrastructure/resource_group_level/modules/trade-finance-manager-api.bicep
```diff
@@ -0,0 +1,247 @@
...
+param secureSettings object = {
+
+}
+
+// These values are taken from an export of Configuration on Dev (& validating with staging).
+@secure()
+param additionalSecureSettings object = {
+  DOCKER_REGISTRY_SERVER_PASSWORD: 'test-value'
+  UKEF_INTERNAL_NOTIFICATION: 'test-value'
+  // TODO:FN-424 All of these were found on dev but not staging, remove if appropriate
```
OxonianCambion, line 28: @abhi-markan I'm not sure what's gone on here. I think I should be able to remove them, but thought I'd check first.
How did they get on the DEV environment?

infrastructure/resource_group_level/modules/trade-finance-manager-api.bicep
```diff
@@ -0,0 +1,247 @@
...
+// These values are taken from GitHub secrets injected in the GHA Action
+@secure()
+param secureConnectionStrings object = {
+  // NOTE that CORS_ORIGIN is not present in the variables exported from dev or staging
+  CORS_ORIGIN: 'test-value'
+  UKEF_TFM_API_SYSTEM_KEY: 'test-value'
+  UKEF_TFM_API_REPORTS_KEY: 'test-value'
+  // TODO:FN-429 Note that this has a value like https://tfs-dev-tfm-fd.azurefd.net 
+  // while in the CLI it is injected as a secret, we can probably calculate it from the Front Door component.
+  TFM_URI: 'test-value'
```
OxonianCambion, line 45: Again, I've put a TODO as I don't think this should be a secret and should just be generated from the Front-Door deployment. (`TFM_URI` isn't used as an input to the Front-Door component.)

infrastructure/resource_group_level/modules/trade-finance-manager-api.bicep
```diff
@@ -0,0 +1,247 @@
...
+  AZURE_NUMBER_GENERATOR_FUNCTION_SCHEDULE: 'test-value'
+  JWT_SIGNING_KEY: 'test-value' // NOTE - in the export this appears to be a slot setting.
```
OxonianCambion, line 47: I don't think slot settings have any meaning for us as we use slots in a somewhat unusual way:
* cloning the deployment slot into a new one, including settings
* deploying to the new one
* swapping
* deleting the non-deployment slot.

This means that the settings will be identical, so it doesn't matter if the parameter is sticky or not.

infrastructure/resource_group_level/modules/trade-finance-manager-api.bicep
```diff
@@ -0,0 +1,247 @@
...
+  DOCKER_ENABLE_CI: 'true'
+  DOCKER_REGISTRY_SERVER_URL: '${containerRegistryName}.azurecr.io'
+  DOCKER_REGISTRY_SERVER_USERNAME: dockerRegistryServerUsername
+  LOG4J_FORMAT_MSG_NO_LOOKUPS: 'true'
+  TZ: 'Europe/London'
+  WEBSITE_HTTPLOGGING_RETENTION_DAYS: '3'
+  WEBSITES_ENABLE_APP_SERVICE_STORAGE: 'false'
+  
+  // All of these were found on dev but not staging
+  DTFS_CENTRAL_API_URL: 'TODO:FN-424' // https://tfs-dev-dtfs-central-api.azurewebsites.net/
```
OxonianCambion, line 84: Again, I think I can remove all these. I do want to check and understand what has gone on though!

infrastructure/resource_group_level/modules/trade-finance-manager-api.bicep
```diff
@@ -0,0 +1,247 @@
...
+  name: storageAccountName
+}
+var storageAccountKey = storageAccount.listKeys().keys[0].value
+
+// Note that in the CLI script, http was used, but the value in the exported config was https.
+var externalApiUrl = 'https://${externalApiHostname}'
+
+var dtfsCentralApiUrl = 'http://${dtfsCentralApiHostname}'
+
+var connectionStringsCalculated = {
```
OxonianCambion, line 124: Again, I think it is strange to have all these values as connection strings. We can change that after we've taken over the environments though.

infrastructure/resource_group_level/modules/trade-finance-manager-api.bicep
OxonianCambion: They will become necessary when the APIM work is merged in. I'll update the code.

infrastructure/resource_group_level/modules/trade-finance-manager-api.bicep
```diff
@@ -0,0 +1,251 @@
...
+var settings = {
+  WEBSITE_DNS_SERVER: azureDnsServerIp
+  WEBSITE_VNET_ROUTE_ALL: '1'
+  PORT: '5000'
+  WEBSITES_PORT: '5000'
+}
+
+// These values are taken from an export of Configuration on Dev (& validating with staging).
+var dtfsCentralApiUrl = 'http://${dtfsCentralApiHostname}'
+// Note that in the CLI script, http was used, but the value in the exported config was https.
```
IgnatG, line 72: in reality, all URLs should be HTTPs. can we change Central-API to follow the same approach?

infrastructure/resource_group_level/modules/trade-finance-manager-api.bicep
IgnatG: agreed

infrastructure/resource_group_level/modules/trade-finance-manager-api.bicep
```diff
@@ -0,0 +1,251 @@
...
+
+resource tfmApi 'Microsoft.Web/sites@2022-09-01' = {
+  name: tfmApiName
+  location: location
+  tags: {
+    Environment: 'Preproduction'
+  }
+  kind: 'app,linux,container'
+  properties: {
+    httpsOnly: false
```
IgnatG, line 167: we can leave it as `false` for now. but we need to look at setting it to `true` afterwards. Not sure if we need to change anything else

infrastructure/resource_group_level/modules/trade-finance-manager-api.bicep
```diff
@@ -0,0 +1,251 @@
...
+}
+
+resource tfmApi 'Microsoft.Web/sites@2022-09-01' = {
+  name: tfmApiName
+  location: location
+  tags: {
+    Environment: 'Preproduction'
+  }
+  kind: 'app,linux,container'
+  properties: {
```
IgnatG, line 166: one item that I realised now: can we add a TODO and a ticket in the backlog that focuses on loading a lot of these common settings from Github variables? this because if one of them changes, (i.e. scmMinTlsVersion is changed to v `1.2` or something like that), then we have to update the codebase for it. whereas if it's coming from a variable, then we only change it there.

infrastructure/resource_group_level/modules/trade-finance-manager-api.bicep
OxonianCambion: ok - I will not make any settings slot specific.

infrastructure/resource_group_level/modules/trade-finance-manager-api.bicep
OxonianCambion: I assume so! @abhi-markan is looking into this. I've created FN-741 to track it.

infrastructure/resource_group_level/modules/trade-finance-manager-api.bicep
OxonianCambion: Apart from making sure clients are updated to use https, I don't think there's anything else. I'm mentioned both steps in FN-741.

infrastructure/resource_group_level/modules/trade-finance-manager-api.bicep
OxonianCambion: Agreed - I had mentally put that task as part of FN-712, but it might be slightly wider in scope. I'll make it clearer in FN-712 - which is still a good place for it - and create a new card as a check & for any additional settings that might be best commonised.

Note that I think it may be better for internal to just have them as a `val` in `main.bicep` rather than a parameter that is set from a GitHub environment variable - even if we give that parameter a default value. (e.g. for `scmMinTlsVersion` I think it would just add an extra layer of indirection for no clear advantage. Seeing as deploying new code will be trivial there's no barrier there and it doesn't seem a value that is useful to inject from GitHub)

Note that I didn't want to do it as I don't know what variation there is in the existing entities:
I think they _should_ be the same, but not all are. e.g. `scmMinTlsVersion` _is_ `1.2` for external API but not for the others. (I suspect what happened was HTTPS access was added for External-API after a security audit, but the changes weren't made for the other services.)

IgnatG (approved):

PR #1847, opened 2023-07-14
fix(SR-9): Add csrf checks to missing Gef and Portal routes

JennyHumphries:
## Introduction
This [warning ](https://github.com/UK-Export-Finance/dtfs2/security/code-scanning?query=is%3Aopen+branch%3Amain+CSRF)shows that some of the endpoints in GEF and Portal are missing CSRF protection.

There were 4 endpoints affected by this in gef:
* Get supporting documents page
* Post supporting documents form
* Upload supporting document
* Delete supporting document

There were 8 endpoints affected by this in portal:
* Eligibility Criteria form - Get, Post and Post with Save and go back
* Eligibility Supporting Documents form - Get, Post and Post with Save and go back
* Eligibility Check your answers - Get and download document

### Gef
* The get and post supporting documents form endpoints work fine when moved behind the existing csrf protection. So I have just done that.
* The upload and delete endpoints were trickier because these post requests are made from within the MOJ component so we are unable to add the csrf token to the request body.
* So instead I have made a separate csrf token called `uploadCsrf`:
    * It's created when the supporting document page is loaded
    * It has a lifetime of 15 minutes before it expires
* When the upload fails,  an error message is shown:
sr-9-error-demo
* When the delete fails, an error message is shown:
sr-9-error-on-delete

### Portal
* Only the Eligibility Supporting Documents form routes failed when I moved the routes behind the existing csrf checks.
* This was because this form has the encoding `enctype="multipart/form-data"`. Which seems to ignore any hidden fields, so `_csrf` was not being added to any of the post request bodies.
* I couldn't change the `enctype` because it's required if you want to allow file uploads in your form.
* So instead I've passed `_csrf` into the query and added a middleware that copies that into they body. Then the form is able to use the same csrf checks as the rest of portal.

## Miscellaneous
* While working on this I found that the webpack builds for `gef-ui` and `portal` did not work.
* It was because I needed to run `npm install` for those projects. But `npm install` also did not work because the package `optimize-css-assets-webpack-plugin` was not compatible with webpack version 5.
* The plugin suggested replacing it with `css-minimizer-webpack-plugin`. Which I have done.
* This change meant that there were no package dependency issues for gef-ui and portal so I was able to remove the `--legacy-peer-deps` flags from the project `install:all` and `ci:all` scripts.

## Automated Checks
* SonarCloud is showing 1 Security Hotspot to check that the content length is safe in `supporting-information-upload.js`. I think this is covered by one of the other SR tickets so I don't plan to do anything for it here.
* SonarCloud is also showing 1 bug because it incorrectly thinks that `new MOJFrontend.MultiFileUpload` in `multi-file-upload.js` isn't being used. I don't think I need to take any action here - let me know if that's wrong!


gef-ui/public/js/fileUpload.js
JennyHumphries, line 1: This has now been moved into the nunjucks file so that we're able to pass in the csrf token

gef-ui/server/utils/helpers.test.js
JennyHumphries, line 1: The tests for the last few methods in this file were incorrectly nested inside the describe for `futureDateInRange` so I've moved them outside.

I haven't changed any of the content of those tests.

The new tests I've added are all in the `describe('getCurrentTimePlusMinutes', () => {` section

gef-ui/templates/partials/manual-inclusion-questionnaire.njk
JennyHumphries, line 1: I don't think this was used anyway. I think it might have been left over when all the separate document pages were combined into `upload-supporting-documents.njk`

gef-ui/templates/partials/upload-supporting-documents.njk
```diff
@@ -108,5 +108,21 @@
 {% endblock %}
 
 {% block scripts %}
-  <script src="/assets/js/fileUpload.js"></script>
+  <script>
+    if (typeof MOJFrontend.MultiFileUpload !== 'undefined') {
+      const onFileDelete = (mojFileUpload, response) => {
```
JennyHumphries, line 113: This was the only way I could find to show an error message when the delete request fails. There didn't seem to be a nicer hook in.

ChristianMcc (commented):
Looking good to me! Few comments that require some other opinions.
Webpack changes look good to me - just need to be careful when deploying to check nothing goes wrong there

gef-ui/server/index.js
```diff
@@ -82,7 +82,11 @@ app.set('trustproxy', true);
 app.use(session(sessionOptions));
 app.use(cookieParser());
 app.use(express.urlencoded({ extended: true }));
-app.use('/', supportingDocuments);
+
+// These routes cannot use the csrf check below so must come before it
+// They implement their own separate csrf check
+app.use('/', supportingInformationUploadRoutes);
```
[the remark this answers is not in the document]
JennyHumphries: Thanks, I've marked it as false positive

gef-ui/templates/partials/upload-supporting-documents.njk
JennyHumphries: I've changed it so that when a delete request succeeds, all of the failed delete messages disappear.

I've had a play around with that and I think it all feels intuitive.

gef-ui/server/index.js
```diff
@@ -82,7 +82,23 @@ app.set('trustproxy', true);
...
+
+// For endpoints requiring file uploads, the csrf token can only be sent as part of the query rather than in the body.
+// This is because these forms must use enctype="multipart/form-data"
+// The csrf middleware below expects to find the token in the body, so we copy it from the query to the body here.
+const copyCsrfTokenFromQueryToBody = (req, res, next) => {
+  if (req.query._csrf && !req.body._csrf) {
```
abhi-markan, line 94: Optional chaining please.

gef-ui/server/index.js
```diff
@@ -82,7 +82,23 @@ app.set('trustproxy', true);
...
-app.use('/', supportingDocuments);
+
+// These routes cannot use the csrf check below so must come before it
+// They implement their own separate csrf check
+app.use('/', supportingInformationUploadRoutes);
+
+// For endpoints requiring file uploads, the csrf token can only be sent as part of the query rather than in the body.
+// This is because these forms must use enctype="multipart/form-data"
+// The csrf middleware below expects to find the token in the body, so we copy it from the query to the body here.
+const copyCsrfTokenFromQueryToBody = (req, res, next) => {
```
abhi-markan, line 93: Please move this middleware to `middleware` directory as a function.

gef-ui/server/utils/helpers.js
```diff
@@ -514,6 +514,8 @@ const futureDateInRange = ({ day, month, year }, days) => {
   return false;
 };
 
+const getCurrentTimePlusMinutes = (minutesToAdd) => new Date(new Date().getTime() + minutesToAdd * 60 * 1000);
```
abhi-markan, line 517: Can we please add `minutesToAdd` void check.

gef-ui/server/utils/helpers.test.js
```diff
@@ -1444,132 +1445,158 @@ describe('futureDateInRange', () => {
...
-    it('Should return expected date object for mock facility', () => {
-      const expected = {
-        date: '2',
-        month: '12',
-        year: '2021',
-      };
-      expect(getFacilityCoverStartDate(MOCK_FACILITY.items[1])).toEqual(expected);
-    });
+describe('getCurrentTimePlusMinutes', () => {
+  it('should add the specified number of minutes to the current time', () => {
```
abhi-markan, line 1451: Please add an assertion with an empty input.

gef-ui/server/routes/supporting-information-upload.js
```diff
@@ -0,0 +1,50 @@
...
+// eslint-disable-next-line consistent-return
+const validateUploadCsrfToken = (req, res, next) => {
+  if (req.session.uploadCsrf
+    && req.session.uploadCsrf.token === req.query.uploadCsrf
+    && new Date() < new Date(req.session.uploadCsrf.expiry)) {
+    next();
+  } else {
+    // MOJ multi-file-upload expects a 200 response when the request is not valid
+    // It will only display the error messages when the response is 200.
+    return res.status(200).send({ error: { message: 'Error. Please refresh your browser and try again' } });
```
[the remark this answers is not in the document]
abhi-markan: Can the error message please be updated to `File upload session expired. Please refresh your browser to upload or delete the files.`

gef-ui/server/routes/supporting-information-upload.js
```diff
@@ -0,0 +1,50 @@
...
+  uploadSupportingDocument, deleteSupportingDocument,
+} = require('../controllers/supporting-information/supporting-documents');
+const { validateRole, validateToken, validateBank } = require('../middleware');
+
+// The following routes cannot use the same csrf checks as the rest of the routes
+// as the Ministry of Justice multi-file-upload component does not allow passing tokens to the request body.
+// So we instead make a separate uploadCsrf token and check it here.
+const router = express.Router();
+
+// eslint-disable-next-line consistent-return
```
abhi-markan, line 14: Please remove inline Eslint rule.

abhi-markan: Please add function documentation.

gef-ui/server/routes/supporting-information-upload.js
```diff
@@ -0,0 +1,50 @@
...
+const validateUploadCsrfToken = (req, res, next) => {
+  if (req.session.uploadCsrf
+    && req.session.uploadCsrf.token === req.query.uploadCsrf
+    && new Date() < new Date(req.session.uploadCsrf.expiry)) {
```
abhi-markan, line 18: `<=`, any thoughts?

gef-ui/server/routes/supporting-information-upload.js
```diff
@@ -0,0 +1,50 @@
...
+    next();
+  } else {
+    // MOJ multi-file-upload expects a 200 response when the request is not valid
+    // It will only display the error messages when the response is 200.
+    return res.status(200).send({ error: { message: 'Error. Please refresh your browser and try again' } });
+  }
+};
+
+const validateFiles = multer({ fileFilter: multerFilter }).single('documents');
+const validateUploadRequest = (req, res, next) => {
```
abhi-markan, line 28: * New line before function declaration.
* Function documentation.

gef-ui/server/routes/supporting-information-upload.js
```diff
@@ -0,0 +1,50 @@
...
+  // eslint-disable-next-line consistent-return
```
abhi-markan, line 29: Please remove inline rule.

gef-ui/templates/partials/upload-supporting-documents.njk
abhi-markan: This <script> needs to be moved to an external script file, CSP violation will be flagged on console.

gef-ui/templates/partials/upload-supporting-documents.njk
```diff
@@ -108,5 +109,23 @@
...
+        if (response.error) {
+          const errorRow = `<div class="govuk-summary-list__row moj-multi-file-upload__row moj-multi-file-upload__error-row"><div class="govuk-summary-list__value moj-multi-file-upload__message">${mojFileUpload.getErrorHtml(response.error)}</div></div>`;
```
abhi-markan, line 116: Is there no other cleaner way of presenting an error message back to the component?

gef-ui/server/index.js
abhi-markan: This middleware has been added as Global middleware therefore effecting all the routes, has this caused other routes to fail?

portal/server/index.js
```diff
@@ -89,7 +88,18 @@ configureNunjucks({
 app.use(express.json());
 app.use(express.urlencoded({ extended: true }));
 app.use(cookieParser());
-app.use('/', eligibilityRoutes);
+
+// For endpoints requiring file uploads, the csrf token can only be sent as part of the query rather than in the body.
+// This is because these forms must use enctype="multipart/form-data"
+// The csrf middleware below expects to find the token in the body, so we copy it from the query to the body here.
+const copyCsrfTokenFromQueryToBody = (req, res, next) => {
```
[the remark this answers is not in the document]
abhi-markan: Has this caused other routes to fail?

abhi-markan (changes requested):

gef-ui/templates/partials/upload-supporting-documents.njk
JennyHumphries: I couldn't find a better way of doing it.

The delete method in `mojFileUpload` didn't seem to handle showing error messages, so the best way I could find to hook in was in this `fileDeleteHook`.

And then `mojFileUpload.getErrorHtml` was the biggest bit of the other error messages HTML I could get directly from `mojFileUpload`. So I've added the  other divs and classes are so it matches the other rows.

portal/server/index.js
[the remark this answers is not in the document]
JennyHumphries: Not as far as I can tell.

I think it should be safe as the change is only made when there's _csrf in the query, and not one already in the body.

gef-ui/server/index.js
JennyHumphries: From what I can tell, I don't think that it has caused any of the other routes to fail.
 The e2e tests all still seem to pass too so I think it should be safe

gef-ui/server/routes/supporting-information-upload.js
ChristianMcc: can you do return next() on line 27 to fix this?

JennyHumphries: Yes! I'll do that for both of them, thanks.

gef-ui/server/routes/supporting-information-upload.js
Zainzzkk: We could use date-fns' `isBefore` function for this too?

ChristianMcc: When doing comparisons between two date objects javascript automatically converts them to their epoch time behind the scenes to do the comparison so it's doing this for us for free :)

ChristianMcc (approved):

gef-ui/templates/partials/upload-supporting-documents.njk
```diff
@@ -108,5 +109,5 @@
 {% endblock %}
 
 {% block scripts %}
-  <script src="/assets/js/fileUpload.js"></script>
+  <script id="multi-file-upload" src="/assets/js/multiFileUpload.js" uploadCsrf="{{ uploadCsrf }}" ></script>
```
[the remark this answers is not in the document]
JennyHumphries: I'm not quite sure what you mean sorry.

The attribute is so that `uploadCsrf` can be accessed from the multiFileUpload script. Is there a better way of passing it in from here?

portal/scripts/multi-file-upload.js
```diff
@@ -0,0 +1,21 @@
+/* eslint-disable no-new */
```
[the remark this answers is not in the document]
abhi-markan: Have removed on a peer call.

gef-ui/templates/partials/upload-supporting-documents.njk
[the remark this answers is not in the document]
abhi-markan: Have updated on a peer call.

portal/scripts/multi-file-upload.js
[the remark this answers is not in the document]
ChristianMcc: I bumped into this on my own work with the upload component so replying here:
Removing these will mean linting fails and there's no easy way round this without deviating from recommended use of this component in: https://design-patterns.service.justice.gov.uk/components/multi-file-upload/.
I think it should be left in so linting doesn't fail and at worst at some point move away from this component if it has multiple issues right now (e.g. csrf and this) but that's definitely for another day!

ChristianMcc (approved):

abhi-markan (approved):

PR #1845, opened 2023-07-13
feat: migrated the `/customers` APIs to APIM

IgnatG:
### Changelog
- Migrated PartyDB/Customers API to APIM
- Remove/replaced all Mulesoft references

### TODO:
- Merge the branch against `main` branch
- Update the env variables in Github and Azure

PR #1851, opened 2023-07-17
Feat/FN-737/update env vars

OxonianCambion:
### Introduction

This card covers 
* Removing environment variables that are no longer used on head of `main`.
* marking environment variables that will soon go
* making the parameters sent to modules self consistent.
### Resolution
The changes:
* remove the unnecessary environment variables
* mark some others as to inspect later with an associated JIRA card.
* prevent an circular dependency implied by the dev environment export by establishing which ones were not necessary
* drive by fix of an inconsistent environment variable


.github/workflows/development_infrastructure.yml
```diff
@@ -1158,9 +1158,6 @@ jobs:
                 APIM_TFS_URL=${{ secrets.APIM_TFS_URL }} \
                 APIM_TFS_KEY=${{ secrets.APIM_TFS_KEY }} \
                 APIM_TFS_VALUE=${{ secrets.APIM_TFS_VALUE }} \
-                MULESOFT_API_PARTY_DB_KEY="${{ secrets.MULESOFT_API_PARTY_DB_KEY}}" \
```
OxonianCambion, line 1161: I think these were missed in https://github.com/UK-Export-Finance/dtfs2/pull/1845

.github/workflows/development_infrastructure.yml
```diff
@@ -1170,7 +1167,7 @@ jobs:
                 COMPANIES_HOUSE_API_URL="${{ env.COMPANIES_HOUSE_API_URL }}" \
                 COMPANIES_HOUSE_API_KEY="${{ env.COMPANIES_HOUSE_API_KEY }}" \
                 ORDNANCE_SURVEY_API_URL="${{ env.ORDNANCE_SURVEY_API_URL }}" \
-                GOV_NOTIFY_EMAIL_RECIPIENT="${{ env.GOV_NOTIFY_EMAIL_RECIPIENT }}" \
+                GOV_NOTIFY_EMAIL_RECIPIENT="${{ secrets.GOV_NOTIFY_EMAIL_RECIPIENT }}" \
```
OxonianCambion, line 1170: I've made this consistent with the other usage. Note that VSCode did complain of using via `env` so I'm sure this is now correct.

infrastructure/resource_group_level/modules/portal-api.bicep
```diff
@@ -26,14 +27,19 @@ param additionalSecureSettings object = {
...
+
 // These values are taken from GitHub secrets injected in the GHA Action
 @secure()
 param secureConnectionStrings object = {
   // NOTE that CORS_ORIGIN is not present in the variables exported from dev or staging
   CORS_ORIGIN: 'test-value'
   AZURE_PORTAL_EXPORT_FOLDER: 'test-value'
   AZURE_PORTAL_FILESHARE_NAME: 'test-value'
-  // TODO:FN-737 *_WORKFLOW_* variables are not needed and can be removed.
+  // TODO:FN-749 Confirm *_WORKFLOW_* variables are not needed and can be removed.
```
IgnatG, line 42: they are definitely not used :) 
`Workflow` was an old system that we were supposed to integrate with. this was replaced with TFM

infrastructure/resource_group_level/modules/portal-api.bicep
```diff
@@ -44,9 +50,8 @@ param secureConnectionStrings object = {
   GOV_NOTIFY_API_KEY: 'test-value'
   GOV_NOTIFY_EMAIL_RECIPIENT: 'test-value'
   DTFS_PORTAL_SCHEDULER: 'test-value'
-  // TODO:FN-737 *_WORKFLOW_* variables are not needed and can be removed.
+  // TODO:FN-749 Confirm *_WORKFLOW_* variables are not needed and can be removed.
```
IgnatG, line 53: same as above

infrastructure/resource_group_level/modules/portal-api.bicep
OxonianCambion: Good stuff - I've updated FN-749 with this comment and will strip them in that so we're consistent with the changes on `main`.

PR #1856, opened 2023-07-18
Feat(FN-426): Add gef-ui to bicep

OxonianCambion:
### Introduction

This card covers:
* Adding the gef-ui to bicep
### Resolution
* gef-ui bicep file added


infrastructure/resource_group_level/modules/gef-ui.bicep
```diff
@@ -0,0 +1,219 @@
...
+// These values are hardcoded in the CLI scripts, derived in the script or set from normal env variables
+var settings = {
+  // from env.
+  COMPANIES_HOUSE_API_URL: 'test-value'
+
+  // derived
+  PORTAL_API_URL: portalApiUrl
+  REDIS_HOSTNAME: redis.properties.hostName
+  REDIS_PORT: redis.properties.sslPort
+  // TODO:FN-425 check if this leaks the REDIS_KEY value?
```
OxonianCambion, line 66: Note that it was so similar to portal-ui, I just coped it.
The only difference was for some reason `gef-ui` doesn't have `WEBSITE_HEALTHCHECK_MAXPINGFAILURES` set.

infrastructure/resource_group_level/modules/gef-ui.bicep
```diff
@@ -0,0 +1,219 @@
...
+  DTFS_CENTRAL_API_KEY: 'test-value'
+  EXTERNAL_API_KEY: 'test-value'
+  PORTAL_API_KEY: 'test-value'
+  TFM_API_KEY: 'test-value'
+}
+
+// These values are taken from GitHub secrets injected in the GHA Action
+@secure()
+param secureConnectionStrings object = {
+  COMPANIES_HOUSE_API_KEY : 'test-value'
```
IgnatG, line 35: interesting. we need to check if this is actually used because GEF-UI shouldn't call the Companies House API directly

OxonianCambion: I've marked for removal in FN-820

PR #1857, opened 2023-07-19
fix(mock-data-loader): Fix TFM failing on mock-data-loader

Zainzzkk:
## Introduction
Mock data loader failing for tfm with 401

## Resolution
* re-insert-mocks user not deleted until all tfm mocks inserted
* New function to delete all re-inserted mocks users from tfm

### DIFF
--- a/utils/mock-data-loader/re-insert-mocks.js
+++ b/utils/mock-data-loader/re-insert-mocks.js
@@ -1,4 +1,4 @@
-const { createAndLogInAsInitialUser, deleteInitialUser } = require('./user-helper');
+const { createAndLogInAsInitialUser, deleteInitialUser, deleteInitialTFMUser } = require('./user-helper');
 
 const cleanAllTables = require('./clean-all-tables');
 const insertMocks = require('./insert-mocks');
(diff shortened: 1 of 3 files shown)


abhi-markan (approved):

utils/mock-data-loader/re-insert-mocks.js
```diff
@@ -29,8 +28,12 @@ const init = async () => {
     roles: ['data-admin'],
     email: 're-insert-mocks-data-loader-tfm@ukexportfinance.gov.uk',
   });
+
+  await cleanAllTablesTfm(tfmToken);
```
ChristianMcc, line 32: When I looked nothing in here actually touches TFM (it goes to central) so doesn't need the token, has that changed? Or are we just future proofing ourselves here?

ChristianMcc (commented):
I'm happy with the changes but this shouldn't have actually fixed any 401s - cleanAllTablesTfm never touched TFM API, it was only insertMocksTfm which created a new user after we cleaned all the tables. So any 401s you were getting would be for a different reason

utils/mock-data-loader/re-insert-mocks.js
Zainzzkk: refactored it for those changes

PR #1860, opened 2023-07-19
fix(SR-12): Stop exposing full error messages and stack traces

JennyHumphries:
## Introduction
This PR is to fix the [Information exposure through a stack trace](https://github.com/UK-Export-Finance/dtfs2/security/code-scanning?query=is%3Aopen+branch%3Amain+stack+trace++) alerts

## Resolution
Where errors are returned from the api methods, instead of returning them, log them and return an error message instead.

### DIFF
--- a/dtfs-central-api/src/v1/controllers/cron-jobs/cron-jobs.controller.js
+++ b/dtfs-central-api/src/v1/controllers/cron-jobs/cron-jobs.controller.js
@@ -8,6 +8,8 @@ exports.deleteAllEstoreLogs = async (req, res) => {
 
     return res.status(200).send();
   } catch (error) {
-    return res.status(500).send(error);
+    return res.status(500).send({
+      error: 'An exception has occured',
+    });
   }
 };
--- a/dtfs-central-api/src/v1/controllers/cron-jobs/cron-jobs.controller.js
+++ b/dtfs-central-api/src/v1/controllers/cron-jobs/cron-jobs.controller.js
@@ -8,6 +8,8 @@ exports.deleteAllEstoreLogs = async (req, res) => {
 
     return res.status(200).send();
   } catch (error) {
+    console.error('CRON job error %O', error);
+
     return res.status(500).send({
       error: 'An exception has occured',
     });
--- a/dtfs-central-api/src/v1/controllers/durable-functions/durable-functions.controller.js
+++ b/dtfs-central-api/src/v1/controllers/durable-functions/durable-functions.controller.js
@@ -8,6 +8,8 @@ exports.deleteAllDurableFunctions = async (req, res) => {
 
     return res.status(200).send();
   } catch (error) {
+    console.error('ACBS DOF error %O', error);
+
     return res.status(500).send({
       error: 'An exception has occured',
     });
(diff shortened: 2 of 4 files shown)


trade-finance-manager-api/src/v1/api.js
```diff
@@ -214,7 +214,8 @@ const updateDeal = async (dealId, dealUpdate) => {
 
     return response.data;
   } catch (error) {
-    return error;
```
JennyHumphries, line 217: There seem to be quite a few other method in this file that have
```
catch(error) {
  return error;
}
```
But that haven't been flagged by the code scanning alerts.

Should I go through and change them to follow this pattern anyway? Or are we safe to assume that the code scanning should have picked up any that we need to worry about?

ChristianMcc: Can you check a couple of these and just make sure that it isn't then returned by the controller that has called this api.js file please? And if it's not then we should be happy to assume the code scanning is doing it's job properly :)

dtfs-central-api/src/v1/controllers/cron-jobs/cron-jobs.controller.js
```diff
@@ -8,6 +8,8 @@ exports.deleteAllEstoreLogs = async (req, res) => {
 
     return res.status(200).send();
   } catch (error) {
-    return res.status(500).send(error);
```
abhi-markan, line 11: Can we please add

```javascript
console.error('CRON job error %O', error);
```

dtfs-central-api/src/v1/controllers/durable-functions/durable-functions.controller.js
```diff
@@ -8,6 +8,8 @@ exports.deleteAllDurableFunctions = async (req, res) => {
 
     return res.status(200).send();
   } catch (error) {
-    return res.status(500).send(error);
```
abhi-markan, line 11: Can we please add

``javascript
console.error('ACBS DOF error %O', error);
```

File: portal/server/api.js [hunk 4]
```diff
@@ -729,7 +729,7 @@ const getUnissuedFacilitiesReport = async (token) => {
     return response.data;
   } catch (error) {
     console.error('Unable to return unissued facilities', { error });
```
abhi-markan, line 731: Can this please be changed to following:

```javascript
console.error('Unable to return unissued facilities %O', error);
```

portal/server/api.js
```diff
@@ -747,7 +747,7 @@ const getUkefDecisionReport = async (token, payload) => {
     return response.data;
   } catch (error) {
     console.error('Unable to return Ukef decision report', { error });
```
abhi-markan, line 749: Can this please be changed to following:

```javascript
console.error('Unable to return UKEF decision report %O', error);
```

trade-finance-manager-api/src/v1/api.js
abhi-markan: If you can go through them as well please that will be great, ideally if they can have the same `console.error` and return statement that will be great.

abhi-markan (changes requested):
Only few minor comments to address please.

PR #1862, opened 2023-07-20
test(DTFS2-5348): Re-instate cypress tests

JennyHumphries:
### Introduction
On this ticket there were two cypress tests that had been removed.

### Resolution
I've copied the tests back into the files mentioned on the ticket.
I've also split them each into two tests as they both test two different cases.
I've tested that they pass locally as well as checking they run on the PR


ChristianMcc (approved):
test is passing on main and this won't affect anything so the rer-run should hopefully just pass 👀

abhi-markan (approved):

Zainzzkk (approved):

PR #1863, opened 2023-07-20
Feat(FN-427): application gateway portal

OxonianCambion:
### Introduction

This card covers 
* Adding the Portal Application Gateway in Bicep
### Resolution
* Portal Application Gateway added in Bicep
* Network Security Groups added and linked


infrastructure/resource_group_level/application-gateway-portal.bicep
```diff
@@ -0,0 +1,287 @@
+param location string
+param environment string
+param gatewaySubnetId string
+param tfsIpId string
+param portalApiHostname string
+param portalUiHostname string
+param gefUiHostname string
+param apiPortalAccessPort int = 0
```
OxonianCambion, line 8: Note that the Api portal seems to only be for `dev`. I've adjusted this and the NSG code to only set it up if has been given a non-zero value. (e.g. `44232`.)

infrastructure/resource_group_level/application-gateway-portal.bicep
```diff
@@ -0,0 +1,287 @@
...
+
+param applicationGatewaySku object = {
+  name: 'WAF_v2'
+  tier: 'WAF_v2'
+}
+
+param autoscaleConfiguration object = {
+  minCapacity: 1
+  maxCapacity: 5
+}
```
OxonianCambion, line 18: I've made some of the values parametrisable but defaulted. There are some others in theory, but I think they are ok inlined and could easily be updated later.

PR #1866, opened 2023-07-24
fix(portal-api-exception-text): Portal-API - error return

Zainzzkk:
## Introduction
* Portal-API was returning error for certain api calls
* Husky was not working and linting was not running on commit

## Resolution
* Changed to return error message and status (rather than full error)
* changes made to let husky and linting run on commit with @abhi-markan


portal-api/src/v1/api.js
```diff
@@ -87,7 +87,7 @@ const deleteDeal = async (dealId) => {
       headers: headers.central,
     });
   } catch (error) {
-    return error;
```
abhi-markan, line 90: Please add console.error to all the return statements in following format:

`console.error('... %O', error)`;

portal-api/src/v1/gef/controllers/files.controller.js
```diff
@@ -160,7 +160,7 @@ exports.downloadFile = async (req, res) => {
     return readStream.pipe(res);
   } catch (error) {
     console.error(`Error downloading file: ${error}`);
```
abhi-markan, line 162: Update console.error to above syntax please.

abhi-markan (changes requested):
Few comments to address.

abhi-markan (approved):

PR #1879, opened 2023-07-26
fix(DTFS2-5656): Add middleware to strip _csrf token from body of requests

ChristianMcc:
## Introduction
In some places portal/tfm/central API were saving the _csrf token from payloads straight to the database

## Resolution
Introduce middleware to remove any _csrf tokens from the body of requests before processing so it's not saved to the database


abhi-markan (dismissed):
LGTM 👍🏼

abhi-markan (approved):

PR #1883, opened 2023-07-27
fix(SR-14): Estore API SSRF protection

ChristianMcc:
## Introduction
Estore API calls were missing validation on some variables used in to construct urls for external API calls to the Estore API.

## Resolution
Add in validation for siteId, exporterName and dealIdentifier to help reduce attack vectors on these variables.

NOTE: I think this should be tested and QA'd carefully on dev/staging with estore to check there are no edge cases of site IDs etc. that this breaks unintentionally


external-api/src/constants/regex.constants.ts
```diff
@@ -1,6 +1,9 @@
+/* eslint-disable no-useless-escape */
```
ChristianMcc, line 1: Added in so I can mimic the regex of the EStore API without changes

abhi-markan (approved):

PR #1940, opened 2023-08-04
feat(dtfs2-5003): TFM UI - use aliases for file paths

francescastocco:
## Introduction

There are cases when we load a module using the following format:
`require('../../../../constants');`
This could be better managed by changing it to
`require('@constants/constants');`

## Resolution

I have used [node's import mapping](https://nodejs.org/docs/latest-v18.x/api/packages.html#subpath-imports), available as standard instead of the [module-alias](https://www.npmjs.com/package/module-alias) package mentioned in the ticket. 

I turned off the following linting rules as they conflicted with javascript module aliasing:
- `import/no-extensions`
- `import/no-unresolved`
We could use [this package](https://www.npmjs.com/package/eslint-import-resolver-alias) to resolve the imports for eslint but it would mean we specify the aliases twice (in the eslint config and package.json)


trade-finance-manager-ui/package.json
```diff
@@ -17,6 +17,19 @@
   },
   "license": "MIT",
   "author": "UK Export Finance",
+  "imports": {
+    "#api": "./server/api.js",
```
abhi-markan, line 21: Can these `#` be converted to `@` to keep it consistent with other projects please.

trade-finance-manager-ui/.eslintrc.js
```diff
@@ -23,6 +23,8 @@ module.exports = {
     }],
     'no-unneeded-ternary': 'off',
     'require-await': 'error',
+    'import/extensions': 'off',
```
abhi-markan, line 26: Can these be converted to `warning` please?

abhi-markan (changes requested):
Good work, couple of comments to address please.

trade-finance-manager-ui/package.json
francescastocco: I have used `#` as this is written in the [node documentation](https://nodejs.org/docs/latest-v18.x/api/packages.html#subpath-imports)
> Entries in the "imports" field must always start with # to ensure they are disambiguated from external package specifiers.

abhi-markan: Understood thanks for the clarification.

abhi-markan (approved):

abhi-markan (changes requested):
Hey @francescastocco could you please look into Coday suggestions?

https://app.codacy.com/gh/UK-Export-Finance/dtfs2/pullRequest?prid=12348773

lmarrai-sw:
Since we're adding Typescript to all services in DTFS2-6774, we should be able to add aliases via Typescript instead now, so I'm going to close this PR without merging :)

PR #1942, opened 2023-08-04
fix(dockerignore): fixed docker ignore paths

abhi-markan:
## Introduction
Fixed docker ignore paths


azure-functions/acbs-function/.dockerignore
```diff
@@ -39,13 +39,13 @@ coverage
...
-**/report/*
+reporting/tfm/report/*
 **/generated_reports
 **/temp/*
-**/actionsheets/*
-**/json/*
-**/amendments/*
-**/exports/*
+utils/**/actionsheets/*
+utils/**/json/*
```
IgnatG, line 46: Just to flag that Docker would only try to containerise the files from its current folder. I.e. the Dockerfile from acbs-functions folder wouldn't have access to utils, TFM, reporting, etc folders

abhi-markan: Hello, yes I agree. I have added it in case we need to any exports or reports we can use these safe paths.

PR #1945, opened 2023-08-08
fix(SR8): `deleteOne` operator NoSQLi mitigation

abhi-markan:
## Introduction
MongoDB query rely on a direct user input, which is susceptible  to NoSQL injection.

## Resolution
* Introduced usage of `$eq` operator to ensure any user input is interpreted as a literal value and not as a query obect.
* Ensure user input is of type `string` and not an object.

## Miscellaneous
Updated `cspell` command execution `flags`.


lmarrai-sw (approved):
The change looks good to me @abhi-markan 

1 thing: your description says
> Ensure user input is of type string and not an object.

How are we doing that in this PR?

dtfs-central-api/src/v1/controllers/portal/deal/delete-deal.controller.js
```diff
@@ -3,11 +3,13 @@ const { findOneDeal } = require('./get-deal.controller');
...
-  findOneDeal(req.params.id, async (deal) => {
-    if (ObjectId.isValid(req.params.id)) {
+  const { id } = req.params;
+
+  findOneDeal(id, async (deal) => {
+    if (ObjectId.isValid(id)) {
       if (deal) {
         const collection = await db.getCollection('deals');
-        const status = await collection.deleteOne({ _id: ObjectId(req.params.id) });
+        const status = await collection.deleteOne({ $eq: { _id: ObjectId(id) } });
```
oscar-richardson-softwire, line 12: I think this line should be
`const status = await collection.deleteOne({ _id: { $eq: ObjectId(id) } });`

abhi-markan: Good spot, have resolved.

portal-api/src/v1/controllers/eligibilityCriteria.controller.js
```diff
@@ -69,6 +69,13 @@ exports.update = async (req, res) => {
...
-  res.status(200).send(status);
+  const { version } = req.params;
+  const versionNumber = Number(version);
+
+  if (typeof versionNumber === 'number') {
+    const status = await collection.deleteOne({ version: { $eq: versionNumber } });
+    res.status(200).send(status);
+  }
+
+  res.status(400).send('Invalid eligibility criteria version number');
```
lmarrai-sw, line 80: I think the `if (typeof versionNumber === 'number')` check will always be `true` here, because `Number(...)` always returns a number (if `version` wasn't a valid number then `Number(version)` will be `NaN` and `typeof NaN` is `number`)

Also, do we need a `return` after line 77?

lmarrai-sw (changes requested):

trade-finance-manager-api/src/v1/controllers/user/user.controller.js
```diff
@@ -95,8 +95,12 @@ exports.incrementFailedLoginCount = async (user) => {
...
-  const status = await collection.deleteOne({ _id: ObjectId(_id) });
+  if (ObjectId.isValid(_id)) {
+    const collection = await db.getCollection('tfm-users');
+    const status = await collection.deleteOne({ _id: { $eq: ObjectId(_id) } });
+
+    callback(null, status);
+  }
 
-  callback(null, status);
+  callback('Invalid TFM user id', 400);
```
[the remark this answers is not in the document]
abhi-markan: `next` will return the error to the calling route and the caller will handle the error as an exception.

portal-api/src/v1/users/controller.js
```diff
@@ -241,8 +241,12 @@ exports.disable = async (_id, callback) => {
...
-  const status = await collection.deleteOne({ _id: ObjectId(_id) });
+  if (ObjectId.isValid(_id)) {
+    const collection = await db.getCollection('users');
+    const status = await collection.deleteOne({ _id: { $eq: ObjectId(_id) } });
 
-  callback(null, status);
+    callback(null, status);
+  }
+
+  callback('Invalid portal user id', 400);
```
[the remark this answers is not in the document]
abhi-markan: Same as above.

portal-api/src/v1/controllers/feedback.controller.js
```diff
@@ -113,12 +113,20 @@ exports.findAll = (req, res) => (
...
     if (!feedback) {
       return res.status(404).send();
     }
     const collection = await db.getCollection('feedback');
-    const status = await collection.deleteOne({ _id: ObjectId(req.params.id) });
+    const status = await collection.deleteOne({ _id: { $eq: ObjectId(id) } });
     return res.status(200).send(status);
   });
+
+  return res.status(400).send();
```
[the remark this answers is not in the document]
abhi-markan: Have refactored.

portal-api/src/v1/controllers/mandatoryCriteria.controller.js
```diff
@@ -78,6 +78,13 @@ exports.update = async (req, res) => {
...
-  res.status(200).send(status);
+  const { version } = req.params;
+  const versionNumber = Number(version);
+
+  if (typeof versionNumber === 'number') {
+    const status = await collection.deleteOne({ version: { $eq: versionNumber } });
+    res.status(200).send(status);
+  }
+
+  res.status(400).send('Invalid mandatory criteria version number');
```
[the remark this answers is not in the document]
abhi-markan: Have refactored.

portal-api/src/v1/controllers/eligibilityCriteria.controller.js
abhi-markan: Have refactored.

trade-finance-manager-api/src/v1/controllers/user/user.controller.js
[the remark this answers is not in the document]
lmarrai-sw: That makes sense, thanks @abhi-markan

portal-api/src/v1/users/controller.js
lmarrai-sw, line 251: We should return on or after line 248, otherwise the `callback` outside the `if` statement will run as well

trade-finance-manager-api/src/v1/controllers/user/user.controller.js
lmarrai-sw, line 105: We should `return` at or after line 102 so the 2nd `callback` doesn't execute as well :)

portal-api/src/v1/users/controller.js
abhi-markan: Good, spot have added.

trade-finance-manager-api/src/v1/controllers/user/user.controller.js
abhi-markan: Have added.

PR #1948, opened 2023-08-09
feat(FN-461): commonise tags

OxonianCambion:
### Introduction

This card covers 
* commonising the tag application
* NOTE: this card does not attempt to enhance the tagging, merely produce the same tagging as is currently there
 
### Resolution
* The `Environment` tag on resources are applied by a Policy
* As the only tags anywhere are `Environment` we can simply remove them everywhere. (Note that if we set `Environment` to a different value, the policy updates them back again!)
* I have just set an empty object, rather than omit the `tags` field to make it clearer where tags can be applied if we do enhance our tagging.


infrastructure/resource_group_level/main.bicep
```diff
@@ -137,9 +137,7 @@ resource containerRegistry 'Microsoft.ContainerRegistry/registries@2023-01-01-pr
 resource logAnalyticsWorkspace 'Microsoft.OperationalInsights/workspaces@2022-10-01' = {
   name: logAnalyticsWorkspaceName
   location: location
-  tags: {
```
OxonianCambion, line 140: A really nice feature is that the `what-if` deploy accurately includes the tagging due to the Policy!

infrastructure/resource_group_level/modules/application-gateway-tfm.bicep
```diff
@@ -81,9 +81,7 @@ var requestRoutingRules = [
 resource applicationGateway 'Microsoft.Network/applicationGateways@2022-11-01' = {
   name: applicationGatewayName
   location: location
-  tags: {
-    Environment: 'Preproduction'
-  }
+  tags: {}
```
OxonianCambion, line 84: I've literally just made this change everywhere with a big regex...

PR #1949, opened 2023-08-09
feat(FN-864): prevent cosmosdb 429 errors - batch collection creation

OxonianCambion:
### Introduction

This card covers
* Preventing 429 errors on first deploy of CosmosDB
### Resolution
* stop the attempted creation of all collections at the same time by putting them in an array and running in a batched loop.

### DIFF
--- a/infrastructure/resource_group_level/modules/cosmosdb.bicep
+++ b/infrastructure/resource_group_level/modules/cosmosdb.bicep
@@ -5,7 +5,7 @@ param privateEndpointsSubnetId string
 param databaseName string
 
 
-@description('Netowrk IPs to permit access to CosmosDB')
+@description('Network IPs to permit access to CosmosDB')
 @secure()
 param allowedIpsString string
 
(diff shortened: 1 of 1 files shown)


infrastructure/resource_group_level/modules/cosmosdb.bicep
```diff
@@ -130,390 +130,368 @@ resource submissionsDb 'Microsoft.DocumentDB/databaseAccounts/mongodbDatabases@2
   }
 }
 
-resource banksCollection 'Microsoft.DocumentDB/databaseAccounts/mongodbDatabases/collections@2023-04-15' = {
```
OxonianCambion, line 133: I've basically put all the resource definitions in an array (stripping out the `parent` as it's always the same.

infrastructure/resource_group_level/modules/cosmosdb.bicep
```diff
@@ -130,390 +130,368 @@ resource submissionsDb 'Microsoft.DocumentDB/databaseAccounts/mongodbDatabases@2
...
+        ]
+      }
     }
-  }
-}
+  } 
+]
+
+// We set a batch size because otherwise Azure tries to create all of the resources in parallel and we get 429 errors.
+@batchSize(4)
```
OxonianCambion, line 488: Then create them in a loop here, using this handy batch size limit. 4 seemed reasonable and worked when I tested it.

IgnatG (approved):

PR #1961, opened 2023-08-14
fix(DTFS2-6615): allow most csp headers in e2e tests

lmarrai-sw:
## Introduction
The bug fixed in !1950 was caused by the `script-src-attr` directive in the CSP header being too strict.

The e2e tests did cover (at least) one journey affected by this bug, but they did not fail because [Cypress removes the CSP header from your response](https://docs.cypress.io/guides/guides/content-security-policy) when it tests your app. You can allow most (but not all) CSP directives through [an experimental Cypress setting](https://docs.cypress.io/guides/references/experiments#Experimental-CSP-Allow-List).

## Resolution

I have configured the e2e tests to allow as many of the CSP directives as possible via an experimental Cypress setting. This PR will check if the e2e tests still pass.

Note that experimental Cypress settings can have breaking changes in Cypress releases **without** a major semver change (see https://docs.cypress.io/guides/references/release-stages#Experimental for more details) but I think this risk is worth it to be able to check that CSP policies haven't broken DTFS.

## Miscellaneous
* Reverted `script-scr-attr` for `trade-finance-manager-ui` to `self` thereby removing `unsafe-hashes` directive.
* Replaced all `onChange` inline script handles to external file event handles for `change` type on both `portal` and `trade-finance-manager-ui`.
* Re-build assets for both UI microservices as mentioned above.
* Introduced `Array.isArray()` validation check for amendments.


e2e-tests/gef/cypress.config.js
```diff
@@ -24,4 +24,5 @@ module.exports = defineConfig({
     baseUrl: 'http://localhost',
     specPattern: 'cypress/e2e/**/*.spec.js',
   },
+  experimentalCspAllowList: ['child-src', 'default-src', 'frame-src', 'form-action', 'script-src', 'script-src-elem'],
```
AlexBramhill, line 27: If we're trying to allow as many csp headers as possible -- could we just set `experimentalCspAllowList: true` ?

lmarrai-sw: That's what you'd expect but that's not the way it works! As per https://docs.cypress.io/guides/references/experiments#Strip-Minimum-CSP-Directives, the way it works is:
- `false`: the whole CSP header gets stripped
- `true`:
   - strips the CSP directives that _always_ break Cypress
   - strips the CSP directives that _conditionally_ break Cypress
   - leaves the CSP directives that don't affect Cypress alone
- an array of directives: strips the CSP directives that _always_ break Cypress
   - strips the CSP directives that _always_ break Cypress
   - strips any CSP directives that _conditionally_ break Cypress that you *do not* list
   - leaves any CSP directives that _conditionally_ break Cypress that you *do* list
   - leaves the CSP directives that don't affect Cypress alone

AlexBramhill (approved):

PR #1955, opened 2023-08-11
fix(SR-8): sanitise db input additional changes

oscar-richardson-softwire:
Additional changes for https://github.com/UK-Export-Finance/dtfs2/pull/1946

PR #1963, opened 2023-08-15
fix(sr-8): update error message name

AlexBramhill:
## Introduction
@abhi-markan suggested renaming `errCode` to `status` in some error messages in https://github.com/UK-Export-Finance/dtfs2/pull/1955#discussion_r1293534151

## Resolution
As per the conversation below: We've kept `errCode` but added a `status` field following existing patterns

### DIFF
--- a/portal-api/src/v1/gef/controllers/facilities.controller.js
+++ b/portal-api/src/v1/gef/controllers/facilities.controller.js
@@ -15,6 +15,7 @@ exports.create = async (req, res) => {
   if (!req.body.type || !req.body.dealId) {
     return res.status(422).send([{ status: 422, errCode: 'MANDATORY_FIELD', errMsg: 'No Application ID and/or facility type sent with request' }]);
   }
+
   if (enumValidationErr) {
     return res.status(422).send(enumValidationErr);
   }
--- a/portal-api/src/v1/gef/controllers/facilities.controller.js
+++ b/portal-api/src/v1/gef/controllers/facilities.controller.js
@@ -12,7 +12,7 @@ const dealsCollectionName = 'deals';
 
 exports.create = async (req, res) => {
   const enumValidationErr = facilitiesCheckEnums(req.body);
-  if (!req.body.type && !req.body.dealId) {
+  if (!req.body.type || !req.body.dealId) {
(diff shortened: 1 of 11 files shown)


portal-api/src/v1/gef/controllers/facilities.controller.js
```diff
@@ -42,7 +42,7 @@ exports.create = async (req, res) => {
       res.status(201).json(response);
     }
   } else {
-    res.status(422).send([{ errCode: 'MANDATORY_FIELD', errMsg: 'No Application ID and/or facility type sent with request' }]);
+    res.status(422).send([{ status: 'MANDATORY_FIELD', errMsg: 'No Application ID and/or facility type sent with request' }]);
```
lmarrai-sw, line 45: @abhi-markan are we sure we want to make this change? `MANDATORY_FIELD`, `company-profile-not-found`, `ENUM_ERROR`, etc. do seem like error codes rather than statuses to me

abhi-markan: Good spot, can we do following please.

```javascript
{ status: 422, errCode: 'MANDATORY_FIELD'....}
```

portal-api/src/v1/gef/controllers/facilities.controller.js
```diff
@@ -1,49 +1,43 @@
...
-    }
-
-    const facilitiesQuery = await db.getCollection(facilitiesCollectionName);
-    const createdFacility = await facilitiesQuery.insertOne(new Facility(req.body));
-
-    const { insertedId } = createdFacility;
+  if (!req.body.type || !req.body.dealId) {
+    return res.status(422).send([{ status: 422, errCode: 'MANDATORY_FIELD', errMsg: 'No Application ID and/or facility type sent with request' }]);
+  }
+  if (enumValidationErr) {
```
abhi-markan, line 18: Missing CRLF please before `if`.

abhi-markan (changes requested):
One minor comment to address please.

portal-api/src/v1/gef/controllers/facilities.controller.js
AlexBramhill, line 18:
```suggestion
  }

  if (enumValidationErr) {
```

abhi-markan (approved):

PR #1964, opened 2023-08-15
feat(FN-930): configure vnet peering

OxonianCambion:
### Introduction

This card covers 
* wiring up vnet peering with the `VNET_UKEF_UKS` vnet on another subscription.
### Resolution
* OUR Side only - adds vnet link
* NOTE: Someone also needs to configure the `VNET_UKEF_UKS` at least for `feature`.

NOTE: I haven't tested this as I can't deploy until Risual increase the number of available static IPs in Azure.

### DIFF
--- a/infrastructure/resource_group_level/main.bicep
+++ b/infrastructure/resource_group_level/main.bicep
@@ -40,6 +40,10 @@ param vnetAddressPrefixes array = [
 ]
 
 param privateEndpointsCidr string = '172.16.20.0/24'
+
+param peeringRemoteVnetSubscriptionId string = '08887298-3821-49f0-8303-f88859c12b9b'
+param peeringRemoteVnetResourceGroupName string = 'UKEF-Firewall-Appliance-UKS'
+param peeringRemoteVnetName string = 'VNET_UKEF_UKS'
 param peeringAddressSpace string = '10.50.0.0/16'
 
 @allowed(['Allow', 'Deny'])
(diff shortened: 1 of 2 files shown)


infrastructure/resource_group_level/modules/vnet.bicep
```diff
@@ -7,10 +7,17 @@ param applicationGatewayCidr string
 
 param appServicePlanName string
 
+// TODO:DTFS-6422 check route table config isn't needed and happens automatically from the route table setup.
 param routeTableId string
 param networkSecurityGroupId string
-// TODO:DTFS-6422 Note that peeringAddressSpace doesn't seem to appear in the old IaC scripts
+
+// TODO:DTFS-6422 Note that the staging name is: "tfs-test-vnet_vnet-ukef-uks". We should ensure we start with that name as an override.
+param peeringVnetName string = 'tfs-${environment}-vnet_vnet-ukef-uks'
```
OxonianCambion, line 15: I think it is best to start with the misnaming on staging, and then change it later.

infrastructure/resource_group_level/main.bicep
```diff
@@ -40,6 +40,10 @@ param vnetAddressPrefixes array = [
 ]
 
 param privateEndpointsCidr string = '172.16.20.0/24'
+
+param peeringRemoteVnetSubscriptionId string = '08887298-3821-49f0-8303-f88859c12b9b'
```
IgnatG, line 44: this one needs to be an environment variable

infrastructure/resource_group_level/main.bicep
```diff
@@ -40,6 +40,10 @@ param vnetAddressPrefixes array = [
...
+param peeringRemoteVnetResourceGroupName string = 'UKEF-Firewall-Appliance-UKS'
```
IgnatG, line 45: same as above, this one needs to be an environment variable

infrastructure/resource_group_level/main.bicep
```diff
@@ -40,6 +40,10 @@ param vnetAddressPrefixes array = [
...
+param peeringRemoteVnetName string = 'VNET_UKEF_UKS'
```
IgnatG, line 46: same as above, this one needs to be an environment variable

infrastructure/resource_group_level/main.bicep
OxonianCambion: Yes - which is why I've made it a `param`, though I need to put in a default for my local deploys and for use until we have the GHA config & secrets set up.
My understanding is that the subscriptionId isn't considered secret.

IgnatG (approved):

PR #1976, opened 2023-08-24
feat(FN-742): Set values from GitHub

OxonianCambion:
### Introduction

This card covers
* input of secure parameters, typically from GitHub secrets
* input of other parameter
### Resolution
* multiple params are added, deviating from the normal naming scheme for clarity
* combination of params into the correct inputs for the app services
* NOTE: I have kept the distinction between the different "classes" of parameters.

### DIFF
--- a/infrastructure/resource_group_level/modules/function-acbs.bicep
+++ b/infrastructure/resource_group_level/modules/function-acbs.bicep
@@ -10,6 +10,8 @@ param nodeDeveloperMode bool
 
 param resourceNameFragment string = 'function-acbs'
 
+param settings object
+
 // These values are taken from GitHub secrets injected in the GHA Action
 @secure()
 param secureSettings object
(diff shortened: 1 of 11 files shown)


infrastructure/resource_group_level/modules/function-acbs.bicep
```diff
@@ -10,24 +10,15 @@ param nodeDeveloperMode bool
 
 param resourceNameFragment string = 'function-acbs'
 
+param settings object
```
OxonianCambion, line 13: Overall, we generally have the following classes of settings:
`settings`: non-sensitive values from `vars` or just hard coded in the main file, observed in the CLI scripts.
`secureSettings`: sensitive values from GH secrets, observed in the CLI scripts.
`additionalSecureSettings`: values that appeared when I performed an export from the `dev` environment, that aren't in the CLI scripts. Perhaps some can be removed.
`staticSettings`: settings that are either hard-coded or can be calculated inline.
`secureConnectionStrings`: connection strings that come from secrets and are in the CLI scripts. As discussed, we may want to move these to normal secureSettings at some point.
`additionalSecureConnectionStrings`: connection strings that appeared when I performed an export on `dev`.

IgnatG (approved):
LGTM

PR #2041, opened 2023-09-06
feat(DTFS2-6182): replace get facilities graphql endpoint with REST endpoint

francescastocco:
## Introduction

## Resolution
Replace the get facilities graphql endpoint with REST endpoint as part of the gql -> rest migration in TFM API.

### DIFF
--- a/trade-finance-manager-api/src/v1/controllers/facility.controller.js
+++ b/trade-finance-manager-api/src/v1/controllers/facility.controller.js
@@ -31,12 +31,14 @@ const getFacilities = async (req, res) => {
   try {
     const queryParams = req.body;
 
+    // TODO DTFS2-6182: This is as current resolver implimentation, but should this be a 400?
     if (!queryParams) {
       return {};
     }
 
     const dbFacilities = await api.getAllFacilities(queryParams);
 
+    // TODO DTFS2-6182: Why is this different from getFacility?
     const facilities = dbFacilities.map((dbFacility) => {
       const { tfmFacilities: facility } = dbFacility;
 
@@ -66,6 +68,7 @@ const getFacilities = async (req, res) => {
 
       if (latestCompletedAmendment?.coverEndDate) {
         const { coverEndDate } = latestCompletedAmendment;
+        // TODO DTFS2-6182: Why is the below * 1000?
         // * 1000 to convert to ms epoch time format so can be correctly formatted by template
         facilityCoverEndDate = format(new Date(coverEndDate * 1000), 'dd MMM yyyy');
         facilityCoverEndDateEpoch = coverEndDate;
--- a/trade-finance-manager-api/src/v1/controllers/facility.controller.js
+++ b/trade-finance-manager-api/src/v1/controllers/facility.controller.js
@@ -1,6 +1,11 @@
+const { format, getUnixTime } = require('date-fns');
+const commaNumber = require('comma-number');
 const api = require('../api');
 const { findOneTfmDeal } = require('./deal.controller');
 const facilityMapper = require('../graphql-mappings/facility');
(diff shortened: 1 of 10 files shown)


trade-finance-manager-api/src/v1/controllers/facility.controller.js
```diff
@@ -22,6 +27,66 @@ const getFacility = async (req, res) => {
...
+
+      if (!facility.ukefFacilityId) {
+        facility.ukefFacilityId = '-';
+      }
+
+      const latestCompletedAmendment = findLatestCompletedAmendment(dbFacility.amendments);
+      let facilityCoverEndDate = '';
+      let facilityCoverEndDateEpoch = '';
+
+      if (REGEX.DATE.test(facility.coverEndDate) && facility.coverEndDate) {
```
ChristianMcc, line 51: Is this regex just the way it was done before? It also feels wrong to check facility.coverEndDate exists after testing it in regex before that
do you reckon we can do something instead where we create a new date from the coverEndDate and check that the date is valid after instead? (sorry I'm reviewing copy pasted code but worth having these noted)

trade-finance-manager-ui/server/controllers/facilities/index.js
```diff
@@ -3,7 +3,8 @@ const CONSTANTS = require('../../constants');
 
 const getFacilities = async (req, res) => {
   const { userToken } = req.session;
-  const apiResponse = await api.getFacilities();
+  // queryParams here?
```
ChristianMcc, line 6: Interesting that they never passed some parameters through here before - I'd leave a comment saying query parameters have been implemented in APIs but aren't used in the frontends currently

trade-finance-manager-api/src/v1/controllers/facility.controller.js
francescastocco: Ah yes, I did refactor this a bit, I've now updated the condition which doesn't make sense - [this](https://github.com/UK-Export-Finance/dtfs2/blob/main/trade-finance-manager-api/src/graphql/resolvers/query-facilities.js) is the code that I am replicating

trade-finance-manager-api/src/v1/controllers/facility.controller.js
```diff
@@ -1,4 +1,105 @@
...
+  try {
+    const queryParams = req.body;
+
+    if (!queryParams) {
+      return {};
+    }
+
+    const dbFacilities = await api.getAllFacilities(queryParams);
+
+    const facilities = dbFacilities.map((dbFacility) => {
```
AlexBramhill, line 40: Could this be a separate function?

trade-finance-manager-ui/server/controllers/facilities/index.js
```diff
@@ -3,7 +3,8 @@ const CONSTANTS = require('../../constants');
 
 const getFacilities = async (req, res) => {
   const { userToken } = req.session;
-  const apiResponse = await api.getFacilities();
+  // Query parameters have ben implemented in APIs but aren't currently used in frontends
```
AlexBramhill, line 6:
```suggestion
  // Query parameters have been implemented in APIs but aren't currently used in frontends
```

AlexBramhill (commented):
reviewed non-mapping files

trade-finance-manager-api/src/v1/controllers/facility.controller.js
```diff
@@ -1,4 +1,105 @@
...
+    return res.status(500).send(error.message);
+  }
+};
+
+const getFacilities = async (req, res) => {
+  try {
+    const queryParams = req.body;
+
+    if (!queryParams) {
+      return {};
```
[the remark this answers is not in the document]
AlexBramhill: I've added a todo to check this -- It's in the existing resolver implimentation

PR #2042, opened 2023-09-06
Feat(DTFS2-6182): replace update facility graphql endpoint with REST endpoint

francescastocco:
## Introduction

## Resolution
Replace the updateFacilities and updateFacilityRiskProfile graphql endpoints with an updateFacility REST endpoint as part of the gql -> rest migration in TFM API.

### DIFF
--- a/trade-finance-manager-ui/server/api.js
+++ b/trade-finance-manager-ui/server/api.js
@@ -78,13 +78,17 @@ const getDeals = async (queryParams) => {
 };
 
 const getFacility = async (id, token) => {
-  const response = await axios({
-    method: 'get',
-    url: `${TFM_API_URL}/v1/facilities/${id}`,
-    headers: generateHeaders(token),
-  });
-
-  return response.data.facility;
+  try {
+    const response = await axios({
+      method: 'get',
+      url: `${TFM_API_URL}/v1/facilities/${id}`,
+      headers: generateHeaders(token),
+    });
+    return response.data.facility;
+  } catch (error) {
+    console.error(error);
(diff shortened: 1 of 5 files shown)


trade-finance-manager-ui/server/api.js
```diff
@@ -112,22 +110,49 @@ const updateParty = async (id, partyUpdate) => {
...
+    const response = await axios({
+      method: 'put',
+      url: `${TFM_API_URL}/v1/facilities/${id}`,
+      headers: generateHeaders(token),
+      data: facilityUpdate,
+    });
+
+    return response.data;
+  } catch (error) {
+    console.error('Unable to update facility request %O', error);
```
ChristianMcc, line 131: I know that some of the error messages have the word request in but can we delete it I think it's misleading

trade-finance-manager-ui/server/api.js
```diff
@@ -112,22 +110,49 @@ const updateParty = async (id, partyUpdate) => {
...
+
+    const response = await axios({
+      method: 'put',
+      url: `${TFM_API_URL}/v1/facilities/${id}`,
+      headers: generateHeaders(token),
+      data: facilityUpdate,
+    });
+    return response.data;
+  } catch (error) {
+    console.error('Unable to update facility risk profile request %O', error);
```
ChristianMcc, line 153: ditto

ChristianMcc (approved):
Nit pick on error message, going to approve so you can merge once fixed (if it unapproves on a new commit I'll re-approve!)

PR #2043, opened 2023-09-06
feat(DTFS2-6182): replace update party graphql endpoint with REST endpoint 

francescastocco:
## Introduction

## Resolution
Replace the update party graphql endpoint with a REST endpoint as part of the gql -> rest migration in TFM API.

### DIFF
--- a/trade-finance-manager-api/src/v1/controllers/party.controller.js
+++ b/trade-finance-manager-api/src/v1/controllers/party.controller.js
@@ -13,7 +13,7 @@ const updateParty = async (req, res) => {
     const updatedDeal = await api.updateDeal(dealId, partyUpdate);
 
     if (updatedDeal.dealSnapshot) {
-      if (await canDealBeSubmittedToACBS(updatedDeal.dealSnapshot.submissionType)) {
+      if (canDealBeSubmittedToACBS(updatedDeal.dealSnapshot.submissionType)) {
         await submitACBSIfAllPartiesHaveUrn(dealId);
       }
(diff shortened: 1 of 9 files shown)


trade-finance-manager-api/src/v1/controllers/party.controller.js
```diff
@@ -0,0 +1,34 @@
...
+    tfm: {
+      parties: req.body,
+    },
+  };
+
+  try {
+    const updatedDeal = await api.updateDeal(dealId, partyUpdate);
+
+    if (updatedDeal.dealSnapshot) {
+      if (await canDealBeSubmittedToACBS(updatedDeal.dealSnapshot.submissionType)) {
```
ChristianMcc, line 16: I'm guessing was just copy pasted but I think this isn't an async function and doesn't need to be awaited

trade-finance-manager-api/src/v1/controllers/tfm.controller.js
```diff
@@ -1,6 +1,7 @@
 const activity = require('../helpers/activity');
 const api = require('../api');
 
+// This is never used - should I delete it?
```
ChristianMcc, line 4: Ah yes the function from deal.controller.js is the only one used right? Happy for this to be deleted

ChristianMcc (changes requested):
Minor async change comment then can merge :)

AlexBramhill (commented):
Review non-mappings files

PR #2044, opened 2023-09-06
Feat(DTFS2-6182): replace update deal graphql endpoints with REST endpoint

francescastocco:
## Introduction

## Resolution
Replace the updateCreditRating, updateLossGivenDefault, updateProbabilityOfDefault, createActivity graphql endpoints with a single `updateDeal` REST endpoint as part of the gql -> rest migration in TFM API.


ChristianMcc (approved):
LGTM, before merging to the other branch can you dismiss all the alerts as 'False positive won't fix' please :)

PR #2050, opened 2023-09-12
feat(DTFS2-6637): remove unused roles from portal api

lmarrai-sw:
## Introduction

We are cleaning up the roles in `portal` and reducing the list to just `maker`, `checker`, `read-only`, `admin`.
There are other roles in `portal-api` that should now be unused, such as `data-admin`, `interface`, `editor`, `ukef_operations`

## Resolution
I've removed the unused roles from `portal-api` (and a couple of references in `dtfs-central-api`, and `utils`)


utils/data-migration/bss-ewcs/migrate-users.js
```diff
@@ -17,7 +18,7 @@ const loadUsersFromFile = () => {
 
 const generateRoles = (v1Role) => {
   if (v1Role === 'Maker/Checker') {
```
AlexBramhill, line 20: Are we keeping this in for legacies sake?

lmarrai-sw: Yep exactly, I don't think we can remove this check because the data being migrated from will still have this role

AlexBramhill (approved):

PR #2055, opened 2023-09-14
Feat(FN-980): List previous reports page

francescastocco:
## Introduction

We need to allow users to see a list of previously submitted reports.

## Resolution

- Added the previous reports page at `/previous-reports`.
- The page has a side navigation for the years that the users bank has uploaded reports for.
- When a user selects a year in the side navigation a query parameter is set for the target year.
- The page will reload and display the list of reports for the target year selected.
- The backend has been mocked until confirmation of using MongoDB/migration to SQL.
- Added E2E tests for this functionality.
- Added unit tests for converting month number to month name.
- Added component test for this page.

## To Do

- Implement backend solution.


e2e-tests/portal/cypress/e2e/pages/utilisation-report-service/previousReports.js
```diff
@@ -0,0 +1,11 @@
+const page = {
+  mainHeading: () => cy.get('[data-cy="main-heading"]'),
+  bodyText: () => cy.get('[data-cy="paragraph"]'),
+  sideNavigation2022Item: () => cy.get('[data-cy="side-navigation-2022"]'),
```
ChristianMcc, line 4: Can we refactor the sideNavigation and listItem functions to take in a parameter of year/month which targets the right year/month in the cy.get please

e2e-tests/portal/cypress/e2e/journeys/payment_officer/previous-reports/list-previous-reports.spec.js
```diff
@@ -0,0 +1,63 @@
+const { previousReports } = require('../../../pages');
+const MOCK_USERS = require('../../../../fixtures/users');
+const relativeURL = require('../../../relativeURL');
+
+// TODO FN-980 update to payment officer role
+const { BANK1_MAKER1 } = MOCK_USERS;
+
+context('List previous utilisation reports', () => {
```
ChristianMcc, line 8: it's a bit of an obvious one but can we have a test that when you first load the page it defaults to the most recent year that has reports please?

portal/component-tests/utilisation-report-service/previous-reports.component-test.js
```diff
@@ -0,0 +1,119 @@
+const pageRenderer = require('../pageRenderer');
+
+const page = 'utilisation-report-service/partials/previous-reports.njk';
+const render = pageRenderer(page);
+
+describe(page, () => {
```
ChristianMcc, line 6: Nice test suite :)

portal/server/controllers/utilisation-report-service/previous-reports/index.js
```diff
@@ -0,0 +1,49 @@
...
+  // get data from the DB (currently using json) - will need to pass in bank id from user
+  try {
+    const { targetYear } = req.query;
+    const navItems = utilisationData.map((utilisation) => ({
+      text: utilisation.year,
+      href: `?targetYear=${utilisation.year}`,
+      attributes: { 'data-cy': `side-navigation-${utilisation.year}` },
+    }));
+
+    const utilisation = targetYear ? utilisationData.find((x) => x.year.toString() === targetYear) : utilisationData[0];
```
ChristianMcc, line 20: Leaving a note that this explicitly needs backend to order the data in descending year, nothing wrong with that just for if I look later!

portal/server/controllers/utilisation-report-service/previous-reports/index.js
```diff
@@ -0,0 +1,49 @@
...
+    const utilisationIndex = targetYear ? utilisationData.findIndex((x) => x.year.toString() === targetYear) : 0;
+    if (navItems.length) {
+      navItems[utilisationIndex].active = true;
+    }
+
+    const reports = navItems.length
```
ChristianMcc, line 26: just to be safe let's chuck a load of optional chaining in here and the below lines e.g. navItems?.length and utilisation?.reports?.map.
When we get this from the API we'll have to think about error cases and if we always have an empty array etc. being returned so a few of these pieces of logic don't fail if an API call fails

portal/server/controllers/utilisation-report-service/previous-reports/index.js
```diff
@@ -0,0 +1,49 @@
...
+  // offset by 1 as January = 1 in Database
+  date.setMonth(monthNumber - 1);
+  return date.toLocaleString('default', { month: 'long' });
+};
+
+const getUtilisationReportDownload = async (req, res) => {
+  // get data from the DB (currently using json) - will need to pass in bank id from user
+  try {
+    const { targetYear } = req.query;
+    const navItems = utilisationData.map((utilisation) => ({
```
ChristianMcc, line 14: ditto here for making sure we always have an array in any error case coming from portal API, would prefer optional chaining to be safe in future if someone changes the API for example

ChristianMcc (changes requested):
Looks good to me! Couple of nit picks on optional chaining + an extra test :)

portal/server/controllers/utilisation-report-service/previous-reports/index.js
```diff
@@ -0,0 +1,49 @@
...
+      ? utilisation.reports.map((report) => ({
+        month: getMonthName(report.month),
+        path: report.path,
+      }))
+      : [];
+    const year = navItems.length
+      ? utilisation.year
+      : undefined;
+
+    return res.render('utilisation-report-service/partials/previous-reports.njk', {
```
francescastocco, line 36: need to add user property here 
`user: req.session.user`

portal/server/controllers/utilisation-report-service/previous-reports/index.js
```diff
@@ -0,0 +1,49 @@
...
+      navItems,
+      reports,
+      year,
+    });
+  } catch (error) {
+    return res.render('partials/problem-with-service.njk');
```
francescastocco, line 42: And also here 
`user: req.session.user`

portal/server/controllers/utilisation-report-service/previous-reports/index.js
francescastocco: Ah yes, I've made a note that the backend should be ordered by descending year and ascending month.
Additionally the backend will return an empty array for years where no reports were submitted but were submitted for years before AND after this year

e2e-tests/portal/cypress/e2e/journeys/payment_officer/previous-reports/list-previous-reports.spec.js
francescastocco: Yes, I've added a test, let me know what you think as I was trying to test without using knowledge of the test data, if that makes sense?

e2e-tests/portal/cypress/e2e/journeys/payment_officer/previous-reports/list-previous-reports.spec.js
francescastocco, line 1: need to delete this

portal/server/routes/utilisation-report-service/index.js
```diff
@@ -0,0 +1,9 @@
+const express = require('express');
+
+const utilisationReportDownloadRoutes = require('./previous-reports');
+
+const router = express.Router();
+
+router.use('/previous-reports', utilisationReportDownloadRoutes);
```
francescastocco, line 7: call this previous reports

ChristianMcc (approved):
Approved but can you try and rerun the tests and see if you get passes, I think this is all fine but the tests are very flakey :(

ChristianMcc (approved):

PR #2062, opened 2023-09-20
fix(DTFS2-6679): display errors on the edit user page

oscar-richardson-softwire:
### Introduction

Currently the page that admins use to edit a user does not display errors if the update fails. Instead it assumes that the update succeeded and redirects the admin to the list of all users (as it does when the update succeeds).

The update can fail if the admin changes the user to have the `read-only` role AND one or more other roles. This is only possible to achieve via the UI in browsers with JavaScript disabled.

Steps to reproduce bug:

1. Disable JavaScript in the browser
2. Log in to Portal UI as the username ‘ADMIN’
3. Go to the ‘Users’ tab
4. Click on username 'BANK1_MAKER1’, taking you to the edit user page
5. Check the ‘Read-only’ box, so that ‘Maker’ and ‘Read-only’ are both checked
6. Click the save button
7. You are redirected to the ‘Users’ tab, but the update has silently failed - ‘BANK1_MAKER1’ still only has the 'Maker’ role

### Resolution

I've changed the code to check if the update fails and display the edit page with the errors (and keeping the in-progress edit data) if so.

### Misc

- At the moment, it appears that no errors would be shown if the update failed due to a reason other than an invalid selection in the UI (e.g., a 500 Internal Server Error). This should be investigated and it should be decided how we want to deal with this situation, and a ticket should be raised
- I haven't been able to write API tests for this change yet, due to VM issues
- We cannot write E2E tests for this change, as the bug is only present when JS is disabled, which isn't possible in the E2E tests

### DIFF
--- a/portal/server/routes/admin/users.js
+++ b/portal/server/routes/admin/users.js
@@ -10,6 +10,21 @@ const {
 
 const router = express.Router();
 
+const getOptionsForRenderingEditUser = async (req, res) => {
+  const { _id, userToken } = requestParams(req);
+  const { user } = req.session;
+
+  const banks = await getApiData(api.banks(userToken), res);
+  const userToEdit = await getApiData(api.user(_id, userToken), res);
+
+  return {
+    _id,
+    banks: banks.sort((bank1, bank2) => bank1.name < bank2.name),
+    displayedUser: userToEdit,
+    user,
+  };
+}
+
 router.get('/users', async (req, res) => {
   const { _id, userToken } = requestParams(req);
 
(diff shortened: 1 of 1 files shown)


portal/server/routes/admin/users.js
```diff
@@ -123,14 +123,37 @@ router.post('/users/edit/:_id', async (req, res) => {
...
+    ...update,
+  };
+
+  return res.render('admin/user-edit.njk', {
+    _id,
+    banks: banks.sort((bank1, bank2) => bank1.name < bank2.name),
+    displayedUser: editedUser,
+    user,
+    validationErrors: formattedValidationErrors,
+  });
```
lmarrai-sw, line 156: Are we able to commonise this with the similar code for `GET /users/edit/:id` above at all?

lmarrai-sw (commented):
We're currently waiting for @oscar-richardson-softwire's docker issues to be resolved so he can write tests

lmarrai-sw (approved):

PR #2070, opened 2023-09-22
Feat(FN-957): Create payment officer role in portal

francescastocco:
## Introduction

Allow admin users to create/edit users to have a new `payment officer` role which will give access to the utilisation reporting service.

## Resolution

- Added payment officer check box to create/edit user form.
- Added a test user in mock data loader.
- Updated component & e2e tests.

### DIFF
--- a/portal/templates/admin/user-edit.njk
+++ b/portal/templates/admin/user-edit.njk
@@ -93,6 +93,13 @@
           attributes: {"data-cy": "role-checker"}
         },
         {
+          value: "payment-officer",
+          text: "Payment Officer",
+          hint: { text: "Report on GEF utilisation and fees" },
+          checked: displayedUser.roles.includes("payment-officer"),
+          attributes: {"data-cy": "role-payment-officer"}
+        },
+        {
           value: "admin",
           text: "Admin",
           hint: { text: "Creates, updates, and removes user accounts" },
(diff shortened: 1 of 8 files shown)


portal/templates/admin/user-edit.njk
```diff
@@ -92,6 +92,13 @@
           checked: displayedUser.roles.includes("checker"),
           attributes: {"data-cy": "role-checker"}
         },
+        {
+          value: "payment-officer",
```
ChristianMcc, line 96: it would be nicer if we passed these values in from the constants in the NJK files but given read-only work didn't do this I'm guessing it was decided against so let's leave it as this, I just like leaving random thoughts on PRs apparently

utils/mock-data-loader/portal/users.js
```diff
@@ -98,6 +98,16 @@ const USERS = [
     roles: [MAKER, CHECKER],
     bank: UKEF_TEST_BANK_1,
   },
+  {
+    username: 'BANK1_PAYMENT_OFFICER1',
+    password: 'AbC!2345',
+    firstname: 'Payton',
```
ChristianMcc, line 104: Great name choice

ChristianMcc (approved):

PR #2074, opened 2023-09-25
feat(FN-1086): Sync env vars with main

OxonianCambion:
### Introduction

This card covers
* updating the environment variables to be consistent with the ones on `main`
### Resolution
* environment variables were updated
* connection strings were updated
* Some minor adjustments made


.github/workflows/development_infrastructure.yml
```diff
@@ -1315,7 +1315,8 @@ jobs:
                 WEBSITE_VNET_ROUTE_ALL=1 \
                 PORT=5000 \
                 WEBSITES_PORT=5000 \
-                RATE_LIMIT_THRESHOLD=${{ vars.RATE_LIMIT_THRESHOLD }}
+                RATE_LIMIT_THRESHOLD=${{ vars.RATE_LIMIT_THRESHOLD }} \
```
OxonianCambion, line 1318: I've updated the CLI scripts for the clearest variables, but I'm not 100% sure they are perfectly up to date.

infrastructure/resource_group_level/main.bicep
```diff
@@ -123,7 +123,7 @@ var externalApiSettings = {
 }
 var externalApiSecureSettings = {
   CORS_ORIGIN: CORS_ORIGIN
-  APIM_TFS_URL: APIM_TFS_URL
+  APIM_TFS_URL: APIM_TFS_URL // TODO:FN-1086 different between dev & GH.
```
OxonianCambion, line 126: Note that the GitHub secret has the value `<blah>/mdm/api/v1/` but the dev export has `<blah>/tfs/api/v1`.
I suspect the secret is wrong.

infrastructure/resource_group_level/main.bicep
```diff
@@ -180,16 +184,15 @@ var tfmApiAdditionalSecureSettings = {
   UKEF_INTERNAL_NOTIFICATION: UKEF_INTERNAL_NOTIFICATION
   DTFS_CENTRAL_API_KEY: DTFS_CENTRAL_API_KEY
   EXTERNAL_API_KEY: EXTERNAL_API_KEY
-  JWT_VALIDATING_KEY: JWT_VALIDATING_KEY
-  PORTAL_API_KEY: PORTAL_API_KEY
+  JWT_VALIDATING_KEY: JWT_VALIDATING_KEY // TODO:FN-1086 JWT_VALIDATING_KEY seems to have been moved to connection strings.
```
OxonianCambion, line 187: As per the comment, `JWT_VALIDATING_KEY` seems to be a connection string value in `dev` now. Is that deliberate?

infrastructure/resource_group_level/main.bicep
```diff
@@ -200,14 +203,12 @@ var tfmApiAdditionalSecureConnectionStrings = {
 }
 
 var portalUiSettings = {
-    RATE_LIMIT_THRESHOLD: RATE_LIMIT_THRESHOLD
+    RATE_LIMIT_THRESHOLD: RATE_LIMIT_THRESHOLD // TODO:FN-1086 30 on dev, 10000 on feature
```
OxonianCambion, line 206: Note that on dev, the rate limit is generally 300. If there is to be variations per-app, then we need to be cleverer here. Otherwise, we'll need up update the value of the secret to the desired one?

infrastructure/resource_group_level/modules/webapps/external-api.bicep
```diff
@@ -88,6 +88,7 @@ module externalApiWebapp 'webapp.bicep' = {
     privateEndpointsSubnetId: privateEndpointsSubnetId
     resourceNameFragment: resourceNameFragment
     scmMinTlsVersion: '1.2'
+    selfHostnameEnvironmentVariable: 'EXTERNAL_API_URL'
```
OxonianCambion, line 91: `external-api` needs to know its own address for emails. We can actually do that fairly easily as we set up the application settings in a separate resource from the webapp.

OxonianCambion:
Note that I've tested this by merging it to the Github deployment branch and pushing that.

.github/workflows/development_infrastructure.yml
```diff
@@ -1324,19 +1325,14 @@ jobs:
...
 
-            echo "External API url: http://$EXTERNAL_API_URL"
-
             az webapp config connection-string set --name tfs-${{ env.environment }}-dtfs-central-api \
               --connection-string-type custom \
               --settings \
                 CORS_ORIGIN=${{ secrets.CORS_ORIGIN }} \
                 MONGO_INITDB_DATABASE="${{ env.database_name }}" \
-                MONGODB_URI="$mongo_connection_string" \
-                EXTERNAL_API_URL="http://$EXTERNAL_API_URL"
```
abhi-markan, line 1339: EXTERNAL_API_URL has been removed, is it not required any more?

abhi-markan: Same for all of the below.

abhi-markan (changes requested):
Looks good, however a comment to address please.

.github/workflows/development_infrastructure.yml
OxonianCambion: Not for `central-api` no.
It had been removed from the `dev` environment variables, so I checked the javascript codebase and it wasn't there either. Therefore, I updated my Bicep & the CLI scripts here.
I've gone through quite carefully and have pruned everything that is clearly not needed, apart from `CORS_ORIGIN` which I'm still unsure of - it isn't in the export from dev.

I've Slacked you the discrepancies I've found.

abhi-markan (approved):

PR #2075, opened 2023-09-25
feat(FN-72): integrate build pipeline

OxonianCambion:
### Introduction

This card covers
* Adding Github Workflow to deploy infrastructure
### Resolution
* GitHub workflow added

NOTE: This PR does not cover deploying the App code.


.github/workflows/bicep_az_deployment_group.yml
```diff
@@ -0,0 +1,232 @@
+name: Perform action for an Az Deployment Group via Bicep
```
OxonianCambion, line 1: Infuriatingly the `arm-deploy` action, doesn't like special characters in secrets so we write our own.

When reviewing, it might make sense to start with `bicep_workflow.yml`, then do `bicep_deploy.yml` then this file, as that's the hierarchy.

.github/workflows/bicep_az_deployment_group.yml
```diff
@@ -0,0 +1,232 @@
...
+        required: true
+      ESTORE_URL:
+        description: webapp environment variable
+        required: true
+
+jobs:
+  configure-command:
+    name: Configure ${{ inputs.environment }} ${{ inputs.action }} command
+    runs-on: [self-hosted, linux, deployment]
+    # TODO: FN-72 NOTE: we don't currently have a `feature` GitHub environment yet so use the `dev` one while we're not doing much
```
OxonianCambion, line 127: I'll spin this off into another card, for when we have a `feature` GH environment ready.

.github/workflows/bicep_az_deployment_group.yml
```diff
@@ -0,0 +1,232 @@
...
+        # Note that we use the plain run command here, having installed Bicep above.
+        # - We don't use the azure/arm-deploy action because it doesn't seem to handle parameters with special characters (like * or =) in a secret at all.
+        # I have tried foo=${{ secrets.bar }}, setting the secret to an env and doing foo="$env", escaping values and lots of other things.
+        # - We don't use the azure/cli action because with the latest versions on 13/9/23 it doesn't seem
+        # to have the bicep dependencies (Alpine Linux) available and so fails with
+        # "ERROR: Error loading shared library ld-linux-x86-64.so.2: No such file or directory (needed by /root/.azure/bin/bicep)"
+        # This seems to be: https://github.com/Azure/bicep/issues/1683
+        # - We can just use a normal run command, but will need to keep an eye on the Azure CLI verion as we'll have that set in our runner. See https://github.com/Azure/cli/issues/11
+        # - Note that for the secrets with special characters, I have used the indirection recommended here:
+        # https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions#example-using-bash
```
OxonianCambion, line 190: Aaarrgh! I spent quite some time wading through all this!

.github/workflows/bicep_deploy.yml
```diff
@@ -0,0 +1,137 @@
+name: Deploy via Bicep
+
+# TODO:FN-72 consider concurrency
```
OxonianCambion, line 3: I'll address this shortly, but might need to think a bit more when we actually have multiple environments controlled by Bicep. Note that `bicep_workflow.yml` has a concurrency limiter anyway.

.github/workflows/bicep_az_deployment_group.yml
```diff
@@ -0,0 +1,232 @@
...
+        type: string
+      resourceGroupName:
+        required: true
+        type: string
+      action:
+        description: What action to perform - validate, what-if or create
+        required: true
+        type: string
+    secrets:
+      AZURE_CREDENTIALS:
```
OxonianCambion, line 18: It's a bit annoying having to list all the secrets repeatedly, but it should make adding the other environments or having different workflows much easier.

.github/workflows/bicep_az_deployment_group.yml
```diff
@@ -0,0 +1,232 @@
...
+      - uses: azure/login@v1
+        with:
+          creds: ${{ secrets.AZURE_CREDENTIALS }}
+
+      - name: Perform deployment ${{ needs.configure-command.outputs.command }} command
+        env:
+          # Note the leading space. We need this to stop Bicep what-if validation failing due to the "["
+          # of the string needing to be escaped if the string is not multi-line (which I believe adds leading spaces as a side effect).
+          # See https://stackoverflow.com/questions/74817321/how-to-escape-leading-square-bracket-in-a-parameter-array-in-bicep
+          # and the issue it mentions for details.
```
OxonianCambion, line 176: Aaargh! Again, very confusing, especially when in combination with the GH secret issue on the exact same parameter!

.github/workflows/bicep_deploy.yml
```diff
@@ -0,0 +1,137 @@
...
+        description: webapp environment variable
+        required: true
+      
+jobs:
+  validate:
+    uses: ./.github/workflows/bicep_az_deployment_group.yml
+    with:
+      environment: ${{ inputs.environment }}
+      resourceGroupName: ${{ inputs.resourceGroupName }}
+      action: validate
```
OxonianCambion, line 127: Validate is pretty handy - it will check permissions etc, and should give a good idea whether it will deploy correctly.
There is potential to use the `what-if` deploy instead (and I've implemented it), put the output on the PR and block the workflow until it has been approved.

.github/workflows/bicep_workflow.yml
```diff
@@ -0,0 +1,76 @@
+name: Deploy Infrastructure using Bicep
+concurrency: bicep-deploy
+
+on:
+  push:
+    branches:
+      - feat/FN-72/integrate-build-pipeline
```
OxonianCambion, line 7: Obviously I'll need to update this!
Note that the current plan is to reflect the behaviour for `dev` and have a different branch for infrastructure deployments, but I have a card to make it integrated and nicer.

infrastructure/resource_group_level/modules/cosmosdb.bicep
```diff
@@ -537,5 +538,21 @@ resource mongoDbPrivateEndpoint 'Microsoft.Network/privateEndpoints@2022-11-01'
   }
 }
 
+// Adding the Zone group sets up automatic DNS for the private link. 
+resource zoneGroup 'Microsoft.Network/privateEndpoints/privateDnsZoneGroups@2022-11-01' = {
```
OxonianCambion, line 542: Note that using Zone Groups is the recommended way of setting up A records, but I think I might have to rewrite this to use the manual method that the CLI has because it isn't quite identical. Worth thinking about.

abhi-markan (approved):

PR #2076, opened 2023-09-25
feat(FN-1133): deploy app code

OxonianCambion:
### Introduction

This card covers
* deploying the app code which is currently separate from deploying the infrastructure.
### Resolution
* app code is deployed in the same way as for the dev environment

### DIFF
--- a/.github/workflows/feature_deploy.yml
+++ b/.github/workflows/feature_deploy.yml
@@ -22,6 +22,10 @@ env:
   acr: ${{ secrets.ACR_REGISTRY_FEATURE }}
   acr_username: ${{ secrets.ACR_USERNAME_FEATURE }}
   acr_password: ${{ secrets.ACR_PASSWORD_FEATURE }}
+  # TODO: FN-72 NOTE: While we are using the dev GH environment, we can't use the RESOURCE_GROUP secret
+  # (It doesn't need to be a secret anyway.) So we define our resource group in an envioronment variable
+  # here and use that. The Azure region is the same so I've left that as using the secret.
+  resource_group: Digital-Feature
 
 jobs:
   deployment-environment:
@@ -86,7 +90,7 @@ jobs:
       with:
         inlineScript: |
           az configure --defaults location=${{ secrets.AZURE_REGION }}
-          az configure --defaults group=${{ secrets.RESOURCE_GROUP }}
+          az configure --defaults group=${{ env.resource_group }}
 
     - name: 7. Slot
       uses: azure/cli@v1.0.0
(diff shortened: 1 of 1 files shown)


.github/workflows/feature_deploy.yml
```diff
@@ -0,0 +1,585 @@
...
+        docker pull ${{ env.acr }}/${{ env.repository }}:$from
+        docker tag ${{ env.acr }}/${{ env.repository }}:$from ${{ env.acr }}/${{ env.repository }}:$to
+        docker push ${{ env.acr }}/${{ env.repository }}:$to
+
+    - name: 5. Azure login
+      uses: azure/login@v1
+      with:
+        creds: ${{ env.credentials }}
+
+    - name: 6. Slot
```
OxonianCambion, line 79: Note that it failed when I just used the same approach as the other deploy scripts - setting the default value for the resource group. When deploying multiple services at the same time, it seemed to not preserve the default resource group. Odd. Anyway, resolvable by putting it in the commands.

I suspect they have issues too?

.github/workflows/feature_deploy.yml
```diff
@@ -0,0 +1,585 @@
...
+          az config get defaults
+          az webapp deployment slot create --name tfs-${{ env.environment }}-${{ env.repository }} --resource-group ${{ env.resource_group }} --slot ${{ github.sha }} --configuration-source tfs-${{ env.environment }}-${{ env.repository }}
+          az webapp deployment slot swap --name tfs-${{ env.environment }}-${{ env.repository }} --resource-group ${{ env.resource_group }}  --slot ${{ github.sha }} --action swap
+          az webapp deployment slot delete --name tfs-${{ env.environment }}-${{ env.repository }} --resource-group ${{ env.resource_group }}  --slot ${{ github.sha }}
+    
+    - name: 7. Restart
+      uses: azure/cli@v1.0.0
+      with:
+        inlineScript: |
+          az webapp restart --name tfs-${{ env.environment }}-${{ env.repository }} --resource-group ${{ env.resource_group }}
```
OxonianCambion, line 92: I still don't understand why we restart?

.github/workflows/feature_deploy.yml
```diff
@@ -0,0 +1,585 @@
...
+  deployment-environment:
+    name: Environment setup
+    runs-on: [self-hosted, linux, deployment]
+    outputs:
+      environment: ${{ env.environment }}
+    steps:
+    - name: Initialise Environment
+      run: echo "Setting up ${{ env.environment }} environment"
+
+  #################### 1. PORTAL UI ####################
```
OxonianCambion, line 30: Note that with all the apps deploying in parallel, we currently get "429 - Too many requests" errors. See https://github.com/UK-Export-Finance/dtfs2/actions/runs/6303025076/job/17111424514

I'll raise another card to address this. Is this an issue when deploying to `dev`? I can't see how it wouldn't be!

abhi-markan (approved):

PR #2077, opened 2023-09-26
fix(cypress): GHA parallel execution

abhi-markan:
## Introduction
Ensure Cypress E2E tests are executed in a parallel fashion.

## Resolution
* Introduced GHA `matrix` strategy for parallel execution.
* Streamlined pipeline execution.

## Miscellaneous
* Lint fixes
* Spelling fixes
* Dependencies updates
* Consistent naming convention


.github/workflows/test.yml
```diff
@@ -273,9 +270,28 @@ jobs:
...
+            "admin/**/*.spec.js",
+            "checker/**/*.spec.js",
+            "dashboard/**/*.spec.js",
+            "feedback/**/*.spec.js",
+            "login/**/*.spec.js",
+            "maker/**/*.spec.js",
+            "number-generator/**/*.spec.js",
+            "read-only/**/*.spec.js",
+            "reports/**/*.spec.js",
+          ]
```
lmarrai-sw, line 290: Does this mean we need to remember to update this list whenever we add new tests or rename old tests? Is there any way we can make sure we've not missed any?

abhi-markan: Yes unfortunately, however in future this will be automated.
As long as we add under of the directories then we don't need to worry about it, only if we add a new directory.

lmarrai-sw (approved):

PR #2078, opened 2023-09-26
feat(FN-966): validate utilisation report content

ChristianMcc:
Want to finish a couple of tests on this so can hold off reviewing for a little bit then will need both Francesca and Abhi/Luke!
## Introduction
FN-966 and FN-1013. This PR addresses once the file is uploaded to the Portal server it is converted to csv and then has a set of data validation rules applied to it - if you need access to these please message me. 

## Resolution
It's fairly long winded to support both xlsx and csv files and provide cell addresses of where any validation error has been made. I'll leave some comments through the PR on any interesting parts.

## Screenshots

image
image


e2e-tests/portal/cypress/e2e/journeys/payment_officer/utilisation-report-upload/can-upload-utilisation-report.spec.js
```diff
@@ -0,0 +1,91 @@
...
+const MOCK_USERS = require('../../../../fixtures/users');
+const relativeURL = require('../../../relativeURL');
+
+const { BANK1_MAKER1 } = MOCK_USERS;
+
+context('Utilisation report upload', () => {
+  describe('Submitting a file to the utilisation report upload', () => {
+    it('Should route to the next page when a file is successfully upload', () => {
+      // TODO FN-965 update to payment officer role
+      cy.login(BANK1_MAKER1);
```
ChristianMcc, line 11: Update this file when the service options page is merged in

portal/server/controllers/utilisation-report-service/utilisation-report-upload/utilisation-report-cell-validators.js
```diff
@@ -0,0 +1,199 @@
...
+  if (!facilityIdObject?.value) {
+    return {
+      errorMessage: 'UKEF facility ID must have an entry',
+      column: facilityIdObject?.column,
+      row: facilityIdObject?.row || rowNumber,
+      value: facilityIdObject?.value,
+      exporter: exporterName,
+    };
+  }
+  if (!/^\d{8,10}$/.test(facilityIdObject?.value)) {
```
ChristianMcc, line 14: Going to move these regex's into the regex constants file

portal/server/controllers/utilisation-report-service/utilisation-report-upload/utilisation-report-validator.js
```diff
@@ -0,0 +1,139 @@
...
+  } else {
+    availableHeaders.push(HEADERS.MONTHLY_FEES_PAID);
+  }
+  return { missingHeaderErrors, availableHeaders };
+};
+
+const validateCsvCellData = (csvData, availableHeaders) => csvData.flatMap((value, index) => {
+  const csvRowDataErrors = [];
+
+  // If we have the UKEF Facility ID header, validate the UKEF Facility ID
```
ChristianMcc, line 77: I'll remove this

portal/server/utils/csv-utils.js
```diff
@@ -0,0 +1,137 @@
+const ExcelJS = require('exceljs');
```
ChristianMcc, line 1: My mind melted slightly going through all the iterations of data processing I needed in this file, if anyone wants to suggest better variable/function naming etc. please do!

portal/server/utils/csv-utils.js
```diff
@@ -0,0 +1,137 @@
...
+        .pipe(
+          csv({
+            mapHeaders: ({ header }) => header.toLowerCase().replace(/\s/g, ' ').trim(),
+            mapValues: ({ value }) => {
+              const lastDashIndex = value.lastIndexOf('-');
+              const cellValue = value.substring(0, lastDashIndex);
+              const cellAddress = value.substring(lastDashIndex + 1);
+
+              // Split the cell address into column and row
+              const [column, row] = cellAddress.match(/([A-Z]+)(\d+)/).slice(1);
```
ChristianMcc, line 60: Will move this regex to somewhere else, also I need to improve it as sonarcloud flagged it as being vulnerable to super-linear runtime due to backtracking

portal/server/utils/csv-utils.js
```diff
@@ -0,0 +1,137 @@
...
+  return new Promise((resolve, reject) => {
+    try {
+      const jsonParsedData = [];
+      csvStream
+        .pipe(
+          csv({
+            mapHeaders: ({ header }) => header.toLowerCase().replace(/\s/g, ' ').trim(),
+            mapValues: ({ value }) => {
+              const lastDashIndex = value.lastIndexOf('-');
+              const cellValue = value.substring(0, lastDashIndex);
```
ChristianMcc, line 56: need to gracefully handle a failure case to split here

portal/templates/utilisation-report-service/utilisation-report-upload/check-the-report.njk
```diff
@@ -0,0 +1,102 @@
...
+                type="file"
+                accept=".xls, .xlsx, .csv"/>
+        </div>
+        <div class="govuk-button-group">
+            <input
+                formaction="?_csrf={{ csrfToken }}"
+                type="submit"
+                class="govuk-button govuk-!-margin-right-1"
+                data-module="govuk-button"
+                value="Continue"
```
lmarrai-sw, line 98: It'd be worth checking with a designer if something along the lines of "Upload" or "Upload and continue" would be better text for the button (here and on the previous screen)

portal/templates/utilisation-report-service/utilisation-report-upload/check-the-report.njk
```diff
@@ -0,0 +1,102 @@
...
+                                {% else %}
+                                    -
+                                {% endif %}
+                            </td>
+                            <td class="govuk-table__cell">
+                                {% if item.column !== null %}
+                                    {{ item.column }}
+                                {% else %}
+                                    -
+                                {% endif %}
```
lmarrai-sw, line 63: Are `item.exporter`, `item.row`, `item.column` never undefined? Could/should we use e.g. `not item.exporter` instead?

lmarrai-sw (commented):
I've only looked at the template file so far, but I won't have time for the rest of the PR today so thought I'd get my comments on the template down anyway :)

portal/server/utils/csv-utils.js
```diff
@@ -0,0 +1,138 @@
...
+      const { csvData, csvDataWithCellAddresses } = parseXlsxToCsvArrays(worksheet);
+
+      fileBuffer = Buffer.from(csvData);
+      csvJson = await xlsxBasedCsvToJsonPromise(csvDataWithCellAddresses);
+    });
+  } else if (file.mimetype === 'text/csv') {
+    fileBuffer = file.buffer;
+    csvJson = await csvBasedCsvToJsonPromise(file.buffer);
+  }
+  return { csvJson, fileBuffer };
```
lmarrai-sw, line 130: Should we error here if the mimetype isn't one of the 2 we can process?

portal/server/utils/csv-utils.test.js
```diff
@@ -0,0 +1,72 @@
...
+
+describe('csv-utils', () => {
+  describe('columnIndexToExcelColumn', () => {
+    it('returns the correct column for an index below 26', async () => {
+      const excelColumnIndex = columnIndexToExcelColumn(2);
+
+      expect(excelColumnIndex).toBe('C');
+    });
+
+    it('returns the correct column for an index below 26', async () => {
```
lmarrai-sw, line 11:
```suggestion
    it('returns the correct column for an index above 26', async () => {
```

portal/server/utils/csv-utils.js
```diff
@@ -0,0 +1,138 @@
...
+const csvBasedCsvToJsonPromise = async (csvBuffer) => {
+  const csvStream = Readable.from(csvBuffer);
+  return new Promise((resolve, reject) => {
+    try {
+      const csvData = [];
+      csvStream
+        .pipe(
+          csv({
+            mapHeaders: ({ header }) => header.toLowerCase().replace(/\s/g, ' ').trim(),
+            mapValues: ({ index, value }) => ({ value, column: columnIndexToExcelColumn(index), row: null }),
```
lmarrai-sw, line 88: I'm curious why the row values are `null` here, but it might make sense when I read the rest of the code

portal/server/controllers/utilisation-report-service/utilisation-report-upload/utilisation-report-cell-validators.js
```diff
@@ -0,0 +1,200 @@
...
+      errorMessage: 'Total fees accrued for the month must be a number',
+      column: totalFeesAccruedObject?.column,
+      row: totalFeesAccruedObject?.row || rowNumber,
+      value: totalFeesAccruedObject?.value,
+      exporter: exporterName,
+    };
+  }
+  if (totalFeesAccruedObject?.value?.length > 15) {
+    return {
+      errorMessage: 'Total fees accrued for the month must be 15 characters or less',
```
lmarrai-sw, line 101: It might be worth making the 15 a more prominent constant? You can re-use it in the error message as well

portal/server/controllers/utilisation-report-service/utilisation-report-upload/utilisation-report-cell-validators.js
```diff
@@ -0,0 +1,200 @@
...
+      errorMessage: 'Monthly fees paid to UKEF must be a number',
+      column: monthlyFeesPaidObject?.column,
+      row: monthlyFeesPaidObject?.row || rowNumber,
+      value: monthlyFeesPaidObject?.value,
+      exporter: exporterName,
+    };
+  }
+  if (monthlyFeesPaidObject?.value.length > 15) {
+    return {
+      errorMessage: 'Monthly fees paid to UKEF must be 15 characters or less',
```
lmarrai-sw, line 132: Same comment as above about 15 :)

portal/server/controllers/utilisation-report-service/utilisation-report-upload/utilisation-report-cell-validators.js
```diff
@@ -0,0 +1,200 @@
...
+      errorMessage: 'Exchange rate must be a number',
+      column: csvDataRow[HEADERS.EXCHANGE_RATE]?.column,
+      row: csvDataRow[HEADERS.EXCHANGE_RATE]?.row || rowNumber,
+      value: csvDataRow[HEADERS.EXCHANGE_RATE]?.value,
+      exporter: csvDataRow[HEADERS.EXPORTER]?.value,
+    };
+  }
+  if (csvDataRow[HEADERS.EXCHANGE_RATE]?.value.length > 15) {
+    return {
+      errorMessage: 'Exchange rate must be 15 characters or less',
```
lmarrai-sw, line 182: Same comment as above about 15

portal/server/controllers/utilisation-report-service/utilisation-report-upload/utilisation-report-cell-validators.test.js
```diff
@@ -0,0 +1,581 @@
...
+          column: 3,
+        },
+      };
+
+      const exchangeRateError = generateExchangeRateError(csvDataRow);
+
+      expect(exchangeRateError).toEqual(null);
+    });
+
+    it('returns an error if payment currency is the different to base currency and exchange rate is null', async () => {
```
lmarrai-sw, line 470:
```suggestion
    it('returns an error if payment currency is different to base currency and exchange rate is null', async () => {
```

portal/server/controllers/utilisation-report-service/utilisation-report-upload/utilisation-report-cell-validators.test.js
```diff
@@ -0,0 +1,581 @@
...
+        value: null,
+        exporter: testExporterName,
+      };
+
+      const exchangeRateError = generateExchangeRateError(csvDataRow);
+
+      expect(exchangeRateError).toEqual(expectedError);
+    });
+
+    it('returns an error if payment currency is the different to base currency and exchange rate is not a number', async () => {
```
lmarrai-sw, line 507:
```suggestion
    it('returns an error if payment currency is different to base currency and exchange rate is not a number', async () => {
```

portal/server/controllers/utilisation-report-service/utilisation-report-upload/utilisation-report-cell-validators.test.js
```diff
@@ -0,0 +1,581 @@
...
+        value: 'abc',
+        exporter: testExporterName,
+      };
+
+      const exchangeRateError = generateExchangeRateError(csvDataRow);
+
+      expect(exchangeRateError).toEqual(expectedError);
+    });
+
+    it('returns an error if payment currency is the different to base currency and exchange rate is more than 15 characters', async () => {
```
lmarrai-sw, line 544:
```suggestion
    it('returns an error if payment currency is different to base currency and exchange rate is more than 15 characters', async () => {
```

portal/server/controllers/utilisation-report-service/utilisation-report-upload/utilisation-report-cell-validators.test.js
```diff
@@ -0,0 +1,581 @@
...
+  generateUkefFacilityIdError,
+  generateBaseCurrencyError,
+  generateFacilityUtilisationError,
+  generateTotalFeesAccruedError,
+  generateMonthlyFeesPaidError,
+  generatePaymentCurrencyError,
+  generateExchangeRateError,
+} = require('./utilisation-report-cell-validators');
+
+describe('utilisation-report-cell-validators', () => {
```
lmarrai-sw, line 11: Great job on the thorough tests

I definitely think it would be nicer if each of these describe blocks had their own test file, to make this file easier to manage and more readable

Honestly I think the `utilisation-report-cell-validators.js` file is too long as well so you could split those into separate files as well, e.g.

```
utilisation-report-cell-validators
|- generate-ukef-facility-id-error.js
|- generate-ukef-facility-id-error.test.js
|- generate-base-currency-error.js
|- generate-base-currency-error.test.js
// etc.
```

portal/server/controllers/utilisation-report-service/utilisation-report-upload/utilisation-report-validator.js
```diff
@@ -0,0 +1,140 @@
+// Errors can look like:
```
lmarrai-sw, line 1: Do we still need this?

portal/server/controllers/utilisation-report-service/utilisation-report-upload/utilisation-report-validator.js
```diff
@@ -0,0 +1,140 @@
...
+    missingHeaderErrors.push({
+      errorMessage: 'Monthly fees paid to UKEF header is missing or spelt incorrectly',
+      column: null,
+      row: null,
+      value: null,
+      exporter: null,
+    });
+  } else {
+    availableHeaders.push(HEADERS.MONTHLY_FEES_PAID);
+  }
```
lmarrai-sw, line 70: I think it'd be good to refactor this to reduce duplication

Maybe something like

```
requiredHeaders.forEach(({header, errorMessageIfMissing}) => {
  if (!headers.includes(header)) {
    missingHeadersErrors.push({
      errorMessage: errorMessageIfMissing,
      column: null,
      row: null,
      value: null,
      exporter: null,
    });
  } else {
    availableHeaders.push(header);
  }
});
```

portal/server/controllers/utilisation-report-service/utilisation-report-upload/utilisation-report-validator.js
```diff
@@ -0,0 +1,140 @@
...
+  return { missingHeaderErrors, availableHeaders };
+};
+
+const validateCsvCellData = (csvData, availableHeaders) => csvData.flatMap((value, index) => {
```
lmarrai-sw, line 74: This function would be good to reduce duplication in as well, maybe we can do something similar to the above comment

portal/server/controllers/utilisation-report-service/utilisation-report-upload/utilisation-report-validator.test.js
```diff
@@ -0,0 +1,125 @@
...
+      expect(missingHeaderErrors.length).toBe(2);
+      expect(missingHeaderErrors[0].errorMessage).toBe('UKEF facility ID header is missing or spelt incorrectly');
+    });
+  });
+
+  describe('validateCsvCellData', () => {
+    // This test mocks out all the function from utilisation-report-cell-validators.js and
+    // tests that if headers are available then the respective cell validator function is called on that data
+
+    it('Calls the generate error functions for headers that are present', async () => {
```
lmarrai-sw, line 80:
```suggestion
    it('calls the generate error functions for headers that are present', async () => {
```

portal/server/controllers/utilisation-report-service/utilisation-report-upload/utilisation-report-validator.test.js
```diff
@@ -0,0 +1,125 @@
...
+
+      const availableHeaders = [HEADERS.UKEF_FACILITY_ID, HEADERS.BASE_CURRENCY];
+
+      validateCsvCellData(csvData, availableHeaders);
+      expect(generateUkefFacilityIdError).toHaveBeenCalledWith({ value: '20001371', column: 'B', row: 1 }, 'Britannia Energy Ltd', 1);
+      expect(generateBaseCurrencyError).toHaveBeenCalledWith({ value: 'GBP', column: 'D', row: 1 }, 'Britannia Energy Ltd', 1);
+      expect(generateFacilityUtilisationError).not.toHaveBeenCalled();
+    });
+
+    it('Calls the generate payment currency and exchange rate error functions even if no headers are present', async () => {
```
lmarrai-sw, line 103:
```suggestion
    it('calls the generate payment currency and exchange rate error functions even if no headers are present', async () => {
```

portal/server/utils/csv-utils.js
lmarrai-sw: Okay my understanding from looking at the code is:
- the validators want to know which row caused the validation to fail, so we can tell the user where the error is
- `csv-parser` doesn't give us the row number in mapValues, so we can't work out the row number here and set it to `null` instead
- when the validators validate the cell, we pass in the `rowNumber` which we've worked out from the JSON array index instead, which lines up with the actual CSV row number
- for XLSX based CSVs, the JSON array index does **not** match the actual row number because XSLXs can have empty rows, so in that case we have to use `row` instead

Is that right? It's a bit tricky to work out from the code if so, is there any way we can simplify it, e.g. can weset the row number of the csv based JSONs in `csvBasedCsvToJsonPromise` somehow instead of leaving them `null` and having to work out the row number later?

portal/server/constants/regex.js
```diff
@@ -3,3 +3,12 @@ const { FILE_UPLOAD } = require('./file-upload');
 exports.COMPANIES_HOUSE_NUMBER_REGEX = /^(([A-Z]{2}|[A-Z]\d{1}|\d{2})(\d{5,6}|\d{4,5}[A-Z]))$/;
 
 exports.FILE_NAME_REGEX = new RegExp(`.+\\.(${FILE_UPLOAD.ALLOWED_FORMATS.join('|')})$`);
+
+// This is flagged as a dangerous regex but is only used on excel cell addresses and not from user input and so is not vulnerable to a DoS attack
+exports.CELL_ADDRESS_REGEX = /([A-Z]+)(\d+)/;
```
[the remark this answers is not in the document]
ChristianMcc: Good suggestion I should just do this :)

portal/templates/utilisation-report-service/utilisation-report-upload/check-the-report.njk
ChristianMcc: Just following designs here but yes we definitely need a content designer for these screens to come review them

portal/templates/utilisation-report-service/utilisation-report-upload/check-the-report.njk
ChristianMcc: They can be undefined depending on the type of error, I tried using `item.column | Default('-')` but was having no luck with it

portal/server/constants/regex.js
```diff
@@ -3,3 +3,12 @@ const { FILE_UPLOAD } = require('./file-upload');
...
+
+exports.UKEF_FACILITY_ID_REGEX = /^\d{8,10}$/;
```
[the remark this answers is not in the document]
ChristianMcc: excel files are dumb and if you have two 0s at the front of the number it will get rid of them, so it's sometimes 8 as they normally start 00XXXXXXXX

portal/jest.config.js
```diff
@@ -1,14 +1,10 @@
...
-    'server/routes/**/*.{js,}',
-    'server/helpers/**/*.{js,}',
-    'scripts/**/*.{js,}',
-  ],
+  collectCoverageFrom: ['server/controllers/**/*.{js,}', 'server/routes/**/*.{js,}', 'server/helpers/**/*.{js,}', 'scripts/**/*.{js,}'],
   coverageDirectory: 'generated_reports/coverage/unit',
   testMatch: ['**/*.test.js', '**/*.component-test.js'],
   moduleNameMapper: {
     '^.+\\.(css|less|scss)$': 'babel-jest',
+    '^exceljs$': 'babel-jest',
```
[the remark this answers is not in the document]
ChristianMcc: The file uses a certain type of imports that means compiling for jest tests fails without using babel-jest for this npm package - I can find a link for this later if you want :)

portal/component-tests/utilisation-report-service/utilisation-report-upload.component-test.js
```diff
@@ -0,0 +1,44 @@
...
+        { errorMessage: 'Facility utilisation must be a number', column: 'C', row: 3, value: 'abc', exporter: 'test exporter' },
+      ],
+      errorSummary: [
+        {
+          text: 'You must correct these errors before you can upload the report',
+          href: '#utilisation-report-file-upload',
+        },
+      ],
+      filename: 'test-file.xlsx',
+    };
```
francescastocco, line 40: Is it worth pulling this out and re-using it for both tests?

francescastocco (changes requested):
Looks good!! Sorry if I've duplicated some of Luke's comments..

portal/server/controllers/utilisation-report-service/utilisation-report-upload/index.js
```diff
@@ -20,29 +23,47 @@ const postUtilisationReportUpload = async (req, res) => {
...
+      ];
+      return res.render('utilisation-report-service/utilisation-report-upload/check-the-report.njk', {
+        validationErrors: csvValidationErrors,
+        errorSummary,
+        filename: req.file.originalname,
+        user: req.session.user,
+        primaryNav: 'utilisation_report_upload',
+      });
+    }
+    req.session.utilisation_report = { fileBuffer, fileName: req.file.originalname };
```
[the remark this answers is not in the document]
ChristianMcc: 1. Yes we will send it away to an API later on (not implemented)
2. Yes our sessions get cleared from redis automatically, I also plan to manually clear the file from the session once it's uploaded

portal/server/utils/csv-utils.js
ChristianMcc: You're pretty much correct in everything you've said. The only way I can think of getting row number in the csvBasedCsvToJsonPromise function is to iterate through the array and add it on in a separate loop (which is what I do later on to get the row number anyway). This would possibly be a bit neater/easier to read at the cost of a bit of performance (that should be completely negligible in this case). I can give that a go and see if it works out nicely!

portal/server/controllers/utilisation-report-service/utilisation-report-upload/index.js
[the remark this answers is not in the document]
lmarrai-sw: Nice, sounds good

portal/server/constants/regex.js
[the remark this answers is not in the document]
lmarrai-sw: Eurgh that's annoying. That might be worth a comment in the code?

portal/jest.config.js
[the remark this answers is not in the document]
lmarrai-sw: Interesting! Thanks

lmarrai-sw (approved):
Thanks @ChristianMcc I love the changes. The only outstanding things from me are:
- do we want to leave a comment in the code about the Excel weirdness that means the facility id regex allows 8 - 10 digits?
- do we want to extract a constant for `15` since a lot of the validators seem to use 15 as the maximum number of characters?
:)

PR #2083, opened 2023-09-27
feat(FN-1113): matrix deploy

OxonianCambion:
### Introduction

This card covers
* deploying application code via a matrix strategy to limit parallelism and avoid throttling errors
### Resolution
* application code is now deployed via a matrix strategy with limited parallelism.

### DIFF
--- a/.github/workflows/feature_deploy.yml
+++ b/.github/workflows/feature_deploy.yml
@@ -110,129 +110,78 @@ jobs:
         with:
           inlineScript: |
             az webapp restart --name tfs-${{ env.environment }}-${{ matrix.repository }} --resource-group ${{ env.resource_group }}
-        
-  #################### 8. Numgen ####################
-  deploy-numgen:
-    needs: deployment-environment
-    name: Number generator
+
+  deploy-functionapps:
+    # We depend on deploy-webapps to reduce the chances of being throttled
+    needs: [deployment-environment, deploy-webapps]
+    name: Deploy Functionapps
     runs-on: [self-hosted, linux, deployment]
+    # TODO: FN-1077 NOTE: we don't currently have a `feature` GitHub environment yet so use the `dev` one while we're not doing much.
+    # Restore the parameterised version throughout!
     environment:
       dev
-    env:
-      repository: azure-function-number-generator
-      path: azure-functions/number-generator-function
-      name: function-number-generator
-      
+    # environment:
+    #   name: ${{ needs.deployment-environment.outputs.environment }}
+    strategy:
+      matrix:
+        repository: 
+          - azure-function-number-generator
+          - azure-function-acbs
+        include:
+          - path: azure-functions/number-generator-function
+            repository: azure-function-number-generator
+            name: function-number-generator
+          - path: azure-functions/acbs-function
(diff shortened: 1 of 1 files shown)


.github/workflows/feature_deploy.yml
```diff
@@ -37,559 +37,151 @@ jobs:
...
+          - portal-ui
+          - portal-api
+          - gef-ui
+          - trade-finance-manager-ui
+          - trade-finance-manager-api
+          - dtfs-central-api
+          - external-api
+        include:
+          - path: portal
+            repository: portal-ui
```
OxonianCambion, line 62: We need this as the `portal-ui` webapp lives in the `portal` folder. However, GHA has quite a neat augmentation mechanism.

.github/workflows/feature_deploy.yml
```diff
@@ -37,559 +37,151 @@ jobs:
...
     # TODO: FN-1077 NOTE: we don't currently have a `feature` GitHub environment yet so use the `dev` one while we're not doing much.
     # Restore the parameterised version throughout!
     environment: dev
     # environment:
     #   name: ${{ needs.deployment-environment.outputs.environment }}
-    env:
-      repository: portal-ui
-      
+    strategy:
+      max-parallel: 4
```
OxonianCambion, line 50: I've tested a deploy with this (and making the functionapps dependent on the webapp job) and it succeeds.
The code's a lot neater too.

I considered combining the functionapps with the webapps, which wouldn't be especially difficult, but with them going away at some point, it is perhaps overkill.

.github/workflows/feature_deploy.yml
```diff
@@ -37,559 +37,151 @@ jobs:
...
-      uses: azure/cli@v1.0.0
-      with:
-        inlineScript: |
-          az functionapp restart --name tfs-${{ env.environment }}-${{ env.name }} --resource-group ${{ env.resource_group }}
-
-  #################### 9. ACBS ####################
-  deploy-acbs:
-    needs: deployment-environment
-    name: ACBS
+      - name: 1. Latest commit hash
```
OxonianCambion, line 64: The only difference is that I need to use `matrix.<blah>` for the matrix parameters.

.github/workflows/feature_deploy.yml
```diff
@@ -37,559 +37,151 @@ jobs:
...
-          az functionapp deployment slot swap --name tfs-${{ env.environment }}-${{ env.name }} --resource-group ${{ env.resource_group }} --slot ${{ github.sha }} --action swap
-          az functionapp deployment slot delete --name tfs-${{ env.environment }}-${{ env.name }} --resource-group ${{ env.resource_group }} --slot ${{ github.sha }}
-
-    - name: 7. Restart
-      uses: azure/cli@v1.0.0
-      with:
-        inlineScript: |
-          az functionapp restart --name tfs-${{ env.environment }}-${{ env.name }} --resource-group ${{ env.resource_group }}
\ No newline at end of file
+      - name: 1. Latest commit hash
```
OxonianCambion, line 139: This is weird.
It seems the original version didn't have the steps indented. I wonder how it worked?

PR #2143, opened 2023-10-02
feat(FN-1203): disable function-acbs in non-prod environments

OxonianCambion:
### Introduction

This card covers
* deploying function-acbs in a disabled state
### Resolution
* adds `enabled` status to `function-acbs` and sets it per-environment

### DIFF
--- a/infrastructure/resource_group_level/modules/function-acbs.bicep
+++ b/infrastructure/resource_group_level/modules/function-acbs.bicep
@@ -1,6 +1,8 @@
 param location string
 param environment string
-param enabled bool
+
+@allowed(['Stopped', 'Running'])
+param state string
 param containerRegistryName string
 param appServicePlanEgressSubnetId string
 param appServicePlanId string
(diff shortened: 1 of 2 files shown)


infrastructure/resource_group_level/modules/function-acbs.bicep
```diff
@@ -83,6 +84,7 @@ resource functionAcbs 'Microsoft.Web/sites@2022-09-01' = {
   tags: {}
   kind: 'functionapp,linux,container'
   properties: {
+    enabled: enabled
```
OxonianCambion, line 87: Note that when deployed this produces the Azure Function App in the `disabled` state. (Not `stopped`).
It isn't clear if the individual functions might still run. They are set to `enabled` though I hope that the overall App being disabled prevents them executing. 

I suspect this means that one can't enable it in the UI and will need to redeploy!

There doesn't seem an obvious way to enable / Start it in the portal. Weirdly, you can click "Stop" and it reports success, but the UI still shows the Stop button, not Start. 

I'm not really sure how to test this. Looking at the metrics, the memory working set is at zero while disabled, so it looks like it is inactive!

Is all this ok?

infrastructure/resource_group_level/modules/function-acbs.bicep
```diff
@@ -85,6 +88,7 @@ resource functionAcbs 'Microsoft.Web/sites@2022-09-01' = {
   properties: {
     httpsOnly: false
     serverFarmId: appServicePlanId
+    state: state
```
OxonianCambion, line 91: Further to the previous comment, I dug deeper and found that actually the documentation is a combination of wrong and missing.

We can set the state, even though the REST interface and Azure docs say we can't.

I've raised and issue with the Azure CLI:
https://github.com/Azure/azure-rest-api-specs/issues/26097

abhi-markan (approved):

PR #2144, opened 2023-10-02
fix(e2e-tests): seed random company names to prevent 3% chance of collision that causes test failure

lmarrai-sw:
## Introduction

@AlexBramhill and I noticed that the `dashboard-deals-filter-on-wrong-pagination-page.spec.js` test in the `portal` e2e tests would randomly fail on PRs that hadn't affected it at all, so we investigated it.

It creates 30 deals with random company names (using `chance.company()` from the `chance` package. It then filters the dashboard by searching by the first company name that was generated, assuming that this should return exactly 1 deal (the first deal).

However, `chance.company()` picks a random company name from a list of roughly 900. That gives ~3% chance that at least one of the other 29 deals have exactly the same company name as the first deal, meaning the test will fail because the dashboard filter will return (at least) 2 deals instead of 1.

## Resolution

I've introduced a new class, `RandomValueGenerator`, for the e2e tests that uses a seeded version of `chance`. This way each test run should be repeatable by using the exact same random company names (assuming the test code and the `chance` code doesn't change). This should eliminate the 3% chance that the test fails, which was a source of flakey e2e test results.

Note there is a chance that in the future we edit the test in such a way that the seeded random value generator will end up generating the same company name for another deal as it does for the first. If this happens, it will happen 100% of the time, so it should be easy to debug. We can then pick a different seed for `chance` to fix the issue.

### DIFF
--- a/e2e-tests/.eslintrc.js
+++ b/e2e-tests/.eslintrc.js
@@ -37,7 +37,7 @@ module.exports = {
     ],
   },
   parserOptions: {
-    ecmaVersion: 12, // es2021
+    ecmaVersion: 14,
   },
   globals: {
     cy: true,
--- a/e2e-tests/portal/cypress/e2e/journeys/dashboard/dashboard-deals/dashboard-deals-sort.spec.js
+++ b/e2e-tests/portal/cypress/e2e/journeys/dashboard/dashboard-deals/dashboard-deals-sort.spec.js
@@ -25,7 +25,7 @@ context('Dashboard Deals filters - filter by multiple fields with multiple value
         let companyName = '';
         // sets one company to lowercase
         if (index === 3) {
-          companyName = randomValueGenerator.companyName().toLowerCase();
+          companyName = randomValueGenerator.companyName({ lowerCase: true });
         } else {
           companyName = randomValueGenerator.companyName();
         }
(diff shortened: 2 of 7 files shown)


e2e-tests/.eslintrc.js
```diff
@@ -37,7 +37,7 @@ module.exports = {
     ],
   },
   parserOptions: {
-    ecmaVersion: 12, // es2021
+    ecmaVersion: 14,
```
lmarrai-sw, line 40: We need ecma script 13 support for the private class properties I have used in `RandomValueGenerator` (the hashes)

e2e-tests/portal/cypress/e2e/journeys/dashboard/dashboard-deals/dashboard-deals-sort.spec.js
```diff
@@ -25,9 +25,9 @@ context('Dashboard Deals filters - filter by multiple fields with multiple value
         let companyName = '';
         // sets one company to lowercase
         if (index === 3) {
-          companyName = chance.company().toLowerCase();
+          companyName = randomValueGenerator.companyName().toLowerCase();
```
abhi-markan, line 28: Can `toLowerCase()` be added as an argument please.

e2e-tests/support/random-value-generator.js
```diff
@@ -0,0 +1,15 @@
+import { Chance } from 'chance';
+
+export class RandomValueGenerator {
```
abhi-markan, line 3: Can we please add some documentation from PR description for context.

AlexBramhill (approved):

abhi-markan (changes requested):
Great work, few comments to address please.

abhi-markan:
One minor comment regarding commit name, please can it match the standard. Thanks

abhi-markan (approved):
LGTM 👍🏼  Good work

PR #2146, opened 2023-10-03
feat(FN-927): wire up tfm-ui url

OxonianCambion:
### Introduction

This card covers
* don't use secret to resolve circular dependency on TFM-UI URL connection string
### Resolution
* we now set the connection strings after the TFM-FD is deployed

### DIFF
--- a/infrastructure/resource_group_level/modules/front-door-tfm.bicep
+++ b/infrastructure/resource_group_level/modules/front-door-tfm.bicep
@@ -140,3 +140,4 @@ resource frontDoorTfm 'Microsoft.Network/frontdoors@2021-06-01' = {
   }
 }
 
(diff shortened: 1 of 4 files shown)


infrastructure/resource_group_level/modules/front-door-tfm.bicep
```diff
@@ -140,3 +140,4 @@ resource frontDoorTfm 'Microsoft.Network/frontdoors@2021-06-01' = {
   }
 }
 
+output defaultHostName string = frontDoorTfm.properties.frontendEndpoints[0].properties.hostName
```
OxonianCambion, line 143: Note that this will correspond to the default value of `tfs-feature=tfm-fd.azurefd.net` and not any custom hostname. I've checked `prod` and it uses the equivalent url.

abhi-markan (approved):

PR #2147, opened 2023-10-05
feat(DTFS2-6715): migrate get team members in tfm to rest

lmarrai-sw:
## Introduction

We want to replace TFM API's TeamMembers GraphQL query with a REST endpoint.

## Resolution

In TFM API, I've created a new endpoint (`GET /v1/teams/{teamid}/members`) to get members of a team.
This validates that `{teamId}` is one of the allowed team ids, then calls DTFS Central API's `GET /v1/users/team/{teamId}` and returns the `_id`, `firstName`, and `lastName` of the results.

I have changed TFM UI to use the new rest endpoint instead of the GraphQL query, and tested the UI pages that use this functionality still work (I've commented this functionality on the ticket)


trade-finance-manager-api/api-tests/common-tests/client-authentication-tests.js
```diff
@@ -0,0 +1,31 @@
+const expectNotAuthenticatedResponse = ({
+  status,
+  body
+}) => {
+  expect(status).toBe(401);
+  expect(body).toStrictEqual({});
+};
+
+const withClientAuthenticationTests = ({
```
AlexBramhill, line 9: Very nice

trade-finance-manager-api/src/v1/teams/routes.js
```diff
@@ -0,0 +1,75 @@
...
+ *                     type: string
+ *                 required:
+ *                   - _id
+ *                   - firstName
+ *                   - lastName
+ *       400:
+ *         description: Bad Request.
+ */
+teamsRoutes.route('/teams/:teamId/members').get(
+  param('teamId').isIn(allTeamIds).withMessage(`teamId must be one of ${allTeamIds.join(', ')}`),
```
AlexBramhill, line 58: Do we not want to use our fancy new validator?

lmarrai-sw: @AlexBramhill and I spoke about this and I have extracted this validator to a new file

AlexBramhill (approved):

PR #2151, opened 2023-10-05
Feat(FN-966): handle weird excel data that strips cell address

ChristianMcc:
## Introduction
Cell addresses weren't present on excel files with some inbuilt data from Bing in the cells. While we don't need to support this we need to avoid crashes as it wasn't getting picked up by the try/catch

## Resolution
Only split up the cell address to get column and row if we match the regex of the cell address


francescastocco (approved):
Nice fix :)

PR #2152, opened 2023-10-06
feat(fn 997): file type validation and virus scanning

ChristianMcc:
## Introduction
We want to bring in virus scanning and validation that the files are the right size/type and aren't password protected.

## Resolution
Introduce a new clamav server used for virus scanning (with the clamscan npm package to interact with it). Add virus scanning middleware which adds an error to `res.locals` for controllers to display the correct error with.
Also use an error which clamav returns for password protected files to return to the user if the file is pw protected (it returns an error as it means it cannot be scanned).

NOTE: I can't false positive test the virus scanning code in an e2e test as it requires an EICAR signature file which cannot be a .csv or .xlsx file. This means that the mimetype checking will refuse the file before it gets to the virus scanning section

This should be merged alongside any ifnrastructure changes for the deployed clamav server


portal/server/controllers/utilisation-report-service/utilisation-report-upload/index.js
```diff
@@ -31,6 +31,14 @@ const getUploadErrors = (req, res) => {
...
     ];
     uploadValidationError = { text: 'You must upload a file' };
+  } else if (res?.locals?.virusScanFailed) {
+    uploadErrorSummary = [
+      {
+        text: 'The selected file could not be uploaded - try again',
+        href: '#utilisation-report-file-upload',
+      },
+    ];
+    uploadValidationError = { text: 'The selected file could not be uploaded – try again' };
```
francescastocco, line 41: Do you think it's worth making this error message more explicit that it failed the virus scan?

.github/workflows/feature_deploy.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main-stbr
+      - feat/FN-997/file-type-validation-and-virus-scanning
```
ChristianMcc, line 7: need to delete this

portal/server/routes/middleware/virusScanUpload/index.js
```diff
@@ -0,0 +1,78 @@
...
+            reject(error);
+          })
+          .on('timeout', (error) => {
+            const timeoutError = error || new Error('Scan timed out');
+            console.error('Clamav virus scan timed out');
+            reject(timeoutError);
+          });
+      });
+
+      if (scanResult?.isInfected === true) {
```
francescastocco, line 55: Can this just be 
```suggestion
      if (scanResult?.isInfected) {
```

francescastocco (changes requested):
Looks good, just a couple of minor things

portal/server/controllers/utilisation-report-service/utilisation-report-upload/index.js
ChristianMcc: If it's essentially a server error in back end I don't really want to give away what's going on/wrong when it might be that nothing is wrong with the file so have left it as generic for now

portal/server/routes/middleware/virusScanUpload/index.js
ChristianMcc: Thought about this - I don't really know what the package returns here. It's supposed to be a boolean but sometimes its null/undefined when it errors so I'm worried about it returning something dumb like a string and me treating it as true when I shouldn't. Pure paranoia but wanted to be safe as possible

ChristianMcc: Good shout, I'll do it in my other PR so I can merge this, busy afternoon :D

PR #2153, opened 2023-10-06
feat(FN-980): list previous reports backend

francescastocco:
## Introduction
We need to allow users to see a list of previously submitted reports.

## Resolution
- Endpoint in DTFS Central API to get DB collection `utilisation-reports`
- Endpoint in Portal API which calls above endpoint and maps data to format required in the frontend
- In Portal call above endpoint
- Started writing some tests but would be easier with the endpoint to create documents in this collection

## To Do
- Finish off tests
- Add some validation for bank ID


portal/server/api.js
```diff
@@ -797,6 +797,24 @@ const getUkefDecisionReport = async (token, payload) => {
   }
 };
 
+const getPreviousUtilisationReportsByBank = async (token, bankId) => {
+  try {
+    const response = await axios({
+      method: 'get',
+      url: `${PORTAL_API_URL}/v1/previous-reports/${bankId}`,
```
ChristianMcc, line 804: Same here for some extra validation on the bankID to check it's the right format

portal-api/src/v1/controllers/utilisation-report-service/previous-reports.controller.js
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  return groupedReports;
+};
+
+const getPreviousReportsByBankId = async (req, res) => {
+  try {
+    const { bankId } = req.params;
+
+    // Add validation for user bank ID here?
```
ChristianMcc, line 66: Can see you've added a comment here so ye just chuck it in a route validation and then in the api files

dtfs-central-api/src/v1/routes/portal-routes.js
```diff
@@ -870,4 +872,6 @@ portalRouter.route('/cron-jobs').delete(cronJobsController.deleteAllEstoreLogs);
 portalRouter.route('/gef/mandatory-criteria/latest').get(mandatoryCriteria.getLatestGefMandatoryCriteria);
 portalRouter.route('/gef/mandatory-criteria/version/:version').get(mandatoryCriteria.getGefMandatoryCriteriaByVersion);
 
+portalRouter.route('/previous-reports/:bankId').get(previousReports.getUtilisationReports);
```
ChristianMcc, line 875: Can we add a validation on the bankId here too? it'll be the same as in portal API just checking the format of it

portal-api/src/v1/routes.js
```diff
@@ -205,4 +207,6 @@ authRouter.get('/validate', (req, res) => {
 // bank-validator
 authRouter.get('/validate/bank', (req, res) => banks.validateBank(req, res));
 
+authRouter.route('/previous-reports/:bankId').get(validateUserHasAtLeastOneAllowedRole({ allowedRoles: [PAYMENT_OFFICER] }), getPreviousReportsByBankId);
```
ChristianMcc, line 210: It's missing  from the rest of this file like in the other route but we should validate bankId here with the validator package just as in central API. I think TFM API has an example of how we can write some validations there

dtfs-central-api/src/v1/controllers/portal/utilisation-report-service/previous-reports.controller.js
```diff
@@ -0,0 +1,28 @@
...
+      $sort: {
+        year: 1,
+        month: 1,
+      },
+    },
+  ]).toArray();
+  if (filteredAndSortedUtilisationReports) {
+    return res.status(200).send(filteredAndSortedUtilisationReports);
+  }
+  return res.status(404).send({ status: 404, message: 'No utilisation reports found' });
```
ChristianMcc, line 23: I'm not sure 404 is right here, 404 to me is we've hit a route that doesn't exist.
If there are no reports isfilteredAndSortedUtilisationReports `[]` or `null` or `undefined`? If [] then that will be truthy in javascript and we'll return the 200 and empty array which is fine.
If it's not [] then we can assume an error happened (do we need a try/catch?) and we can send back 500 and Error finding reports?

dtfs-central-api/src/v1/routes/portal-routes.js
```diff
@@ -870,4 +872,6 @@
...
```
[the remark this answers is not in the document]
ChristianMcc: Is there global rate-limiting on this that means this is fine? the DTFS team implemented some rate limiting a few weeks ago so we should check with them if we can ignore this or if we need to do anything :)

ChristianMcc (changes requested):
Some initial thoughts, still going through the grouping logic in portal-API but will come back to it later :)

dtfs-central-api/src/v1/controllers/portal/utilisation-report-service/previous-reports.controller.js
francescastocco: Yes yes, good point!

dtfs-central-api/src/v1/routes/portal-routes.js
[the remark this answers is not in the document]
francescastocco: Yeh the DTFS team did implement this, I will test locally but probably a false positive

portal-api/src/v1/controllers/utilisation-report-service/previous-reports.controller.js
```diff
@@ -0,0 +1,66 @@
...
+      month: getMonthName(report.month),
+      path: report.path
+    }));
+  return {
+    year,
+    reports: reportsByYear,
+  };
+});
+
+const populateOmittedYears = (reportsGroupedByYear, years) => {
```
ChristianMcc, line 29: Minor nit-pick, can this return the array back so we reassign to a new variable in the function below please? little more obvious than a side effect of changing the array passed in etc.

ChristianMcc (commented):
Looks good to me, I have one nit-pick that if you fix quickly then I'll approve :)
Agreed adding in missing years is a bit annoying but think your curent solution here is a good one!

portal-api/src/v1/controllers/utilisation-report-service/previous-reports.controller.js
francescastocco: On it

PR #2154, opened 2023-10-06
fix(FN-966): fix back button on upload page

ChristianMcc:
## Introduction 
The back button was routed to '/utilisation-report-service' instead of '/utilisation-report-upload'

## Resolution
Fix the routing to '/utilisation-report-upload'


francescastocco (approved):
Good spot

PR #2155, opened 2023-10-07
feat(FN-972): file type validation on re upload

ChristianMcc:
## Introduction
We need the file type validation on the check the report page
NOTE: Don't merge this before FN-997!

## Resolution
The file type validation was already there but we were missing the frontend logic of the upload errors on the check-the-report page.

image


francescastocco (dismissed):
Nice :)

francescastocco (approved):

PR #2157, opened 2023-10-09
feat(FN-857): update waf policy application

OxonianCambion:
### Introduction

This card covers
* Enabling the `prod` environment to have the WAF rules applied to Portal without the IP restriction
### Resolution
* Create a second set of WAF rules, only on `prod` and apply them to Portal

### DIFF
--- a/infrastructure/resource_group_level/modules/waf-policies.bicep
+++ b/infrastructure/resource_group_level/modules/waf-policies.bicep
@@ -12,9 +12,21 @@ param matchVariable string
 param rejectAction string
 param applyWafRuleOverrides bool = false
 param restrictAccessToUkefIps bool
+
+type RuleSetType = 'DefaultRuleSet' | 'Microsoft_DefaultRuleSet'
+type RuleSet = {
+  ruleSetType: RuleSetType
+  ruleSetVersion: string
+}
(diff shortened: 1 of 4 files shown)


.github/workflows/bicep_workflow_feature.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - feat/DTFS2-6422
+      - feat/FN-857/update-waf-policy
```
OxonianCambion, line 7: I'll remove this before merging!

infrastructure/resource_group_level/modules/waf-policies.bicep
```diff
@@ -12,9 +12,21 @@ param matchVariable string
 param rejectAction string
 param applyWafRuleOverrides bool = false
 param restrictAccessToUkefIps bool
+
+type RuleSetType = 'DefaultRuleSet' | 'Microsoft_DefaultRuleSet'
+type RuleSet = {
+  ruleSetType: RuleSetType
+  ruleSetVersion: string
+}
+param ruleSet RuleSet
```
OxonianCambion, line 21: Bicep now supports Types!

abhi-markan (approved):

PR #2160, opened 2023-10-12
feat(dtfs2-6716):replace graphql update task in tfm with rest

AlexBramhill:
Introduction
We want to replace TFM API's update task GraphQL query with a REST endpoint.

Resolution
In TFM API, I've created a new endpoint (PUT /deals/:dealId/tasks/:groupId/:taskId). This has existing dealId validation applied to it, as well as additional basic groupId validation applied.


trade-finance-manager-api/src/v1/validation/route-validators/route-validators.js
```diff
@@ -5,10 +5,8 @@ const userParamValidation = param('user').isMongoId().withMessage('The User ID (
 const facilityIdValidation = param('facilityId').isMongoId().withMessage('The Facility ID (facilityId) provided should be a Mongo ID');
 const amendmentIdValidation = param('amendmentId').isMongoId().withMessage('The Amendment ID (amendmentId) provided should be a Mongo ID');
 const dealIdValidation = param('dealId').isMongoId().withMessage('The Deal ID (dealId) provided should be a Mongo ID');
-const partyURNValidation = param('urn')
-  .isString()
-  .matches(/^\d+$/)
-  .withMessage('The party URN (urn) provided should be a string of numbers');
+const groupIdValidation = param('groupId').isInt().withMessage('The Group ID (groupId) provided should be an integer');
```
AlexBramhill, line 8: This is some very basic validation added in here -- Looking at the code, I can tell that groupId always has to be an int.

lmarrai-sw (approved):
thanks @AlexBramhill

PR #2165, opened 2023-10-16
feat(mongodb): Fix vulnerable version of `mongodb`

abhi-markan:
## Introduction ✏️ 
Ensure `mongodb` vulnerable version is no longer in use.

## Resolution ✔️ 
* Upgraded `mongodb` to `4.17.0` from `4.2.0` as per GH advisory

## Miscellaneous ➕ 
* Dependencies updates


ChristianMcc:
This is a little annoying considering cosmosdb is currently deployed on version 4.2 but if it's to do with how the package interacts with the DB maybe having differing versions won't cause any issues. 
It looks like we were on 4.2.0 before but looking at https://www.mongodb.com/docs/manual/release-notes/4.2/ is 4.2.24 safer and hopefully more in line with what's deployed on azure?

dtfs-central-api/package.json
```diff
@@ -38,7 +38,7 @@
     "express-rate-limit": "^6.11.2",
     "moment": "^2.29.4",
     "mongo-dot-notation": "3.1.0",
-    "mongodb": "4.2.0",
+    "mongodb": "4.17.0",
```
lmarrai-sw, line 41: I'm a bit confused here: the description says we're downgrading to 4.2.0 but this looks like we're upgrading to 4.17.0

lmarrai-sw (approved):

PR #2180, opened 2023-10-18
feat(FN-997): add in correct check for clamav scanning being enabled

ChristianMcc:
## Introduction
We couldn't disable clamav scanning on feature env as the env variable check was wrong

## Resolution
Fix it to check that it matches on 'true' as env variables are strings


francescastocco (approved):
We had some config files on APIM to validate our env variables, I can't really remember all the reasons this was better but do you think it's worth introducing to dtfs2?

PR #2186, opened 2023-10-19
Feat(FN-997): improve password protected errors

ChristianMcc:
## Introduction
We were returning something went wrong for some errors that including password protected files

## Resolution
For these errors we want to return and error message on the page. This might have the copy udpated in the future


francescastocco (dismissed):
Looks good!

francescastocco (approved):

PR #2188, opened 2023-10-23
fix(DTFS2-6746): do not remove parties after updating bond beneficiary or issuer urns

lmarrai-sw:
## Introduction

After TFM UI updates the URN for the bond issuer/beneficiary, it calls `api.updateParty(dealId, deal.parties);` where `deal` is the response from `GET deals/:dealId` from TFM API (see `postTfmFacility` in TFM-UI for this). It does this because when TFM API receives updates a party, it then checks if the deal can be created in ACBS which relies on the parties that require URNs having URNs set and (for BSS/EWCS deals) the bond issuer and beneficiary having URNs set.

However `deal.parties` is always `undefined` on the response from `GET deals/:dealId` from TFM API.

The GraphQL endpoints that we had before DTFS2-6182 meant that no change was made to the deal's parties when we call `api.updateParty(dealId, undefined);` (but the logic to create the deal in ACBS still worked), but the REST endpoints we changed to in DTFS2-6182 delete the deal's parties when we call `api.updateParty(dealId, undefined);`,

This causes a bug that deletes the deal's party info after the bond issuer/beneficiary URNs are set.

## Resolution

I've changed the TFM UI to call `api.updateParty(dealId, deal.tfm.parties);` where `deal` is the response from `GET deals/:dealId` from TFM API, since `deal.tfm.parties` are the deal's parties.

I've added 2 e2e tests that check the exporter URN is not deleted when the bond issuer/beneficiary URNs are set

### DIFF
--- a/e2e-tests/tfm/cypress/e2e/journeys/case/parties/bond-beneficiary.spec.js
+++ b/e2e-tests/tfm/cypress/e2e/journeys/case/parties/bond-beneficiary.spec.js
@@ -208,6 +208,29 @@ context('Bond beneficiary URN - User can add, edit, confirm and submit URN to th
 
         pages.partiesPage.bondBeneficiaryEditLink().should('exist');
       });
+
+      it('should not remove the exporter party URN when the beneficiary URNs are submitted', () => {
+        const exporterUrn = CONSTANTS.PARTY_URN.ANOTHER_VALID;
+        submitExporterUrn(exporterUrn);
+        pages.partiesPage.bondBeneficiaryEditLink().click();
+        pages.bondBeneficiaryPage.urnInput(1).clear();
+        pages.bondBeneficiaryPage.urnInput(2).clear();
+        pages.bondBeneficiaryPage.urnInput(1).type(partyUrn[0]);
+        pages.bondBeneficiaryPage.urnInput(2).type(partyUrn[1]);
+
+        pages.bondBeneficiaryPage.saveButton().click();
+        pages.bondBeneficiaryPage.saveButton().click();
+
+        pages.partiesPage.exporterUrn().should('contain', exporterUrn);
+      });
+
+      function submitExporterUrn(urn) {
+        pages.partiesPage.exporterEditLink().click();
+        pages.exporterPage.urnInput().clear();
+        pages.exporterPage.urnInput().type(urn);
+        pages.exporterPage.saveButton().click();
+        pages.exporterPage.saveButton().click();
+      }
     });
 
     describe('when the TFM user is NOT in `BUSINESS_SUPPORT` team', () => {
--- a/e2e-tests/tfm/cypress/e2e/journeys/case/parties/bond-issuer.spec.js
+++ b/e2e-tests/tfm/cypress/e2e/journeys/case/parties/bond-issuer.spec.js
@@ -208,6 +208,29 @@ context('Bond issuer URN - User can add, edit, confirm and submit URN to the TFM
 
         pages.partiesPage.bondIssuerEditLink().should('exist');
       });
+
+      it('should not remove the exporter party URN when the issuer URNs are submitted', () => {
+        const exporterUrn = CONSTANTS.PARTY_URN.ANOTHER_VALID;
+        submitExporterUrn(exporterUrn);
+        pages.partiesPage.bondIssuerEditLink().click();
+        pages.bondIssuerPage.urnInput(1).clear();
+        pages.bondIssuerPage.urnInput(2).clear();
+        pages.bondIssuerPage.urnInput(1).type(partyUrn[0]);
+        pages.bondIssuerPage.urnInput(2).type(partyUrn[1]);
+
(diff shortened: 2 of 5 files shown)


e2e-tests/tfm/cypress/e2e/journeys/case/parties/bond-beneficiary.spec.js
```diff
@@ -208,6 +208,29 @@ context('Bond beneficiary URN - User can add, edit, confirm and submit URN to th
...
+        const exporterUrn = CONSTANTS.PARTY_URN.ANOTHER_VALID;
+        submitExporterUrn(exporterUrn);
+        pages.partiesPage.bondBeneficiaryEditLink().click();
+        pages.bondBeneficiaryPage.urnInput(1).clear();
+        pages.bondBeneficiaryPage.urnInput(2).clear();
+        pages.bondBeneficiaryPage.urnInput(1).type(partyUrn[0]);
+        pages.bondBeneficiaryPage.urnInput(2).type(partyUrn[1]);
+
+        pages.bondBeneficiaryPage.saveButton().click();
+        pages.bondBeneficiaryPage.saveButton().click();
```
oscar-richardson-softwire, line 222: Why do we do this twice?

e2e-tests/tfm/cypress/e2e/journeys/case/parties/bond-beneficiary.spec.js
```diff
@@ -208,6 +208,29 @@ context('Bond beneficiary URN - User can add, edit, confirm and submit URN to th
...
+
+        pages.partiesPage.exporterUrn().should('contain', exporterUrn);
+      });
+
+      function submitExporterUrn(urn) {
+        pages.partiesPage.exporterEditLink().click();
+        pages.exporterPage.urnInput().clear();
+        pages.exporterPage.urnInput().type(urn);
+        pages.exporterPage.saveButton().click();
+        pages.exporterPage.saveButton().click();
```
oscar-richardson-softwire, line 232: Again, why do we do this twice?

e2e-tests/tfm/cypress/e2e/journeys/case/parties/bond-issuer.spec.js
```diff
@@ -208,6 +208,29 @@ context('Bond issuer URN - User can add, edit, confirm and submit URN to the TFM
...
+        const exporterUrn = CONSTANTS.PARTY_URN.ANOTHER_VALID;
+        submitExporterUrn(exporterUrn);
+        pages.partiesPage.bondIssuerEditLink().click();
+        pages.bondIssuerPage.urnInput(1).clear();
+        pages.bondIssuerPage.urnInput(2).clear();
+        pages.bondIssuerPage.urnInput(1).type(partyUrn[0]);
+        pages.bondIssuerPage.urnInput(2).type(partyUrn[1]);
+
+        pages.bondIssuerPage.saveButton().click();
+        pages.bondIssuerPage.saveButton().click();
```
oscar-richardson-softwire, line 222: Why twice?

e2e-tests/tfm/cypress/e2e/journeys/case/parties/bond-issuer.spec.js
```diff
@@ -208,6 +208,29 @@ context('Bond issuer URN - User can add, edit, confirm and submit URN to the TFM
...
+
+        pages.partiesPage.exporterUrn().should('contain', exporterUrn);
+      });
+
+      function submitExporterUrn(urn) {
+        pages.partiesPage.exporterEditLink().click();
+        pages.exporterPage.urnInput().clear();
+        pages.exporterPage.urnInput().type(urn);
+        pages.exporterPage.saveButton().click();
+        pages.exporterPage.saveButton().click();
```
oscar-richardson-softwire, line 232: Why twice?

e2e-tests/tfm/cypress/e2e/journeys/case/parties/bond-beneficiary.spec.js
lmarrai-sw: The first time you click save takes you to a summary page that asks you to confirm your input. The second time you click save does the actual saving

e2e-tests/tfm/cypress/e2e/journeys/case/parties/bond-beneficiary.spec.js
lmarrai-sw: The first time you click save takes you to a summary page that asks you to confirm your input. The second time you click save does the actual saving

e2e-tests/tfm/cypress/e2e/journeys/case/parties/bond-issuer.spec.js
lmarrai-sw: The first time you click save takes you to a summary page that asks you to confirm your input. The second time you click save does the actual saving

oscar-richardson-softwire (approved):
LGTM :)

PR #2189, opened 2023-10-23
feat(DTFS2-6743): send email on login

oscar-richardson-softwire:
### Introduction

As part of implementing 2FA for Portal UI, we need to email a sign in link to a user when they login. This PR adds the logic to do this.

### Resolution

Added the logic to send the email (and added a new template for the email in Notify). This logic is currently turned on/off with a feature flag, and the sign in link is currently a placeholder.

Screenshot of the email:

image


portal-api/src/v1/users/login.controller.js
```diff
@@ -33,3 +35,12 @@ module.exports = (username, password) =>
...
   });
+
+module.exports.sendSignInLinkEmail = async (emailAddress, firstName, signInLink) => sendEmail(
+  CONSTANTS.EMAIL_TEMPLATE_IDS.SIGN_IN_LINK,
+  emailAddress,
+  {
+    firstName,
+    signInLink,
+  },
+);
```
lmarrai-sw, line 46: Can we make the "30 minutes" expiry timeframe a parameter as well please @oscar-richardson-softwire? We'll want to make that a constant in our code so we can easily change the expiry time

portal-api/src/constants/email-template-ids.js
```diff
@@ -3,11 +3,13 @@ const BLOCKED = '82506983-cb85-4f33-b962-922b850be7ac';
 const UNBLOCKED = '44959d08-6389-4f27-a6be-2faae8bea711';
 const PASSWORD_UPDATE = '41235821-7e52-4d63-a773-fa147362c5f0';
 const UPDATE_STATUS = '718beb52-474e-4f34-a8d7-ab0e48cdffce';
+const SIGN_IN_LINK = '2eab0ad2-eb92-43a4-b04c-483c28a4da18';
```
lmarrai-sw, line 6: Thanks for the screenshot.

Are we going to need to do any extra styling work to make `placeholderSignInLink` display as a link when we implement it, or will it happen automatically if it is a URL?

lmarrai-sw: I've added a question for Adam to the Figma designs because I couldn't tell from the designs how we wanted the URL to be formatted

lmarrai-sw, line 6: I've just realised that
> If you did not create an account for UKEF export insurance, please contact us.
is definitely _not_ the right text - that's talking about EXIPS instead of DTFS. Hopefully when you get a response about the footer it will resolve this too - if not then we should get some new text for that

portal-api/src/v1/users/login.controller.js
oscar-richardson-softwire: Added

portal-api/src/constants/email-template-ids.js
oscar-richardson-softwire: Adam's decided that we want to have link text link so: [Click here to sign in to your account](url), which solves this problem. I've made this change to the template

oscar-richardson-softwire: This has been changed to `If you did not create an account for UKEF Digital Trade Finance, please contact us.`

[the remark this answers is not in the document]
oscar-richardson-softwire: I've asked Adam both of these questions, will let you know when I have a response

portal-api/unit.jest.config.js
```diff
@@ -1,5 +1,5 @@
 module.exports = {
   collectCoverageFrom: ['src/**/*.{js,}'],
   coverageDirectory: 'generated_reports/coverage/unit',
-  testMatch: ['**/*.test.js'],
+  testMatch: ['**/login.controller.test.js'],
```
lmarrai-sw, line 4:
```suggestion
  testMatch: ['**/*.test.js'],
```
I think this was a mistake :)

portal-api/src/v1/users/routes.js
```diff
@@ -186,6 +186,22 @@ module.exports.remove = (req, res, next) => {
...
+    if (status === 201) {
+      // TODO DTFS2-6680: Add success logic here.
+    } else {
+      // TODO DTFS2-6680: Add failure logic here.
+    }
+  } else {
+    next(usernameOrPasswordIncorrect);
+  }
+};
+module.exports.sendSignInLinkEmailAndHandleErrors = sendSignInLinkEmailAndHandleErrors;
```
lmarrai-sw, line 203: Do we only export this just so we can test it?

portal-api/src/v1/users/routes.js
```diff
@@ -209,20 +225,7 @@ module.exports.login = async (req, res, next) => {
...
-        if (status === 201) {
-          // TODO DTFS2-6680: Add success logic here.
-        } else {
-          // TODO DTFS2-6680: Add failure logic here.
-        }
-      } else {
-        next(usernameOrPasswordIncorrect);
-      }
-    });
+    findByUsername(username, sendSignInLinkEmailAndHandleErrors(next));
```
lmarrai-sw, line 228: Do we not need to `await` `findByUsername`?

portal-api/src/v1/email/index.test.js
```diff
@@ -0,0 +1,94 @@
+const mockNotifyClientSendEmail = jest.fn();
+
+jest.mock('notifications-node-client', () => ({
+  NotifyClient: jest.fn().mockImplementation(() => ({ sendEmail: mockNotifyClientSendEmail })),
+}));
+
+const sendEmail = require('./index');
+
+describe('sendEmail', () => {
```
lmarrai-sw, line 9: Thanks for adding these tests @oscar-richardson-softwire, nice improvement to the code :)

portal-api/src/v1/users/routes.js
lmarrai-sw: @oscar-richardson-softwire and I spoke about this on a call: the answer is yes, and we discussed how we could test it by through `login` by mocking `findByUsername` instead. We're not going to do so in this PR because this code is about to change anyway

portal-api/src/v1/users/routes.js
lmarrai-sw: I spoke to @oscar-richardson-softwire on a call about this and the answer is yes but we're not going to do so in this PR because this code is about to change anyway

lmarrai-sw (dismissed):

lmarrai-sw (dismissed):

portal-api/src/constants/email-template-ids.js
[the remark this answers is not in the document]
lmarrai-sw: Adam responded that we want to use `firstName` and `lastName`, which means a code change. We're going to merge this as-is and then make a new PR for the change

lmarrai-sw (approved):

PR #2192, opened 2023-10-24
feat(FN-1294): add payment report officer nav items to shared nav bar

toddym42:
## Introduction
- use the standard DTFS header for utilisation report pages
- add the utilisation report nav items to the main nav bar for users with the `'payment-report-officer'` role.
  - Report GEF utilisation and fees
  - Previous GEF reports

## Resolution
- removed the custom `utilisation-report-service/index.njk` page wrapper and replaced use with the generic `index.njk`. This means all pages now use the same header
- refactored `'payment-officer'` role to `'payment-report-officer'`, and added permission to all projects
- refactored logic in `portal/templates/_partials/primary-navigation.njk` that determines which nav items are displayed. Now checks each nav item to see if you have permission to view it, where previously we rendered a different set based on the role (or roles) the user had. This prevents the need to have to think about combinations of roles, and the order in which we'd need to check the combinations to see what should be rendered


portal/component-tests/partials/primary-navigation.component-test.js
```diff
@@ -1,50 +1,86 @@
...
+    itDoesNotRenderAReportsLink();
+    itDoesNotRenderAUsersLink();
+  });
+
+  describe('viewed with no roles', () => {
+    beforeAll(() => {
+      wrapper = render({ user: { roles: [] } });
+    });
+
+    itDoesNotRenderAHomeLink();
```
ChristianMcc, line 80: In previous behaviour a user would have a home link only, while I'm not sure if this is right or wrong I'd prefer to not change existing behaviour so can we keep that logic please?

utils/mock-data-loader/portal/users.js
```diff
@@ -1,174 +1,172 @@
...
-} = require('./roles');
+const ROLES = require('./roles');
 
 const UKEF_TEST_BANK_1 = BANKS.find((bank) => bank.name === 'UKEF test bank (Delegated)');
 const UKEF_TEST_BANK_2 = BANKS.find((bank) => bank.name === 'UKEF test bank (Delegated) 2');
 const UKEF_GEF_ONLY_BANK = BANKS.find((bank) => bank.name === 'GEF Only Bank');
 
-const USERS = [
-  {
+const USERS = {
```
ChristianMcc, line 8: big fan of this change, nice one :) 
In insert-mocks.js we iterate over this USERS array and as it's now an object we'll have to iterate over Object.entries or something here to turn this object into an array to insert all the users into the DB.
Hard to test without having a local setup but if you check to fix there first can check your branch out and test it if you don't have a working setup by then!

ChristianMcc (changes requested):
Nice work! Small behaviour question and then need to fix the 'mock-data-loader' (the seeding of data) with your change to the array -> object which I like

portal/component-tests/partials/primary-navigation.component-test.js
```diff
@@ -1,50 +1,86 @@
-const { ADMIN, MAKER, CHECKER } = require('../../server/constants/roles');
-const { NON_MAKER_OR_CHECKER_OR_ADMIN_ROLES } = require('../../test-helpers/common-role-lists');
+const ROLES = require('../../server/constants/roles');
```
JoshBinns2000, line 1: Minor change but maybe this could be replaced with `const { ROLES } = require('../../server/constants')`

ChristianMcc:
There's a few failing portal tests to do with access that are probably from the rename. Personally I'd wait to debug them with a real setup before looking into them so you can test this.
As far as I'm concerned this is good to go with a couple of tests to fix so when you have a setup and can test this ticket will approve and can merge then!
@JoshBinns2000 you can rebase on top of this as it is now to get these changes if you haven't as I don't foresee many other changes happening :)

toddym42:
> There's a few failing portal tests to do with access that are probably from the rename. Personally I'd wait to debug them with a real setup before looking into them so you can test this. As far as I'm concerned this is good to go with a couple of tests to fix so when you have a setup and can test this ticket will approve and can merge then! @JoshBinns2000 you can rebase on top of this as it is now to get these changes if you haven't as I don't foresee many other changes happening :)

@ChristianMcc - I'd looked into the test failures and apparently they're all existing ones, and I believe they're already in the process of being fixed separately. I have my access now, so will double-check for sure, and check that this is all working as we'd expect.

ChristianMcc (approved):

ChristianMcc (approved):

PR #2194, opened 2023-10-27
feat(FN-1363): add role specific login routing

JoshBinns2000:
## Introduction
After a successful login attempt, a user who *only* has the `'payment-report-officer'` role should be redirected to the 'Report GEF utilisation and fees' page. Otherwise, they should be redirected to 'Home'.

## Resolution
- added a login controller to abstract redirecting behaviour into its own function
- updated the `portal/server/helpers/getUserRoles.test.js` function to test the newly added roles and `null` or `undefined` input arguments
- updated e2e tests for the login journey


portal/server/helpers/getUserRoles.test.js
```diff
@@ -11,14 +11,28 @@ describe('getUserRoles', () => {
       });
 
       it(`should return false when roles does NOT include ${role}`, () => {
-        const result = getUserRoles(['']);
+        const result = getUserRoles([]);
```
JoshBinns2000, line 14: I'm not convinced about the way this test works. I think it could probably be replaced with a single test that just makes sure all the results are false if an empty array is passed in. Further, this test sounds more like you should pass in an array which includes all roles **expect** the role currently being tested. I'm curious to know what the rest of you think

portal/server/helpers/getUserRoles.js
```diff
@@ -1,6 +1,16 @@
 const { MAKER, CHECKER, PAYMENT_REPORT_OFFICER, ADMIN, READ_ONLY } = require('../constants/roles');
 
+const defaultResult = {
```
francescastocco, line 3: This is a good addition, a minor thing but wonder if it's worth renaming to make it explicit that the default has no roles? Eg. `noRolesDefault`

portal/server/helpers/getUserRoles.test.js
francescastocco, line 1: Nice job on improving these tests

francescastocco (changes requested):
This looks good, thanks! Can you check the e2e tests too as I think there are some where the test itself is unrelated to this change but there is an assertion for where the user is redirected.

I think we should also be able to delete the functionality related to the service options page along with this ticket

e2e-tests/portal/cypress/e2e/journeys/login/login.spec.js
```diff
@@ -65,10 +71,34 @@ context('Login', () => {
...
+  it('A successful login with read-only role takes the user to the /dashboard/deals/0 page', () => {
+    cy.login(BANK1_READ_ONLY1);
+
+    cy.url().should('eq', relative('/dashboard/deals/0'));
+  });
+
+  it('A successful login with payment report officer role takes the user to the /utilisation-report-upload page', () => {
+    cy.login(BANK1_PAYMENT_REPORT_OFFICER1);
+
+    cy.url().should('eq', relative('/utilisation-report-upload'));
```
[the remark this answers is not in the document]
ChristianMcc: This could still be an it.each(), although not very worried about it I'd personally approve and leave it up to you!

ChristianMcc (approved):
LGTM! Nice work :) might want to wait for francesca to have a second review before merging as she is closer to this

francescastocco (dismissed):
Nice!

e2e-tests/portal/cypress/e2e/journeys/login/login.spec.js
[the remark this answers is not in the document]
JoshBinns2000: We discussed this in person and I forgot to add a comment. Cypress doesn't support `it.each` by default, but there does exist a node package to do it if we ever wanted to (see https://github.com/bahmutov/cypress-each)

francescastocco (changes requested):
Looks good to me - just one more thing, can you delete the `service-options` files too as we don't need these and shouldn't refer to them anywhere else

francescastocco (approved):
Nice well done 🎉

PR #2195, opened 2023-10-27
feat(FN-967): upload file to azure and db

ChristianMcc:
## Introduction
We need to persist the data and save the data in the utilisation report to the database.

## Resolution
Post the data through to the central API to save the details of the report uploaded and the data in the report.


dtfs-central-api/src/v1/validation/utilisation-report-service/utilisation-report-validator.js
```diff
@@ -0,0 +1,94 @@
...
+/**
+ * Validates the month of the utilisation report. Returns null if valid, otherwise returns an error message.
+ * @param {Integer} month - Month of utilisation report.
+ * @returns {String | null} - Error message or null if valid.
+ */
+const validateMonth = (month) => {
+  if (!month) {
+    return 'Month is required';
+  }
+  if (month < 1 || month > 12) {
```
toddym42, line 22: Could add some safety here to reject non-numbers, or numbers with decimal place. e.g.
```
if (!Number.isInteger(month) || month < 1 || month > 12) {
```
Same comment for the year below.

dtfs-central-api/src/v1/validation/utilisation-report-service/utilisation-data-validator.js
```diff
@@ -0,0 +1,133 @@
+const validator = require('validator');
+const REGEXES = require('../../../constants/regex');
+
+/**
+ * Validates the UKEF ID to match the UKEF facility ID regex, returns an error message or null if valid.
+ * @param {Integer | String | null} ukefId - ukef facility ID.
```
toddym42, line 6: For all the validators, we could have the value we're checking as `unknown`. This better represents what we know about the value, and should also prompt us to check for the required type before calling any functions on it.

dtfs-central-api/src/v1/validation/utilisation-report-service/utilisation-report-validator.js
```diff
@@ -0,0 +1,94 @@
...
+  }
+  if (year < 2020 || year > 2100) {
+    return 'Year must be between 2020 and 2100';
+  }
+  return null;
+};
+
+/**
+ * Validates the file path of the utilisation report in azure. Returns null if valid, otherwise returns an error message.
+ * @param {Integer} filePath - file path of the utilisation report in azure.
```
toddym42, line 45: Expected type here is `string` instead of `Integer`. As mentioned elsewhere though, maybe we could have this as `unknown`.

portal-api/src/v1/controllers/utilisation-report-service/utilisation-report-upload.controller.js
```diff
@@ -1,29 +1,43 @@
...
+
+    if (saveDataResponse.status !== 201) {
+      const status = saveDataResponse.status || 500;
+      console.error('Failed to save utilisation report: %O', saveDataResponse);
+      return res.status(status).send({ status, data: 'Failed to save utilisation report' });
+    }
+
+    sendEmailToPdcInputtersEmail(parsedUser.bank.bankName, month, year);
+
+    return res.status(200).send({ status: 200, data: 'Successfully saved utilisation report' });
```
toddym42, line 36: is there any reason to return 200 instead of the original 201 response here?

It doesn't look like we usually include a `status` attribute for success responses. Should we instead return the body of the original success response (i.e. `{ dateUploaded: <some_date_value>}`)

portal/server/api.js
```diff
@@ -797,22 +797,38 @@ const getUkefDecisionReport = async (token, payload) => {
...
+    const formData = new FormData();
+    formData.append('reportData', JSON.stringify(csvData));
+
+    formData.append('user', JSON.stringify(uploadingUser));
+    formData.append('month', month);
+    formData.append('year', year);
+
+    // TODO get the csvFileBuffer from the csv in the user's session
+    const x = [1, 2, 3];
+    const buffer = Buffer.from(x);
```
toddym42, line 811: Still TODO as part of this ticket?

portal/server/api.js
```diff
@@ -797,22 +797,38 @@ const getUkefDecisionReport = async (token, payload) => {
...
+      data: formData.getBuffer(),
+      maxContentLength: Infinity,
+      maxBodyLength: Infinity,
     });
 
-    return response.data;
+    return response;
   } catch (error) {
-    console.error('Unable to upload utilisation report %s', error);
+    console.error('Unable upload utilisation report %s', error);
```
toddym42, line 831: This looks like it was removed by mistake.

portal/server/controllers/utilisation-report-service/utilisation-report-upload/index.js
```diff
@@ -96,8 +96,14 @@ const postUtilisationReportUpload = async (req, res) => {
         primaryNav: 'utilisation_report_upload',
       });
     }
-    // TODO FN-970 Populate month, year
```
toddym42, line 99: Should this comment remain here? It looks like we're still hard-coding the month and year values.

dtfs-central-api/src/v1/validation/utilisation-report-service/utilisation-report-validator.js
```diff
@@ -0,0 +1,94 @@
...
+      return 'File path must be a string';
+    }
+  }
+  return null;
+};
+
+/**
+ * Validates the utilisation report data. Returns an array of error messages.
+ * @param {Array} utilisationReportData - array of json objects representing utilisation report data.
+ * @returns {Array} - Array of string error messages.
```
toddym42, line 63: We could be more specific about the types here, using `Array<object>` and `Array<string>`. It looks like the `object[]` and `string[]` is more commonly used in the project though.

dtfs-central-api/src/constants/db_collections.js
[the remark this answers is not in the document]
ChristianMcc: snake case isn't one at all so not sure how i did this, we're doing kebab-case mostly but will change this to camelCase just for the constants files

dtfs-central-api/src/constants/regex.js
```diff
@@ -0,0 +1,5 @@
+exports.UKEF_FACILITY_ID_REGEX = /^\d{8,10}$/;
+
+exports.CURRENCY_NUMBER_REGEX = /^\d+(\.\d{1,2})?$/;
```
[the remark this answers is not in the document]
ChristianMcc: Tempted to go with no because I want loose validation in central API, but this could be a thing in portal when we first do the validation.
I'm mostly doing data validation from a security perspective in central API rather than business logic validation but happy to chat about this

dtfs-central-api/src/services/repositories/utilisation-data-repo.js
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * Saves the data from the utilisation report to the database.
+ * @param {Object} reportData - Array of data that has been turned to json objects.
+ * @param {Integer} month - Month of utilisation report, integer between 1 and 12.
+ * @param {Integer} year - Year of utilisation report, integer greater than 2020.
+ * @param {Object} bank - Object representing bank the report belongs to.
+ * @param {String} reportId - Id of the report details database document.
+ * @returns {Object} - result of the insertMany save operation.
+ */
```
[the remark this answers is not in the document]
ChristianMcc: Good shout, not going ot reply to all the JSDoc style comments but will fix all :)

dtfs-central-api/src/v1/controllers/utilisation-report-service/upload-utilisation-report.controller.js
```diff
@@ -0,0 +1,45 @@
...
+const { saveUtilisationReportDetails } = require('../../../services/repositories/utilisation-reports-repo');
+const {
+  validateUtilisationReportData,
+  validateMonth,
+  validateYear,
+  validateFilePath,
+} = require('../../validation/utilisation-report-service/utilisation-report-validator');
+
+const validatePayload = (reportData, month, year, filePath) => {
+  const validationErrors = validateUtilisationReportData(reportData);
```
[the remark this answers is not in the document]
ChristianMcc: Yeah one thing we might be looking at but haven't sold on is using express-validator. Will be happy to refactor if we do as it's a lot nicer than this

dtfs-central-api/src/services/repositories/utilisation-data-repo.test.js
```diff
@@ -0,0 +1,50 @@
+const { saveUtilisationData } = require('./utilisation-data-repo');
+const db = require('../../drivers/db-client');
+const { DB_COLLECTIONS } = require('../../constants/db_collections');
+
+describe('utilisation-data-repo', () => {
+  it('returns null when a correct month is provided', async () => {
```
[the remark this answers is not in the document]
ChristianMcc: classic copy paste, oops

dtfs-central-api/src/v1/controllers/utilisation-report-service/upload-utilisation-report.controller.js
```diff
@@ -0,0 +1,45 @@
...
+    const { bank } = user;
+
+    // If the are any data type errors in the report data, return 400
+    const validationErrors = validatePayload(reportData, month, year, filePath);
+    if (validationErrors.length > 0) {
+      console.error('Failed to save utilisation report, validation errors: %O', validationErrors);
+      return res.status(400).send(validationErrors);
+    }
+
+    // TODO: FN-967 want to 429 here if a report details already exists
```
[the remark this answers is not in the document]
ChristianMcc: In my mind put is idempotent e.g. if you call it lots of times then we only ever get one set of data. Post we would just keep creating new entries of the data regardless of what existed. Also I meant 409 not 429 here*

portal/server/api.js
ChristianMcc: Doing separately as a separate sub task francesca is doing

dtfs-central-api/src/services/repositories/utilisation-reports-repo.test.js
```diff
@@ -0,0 +1,42 @@
+const { saveUtilisationReportDetails } = require('./utilisation-reports-repo');
+const db = require('../../drivers/db-client');
+const { DB_COLLECTIONS } = require('../../constants/db_collections');
```
toddym42, line 3: Looks like this needs to end `require('../../constants/dbCollections');` now.

dtfs-central-api/src/v1/controllers/utilisation-report-service/upload-utilisation-report.controller.js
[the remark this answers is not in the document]
toddym42: Ah, I see where your coming from. Usually though I would expect a PUT request to update the resource if it's different (which I don't think we want to do here?), and also for the resource ID to be included in the API endpoint (e.g. `/utilisation-reports/<report_id_here>` for PUT and just `/utilisation-reports`, as we have, for POST) (summary [here](https://restfulapi.net/rest-put-vs-post/)). 

I guess neither method perfectly matches our particular use case, so I'm happy to go with what you suggest.

dtfs-central-api/src/v1/routes/portal-routes.js
```diff
@@ -875,6 +876,38 @@ portalRouter.route('/cron-jobs').delete(cronJobsController.deleteAllEstoreLogs);
...
+ *                 - type: object
+ *                   properties:
+ *                     dateUploaded:
+ *                       example: 2023-10-27T08:07:40.028Z
+ *       500:
+ *         description: Internal server error
+ *       400:
+ *         description: Invalid payload
+ *       429:
+ *         description: Server conflict
```
toddym42, line 908: Reminder to update this to 409 if we update the controller (as per your comment).

toddym42 (approved):

dtfs-central-api/src/v1/validation/utilisation-report-service/utilisation-data-validator.js
francescastocco, line 1: Can we commonise some of these validations and pass in the field name for the error? Eg. currency and monetary values?

dtfs-central-api/src/v1/validation/utilisation-report-service/utilisation-report-validator.test.js
```diff
@@ -0,0 +1,99 @@
...
+      expect(validationError).toEqual(null);
+    });
+
+    it('returns an error when no month is provided', async () => {
+      const validationError = validateMonth(undefined);
+
+      expect(validationError).toEqual('Month is required');
+    });
+
+    it('returns null when an incorrect month is provided', async () => {
```
francescastocco, line 19: I think all the 3rd tests should be named `returns an error...`
Wondering if we can commonise the tests too?

dtfs-central-api/src/v1/validation/utilisation-report-service/utilisation-data-validator.test.js
```diff
@@ -0,0 +1,172 @@
...
+      const validationError = validateUkefId('28374891', 1);
+
+      expect(validationError).toEqual(null);
+    });
+
+    it('returns null when no UKEF ID is provided', async () => {
+      const validationError = validateUkefId(undefined, 1);
+
+      expect(validationError).toEqual(null);
+    });
```
francescastocco, line 24: Is this one correct? If no facility ID is provided would we not expect to throw an error? Same for the other tests when nothing is provided - I think I understood that Payment Currency could be empty and we would assume base currency.

dtfs-central-api/src/services/repositories/utilisation-reports-repo.test.js
francescastocco, line 3: Just a note to update the path

francescastocco (changes requested):
Looks good :)

dtfs-central-api/src/v1/controllers/utilisation-report-service/upload-utilisation-report.controller.js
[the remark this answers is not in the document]
francescastocco: I think I see this as a POST as it's the first time we create the records in the DB, and when we update we'd use PUT. 
If we called other POST endpoints (eg. creating a new user) multiple times we wouldn't want to keep creating new data entries either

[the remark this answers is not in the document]
ChristianMcc: I'll change to POST :D

dtfs-central-api/src/services/repositories/utilisation-reports-repo.test.js
ChristianMcc: VSCode not smart enough :(

dtfs-central-api/src/v1/validation/utilisation-report-service/utilisation-data-validator.test.js
ChristianMcc: So here I wanted to add in validation to be safe for putting information in the database but I didn't want to include business logic - I don't want the case where DTFS API has stronger validation than portal and they need to be kept in sync. 
E.g. if portal validation passes I don't want DTFS central API to fail and the user journey to fail

Can chuck a comment at the top of the validations file if helpful to communicate that!

dtfs-central-api/src/v1/validation/utilisation-report-service/utilisation-report-validator.test.js
ChristianMcc: I'll have a think about this one, might tackle it in FN-978 if that's alright!

dtfs-central-api/src/v1/validation/utilisation-report-service/utilisation-data-validator.test.js
francescastocco: Makes sense!

dtfs-central-api/src/v1/routes/portal-routes.js
```diff
@@ -875,6 +876,38 @@ portalRouter.route('/cron-jobs').delete(cronJobsController.deleteAllEstoreLogs);
...
+ *                     dateUploaded:
+ *                       example: 2023-10-27T08:07:40.028Z
+ *       500:
+ *         description: Internal server error
+ *       400:
+ *         description: Invalid payload
+ *       409:
+ *         description: Server conflict
+ */
+portalRouter.route('/utilisation-reports').put(utilisationReportUpload.postUtilisationReportData);
```
toddym42, line 910: I think this should be `.post` now, and in the doc comment above?

francescastocco (dismissed):

toddym42 (approved):

dtfs-central-api/src/v1/routes/portal-routes.js
ChristianMcc: good catch!

toddym42 (approved):

dtfs-central-api/src/v1/controllers/utilisation-report-service/upload-utilisation-report.controller.js
```diff
@@ -0,0 +1,45 @@
...
+  if (filePathValidationError) validationErrors.push(filePathValidationError);
+  return validationErrors;
+};
+
+const postUtilisationReportData = async (req, res) => {
+  try {
+    const {
+      reportData, month, year, user, filePath
+    } = req.body;
+    const { bank } = user;
```
toddym42, line 26: Do we also need to validate that the `bank.id` isn't `*` (as it would be for user's with access to all banks, like Admins)? This has come up after discussion around whether Admins should be able to see the two GEF report nav items. It isn't hard-coded that admins have access to all banks though, so I guess this could technically be an issue for every user.

Re the nav bar issue, I've suggested that this be addressed in a separate ticket. Maybe that ticket could cover handling of access to multiple banks in other places too (like this). What do you think?

ChristianMcc: Ah yes, ideally I don't want to have to deal with this as much. We should be only allowing admins to have * (which we don't atm) and I'm not sure we want to even support admins being able to 'upload a report' for a bank, aka they shouldn't really have access to the upload.

All in all I like your suggestion of yes let's tackle this separately and we'll need to catch up with DTFS team not just E2ETF team for this

toddym42 (approved):

PR #2199, opened 2023-10-30
fix(DTFS2-6759): fix token expiry

AlexBramhill:
# Introduction
Currently, token expiry is non-functional. We are setting the token expiry based on Date.now() which uses milliseconds since epoch. JWTs work with seconds since epoch ([RFC 7519: JSON Web Token (JWT) (rfc-editor.org)](https://www.rfc-editor.org/rfc/rfc7519#section-4.1.6))

# Resolution
This fix removes setting `ias` as it is unneeded. 

# Misc
I have checked to see what happens when a token is no longer valid in `portal-api`. Portal logs the user out and their session data is deleted.

### DIFF
--- a/portal-api/src/crypto/utils.js
+++ b/portal-api/src/crypto/utils.js
@@ -61,7 +61,6 @@ function issueJWTWithExpiryAndPayload({ user, sessionIdentifier = crypto.randomB
 
   const payload = {
     sub: _id,
-    iat: Date.now(),
     sessionIdentifier,
     ...additionalPayload,
   };
--- a/portal-api/src/crypto/utils.test.js
+++ b/portal-api/src/crypto/utils.test.js
@@ -26,9 +26,17 @@ describe('crypto utils', () => {
 
   const DATE_NOW_IN_UNIX_TIME = Math.floor(Date.now().valueOf() / 1000);
 
-  const SECONDS_IN_30_MINUTES = 30 * 60;
+  const SECONDS_IN_105_MINUTES = 105 * 60;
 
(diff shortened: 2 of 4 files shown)


AlexBramhill:
Note: I wont merge this until I've merged [this ticket](https://github.com/UK-Export-Finance/dtfs2/pull/2198).

portal-api/src/crypto/utils.js
```diff
@@ -61,7 +61,6 @@ function issueJWTWithExpiryAndPayload({ user, sessionIdentifier = crypto.randomB
 
   const payload = {
     sub: _id,
-    iat: Date.now(),
```
lmarrai-sw, line 64: Just to double-check I've understood:
- `jsonwebtoken` was calculating the `exp` claim from the `iat` and `expiresIn` that we specified (and the `iat` was out by a factor of 1000 so our `exp` claim was incorrect)
- by not providing the `iat` ourselves:
  - `jsonwebtoken` will calculate `exp` from `expiresIn` correctly
  - `jsonwebtoken` will still calculate the `iat` claim for the JWT itself and include it

Is that right?

lmarrai-sw (approved):

portal-api/src/crypto/utils.js
AlexBramhill: Exactly 
https://github.com/auth0/node-jsonwebtoken

Generated jwts will include an `iat` (issued at) claim by default unless `noTimestamp` is specified. If `iat` is inserted in the payload, it will be used instead of the real timestamp for calculating other things like `exp` given a timespan in `options.expiresIn`.

PR #2201, opened 2023-10-31
fix(DTFS-6752): improve css for check your email screen

oscar-richardson-softwire:
### Introduction

In [DTFS2-6733](https://ukef-dtfs.atlassian.net/browse/DTFS2-6733) we built the “Check your email screen” and used the CSS class button-as-link govuk-!-padding-0 moj-sub-navigation__item for the “Request a new sign in link” button. However moj-subnavigation__item is not appropriate for this element because it is not a sub navigation. We should improve the CSS for this element.

### Resolution

- Added the `padding: 0` CSS reset to the `button-as-link` class and checked it did not mess up the (1) other instance where the `button-as-link` class is used.
- Noticed that in the other instance where `button-as-link` was used, the `gov-uk-link` class was also used. Decided to include the additional styling from `gov-uk-link` that wasn't already in `button-as-link` in the latter class, for clarity and so that the developer doesn't have to add multiple classes to ensure that a button is styled as a GOV UK link.
- Removed the `govuk-!-margin-bottom-0` class from the other instance where `button-as-link` was used, as it wasn't doing anything (I tested this on multiple screen sizes).
- Used the `govuk-!-font-size-19` class to correctly set the font size for both instances - I decided not to include this CSS in the `button-as-link` class as we may want links of different font sizes.


portal/public/js/mojFrontend.js.LICENSE.txt
```diff
@@ -1,10 +1,10 @@
...
+ * jQuery JavaScript Library v3.7.1
  * https://jquery.com/
  *
  * Copyright OpenJS Foundation and other contributors
  * Released under the MIT license
  * https://jquery.org/license
  *
- * Date: 2023-05-11T18:29Z
+ * Date: 2023-08-28T13:37Z
  */
```
lmarrai-sw, line 10: Why did this and `portal/public/js/mojFrontend.js.map` change?

portal/styles/button.scss
```diff
@@ -4,4 +4,8 @@
   text-decoration: underline;
   cursor: pointer;
   color: $govuk-link-colour;
+  padding: 0;
+  -webkit-font-smoothing: antialiased;
+  -moz-osx-font-smoothing: grayscale;
```
lmarrai-sw, line 9: I expected we'd want `inherit` for these

portal/styles/button.scss
```diff
@@ -4,4 +4,8 @@
...
+  font-family: GDS Transport, arial, sans-serif;
```
lmarrai-sw, line 10: I expected we'd want `font: inherit` here instead

I hope doing that also means we don't need to use `govuk-!-font-size-19` anymore either

portal/styles/button.scss
```diff
@@ -4,4 +4,8 @@
   text-decoration: underline;
   cursor: pointer;
   color: $govuk-link-colour;
+  padding: 0;
```
lmarrai-sw, line 7: In the introduction you mentioned
> Noticed that in the other instance where `button-as-link` was used, the `gov-uk-link` class was also used. Decided to include the additional styling from `gov-uk-link` that wasn't already in `button-as-link` in the latter class, for clarity and so that the developer doesn't have to add multiple classes to ensure that a button is styled as a GOV UK link.

I'm a bit wary of this because what if the `gov-uk-link` class changes? Do we have any way of using Sass to make `button-as-link` extend the `gov-uk-link` class?

If we can't find a way to keep it in sync automatically then we might be better off having to add both classes every time

portal/public/js/mojFrontend.js.LICENSE.txt
lmarrai-sw: @oscar-richardson-softwire  and I looked into this: 2 months ago we updated the version of jquery in portal from 3.7.0 to 3.7.1 but we have not done a webpack build since then

portal/styles/button.scss
lmarrai-sw: This change is actually from trying to merge the `govuk-link` class into this class. @oscar-richardson-softwire's going to try to make that more maintainable with SASS features

portal/styles/button.scss
oscar-richardson-softwire: Great suggestion, I've made this change. You're absolutely right that this negates the need for `govuk-!-font-size-19` too

portal/styles/button.scss
oscar-richardson-softwire: Great shout, have used the SASS @extend syntax

lmarrai-sw (dismissed):

lmarrai-sw (approved):

AlexBramhill (approved):

PR #2203, opened 2023-10-31
feat(DTFS2-5659): added tests for sorting Facility Stage column 

vamshidharukef:
## Introduction
Some tests were missing for Facility Stage column

## Resolution
Build tests for sorting facility stage column under e2e-tests\tfm\cypress\e2e\journeys\facility\facilities.spec.js


lmarrai-sw (approved):
Thanks for this @vamshidharukef

PR #2263, opened 2023-11-02
fix(private-runners): Moved workflow jobs to public runners

abhi-markan:
## Introduction ✏️ 
GHA jobs backlog is being noticed due to rise in jobs requests.
All non-deployment and IaC jobs should be executed on public runners.

## Resolution ✔️ 
* Moved SCA and Lint jobs to public runners `ubuntu-latest`.

## Miscellaneous ➕ 
* Dependencies updates


ChristianMcc (approved):
LGTM

PR #2266, opened 2023-11-02
feat(DTFS2-6762): create environment variable for Portal website domain in Portal API

oscar-richardson-softwire:
### Introduction

In the new Portal 2FA flow, the sign in link needs to link to the Portal website to log the user in. The Portal website URL varies per-environment, so we need to introduce an environment variable that configures the Portal website URL and use this for the sign in link that we send to the user.

### Resolution

- Added a new environment variable PORTAL_UI_URL
- @abhi-markan will need to add this to GitHub and ensure it has the correct value per environment

### Misc
**NOTE**: This branch was created from Luke's branch https://github.com/UK-Export-Finance/dtfs2/pull/2200 (which has not yet been merged), so also contains those changes. The only additional changes I have made are to the files:

- .env.sample
- portal-api/src/config/sign-in-link.config.js
- portal-api/src/v1/users/sign-in-link.service.js
- portal-api/src/v1/users/sign-in-link.service.test.js


lmarrai-sw:
@oscar-richardson-softwire do we definitely want to merge _all_ of these commits into the `feat/stbr-infra` branch?

oscar-richardson-softwire:
> @oscar-richardson-softwire do we definitely want to merge _all_ of these commits into the `feat/stbr-infra` branch?

Good point, I think we only want the 3 most recent commits (i.e., the changes from my 6762 branch and your 6750 branch)?

PR #2269, opened 2023-11-02
feat(FN-980): fix utilisation reports sorting from db

francescastocco:
## Introduction

After putting test data in the DB we were seeing an error that there was no corresponding composite index to sort the data.

## Resolution

Use javascript array sorting

## To Do

- Update test data to match the schema
- Fix tests


dtfs-central-api/src/v1/controllers/utilisation-report-service/previous-reports.controller.js
```diff
@@ -1,22 +1,18 @@
...
-    ]).toArray();
+    const { bankId } = req.params;
+
+    const utilisationReportsCollection = await db.getCollection(DB_COLLECTIONS.UTILISATION_REPORTS);
+    const filteredUtilisationReports = await utilisationReportsCollection
+      .find({ 'bank.id': { $eq: bankId } })
+      .toArray();
+
+    const filteredAndSortedUtilisationReports = filteredUtilisationReports
+      .sort((a, b) => a.year - b.year || a.month - b.month);
```
toddym42, line 14: I always find the native js `.sort` functions unintuitive, and this becomes more so when there is more that one attribute taken into account.

I think this is doing "sort by year ascending, then by month ascending"? If that's the case we could make use of lodash to make this a bit easier to understand. e.g.
```
const sortBy = require('lodash/sortBy');
...
const filteredAndSortedUtilisationReports = sortBy(filteredUtilisationReports, ['year', 'month']);
```
`sortBy` orders ascending by default. if we wanted to order anything descending we could instead use `orderBy` like this
```
I think this makes things much easier to understand, which hopefully would lead to less errors in future.
const orderBy = require('lodash/orderBy');
...
const filteredAndSortedUtilisationReports = orderBy(filteredUtilisationReports, ['year', 'month'], ['asc', 'desc']);
```

portal-api/api-tests/v1/utilisation-report-service/previous-reports.api-test.js
```diff
@@ -7,7 +7,7 @@ const { withClientAuthenticationTests } = require('../../common-tests/client-aut
 const { withRoleAuthorisationTests } = require('../../common-tests/role-authorisation-tests');
 const { PAYMENT_REPORT_OFFICER } = require('../../../src/v1/roles/roles');
 
-const collectionName = 'utilisation-reports';
+const collectionName = 'utilisationReports';
```
toddym42, line 10: Could we add a `DB_COLLECTIONS` const to the `portal-api` project to use here instead (and the next file too)?

portal-api/src/v1/controllers/utilisation-report-service/previous-reports.controller.js
```diff
@@ -39,15 +39,21 @@ const getPreviousReportsByBankId = async (req, res) => {
...
+    }
+
     const years = getYears(data);
     const groupedReports = getReportsGroupedByYear(years, data);
     const reportsGroupedByYear = populateOmittedYears(groupedReports, years);
 
-    res.status(200).send(reportsGroupedByYear.sort((a, b) => b.year - a.year));
+    return res.status(200).send(reportsGroupedByYear.sort((a, b) => b.year - a.year));
   } catch (error) {
     console.error('Unable to get previous reports %s', error);
```
toddym42, line 55: I think we could remove `%s` to potentially get more information about the error here (the entire object instead of just the `error.toString()`. We could change this to `%O` to get the full object, but I would just remove the placeholder completely, i.e.
```
console.error('Unable to get previous reports:', error);
```
which will do pretty much the same thing.

toddym42 (approved):
A couple of suggestions, but nothing blocking : )

dtfs-central-api/src/v1/controllers/utilisation-report-service/previous-reports.controller.js
francescastocco: Yep agreed that this is hard to read & understand :) here I am sorting year and month ascending but at some point I do sort them by descending year so will add orderBy in there

dtfs-central-api/package.json
```diff
@@ -37,6 +37,7 @@
     "express-mongo-sanitize": "^2.2.0",
     "express-rate-limit": "^6.11.2",
     "express-validator": "^7.0.1",
+    "lodash.sortby": "^4.7.0",
```
toddym42, line 40: Lodash actually discourage per method packages like this as they will be removed in future versions (full reasoning [here](https://lodash.com/per-method-packages)), so I think we we should just add the whole package here.

This won't necessarily increase the build size as we can still use individual functions where required. e.g.
```
const sortBy = require('lodash/sortBy');
```
which means only the `sortBy` function will appear in the build output.

This also means we can more easily use further lodash functions in future without the need to keep adding individual packages.

dtfs-central-api/src/services/repositories/utilisation-reports-repo.test.js
```diff
@@ -37,6 +96,34 @@ describe('utilisation-reports-repo', () => {
...
+      year: 2021,
+      dateUploaded: expect.any(Date),
+      path: 'test path',
+      uploadedBy: {
+        id: '123',
+        name: 'test user',
+      },
+    }];
+
+    it('filters and sorts the data correctly', async () => {
```
toddym42, line 114: I don't think we should mention `filters` here as we are mocking the db collection response so the data is not filtered. I guess we can only do this in tests where an actual DB is spun up?

dtfs-central-api/src/services/repositories/utilisation-reports-repo.test.js
```diff
@@ -37,6 +96,34 @@ describe('utilisation-reports-repo', () => {
...
+        toArray: toArraySpy,
+      }));
+      const getCollectionMock = jest.fn(() => ({
+        find: findSpy,
+      }));
+      jest.spyOn(db, 'getCollection').mockImplementation(getCollectionMock);
+
+      const utilisationReports = await getUtilisationReportDetails(bankId);
+      expect(utilisationReports[0]).toEqual(mockUtilisationReports[3]);
+      expect(utilisationReports[utilisationReports.length - 1]).toEqual(mockUtilisationReports[0]);
```
toddym42, line 126: It's not immediately clear what we're asserting on here, and I think we could go further to assert that the data is correctly ordered by year then month.

We could instead do something like this:
```
const testUtilisationReport = {
  bank: {
    id: '124',
    name: 'test bank',
  },
  month: 4,
  year: 2021,
  dateUploaded: expect.any(Date),
  path: 'test path',
  uploadedBy: {
    id: '123',
    name: 'test user',
  },
};

it('sorts the data by year then month', async () => {
  // Arrange
  const bankId = testUtilisationReport.bank.id;
  const report1 = { ...testUtilisationReport, month: 2, year: 2022 };
  const report2 = { ...testUtilisationReport, month: 3, year: 2021 };
  const report3 = { ...testUtilisationReport, month: 1, year: 2022 };
  const report4 = { ...testUtilisationReport, month: 2, year: 2021 };

  const mockUtilisationReports = [report1, report2, report3, report4];

  spyOn(db, 'getCollection').mockImplementation(() => ({
    find: jest.fn(() => ({
      toArray: jest.fn(() => mockUtilisationReports),
    })),
  }));

  // Act
  const response = await getUtilisationReportDetails(bankId);

  // Assert
  const expectedResponse = [report4, report2, report3, report1];
  expect(response).toEqual(expectedResponse);
});
```
The difference here is that when I create the mock reports in the test I'm only setting the values that I care about for the test (`month` and `year` in this case) so it's clear what we're testing. I can then use these mock reports to more easily check the entire response.

portal-api/package-lock.json
```diff
@@ -24,6 +24,7 @@
         "filesize": "^8.0.7",
         "joi": "^17.10.2",
         "jsonwebtoken": "^9.0.2",
+        "lodash.orderby": "^4.6.0",
```
toddym42, line 27: Same comment here about adding the full lodash package here.

portal-api/src/v1/controllers/utilisation-report-service/previous-reports.controller.js
```diff
@@ -39,15 +40,22 @@ const getPreviousReportsByBankId = async (req, res) => {
...
+
+    if (status !== 200) {
+      console.error('Unable to get previous reports');
+      return res.status(500).send(data);
+    }
+
     const years = getYears(data);
     const groupedReports = getReportsGroupedByYear(years, data);
     const reportsGroupedByYear = populateOmittedYears(groupedReports, years);
+    const reportsOrderedByDescendingYear = orderBy(reportsGroupedByYear, ['year'], ['desc']);
```
toddym42, line 53: Would be good to add a test to check this new logic.

dtfs-central-api/package.json
francescastocco: Ah, good to know, thanks!

dtfs-central-api/src/services/repositories/utilisation-reports-repo.test.js
francescastocco: Ah yes I like this much more, I wanted something which made it more obvious but didn't know how to do it, thanks 🙏

dtfs-central-api/src/services/repositories/utilisation-reports-repo.test.js
francescastocco: Yep, very true

portal-api/src/v1/controllers/utilisation-report-service/previous-reports.controller.js
```diff
@@ -1,11 +1,23 @@
+const orderBy = require('lodash/orderBy');
 const api = require('../../api');
 const { getMonthName } = require('../../../utils/getMonthName');
 
+/**
+ * Returns a set of unique years based on reports from the database.
+ * @param {Object[]} reports - reports from the database
+ * @returns {Number[]} - unique set of years
```
toddym42, line 8: I think the type here would just be the primitive `number` instead of the `Number` interface (and this would probably be the case for all the types we use unless we've specifically used a Class or Interface).

toddym42 (approved):

PR #2270, opened 2023-11-03
feat(FN-1432): handle formulas and new lines in excel cells

ChristianMcc:
## Introduction
The current parsing could not handle formulas in excel files and would return [Object object]. It also couldn't handle new lines in cells

## Resolution
Replace new lines with spaces so that we can correctly convert to a csv format for parsing without worrying about extra new line characters in the middle of a row.
Check if the cell value comes from a formula looking for cell.value?.result before defaulting to cell.value if that is not the case to handle excel formulas. This also works for lookups to other sheets as they count as formulas in the same way by the excel parsing library

### DIFF
--- a/portal/server/utils/csv-utils.js
+++ b/portal/server/utils/csv-utils.js
@@ -19,7 +19,7 @@ const columnIndexToExcelColumn = (index) => {
  * @returns {string | number} - cell value.
  */
 const extractCellValue = (cell) => {
-  const cellValue = cell.value?.result ? cell.value.result : cell.value;
+  const cellValue = cell.value?.result ?? cell.value;
   const cellValueWithoutNewLines = typeof cellValue === 'string' ? cellValue?.replace(/\r\n|\r|\n/g, ' ').trim() : cellValue;
   return cellValueWithoutNewLines;
 };
--- a/portal/server/utils/csv-utils.js
+++ b/portal/server/utils/csv-utils.js
@@ -20,7 +20,7 @@ const columnIndexToExcelColumn = (index) => {
  */
 const extractCellValue = (cell) => {
   const cellValue = cell.value?.result ? cell.value.result : cell.value;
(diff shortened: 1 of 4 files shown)


portal/server/utils/csv-utils.js
```diff
@@ -13,6 +13,17 @@ const columnIndexToExcelColumn = (index) => {
...
 };
 
+/**
+ * Extracts the value in the cell of an excel cell and removes any new lines so that it doesn't affect parsing as a csv.
+ * @param {Object} cell - excel cell.
+ * @returns {string | number} - cell value.
+ */
+const extractCellValue = (cell) => {
+  const cellValue = cell.value?.result ? cell.value.result : cell.value;
+  const cellValueWithoutNewLines = typeof cellValue === 'string' ? cellValue?.replace(/\n/g, ' ') : cellValue;
```
toddym42, line 23: Could use `/(\r\n|\r|\n)+/g` for the regex here to account for other possible new line characters

portal/server/utils/csv-utils.test.js
```diff
@@ -103,4 +109,38 @@ describe('csv-utils', () => {
...
+    it('returns the cell value', async () => {
+      const cellValue = { value: 'GBP' };
+
+      const extractedValue = extractCellValue(cellValue);
+
+      expect(extractedValue).toEqual('GBP');
+    });
+
+    it('returns the cell value without new lines', async () => {
+      const cellValue = { value: 'GBP\na' };
```
toddym42, line 123: Would be nice to also test the case of multiple new line chars in different places.

toddym42 (approved):

portal/server/utils/csv-utils.js
ChristianMcc: Yeah I thought about this although the library only cares about `\n` as the condition for a new line in csv. Maybe it's helpful for me to replace \r\n with a space instead thought as otherwise it'll be '\r ' which is probably fine but weird!

toddym42 (approved):

francescastocco (approved):
Looks good to me :)

PR #2271, opened 2023-11-03
feat(dtfs2-6757): add sign in link validation

AlexBramhill:


portal-api/api-tests/v1/users/post-sign-in-link.api-test.js
```diff
@@ -31,7 +31,7 @@ jest.mock('node:crypto', () => ({
   const hash = '0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef';
   const salt = 'abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789';
   const saltBytes = Buffer.from(salt, 'hex');
-  const signInCode = '0a1b2c3d4e5f67890a1b2c3d4e5f6789';
+  const signInToken = '0a1b2c3d4e5f67890a1b2c3d4e5f6789';
```
AlexBramhill, line 34: @lmarrai-sw I've renamed this to `signInToken` to match `resetPasswordToken`

portal-api/src/v1/users/sign-in-link.controller.js
```diff
@@ -1,10 +1,45 @@
...
+      const { user } = req;
+
+      const isValidSignInToken = await this.#signInLinkService.isValidSignInToken({ username: user.username, signInToken });
+
+      if (!isValidSignInToken) {
+        throw new Error(`Invalid sign in token for user: ${user.username}`);
+      }
+
+      const { sessionIdentifier, ...tokenObject } = utils.issueValid2faJWT(user);
+      await updateLastLogin(user, sessionIdentifier, () => {});
```
AlexBramhill, line 25: I really dislike this empty callback

portal-api/src/v1/users/sign-in-link.service.js
```diff
@@ -17,37 +18,51 @@ class SignInLinkService {
...
       userEmail,
       userFirstName,
       userLastName,
     });
   }
 
-  #createSignInCode() {
+  async isValidSignInToken({ username, signInToken }) {
+    return new Promise((resolve, reject) => {
+      findByUsername(username, (error, user) => {
```
AlexBramhill, line 34: Again, I dislike the callback structure we are using here. I'm tempted to depreciate `findByUsername` and make its implementation call our new `user repo` (pending how many extra failing tests this is and others thoughts on the matter)

portal-api/src/v1/users/sign-in-link.service.js
```diff
@@ -17,37 +18,51 @@ class SignInLinkService {
...
-  #createSignInCode() {
+  async isValidSignInToken({ username, signInToken }) {
+    return new Promise((resolve, reject) => {
+      findByUsername(username, (e, user) => {
+        if (e || !user) {
+          const error = new Error(`Failed to find user: ${username}`);
+          error.cause = e || 'User object not defined';
+          reject(error);
+        }
+        const { hash, salt } = user.signInToken;
```
lmarrai-sw, line 40: We'll reject with an unexpected error here if `user.signInToken` is `undefined` - is that what we want?

portal-api/src/v1/users/sign-in-link.controller.js
```diff
@@ -1,10 +1,45 @@
...
+  async validateSignInLink(req, res) {
+    try {
+      const { signInToken } = req.params;
+      const { user } = req;
+
+      const isValidSignInToken = await this.#signInLinkService.isValidSignInToken({ username: user.username, signInToken });
+
+      if (!isValidSignInToken) {
+        throw new Error(`Invalid sign in token for user: ${user.username}`);
+      }
```
lmarrai-sw, line 22: I think we want this case to cause a 403 error instead of a 500, what do you think?

portal-api/src/v1/users/sign-in-link.controller.js
```diff
@@ -1,10 +1,45 @@
...
+const { sanitizeUser } = require('./sanitizeUserData');
+
 class SignInLinkController {
   #signInLinkService;
 
   constructor(signInLinkService) {
     this.#signInLinkService = signInLinkService;
   }
 
+  async validateSignInLink(req, res) {
```
lmarrai-sw, line 13: I'd argue we do more than validate a sign in link, we complete the login and issue a full JWT. I see why you've gone for this name because validating the sign in link is the main step but if this was a 1FA login endpoint we wouldn't call it `validateUsernameAndPassword` which makes me think we might be slightly better off with a different name

portal-api/src/v1/users/sign-in-link.controller.js
```diff
@@ -1,10 +1,45 @@
...
+    try {
+      const { signInToken } = req.params;
+      const { user } = req;
```
lmarrai-sw, line 16:
```suggestion
      const { params: { signInToken }, user } = req;
```
^ A minor suggestion, I don't know if you have a preference

portal-api/src/v1/users/sign-in-link.controller.js
lmarrai-sw: I know you've not changed the URL in this PR but maybe the same idea applies to the URL, which is currently
```
POST /users/me/validate-sign-in-link/:signInToken
```

Given the endpoint that creates the tokens is `POST /users/me/sign-in-link`, maybe we want this endpoint to be something like

```
POST /users/me/sign-in-link/:signInToken/login
```

What do you think?

portal-api/src/v1/users/sign-in-link.controller.js
AlexBramhill: I have no strong preference either way, updated

portal-api/src/v1/users/sign-in-link.controller.js
AlexBramhill: let me know what you think about the new name

portal-api/src/crypto/utils.js
```diff
@@ -19,7 +19,7 @@ const PRIV_KEY = Buffer.from(process.env.JWT_SIGNING_KEY, 'base64').toString('as
 function validPassword(password, hash, salt) {
   const hashVerify = crypto.pbkdf2Sync(password, salt, 10000, 64, 'sha512').toString('hex');
 
-  return hash === hashVerify;
+  return crypto.timingSafeEqual(hash, hashVerify);
 }
```
lmarrai-sw, line 23: I like that we're thinking to change this but it also makes me a bit nervous because `timingSafeEqual` throws if `hash` and `hashVerify` are not the same length, and user passwords have a weird edge case where new users don't have a password set to start with - is that going to cause us to throw here, or does the code not get this far if the user doesn't have a hash in the db yet?

lmarrai-sw: Also, shall we change it in tfm-api too?

portal-api/src/v1/users/sign-in-link.service.js
```diff
@@ -17,37 +17,47 @@ class SignInLinkService {
...
       userLastName,
     });
   }
 
-  #createSignInCode() {
+  async isValidSignInToken({ userId, signInToken }) {
+    const user = await this.#userRepository.findById(userId);
+    if (!user.hash || !user.salt) {
+      throw new Error('User does not have a valid sign in token.');
+    }
```
lmarrai-sw, line 35: Why do we have this check?

portal-api/src/v1/users/sign-in-link.service.js
```diff
@@ -17,37 +17,47 @@ class SignInLinkService {
...
+
+    const { hash, salt } = user.signInToken;
+    return this.#hasher.verifyHash({ target: signInToken, hash: Buffer.from(hash, 'hex'), salt: Buffer.from(salt, 'hex') });
```
lmarrai-sw: Maybe something you want to worry about in the next PR, or later, but there's a bit of a mismatch between saving and reading the hash and salt

When we save the hash and salt, we give the repository the bytes and it decides to convert them to hex strings to be stored in the db.
When we retrieve the hash and salt, the repository gives us the hex strings and we need to know that they need to be converted back to bytes

Ideally it would be consistent, i.e.
1. we give the repository bytes to save and it gives us back bytes when we read, or
2. we give the repository hex strings to save and it gives us back hex strings when we read

Personally I prefer (1) because the bytes are the actual hash and salt and the hex strings are just the format we choose to persist that info in the db

portal-api/src/v1/users/sign-in-link.controller.js
```diff
@@ -1,10 +1,52 @@
...
+    try {
+      const {
+        params: { signInToken },
+        user,
+      } = req;
+
+      const isValidSignInToken = await this.#signInLinkService.isValidSignInToken({ userId: user._id, signInToken });
+
+      if (!isValidSignInToken) {
+        const invalidSignInTokenMessage = `Invalid sign in token for user: ${user.username}`;
```
lmarrai-sw, line 23: Minor question, but is there a reason we choose to log and respond with the username instead of the id?

portal-api/src/v1/users/controller.js
```diff
@@ -93,23 +94,30 @@ exports.list = async (callback) => {
...
-
-  const collection = await db.getCollection('users');
-
-  collection.findOne({ _id: { $eq: ObjectId(_id) } }, callback);
 };
 
+/**
+ * @deprecated Use findByUsername inside user repository instead
+ */
 exports.findByUsername = async (username, callback) => {
```
lmarrai-sw, line 113: The error-case behaviours for `findOne` and `findByUsername` have changed from this - have we been able to check that the differences are safe? :)

portal-api/src/v1/users/sign-in-link.controller.js
lmarrai-sw: I like it, thanks @AlexBramhill

portal-api/src/v1/users/sign-in-link.service.js
AlexBramhill: We actually never have checked that this exists in the database prior to this step. On the off chance that we've somehow got here and failed to save this to the database (currently error handled), we should be prepared to handle it here too do you think? I didn't want to base the implimentation around other functions error handling

portal-api/src/crypto/utils.js
AlexBramhill: It is indeed going to throw here -- I think a solution is to try catch here, and return false in the catch. This would ignore the time safe equals (https://github.com/nodejs/node/issues/17178) but I feel this is not an issue

AlexBramhill: (this implimentation is not worse than the previous implimentation, dispite the try/catch block here)

portal-api/src/v1/users/sign-in-link.controller.js
AlexBramhill: Minor oversight

portal-api/src/v1/users/controller.js
AlexBramhill: reverted this back

portal-api/src/v1/users/sign-in-link.service.js
AlexBramhill: Updated to match first option

portal-api/src/crypto/utils.test.js
```diff
@@ -1,101 +1,165 @@
...
+        const result = validPassword(inputPassword, inputHash, inputSalt);
+        expect(result).toEqual(false);
+      });
     });
   });
 
-  describe('issueValid2faJWT', () => {
-    it('should throw an error if no existing session identifier is present', () => {
-      expect(() => issueValid2faJWT(USER)).toThrow('User does not have a session identifier');
+  describe('issue JTW tokens', () => {
```
lmarrai-sw, line 70:
```suggestion
  describe('issue JWTs', () => {
```

portal-api/src/crypto/utils.test.js
```diff
@@ -1,101 +1,165 @@
...
+        name: 'HSBC',
+        emails: ['maker1@ukexportfinance.gov.uk', 'maker2@ukexportfinance.gov.uk'],
+      },
+    };
+
+    const EXISTING_SESSION_IDENTIFIER = crypto.randomBytes(32).toString('hex');
+
+    const USER_WITH_EXISTING_SESSION_IDENTIFIER = { ...USER, sessionIdentifier: EXISTING_SESSION_IDENTIFIER };
+
+    const DATE_NOW_IN_UNIX_TIME = Math.floor(Date.now().valueOf() / 1000);
```
lmarrai-sw, line 98: I think this will execute before `jest.useFakeTimers`

lmarrai-sw (approved):
Thanks @AlexBramhill - I'm happy if you want to merge first and raise a separate PR for extra tests and resolving my comments

PR #2272, opened 2023-11-03
feat(DTFS2-6762): add environment variable portal-ui-url

MarRobSoftwire:
## Introduction
The new environment variable `PORTAL_UI_URL` is added by #2265 & needs to be added to the bicep files

## Resolution

* Added `PORTAL_UI_URL` to `.github/workflows/bicep_az_deployment_group.yml`, `.github/workflows/bicep_workflow_feature.yml` and `infrastructure/resource_group_level/main.bicep`

### DIFF
--- a/.github/workflows/bicep_az_deployment_group.yml
+++ b/.github/workflows/bicep_az_deployment_group.yml
@@ -122,9 +122,6 @@ on:
       PDC_INPUTTERS_EMAIL_RECIPIENT:
         description: webapp environment variable
         required: true
-      PORTAL_UI_URL:
-        description: webapp environment variable
-        required: true
 
 jobs:
   configure-command:
--- a/.github/workflows/bicep_az_deployment_group.yml
+++ b/.github/workflows/bicep_az_deployment_group.yml
@@ -122,6 +122,9 @@ on:
       PDC_INPUTTERS_EMAIL_RECIPIENT:
         description: webapp environment variable
         required: true
+      PORTAL_UI_URL:
+        description: webapp environment variable
+        required: true
 
 jobs:
   configure-command:
@@ -233,4 +236,5 @@ jobs:
               ESTORE_URL=${{ secrets.ESTORE_URL }} \
               PDC_INPUTTERS_EMAIL_RECIPIENT=${{ secrets.PDC_INPUTTERS_EMAIL_RECIPIENT }} \
               RATE_LIMIT_THRESHOLD=${{ vars.RATE_LIMIT_THRESHOLD}} \
-              MAX_UTILISATION_REPORT_FILE_SIZE=${{ vars.MAX_UTILISATION_REPORT_FILE_SIZE }}
+              MAX_UTILISATION_REPORT_FILE_SIZE=${{ vars.MAX_UTILISATION_REPORT_FILE_SIZE }} \
+              PORTAL_UI_URL=${{ vars.PORTAL_UI_URL }}
--- a/.github/workflows/bicep_workflow_feature.yml
+++ b/.github/workflows/bicep_workflow_feature.yml
@@ -75,4 +75,3 @@ jobs:
       SESSION_SECRET: ${{ secrets.SESSION_SECRET }}
       ESTORE_URL: ${{ secrets.ESTORE_URL }}
       PDC_INPUTTERS_EMAIL_RECIPIENT: ${{ secrets.PDC_INPUTTERS_EMAIL_RECIPIENT }}
-      PORTAL_UI_URL: ${{ vars.PORTAL_UI_URL }}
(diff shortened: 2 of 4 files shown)


infrastructure/resource_group_level/main.bicep
```diff
@@ -189,9 +191,10 @@ var tfmApiAdditionalSecureConnectionStrings = {
 }
 
 var portalUiSettings = {
-    RATE_LIMIT_THRESHOLD: RATE_LIMIT_THRESHOLD // TODO:FN-1086 30 on dev, 10000 on feature
-    COMPANIES_HOUSE_API_URL: COMPANIES_HOUSE_API_URL
-    MAX_UTILISATION_REPORT_FILE_SIZE: MAX_UTILISATION_REPORT_FILE_SIZE
+  RATE_LIMIT_THRESHOLD: RATE_LIMIT_THRESHOLD // TODO:FN-1086 30 on dev, 10000 on feature
+  COMPANIES_HOUSE_API_URL: COMPANIES_HOUSE_API_URL
+  MAX_UTILISATION_REPORT_FILE_SIZE: MAX_UTILISATION_REPORT_FILE_SIZE
+  PORTAL_UI_URL: PORTAL_UI_URL
```
lmarrai-sw, line 197: `portal-api` is the service that needs to know this info, not `portal-ui`, so I'm guessing it should be in `portalApiSecureSettings` instead

francescastocco (changes requested):
Nice, there are just a few changes to make as we handle secrets and variables slightly differently

.github/workflows/bicep_workflow_feature.yml
```diff
@@ -46,7 +46,6 @@ jobs:
       REMOTE_VNET_NAME_VPN: ${{ secrets.REMOTE_VNET_NAME_VPN }}
       VNET_ADDRESS_PREFIX: ${{secrets.VNET_ADDRESS_PREFIX}}
       UKEF_VPN_IPS: ${{ secrets.UKEF_VPN_IPS }}
-
```
OxonianCambion, line 49: That was there to give a delineation between "infrastructure" values and application environment variables. I don't feel strongly about it. Either leave it as is or put a comment saying as such.

.github/workflows/bicep_az_deployment_group.yml
```diff
@@ -233,4 +233,5 @@ jobs:
               ESTORE_URL=${{ secrets.ESTORE_URL }} \
               PDC_INPUTTERS_EMAIL_RECIPIENT=${{ secrets.PDC_INPUTTERS_EMAIL_RECIPIENT }} \
               RATE_LIMIT_THRESHOLD=${{ vars.RATE_LIMIT_THRESHOLD}} \
-              MAX_UTILISATION_REPORT_FILE_SIZE=${{ vars.MAX_UTILISATION_REPORT_FILE_SIZE }}
+              MAX_UTILISATION_REPORT_FILE_SIZE=${{ vars.MAX_UTILISATION_REPORT_FILE_SIZE }} \
+              PORTAL_UI_URL=${{ vars.PORTAL_UI_URL }}
```
OxonianCambion, line 237: I think using `vars` for this is a good idea, as the URL isn't secret. Even though vars are a beta feature, I think it is sound.

Note that you'll need to get @abhi-markan  to add the correct values to the different environments.

Ideally, it should be:
```
dev: https://tfs-dev-fd.azurefd.net/
feature: https://tfs-feature-fd.azurefd.net/
staging: https://tfs-staging-fd.azurefd.net/
prod: https://www.get-a-guarantee-for-export-finance.service.gov.uk/
```
(Note that I you should check the prod value.)

HOWEVER, because there isn't a `feature` environment on GitHub yet, you have a problem.
Because it isn't used on `dev` yet, you could do the following:
```
dev: https://tfs-feature-fd.azurefd.net/
staging: https://tfs-staging-fd.azurefd.net/
prod: https://www.get-a-guarantee-for-export-finance.service.gov.uk/
```

And keep track that the `dev` will need updating before the `main-stbr` branch is merged to `main`.

infrastructure/resource_group_level/main.bicep
```diff
@@ -145,6 +146,7 @@ var dtfsCentralApiAdditionalSecureSetting = {
 
 var portalApiSettings = {
   RATE_LIMIT_THRESHOLD: RATE_LIMIT_THRESHOLD
+  PORTAL_UI_URL: PORTAL_UI_URL
```
OxonianCambion, line 149: Note that because we are in control of the front-door deployment, we can actually get the value ourselves and avoid the GitHub variable. If there wasn't the issue of selecting the right url for the `prod` deployment I would insist on following the same non-trivial mechanism used by the `tfmUiUrl` setting. This is a little fiddly though as you need to avoid a circular dependency and set the value _after_ the front-door deployment. (Weirdly, the tfmUiUrl uses a connection string too.)

Don't do that now, but raise a card under FN-692 to at least consider it. Use FN-927 (The TFM_UI_URL one) as a template for it. Grab me if that's not clear!

OxonianCambion (approved):
Don't forget to get @abhi-markan to add the value to `dev` before you merge or it will break!

infrastructure/resource_group_level/main.bicep
MarRobSoftwire: See FN-1479

francescastocco (approved):

PR #2277, opened 2023-11-06
feat(FN-1472): remove commas from cell values

ChristianMcc:
## Introduction
We had issues parsing cell values with commas due to us going to csv first and then parsing the csv to JSON

## Resolution
No commas are needed in the values so strip the values of any commas. Added benefit that if someone has a string representing a number with commas they are removed and it is easier to parse


toddym42 (approved):
Looks good to me

PR #2282, opened 2023-11-07
fix: lock down report upload journey

ChristianMcc:
## Introduction
Trying to access one of the pages in the journey of the utilisation report upload (e.g. if you had it bookmarked) would crash the page as the utilisation report would not be in the session.

## Resolution
Route the user back to the start of the journey if the user does not have a utilisationReport in their session.

### DIFF
--- a/e2e-tests/portal/cypress/e2e/journeys/payment-report-officer/utilisation-report-upload/confirm-and-send.spec.js
+++ b/e2e-tests/portal/cypress/e2e/journeys/payment-report-officer/utilisation-report-upload/confirm-and-send.spec.js
@@ -5,32 +5,43 @@ const relativeURL = require('../../../relativeURL');
 const { BANK1_PAYMENT_REPORT_OFFICER1 } = MOCK_USERS;
 
 context('Confirm and send', () => {
-  beforeEach(() => {
-    cy.login(BANK1_PAYMENT_REPORT_OFFICER1);
-    cy.visit(relativeURL('/utilisation-report-upload'));
+  describe('After logging in and submitting a valid file', () => {
+    beforeEach(() => {
+      cy.login(BANK1_PAYMENT_REPORT_OFFICER1);
+      cy.visit(relativeURL('/utilisation-report-upload'));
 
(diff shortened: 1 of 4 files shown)


e2e-tests/portal/cypress/e2e/journeys/payment-report-officer/utilisation-report-upload/confirm-and-send.spec.js
```diff
@@ -5,32 +5,43 @@ const relativeURL = require('../../../relativeURL');
...
-    confirmAndSend.currentUrl().should('contain', '/utilisation-report-upload');
+      confirmAndSend.mainHeading().should('not.exist');
+      confirmAndSend.currentUrl().should('contain', '/confirmation');
+    });
   });
 
-  it('Should route to the Confirmation page when the Confirm and Send button is selected', () => {
-    confirmAndSend.confirmAndSendButton().click();
+  describe('After logging in but not submitting a file', () => {
+    it('Should route to the Upload Report page when you try and access the confirm and send page directly', () => {
```
ChristianMcc, line 40: Need to check if the express session will still exist at this point with a fresh login or if I'll need to force a logout first to destroy the session

portal/server/utils/csv-utils.js
```diff
@@ -20,7 +20,12 @@ const columnIndexToExcelColumn = (index) => {
  */
 const extractCellValue = (cell) => {
   const cellValue = cell.value?.result ?? cell.value;
-  const cellValueWithoutNewLines = typeof cellValue === 'string' ? cellValue?.replace(/\r\n|\r|\n/g, ' ').replace(/,/g, '').trim() : cellValue;
+  const cellValueWithoutNewLines = typeof cellValue === 'string'
+    ? cellValue
+      ?.replace(/\r\n|\r|\n/g, ' ')
```
toddym42, line 25: I know this was just a prettier change, but I think we can remove the `?` here as the `cellValue` will definitely be defined when the type is `'string'`.

toddym42 (approved):

JoshBinns2000:
lgtm

toddym42 (approved):

PR #2283, opened 2023-11-07
fix(DTFS2-6520): made aria-labels in Deal page tabs headings more consistent

avaitonis:
### Introduction
Visible H2 text didn't match aria-label in Deal underwriting tab.

### Resolution
 - Removed `aria-label` for all Deal page tabs. 
 - Changed Underwriting tab header text from `Deal` to `Underwriting`. Agreed with Tim.
 - Markup cleanup


trade-finance-manager-ui/templates/case/activity/activity.njk
```diff
@@ -17,7 +17,7 @@
 
     <div class="govuk-grid-column-full">
 
-      <h2 class="govuk-heading-xl govuk-!-margin-top-6 govuk-!-margin-bottom-5" data-cy="activity-heading" aria-label="Activity and comments">Activity and comments</h2>
+      <h2 class="govuk-heading-xl govuk-!-margin-top-6 govuk-!-margin-bottom-5" data-cy="activity-heading" aria-label="Deal activity and comments">Activity and comments</h2>
```
lmarrai-sw, line 20: For all of these changes, @avaitonis we could just remove the `aria-label` and the actual text will be read out instead.

Do you think the actual text isn't clear enough and the text in the `aria-label` is better? If so then we probably want to change the actual text too (and can talk to the DTFS service designer about that)

lmarrai-sw (changes requested):

trade-finance-manager-ui/templates/case/activity/activity.njk
avaitonis: Hi @lmarrai-sw ,

aria-labels are removed, text copy change agreed with Tim.

lmarrai-sw (approved):
Thanks for this @avaitonis

lmarrai-sw (approved):
Thanks @avaitonis

PR #2284, opened 2023-11-07
feat(FN-1163): add report due email job

toddym42:
## Ticket

[FN-1163](https://ukef-dtfs.atlassian.net/browse/FN-1163)

## Introduction

Banks report facility utilisation for the previous month from the 1st of each month until the 10th business day.

If a bank has not already submitted their report by the 10th business day they should be sent a follow-up email to inform them that the report is due today.

## Resolution

* Add new job (`send-report-due-emails-job`) to the `portal-api` scheduler to check each day if the chaser emails should be sent (we can't have a cron schedule based on business days, so we instead schedule the job to run each day and check if today is the 10th business day)
  * On the 10th business day, send email to all banks that have not already submitted a report for the current period and have a valid email set to the `paymentOfficerTeam.email` property

## To Do

* Currently makes use of the central api `/portal/previous-reports/:bankId` route, there this is currently being changed in a separate ticket. Will need updating when changes are complete.


.env.sample
```diff
@@ -39,6 +39,7 @@ SESSION_SECRET=
 # UTILISATION REPORT
 PORTAL_REPORTING_DUE_DATE_BUSINESS_DAYS_FROM_START_OF_MONTH=10
 PORTAL_REPORTING_PERIOD_START_EMAIL_SCHEDULE='0 7 1 * *'
+PORTAL_REPORT_DUE_EMAIL_SCHEDULE='0 7 * * *'
```
toddym42, line 42: Will add to the `feat/stbr-infra` branch separately.

dtfs-central-api/api-tests/v1/bank/banks-get.api-test.js
```diff
@@ -21,14 +21,8 @@ describe('/v1/bank', () => {
...
-        {
-          _id: createdBarclaysBank._id,
-          ...BANKS.BARCLAYS,
-        },
-        {
-          _id: createdHsbcBank._id,
-          ...BANKS.HSBC,
-        },
+        { ...BANKS.BARCLAYS, _id: createdBarclaysBank._id },
+        { ...BANKS.HSBC, _id: createdHsbcBank._id },
```
JoshBinns2000, line 25: I'm not sure how I feel about this change. Whenever you're in MongoDB the `_id` field comes first, so I do think the previous way this was written is easier to follow. My main concern is if someone copies this layout in the future they might add more than two fields in the single line which will make it very hard to read (I only say this because I could maybe see myself doing that if I were tired xD)

portal-api/src/scheduler/jobs/helpers/utilisation-report-helpers.js
```diff
@@ -0,0 +1,47 @@
...
+  return addBusinessDaysWithHolidays(startOfMonth(new Date()), process.env.PORTAL_REPORTING_DUE_DATE_BUSINESS_DAYS_FROM_START_OF_MONTH, bankHolidays);
+};
+
+/**
+ * Returns the utilisation report due date for the current month in 'do MMMM yyyy' format
+ * @returns {Promise<string>}
+ */
+const getFormattedReportDueDate = async () => {
+  const reportDueDate = await getReportDueDate();
+  return format(reportDueDate, 'do MMMM yyyy');
```
JoshBinns2000, line 20: I don't know what the correct approach should be here, but the designs I've interacted with use `'d MMMM yyyy'` - maybe worth asking?

portal-api/src/scheduler/jobs/helpers/utilisation-report-helpers.js
```diff
@@ -0,0 +1,47 @@
...
+};
+
+/**
+ * Returns the month (1-indexed) and year of the current report period (i.e. the previous month)
+ * @returns {{ month: number, year: number }}
+ */
+const getReportPeriodMonthAndYear = () => {
+  const lastMonthDate = subMonths(startOfMonth(new Date()), 1);
```
JoshBinns2000, line 28: I don't dislike the use of `date-fns`, but I'm curious as to why this is better than something like:
```
const lastMonth = new Date().getMonth() - 1;
const oneIndexedMonth = lastMonth + 1;
```

portal-api/src/scheduler/jobs/helpers/utilisation-report-helpers.test.js
```diff
@@ -0,0 +1,116 @@
...
+
+jest.mock('../../../external-api/bank-holidays');
+
+const originalProcessEnv = process.env;
+
+describe('utilisation-report-helpers', () => {
+  beforeEach(() => {
+    process.env = { ...originalProcessEnv };
+    jest.useRealTimers();
+  });
```
JoshBinns2000, line 12: Does `jest.useFakeTimers()` not need to be reset after it is called (ie. do you not need an `afterAll` function)?

portal-api/src/scheduler/jobs/index.js
```diff
@@ -1,6 +1,7 @@
+const sendReportDueEmailsJob = require('./send-report-due-emails-job');
 const sendReportingPeriodStartEmailsJob = require('./send-reporting-period-start-emails-job');
 
 /**
  * @type {typeof import('../types/scheduler-job').SchedulerJob[]}
  */
-module.exports = [sendReportingPeriodStartEmailsJob];
+module.exports = [sendReportDueEmailsJob, sendReportingPeriodStartEmailsJob];
```
JoshBinns2000, line 7: Maybe put the elements of the array onto new lines to match the formatting elsewhere

portal-api/src/scheduler/jobs/send-report-due-emails-job/index.js
```diff
@@ -0,0 +1,113 @@
...
+ */
+const sendEmailForBank = async (bank, reportPeriod) => {
+  const { name: bankName, paymentOfficerTeam } = bank;
+  const paymentOfficerTeamEmail = paymentOfficerTeam?.email;
+
+  try {
+    if (!hasValue(paymentOfficerTeamEmail)) {
+      console.warn(`Not sending ${EMAIL_NAME} email to '${bankName}' - no payment officer team email set`);
+    } else if (!isValidEmail(paymentOfficerTeamEmail)) {
+      console.error(`Failed to send ${EMAIL_NAME} email to '${bankName}' - invalid payment officer team email '${paymentOfficerTeamEmail}'`);
```
JoshBinns2000, line 25: Don't we want to throw an error to trigger the catch of this block? Technically reaching either of these conditions results in an error where the email isn't sent unless I'm misunderstanding

portal-api/src/scheduler/jobs/send-report-due-emails-job/index.js
```diff
@@ -0,0 +1,113 @@
...
+  // TODO FN-1163 - check how this endpoint has been updated
+  const reportsResponse = await api.getUtilisationReports(bank.id);
+
+  if (reportsResponse?.status !== 200) {
+    const error = reportsResponse?.data ?? 'unknown error';
+    throw new Error(`Failed to get utilisation reports for ${bank.name} (bank ID: ${bank.id}): %s`, error);
+  }
+
+  const reportPeriod = getReportPeriodMonthAndYear();
+  return reportsResponse.data.some((report) => report.month === reportPeriod.month && report.year === reportPeriod.year);
```
JoshBinns2000, line 53: Could we maybe destructure `reportsResponse` from line 45 into `{ status, data }` and change the condition to `status && status !== 200`?

portal-api/src/scheduler/jobs/send-report-due-emails-job/index.js
```diff
@@ -0,0 +1,113 @@
...
+/**
+ * Attempts to send the emails if the report for the current period is due today
+ */
+const sendEmailsOnReportDueDate = async () => {
+  console.info(`Checking if ${EMAIL_NAME} emails should be sent today`);
+
+  const today = new Date();
+  const reportDueDate = await getReportDueDate();
+
+  if (isSameDay(today, reportDueDate)) {
```
JoshBinns2000, line 94: I like this usage of `date-fns` :)

portal-api/src/scheduler/jobs/send-report-due-emails-job/index.test.js
```diff
@@ -0,0 +1,146 @@
...
+
+const sendReportDueEmailsJobTask = sendReportDueEmailsJob.init().task;
+
+const NOV_2023_REPORT_DUE_DATE = new Date('2023-11-15');
+
+describe('sendReportDueEmailsJob', () => {
+  beforeEach(() => {
+    jest.resetAllMocks();
+    jest.useRealTimers();
+  });
```
JoshBinns2000, line 27: Same as above for possibly needing an `afterAll` statement

portal-api/src/scheduler/jobs/send-report-due-emails-job/index.test.js
```diff
@@ -0,0 +1,146 @@
...
+
+    externalApi.getBankHolidayDatesForRegion.mockResolvedValue([]);
+
+    api.getAllBanks.mockResolvedValue([MOCK_BANKS.HSBC]);
+
+    const existingReport = {
+      ...MOCK_UTILISATION_REPORT,
+      // 'month' should be 1-indexed, Date.getMonth() is 0-indexed, so this sets 'month' to the previous month
+      month: NOV_2023_REPORT_DUE_DATE.getMonth(),
+      year: NOV_2023_REPORT_DUE_DATE.getFullYear(),
```
JoshBinns2000, line 56: I've made some comments about `MOCK_UTILISATION_REPORT` below, but I think those changes can be implemented to improve how `existingReport` is defined

portal-api/test-helpers/mock-utilisation-reports.js
```diff
@@ -0,0 +1,16 @@
...
+  },
+  month: 1,
+  year: 2021,
+  dateUploaded: new Date('2023-11-07'),
+  path: 'test path',
+  uploadedBy: {
+    id: '123',
+    name: 'test user',
+  },
+};
```
JoshBinns2000, line 14: Considering your use case of this mock, you overwrite `month` and `year` but not `dateUploaded`. It may be more worthwhile to instead make a function that generates a report given a month and year. This will then make sure the `dateUploaded` matches what you would expect it to be and doesn't cause any issues down the line with those two fields not matching up

portal-api/src/scheduler/jobs/send-reporting-period-start-emails-job/index.test.js
```diff
@@ -1,9 +1,9 @@
 const { produce } = require('immer');
-const sendReportingPeriodStartEmailsJob = require('./index');
+const sendReportingPeriodStartEmailsJob = require('.');
 const api = require('../../../v1/api');
 const externalApi = require('../../../external-api/bank-holidays');
 const sendEmail = require('../../../external-api/send-email');
-const BANKS = require('../../../../test-helpers/mock-banks');
+const MOCK_BANKS = require('../../../../test-helpers/mock-banks');
```
JoshBinns2000, line 6: I like this change :)

JoshBinns2000 (changes requested):

portal-api/src/scheduler/jobs/helpers/utilisation-report-helpers.js
toddym42: Thanks, will update to match elsewhere 👍

portal-api/src/scheduler/jobs/helpers/utilisation-report-helpers.js
toddym42: I can update this to remove `startOfMonth` though as that's not needed. This was used to handle an edge case where the day of the current month were more than the new previous month. `subMonths` is clever enough to handle this case though. e.g. if I do
```
const lastMonthDate = subMonths(new Date('2023-03-31'), 1);
```
it will return a `Date('2023-02-28')` and not fail, or instead do anything weird like return `Date('2023-03-03')`.

portal-api/src/scheduler/jobs/index.js
JoshBinns2000: I didn't realise prettier formatted it this way - if it does then no issue :)

portal-api/src/scheduler/jobs/send-report-due-emails-job/index.js
JoshBinns2000: Nope I agree with that reasoning :)

JoshBinns2000 (dismissed):

.env.sample
```diff
@@ -39,6 +39,7 @@ SESSION_SECRET=
 # UTILISATION REPORT
 PORTAL_REPORT_DUE_DATE_BUSINESS_DAYS_FROM_START_OF_MONTH=10
 PORTAL_REPORTING_PERIOD_START_EMAIL_SCHEDULE='0 7 1 * *'
+PORTAL_REPORT_DUE_EMAIL_SCHEDULE='0 7 * * *'
```
ChristianMcc, line 42: It doesn't affect local/this PR but do we have a time we want to send this given to us by product?

.github/workflows/development_infrastructure.yml
```diff
@@ -1478,7 +1478,8 @@ jobs:
                 WEBSITES_PORT=5000 \
                 RATE_LIMIT_THRESHOLD=${{ vars.RATE_LIMIT_THRESHOLD }} \
                 PORTAL_REPORTING_PERIOD_START_EMAIL_SCHEDULE=${{ vars.PORTAL_REPORTING_PERIOD_START_EMAIL_SCHEDULE }} \
-                PORTAL_REPORT_DUE_DATE_BUSINESS_DAYS_FROM_START_OF_MONTH=${{ vars.PORTAL_REPORT_DUE_DATE_BUSINESS_DAYS_FROM_START_OF_MONTH }}
+                PORTAL_REPORT_DUE_DATE_BUSINESS_DAYS_FROM_START_OF_MONTH=${{ vars.PORTAL_REPORT_DUE_DATE_BUSINESS_DAYS_FROM_START_OF_MONTH }} \
```
ChristianMcc, line 1481: Can we call this UTILISATION_REPORT_DUE_DATE.... instead please? I can imagine we want to commonise this variable in TFM at some point when we are saying when stuff is due
Sorry this will affect the bicep PR too!

portal-api/src/scheduler/jobs/helpers/utilisation-report-helpers.js
```diff
@@ -0,0 +1,47 @@
+const { startOfMonth, format, subMonths } = require('date-fns');
+const externalApi = require('../../../external-api/api');
```
ChristianMcc, line 2: I've been thinking and I think I prefer having notify logic in external API for this as we don't care about response times etc., though given on Francesca's work it brings in a delay of another HTTPS call i'm a little more apprehensive. Let's leave it like this but discuss it next week if we reckon this is neater and worth the small performance hit on portal routes

portal-api/src/scheduler/jobs/helpers/utilisation-report-helpers.test.js
```diff
@@ -0,0 +1,118 @@
+const { getReportDueDate, getFormattedReportDueDate, getReportPeriodMonthAndYear, getFormattedReportPeriod } = require('./utilisation-report-helpers');
```
ChristianMcc, line 1: Nice tests :)

portal-api/src/scheduler/jobs/send-report-due-emails-job/index.js
```diff
@@ -0,0 +1,113 @@
...
+  const paymentOfficerTeamEmail = paymentOfficerTeam?.email;
+
+  try {
+    if (!hasValue(paymentOfficerTeamEmail)) {
+      console.warn(`Not sending ${EMAIL_NAME} email to '${bankName}' - no payment officer team email set`);
+    } else if (!isValidEmail(paymentOfficerTeamEmail)) {
+      console.error(`Failed to send ${EMAIL_NAME} email to '${bankName}' - invalid payment officer team email '${paymentOfficerTeamEmail}'`);
+    } else {
+      await sendEmail(EMAIL_TEMPLATE_IDS.UTILISATION_REPORT_DUE_TODAY, paymentOfficerTeamEmail, {
+        recipient: paymentOfficerTeam.teamName,
```
ChristianMcc, line 28: We'll want to send the email even if we don't have a teamName, but just because we won't really look at what emails are sent we'll not know if teamName isn't populated unless a bank complains.
Can we add a simple console.warn on a check if teamName is null somewhere?

portal-api/src/scheduler/jobs/send-report-due-emails-job/index.js
```diff
@@ -0,0 +1,113 @@
+const { format, isSameDay } = require('date-fns');
```
ChristianMcc, line 1: All LGTM, only comments will probably be on how it's commonised with the late report job so look forward to that excitement!

ChristianMcc (commented):
Nice work, only a couple of nit picks. Hopefully if you agree with the variable rename then it's not too much pain to update!

.github/workflows/development_infrastructure.yml
toddym42: Have updated all the new env vars to being `UTILISATION_REPORT...` 👍

portal-api/src/scheduler/jobs/helpers/utilisation-report-helpers.js
toddym42: It feels like `external-api` should just be a shared package that provides an interface for external providers. That way we'd have the best of both worlds - no extra API requests, and shared code across all of the dtfs2 packages.

.env.sample
toddym42: I'll check 👍

ChristianMcc (approved):

PR #2289, opened 2023-11-08
feat(DTfS2-6784): update built mojFrontend files to reflect jQuery version bump

oscar-richardson-softwire:
### Introduction

We currently have changes in the source files that are not reflected in the built files. In the portal service, we have changed from version 3.7.0 to 3.7.1 of jQuery (which is a dependency of mojFrontend), but the built files have not been updated to reflect this.

We need to run npm run build in the portal service to update the built files. This initially breaks mojFrontend, but after some investigation it was found that the integrity attributes in the script tags that use the mojFrontend script needed to be updated to reflect the new hash of the rebuilt file.

In order to prevent a similar investigation in the future, we should document in the README how the build process works and the steps that need to be taken by the developer.

### Resolution

- investigated why updating mojFrontend broke it - found hashes needed updating
- updated hashes in script tags
- added documentation about the build process


lmarrai-sw (approved):
thanks @oscar-richardson-softwire

PR #2291, opened 2023-11-08
feat(FN-1295): back link and error positioning

JoshBinns2000:
# Introduction

The position of the error summary box and back button on some of the utilisation report upload pages doesn't match the designs.

# Resolution

The back link is placed into the top-level `index.njk` function so that it appears directly below the header. The back link can be used in a specific njk file by setting the `backLink` object via `{% set backLink = { href: '/[path]' } %}`. The back link will not appear if your njk template file doesn't include this line. The resulting appearance of `confirm-and-send.njk` and `check-the-report.njk` are shown below.

The error summary box is moved above the `<h1>` in the relevant njk file.

image
image

### DIFF
--- a/portal/templates/index.njk
+++ b/portal/templates/index.njk
@@ -33,6 +33,17 @@
 
     <div class="govuk-width-container">
 
+      {% from "govuk/components/back-link/macro.njk" import govukBackLink %}
+      {% if backLink and backLink.href %}
+        {{ govukBackLink({
+          text: "Back",
+          href: backLink.href,
+          attributes: {
+              "data-cy": "back-link"
+          }
+        }) }}
+      {% endif %}
+
       <main class="govuk-main-wrapper govuk-main-wrapper--auto-spacing" id="main-content" role="main">
         {% block content %}{% endblock %}
       </main>
--- a/portal/templates/index.njk
+++ b/portal/templates/index.njk
@@ -21,6 +21,7 @@
   <body class="govuk-template__body ">
 
     <script src="/assets/js/jsEnabled.js" type="text/javascript" integrity="sha512-+oUPyU7P4RBV7u2B6RkjEWzdtDSbzvPyIOKS7YelVhYUftt4L261FozUDCXcQG6ITqkIPfZW91okPyupbLd0xA==" crossorigin="anonymous"></script>
+    {% from "govuk/components/back-link/macro.njk" import govukBackLink %}
     {% include "_partials/skip-link.njk" %}
 
     {% if user and user.roles and user.roles.includes('admin') %}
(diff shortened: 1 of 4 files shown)


portal/templates/index.njk
```diff
@@ -33,6 +33,17 @@
 
     <div class="govuk-width-container">
 
+      {% from "govuk/components/back-link/macro.njk" import govukBackLink %}
```
toddym42, line 36: I think I would expect all imports at the top of the file. Does that work here?

portal/templates/index.njk
```diff
@@ -33,6 +33,17 @@
...
+      {% if backLink and backLink.href %}
+        {{ govukBackLink({
+          text: "Back",
+          href: backLink.href,
+          attributes: {
+              "data-cy": "back-link"
```
toddym42, line 42: Indenting is slightly off here. Looks like 2 spaces in this file, though I see this is not consistent across all .njk files.

portal/templates/index.njk
```diff
@@ -33,6 +33,17 @@
 
     <div class="govuk-width-container">
 
+      {% from "govuk/components/back-link/macro.njk" import govukBackLink %}
+      {% if backLink and backLink.href %}
+        {{ govukBackLink({
```
toddym42, line 38: Are you able to add a screenshot of how this looks now? Always useful to include a screenshot for visual changes so people can more easily comment on the design as well as the technical implementation.

portal/templates/index.njk
```diff
@@ -33,6 +33,17 @@
 
     <div class="govuk-width-container">
 
+      {% from "govuk/components/back-link/macro.njk" import govukBackLink %}
+      {% if backLink and backLink.href %}
```
toddym42, line 37: Should we also make the same change in the `gef-ui` package?

JoshBinns2000: Discussed with @ChristianMcc : we would need to run this change past dtfs and the back link is apparently working in gef-ui, so let's not also change gef-ui

portal/templates/index.njk
JoshBinns2000: I've moved it to the top level of `<script>` but I don't believe I can move it any higher than that

toddym42 (approved):
Looks great : )

francescastocco (commented):
Nice find 👏 in portal the back link is also used on the `user-edit.njk` template, does that fall under 'check with dtfs' @ChristianMcc too?

francescastocco:
> Nice find 👏 in portal the back link is also used on the `user-edit.njk` template, does that fall under 'check with dtfs' too @ChristianMcc?

Oops, I missed that you'd already done this :)

francescastocco (approved):

PR #2295, opened 2023-11-09
fix(DTFS2-6788): nunjucks boolean check

abhi-markan:
## Introduction ✏️ 
EC are being shown as all failed on TFM.

## Resolution ✔️ 
* Updated condition to Boolean instead of a String.

## Miscellaneous ➕ 
* Dependencies updates


trade-finance-manager-ui/templates/case/deal/_macros/eligibility-criteria-answer-tag.njk
```diff
@@ -1,12 +1,10 @@
 {% from "govuk/components/tag/macro.njk" import govukTag %}
 
 {% macro render(params) %}
-
-  {% if params.answer === 'true' %}
+  {% if params.answer === true %}
```
lmarrai-sw, line 4: @abhi-markan do we need `=== true` or can we just use `if params.answer`?

lmarrai-sw (dismissed):

lmarrai-sw (dismissed):

lmarrai-sw (dismissed):

lmarrai-sw (approved):

PR #2296, opened 2023-11-09
fix(DTFS2-6789): allow read-only users to have access to all banks

lmarrai-sw:
## Introduction

When we added the `read-only` role, we thought that only the `admin` role was able to have access to all banks (when the user object has `bank.id = '*'`). However, since deploying, we have realised that the `read-only` role should be able to have access to all banks too.

## Resolution

The `validateBank` middleware in `portal` and `gef-ui` now allows users to access all banks if they have `bank.id === '*'` and the user has the `admin` _or_ `read-only` role 

## Misc

There is a `isSuperUser` function in `portal`, `portal-api`, and `gef-ui` that checks `user.bank.id === '*'` but does not check the user roles so they do not need to change for this fix

### DIFF
--- a/gef-ui/server/middleware/validateBank/index.js
+++ b/gef-ui/server/middleware/validateBank/index.js
@@ -1,4 +1,5 @@
 const { ADMIN, READ_ONLY } = require('../../constants/roles');
+const { ALL_BANKS_ID } = require('../../constants');
 const api = require('../../services/api');
 
 const validRolesForAccessingAllBanks = [
@@ -6,9 +7,14 @@ const validRolesForAccessingAllBanks = [
   READ_ONLY,
 ];
 
+/**
+ * Returns `true` if the user has permission to access data from all banks, and `false` otherwise.
+ * @param {{ bank: {id: string}, roles: string[] }} user
+ * @returns {Boolean}
+ */
 const userCanAccessAllBanks = (user) => {
   const userBankId = user?.bank?.id;
-  if (userBankId !== '*') {
+  if (userBankId !== ALL_BANKS_ID) {
     return false;
   }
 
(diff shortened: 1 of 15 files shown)


AlexBramhill (dismissed):

gef-ui/server/middleware/validateBank/index.js
```diff
@@ -1,17 +1,30 @@
...
+const { ADMIN, READ_ONLY } = require('../../constants/roles');
 const api = require('../../services/api');
 
-const ukefAdminRoles = [ADMIN];
+const validRolesForAccessingAllBanks = [
+  ADMIN,
+  READ_ONLY,
+];
+
+const userCanAccessAllBanks = (user) => {
```
abhi-markan, line 9: Can you please add function documentation.

gef-ui/server/middleware/validateBank/index.js
```diff
@@ -1,17 +1,30 @@
...
+  const userBankId = user?.bank?.id;
+  if (userBankId !== '*') {
```
abhi-markan, line 11: Can we pick `*` from Constant?

portal/server/routes/middleware/validateBank/index.js
```diff
@@ -1,18 +1,30 @@
...
+  READ_ONLY,
+];
 
-const validateBank = async (req, res, next) => {
-  const { _id: dealId } = req.params;
-  const { id: bankId } = req.session.user.bank;
-  const token = req.session.userToken;
+const userCanAccessAllBanks = (user) => {
+  const userBankId = user?.bank?.id;
+  if (userBankId !== '*') {
```
abhi-markan, line 11: Constant please?

abhi-markan (changes requested):
3 minor comments to address please.

abhi-markan (approved):

PR #2298, opened 2023-11-09
feat(FN-978): handle report upload failures and 409 on double submission

ChristianMcc:
## Introduction
We need to return errors when uploading the report fails. 

## Resolution
Return the error page for any errors that happen during the submission process.
As part of this I have implemented 409s in the APIs if you try and submit the report when you already have a report saved.
Also I redid the url structure for the /utilisation-reports endpoints to /banks/:bankId/utilisation-reports. In a separate PR I'll look at redoing the POST url structure (to mimic this change of being bank centric)


dtfs-central-api/src/services/repositories/utilisation-reports-repo.js
```diff
@@ -32,6 +32,19 @@
...
+/**
+ * Saves the utilisation report details but not data to the database.
+ * @param {Object} bankId - Id of the bank.
+ * @param {number} month - Month of utilisation report.
+ * @param {number} year - Year of utilisation report.
+ * @returns {Object | null} - Utilisation report details matching the bank/month/year combo or null if it doesn't exist.
+ */
+const getUtilisationReportDetailsForMonthAndYear = async (bankId, month, year) => {
+  const utilisationReportDetailsCollection = await db.getCollection(DB_COLLECTIONS.UTILISATION_REPORTS);
+  const matchingReportDetails = await utilisationReportDetailsCollection.findOne({ 'bank.id': bankId, month, year });
```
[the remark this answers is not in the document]
ChristianMcc: I call `validatePayload` in `postUtilisationReportData` which validates month and year, it looks like I'm not validating bankId within central API so I'll need to fix that

ChristianMcc:
For my E2E tests I also need to have Josh's changes in to allow me to reset the database otherwise tests will start failing in the same spec

portal-api/src/v1/validation/validateIds.js
```diff
@@ -44,10 +44,22 @@ const isValidCompaniesHouseNumber = (companiesHouseNumber) => isValidRegex(COMPA
...
 
+/**
+ * isValidMonth
+ * checks if month is an integer between 1 and 12
+ * @param {Number} month
+ * @returns {Boolean} asserts if regex is matched
+ */
+const isValidMonth = (month) => Number.isInteger(month) && month >= 1 && month <= 12;
+
+const isValidYear = (year) => Number.isInteger(year) && year >= 2000 && year <= 2100;
```
ChristianMcc, line 55: add function documentation

dtfs-central-api/src/services/repositories/utilisation-reports-repo.js
```diff
@@ -32,6 +32,19 @@ const saveUtilisationReportDetails = async (bank, month, year, csvFilePath, uplo
   return { reportId: savedDetails?.insertedId?.toString(), dateUploaded: utilisationReportInfo.dateUploaded };
 };
 
+/**
+ * Saves the utilisation report details but not data to the database.
+ * @param {Object} bankId - Id of the bank.
```
toddym42, line 37: Should this be type `{string}`?

dtfs-central-api/src/v1/controllers/utilisation-report-service/previous-reports.controller.js
```diff
@@ -1,14 +1,22 @@
...
+      const utilisationReport = await getUtilisationReportDetailsForMonthAndYear(bankId, month, year);
+      utilisationReports = utilisationReport ? [utilisationReport] : [];
+    } else {
+      utilisationReports = await getUtilisationReportDetails(bankId);
+    }
 
     res.status(200).send(utilisationReports);
   } catch (error) {
-    console.error('Unable to get utilisation reports %s', error);
+    console.error('Unable to get utilisation reports: ', error);
```
toddym42, line 19: Nit: can remove the space in `reports: `. When you just add `error` as a second argument this is done for you.

dtfs-central-api/src/v1/controllers/utilisation-report-service/upload-utilisation-report.controller.js
```diff
@@ -32,7 +32,13 @@ const postUtilisationReportData = async (req, res) => {
       return res.status(400).send(validationErrors);
     }
 
-    // TODO: FN-967 want to 429 here if a report details already exists
+    // If a report already exists for this month/year/bank combo, return 409
```
toddym42, line 35: This comment feels a bit unnecessary as the intention seems clear enough from the code. I think `existingReports` should be singular though.

dtfs-central-api/src/v1/routes/portal-routes.js
```diff
@@ -908,6 +908,6 @@ portalRouter.route('/gef/mandatory-criteria/version/:version').get(mandatoryCrit
  *         description: Server conflict
  */
 portalRouter.route('/utilisation-reports').post(utilisationReportUpload.postUtilisationReportData);
-portalRouter.route('/previous-reports/:bankId').get(validation.bankIdValidation, handleValidationResult, previousReports.getUtilisationReports);
+portalRouter.route('/banks/:bankId/utilisation-reports').get(validation.bankIdValidation, handleValidationResult, previousReports.getUtilisationReports);
```
toddym42, line 911: Should this go in `bank-routes.js` now? If so we should change also it to `'/bank/:bankId/utilisation-reports'` (singular `bank`).

portal-api/src/v1/controllers/utilisation-report-service/previous-reports.controller.js
```diff
@@ -70,19 +70,14 @@ const getPreviousReportsByBankId = async (req, res) => {
...
-
-    const reports = mapAndSortReports(data);
+    const sortedReports = mapAndSortReports(reports);
 
-    return res.status(200).send(reports);
+    return res.status(200).send(sortedReports);
   } catch (error) {
     console.error('Unable to get previous reports %O', error);
-    return res.status(500).send({ status: 500, message: 'Failed to get previous reports' });
+    return res.status(500).send({ status: error.status ?? 500, message: 'Failed to get previous reports' });
```
toddym42, line 80: I think this error would be of type AxiosError, in which case I think the status would come from `error.response?.status`

portal-api/src/v1/validation/validateIds.js
```diff
@@ -44,10 +44,22 @@ const isValidCompaniesHouseNumber = (companiesHouseNumber) => isValidRegex(COMPA
  */
 const isValidBankId = (bankId) => isValidRegex(/^\d+$/, bankId);
 
+/**
+ * isValidMonth
+ * checks if month is an integer between 1 and 12
+ * @param {Number} month
+ * @returns {Boolean} asserts if regex is matched
```
toddym42, line 51: Would use lower case `number` and `boolean` here as primitives are used in these functions.

dtfs-central-api/src/v1/routes/portal-routes.js
ChristianMcc: While some of this is preference on single vs plural and the preference in this codebase is plural, I also think plural nouns are better for this kind of thing. We are picking a bank from a collection of banks and getting utilisation-reports from that bank.

For bank-routes now I considered this and was struggling with getting my head round the portal and tfm sets of routes which exist due to the different collections, but if they also just contained 'portal related' routes in an annoying sort of way. I think I'm happy to move it to bank-routes and it's a good suggestion just thinking out loud :D

ChristianMcc: Oh no, under /bank its /bank instead of /banks :( I should probably fix this but might raise a separate tech debt ticket to refactor (we did some pluralisation of nouns refactoring not that long ago). Looks like for now it'll have to just be /bank, sad times

ChristianMcc: I noticed inyour PR the get all banks endpoints /bank

toddym42: If we do refactor `bank-routes.js`to use `/banks` instead of `/bank` we should wait until this PR has gone through (to remove the duplicate bank routes from `portal-routes.js`) as `portal-routes.js` is currently already using `/banks`.

ChristianMcc: Yep can do it separately as will need to run it by DTFS team first anyway

dtfs-central-api/src/v1/controllers/utilisation-report-service/previous-reports.controller.js
toddym42, line 1: Wondering if the name of this file should be something like `get-utilisation-reports.controller.js`? The name `previous-reports` seems too tightly coupled to one of the pages this is used for in the front-end.

dtfs-central-api/src/v1/swagger-definitions/utilisationReport.js
```diff
@@ -0,0 +1,41 @@
...
+ *   UtilisationReport:
+ *     type: object
+ *     properties:
+ *       id:
+ *         type: string
+ *         example: '9'
+ *         description: Bank id. Separate MongoDB _id
+ *       month:
+ *         type: integer
+ *         example: 1
```
toddym42, line 13: Maybe worth adding a description here to clarify that this is 1-indexed?

portal-api/src/v1/api.js
```diff
@@ -268,23 +268,35 @@ const saveUtilisationReport = async (reportData, month, year, user, filePath) =>
...
+    }
+
+    const queryString = month ? `?month=${month}${year ? `&year=${year}` : ''}` : '';
+
     const response = await axios({
       method: 'get',
-      url: `${DTFS_CENTRAL_API_URL}/v1/portal/previous-reports/${bankId}`,
+      url: `${DTFS_CENTRAL_API_URL}/v1/bank/${bankId}/utilisation-reports${queryString}`,
       headers: headers.central,
     });
```
toddym42, line 294: I think you could use the `param` config attribute in the axios request to set the query params instead of hand rolling it. So this could instead be:
```
const response = await axios.get(`${DTFS_CENTRAL_API_URL}/v1/bank/${bankId}/utilisation-reports`, {
  params: { month, year },
  headers: headers.central,
});
```

ChristianMcc: Ah didn't know this existed, so much nicer!

dtfs-central-api/src/v1/routes/bank-routes.js
```diff
@@ -4,6 +4,10 @@ const bankRouter = express.Router();
 
 const getBankController = require('../controllers/bank/get-bank.controller');
 const createBankController = require('../controllers/bank/create-bank.controller');
+const previousReports = require('../controllers/utilisation-report-service/get-utilisation-reports.controller');
```
toddym42, line 7: This should be `getUtilisationReportsController` now.

francescastocco (approved):

PR #2299, opened 2023-11-09
feat(FN-970): fix month indexing in session

francescastocco:
## Introduction

Report month is stored in the session, it was indexed at 0 but we want to use index 1 to make it easier to read

## Resolution

When we set the month increase the value by 1

### DIFF
--- a/portal/jest.config.js
+++ b/portal/jest.config.js
@@ -1,7 +1,7 @@
 module.exports = {
   collectCoverageFrom: ['server/controllers/**/*.{js,}', 'server/routes/**/*.{js,}', 'server/helpers/**/*.{js,}', 'scripts/**/*.{js,}'],
   coverageDirectory: 'generated_reports/coverage/unit',
-  testMatch: ['**/getOneIndexedMonth.test.js'],
+  testMatch: ['**/*.test.js', '**/*.component-test.js'],
   moduleNameMapper: {
     '^.+\\.(css|less|scss)$': 'babel-jest',
     '^exceljs$': 'babel-jest',
--- a/portal/jest.config.js
+++ b/portal/jest.config.js
@@ -1,7 +1,7 @@
 module.exports = {
   collectCoverageFrom: ['server/controllers/**/*.{js,}', 'server/routes/**/*.{js,}', 'server/helpers/**/*.{js,}', 'scripts/**/*.{js,}'],
   coverageDirectory: 'generated_reports/coverage/unit',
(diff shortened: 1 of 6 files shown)


ChristianMcc (dismissed):
LGTM :)

portal/server/controllers/utilisation-report-service/utilisation-report-upload/utilisation-report-status.js
```diff
@@ -20,7 +20,7 @@ const getBankHolidays = async (userToken) => {
 const getReportPeriod = () => {
   const lastMonth = subMonths(new Date(), 1);
   const reportPeriod = format(lastMonth, 'MMMM yyyy');
-  const month = getMonth(lastMonth);
+  const month = getMonth(lastMonth) + 1; // Account for javascript 0-index for months
```
toddym42, line 23: I guess we're going to have to do this a lot because javascript. Maybe it'd be worth creating a shared `getOneIndexMonth` function that we can use in all these cases. Then we also wouldn't need to add a comment to explain the reason.

toddym42 (dismissed):

portal/server/controllers/utilisation-report-service/utilisation-report-upload/utilisation-report-status.js
francescastocco: Yep, I've added it :)

portal/jest.config.js
```diff
@@ -1,7 +1,7 @@
 module.exports = {
   collectCoverageFrom: ['server/controllers/**/*.{js,}', 'server/routes/**/*.{js,}', 'server/helpers/**/*.{js,}', 'scripts/**/*.{js,}'],
   coverageDirectory: 'generated_reports/coverage/unit',
-  testMatch: ['**/*.test.js', '**/*.component-test.js'],
+  testMatch: ['**/getOneIndexedMonth.test.js'],
```
toddym42, line 4: Was this intentional?

toddym42, line 4: Looks like this needs to be reverted?

ChristianMcc: fwiw for unit tests I use the 'jest runner' package in vscode to run these tests easily. saves having to do this kind of thing - note it'll only pick up file called .test.js so component tests and api-tests I will rename to get jest runner to pick up.

Having written this down I'm sure you can configure the extension to not have to do that but I've never bothered

toddym42 (approved):

PR #2301, opened 2023-11-10
fix(gef-readonly): added read-only to GEF deals users

abhi-markan:
## Introduction ✏️ 
`read-only` user is unable to access `GEF` deal on Portal.

## Resolution ✔️ 
* Added `READ_ONLY` to valid list of users with bank set as `*`.
* Added `console.error` when unauthorised.
* Added documentation.


AlexBramhill (dismissed):

avaitonis (approved):
LGTM

PR #2302, opened 2023-11-10
Feat/dtfs2 6781/add typescript compatibility to portal api

MarRobSoftwire:
## Introduction
Currently `portal-api` does not use typescript but it would improve efficiency & reduce bugs to allow all the services to accept typescript files

## Resolution
- updated the `package.json`
    - changed the `start` script to `npx ts-node server/index.ts`
- updated the `init.sh` file to use `ts-node`
- Added `tsconfig.json`
    - This allows but does not check js files 
- Added `ts-config.eslint.json`
- Changed `src/index` and `src/azure-env/index` from js to ts to verify the change works.
    - replaced `require`/`module.exports` with `import`/`export`

## Note
- I haven't been able to run this on my VM yet, but it compiles to the point of needed environment variables.
    - When running this locally without environment variables I get a `TypeError: Cannot read properties of undefined (reading 'substring')` rather than `InvalidEnvironmentVariableError` for the missing `GOV_NOTIFY_API_KEY`
- This is branched off #2279, this PR covers the changes to `portal-api`


dtfs-central-api/Dockerfile
```diff
@@ -27,6 +27,7 @@ WORKDIR /app
 COPY package.json .
 COPY package-lock.json .
 RUN npm ci --legacy-peer-deps
+RUN npm install -g typescript
```
MarRobSoftwire, line 30: These need adding to relevant PRs, just added here to check pipeline

docker-compose.gha.yml
```diff
@@ -167,8 +167,7 @@ services:
       - TZ=Europe/London
       - NODE_ENV=${NODE_ENV}
       - RATE_LIMIT_THRESHOLD
-
-    entrypoint: npx nodemon src/index.js -L
+    entrypoint: npx ts-node src/index.ts
```
MarRobSoftwire, line 170: ```npx nodemon --exec npx ts-node src/index.ts```

lmarrai-sw (dismissed):

lmarrai-sw (approved):

PR #2305, opened 2023-11-13
feat(DTFS2-6770): allow user to resend sign in link

lmarrai-sw:


portal/server/controllers/login/check-your-email.js
```diff
@@ -0,0 +1,48 @@
...
+module.exports.sendNewSignInLink = async (req, res) => {
+  const { session: { numberOfSendSignInLinkAttemptsRemaining } } = req;
+
+  if (!numberOfSendSignInLinkAttemptsRemaining || numberOfSendSignInLinkAttemptsRemaining < 0) {
+    console.error(`User attempted to send a new sign in link without any attempts remaining. Attempts remaining count is ${numberOfSendSignInLinkAttemptsRemaining}.`);
+    return res.status(403).render('_partials/problem-with-service.njk');
+  }
+
+  // TODO DTFS2-6770: Should we worry about concurrency?
+  // TODO DTFS2-6770: Should we move this count to the database?
```
AlexBramhill, line 35: I think this will have to be moved to the database (maybe if we go ahead with the changes as a result of webmail). What do you think?

portal/server/controllers/login/check-your-email.js
```diff
@@ -0,0 +1,48 @@
...
+
+module.exports.sendNewSignInLink = async (req, res) => {
+  const { session: { numberOfSendSignInLinkAttemptsRemaining } } = req;
+
+  if (!numberOfSendSignInLinkAttemptsRemaining || numberOfSendSignInLinkAttemptsRemaining < 0) {
+    console.error(`User attempted to send a new sign in link without any attempts remaining. Attempts remaining count is ${numberOfSendSignInLinkAttemptsRemaining}.`);
+    return res.status(403).render('_partials/problem-with-service.njk');
+  }
+
+  // TODO DTFS2-6770: Should we worry about concurrency?
```
AlexBramhill, line 34: Are you meaning for the same user?

portal/server/routes/login/index.js
```diff
@@ -60,9 +62,12 @@ router.post('/login', async (req, res) => {
   try {
     const tokenResponse = await api.login(email, password);
 
-    const { token, loginStatus } = tokenResponse;
+    const { token, loginStatus, user: { email: userEmail } } = tokenResponse;
     req.session.userToken = token;
     req.session.loginStatus = loginStatus;
+    req.session.numberOfSendSignInLinkAttemptsRemaining = 2;
+    // TODO DTFS2-6770: Are we right to avoid setting `user.email` (in case this triggers UI elements)?
```
AlexBramhill, line 69: Interesting, I didn't set this originally as it wasnt needed. I'm assuming this is due to the requirement to resend email.

As you mentioned - this does draw the top bar if set as `user.email`. This solution is good, i think, but I think it might be worth commenting why we are doing this as `userEmail`. What do you think?

portal/server/routes/middleware/validatePartialAuthToken/index.js
```diff
@@ -0,0 +1,38 @@
...
+ * @param {Object} res Response object
+ * @param {Object} next Next object
+ */
+const validatePartialAuthToken = async (req, res, next) => {
+  try {
+    const { userToken } = req.session;
+
+    if (!userToken) {
+      console.info('User has not logged in, destroying the session');
+      return destroySessionAndRedirectToStart(req, res);
```
AlexBramhill, line 15: is there a reason we need to destroy an empty session here out of interest?

portal/server/api.js
```diff
@@ -52,33 +52,23 @@ const sendSignInLink = async (token) => axios({
...
-    });
+const loginWithSignInLink = async ({ token: requestAuthToken, signInToken }) => {
+  const response = await axios({
+    method: 'post',
+    url: `${PORTAL_API_URL}/v1/users/me/sign-in-link/${signInToken}/login`,
+    headers: {
+      'Content-Type': 'application/json',
+      Authorization: requestAuthToken,
+    },
+    data: { signInToken },
```
AlexBramhill, line 63: out of curiosity, why are we needing a payload here at all?

portal/server/generateApp.js
```diff
@@ -124,6 +124,7 @@ const generateApp = () => {
   // error handler
   app.use((error, req, res, next) => {
     if (error.code === 'EBADCSRFTOKEN') {
+      console.error('The user\'s CSRF token is incorrect, redirecting the user to /.');
```
AlexBramhill, line 127: this reads slightly odd, happy to keep as is, but could also use `PORTAL_API_URL`?

portal/templates/login/we-have-sent-you-another-link.njk
```diff
@@ -11,7 +11,7 @@
 
   <div class="govuk-grid-row govuk-body">
     <div class="govuk-grid-column-three-quarters">
-      <p class="govuk-!-margin-bottom-6">We've sent you another email with a link to {{emailAddress}}.</p>
+      <p class="govuk-!-margin-bottom-6" data-cy="sign-in-link-target-email-address">We've sent you another email with a link to {{ signInLinkTargetEmailAddress }}.</p>
```
AlexBramhill, line 14: Do we want to rename this var to something like `obscuredSignInLinkTargetEmailAddress`

portal/server/routes/middleware/validatePartialAuthToken/index.test.js
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  function itDestroysTheSession() {
+    it('destroys the session', async () => {
+      await validatePartialAuthToken(req, res, next);
+      expect(req.session.destroy).toHaveBeenCalled();
+    });
+  }
+
+  async function itRedirects() {
```
MarRobSoftwire, line 90: Should the name specify redirecting to login?

portal-api/api-tests/v1/users/users.api-test.js
```diff
@@ -31,18 +32,18 @@ describe('a user', () => {
     await wipeDB.deleteUser(MOCK_USER);
   });
 
-  afterAll(async () => {
-    await wipeDB.wipe(['users']);
-  });
+  // afterAll(async () => {
```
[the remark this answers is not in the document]
AlexBramhill: Woops!

portal/server/routes/middleware/validatePartialAuthToken/index.test.js
AlexBramhill: updated!

e2e-tests/portal/cypress/e2e/journeys/login/resend-sign-in-link.spec.js
```diff
@@ -0,0 +1,85 @@
...
+            expect(response.status).to.eq(403);
+          });
+        });
+    });
+
+    it('The user is shown the email address that sign in links are being sent to after resending the link 2 times', () => {
+      checkYourEmail.sendNewSignInLink();
+      checkYourEmail.sendNewSignInLink();
+      checkYourEmail.obscuredSignInLinkTargetEmailAddressText().should('contain', userAnonymisedEmailAddress);
+      checkYourEmail.visit();
```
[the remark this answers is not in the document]
AlexBramhill: agree, this is covered below in `The session remembers how many sign in attempts are remaining` tests

MarRobSoftwire (approved):

portal-api/test-helpers/api-test-helpers/database/user-repository.js
```diff
@@ -0,0 +1,87 @@
...
+    throw new Error(`Failed to create logged in user session for user: ${user.username}: ${e}`);
+  }
+};
+
+const createLoggedInUserSession = async (user) => {
+  try {
+    const userCollection = await db.getCollection('users');
+    const userFromDatabase = await userCollection.findOne({ username: { $eq: user.username } }, { collation: { locale: 'en', strength: 2 } });
+
+    const sessionIdentifier = crypto.randomBytes(32).toString('hex');
```
oscar-richardson-softwire, line 77: Can we commonize these lines in this function and the previous one into a function like `getUserAndSessionIdFromDb`, for instance?

portal-api/test-helpers/api-test-helpers/database/database.config.js
```diff
@@ -0,0 +1,12 @@
+const dotenv = require('dotenv');
+
+dotenv.config();
+
+const dbName = process.env.MONGO_INITDB_DATABASE;
+// The below is different than MONGOD_URI env var as mock-data-loader does not run within the container
```
oscar-richardson-softwire, line 6: I think this means that this is a different value from the `MONGODB_URI` environment variable in the root-level portal-api .env file (i.e., there'll be another .env file within `portal-api/test-helpers` with a different value for this env var). If I'm right in thinking this, I would change comment to something like:

`// The below value is different in the portal-api/test-helpers .env file compared to the root-level portal-api .env file as the mock-data-loader does not run within the container`

portal-api/test-helpers/api-test-helpers/database/user-repository.js
AlexBramhill: I'm deliberately being a bit more expressive of whats going on here (as it's for tests rather than code) -- also they have enough variation between lines imo to justify not being commonised. What do you think?

portal-api/test-helpers/api-test-helpers/database/database.config.js
AlexBramhill: I've updated this -- hopefully makes sense

PR #2314, opened 2023-11-14
feat(FN-1164): add report overdue chaser email job

toddym42:
## Ticket

[FN-1164](https://ukef-dtfs.atlassian.net/browse/FN-1164)

## Introduction

Banks report facility utilisation for the previous month from the 1st of each month until the 10th business day.

If a bank's report is overdue (a certain number of configurable business days into the month) they should be sent a follow-up email to inform them that the report is overdue.

## Resolution

* Add new job (`send-report-overdue-emails-job`) to the `portal-api` scheduler to check each day if the chaser emails should be sent (we can't have a cron schedule based on business days, so we instead schedule the job to run each day and check if today is the 10th business day)
* On the configured business day, send email to all banks that have not already submitted a report for the current period and have a valid email set to the `paymentOfficerTeam.email` property
* Refactored existing jobs to used new shared code
* Updated `send-reporting-period-start-emails-job` to `send-report-submission-period-start-emails-job`. This ensures we use common terminology:
  * reporting period - the date range that we are providing reporting detail for (usually the previous month)
  * submission period - the date range that we expect banks to submit their reports for the report period (e.g. the 1st on the current month until the 10th business day)

## To Do

* Currently makes use of the central api `/portal/previous-reports/:bankId` route, which is currently being changed in a separate ticket. Will need updating when changes are complete.


.env.sample
```diff
@@ -38,8 +38,10 @@ SESSION_SECRET=
 
 # UTILISATION REPORT
 UTILISATION_REPORT_DUE_DATE_BUSINESS_DAYS_FROM_START_OF_MONTH=10
-UTILISATION_REPORT_REPORTING_PERIOD_START_EMAIL_SCHEDULE='0 7 1 * *'
+UTILISATION_REPORT_OVERDUE_CHASER_DATE_BUSINESS_DAYS_FROM_START_OF_MONTH=20
+UTILISATION_REPORT_SUBMISSION_PERIOD_START_EMAIL_SCHEDULE='0 7 1 * *'
```
toddym42, line 42: Note additional change of
```
UTILISATION_REPORT_REPORTING_PERIOD_START_EMAIL_SCHEDULE
```
to
```
UTILISATION_REPORT_SUBMISSION_PERIOD_START_EMAIL_SCHEDULE
```
as explained in the PR description. Reminder to self that I will need to get the GitHub env var updated when this PR is merged.

portal-api/src/scheduler/jobs/helpers/utilisation-report-helpers.js
```diff
@@ -41,6 +59,24 @@ const getFormattedReportPeriod = () => {
   return format(lastMonthDate, 'MMMM yyyy');
 };
 
+/**
+ * Checks if a utilisation report has already been submitted by the specified bank for the current reporting period
+ * @param bank {object} - the bank to check
+ * @returns {Promise<boolean>}
+ */
+const getIsReportSubmitted = async (bank) => {
+  // TODO FN-1164 - check how this endpoint has been updated
```
toddym42, line 68: Still need to check this.

toddym42: @ChristianMcc - I think after your recent changes I can still use this function to find all existing reports (I think this was previously point at a `previous-reports` endpoint but now points to the new `/v1/bank/${bankId}/utilisation-reports` one). Let me know if this isn't right.

portal-api/src/scheduler/jobs/helpers/utilisation-report-helpers.js
```diff
@@ -22,6 +29,17 @@ const getFormattedReportDueDate = async () => {
   return format(reportDueDate, 'd MMMM yyyy');
 };
 
+/**
+ * Returns the utilisation report chaser date for the current month - i.e. the date that a follow-up email should be
+ * sent to the bank to chase a report if not received by the due date
+ * @returns {Promise<Date>}
+ */
+const getReportOverdueChaserDate = async () => {
+  const bankHolidays = await externalApi.bankHolidays.getBankHolidayDatesForRegion('england-and-wales');
```
ChristianMcc, line 38: should 'england-and-wales' be a constant somewhere so it's easy to update if the format of the JSON ever changes? (I know this is after having done this multiple times and i've only just caught it)

portal-api/src/scheduler/jobs/helpers/utilisation-report-helpers.js
ChristianMcc: yep this is right, our method of abstraction to the api file actually working out nicely here for once! However you can now query my month and year so it's probably worth doing that query and then checking that exists rather than then checking the response has something that matches the right year/month combo at the end of this function

ChristianMcc, line 68: (so can remove this)

portal-api/src/scheduler/jobs/helpers/utilisation-report-helpers.js
```diff
@@ -41,6 +59,24 @@ const getFormattedReportPeriod = () => {
...
+/**
+ * Checks if a utilisation report has already been submitted by the specified bank for the current reporting period
+ * @param bank {object} - the bank to check
+ * @returns {Promise<boolean>}
+ */
+const getIsReportSubmitted = async (bank) => {
+  // TODO FN-1164 - check how this endpoint has been updated
+  const reportsResponse = await api.getUtilisationReports(bank.id);
+
+  if (reportsResponse?.status !== 200) {
```
ChristianMcc, line 71: I refactored the endpoint to just return data/throw though so this will need to be refactored to account for this

portal-api/src/scheduler/jobs/index.js
```diff
@@ -1,7 +1,8 @@
 const sendReportDueEmailsJob = require('./send-report-due-emails-job');
```
ChristianMcc, line 1: It's a nit pick and I think we're fine to change but just on naming here:
ReportDue is kind of true for all of them, I think this would be sendReportSubmissionPeriodEndEmailsJob or something instead.
Feel free to just resolve this as I think the rest of this section of code explains all the functionality well

portal-api/src/scheduler/jobs/send-report-due-emails-job/index.js
```diff
@@ -1,101 +1,35 @@
...
-const { hasValue, isValidEmail } = require('../../../utils/string');
+const { getReportDueDate, getFormattedReportPeriod, sendEmailToAllBanksWhereReportNotReceived } = require('../helpers/utilisation-report-helpers');
 const sendEmail = require('../../../external-api/send-email');
 const EMAIL_TEMPLATE_IDS = require('../../../constants/email-template-ids');
 
 const { UTILISATION_REPORT_DUE_EMAIL_SCHEDULE } = process.env;
 
-const EMAIL_NAME = 'GEF utilisation report due';
-
-/**
```
ChristianMcc, line 12: nice commonisation into the helpers :)

ChristianMcc (changes requested):
Nice work, good commonisation :) just a few comments mostly on the new function you mentioned

JoshBinns2000 (commented):
It looks pretty good to me :)

.env.sample
```diff
@@ -38,8 +38,10 @@ SESSION_SECRET=
 
 # UTILISATION REPORT
```
toddym42, line 39: To confirm if the time of sending these emails will change from 7am to 8am.

.env.sample
```diff
@@ -38,8 +38,10 @@ SESSION_SECRET=
...
 UTILISATION_REPORT_DUE_DATE_BUSINESS_DAYS_FROM_START_OF_MONTH=10
-UTILISATION_REPORT_REPORTING_PERIOD_START_EMAIL_SCHEDULE='0 7 1 * *'
+UTILISATION_REPORT_OVERDUE_CHASER_DATE_BUSINESS_DAYS_FROM_START_OF_MONTH=20
```
toddym42, line 41: To confirm if the overdue chaser date will change from 20 business days to 15.

portal-api/src/scheduler/jobs/helpers/utilisation-report-helpers.js
```diff
@@ -57,10 +93,80 @@ const getEmailRecipient = (paymentOfficerTeam, bankName) => {
   return paymentOfficerTeam.teamName;
 };
 
+/**
+ * Sends the email to the specified bank when a valid payment officer team email is present
+ * @param emailDescription {string} - a description of the email (used for logging)
+ * @param sendEmailCallback {SendEmailCallback} - callback function that sends the required email.
```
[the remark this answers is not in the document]
toddym42: The type `SendEmailCallback` is declared at the top of the file. i.e.
```
/**
 * @typedef {({ emailAddress: string, recipient: string }) => Promise<void>} SendEmailCallback
 */
```
I could have done this inline here, but the type def was quite long, and used in two places (see function below) so made sense to abstract it.

I could have just used `Function` but that is very generic and doesn't tell you anything about the function (importantly that there are two parameters, `emailAddress` and `recipient`, provided to it).

portal-api/src/scheduler/jobs/send-report-overdue-emails-job/index.test.js
```diff
@@ -0,0 +1,88 @@
...
+describe('sendReportOverdueEmailsJob', () => {
+  afterEach(() => {
+    process.env = { ...originalProcessEnv };
+    jest.resetAllMocks();
+    jest.useRealTimers();
+  });
+
+  it('does not send emails when report chaser is not due today', async () => {
+    // Arrange
+    process.env.UTILISATION_REPORT_OVERDUE_CHASER_DATE_BUSINESS_DAYS_FROM_START_OF_MONTH = 15;
```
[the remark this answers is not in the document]
toddym42: It would have effects on other tests, but I have the following the in the `afterEach` function that resets it after each test:
```
process.env = { ...originalProcessEnv };
```

portal-api/src/scheduler/jobs/helpers/utilisation-report-helpers.js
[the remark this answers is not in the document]
JoshBinns2000: Ah nice - I hadn't clocked the definition at the top

portal-api/src/scheduler/jobs/send-report-overdue-emails-job/index.test.js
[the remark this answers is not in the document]
JoshBinns2000: Ah very nice :)

portal-api/src/scheduler/jobs/send-report-due-emails-job/index.test.js
```diff
@@ -2,10 +2,9 @@ const { subDays } = require('date-fns');
 const { produce } = require('immer');
 const sendReportDueEmailsJob = require('.');
```
francescastocco, line 3: I haven't come across `require('.')` before, how does it work?

francescastocco (commented):
Nice work on commonising these emails!

portal-api/src/scheduler/jobs/send-report-due-emails-job/index.test.js
toddym42: This is because the file I'm referring to is an `index.js` file. Similarly to how I can just refer to the directory name when importing an `index.js` file from another directory, I can just use `'.'` when referring to an `index.js` file in the same directory. This is the same as `'./index'`.

PR #2317, opened 2023-11-14
feat(FN-1134): upload correct months report

JoshBinns2000:
# Introduction

We want to make sure the .xlsx file that the payment report officer is trying to upload contains the current report period in the filename (ie. if the report period is October 2023, the filename should contain either `Oct_2023` or `October_2023`). We want different behaviour for when the filename does not contain a report period versus when it does contain a report period, but that report period does not match the current due report period.

# Resolution

Add a function to validate whether or not a filename contains:
- any report period at all
- the correct report period

Also updates the existing e2e tests for being able to upload a report and updated the names of the existing fixture files in `e2e-tests/portal`. This is because the filename validation comes before the csv extraction errors, so those files need to have a valid filename but invalid body. The files labelled as `valid-*` have a valid body, but may not have a valid name. The files labelled as `invalid-*` should have a valid name, but invalid body.


portal/server/controllers/utilisation-report-service/utilisation-report-upload/index.js
```diff
@@ -14,7 +14,7 @@ const getUtilisationReportUpload = async (req, res) => {
       const reportPeriod = format(new Date(year, month - 1), 'MMMM yyyy');
       return { ...dueReport, reportPeriod };
     });
-    const { reportPeriod, month, year } = dueReportsWithDetails.at(-1);
+    const { reportPeriod, month, year } = dueReportsWithDetails[0];
     req.session.utilisationReport = { reportPeriod, month, year };
     const nextDueReportDueDate = await getReportDueDate(userToken, new Date(year, month - 1));
```
JoshBinns2000, line 19: Note to self: this is technically working, but is kind of incorrect

JoshBinns2000: This "bug" gets resolved as part of 1089, but I'm going to bring some of that functionality into this ticket

portal/server/controllers/utilisation-report-service/utilisation-report-upload/utilisation-report-validator.js
```diff
@@ -82,8 +82,45 @@ const validateCsvData = (csvData) => {
...
+
+  const { containsReportPeriod, monthInFilename } = filenameContainsReportingPeriod(filename);
+  if (!containsReportPeriod) {
+    const filenameError = `The selected file must contain the reporting period as part of its name, for example '${expectedFilenameReportPeriod}'`;
+    return { filenameError };
+  }
+
+  const currentReportPeriodYear = currentReportPeriod.split(' ').at(-1); // Expected format 'MMMM yyyy'
+  const filenameReportPeriod = `${monthInFilename}_${currentReportPeriodYear}`;
+  const filenameReportPeriodMatches = filenameReportPeriod === expectedFilenameReportPeriod;
```
toddym42, line 110: I don't think this validates that the filename contains the correct year, as we're just using the expected year in `filenameReportPeriod`?

If we can ignore the need to check for month short names I think this logic could be a lot simpler. e.g.
```
const validateFilenameContainsReportPeriod = (filename, currentReportPeriod) => {
  const expectedFilenameReportPeriod = currentReportPeriod.replace(' ', '_');

  const filenameContainsReportPeriod = filename.toLowerCase().includes(expectedFilenameReportPeriod.toLowerCase());

  if (!filenameContainsReportPeriod) {
    const filenameError = `The selected file must contain the current reporting period as part of its name - i.e. '${expectedFilenameReportPeriod}'`;
    return { filenameError };
  }

  return {};
};
```
If we do need to also check the short name we should be able to extract the long name from the `currentReportPeriod`, map to the short name, and check that too.

e2e-tests/portal/cypress/e2e/journeys/payment-report-officer/utilisation-report-upload/can-upload-utilisation-report.spec.js
```diff
@@ -14,18 +20,40 @@ context('Utilisation report upload', () => {
...
 
+    it('should display an error if the file selected does not contain the current report period', () => {
+      cy.login(BANK1_PAYMENT_REPORT_OFFICER1);
+      cy.visit(relativeURL('/utilisation-report-upload'));
+
+      utilisationReportUpload.utilisationReportFileInput().attachFile('valid-utilisation-report-September_2023.xlsx');
+      utilisationReportUpload.continueButton().click();
+
+      utilisationReportUpload.utilisationReportFileInputErrorMessage().should('have.length', 1);
+      utilisationReportUpload.utilisationReportFileInputErrorMessage().contains('The selected file must be the February 2023 report');
```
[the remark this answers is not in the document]
JoshBinns2000: Unfortunately it needs to be in the `beforeEach` as many of the tests need to upload a valid file, and a valid file now needs to include the correct reporting period in its name. As these files are taken from fixtures and I want them to remain static, I use the `beforeEach` to enforce in each test that the due report period is February 2023. There is the added issue that some tests successfully upload a file, whilst some don't, so it is easier to fix the state of the submitted reports before each test runs

portal/server/controllers/utilisation-report-service/utilisation-report-upload/index.js
```diff
@@ -14,14 +14,22 @@ const getUtilisationReportUpload = async (req, res) => {
...
+    // TODO: FN-1089 will make this section better
+    if (dueReportsWithDetails.length === 1) {
+      const nextDueReportDueDate = await getReportDueDate(userToken, new Date(year, month - 1));
+      return res.render('utilisation-report-service/utilisation-report-upload/utilisation-report-upload.njk', {
+        user,
+        primaryNav: 'utilisation_report_upload',
+        dueReportsWithDetails,
+        nextDueReportDueDate,
+      });
+    }
```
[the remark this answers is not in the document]
JoshBinns2000: I've done a refactor of this section - please take another look

portal/server/controllers/utilisation-report-service/utilisation-report-upload/utilisation-report-validator.js
JoshBinns2000: I have refactored this quite significantly so it actually checks the format we expect now - let me know what you

portal/server/controllers/utilisation-report-service/utilisation-report-upload/index.js
```diff
@@ -77,14 +85,23 @@ const renderPageWithError = (req, res, errorSummary, validationError) => {
...
       return renderPageWithError(req, res, uploadErrorSummary, uploadValidationError);
     }
 
     // File is valid so we can start processing and validating its data
+    const { reportPeriod } = utilisationReport;
+    const filename = req.file.originalname;
+    const { filenameError } = validateFilenameContainsReportPeriod(filename, reportPeriod);
+    if (filenameError) {
+      const extractDataErrorSummary = [{ text: filenameError, href: '#utilisation-report-file-upload' }];
+      const extractDataError = { text: filenameError };
```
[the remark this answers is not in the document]
JoshBinns2000: Discussed: filename error logic has been added to `getUploadErrors`, `renderPageWithError` needs to be able to render two separate possible templates and the extraction is being left alone for now (unless we decide otherwise in later discussion)

portal/server/controllers/utilisation-report-service/utilisation-report-upload/utilisation-report-validator.js
```diff
@@ -82,8 +82,60 @@ const validateCsvData = (csvData) => {
...
+  const currentReportPeriodYear = dueReportPeriod.split(' ').at(-1);
+  const longFilenameReportPeriod = `${monthInFilename.longName}_${currentReportPeriodYear}`;
+  const shortFilenameReportPeriod = `${monthInFilename.shortName}_${currentReportPeriodYear}`;
+  const lowerCaseFilenameWithUnderscores = filename.toLowerCase().replaceAll('-', '_');
+  if (lowerCaseFilenameWithUnderscores.includes(longFilenameReportPeriod)) {
+    return {};
+  }
+  if (lowerCaseFilenameWithUnderscores.includes(shortFilenameReportPeriod)) {
+    return {};
+  }
```
toddym42, line 129: As discussed, I think we could do this with a regex.

portal/server/controllers/utilisation-report-service/utilisation-report-upload/utilisation-report-validator.js
```diff
@@ -82,8 +82,60 @@ const validateCsvData = (csvData) => {
...
+
+  const filenameError = `The selected file must be the ${dueReportPeriod} report`;
```
toddym42, line 131: We should make clear here that the filename is the issue, otherwise the user may be confused why it's failed (e.g. e.g they had the correct month but no year in the filename.

portal/server/controllers/utilisation-report-service/utilisation-report-upload/index.js
```diff
@@ -1,62 +1,98 @@
...
 const { extractCsvData, removeCellAddressesFromArray } = require('../../../utils/csv-utils');
-const { validateCsvData } = require('./utilisation-report-validator');
+const { validateCsvData, validateFilenameContainsReportPeriod } = require('./utilisation-report-validator');
 const { getReportDueDate } = require('./utilisation-report-status');
 const api = require('../../../api');
 
+/**
+ * Returns an array of due report dates including the one-indexed month,
+ * the year and the report period with format 'MMMM yyyy'
+ * @param {Object} userToken - Token to validate session
```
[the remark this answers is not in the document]
JoshBinns2000: Francesca listed it as a string in the JSDocs of `utilisation-report-status.js` and there is only one place where it is described as a string, so I went with object

portal/server/controllers/utilisation-report-service/utilisation-report-upload/utilisation-report-validator.js
JoshBinns2000: This message was taken from the designs rather than by my decision. Not to mention that it will make it more complex if I am checking whether or not the year or month individually are incorrect

portal/server/controllers/utilisation-report-service/utilisation-report-upload/index.js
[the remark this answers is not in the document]
JoshBinns2000: They are supposed to be `string`

portal/server/controllers/utilisation-report-service/utilisation-report-upload/utilisation-report-validator.js
JoshBinns2000: Done

portal/server/constants/monthNames.js
```diff
@@ -0,0 +1,52 @@
+const LOWER_CASE_MONTH_NAMES = {
```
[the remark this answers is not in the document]
JoshBinns2000: Yeh I wanted to do this originally, and changed it to help my old implementation. The new regex approach means I can do this easily

toddym42 (approved):
Looks good 👍

toddym42 (approved):

PR #2319, opened 2023-11-15
feat(FN-1507): update title on upload report page

ChristianMcc:
## Introduction
We need to update the title of the upload report page

## Resolution
Update the title of the upload report page

### DIFF
--- a/portal-api/api-tests/api.js
+++ b/portal-api/api-tests/api.js
@@ -24,7 +24,11 @@ const postMultipartForm = (app, { url, data, files, headers }) => {
   }
 
   Object.entries(data).forEach(([fieldname, value]) => {
-    requestInProgress.field(fieldname, value);
+    if (typeof value === 'object') {
+      requestInProgress.field(fieldname, JSON.stringify(value));
+    } else {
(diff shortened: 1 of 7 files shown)


portal-api/api-tests/api.js
```diff
@@ -24,7 +24,11 @@ const postMultipartForm = (app, { url, data, files, headers }) => {
   }
 
   Object.entries(data).forEach(([fieldname, value]) => {
-    requestInProgress.field(fieldname, value);
+    if (typeof value === 'object') {
+      requestInProgress.field(fieldname, JSON.stringify(value));
+    } else {
+      requestInProgress.field(fieldname, value);
+    }
```
JoshBinns2000, line 31: Rather than branching could we possibly do something like:
```
const valueAsJson = typeof value === 'object' ? JSON.stringify(value) : value;
requestInProgress.field(fieldname, valueAsJson);
```

JoshBinns2000 (commented):
Looks pretty good to me! Only got the one comment about using a ternary because I think that looks a bit cleaner

JoshBinns2000 (approved):

PR #2327, opened 2023-11-17
fix: fix mock-data-loader clearing gef-eligibilityCriteria

lmarrai-sw:
## Introduction

I noticed that the mock data loader does not clear the `gef-eligibilityCriteria` collection when it runs.

## Resolution

I corrected the API call that mock-data-loader uses to delete GEF eligibility criteria so that it does clear the collection when it runs


ChristianMcc (approved):
LGTM :)

AlexBramhill (approved):

PR #2328, opened 2023-11-17
feat(FN-967): upload report to azure file storage

francescastocco:
## Introduction
Once files are uploaded in portal they need to be stored as csv's in Azure File Storage

## Resolution
- Use existing `uploadFile` in portal-api fileshare to upload csv to Azure
- Return file info and store in mongo db


dtfs-central-api/src/services/repositories/utilisation-reports-repo.js
```diff
@@ -2,16 +2,20 @@ const sortBy = require('lodash/sortBy');
 const db = require('../../drivers/db-client');
 const { DB_COLLECTIONS } = require('../../constants/dbCollections');
 
+/**
+ * @typedef {folder: string, filename: string, fullPath: string, url: string, contentType: string} AzureFileInfo
```
ChristianMcc, line 6: I noticed in some tests our mock data doesn't have contentType and we don't care much about it, should we put it as an optional here? I'm not familiar with JSDoc types but assuming that `string | null` or `contentType?` is something we would do

dtfs-central-api/src/v1/validation/utilisation-report-service/utilisation-report-validator.js
```diff
@@ -42,20 +42,41 @@ const validateYear = (year) => {
...
-  if (!filePath) {
-    return 'File path is required';
+const validateFileInfo = (fileInfo) => {
+  if (!fileInfo) {
+    return 'File info is required';
   }
-  if (filePath) {
-    if (typeof filePath !== 'string') {
-      return 'File path must be a string';
+  if (fileInfo) {
```
ChristianMcc, line 53: we don't need this as we have an early return on line 51

dtfs-central-api/src/v1/validation/utilisation-report-service/utilisation-report-validator.test.js
```diff
@@ -43,23 +43,34 @@ describe('utilisation-report-validator', () => {
...
 
-      expect(validationError).toEqual('File path is required');
+      expect(validationError).toEqual(['File info is required']);
     });
 
-    it('returns an error when an incorrect file path is provided', async () => {
-      const validationError = validateFilePath(14);
-
-      expect(validationError).toEqual('File path must be a string');
+    it('returns an array of errors if the file info has any errors', async () => {
```
ChristianMcc, line 64: If I'm nit-picking I'd probably want this as well as separate tests for each condition 👀

portal-api/src/drivers/fileshare.js
```diff
@@ -1,20 +1,31 @@
...
+const getConfig = (fileshare = FILESHARES.PORTAL) => {
+  let config;
+  switch (fileshare) {
+    case FILESHARES.WORKFLOW:
+      config = AZURE_WORKFLOW_FILESHARE_CONFIG;
+      break;
+    case FILESHARES.UTILISATION_REPORTS:
+      config = AZURE_UTILISATION_REPORTS_FILESHARE_CONFIG;
+      break;
+    default:
```
ChristianMcc, line 22: I'd probably add `case FILESHARES.PORTAL` here as well and let it ffall through to the default blow just to be clear it exists as an option

portal-api/src/v1/controllers/utilisation-report-service/utilisation-report-upload.controller.js
```diff
@@ -62,19 +64,54 @@ const sendEmailToBankPaymentOfficerTeam = async (reportPeriod, bankId, submitted
...
-    // const file = req.file;
-
-    // if (!file) return res.status(400).send();
-    // const path = await saveFileToAzure(req.file, month, year, bank);
+    if (!file) return res.status(400).send();
+    const { fileInfo, error } = await saveFileToAzure(file, parsedUser.bank.id);
 
-    const saveDataResponse = await api.saveUtilisationReport(parsedReportData, month, year, parsedUser, 'a file path');
+    if (error) {
+      const status = 500;
```
ChristianMcc, line 107: could probably inline this and change it to just send back the string message. I know we do this elsewhere of status and data but axios already wraps a return message in status and data so don't see the need for it. Not sure if I'm missing something here though

portal-api/src/v1/controllers/utilisation-report-service/utilisation-report-upload.controller.js
```diff
@@ -62,19 +64,54 @@ const sendEmailToBankPaymentOfficerTeam = async (reportPeriod, bankId, submitted
...
+      return res.status(status).send({ status, data: 'Failed to save utilisation report to Azure' });
+    }
+    const azureFileStorage = {
+      ...fileInfo,
+      mimetype: file.mimetype,
```
ChristianMcc, line 112: what do we need this for?

portal-api/src/v1/controllers/utilisation-report-service/utilisation-report-upload.controller.test.js
```diff
@@ -0,0 +1,45 @@
+const { when } = require('jest-when');
```
ChristianMcc, line 1: I've recently fixed the api tests for this function, which now do a happy path through the utilisation-report-upload.controller. I'm not sure what we do with the other uploading to azure type functions and if the api-tests just upload to azure but we should try copy that so they work. 
Grab me offline if you want to chat about this

portal-api/unit.jest.config.js
```diff
@@ -1,5 +1,5 @@
 module.exports = {
   collectCoverageFrom: ['src/**/*.{js,}'],
   coverageDirectory: 'generated_reports/coverage/unit',
-  testMatch: ['**/*.test.js'],
+  testMatch: ['**/utilisation-report-upload.controller.test.js'],
```
ChristianMcc, line 4: can revert this

ChristianMcc (commented):
You haven't posted in pull-requests yet but I've reviewed it anyway as I saw it here :D 

Great work, a few small comments but implementation looks good. Happy to have a look at the infra PR whenever it's ready for the new env vars

portal-api/unit.jest.config.js
francescastocco: oops

portal-api/src/v1/controllers/utilisation-report-service/utilisation-report-upload.controller.js
francescastocco: the content type/mimetype is needed to download a file from azure

dtfs-central-api/src/services/repositories/utilisation-reports-repo.js
francescastocco: I've added it to tests now, as it's needed for downloading the file I don't think it should be nullable :)

dtfs-central-api/src/services/repositories/utilisation-reports-repo.js
```diff
@@ -19,7 +23,7 @@ const saveUtilisationReportDetails = async (month, year, csvFilePath, uploadedBy
     month: Number(month),
     year: Number(year),
     dateUploaded: new Date(),
-    path: csvFilePath,
```
toddym42, line 22: I _think_ this `path` value was what the `previous-reports.njk` template was previously using for the link?

dtfs-central-api/src/v1/controllers/utilisation-report-service/upload-utilisation-report.controller.js
```diff
@@ -4,19 +4,19 @@ const {
...
+const validatePayload = (reportData, month, year, fileInfo, bankId) => {
   const validationErrors = validateUtilisationReportData(reportData);
   const monthValidationError = validateMonth(month);
   if (monthValidationError) validationErrors.push(monthValidationError);
   const yearValidationError = validateYear(year);
   if (yearValidationError) validationErrors.push(yearValidationError);
-  const filePathValidationError = validateFilePath(filePath);
-  if (filePathValidationError) validationErrors.push(filePathValidationError);
+  const fileInfoValidationError = validateFileInfo(fileInfo);
+  if (fileInfoValidationError) validationErrors.push(...fileInfoValidationError);
```
toddym42, line 19: I think this should be `if (fileInfoValidationErrors.length)` as an empty array is truthy.

dtfs-central-api/src/v1/validation/utilisation-report-service/utilisation-report-validator.js
```diff
@@ -42,20 +42,45 @@ const validateYear = (year) => {
 };
 
 /**
- * Validates the file path of the utilisation report in azure. Returns null if valid, otherwise returns an error message.
- * @param {unknown} filePath - file path of the utilisation report in azure.
- * @returns {String | null} - Error message or null if valid.
+ * Validates the details of the file storage for the utilisation report in azure. Returns null if valid, otherwise returns an error message.
```
toddym42, line 45: The last sentence should be something like
```
Returns an array of errors, with an empty array if valid.
```

portal-api/api-tests/v1/utilisation-report-service/utilisation-report-upload.api-test.js
```diff
@@ -8,6 +8,21 @@ const { DB_COLLECTIONS } = require('../../fixtures/constants');
...
 const { as, postMultipartForm } = require('../../api')(app);
 const { PAYMENT_REPORT_OFFICER } = require('../../../src/v1/roles/roles');
+const { uploadFile } = require('../../../src/drivers/fileshare');
+
+const mockFileUpload = {
+  folder: 'folder_name',
+  filename: 'test_file.csv',
+  fullPath: 'folder_name/test_file.csv',
+  url: 'https://azure/utilisation-reports/folder_name/test_file.csv',
+};
```
toddym42, line 18: We create a mock value for this `AzureFileInfo` type in quite a few places. Are we able to create a `MOCK_AZURE_FILE_INFO` (or similar) in `portal-api/test-helpers`?

This would also make things much easier to update if the type ever changed as we'd only need to update one place to get all out tests using the updated type. We should also add this to `MOCK_UTILISATION_REPORT` as that should now have the extra `azureFileStorage` too.

portal-api/package.json
```diff
@@ -25,7 +25,7 @@
     "lint:fix": "eslint ./src --fix",
     "start": "node src/index.js",
     "start:dev": "npx nodemon src/index.js",
-    "unit-test": "jest --coverage --verbose --config=unit.jest.config.js"
+    "unit-test": "jest --verbose --config=unit.jest.config.js"
```
toddym42, line 28: Guessing this was a temporary change for local testing? I see `dtfs-central-api` has a `"unit-test-quick"` command, so maybe we could add that here too so we don't need to manually amend this when testing locally.

portal-api/src/drivers/fileshare.js
toddym42: Do we ever expect this default case to be used? If not then maybe we should throw an error instead as this could lead to unexpected results.

portal-api/src/drivers/fileshare.js
```diff
@@ -100,7 +114,7 @@ const uploadFile = async ({ fileshare, folder, filename, buffer, allowOverwrite
 
   const fileClient = await directoryClient.getFileClient(`${filename}`);
 
-  const existingFileProps = await fileClient.getProperties().catch(() => {});
+  const existingFileProps = await fileClient.getProperties().catch(() => { });
```
toddym42, line 117: When we get round to running prettier on everything I think it would remove the added space here and below.

portal-api/src/scheduler/types/scheduler-job.d.ts
```diff
@@ -9,7 +9,7 @@ export type SchedulerJob = {
      */
     message: string;
     /**
-     * The task to be run by the job
+     * The task to be run by the jobs
```
toddym42, line 12: This comment is in reference to this type, being a single job, so I don't think this should be plural?

portal-api/src/v1/controllers/utilisation-report-service/utilisation-report-upload.controller.js
```diff
@@ -62,35 +64,74 @@ const sendEmailToBankPaymentOfficerTeam = async (reportPeriod, bankId, submitted
...
+      filename: formatFilenameForSharepoint(originalname),
+      buffer,
+      allowOverwrite: true,
+    });
+
+    console.info(`Successfully saved utilisation report for bank: ${bankId}`);
+    return { fileInfo, error: false };
+  } catch (error) {
+    console.error('Failed to save utilisation report: %O', error);
+    return { fileInfo: null, error: true };
```
toddym42, line 91: I think we should just throw here and let the error bubble up to the initial controller function where `res.status` is set. This prevents the need to remember to check the response for success. We then know that if the call to `saveFileToAzure` is successful the the response will be the populated `fileInfo`. So this function would then be
```
 /**
 * Saves file to Azure in utilisation-reports ShareClient, returns the file storage info
 * @param {object} file
 * @param {string} bankId - bank id as a string
 * @returns {Promise<object>} - azure storage details with folder & file name, full path & url.
 */
const saveFileToAzure = async (file, bankId) => {
  try {
    console.info(`Attempting to save utilisation report for bank: ${bankId}`);
    const { originalname, buffer } = file;

    const fileInfo = await uploadFile({
      fileshare: FILESHARES.UTILISATION_REPORTS,
      folder: bankId,
      filename: formatFilenameForSharepoint(originalname),
      buffer,
      allowOverwrite: true,
    });

    console.info(`Successfully saved utilisation report for bank: ${bankId}`);
    return fileInfo;
  } catch (error) {
    console.error('Failed to save utilisation report: %O', error);
    throw error;
  }
};
```
and when calling in `uploadReportAndSendNotification`, instead of
```
const { fileInfo, error } = await saveFileToAzure(file, parsedUser.bank.id);
if (error) {
  const status = 500;
  return res.status(status).send({ status, data: 'Failed to save utilisation report to Azure' });
}
```
we can just do
```
const fileInfo = await saveFileToAzure(file, parsedUser.bank.id);
```
we don't need to replace the `if (error) {` with a try/catch as the `uploadReportAndSendNotification` already has its own and handles the error with an appropriate response.

portal-api/src/v1/controllers/utilisation-report-service/utilisation-report-upload.controller.js
```diff
@@ -62,35 +64,74 @@ const sendEmailToBankPaymentOfficerTeam = async (reportPeriod, bankId, submitted
...
+
     const { reportPeriod, reportData, month, year, user } = req.body;
     const parsedReportData = JSON.parse(reportData);
     const parsedUser = JSON.parse(user);
 
-    // TODO: FN-967 save file to azure
-    // const file = req.file;
-
-    // if (!file) return res.status(400).send();
+    if (!file) return res.status(400).send();
```
toddym42, line 103: nit: As a rule of thumb I almost never use an `if` block without using `{}` to scope the change. This helps prevent unexpected errors as the `{}` make absolutely clear what is being executed when the `if` predicate is true.
```
if (!file) {
}
```

portal-api/src/v1/controllers/utilisation-report-service/utilisation-report-upload.controller.js
```diff
@@ -62,35 +64,74 @@ const sendEmailToBankPaymentOfficerTeam = async (reportPeriod, bankId, submitted
...
       console.error('Failed to save utilisation report: %O', saveDataResponse);
       return res.status(status).send({ status, data: 'Failed to save utilisation report' });
     }
-    const submittedDateUtc = new Date().toISOString();
     await sendEmailToPdcInputtersEmail(parsedUser?.bank?.name, reportPeriod);
-    const { paymentOfficerEmail } = await sendEmailToBankPaymentOfficerTeam(reportPeriod, parsedUser?.bank?.id, submittedDateUtc, parsedUser);
+    const { paymentOfficerEmail } = await sendEmailToBankPaymentOfficerTeam(
+      reportPeriod,
+      parsedUser?.bank?.id,
+      new Date(saveDataResponse.data.dateUploaded),
```
toddym42, line 132: I think the `sendEmailToBankPaymentOfficerTeam` is expecting a string here, though it might make more sense to pass the date around instead (as you've done) until we need it formatted. If we keep this change we should also update the type of the function and update where the value is used to use a Date instead of a string.

portal-api/src/v1/middleware/utilisation-report-file-upload.js
```diff
@@ -0,0 +1,23 @@
+const multer = require('multer');
+const { FILE_UPLOAD } = require('../../constants');
+
+const fileFilter = (req, file, cb) => {
+  const fileExtension = file.originalname.match(/\.[^.]*$/g);
```
toddym42, line 5: Node has a built in function to allow us to get the file extension, so this can be
```
const path = require('path');
...
const fileExtension = path.extname(file.originalname);
```

portal-api/src/v1/middleware/utilisation-report-file-upload.js
```diff
@@ -0,0 +1,23 @@
...
+      message: 'file type is not allowed',
+    };
+
+    req.filesNotAllowed = req.filesNotAllowed ? req.filesNotAllowed.concat(fileError) : [fileError];
+  }
+
+  cb(null, allowed);
+};
+
+const utilisationReportFileUpload = multer({ fileFilter, limits: { fileSize: FILE_UPLOAD.MAX_FILE_SIZE } }).single('csvFile');
```
toddym42, line 21: Would be good to add some tests for this middleware.

toddym42 (commented):
Nice work. Mostly just some comments on code style.

dtfs-central-api/src/services/repositories/utilisation-reports-repo.js
```diff
@@ -2,15 +2,19 @@ const sortBy = require('lodash/sortBy');
...
+ */
+
 /**
  * Saves the utilisation report details but not data to the database.
  * @param {number} month - Month of utilisation report, integer between 1 and 12.
  * @param {number} year - Year of utilisation report, integer greater than 2020.
- * @param {String} csvFilePath - Path to the csv file.
- * @param {Object} uploadedByUser - Object representing the user who uploaded the report.
- * @returns {Object} - Object containing reportId and dateUploaded.
+ * @param {AzureFileInfo} azureFileStorage - Azure storage details for csv file.
```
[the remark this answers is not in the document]
francescastocco: Yep good point, I've renamed azureFileStorage --> azureFileInfo

portal-api/package.json
francescastocco: Yep!

dtfs-central-api/src/services/repositories/utilisation-reports-repo.js
```diff
@@ -2,15 +2,24 @@ const sortBy = require('lodash/sortBy');
 const db = require('../../drivers/db-client');
 const { DB_COLLECTIONS } = require('../../constants/dbCollections');
 
+/**
+ * @typedef {object} AzureFileInfo
```
JoshBinns2000, line 6: At a glance, it looks like `{Object}` is used more frequently than `{object}` in the code base, and I think the preferred choice is `Object` with a capital O

toddym42: That does seem to be the way, but I'd argue that the lower-case `object` is more accurate. I detailed some reasoning for this in a previous review [here](https://github.com/UK-Export-Finance/dtfs2/pull/2314#discussion_r1394187937).

portal-api/src/drivers/fileshare.js
toddym42: Just realised I think we need to return the `userDefinedConfig` if defined, only reverting to env var defined config if not defined.

ChristianMcc (commented):
LGTM! Just a small comment on some of the login tests please

e2e-tests/portal/cypress/e2e/journeys/login/login.spec.js
```diff
@@ -116,63 +116,18 @@ context('Login', () => {
...
     });
-  } else {
-    it('A successful login takes the user to the /dashboard page', () => {
-      cy.enterUsernameAndPassword(BANK1_MAKER1);
-
-      cy.url().should('eq', relative('/dashboard/deals/0'));
-    });
   }
 
-  it('A successful login with checker role takes the user to the /dashboard/deals/0 page', () => {
```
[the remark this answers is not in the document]
francescastocco: Josh is fixing these separately, but it looks like all of these tests are duplicated versions of the tests inside the condition on line 67 `if (Cypress.env('DTFS_FF_MAGIC_LINK')) {`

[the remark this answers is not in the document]
francescastocco: I've removed them for now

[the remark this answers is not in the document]
ChristianMcc: Okay cool sounds good :)

ChristianMcc (approved):

PR #2329, opened 2023-11-17
chore: add typescript parser to portal-api to fix linting error

toddym42:
A `.d.ts` type file was recently added to the portal-api project https://github.com/UK-Export-Finance/dtfs2/blob/main-stbr/portal-api/src/scheduler/types/scheduler-job.d.ts

This change adds a typescript parser to eslint so that it can correctly parse these files. This resolves a linting error where eslint wasn't able to recognise the `type` keyword.


francescastocco (approved):
🚀

PR #2330, opened 2023-11-20
feat(FN-1548): add business days bug fix

JoshBinns2000:
# Introduction

The old `addBusinessDaysWithHolidays` function would not account for if the starting day supplied was a weekend. This would lead to incorrect behaviour if this case was ever encountered. Similarly, the implementation of this function was not performed correctly as 10 days were being added in order to obtain the 10th business day from the start of the month. However, only 9 days should be added. This also caused bugs related to FN-1162, FN-1163 and FN-1164 which this PR fixes.

# Resolution

- Remove the old `addBusinessDaysWithHolidays` function
- Add a new `getBusinessDayOfMonth` function to both `portal` and `portal-api` which returns the Nth business day of the month-year combination and array of holiday dates provided
- Update unit tests in both `portal` and `portal-api`


portal-api/src/scheduler/jobs/helpers/utilisation-report-helpers.js
```diff
@@ -18,7 +18,8 @@ const DEFAULT_PAYMENT_OFFICER_TEAM_NAME = 'Team';
 const getReportDueDate = async () => {
   const bankHolidays = await externalApi.bankHolidays.getBankHolidayDatesForRegion(BANK_HOLIDAY_REGION.ENGLAND_AND_WALES);
   const businessDaysToAdd = process.env.UTILISATION_REPORT_DUE_DATE_BUSINESS_DAYS_FROM_START_OF_MONTH;
-  return addBusinessDaysWithHolidays(startOfMonth(new Date()), businessDaysToAdd, bankHolidays);
+  const businessDayIndex = businessDaysToAdd + 1;
```
ChristianMcc, line 21: I think I'd expect this function to just call something that gives you bank the 10th business day, and not have to worry about adding/minusing anything

portal-api/src/scheduler/jobs/helpers/utilisation-report-helpers.js
```diff
@@ -18,7 +18,8 @@ const DEFAULT_PAYMENT_OFFICER_TEAM_NAME = 'Team';
...
+  return getBusinessDayByIndex(startOfMonth(new Date()), bankHolidays, businessDayIndex);
```
ChristianMcc, line 22: Given the above I'd probably still expect this to be `addBusinessDaysWithHolidays` and for any 1st of month implementation issues to be within the function and not visible to the caller of this function

ChristianMcc (changes requested):
Most of the code looks good but could we refactor for the couple of comments and I'll re-review after that please :)

portal-api/src/scheduler/jobs/helpers/utilisation-report-helpers.js
```diff
@@ -17,8 +17,12 @@ const DEFAULT_PAYMENT_OFFICER_TEAM_NAME = 'Team';
...
 const getReportDueDate = async () => {
   const bankHolidays = await externalApi.bankHolidays.getBankHolidayDatesForRegion(BANK_HOLIDAY_REGION.ENGLAND_AND_WALES);
-  const businessDaysToAdd = process.env.UTILISATION_REPORT_DUE_DATE_BUSINESS_DAYS_FROM_START_OF_MONTH;
-  return addBusinessDaysWithHolidays(startOfMonth(new Date()), businessDaysToAdd, bankHolidays);
+  const businessDay = process.env.UTILISATION_REPORT_DUE_DATE_BUSINESS_DAYS_FROM_START_OF_MONTH;
+
+  const reportDate = new Date();
+  const oneIndexedReportMonth = reportDate.getMonth() + 1;
+  const reportYear = reportDate.getFullYear();
+  return getBusinessDayOfMonth(oneIndexedReportMonth, reportYear, bankHolidays, businessDay);
```
ChristianMcc, line 25: I like this a lot more!

portal/server/helpers/getBusinessDayOfMonth.js
```diff
@@ -0,0 +1,79 @@
...
+ * If the index is one, and the start date is a business
+ * day, it will return the date provided as that is the
+ * first-indexed business day.
+ * @param {number} month - The month (one-indexed)
+ * @param {number} year - The year
+ * @param {Date[]} holidays - A list of dates which should be excluded as holidays
+ * @param {number} businessDay - The one-indexed business day to get
+ * @returns {Date}
+ */
+const getBusinessDayOfMonth = (month, year, holidays, businessDay) => {
```
toddym42, line 58: I think this be a little cleaner if instead of `month` and `year` we just passed in a Date object which represents a date in the relevant month (maybe called something like `dateInMonth`. Then where this is called we don't need to worry about passing the values in separately, and also we don't need to remember that the month needs to be one-indexed. That logic can instead all be taken care of here.

portal-api/src/scheduler/jobs/helpers/utilisation-report-helpers.js
```diff
@@ -17,8 +17,12 @@ const DEFAULT_PAYMENT_OFFICER_TEAM_NAME = 'Team';
...
 const getReportDueDate = async () => {
   const bankHolidays = await externalApi.bankHolidays.getBankHolidayDatesForRegion(BANK_HOLIDAY_REGION.ENGLAND_AND_WALES);
-  const businessDaysToAdd = process.env.UTILISATION_REPORT_DUE_DATE_BUSINESS_DAYS_FROM_START_OF_MONTH;
-  return addBusinessDaysWithHolidays(startOfMonth(new Date()), businessDaysToAdd, bankHolidays);
+  const businessDay = Number.parseInt(process.env.UTILISATION_REPORT_DUE_DATE_BUSINESS_DAYS_FROM_START_OF_MONTH, 10);
+
+  const reportDate = new Date();
+  const oneIndexedReportMonth = reportDate.getMonth() + 1;
+  const reportYear = reportDate.getFullYear();
+  return getBusinessDayOfMonth(oneIndexedReportMonth, reportYear, bankHolidays, businessDay);
```
ChristianMcc, line 25: Just overheard your conversation, very least would be calling this getBusinessDayOfOneIndexedMonth, best case is probably what you just discussed with Mike of passing in only a date!

portal-api/src/utils/date.js
```diff
@@ -1,26 +1,72 @@
...
+const getFirstBusinessDayOfMonth = (dateInMonth, holidays) => {
+  let firstBusinessDay = startOfMonth(dateInMonth);
+  if (isWeekend(firstBusinessDay)) {
+    firstBusinessDay = addOneBusinessDayWithHolidays(firstBusinessDay, holidays);
+  }
+  while (isHoliday(firstBusinessDay, holidays)) {
+    firstBusinessDay = addOneBusinessDayWithHolidays(firstBusinessDay, holidays);
+  }
+  return firstBusinessDay;
+};
```
toddym42, line 43: I don't think we need the while loop if we're using the `addOneBusinessDayWithHolidays` function? Could this just be
```
const getFirstBusinessDayOfMonth = (dateInMonth, holidays) => {
  const firstDayOfMonth = startOfMonth(dateInMonth);

  if (isWeekend(firstDayOfMonth) || isHoliday(firstDayOfMonth, holidays)) {
    return addOneBusinessDayWithHolidays(firstDayOfMonth, holidays);
  }

  return firstDayOfMonth;
};
```
?

portal-api/src/utils/date.js
```diff
@@ -1,26 +1,72 @@
...
+ * businessDay index, it will return the business day with
+ * that index. For example, a businessDay index of 1 will
+ * return the first business day of the month.
+ * @param {Date} dateInMonth - A date in the required month
+ * @param {Date[]} holidays - A list of dates which should be excluded as holidays
+ * @param {number} businessDay - The one-indexed business day to get
+ * @returns {Date}
+ */
+const getBusinessDayOfMonth = (dateInMonth, holidays, businessDay) => {
+  if (typeof businessDay !== 'number' || businessDay < 1) {
```
toddym42, line 56: Could make this even more strict with
```
if (!Number.isInteger(businessDay) || businessDay < 1) {
```

portal-api/src/utils/date.test.js
```diff
@@ -1,106 +1,146 @@
...
+
+      expect(() => getBusinessDayOfMonth(dateInMonth, holidays, businessDay)).toThrow(new Error('Error getting business day: business day must be a positive number'));
+    });
+
+    it('should throw an error when businessDay is not a number', () => {
+      const dateInMonth = new Date('2023-01-01');
+      const holidays = [];
+      const businessDay = 'first';
+
+      expect(() => getBusinessDayOfMonth(dateInMonth, holidays, businessDay)).toThrow(new Error('Error getting business day: business day must be a positive number'));
```
toddym42, line 97: Would prettier format these long lines?

portal/server/helpers/getBusinessDayOfMonth.js
toddym42, line 1: Same comments here as the portal-api version.

portal-api/src/utils/date.test.js
JoshBinns2000: Yes - fixed the same thing in `getBusinessDayOfMonth.test.js`

toddym42 (approved):

PR #2331, opened 2023-11-20
Fix: use mongodb transactions

ChristianMcc:
## Introduction
We weren't using transactions when saving data so if one call saved data could be in a bad state

## Resolution
Use mongodb transactions and add some misc. fixes/logging

### DIFF
--- a/dtfs-central-api/src/drivers/db-client.js
+++ b/dtfs-central-api/src/drivers/db-client.js
@@ -26,7 +26,17 @@ const getConnection = async () => {
   return connection;
 };
 
+const getClient = async () => {
+  if (!connection) {
+    connection = await dbConnect();
+  }
+
+  return client;
+};
+
 module.exports.get = getConnection;
+
+module.exports.getClient = getClient;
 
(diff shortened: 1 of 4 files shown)


dtfs-central-api/src/drivers/db-client.js
```diff
@@ -26,8 +26,18 @@ const getConnection = async () => {
   return connection;
 };
 
+const getClient = async () => {
+  if (!connection) {
+    connection = await dbConnect();
```
toddym42, line 31: Can just do `await dbConnect();` here.

dtfs-central-api/src/drivers/db-client.js
```diff
@@ -26,8 +26,18 @@ const getConnection = async () => {
   return connection;
 };
 
+const getClient = async () => {
```
toddym42, line 29: Maybe move this below `module.exports.get = getConnection;` to the function declarations are grouped together with their `module.export`.

toddym42 (approved):

toddym42 (approved):

PR #2332, opened 2023-11-21
feat(FN-967): add an environment variable for utilisation reports shareclient

francescastocco:
## Introduction
We need an environment variable for the utilisation reports shareclient in Azure File Storage

## Resolution
Added to Bicep
Add to github actions


OxonianCambion (approved):

cspell.json
```diff
@@ -124,7 +124,8 @@
     "wordprocessingml",
     "stbr",
     "ukwest",
-    "tfsproduction"
+    "tfsproduction",
+    "clamav"
```
toddym42, line 128: This list is mostly in alphabetical order except the last few entries. Can we put this in the relevant place?

PR #2333, opened 2023-11-21
feat(FN-976): download utilisation reports

toddym42:
## Ticket

[FN-976](https://ukef-dtfs.atlassian.net/browse/FN-976)

## Introduction

When banks upload utilisation reports they are saved to Azure Storage. There is an existing page (`/previous-reports`) that allows you to view previously uploaded reports, but currently clicking on a report does nothing. This change implements the functionality to download the CSV report when clicked.

## Resolution

* Add `utilisation-report-download` GET routes to `portal` and `portal-api`. This route includes the word `download` to distinguish from fetching utilisation report details we have stored in the `utilisationReport` collection of the MongoDB database.
* The routes also include the `bankId`, which needs to match the user's bank. The `bankId` is used as the folder name in Azure Storage, so if the user requests a file for a bank other than their own then it will not be found.
* The routes also include the MongoDB ID of the report details, which is used to fetch metadata about the file, such as the filename and mimetype.

## To Do

This ticket relies on functionality implemented in [FN-976](https://ukef-dtfs.atlassian.net/browse/FN-967) (uploading a utilisation report). This PR will be rebased again once [the PR for FN-967](https://github.com/UK-Export-Finance/dtfs2/pull/2328) has been merged into `main-stbr`.


dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-report.controller.js
```diff
@@ -0,0 +1,21 @@
...
+
+    if (!utilisationReport) {
+      return res.status(404).send();
+    }
+
+    return res.status(200).send(utilisationReport);
+  } catch (error) {
+    const errorMessage = `Failed to get utilisation report with _id '${_id}'`;
+    console.error(errorMessage, error);
+    return res.status(500).send(errorMessage);
```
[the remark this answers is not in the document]
toddym42: `_id` is validated in middleware.

portal/server/api.js
```diff
@@ -871,17 +871,23 @@
   return response.data;
 };
 
+const downloadUtilisationReport = async (userToken, bankId, _id) => {
+  const response = await axios.get(`${PORTAL_API_URL}/v1/banks/${bankId}/utilisation-report-download/${_id}`, {
+    responseType: 'stream',
+    headers: { Authorization: userToken },
+  });
```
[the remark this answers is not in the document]
toddym42: `bankId` and `_id` values are validated in middleware.

portal-api/src/v1/api.js
```diff
@@ -302,6 +302,23 @@
...
 
+const getUtilisationReportById = async (_id) => {
+  try {
+    if (!isValidMongoId(_id)) {
+      throw new Error(`Invalid MongoDB _id provided: '${_id}'`);
+    }
+
+    const response = await axios.get(`${DTFS_CENTRAL_API_URL}/v1/portal/utilisation-reports/${_id}`, {
+      headers: headers.central,
+    });
```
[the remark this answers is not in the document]
toddym42: `_id` is validated in middleware.

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-report.controller.js
```diff
@@ -0,0 +1,21 @@
...
+    const utilisationReport = await getUtilisationReportDetailsById(_id);
+
+    if (!utilisationReport) {
+      return res.status(404).send();
+    }
+
+    return res.status(200).send(utilisationReport);
+  } catch (error) {
+    const errorMessage = `Failed to get utilisation report with _id '${_id}'`;
+    console.error(errorMessage, error);
```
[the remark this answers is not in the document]
toddym42: `_id` is validated in middleware.

portal-api/src/v1/api.js
```diff
@@ -302,6 +302,23 @@
...
+
+    return response.data;
+  } catch (error) {
+    console.error(`Unable to get utilisation report with MongoDB _id '${_id}'`, error);
```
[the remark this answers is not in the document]
toddym42: `_id` is validated in middleware.

dtfs-central-api/api-tests/mocks/utilisationReports.js
```diff
@@ -0,0 +1,19 @@
...
+
+const MOCK_UTILISATION_REPORT = {
+  bank: {
+    id: MOCK_BANKS.HSBC.id,
+    name: MOCK_BANKS.HSBC.name,
+  },
+  month: 11,
+  year: 2023,
+  dateUploaded: new Date('2023-11-15'),
+  path: 'report.csv',
```
ChristianMcc, line 12: will this need updating between your and francesca's work?

dtfs-central-api/src/services/repositories/utilisation-reports-repo.test.js
```diff
@@ -1,4 +1,4 @@
-const { saveUtilisationReportDetails, getUtilisationReportDetails } = require('./utilisation-reports-repo');
+const { saveUtilisationReportDetails, getUtilisationReportDetailsByBankId } = require('./utilisation-reports-repo');
```
ChristianMcc, line 1: Shall we add a super simple test to this file testing the collection is called with the right ID?

dtfs-central-api/src/v1/routes/portal-routes.js
```diff
@@ -892,4 +893,37 @@ portalRouter.route('/gef/mandatory-criteria/version/:version').get(mandatoryCrit
...
+ *                 - type: object
+ *                   properties:
+ *                     _id:
+ *                       example: 123456abc
+ *       400:
+ *         description: Bad request
+ *       404:
+ *         description: Not found
+ */
+portalRouter.route('utilisation-reports/:_id').get(getUtilisationReportController.getUtilisationReportById);
```
ChristianMcc, line 927: from your other comments I think I was expecting the _id validation to happen here?

portal-api/api-tests/insertUtilisationReportDetails.js
```diff
@@ -4,13 +4,13 @@ const { DB_COLLECTIONS } = require('./fixtures/constants');
 const insertOneUtilisationReportDetails = async (report) => {
   console.info('inserting one utilisation report details');
   const collection = await db.getCollection(DB_COLLECTIONS.UTILISATION_REPORTS);
-  await collection.insertOne(report);
+  return collection.insertOne(report);
```
ChristianMcc, line 7: I have a preference for explicitly awaiting in the low down functions like here so it's clear to the reader what the asynchronous part of the function is. In this case it makes no difference as it's the last line and there aren't any other processes happening after it but if anyone was to add anything it could be weird behaviour if they forget to put an await back in.

Aware in this case it's more of a style nitpick than a behaviour nitpick - if you have any reasons on why you prefer what you've done would be keen to hear them!

portal-api/api-tests/insertUtilisationReportDetails.js
```diff
@@ -4,13 +4,13 @@ const { DB_COLLECTIONS } = require('./fixtures/constants');
...
 };
 
 const insertManyUtilisationReportDetails = async (reports) => {
   console.info('inserting many utilisation report details');
   const collection = await db.getCollection(DB_COLLECTIONS.UTILISATION_REPORTS);
-  await collection.insertMany(reports);
+  return collection.insertMany(reports);
```
ChristianMcc, line 13: ditto

portal-api/src/v1/routes.js
```diff
@@ -251,7 +251,18 @@ authRouter
...
+  );
+
+authRouter
+  .route('/banks/:bankId/utilisation-report-download/:_id')
+  .get(
+    validateUserHasAtLeastOneAllowedRole({ allowedRoles: [PAYMENT_REPORT_OFFICER] }),
+    bankIdValidation,
+    handleValidationResult,
+    validateUserAndBankIdMatch,
+    validateMongoId,
```
ChristianMcc, line 264: we're calling this after the handleValidationResult so we won't handle any validation errors from this here

ChristianMcc (commented):
Leaving some comments, haven't finished reviewing it but hope these are still helpful!

portal/server/routes/middleware/validateMongoId/index.js
```diff
@@ -0,0 +1,20 @@
...
+ */
+const validateMongoId = (req, res, next) => {
+  const { _id } = req.params;
+
+  if (!isValidMongoId(_id)) {
+    return res.redirect('/not-found');
+  }
+
+  return next();
+};
```
JoshBinns2000, line 18: This is the middleware that validates `_id` @ChristianMcc

portal/server/routes/utilisation-report-service/utilisation-report-download/index.js
```diff
@@ -0,0 +1,16 @@
...
+const { PAYMENT_REPORT_OFFICER } = require('../../../constants/roles');
+const { getReportDownload } = require('../../../controllers/utilisation-report-service');
+
+const router = express.Router();
+
+router.get(
+  'banks/:bankId/utilisation-report-download/:_id',
+  [validateToken, validateRole({ role: [PAYMENT_REPORT_OFFICER] })],
+  validateBankIdForUser,
+  validateMongoId,
```
JoshBinns2000, line 12: And this is where the middleware is implemented @ChristianMcc

portal/templates/utilisation-report-service/previous-reports/previous-reports.njk
```diff
@@ -28,10 +28,15 @@
                 {% else %}
                     <p class="govuk-body" data-cy="paragraph">Reports are downloaded as CSV files.</p>
                     <ul class="govuk-list">
-                        {% for item in reports %}
+                        {% for report in reports %}
```
JoshBinns2000, line 31: Why this change?

portal/server/routes/middleware/validateMongoId/index.js
```diff
@@ -0,0 +1,20 @@
+const { isValidMongoId } = require('../../../validation/validate-ids');
+
+/**
+ * Redirects the user to '/not-found' if an invalid MongoDB '_id' path param is
+ * provided
+ * @param {object} req - The request object
+ * @param {object} res - The response object
+ * @param {Function} next - The next middleware function
```
JoshBinns2000, line 8: I've started using different types for these as of late just because we know this is using express:
```
* @ param {import('express').Request} req - Express request object
* @ param {import('express').Response} res - Express response object
* @ param {import('express').NextFunction} req - Express next function
```

JoshBinns2000: There are a few more places where the JSDocs have been written like this in this PR, so if you agree with me then don't forget to change it in those places too :)

portal/server/routes/middleware/validateMongoId/index.js
```diff
@@ -0,0 +1,20 @@
...
+ */
+const validateMongoId = (req, res, next) => {
+  const { _id } = req.params;
+
+  if (!isValidMongoId(_id)) {
+    return res.redirect('/not-found');
+  }
```
JoshBinns2000, line 15: I don't know if this is in the scope of this ticket, but are we supposed to redirect to 'not-found'? Won't this render a page that says 'Page not found' but it's actually an error of the file not existing rather than the page not being found? I'd be expecting to see an error on the page saying "File does not exist" or something like that, but I'm not fully sure what the intended behaviour is

portal/server/routes/middleware/validateBankIdForUser/index.js
```diff
@@ -0,0 +1,19 @@
...
+ * @param {object} res - The response object
+ * @param {Function} next - The next middleware function
+ */
+const validateBankIdForUser = (req, res, next) => {
+  const { user } = req;
+  const { bankId } = req.params;
+
+  if (user.bank.id !== bankId) {
+    return res.redirect('/not-found');
+  }
```
JoshBinns2000, line 14: Same comment as below for this redirect

JoshBinns2000 (commented):
Left a few comments. I put two in to point out where the validation of the `_id` mongo parameter happens for @ChristianMcc, and a few others about JSDocs and the middleware redirect behaviour

francescastocco (commented):
Looks good :)

portal/server/routes/middleware/validateMongoId/index.js
toddym42: Agreed, this is much better. We'll need to add `@types/express` to the project though for the IDE to actually recognise these types and give us some type safety and auto-complete. Will add as part of this PR 👍

portal/server/routes/middleware/validateMongoId/index.js
toddym42: I think the only case of this happening would be if the user manually entered the URL, so I think it's fair to redirect to `/not-found` in that case. There is precedent for this in `validateBank` which validates to `/not-found` when that check fails.

I think we do need to validate here to prevent the user entering something dodgy which then gets included in the upstream request to the portal-api.

Again, there is some inconsistency though where some validators at the UI server level will handle the result themselves, and others (such as `virusScanUpload`) will just set some `res.locals` value and let the controller handle the result. I guess that's more useful in the case where we want to show a specific page.

@ChristianMcc - might need your steer on this one. Would be good to discuss when you have time.

portal/templates/utilisation-report-service/previous-reports/previous-reports.njk
toddym42: `report` is more descriptive.

portal/server/routes/middleware/validateMongoId/index.js
ChristianMcc: Few points before I give a full answer:
- Definitely need to validate this as you said
- In the virusScanUpload etc. we mainly do this where we want to give custom error messages or errors that are likely to happen and need to stay on the same page

For this one I'm a little torn between routing to `/not-found` or not re-routing and just returning 'something went wrong' page. Given we should be confident in our system that we wouldn't give bad links then inclined to agree with Mike that this is only hit by manually entering a bad URL so I'll probably stick with this `not-found` approach!

JoshBinns2000 (approved):
Happy with the discussions around the comments I made :)

portal/server/api.js
```diff
@@ -871,17 +871,23 @@ const getDueReportDatesByBank = async (token, bankId) => {
   return response.data;
 };
 
+const downloadUtilisationReport = async (userToken, bankId, _id) => {
```
ChristianMcc, line 874: while it's already validated earlier and so is useless I'm semi thinking we add bankId validation here as it's in every function in this file as an SSRF fail safe. Reckon that's overkill or helpful for someone reading this in the future thinking it's missing?

portal-api/src/v1/controllers/utilisation-report-service/utilisation-report-download.controller.js
```diff
@@ -0,0 +1,58 @@
...
+
+    const bufferedFile = await fileshare.readFile({
+      fileshare: FILESHARES.UTILISATION_REPORTS,
+      folder: bankId,
+      filename,
+    });
+
+    if (bufferedFile.error) {
+      console.error(`Failed to get utilisation report for download with _id '${_id}' from Azure Storage`, bufferedFile.error);
+      return res.status(404).send();
```
ChristianMcc, line 41: Do we know it doesn't exist here?
I think I'd go more for if we get no hit on the database metadata then we 404 as it not existing otherwise it's a 500 as something went wrong but we think the file should exist

ChristianMcc (changes requested):
Nice work, few minor comments to go with your rebasing :)

portal/api-tests/utilisation-report-routes.api-test.js
```diff
@@ -0,0 +1,35 @@
...
+  validateToken: () => true,
+}));
+
+const { withRoleValidationApiTests } = require('./common-tests/role-validation-api-tests');
+const app = require('../server/createApp');
+const { get } = require('./create-api').createApi(app);
+const { ROLES } = require('../server/constants');
+const MOCK_BANKS = require('../test-helpers/mock-banks');
+
+describe('utilisation-report routes', () => {
```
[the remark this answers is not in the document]
toddym42: I think so.

portal/server/routes/utilisation-report-service/utilisation-report-download/index.js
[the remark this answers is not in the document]
toddym42: None of these validators currently use `express-validator` so `handleValidation` isn't currently need. We don't currently have thehave the `handleValidation` function in this project, and I guess this is because we're in the UI layer so wouldn't just return a `400` status response along with the errors. We would instead need to do a redirect, or render an error page, and this might be different depending on the particular validation handler. What do you think?

portal/server/api.js
toddym42: Yeah, I didn't add as it's already in the route handler. For that reason I probably wouldn't duplicate here, but I can see the argument for being safe too, so happy to go with your call on this. Let me know.

ChristianMcc: Hmm, I think leave as it is, validation at the router level is the better approach anyway so I guess we should move towards that!

portal/server/routes/utilisation-report-service/utilisation-report-download/index.js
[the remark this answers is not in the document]
ChristianMcc: Ah I've just seen validateMongoId and us in portal not portal-api. As it's just all going to not found and we don't need good errors that's fine - ignore my previous comment.
Could I get you to add in a console.warn or something on both the mongoId/bankId validation so we have a record of it happening in logs please?

portal-api/src/v1/controllers/utilisation-report-service/utilisation-report-download.controller.js
toddym42: Yeah, makes sense. This could be any error here so will update to the generic 500 👍

ChristianMcc (approved):
Lovely work!

ChristianMcc (approved):

PR #2336, opened 2023-11-22
fix(DTFS2-NONE): fix pipeline

AlexBramhill:
# Introduction
We have changed mock data loader to interact directly with mongodb. This has caused issues in the pipeline as we are not allowing enough time for mongodb to spin up

# Resolution
I've bumped up the existing wait timer

# Misc
Removed depreciated options in mongodb connection


ChristianMcc (approved):
I love longer sleeps
If it fixes it seems like a small increase compared to how long most of these steps take anyway :)

PR #2337, opened 2023-11-22
feat(FN-1041): add report period details to TFM 'Bank reports' screen

toddym42:
## Ticket

[FN-1041](https://ukef-dtfs.atlassian.net/browse/FN-1041)

## Introduction

The "Bank reports" screen in TFM should show show details of the current utilisation report reporting period, and the date that this month's report is due.

## Resolution

* Add a basic page outline for the "Bank reports" page and add a link to the nav bar
* Add API calls through TFM UI -> TFM API -> External API to get details of UK bank holidays
* Use bank holiday details to calculate the current utilisation report reporting period, and the date that this month's report is due.
* Add to new "Bank reports" page

### DIFF
--- a/portal-api/src/external-api/bank-holidays.js
+++ b/portal-api/src/external-api/bank-holidays.js
@@ -11,7 +11,7 @@ const headers = {
 
 /**
  * Resolves to the response of `GET /bank-holidays` from external-api.
- * @returns {Promise<import('axios').AxiosResponse>}
+ * @returns {import('./types').BankHolidaysResponseBody}
  */
 const getBankHolidays = async () => {
   const response = await axios.get(`${EXTERNAL_API_URL}/bank-holidays`, { headers });
@@ -20,7 +20,7 @@ const getBankHolidays = async () => {
 
 /**
  * Fetches a list of bank holiday dates for the specified UK region
- * @param {'england-and-wales' | 'scotland' | 'northern-ireland'} region
+ * @param {import('./types').BankHolidayRegion} region
  * @returns {Promise<Date[]>}
  */
 const getBankHolidayDatesForRegion = async (region) => {
--- a/portal-api/src/external-api/types/bank-holidays.d.ts
+++ b/portal-api/src/external-api/types/bank-holidays.d.ts
@@ -1,0 +1,31 @@
+export type BankHolidayRegion = 'england-and-wales' | 'scotland' | 'northern-ireland';
+
+type BankHolidayEvent = {
+  /*
+   * Title of the event - e.g. 'New Year's Day'
+   */
+  title: string;
+  /*
+   * Date of the event ISO 8601 format - e.g. '2022-01-03'
+   */
+  date: string;
+  /*
+   * Additional notes about the event - e.g. 'Substitute day'
+   */
+  notes: string;
+  /*
+   * Indicates whether the bank holiday is a suitable occasion for hanging bunting
+   */
+  bunting: boolean;
+};
+
+type BankHolidayRegionEvents = {
+  division: BankHolidayRegion;
+  events: BankHolidayEvent[];
+};
+
+export type BankHolidaysResponseBody = {
+  'england-and-wales': BankHolidayRegionEvents;
+  scotland: BankHolidayRegionEvents;
+  'northern-ireland': BankHolidayRegionEvents;
+};
--- a/trade-finance-manager-ui/server/api.test.js
+++ b/trade-finance-manager-ui/server/api.test.js
@@ -1,17 +1,43 @@
-import api from './api';
-
 const axios = require('axios');
+const MockAdapter = require('axios-mock-adapter');
+const api = require('./api');
+const MOCK_BANK_HOLIDAYS = require('./test-mocks/mock-bank-holidays');
+const { getUkBankHolidays } = require('./api');
 
-jest.mock('axios');
+const mockAxios = new MockAdapter(axios);
+
+console.error = jest.fn();
 
 afterEach(() => {
   jest.clearAllMocks();
+  mockAxios.reset();
 });
 
 describe('getFacilities()', () => {
   it('returns the correct response', async () => {
-    axios.get.mockReturnValue(Promise.resolve({ data: { status: 200 } }));
+    mockAxios.onGet().reply(200);
     const response = await api.getFacilities();
     expect(response).toEqual({ facilities: [] });
+  });
+});
+
+describe('getUkBankHolidays', () => {
+  it('gets the bank holidays', async () => {
+    // Arrange
+    mockAxios.onGet().reply(200, MOCK_BANK_HOLIDAYS);
+
+    // Act
+    const response = await getUkBankHolidays('user-token');
+
+    // Assert
+    expect(response).toEqual(MOCK_BANK_HOLIDAYS);
+  });
+
+  it('throws when the api TFM API request fails', async () => {
+    // Arrange
+    mockAxios.onGet().reply(404);
+
+    // Act / Assert
+    await expect(getUkBankHolidays('user-token')).rejects.toThrowError('Request failed with status code 404');
   });
 });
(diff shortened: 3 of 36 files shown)


portal-api/src/external-api/bank-holidays.js
```diff
@@ -11,7 +11,7 @@ const headers = {
 
 /**
  * Resolves to the response of `GET /bank-holidays` from external-api.
- * @returns {Promise<import('axios').AxiosResponse>}
+ * @returns {object}
```
JoshBinns2000, line 14: Seeing as the type of bank holidays is fixed in external API, can be add a JSDoc typedef for it to use here?

trade-finance-manager-ui/server/controllers/utilisation-reports/index.js
```diff
@@ -0,0 +1,23 @@
...
+    const reportDueDate = await getFormattedReportDueDate(userToken);
+
+    return res.render('utilisation-reports/utilisation-reports.njk', {
+      activePrimaryNavigation: 'utilisation reports',
+      reportPeriod,
+      reportDueDate,
+    });
+  } catch (error) {
+    console.error('Error rendering utilisation reports page', error);
+    return res.redirect('/not-found');
```
JoshBinns2000, line 17: Can we re-introduce the 'something went wrong' page like we have in portal? Feels like `not-found` is dishonest from a user's point of view

trade-finance-manager-ui/server/helpers/date.test.js
```diff
@@ -0,0 +1,111 @@
...
+          { businessDay: 13, expectedResult: new Date('2023-11-17') },
+          { businessDay: 14, expectedResult: new Date('2023-11-20') },
+          { businessDay: 15, expectedResult: new Date('2023-11-21') },
+          { businessDay: 16, expectedResult: new Date('2023-11-22') },
+          { businessDay: 17, expectedResult: new Date('2023-11-23') },
+          { businessDay: 18, expectedResult: new Date('2023-11-24') },
+          { businessDay: 19, expectedResult: new Date('2023-11-27') },
+          { businessDay: 20, expectedResult: new Date('2023-11-28') },
+          { businessDay: 21, expectedResult: new Date('2023-11-29') },
+          { businessDay: 22, expectedResult: new Date('2023-11-30') },
```
JoshBinns2000, line 49: Do we really need all these test cases? Adding all these cases implies to me that we expect different behaviour when the `businessDay` goes into the next month (like throwing an error because the 23rd business day is December 1st). We don't do that as far as I can tell so I don't think we should have this many test cases

trade-finance-manager-ui/server/routes/index.test.js
```diff
@@ -19,14 +20,15 @@ describe('routes index', () => {
...
     expect(use).toHaveBeenCalledWith('/case', validateUser, caseRoutes);
     expect(use).toHaveBeenCalledWith('/deals', validateUser, dealsRoutes);
     expect(use).toHaveBeenCalledWith('/facilities', validateUser, facilitiesRoutes);
     expect(use).toHaveBeenCalledWith('/feedback', feedbackRoutes);
     expect(use).toHaveBeenCalledWith('/thank-you-feedback', feedbackThankYouRoutes);
     expect(use).toHaveBeenCalledWith('/user', userRoutes);
+    expect(use).toHaveBeenCalledWith('/utilisation-reports', validateUser, utilisationReportsRoutes);
     expect(use).toHaveBeenCalledWith('/', footerRoutes);
   });
 });
```
JoshBinns2000, line 34: I really dislike these test files as they feel redundant, but I suppose we can't really remove them at the moment

trade-finance-manager-ui/server/routes/utilisation-reports/index.test.js
```diff
@@ -0,0 +1,19 @@
...
+  afterEach(() => {
+    jest.resetAllMocks();
+  });
+
+  it('should setup routes with controllers', () => {
+    expect(get).toHaveBeenCalledTimes(1);
+
+    expect(get).toHaveBeenCalledWith('/', getUtilisationReports);
+  });
+});
```
JoshBinns2000, line 19: Should we even bother adding this test? I know it is done in the other files but I don't see the point in adding them for our new stuff (other than to remain consistent, which I think isn't a good enough reason when the test is this redundant)

trade-finance-manager-ui/server/services/utilisation-report-service.js
```diff
@@ -0,0 +1,49 @@
...
+/**
+ * Fetches a list of bank holiday dates for the specified UK region
+ * @param {string} userToken - Token to validate session
+ * @param {'england-and-wales' | 'scotland' | 'northern-ireland'} region
+ * @returns {Promise<Date[]>}
+ */
+const getBankHolidays = async (userToken, region) => {
+  const bankHolidays = await api.getUkBankHolidays(userToken);
+  return bankHolidays[region].events.map((event) => new Date(event.date));
+};
```
JoshBinns2000, line 15: Could we maybe make it so that `region` is instead an array of `regions` so that, if you wanted, you could get the holidays for england and wales and northern island in the same call of this function?

trade-finance-manager-ui/templates/_partials/primary-navigation.njk
```diff
@@ -4,22 +4,30 @@
...
         href: "/facilities",
         active: (activePrimaryNavigation === 'all facilities'),
         attributes: {
           "data-cy": "all-facilities-link"
         }
-    }
+      }
     ]
   }) }}
 {% endif %}
```
JoshBinns2000, line 33: Just in case anyone comments on these changes - I'm updating it all in FN-1404 (specifically only rendering certain navItems depending on the user team)

trade-finance-manager-ui/server/services/utilisation-report-service.js
toddym42: Not a bad idea, but we don't need it currently so it doesn't feel worth spending the time implementing it. If we do need it in the future we can just implement it then.

trade-finance-manager-ui/server/helpers/date.js
```diff
@@ -0,0 +1,69 @@
...
+  for (let i = 1; i < businessDay; i += 1) {
+    result = addOneBusinessDayWithHolidays(result, holidays);
+  }
+
+  return result;
+};
+
+module.exports = {
+  getBusinessDayOfMonth,
+};
```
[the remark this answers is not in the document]
toddym42: It's something we're definitely planning to explore with Abhi when he's back from leave 👍

trade-finance-manager-ui/server/api.test.js
```diff
@@ -1,17 +1,43 @@
...
 const axios = require('axios');
+const MockAdapter = require('axios-mock-adapter');
+const api = require('./api');
+const MOCK_BANK_HOLIDAYS = require('./test-mocks/mock-bank-holidays');
+const { getUkBankHolidays } = require('./api');
+
+const mockAxios = new MockAdapter(axios);
 
-jest.mock('axios');
+console.error = jest.fn();
```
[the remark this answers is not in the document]
JoshBinns2000: I think it is worth keeping in as suppressing the error logs is helpful for tests that can be run locally as it makes it easier to find which tests have failed

portal-api/src/external-api/types/bank-holidays.d.ts
```diff
@@ -0,0 +1,31 @@
+export type BankHolidayRegion = 'england-and-wales' | 'scotland' | 'northern-ireland';
+
+type BankHolidayEvent = {
+  /*
+   * Title of the event - e.g. 'New Year's Day'
+   */
+  title: string;
+  /*
+   * Date of the event ISO 8601 format - e.g. '2022-01-03
```
JoshBinns2000, line 9: This is missing a closing quotation on the example date I think

JoshBinns2000 (dismissed):
I've added one comment but am happy to approve without fixing that piece of documentation. Especially like the addition of the type declaration file to improve the JSDocs

portal-api/src/external-api/types/bank-holidays.d.ts
toddym42: 😢 

I'll fix in FN-1042

ChristianMcc (approved):

PR #2340, opened 2023-11-23
chore: only use --legacy-peer-deps on TFM UI package

toddym42:
Removes use of the `--legacy-peer-deps` flag on repos that don't need it.

It is only required in the `trade-finance-manager-ui` project, so a `.npmrc` file has been added to configure it to always be used in that project (e.g. if a user runs `npm install` in the project, it will automatically apply the `--legacy-peer-deps` flag.


.github/workflows/sca.yml
```diff
@@ -63,7 +63,7 @@ jobs:
 
       - name: Dependencies
         working-directory: ./
-        run: npm ci --legacy-peer-deps
+        run: npm ci
```
toddym42, line 66: I wasn't as sure about removing the flag here as I'm not entirely sure what this will run on.

trade-finance-manager-ui/.npmrc
```diff
@@ -0,0 +1 @@
+legacy-peer-deps=true
```
toddym42, line 1: This ensures the `--legacy-peer-deps` flag is always applied to relevant npm commands in this project.

ChristianMcc (approved):
LGTM and agree with having npm ci at the top level in the sca.yml workflow so no changes needed :)

PR #2341, opened 2023-11-23
feat(FN-1089): fix report uploadedBy schema

JoshBinns2000:
# Introduction

FN-1089 uses the `firstname` and `surname` properties from the `user` who uploaded the report. However, the `uploadedBy` user was stored in a way that did not match the other `user`s stored in portal.

# Resolution

Change how the utilisation reports are stored so that `firstname` and `surname` from the `user` are stored with the report. Updates other parts of the code where `uploadedBy.name` is used to match this new format.


ChristianMcc (approved):
nice

PR #2342, opened 2023-11-23
feat(DTFS2-6461): change p to h2 tags for visual headings on TFM deal page

MarRobSoftwire:
## Introduction
The tfm deal page has visual headings `Facilities`, `Bank Security` and `Bank Application or Notice` which are not semantically marked up with h tags. This is problematic for blind screen reader users

## Resolution
Replaced the `p` tags with `h2` tags.

### Before
image_720

### After
image_720

## Notes
I cannot render this locally but have used inspect element on the dev site to check the change. There is a slight change to the styles with `margin-block-start` and `margin-block-end` going from `1em` to `.83em`.

### DIFF
--- a/trade-finance-manager-ui/templates/case/deal/deal.njk
+++ b/trade-finance-manager-ui/templates/case/deal/deal.njk
@@ -21,7 +21,7 @@
 
   <div class="govuk-grid-row" data-cy="deal-facilities">
     <div class="govuk-grid-column-full separator-line deal">
-      <p class="ukef-heading-l" id="deal-ukef-heading">Facilities</p>
+      <h2 class="ukef-heading-l" id="deal-ukef-heading">Facilities</h2>
 
         {% if deal.dealType == 'GEF' %}
           {% if deal.isFinanceIncreasing %}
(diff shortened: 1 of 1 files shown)


MarRobSoftwire:
I checked the rest of the code for `p` tags using a `ukef-heading` style and found a couple of examples
* `trade-finance-manager-ui\templates\case\parties\_macros\parties-key-value-array-grid-row.njk`
* `trade-finance-manager-ui\templates\case\parties\_macros\parties-key-value-grid-row.njk`
These each use `ukef-heading-xs` but I cant tell if semantically this should be a heading

lmarrai-sw:
> I checked the rest of the code for `p` tags using a `ukef-heading` style and found a couple of examples
> 
> * `trade-finance-manager-ui\templates\case\parties\_macros\parties-key-value-array-grid-row.njk`
> * `trade-finance-manager-ui\templates\case\parties\_macros\parties-key-value-grid-row.njk`
> 
> These each use `ukef-heading-xs` but it looks like they're key / value pairs in a table so probably doesn't want to be a heading

I'll run these locally to check them, thanks for flagging them @MarRobSoftwire

lmarrai-sw (approved):

PR #2343, opened 2023-11-23
feat(DTFS2-none): remove aria label tests and update underwriting heading test

MarRobSoftwire:
## Introduction
#2283 didnt update the e2e tests 

##Resolution
* remove redundant checks for `aria-labels` that have been removed
* update check for `deal` heading to `underwriting`


lmarrai-sw (approved):
Thanks @MarRobSoftwire

PR #2344, opened 2023-11-23
fix: add back eslint parser typescript defs

ChristianMcc:
## Introduction
Fix any issues from the most recent main -> main-stbr merge

### DIFF
--- a/portal/templates/admin/user-edit.njk
+++ b/portal/templates/admin/user-edit.njk
@@ -7,6 +7,7 @@
 
 
 {% extends "index.njk" %}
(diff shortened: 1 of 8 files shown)


portal/templates/admin/user-edit.njk
JoshBinns2000, line 1: Lines 21 to 27 shouldn't be there and line 10 should be - this is to do with the positioning of the back-link

ChristianMcc: good catch! fixed :)

PR #2345, opened 2023-11-24
fix(DTFS2-6465): moving headers to date form component

avaitonis:
### Introduction
Date form component consist from 3 text input fields with 3 labels: day, month, year. These 3 elements need 1 shared/common label. 

### Resolution
- All date components in TFM are used in pages with single question. Moving page header to `govukDateInput` fieldset label.
- `govukDateInput` `fieldset.label` doesn't support custom attributes, to have our `data-cy` I had to add extra `span` element.
- One govukDateInput uses different group approach because it was not possible to add custom element after heading.

### DIFF
--- a/trade-finance-manager-ui/templates/case/amendments/amendment-add-banks-decision-effective-date.njk
+++ b/trade-finance-manager-ui/templates/case/amendments/amendment-add-banks-decision-effective-date.njk
@@ -20,71 +20,78 @@
       }) }}
   {% endif %}
 
-  <h1 class="govuk-heading-l govuk-!-margin-top-4 govuk-!-margin-bottom-2" data-cy="amendment--banks-decision-effective-date-heading" aria-label="What date will the amendment be effective from?">What date will the amendment be effective from?</h1>
-   <div class="govuk-grid-row">
+  <div class="govuk-grid-row">
     <div class="govuk-grid-column-full">
-      <form method="POST">
+      <form method="POST" autocomplete="off" class="govuk-!-margin-top-4">
         <input type="hidden" name="_csrf" value="{{ csrfToken }}">
-          {{ govukDateInput({
-            id: "bankDecisionDate",
-            namePrefix: "amendment--bank-decision-date",
-            hint: {
-              text: "For example, 31 3 1980",
-              attributes: {
-                "data-cy": "amendment--receive-bank-decision-date-hint"
-              }
-            },
-            errorMessage: errors and errors.fieldErrors.bankDecisionDate and {
-              text: errors.fieldErrors.bankDecisionDate.text,
-              attributes: {
-                'data-cy': 'amendment--inline-error'
-              }
-            },
-            items: [
-              {
-                label: "Day",
-                classes: errors.fieldErrors.bankDecisionDate and "govuk-input--error govuk-input--width-2" or "govuk-input--width-2",
-                name: "day",
-                value: bankDecisionDateDay,
-                attributes: {
-                  'data-cy': "amendment--bank-decision-date-day"
-                }
-              },
-              {
-                label: "Month",
-                classes: errors.fieldErrors.bankDecisionDate and "govuk-input--error govuk-input--width-2" or "govuk-input--width-2",
-                name: "month",
-                value: bankDecisionDateMonth,
-                attributes: {
-                  'data-cy': "amendment--bank-decision-date-month"
-                }
-              },
-              {
-                label: "Year",
-                classes: errors.fieldErrors.bankDecisionDate and "govuk-input--error govuk-input--width-4" or "govuk-input--width-4",
-                name: "year",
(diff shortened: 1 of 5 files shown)


trade-finance-manager-ui/templates/case/amendments/amendment-cover-end-date.njk
```diff
@@ -20,59 +20,67 @@
...
-        }
-      },
-      items: [
-        {
-          label: "Day",
-          classes: errors.fieldErrors.coverEndDate and "govuk-input--error govuk-input--width-2" or "govuk-input--width-2",
-          name: "day",
-          value: coverEndDateDay,
+
+    <fieldset class="govuk-fieldset" role="group" aria-describedby="amendmentRequestDate-hint">
```
lmarrai-sw, line 27:
```suggestion
    <fieldset class="govuk-fieldset" role="group" aria-describedby="coverEndDate-hint">
```

I think the `aria-describedby` attribute needs to have the id of the `hint` element, which is `coverEndDate-hint`

trade-finance-manager-ui/templates/case/amendments/amendment-cover-end-date.njk
```diff
@@ -20,59 +20,67 @@
...
+
+      <legend class="govuk-fieldset__legend">
+        <h1 class="govuk-heading-l govuk-!-margin-top-4 govuk-!-margin-bottom-0" data-cy="amendment--cover-end-date-heading">Enter the new cover end date</h1>
+      </legend>
+
+      <div class="govuk-body govuk-!-margin-bottom-4" data-cy="amendment--current-cover-end-date">
+        <p class="govuk-body govuk-!-margin-bottom-0 govuk-!-font-weight-bold">Current cover end date</p>
+        {{currentCoverEndDate}}
```
lmarrai-sw, line 35: I know this wasn't mentioned in this ticket, but do you think we should change this to a description list (`<dl><dt><dd>`)?

trade-finance-manager-ui/templates/case/amendments/amendment-add-banks-decision-effective-date.njk
```diff
@@ -20,71 +20,78 @@
       }) }}
   {% endif %}
 
-  <h1 class="govuk-heading-l govuk-!-margin-top-4 govuk-!-margin-bottom-2" data-cy="amendment--banks-decision-effective-date-heading" aria-label="What date will the amendment be effective from?">What date will the amendment be effective from?</h1>
-   <div class="govuk-grid-row">
+  <div class="govuk-grid-row">
     <div class="govuk-grid-column-full">
-      <form method="POST">
+      <form method="POST" autocomplete="off" class="govuk-!-margin-top-4">
```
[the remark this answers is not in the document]
avaitonis: Not sure, but it is true that most of our inputs should be unique.

We have `autocomplete="off"` in 99 forms out of 116.

trade-finance-manager-ui/templates/case/amendments/amendment-add-banks-decision-receive-date.njk
```diff
@@ -20,16 +20,21 @@
...
   {% endif %}
 
-  <h1 class="govuk-heading-l govuk-!-margin-top-4 govuk-!-margin-bottom-2" data-cy="amendment--banks-decision-receive-date-heading" aria-label="What date did UKEF receive the bank’s decision??">What date did UKEF receive the bank’s decision?</h1>
-
-
-   <div class="govuk-grid-row">
+  <div class="govuk-grid-row">
     <div class="govuk-grid-column-full">
-      <form method="POST">
+      <form method="POST" autocomplete="off" class="govuk-!-margin-top-4">
```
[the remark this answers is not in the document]
avaitonis: I moved H1 inside of fieldset and legend elements and this causes interesting change in margin behaviour.

I spent quite some time playing with css, but margin for form element was best solution that I found.

trade-finance-manager-ui/templates/case/amendments/amendment-cover-end-date.njk
avaitonis: yes, it is single item, but dl might be good, changed.

trade-finance-manager-ui/templates/case/amendments/amendment-cover-end-date.njk
avaitonis: Thanks, fixed.

trade-finance-manager-ui/templates/case/amendments/amendment-add-banks-decision-effective-date.njk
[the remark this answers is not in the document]
lmarrai-sw: Okay thanks

lmarrai-sw (approved):
Thanks for this @avaitonis

PR #2346, opened 2023-11-24
fix: main-stbr tests

JoshBinns2000:
Introduces some fixes for failing tests following the merge of `main` into `main-stbr`


e2e-tests/tfm/cypress/e2e/journeys/case/activities/activities.spec.js
```diff
@@ -61,9 +61,6 @@ context('Users can create and submit comments', () => {
   describe('add a comment', () => {
     it('should show correct heading and aria-label for activities tab heading', () => {
       activitiesPage.mainHeading().contains('Activity and comments');
-      activitiesPage.mainHeading().invoke('attr', 'aria-label').then((label) => {
```
ChristianMcc, line 64: This feels like it's checking an aria-label and was added for a reason, even if it's just hte same as the heading. Is this also gone on main?

ChristianMcc (commented):
just the comments about the aria-label parts of the tests and if they are also gone from main :)

e2e-tests/tfm/cypress/e2e/journeys/case/activities/activities.spec.js
toddym42: Not sure if this is why it was removed, but heading elements should have `aria-label` attributes. They should be for [interactive elements only](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label), and where the default a11y content doesn't describe it accurately. In this case it's not interactive, and is the same text as the actual heading text, so should be removed.

JoshBinns2000: This came from this PR: #2343

toddym42: To add, it looks like the `aria-label` has now been removed from the header in [activity.njk](https://github.com/UK-Export-Finance/dtfs2/blob/fix/main-stbr-tests/trade-finance-manager-ui/templates/case/activity/activity.njk#L20) so it was correct to remove the check from the test.

ChristianMcc (approved):

PR #2353, opened 2023-11-28
feat(FN-1536): add dtfs-central-api endpoint to set report status

JoshBinns2000:
# Introduction

We want an endpoint which the TFM API can use to mark a report's status as either `REPORT_NOT_RECEIVED` or `PENDING_RECONCILIATION`.

# Resolution

- Adds the required `PUT` endpoint which accepts an array of `reportsWithStatus` in the request body
- If the report already exists, the status of that report is modified*
- If the report does not exist and the status to set is `PENDING_RECONCILIATION`, a dummy report is inserted to the database with the `azureFileInfo` being set to `undefined`
- Adds tests to ensure the endpoint works as expected

*If the report exists but does not have a well-defined `azureFileInfo`, it is assumed that this was a dummy report generated by this endpoint. If the status to set in the payload is `REPORT_NOT_RECEIVED` and the existing report is a dummy report, the report is deleted.


dtfs-central-api/.eslintrc.js
```diff
@@ -32,12 +33,21 @@ module.exports = {
...
         functions: false
       }
+    ],
+    'import/extensions': [
+      'error',
+      'ignorePackages',
+      {
+        js: 'never',
+        ts: 'never'
+      }
```
JoshBinns2000, line 43: I added this to remove the errors from lines like `export * from './mock-utilisation-reports'`

dtfs-central-api/src/v1/routes/tfm-routes.js
```diff
@@ -722,4 +724,54 @@ tfmRouter.route('/users/team/:teamId')
...
+ *                       example: 1
+ *                       description: One-indexed month
+ *                     year:
+ *                       type: number
+ *                       example: 2023
+ *                       description: The full year
+ *                     bankId:
+ *                       type: string
+ *                       example: '123'
+ *                       description: The bank id
```
JoshBinns2000, line 764: I'm not sure how I am supposed to document this

PR #2356, opened 2023-11-28
fix(DTFS2-6822): fix linting in typescript services; remove typescript support from webpack builds

lmarrai-sw:
## Introduction

Since we introduced Typescript support to the repos, we've had a few issues with linting

## Resolution

I have:
- installed `@typescript-eslint/eslint-plugin` and `@typescript-eslint/eslint-parser` in the root package.json and configured these to be used by all of the (non-Azure) services for TS files _only_
  - also upgraded these packages in `external-api` where they are also imported
- configured all of the (non-Azure) services to extend the `plugin:@typescript-eslint/recommended-type-checked` config 
  - except for `external-api` because it had too many linting errors with this config to fix in this PR. I will create a separate ticket to add this to `external-api`.
- changed how the `azure-env/index.ts` file is used (the function now runs when the file is imported, instead of having to import the function and run it) to avoid linting errors asking for all other imports to run before the function is called
- installed `@types/express` in all (non-Azure) packages that did not already have it
- removed support for Typescript files in the `scripts` directories of `portal`, `gef-ui`, and `tfm-ui`, because `npm run build` was generating very different output with the TS support and we haven't been able to test that the new output works correctly. I'll raise a separate ticket for adding TS support to the webpack builds
- replaced the `optimize-css-assets-webpack` webpack plugin in `tfm-ui` with the `css-minimizer-webpack-plugin`
  - I had to do this because the `optimize-css-assets-webpack` plugin does not support webpack v5 and we use webpack v5 in tfm-ui: so `npm` would not let me install new packages in `tfm-ui` since it could not resolve the dependencies with these 2 conflicting restrictions
- removed some packages from `tfm-ui` that haven't been used since we removed the GraphQL endpoints from `tfm-api`
- installed `tslib` in any of the TS services that didn't have it, since our TS output assumes that package is available


external-api/.eslintrc.js
```diff
@@ -5,6 +5,7 @@ module.exports = {
     'eslint:recommended',
     'plugin:import/recommended',
     'plugin:import/typescript',
+    // TODO DTFS2-6822: Raise a ticket to increase this to 'plugin:@typescript-eslint/recommended-type-checked'
```
lmarrai-sw, line 8: Flagging TODO

portal/webpack.common.config.js
```diff
@@ -4,7 +4,7 @@ const CssMinimizerPlugin = require('css-minimizer-webpack-plugin');
 
 module.exports = {
   entry: {
-    main: './scripts/main.ts',
+    main: './scripts/main.js',
```
lmarrai-sw, line 7: I should raise a ticket to add TS support to the webpack files

gef-ui/package-lock.json
```diff
@@ -12326,127 +12478,6 @@
...
-      },
-      "engines": {
-        "node": ">=8"
-      }
-    },
-    "node_modules/ts-loader/node_modules/yallist": {
-      "version": "4.0.0",
-      "resolved": "https://registry.npmjs.org/yallist/-/yallist-4.0.0.tgz",
-      "integrity": "sha512-3wdGidZyq5PB084XLES5TpOSRA3wjXAlIWMhum2kRcv/41Sn2emQ0dycQW4uZXLejwKvg6EsvbdlVL+FYEct7A=="
-    },
```
oscar-richardson-softwire, line 12449: Why were all of these removed?

gef-ui/.eslintrc.js
```diff
@@ -1,3 +1,7 @@
+const baseParserOptions = {
+  ecmaVersion: 2020,
+};
+
```
oscar-richardson-softwire, line 4: Nice and DRY, like it 👍

portal-api/tsconfig.eslint.json
```diff
@@ -3,7 +3,7 @@
     "include": [
         "src",
         "api-tests",
-        "_mocks_",
+        "__mocks__",
```
oscar-richardson-softwire, line 6: Good catch 👍

portal/package-lock.json
```diff
@@ -3589,15 +3588,6 @@
...
     },
-    "node_modules/ajv-keywords": {
-      "version": "3.5.2",
-      "resolved": "https://registry.npmjs.org/ajv-keywords/-/ajv-keywords-3.5.2.tgz",
-      "integrity": "sha512-5p6WTN0DdTGVQk6VjcEju19IgaHudalcfabD7yhDGeA6bcQnmL+CpveLJq/3hvfwd1aof6L386Ougkx6RfyMIQ==",
-      "dev": true,
-      "peerDependencies": {
-        "ajv": "^6.9.1"
-      }
-    },
```
oscar-richardson-softwire, line 3600: Why was this removed?

portal/package-lock.json
```diff
@@ -12888,20 +12746,6 @@
...
-      "dev": true,
-      "peer": true,
-      "bin": {
-        "tsc": "bin/tsc",
-        "tsserver": "bin/tsserver"
-      },
-      "engines": {
-        "node": ">=14.17"
-      }
-    },
```
oscar-richardson-softwire, line 12904: Why was this removed?

trade-finance-manager-api/tsconfig.eslint.json
```diff
@@ -1,11 +1,8 @@
 {
     "extends": "./tsconfig.json",
     "include": [
-        "server",
+        "src",
         "api-tests",
-        "component-tests",
-        "scripts",
-        "test-helpers",
```
oscar-richardson-softwire, line 8: Good catch, this directories don't exist so I'm guessing this tsconfig was originally copied across from another service and not changed or something.

gef-ui/package-lock.json
```diff
@@ -46,7 +46,6 @@
         "path": "0.12.7",
         "redis": "^3.1.2",
         "sanitize-html": "^2.11.0",
-        "ts-loader": "^9.5.0",
```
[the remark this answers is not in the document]
lmarrai-sw: Yep that's exactly right

trade-finance-manager-api/tsconfig.eslint.json
lmarrai-sw: Yep I think exactly that

portal/package-lock.json
lmarrai-sw: It's hard to say on this one, since it was done by `npm` automatically

portal/webpack.common.config.js
lmarrai-sw: https://ukef-dtfs.atlassian.net/browse/DTFS2-6831

PR #2358, opened 2023-11-29
fix(FN-976): fix report download link

toddym42:
## Introduction

Various fixes to get utilisation report downloads working:

* Updated `utilisation-reports/:_id` endpoint in central API to start with a `/`
* Send all report details on `previous-reports` endpoint and do mapping in UI layer (previously only had month name and (now removed) path, so couldn't download report by ID and set correct filename in anchor link `download` attribute)
* changed return type of `api.downloadUtilisationReport` to include entire axios response instead of just the data, so headers can also be used to populate the file `Content-Disposition` and `Content-Type`
* Fixed `validateBankIdForUser` in Portal UI to get user from `req.session` instead of just `req` (which is how we access in the API layer)


portal-api/api-tests/v1/utilisation-report-service/previous-reports.api-test.js
```diff
@@ -31,23 +31,23 @@ describe('GET /v1/previous-reports/:bankId', () => {
         bank,
         month: 1,
         year,
-        dateUploaded: new Date(year, 0),
+        dateUploaded: new Date(year, 0).toISOString(),
```
JoshBinns2000, line 34: I'm slightly concerned about doing it this way as the reports we actually feed into the database use a `Date` object for `dateUploaded` rather than an ISO string. It might be better to use a function in line 88 to convert the dates in `expectedResponse` to their respective ISO strings. Alternatively, we can use a function in `JSON.parse` to re-convert ISO strings to `Date` objects like this https://stackoverflow.com/questions/4511705/how-to-parse-json-to-receive-a-date-object-in-javascript

portal-api/src/v1/controllers/utilisation-report-service/previous-reports.controller.test.js
JoshBinns2000, line 1: Does this file not need to update `reports` to have `azureFileInfo` rather than `path`?

portal/server/controllers/utilisation-report-service/utilisation-report-download/index.js
```diff
@@ -1,18 +1,17 @@
...
   const { userToken } = req.session;
   const { bankId, _id } = req.params;
 
-  const fileData = await getApiData(api.downloadUtilisationReport(userToken, bankId, _id), res);
+  const { data, headers } = await api.downloadUtilisationReport(userToken, bankId, _id);
 
-  res.set('Content-Disposition', fileData.headers['Content-Disposition']);
-  res.set('Content-Type', fileData.headers['Content-Type']);
+  res.set('Content-Disposition', headers['content-disposition']);
+  res.set('Content-Type', headers['content-type']);
```
JoshBinns2000, line 11: In one of the api tests above, you use `'content-disposition'` and `'content-type'` for the headers instead of the capitalised versions. Which one are you expecting to use?

portal/api-tests/utilisation-report-routes.api-test.js
```diff
@@ -10,8 +10,8 @@ jest.mock('../server/api', () => ({
   validateToken: () => true,
   downloadUtilisationReport: jest.fn().mockResolvedValue({
     headers: {
-      'Content-Disposition': `attachment; filename=report.csv`,
-      'Content-Type': 'text/csv',
+      'content-disposition': `attachment; filename=report.csv`,
+      'content-type': 'text/csv',
```
JoshBinns2000, line 14: Linked to the comment below

JoshBinns2000 (commented):
Looks mostly fine - just a few points of clarification and suggestions from me

portal-api/api-tests/v1/utilisation-report-service/previous-reports.api-test.js
```diff
@@ -84,30 +97,9 @@ describe('GET /v1/previous-reports/:bankId', () => {
...
-      },
-    ];
-
-    const response = await as(aPaymentReportOfficer).get(previousReportsUrl(matchingBankId));
+  it('returns a 200 response for a valid request', async () => {
+    const { status } = await as(aPaymentReportOfficer).get(previousReportsUrl(matchingBankId));
 
-    expect(response.status).toEqual(200);
-    expect(JSON.parse(response.text)).toEqual(expectedResponse);
+    expect(status).toEqual(200);
```
JoshBinns2000, line 103: Are we sure about removing the check of the return body type? If the date is that big of a pain maybe only check the month, year and azureFileInfo?

portal-api/src/v1/controllers/utilisation-report-service/previous-reports.controller.test.js
```diff
@@ -1,117 +1,65 @@
...
-    reports: [{
-      month: 'January',
+  const year2020Reports = [
+    {
+      bankId: '9',
+      month: 12,
+      year: 2020,
+      dateUploaded: '2023-02-01T00:00',
+      uploadedById: '1',
       path: 'www.abc.com',
```
JoshBinns2000, line 11: Should this not also be azureFileInfo?

portal-api/api-tests/v1/utilisation-report-service/previous-reports.api-test.js
toddym42: I'm pretty sure. Happy to wait for @ChristianMcc to approve instead though if you're uncomfortable with this.

toddym42: Found a way to assert on the response without having to manually map all the `ObjectId` (for mongo `_id` values) and `Date` objects! Maybe there is some value in a single happy path test to at least make sure the correct controller is called. Either way I think it would still be interesting to have this discussion.

JoshBinns2000: lol I love the hack of stringifying and then posing as JSON dad

portal/server/controllers/utilisation-report-service/utilisation-report-download/index.js
```diff
@@ -1,18 +1,17 @@
 const stream = require('stream');
 const api = require('../../../api');
-const { getApiData } = require('../../../helpers');
 
 const getReportDownload = async (req, res) => {
   const { userToken } = req.session;
   const { bankId, _id } = req.params;
 
-  const fileData = await getApiData(api.downloadUtilisationReport(userToken, bankId, _id), res);
+  const { data, headers } = await api.downloadUtilisationReport(userToken, bankId, _id);
```
ChristianMcc, line 8: does this function need to be wrapped in try/catch?

portal-api/api-tests/v1/utilisation-report-service/previous-reports.api-test.js
ChristianMcc: I think I would have been fine with just checking status and treating this as just a green/red on can we get a good response. The stringify/parse hack is nicer though for sure and a nicer form of regression testing so works for me :)

ChristianMcc (commented):
LGTM - just one small try/catch check

portal/server/controllers/utilisation-report-service/utilisation-report-download/index.js
```diff
@@ -1,17 +1,28 @@
 const stream = require('stream');
 const api = require('../../../api');
 
+/**
+ * Fetches a utilisation report DSV file for download
```
ChristianMcc, line 5: CSV*

ChristianMcc (approved):

PR #2359, opened 2023-11-29
fix: ensure pipeline fails if dtfs-central and external api tests fail

lmarrai-sw:
## Introduction

@francescastocco and @JoshBinns2000 noticed that the pipeline was not failing if DTFS Central API or External API API tests were failing. We tracked this down to a test teardown that would always set the exit code to 0 regardless of the test result.

## Resolution

I've removed the test teardown files that would set the exit code to 0.
I've removed references to the "api-test-file" scripts as they're not needed at all, the "api-test" script does the same job

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,7 +131,7 @@ npm run api-test
 #### Run a Single API Test :heavy_check_mark:
 
 ```shell
-npm run api-test-file "**/*/deals-party-db.api-test.js"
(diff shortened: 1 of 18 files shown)


francescastocco (dismissed):
LGTM, thanks for fixing :)

francescastocco (approved):
🎉

PR #2360, opened 2023-11-29
[DO NOT MERGE]: Test PR to check that the pipeline fails if dtfs-central-api and external-api API tests fail

lmarrai-sw:
**This PR should not be merged**

#2359 fixes the pipeline to fail if the DTFS Central API or External API tests fail

This PR **breaks** the API tests in DTFS Central and External API to make sure the pipeline fails.

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,7 +131,7 @@ npm run api-test
 #### Run a Single API Test :heavy_check_mark:
 
 ```shell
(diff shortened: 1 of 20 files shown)


lmarrai-sw:
The DTFS Central API and External API API tests failed and caused the pipeline to fail, as intended (I then cancelled the rest of the test runs) so this PR can be closed

PR #2362, opened 2023-11-30
feat(FN-1618): add dtfs-central-api endpoint to set report status

JoshBinns2000:
# Introduction

We want an endpoint which the TFM API can use to mark a report's status as either `REPORT_NOT_RECEIVED` or `PENDING_RECONCILIATION`.

# Resolution

- Adds the required `PUT` endpoint which accepts an array of `reportsWithStatus` in the request body
- If the report already exists, the status of that report is modified*
- If the report does not exist and the status to set is `PENDING_RECONCILIATION`, a dummy report is inserted to the database with the `azureFileInfo` being set to `undefined`
- Adds tests to ensure the endpoint works as expected

*If the report exists but does not have a well-defined `azureFileInfo`, it is assumed that this was a dummy report generated by this endpoint. If the status to set in the payload is `REPORT_NOT_RECEIVED` and the existing report is a dummy report, the report is deleted.


dtfs-central-api/src/v1/controllers/utilisation-report-service/tfm-put-utilisation-report-status.controller.ts
```diff
@@ -0,0 +1,135 @@
...
+  ReportFilter,
+  ReportStatus,
+} from './types';
+import * as db from '../../../drivers/db-client';
+import { DB_COLLECTIONS } from '../../../constants/dbCollections';
+
+const REPORT_STATUS: Record<ReportStatus, ReportStatus> = {
+  PENDING_RECONCILIATION: 'PENDING_RECONCILIATION',
+  REPORT_NOT_RECEIVED: 'REPORT_NOT_RECEIVED',
+};
```
JoshBinns2000, line 21: I originally did this as an enum, but I don't know of a way to tell an enum that it needs to match the type of `ReportStatus`

dtfs-central-api/api-tests/v1/utilisation-report/tfm-put-utilisation-report-status.api-test.js
JoshBinns2000, line 1: This test is quite granular and, although it doesn't need to be, I'd like to keep it that way as the way this function interacts with the database is pretty important

francescastocco (dismissed):
This looks good :)

dtfs-central-api/src/v1/controllers/utilisation-report-service/tfm-put-utilisation-report-status.controller.ts
francescastocco: Can you make the type `ReportStatus` an enum and then remove this variable?

dtfs-central-api/src/services/repositories/utilisation-report-status-repo.ts
```diff
@@ -0,0 +1,116 @@
...
+};
+
+const createOrSetReportAsReceived = (reportDetails: ReportDetails, filter: ReportFilter, collection: Collection): Promise<UpdateResult> => {
+  const placeholderUtilisationReport: PlaceholderUtilisationReport = {
+    month: reportDetails.month,
+    year: reportDetails.year,
+    bank: {
+      id: reportDetails.bankId,
+    },
+    azureFileInfo: undefined,
```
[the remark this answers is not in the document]
JoshBinns2000: If I wanted to populate the full bank details, I would need to get the bank from the database as part of this function (I don't want to pass the entire bank in the payload)

dtfs-central-api/src/v1/controllers/utilisation-report-service/tfm-put-utilisation-report-status.controller.ts
JoshBinns2000: @toddym42 I can't remember exactly what we discussed but what was the reason we had for not using enums?

dtfs-central-api/api-tests/v1/utilisation-report/tfm-put-utilisation-report-status.api-test.js
```diff
@@ -0,0 +1,243 @@
+const wipeDB = require('../../wipeDB');
+const app = require('../../../src/createApp');
+const api = require('../../api')(app);
+const db = require('../../../src/drivers/db-client');
+const { mockUtilisationReports } = require('../../mocks/tfm-utilisation-reports/mock-utilisation-reports').default;
```
JoshBinns2000, line 5: This doesn't need to be a `default` anymore

dtfs-central-api/src/types/tfm/teams.ts
```diff
@@ -0,0 +1,8 @@
+type TeamId =
+ | 'PDC_READ';
```
JoshBinns2000, line 2: I need to finish this list

dtfs-central-api/src/types/utilisation-report-status.ts
```diff
@@ -0,0 +1,46 @@
...
+  | {
+      month: number;
+      year: number;
+      'bank.id': string;
+    };
+
+export type UtilisationReport = {
+  bank: {
+    id: string;
+    name?: string;
```
JoshBinns2000, line 37: I've added this as optional for the sake of the placeholder report - I'm planning on re-thinking how I implement this though

dtfs-central-api/unit.jest.config.js
```diff
@@ -1,5 +1,7 @@
 module.exports = {
+  preset: 'ts-jest',
+  testEnvironment: 'node',
```
JoshBinns2000, line 3: I can remove `testEnvironment`

dtfs-central-api/src/services/repositories/utilisation-report-status-repo.ts
```diff
@@ -0,0 +1,120 @@
...
+  const placeholderUtilisationReport: UtilisationReport = {
+    month: reportDetails.month,
+    year: reportDetails.year,
+    bank: {
+      id: reportDetails.bankId,
+    },
+    azureFileInfo: null,
+    uploadedBy: user,
+    dateUploaded: new Date(),
+  };
```
JoshBinns2000, line 58: If I want to add the `bank.name` field here, I would need to get the bank from the database using the `bankId` provided. I'm not sure if it is worthwhile to do this for the sake of the placeholder

dtfs-central-api/src/services/repositories/utilisation-report-status-repo.test.ts
```diff
@@ -0,0 +1,249 @@
...
+      // Assert
+      expect(updateOneSpy).toHaveBeenCalledWith(expectedFilter, {
+        $set: {
+          status,
+        },
+      });
+    });
+
+    it("should call updateOne to set the status as 'RECONCILIATION_COMPLETED' when the status is 'RECONCILIATION_COMPLETED'", () => {
+      // Arrange
```
JoshBinns2000, line 218: Placeholder report should be moved here

dtfs-central-api/src/services/repositories/utilisation-report-status-repo.test.ts
```diff
@@ -0,0 +1,249 @@
...
+    };
+    const placeholderReport: UtilisationReport = {
+      month: 1,
+      year: 2023,
+      bank: {
+        id: '123',
+      },
+      azureFileInfo: null,
+      uploadedBy: mockTfmUser,
+      dateUploaded: new Date(),
```
JoshBinns2000, line 188: Mock the system time in order to avoid test being flakey

dtfs-central-api/src/services/repositories/utilisation-report-status-repo.test.ts
```diff
@@ -0,0 +1,249 @@
...
+    } as UtilisationReport;
```
JoshBinns2000, line 189: Remove `as UtilisationReport`

dtfs-central-api/src/services/repositories/utilisation-report-status-repo.test.ts
```diff
@@ -0,0 +1,249 @@
...
+    const reportDetails: ReportDetails = {
+      month: 1,
+      year: 2023,
+      bankId: '123',
+    };
+    const placeholderReport: UtilisationReport = {
+      month: 1,
+      year: 2023,
+      bank: {
+        id: '123',
```
JoshBinns2000, line 184: Add bank name

dtfs-central-api/src/v1/controllers/utilisation-report-service/tfm-put-utilisation-report-status.controller.ts
```diff
@@ -0,0 +1,34 @@
...
+import { Collection, DeleteResult, UpdateResult } from 'mongodb';
+
+import db from '../../../drivers/db-client';
+import { DB_COLLECTIONS } from '../../../constants/dbCollections';
+import { PutReportStatusRequestBody } from '../../../types/utilisation-report-status';
+import { setReportStatusByReportId, setReportStatusByReportDetails } from '../../../services/repositories/utilisation-report-status-repo';
+
+export const putUtilisationReportStatus = async (req: Request<{}, {}, PutReportStatusRequestBody>, res: Response) => {
+  try {
+    const { reportsWithStatus, user } = req.body;
```
JoshBinns2000, line 11: Does `user` exist on `req.session`?

dtfs-central-api/src/services/repositories/utilisation-report-status-repo.ts
```diff
@@ -0,0 +1,120 @@
...
+import { TFMUser } from '../../types/users';
+
+const REPORT_STATUS: Record<ReportStatus, ReportStatus> = {
+  RECONCILIATION_COMPLETED: 'RECONCILIATION_COMPLETED',
+  REPORT_NOT_RECEIVED: 'REPORT_NOT_RECEIVED',
+};
+
+const logWarningMessage = (status: ReportStatus) => console.error(`The status '${status}' is not supported by '/v1/utilisation-reports/set-status'`);
+
+const setReportStatusByReportId = (id: string, status: ReportStatus, collection: Collection): Promise<UpdateResult | void> => {
```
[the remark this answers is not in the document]
JoshBinns2000: My reason for doing it this way was that I didn't want to make the `db.getCollection` call each time I processed an element of `reportsWithStatus`. I suppose I could instead create an array of sanitised update requests and then provide that array to this file so that the collection logic is the only thing handled here

dtfs-central-api/src/services/repositories/utilisation-report-status-repo.ts
JoshBinns2000: I discussed this with Christian earlier and he did say that we should get the bank name from the database in this case

dtfs-central-api/src/types/utilisation-report-status.ts
```diff
@@ -0,0 +1,46 @@
+import { ObjectId } from 'mongodb';
+import { TFMUser } from './users';
+
+export type ReportStatus = 'RECONCILIATION_COMPLETED' | 'REPORT_NOT_RECEIVED';
```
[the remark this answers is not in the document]
JoshBinns2000: I only added these statuses for now as they were the only ones I needed and wanted to leave it as something to add in future as and when we need the extra statuses (so that someone doesn't come and look at this PR in the future and wonder why I added more types which I didn't need at that point). I'm not sure whether or not that logic is the best practice though, so I am happy to add the rest of the statuses

dtfs-central-api/src/types/utilisation-report-status.ts
```diff
@@ -0,0 +1,46 @@
...
+};
+
+type ReportWithStatus = {
+  status: ReportStatus;
+  report: ReportDetails | ReportId;
+};
+
+export type PutReportStatusRequestBody = {
+  user: TFMUser;
+  reportsWithStatus: ReportWithStatus[];
```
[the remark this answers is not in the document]
JoshBinns2000: I didn't know what the best practice was with respect to defining types inside specific `.ts` files (for example, I think each `external-api` file has its own `interfaces` directory rather than a set of shared types). If we are happy to have local types in cases where the type is only needed for that one file then I'm happy to do that

dtfs-central-api/src/v1/controllers/utilisation-report-service/tfm-put-utilisation-report-status.controller.ts
JoshBinns2000: Yeh that was what I thought as well - I already made that change in a local commit

dtfs-central-api/src/v1/routes/tfm-routes.js
```diff
@@ -722,4 +724,33 @@ tfmRouter.route('/users/team/:teamId')
...
+ *       204:
+ *         description: OK
+ *       400:
+ *         description: Bad request
+ */
+tfmRouter
+  .route('/utilisation-reports/set-status')
+  .put(
+    tfmPutUtilisationReportStatusController.putUtilisationReportStatus,
+  );
```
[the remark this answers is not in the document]
JoshBinns2000: I believe this should be specific to `tfm` as there isn't a non-`tfm` case where this endpoint should be accessed (that was my logic behind it, anyway). With respect to the other `utilisation-reports` routes, these can be accessed by both `portal` and `tfm` so they should be left as general (such as downloading the report, for instance)

dtfs-central-api/src/types/utilisation-report-status.ts
toddym42: Edit: re slack discussion about DB types, we should have this type represent the DB collection type, so be in `src/types/db-models/utilisation-reports.d.ts`

dtfs-central-api/src/types/tfm/teams.ts
```diff
@@ -0,0 +1,34 @@
+export type Team = { email: string } & (
+  | {
+      id: 'UNDERWRITER_SUPPORT';
+      name: 'Underwriter support';
```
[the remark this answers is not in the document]
toddym42: True, but in terms of the type we don't really care what the name is, just that it's a string. There is no benefit to including these in the type as there is no logic based off of it (at least there shouldn't be), so we don't care if they change. The id we DO care about because we base logic of it and need to know if it changes because it will break things.

dtfs-central-api/api-tests/v1/utilisation-report/tfm-put-utilisation-report-status.api-test.js
```diff
@@ -0,0 +1,243 @@
...
+const db = require('../../../src/drivers/db-client');
+const { mockUtilisationReports } = require('../../mocks/tfm-utilisation-reports/mock-utilisation-reports');
+const { DB_COLLECTIONS } = require('../../../src/constants/dbCollections');
+
+console.error = jest.fn();
+
+const ReportStatus = {
+  REPORT_NOT_RECEIVED: 'REPORT_NOT_RECEIVED',
+  RECONCILIATION_COMPLETED: 'RECONCILIATION_COMPLETED',
+};
```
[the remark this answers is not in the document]
toddym42: Note I've got this [same constant in one of my PRs](https://github.com/UK-Export-Finance/dtfs2/pull/2366/files#diff-483f4f47a33f35492bcfa1e1c50e6a2698bcd6df0499e2eebdc12da333849239), so we should make sure we're the same so we don't duplicate. I'm guessing my changes will probably go in first, so let me know if you want to me change anything in mine before I merge.

dtfs-central-api/src/v1/routes/tfm-routes.js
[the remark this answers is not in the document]
toddym42: My feeling on this is that the central API should be general, so shouldn't necessarily know what it is being used for. Since utilisation reports are not TFM specific then the route should be too. I realise though that that may not be the logic that has been used consistently in the project though.

@ChristianMcc - would be good to get your thoughts on this one.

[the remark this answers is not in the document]
toddym42: My feeling on this is that the central API should be general, so shouldn't necessarily know what it is being used for. Since utilisation reports are not TFM specific then the route should be too. I realise though that that may not be the logic that has been used consistently in the project though.

@ChristianMcc - would be good to get your thoughts on this one.

dtfs-central-api/src/types/utilisation-report-status.ts
```diff
@@ -0,0 +1,46 @@
...
+    id: string;
+    name?: string;
+  };
+  month: number;
+  year: number;
+  azureFileInfo: {
+    fullPath: string;
+  } | null;
+  dateUploaded: Date;
+  uploadedBy: TFMUser;
```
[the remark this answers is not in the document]
toddym42: Yeah, I'd just defined what we need in the DB, then when we set the values we need to work out how to populate it based on a TFM or portal user.

dtfs-central-api/src/v1/controllers/utilisation-report-service/tfm-put-utilisation-report-status.controller.ts
toddym42: Details sent over slack

dtfs-central-api/src/services/repositories/utilisation-report-status-repo.ts
```diff
@@ -0,0 +1,120 @@
...
+  UtilisationReport,
+} from '../../types/utilisation-report-status';
+import { TFMUser } from '../../types/users';
+
+const REPORT_STATUS: Record<ReportStatus, ReportStatus> = {
+  RECONCILIATION_COMPLETED: 'RECONCILIATION_COMPLETED',
+  REPORT_NOT_RECEIVED: 'REPORT_NOT_RECEIVED',
+};
+
+const logWarningMessage = (status: ReportStatus) => console.error(`The status '${status}' is not supported by '/v1/utilisation-reports/set-status'`);
```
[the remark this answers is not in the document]
JoshBinns2000: My guess is that it will need to set the status to more general statuses down the line, hence the general naming, but for now it only supports two of the statuses (it could easily support the other statuses though)

dtfs-central-api/src/services/repositories/utilisation-report-status-repo.ts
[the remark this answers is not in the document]
JoshBinns2000: I do think this will require more discussion as the general repo file doesn't help me too much here (primarily because it is written in js and I don't want to start rewriting the existing code)

dtfs-central-api/src/types/users.ts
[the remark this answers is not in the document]
JoshBinns2000: It won't only be for TFM - I expect it to also include things like `PortalUsers` and `GefUser`

dtfs-central-api/api-tests/mocks/tfm-utilisation-reports/mock-utilisation-reports.ts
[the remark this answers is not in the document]
JoshBinns2000: File has been deleted

dtfs-central-api/src/constants/dbCollections.js
```diff
@@ -2,6 +2,7 @@ const DB_COLLECTIONS = {
   BANKS: 'banks',
   UTILISATION_REPORTS: 'utilisationReports',
   UTILISATION_DATA: 'utilisationData',
+  BANKS: 'banks',
```
JoshBinns2000, line 5: This got added as part of the rebase

dtfs-central-api/src/types/utilisation-reports.d.ts
```diff
@@ -8,12 +8,36 @@ export type UtilisationReportReconciliationStatus =
 
 export type UtilisationReportReconciliationSummaryItem = {
```
toddym42, line 9: I think this type may have had most of its properties removed in a merge? It should be of type:
```
export type UtilisationReportReconciliationSummaryItem = {
  reportId?: ObjectId;
  bank: {
    id: string;
    name: string;
  };
  status: UtilisationReportReconciliationStatus;
  dateUploaded?: Date;
  totalFacilitiesReported?: number;
  facilitiesLeftToReconcile?: number;
};
```

dtfs-central-api/src/types/users.ts
```diff
@@ -0,0 +1,10 @@
+import { Team } from './tfm/teams';
+
+export type TfmUser = {
```
toddym42, line 3: I _think_ this represents something from the `tfm-users` collection? If so we should put it in `db-model/tfm-users.d.ts` and add `WithId`.

dtfs-central-api/src/v1/controllers/utilisation-report-service/tfm-put-utilisation-report-status.controller.ts
```diff
@@ -0,0 +1,78 @@
...
+    const session = client.startSession();
+    await session.withTransaction(async () => {
+      await updateManyUtilisationReportStatuses(updateInstructions, uploadedByUserDetails);
+    });
+    await session.endSession();
+
+    return res.sendStatus(200);
+  } catch (error) {
+    console.error('Error updating utilisation report status:', error);
+    return res.status(400).send({ error: 'Update utilisation report status request failed' });
```
toddym42, line 76: 400 implies a client error, which I don't think we can guarantee here. I think we should use `500` in the default case. Use of `400` elsewhere seems to be where a payload issue has been identified.

it's the creation of `uploadedByUserDetails` and `updateInstructions` that will throw errors based on an invalid payload, so we should wrap those and return a 400 in that case.

dtfs-central-api/src/v1/routes/tfm-routes.js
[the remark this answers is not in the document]
ChristianMcc: I'd go with the utilisation-reports route partly just because it keeps all this logic together in one file and I think that's quite useful when it's a contained feature. And it lines up with me agreeing that central API should be more general than catering exactly to it's consumers

toddym42 (commented):
Nice work getting some files into TypeScript. Just a few more thoughts, and looks like there are some new lint errors that need to be addressed too.

dtfs-central-api/src/services/repositories/utilisation-reports-repo/update-report-status.ts
```diff
@@ -0,0 +1,96 @@
...
+  ReportFilterWithReportId,
+  UpdateUtilisationReportStatusInstructions,
+} from '../../../types/utilisation-reports';
+import { UploadedByUserDetails, UtilisationReport } from '../../../types/db-models/utilisation-reports';
+import { UTILISATION_REPORT_RECONCILIATION_STATUS } from '../../../constants';
+import { DB_COLLECTIONS } from '../../../constants/dbCollections';
+import db from '../../../drivers/db-client';
+import { getBankNameById } from '../banks-repo';
+
+export const setReportAsCompleted = (utilisationReportsCollection: Collection, filter: ReportFilterWithReportId) =>
```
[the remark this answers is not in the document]
JoshBinns2000: I just decided to remove the exports and update the tests to only test the one exported function

dtfs-central-api/.eslintrc.js
```diff
@@ -1,6 +1,6 @@
 module.exports = {
   plugins: ['@typescript-eslint'],
-  extends: ['airbnb-base', 'prettier'],
```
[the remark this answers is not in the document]
JoshBinns2000: It was not intentional

dtfs-central-api/api-test.jest.config.js
```diff
@@ -1,9 +1,10 @@
 const commonSettings = require('./api-test-common.jest.config');
 
 module.exports = {
+  preset: 'ts-jest',
   globalTeardown: './test-teardown-globals.js',
   collectCoverageFrom: ['src/**/*.{js,}'],
```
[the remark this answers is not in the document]
JoshBinns2000: Happy to update it to this, but what is the benefit is to including `.ts` files in `collectCoverageFrom`?

dtfs-central-api/src/constants/tfmTeams.ts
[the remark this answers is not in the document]
JoshBinns2000: I think I just included it as something I'm expecting to need (especially seeing as I defined the `Team` type and the `TfmUser` type which includes `teams: Team[]`)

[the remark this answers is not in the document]
JoshBinns2000: Also, with respect to file names, every file in constants uses camel case except for the `utilisation-report-reconciliation-status` file - I don't understand why we now want to use kebab case for these files? Unless your logic is based on the files around the rest of `dtfs-central-api`, at which point I am happy to agree

dtfs-central-api/src/types/utilisation-reports.d.ts
JoshBinns2000: Yeh I'm not sure how that happened - oh well!

dtfs-central-api/src/types/db-models/utilisation-reports.d.ts
```diff
@@ -0,0 +1,31 @@
...
+export type AzureFileInfo = {
+  folder: string;
+  fullPath: string;
+  filename: string;
+  url: string;
+  mimetype: string;
+};
+
+export type UploadedByUserDetails = {
+  _id?: string;
```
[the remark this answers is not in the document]
JoshBinns2000: I put it as optional for now as I couldn't remember whether or not we stored the user id in the report, but I'm pretty sure we do so I will remove this and update the controller to extract the id

dtfs-central-api/api-test.jest.config.js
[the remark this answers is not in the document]
toddym42: This is used to calculate the percentage of code that is covered by tests (see [here](https://jestjs.io/docs/configuration#collectcoveragefrom-array)). In some projects there will be a percentage requirement (e.g. at least 80% of code must be covered by tests). In this project it looks like it is just used for information purposes.

dtfs-central-api/src/constants/tfmTeams.ts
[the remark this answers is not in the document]
toddym42: If it's not used we should remove it. We have the details in this PR now, so we can easily copy from here if we need to add it back in in future.

dtfs-central-api/src/v1/routes/tfm-routes.js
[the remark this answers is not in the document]
JoshBinns2000: Sounds good to me then

dtfs-central-api/src/services/repositories/utilisation-reports-repo/update-report-status.ts
```diff
@@ -0,0 +1,96 @@
...
+      $setOnInsert: placeholderUtilisationReport,
+    },
+    { upsert: true },
+  );
+};
+
+export const setToNotReceivedOrDeleteReport = async (utilisationReportsCollection: Collection, filter: ReportFilter): Promise<UpdateResult | DeleteResult> => {
+  const report = await utilisationReportsCollection.findOne(filter);
+  if (!report) {
+    throw new Error("Cannot set report to 'NOT_RECEIVED': report does not exist");
```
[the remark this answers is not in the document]
JoshBinns2000: Tests have been refactored so this should be fixed now

dtfs-central-api/src/constants/dbCollections.js
```diff
@@ -22,8 +22,8 @@ const DB_COLLECTIONS = {
   TFM_USERS: 'tfm-users',
   TRANSACTIONS: 'transactions',
   USERS: 'users',
-  UTILISATION_REPORTS: 'utilisationReports',
   UTILISATION_DATA: 'utilisationData',
+  UTILISATION_REPORTS: 'utilisationReports',
```
JoshBinns2000, line 26: I moved this so that it is in alphabetical order

dtfs-central-api/src/types/db-models/tfm-user.ts
```diff
@@ -0,0 +1,11 @@
...
+import { Team } from '../tfm/teams';
+
+export type TfmUser = WithId<{
+  username: string;
+  email: string;
+  teams: Team[];
+  timezone: string;
+  firstName: string;
+  lastName: string;
+}>;
```
toddym42, line 11: As discussed, the type of this user for incoming requests from TFM is slightly different to what we have in the DB. The DB has these fields:
image
but the incoming user from TFM has a subset of these, based on the below function in the TFM API
```
const mapUserData = (user) => ({
  username: user.username,
  firstName: user.firstName,
  lastName: user.lastName,
  email: user.email,
  timezone: user.timezone,
  teams: user.teams,
  lastLogin: user.lastLogin,
  status: user.status,
  _id: user._id,
});
```
so we should have a type outside of db-models to represent this.

Note in this type I think `lastLogin` is of type `number` and `_id` is of type `string`.

dtfs-central-api/src/v1/controllers/utilisation-report-service/put-utilisation-report-status.controller.ts
```diff
@@ -0,0 +1,97 @@
...
+};
+
+class PayloadError extends Error {
+  status: number;
+
+  constructor(message: string) {
+    super(message);
+    this.status = 400;
+  }
+}
```
toddym42, line 28: Nice use of a custom error here : )

I think we may also need to add
```
this.name = this.constructor.name;
```
in the constructor (see other examples in Portal API).

It might also be nice to make this shared so we can use this in other places where there are payload issues. If so I would put this in `src/v1/errors` (to match what we've done in other projects like Portal API).

dtfs-central-api/src/v1/routes/utilisation-reports-routes.js
```diff
@@ -128,4 +129,37 @@ utilisationReportsRouter
...
+ *                     - $ref: '#/definitions/UtilisationReportStatusWithReportId'
+ *                     - $ref: '#/definitions/UtilisationReportStatusWithBankId'
+ *               user:
+ *                 $ref: '#/definitions/TFMUser'
+ *     responses:
+ *       200:
+ *         description: OK
+ *       400:
+ *         description: Bad request
+ */
```
toddym42, line 158: Should add the default 500 error here too.

dtfs-central-api/src/v1/controllers/utilisation-report-service/put-utilisation-report-status.controller.ts
```diff
@@ -0,0 +1,97 @@
...
+    await session.withTransaction(async () => {
+      await updateManyUtilisationReportStatuses(updateInstructions, uploadedByUserDetails);
+    });
+    await session.endSession();
+
+    return res.sendStatus(200);
+  } catch (error: any) {
+    console.error('Error updating utilisation report status:', error);
+    if (error instanceof PayloadError) {
+      return res.status(error.status).send({ error: 'Update utilisation report status request failed' });
```
toddym42, line 93: As we now know this is a `PayloadError` I think it's safe to return the `error.message` in the request. e.g.
```
return res.status(error.status).send({ error: error.message });
```

dtfs-central-api/src/types/db-models/tfm-user.ts
JoshBinns2000: Discussed: we are going to keep the `tfm-user.d.ts` `db-model` type, but define a new `tfm-session-user.d.ts` type which omits certain properties from the DB model type

toddym42 (approved):

PR #2369, opened 2023-12-01
feat(FN-1501): add new collections and fileshare to bicep infra

francescastocco:
## Introduction
- In E2ETF we have created some new collections and a file share which need to be reflected in the Bicep code

### DIFF
--- a/infrastructure/resource_group_level/modules/storage.bicep
+++ b/infrastructure/resource_group_level/modules/storage.bicep
@@ -136,7 +136,7 @@ resource defaultBlobService 'Microsoft.Storage/storageAccounts/blobServices@2022
 resource defaultFileService 'Microsoft.Storage/storageAccounts/fileServices@2022-09-01' = {
   parent: storageAccount
   name: 'default'
-  // TODO:FN-693 Note that the extant envrionments don't have
+  // TODO:FN-693 Note that the extant environments don't have
   // 7 day soft deletes enabled. We may want to enable this functionality.
   properties: {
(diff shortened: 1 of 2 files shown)


infrastructure/resource_group_level/modules/storage.bicep
```diff
@@ -189,6 +189,16 @@ resource portalFileShare 'Microsoft.Storage/storageAccounts/fileServices/shares@
   }
 }
 
+resource utilisationReportsFileShare 'Microsoft.Storage/storageAccounts/fileServices/shares@2022-09-01' = {
+  parent: defaultFileService
+  name: 'utilisation-reports'
+  properties: {
+    accessTier: 'TransactionOptimized'
```
OxonianCambion, line 196: Any particular reason why you are going with TransactionOptimised rather than Hot?
https://learn.microsoft.com/en-us/azure/storage/files/storage-files-planning#storage-tiers

I doubt it makes much difference!

francescastocco: No reason, other than that being used for the other file shares

OxonianCambion (approved):

infrastructure/resource_group_level/modules/storage.bicep
OxonianCambion: It might be worth a more general tech-debt ticket.
How often are the files accessed? https://learn.microsoft.com/en-us/azure/storage/files/understanding-billing#differences-in-standard-tiers

PR #2420, opened 2023-12-04
feat(FN-1501): Update deployment scripts

francescastocco:
## Introduction :pencil2:
Update deployment scripts to include E2ETF infra changes

- Added environment variables & secrets
- Added new DB collections
- Added new file share
- Created subnet for ClamAv App Container
- Created ClamAv App Container

- Tested by manually running commands in Azure CLI and switching to ClamAv App Container in feature env


.env.sample
```diff
@@ -38,6 +38,7 @@ UKEF_TFM_API_REPORTS_KEY=
 SESSION_SECRET=
 
 # UTILISATION REPORT
+MAX_UTILISATION_REPORT_FILE_SIZE=1048576
```
AlexBramhill, line 41: Could we update this to have the unit in the var name please?

francescastocco: Yes, good idea 🧠

.github/workflows/development_infrastructure.yml
```diff
@@ -493,6 +510,12 @@ jobs:
             az cosmosdb mongodb collection show   --account tfs-${{ env.environment }}-mongo --database-name ${{ env.database_name }} --name feedback || \
             az cosmosdb mongodb collection create --account tfs-${{ env.environment }}-mongo --database-name ${{ env.database_name }} --name feedback --shard "_id"
 
+            az cosmosdb mongodb collection show   --account tfs-${{ env.environment }}-mongo --database-name ${{ env.database_name }} --name utilisationData || \
```
ChristianMcc, line 513: Nothing to do with the review just sharing an opinion:
I'd much rather we defined stuff like this (db schema) in our application code on startup of central API or something rather than here in the infra

.github/workflows/development_infrastructure.yml
```diff
@@ -757,6 +786,64 @@ jobs:
...
+      #   uses: azure/cli@v1.0.0
+      #   with:
+      #     inlineScript: |
+      #       az monitor log-analytics workspace create --workspace-name ${{ secrets.RESOURCE_GROUP}}-Logs-Workspace
+      
+      - name: Managed Environment
+        uses: azure/cli@v1.0.0
+        with:
+          inlineScript: |
+            acaClamAvSubnetId=$(az vnet show --name tfs-${{ env.app_service_plan }}-vnet --query "[?name=='${{ env.environment }}-clamav'].id" --output tsv)
```
ChristianMcc, line 819: Ah I guess we should have tested this - should we just test this against the deployed BICEP clamav to see if it returns a sensible looking id?

ChristianMcc (commented):
Nice work! One thing that I'm wondering whether we should test before committing?
Also will chat to you in person re a plan for clamav, roughly I think:
- Deploy to dev
- Use in dev for QA
- Point staging to dev
- -eventually bring down both and get every environment pointing at feature except prod which will have its own one which will be created using these scripts?

I have no idea if we can peer between vnets :)

.github/workflows/development_infrastructure.yml
francescastocco: Good idea, this one didn't work but tried another command which has worked

ChristianMcc (approved):
Nice work :D

PR #2421, opened 2023-12-04
fix(FN-1686): update error copy for large file

ChristianMcc:
## Introduction :pencil2:
Needed to update the copy to match designs when a large file was uploaded

Updated the file too large copy


francescastocco (approved):
This is an amazing change, 10/10

PR #2422, opened 2023-12-05
fix(DTFS-6466): accessibility - added text for screen readers to make input field prefix/suffix known

avaitonis:
## Introduction :pencil2:
Input field in `govukInput` component can have prefix and suffix, but it is not read by screen readers.

Added:
  - `<span class="govuk-visually-hidden"> in ' + currency + '</span>'` 
  - `<span class="govuk-visually-hidden"> in %</span>?'`


lmarrai-sw (approved):
Thanks @avaitonis - do you think it would be worth having a macro for `<span class="govuk-visually-hidden">{{params.text}}</span>` to commonise it so it's easier to change in future?

lmarrai-sw:
Replying to myself
> do you think it would be worth having a macro for `<span class="govuk-visually-hidden">{{params.text}}</span>` to commonise it so it's easier to change in future?

Since other PRs we have up at the moment also add this span, we couldn't commonise them in this PR anyway, so I'm going to merge as-is. It would be nice to do this in future though

PR #2424, opened 2023-12-05
fix: fix about-facility UI test failing after 5/12/2023

lmarrai-sw:
## Introduction :pencil2:
The `controllers/about-facility > Validate About Facility > redirects user to provided facility page if all of method passes test in gef-ui/server/controllers/about-facility/index.test.js` started failing today.

The facility used by the test has `shouldCoverStartOnSubmission = true` so the cover start date is `5/12/2023` today. The cover end date for the facility is hard-coded to `5/12/2023`, and the cover end date has to be strictly after the cover start date, so the validation (and test) fails.

I've changed this test to always use a future date for the cover end date, so the test passes regardless of today's date.

I fixed a typo in one of the variables in the same test file.

### DIFF
--- a/gef-ui/server/controllers/about-facility/index.test.js
+++ b/gef-ui/server/controllers/about-facility/index.test.js
@@ -100,7 +100,7 @@ describe('controllers/about-facility', () => {
     const tomorrow = add(now, { days: 1 });
     const yesterday = sub(now, { days: 1 });
     const threeMonthsAndOneDayFromNow = add(now, { months: 3, days: 1 });
-    const oneDayoneDayLessThanThreeMonthsFromNow = sub(add(now, { months: 3 }), { days: 1 });
+    const oneDayLessThanThreeMonthsFromNow = sub(add(now, { months: 3 }), { days: 1 });
 
     it('redirects user to application page if save and return is set to true', async () => {
       mockRequest.query.saveAndReturn = 'true';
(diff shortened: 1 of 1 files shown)


avaitonis (approved):
LGTM

AlexBramhill (approved):

gef-ui/server/controllers/about-facility/index.test.js
```diff
@@ -671,12 +671,9 @@ describe('controllers/about-facility', () => {
       mockRequest.body.shouldCoverStartOnSubmission = 'true';
       mockRequest.body.hasBeenIssued = 'true';
       mockRequest.body.monthsOfCover = '10';
-      mockRequest.body['cover-start-date-day'] = '01';
```
ChristianMcc, line 674: out of curiosity does cover-start-date get overwritten if we have `shouldCoverStartOnSubmission = true`?

ChristianMcc (approved):
Nice, added question just for my understanding :)

gef-ui/server/controllers/about-facility/index.test.js
lmarrai-sw: Yes, by `portal-api` when it receives the PUT request from `gef-ui` (you can see this in the constructor in `portal-api\src\v1\gef\models\facilities.js`)

PR #2433, opened 2023-12-08
feat(FN-1620): Update csv validation for fees paid currency

ChristianMcc:
## Introduction :pencil2:
We need to add validation for fees paid to UKEF currency and add validation that if one of exchange rate or payment currency are present the other should be present.

Added the new header to validate and validation that it's an ISO4217 currency. Updated name of `exchange rate` to `payment exchange rate`. Added validation to `payment currency` that if `payment exchange rate` is present it also must be present.


ChristianMcc:
Need to fix the E2E tests with some updated valid csvs :)

portal/server/controllers/utilisation-report-service/utilisation-report-upload/utilisation-report-cell-validators/generate-payment-currency-error.js
```diff
@@ -1,16 +1,28 @@
...
+const generatePaymentCurrencyError = (csvDataRow) => {
+  if (
+    !csvDataRow[UTILISATION_REPORT_HEADERS.PAYMENT_EXCHANGE_RATE]?.value && !csvDataRow[UTILISATION_REPORT_HEADERS.PAYMENT_CURRENCY]?.value
+  ) {
     return null;
   }
-  if (!validator.isISO4217(paymentCurrencyObject?.value)) {
+  if (!csvDataRow[UTILISATION_REPORT_HEADERS.PAYMENT_CURRENCY]?.value) {
+    return {
+      errorMessage: 'Payment currency must have an entry',
```
toddym42, line 12: Seeing as the first `if` block returns `null` if both the exchange rate and currency are missing, I guess this is only required when there is an exchange rate? Would it be worth making that clear in the message here? e.g. something like
```
errorMessage: 'Payment currency must have an entry when an exchange rate is supplied',
```
if so, we could make the same change in the in `generatePaymentExchangeRateError`. e.g.
```
errorMessage: 'Payment exchange rate must have an entry when a payment currency is supplied',
```

toddym42 (approved):

portal/server/controllers/utilisation-report-service/utilisation-report-upload/utilisation-report-cell-validators/generate-payment-currency-error.js
ChristianMcc: Yeah I think that is a good idea, I'm just using the copy I'm given but I'll do this and suggest it as a change on the ticket (and revert if I get shot down!)

toddym42 (approved):

PR #2436, opened 2023-12-11
style(FN-1798): standardise prettier config and add to pre-commit commit hook

toddym42:
## Introduction :pencil2:

Prettier is an opinionated code formatter that can help keep a consistent code formatting style in the project. There are currently the following issues with how it is used:

* The prettier package and config have been added in the root project, but some individual sub-projects (external-api and portal-api) have their own version of prettier and config added. This may lead to inconsistencies if the versions and config diverge
* Sometimes running prettier on files results in conflicts with the existing eslint config, so prettier changes need to be reverted to keep the linter happy
* Running of prettier is not enforced in any way (either with commit hooks or CI)

* Remove the prettier package and config from sub-projects so that formatting across the entire codebase will be based on the root directory prettier installation and config. Updated the README to clarify this point.
* Add [eslint-config-prettier](https://github.com/prettier/eslint-config-prettier) package to the elint config in all projects - this resolves any conflicts between eslint and prettier so we can ensure that prettier can always be run
* Add running of prettier to the commit hook so that this is always run on staged files even if the dev hasn't run the formatter manually
* Move the running of lint-staged to a "pre-commit" hook instead of the "commit-msg" hook. This means that any eslint or prettier fixes that get done as part of the commit hook are done before the commit is submitted, meaning they will be included in commit you are submitting. As lint-staged fixes are currently included in the "commit-msg" hook, this means you you would add a commit, then need to add a further separate commit with any eslint or prettier changes, which may lead to these changes not seen and committed. (Further details of the different types of commit hooks [here](https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks))
* Add a prettier check to the CI pipeline to check that all files are formatted according to our configured prettier formatting style


azure-functions/acbs-function/.eslintrc.js
```diff
@@ -1,25 +1,33 @@
 module.exports = {
-  extends: 'airbnb-base',
+  extends: ['airbnb-base', 'prettier'],
```
toddym42, line 2: Adding `'prettier'` here (available from the `eslint-config-prettier` package adding in the root `package.json`) is what resolves any conflicts between eslint and prettier formatting rules.

portal/.eslintrc.js
```diff
@@ -38,13 +49,8 @@ module.exports = {
...
     {
       files: ['*.ts'],
-      extends: [
-        'airbnb-base',
-        'plugin:@typescript-eslint/recommended-type-checked',
-      ],
-      plugins: [
-        '@typescript-eslint',
-      ],
+      extends: ['airbnb-base', 'plugin:@typescript-eslint/recommended-type-checked'],
```
lmarrai-sw, line 52:
```suggestion
      extends: ['airbnb-base', 'plugin:@typescript-eslint/recommended-type-checked', 'prettier'],
```

trade-finance-manager-ui/.eslintrc.js
```diff
@@ -39,13 +50,8 @@ module.exports = {
...
     {
       files: ['*.ts'],
-      extends: [
-        'airbnb-base',
-        'plugin:@typescript-eslint/recommended-type-checked',
-      ],
-      plugins: [
-        '@typescript-eslint',
-      ],
+      extends: ['airbnb-base', 'plugin:@typescript-eslint/recommended-type-checked'],
```
lmarrai-sw, line 53:
```suggestion
      extends: ['airbnb-base', 'plugin:@typescript-eslint/recommended-type-checked', 'prettier'],
```

toddym42:
Closing for now due to test failures and complications this will cause for the `main-stbr` feature branch. Additional eslint config also to be added (to show errors when code is not formatted in prettier style). Possibly look to merge after `main-stbr` has been merged into `main`.

PR #2438, opened 2023-12-11
refactor: move all utilisation report related routes to 'utilisation-…

toddym42:
Two main goals with this PR (a commit for each), both related to changes in the Central API:

* move all utilisation report related routes from `portal-routes.js` to `utilisation-reports-routes.js`
* add type safety to DB collection names (by use of a `DbCollectionName` type, and full use of the `DB_COLLECTION` constant instead of hard-coding collection names)


ChristianMcc (approved):
Nice one, might get a couple of conflicts with this but I think main won't be touching DB interactions much at all before we merge this in! :)

PR #2442, opened 2023-12-12
feat(FN-1043): add all required data to TFM 'Bank reports' table

toddym42:
## Ticket
[FN-1043](https://ukef-dtfs.atlassian.net/browse/FN-1043)

## Introduction :pencil2:
Adds missing details to the TFM "Bank reports" screen:

image

* set status on summary reponse:
  * set as `REPORT_NOT_RECEIVED` when no report in the DB
  * set to the report status when there is a report in the DB
* where there is a report, also calculate and return additonal required fields:
  * `dateUploaded` for the "Date report received" column
  * `totalFacilitiesReports` for the "Total facilities reported" column - this is the sum of all entries in the `utilisationData` table for the report, unique by `facilityId`
  * `facilitiesLeftToReconcile` for the "Facilities left to reconcile" column - this is currently using the same value as `totalFacilitiesReports`. To calculate this properly we need to know the status of the data item, which will come in a later ticket


dtfs-central-api/src/drivers/db-client.js
```diff
@@ -37,9 +39,9 @@ const getClient = async () => {
 module.exports.getClient = getClient;
 
 /**
- * @template TDoc
- * @param {DbCollectionName} collectionName
- * @return {Promise<import('mongodb').Collection<TDoc>>}
+ * @template {DbCollectionName} TCollectionName
+ * @param {TCollectionName} collectionName
+ * @return {Promise<import('mongodb').Collection<DbModel<TCollectionName>>>}
```
toddym42, line 43: This change, using a new `DbModel` I've created, means that when we fetch a document (or documents) from the data base it will use the `collectionName` you specify to work out what the shape of that object will be. e.g. if we do
```
db.getCollection('banks');
```
then any documents we find off the back of any queries on that collection will have the `Bank` type from `types/db-models/banks.d.ts`.

dtfs-central-api/src/types/db-models/utilisation-data.d.ts
```diff
@@ -0,0 +1,35 @@
...
+   * year of the start of the report period
+   * example: 2023
+   */
+  year: number;
+  exporter: string;
+  baseCurrency: string;
+  facilityUtilisation: number;
+  totalFeesAccruedForTheMonth: number;
+  monthlyFeesPaidToUkef: number;
+  paymentCurrency: string;
```
toddym42, line 28: @ChristianMcc - I wasn't confident enough on my understanding of these values to add a description, so could use your expertise here.

trade-finance-manager-ui/server/controllers/utilisation-reports/helpers/reconciliation-summary-helper.js
```diff
@@ -23,6 +25,7 @@ const getReportReconciliationSummaryViewModel = (summaryApiResponse) =>
   summaryApiResponse.map((item) => ({
     ...item,
     displayStatus: reconciliationStatusCodeToDisplayStatus[item.status],
+    formattedDateUploaded: item.dateUploaded ? format(parseISO(item.dateUploaded), 'd MMM yyyy') : undefined,
```
JoshBinns2000, line 28: I think this can be shortened to `item.dateUploaded && format(parseIso(item.dateUploaded), 'd MMM yyyy');`

dtfs-central-api/src/services/repositories/utilisation-reports-repo.js
```diff
@@ -39,26 +39,25 @@
...
+ * @param {number} month - Month of utilisation report reporting period
+ * @param {number} year - Year of utilisation report reporting period
+ * @returns {Promise<UtilisationReport | null>} - Utilisation report details matching the bank/month/year combo or null if it doesn't exist.
  */
-const getUtilisationReportDetailsForMonthAndYear = async (bankId, month, year) => {
+const getUtilisationReportDetailsByBankIdMonthAndYear = async (bankId, month, year) => {
   const utilisationReportDetailsCollection = await db.getCollection(DB_COLLECTIONS.UTILISATION_REPORTS);
-  const matchingReportDetails = await utilisationReportDetailsCollection.findOne({ 'bank.id': bankId, month, year });
-  return matchingReportDetails;
+  return await utilisationReportDetailsCollection.findOne({ 'bank.id': bankId, month, year });
```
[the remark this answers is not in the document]
toddym42: I'm not exactly sure how mongo with handle anything dodgy, but we should be validating these values if they are provided in a request (e.g. by the route express validators or otherwise). For the new use case where I've used this the values are generated internally (see "getReportPeriodStartForSubmissionMonth" method) so that's not an issue.

dtfs-central-api/src/utils/mongodb.js
```diff
@@ -0,0 +1,17 @@
...
+
+/**
+ * Returns a copy of the provided object, but without the MongoDb '_id' property
+ * @template {object} TObject
+ * @param {WithId<TObject>} objectWithId - the object to remove the '_id' from
+ * @returns {WithoutId<TObject>}
+ */
+const withoutMongoId = (objectWithId) => {
+  const { _id, ...rest } = objectWithId;
+  return rest;
```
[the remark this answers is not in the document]
toddym42: I see what you're saying, but this is a pretty standard convention. I've removed this file anyway as I removed it's only use.

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-reports-reconciliation-summary.controller/index.js
```diff
@@ -0,0 +1,93 @@
...
+
+/**
+ * @typedef {import('../../../../types/db-models/banks').Bank} Bank
+ * @typedef {import('../../../../types/date').IsoMonthStamp} IsoMonthStamp
+ * @typedef {import('../../../../types/utilisation-reports').UtilisationReportReconciliationSummaryItem} UtilisationReportReconciliationSummaryItem
+ */
+
+/**
+ * @param {IsoMonthStamp} submissionMonth
+ * @returns {{ month: number, year: number }}
```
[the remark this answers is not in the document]
toddym42: With TypeScript types a `;` would be used, but apparently with JSDoc you should use a `,` (see all the examples [here](https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html)).

This wasn't actually intentional as usually I've been using the TypeScript syntax, but apparently that wasn't right 🤷‍♂️

trade-finance-manager-ui/server/controllers/utilisation-reports/helpers/reconciliation-summary-helper.js
toddym42: I think in most cases this would result in what we want, but it doesn't behave exactly the same in every case. e.g. if `item.dateUploaded` was the empty string then
```
item.dateUploaded && format(parseISO(item.dateUploaded)
```
would result in `''` instead of `undefined`.

dtfs-central-api/src/utils/date.js
```diff
@@ -1,14 +1,26 @@
...
+ */
+const getOneIndexedMonth = (date) => date.getMonth() + 1;
+
 const ISO_MONTH_REGEX = /^\d{4}-\d{2}$/;
 
 /**
  * Checks whether the provided value is an ISO month string in format 'yyyy-MM'
  * @param {unknown} value - the value to test
- * @returns {boolean}
+ * @returns {value is IsoMonthStamp}
```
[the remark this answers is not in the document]
toddym42: This is a [type predicate](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates) and the idea is that it tells the compiler what the type is if the function returns true, without you additionally having to cast the value if the function returns true.

I think though that JSDoc has limited support for it, and maybe it doesn't add the same value for alias (like `type IsoMonthStamp = string;`). I'll revert.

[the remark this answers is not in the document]
JoshBinns2000: Oo I didn't know type predicates could work like that - very nice

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-reports-reconciliation-summary.controller/index.js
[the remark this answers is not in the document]
JoshBinns2000: Oh well - at least we know now!

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-reports-reconciliation-summary.controller.js
```diff
@@ -1,39 +1,87 @@
...
+  };
+};
+
+/**
+ * @param {Bank[]} banks
+ * @param {IsoMonthStamp} submissionMonth
+ * @return {Promise<UtilisationReportReconciliationSummaryItem[]>}
+ */
+const generateReconciliationSummary = async (banks, submissionMonth) =>
+  Promise.all(banks.map(async (bank) => await generateReconciliationSummaryItem(bank, submissionMonth)));
```
JoshBinns2000, line 71: Nice one-liner (and it's always nice to see `Promise.all`)

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-reports-reconciliation-summary.controller.test.js
```diff
@@ -57,6 +64,49 @@ describe('getUtilisationReportsReconciliationSummary', () => {
...
+    const barclaysReport = {
+      ...MOCK_UTILISATION_REPORT,
+      bank: { id: MOCK_BANKS.BARCLAYS.id, name: MOCK_BANKS.BARCLAYS.name },
+      status: UTILISATION_REPORT_RECONCILIATION_STATUS.RECONCILIATION_IN_PROGRESS,
+    };
+    getUtilisationReportDetailsByBankIdMonthAndYear.mockImplementation(async (bankId) => {
+      switch (bankId) {
+        case MOCK_BANKS.BARCLAYS.id:
+          return barclaysReport;
+        case MOCK_BANKS.HSBC.id:
```
JoshBinns2000, line 86: If there is no specific behaviour for the HSBC case, why not just capture that in the `default`?

trade-finance-manager-ui/templates/utilisation-reports/_macros/report-reconciliation-table.njk
```diff
@@ -9,8 +9,8 @@
         <th scope="col" class="govuk-table__header" aria-sort="ascending">Bank</th>
         <th scope="col" class="govuk-table__header" aria-sort="none">Status</th>
         <th scope="col" class="govuk-table__header" aria-sort="none">Date report received</th>
-        <th scope="col" class="govuk-table__header" aria-sort="none">Total facilities reported</th>
-        <th scope="col" class="govuk-table__header" aria-sort="none">Facilities left to reconcile</th>
+        <th scope="col" class="govuk-table__header" aria-sort="none">Total fees reported</th>
+        <th scope="col" class="govuk-table__header" aria-sort="none">Reported fees left to reconcile</th>
```
JoshBinns2000, line 13: Are there any e2e-tests which need to be updated to match the new text in this file?

dtfs-central-api/src/helpers/mongodb.ts
```diff
@@ -0,0 +1,7 @@
+import { Document, WithoutId } from "mongodb";
```
JoshBinns2000, line 1: Should throw a linting error here, but this needs single quotes

dtfs-central-api/src/helpers/validation.ts
JoshBinns2000, line 1: File needs a newline at the end

dtfs-central-api/src/utils/string.ts
```diff
@@ -0,0 +1,5 @@
+export const isString = (value: unknown): value is string => typeof value === 'string' || value instanceof String;
+
+export const isEmptyString = (str: unknown) => !str || (isString(str) && !str.trim().length);
+
+export const hasValue = (str: unknown) => str && !isEmptyString(str);
```
JoshBinns2000, line 5: What do we mean by `hasValue`?

dtfs-central-api/api-tests/v1/utils/string.api-test.js
JoshBinns2000, line 1: Could we change this to a unit test instead of an api test? Normally I would hesitate to do this given that this may conflict with DTFS, but considering you changed the `string.js` file to `string.ts` I think we should also change the test file

dtfs-central-api/src/utils/string.ts
toddym42: Looks like it means defined and not empty. This is just refactored to TS from existing code, so isn't something I'd want to spend too long refactoring further now.

dtfs-central-api/api-tests/v1/utils/string.api-test.js
JoshBinns2000: Discussed - as we would end up refactoring the entire file to be typescript, it's not worth changing the file at this stage

JoshBinns2000 (approved):
The tests which are failing don't appear to be linked to this PR - happy to merge :)

PR #2444, opened 2023-12-13
feat(DTFS2-6863): added table headers for amendment tasks

vamshidharukef:
## Introduction :pencil2:
Table headers missing when facility is amended. Screen readers will not know that the column headings from the first table are intended to apply to the later tables as well, so screen readers will not be able to tell screen reader users what the columns mean for any table apart from the first one.

Add table headers for each section when facility is amended.


avaitonis (approved):
LGTM - component test could be improved, but this is not required for me

PR #2445, opened 2023-12-13
feat(FN-1606): change encoding from utf-8 to latin1

JoshBinns2000:
## Introduction :pencil2:
Strange characters can appear when users open a downloaded `.csv` file in excel due to excel using `ANSI` encoding to open the file (the file they download is encoded in `utf-8`).

Changes the file upload* and download to encode the file using `latin1`.

*still need to test whether or not we only need to change the download encoding rather than the upload as it would be better to keep everything in `utf-8` unless it is downloaded


francescastocco (approved):
This is a great change 👍

PR #2446, opened 2023-12-14
fix(DTFS2-6472): improve /deals page sorting

avaitonis:
## Introduction :pencil2:
TFM sortable table headers have up and down arrows. These arrows are read by screen readers and it is not recommended.

At the moment we use multiple 2 sorting mechanisms:
- POST based for /deals
- MOJ javascript element for /facilities

Both pages has accessibility issue, but fixing /facilities will be in separate PR.

  - added <span aria-hidden="true">::after</span> to sort buttons to match https://www.w3.org/WAI/ARIA/apg/patterns/table/examples/sortable-table/ and DAC report requirement.
  - extra - enabled sorting on `Date received` column . This allows display of default ordering
  - extra - fixed sort arrow styling to match MOJ
  - fixed aria-sort to have none if sorting is not active for the column
  - added table caption for screen readers - caption exist in W3 example, but needs review by our UX team.


vamshidharukef (approved):
LGTM

abhi-markan (approved):

PR #2447, opened 2023-12-14
fix(DTFS2-6876): fix order dependency in Portal login journey E2E tests

oscar-richardson-softwire:
## Introduction :pencil2:

Currently, `resend-sign-in-link.spec.js` in the Portal login journey E2E tests does not reset the user status and sign in link send count after running all of the tests. This creates an order dependency, whereby running `resend-sign-in-link.spec.js` first and then `login.spec.js` will cause the later to fail because the `BANK1_MAKER1` user is still blocked as a result of the first test suite.

- added an `after` block in `resend-sign-in-link.spec.js` to reset the `BANK1_MAKER1` after all the tests have run.

### DIFF
--- a/e2e-tests/portal/cypress/e2e/journeys/login/resend-sign-in-link.spec.js
+++ b/e2e-tests/portal/cypress/e2e/journeys/login/resend-sign-in-link.spec.js
@@ -3,6 +3,7 @@ const relative = require('../../relativeURL');
 const MOCK_USERS = require('../../../../../e2e-fixtures');
 
 const { BANK1_MAKER1 } = MOCK_USERS;
+const { username } = BANK1_MAKER1;
 const FIRST_SIGN_IN_TOKEN = '6569ca7a6fd828f925e07c6e';
 
 const userAnonymisedEmailAddress = 'm***1@ukexportfinance.gov.uk';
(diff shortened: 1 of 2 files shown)


e2e-tests/portal/cypress/e2e/journeys/login/resend-sign-in-link.spec.js
```diff
@@ -17,14 +18,17 @@ context('Resending sign in links', () => {
...
       cy.getUserByUsername(username).then(({ _id }) => {
         bank1Maker1Id = _id;
       });
       cy.resetPortalUserStatusAndNumberOfSignInLinks(username);
       cy.enterUsernameAndPassword(BANK1_MAKER1);
     });
 
+    after(() => {
+      cy.resetPortalUserStatusAndNumberOfSignInLinks(username);
+    });
```
lmarrai-sw, line 30: Cypress recommends in their best practice doc that we shouldn't really use an `after` block, and instead we should have a `before` block in `login.spec.js`: https://docs.cypress.io/guides/references/best-practices#Using-after-Or-afterEach-Hooks

lmarrai-sw (approved):

PR #2449, opened 2023-12-15
fix(DTFS2-6672): check user authorisation when updating details

MarRobSoftwire:
## Introduction :pencil2:
In `portal-api`,  `PUT /v1/users/:_id` does not check that the logged in user has the authorisation to update the details. Whilst the end-users shouldn't have direct access to the api this should still be added
* Check if the user has role `ADMIN` or is updating their own password
* Added api & unit tests

### DIFF
--- a/portal-api/api-tests/v1/users/users.api-test.js
+++ b/portal-api/api-tests/v1/users/users.api-test.js
@@ -1,5 +1,5 @@
 const databaseHelper = require('../../database-helper');
-const { setUpApiTestUser } = require('../../api-test-users');
+const testUserCache = require('../../api-test-users');
 
 const app = require('../../../src/createApp');
 const { as } = require('../../api')(app);
@@ -9,6 +9,8 @@ const { READ_ONLY, MAKER, CHECKER } = require('../../../src/v1/roles/roles');
 const { NON_READ_ONLY_ROLES } = require('../../../test-helpers/common-role-lists');
 const { LOGIN_STATUSES } = require('../../../src/constants');
 const { createPartiallyLoggedInUserSession, createLoggedInUserSession } = require('../../../test-helpers/api-test-helpers/database/user-repository');
+const { ADMIN } = require('../../../src/v1/roles/roles');
+const { STATUS } = require('../../../src/constants/user');
 
 const aMaker = users.find((user) => user.username === 'MAKER');
 const MOCK_USER = { ...aMaker, username: 'TEMPORARY_USER' };
@@ -20,11 +22,15 @@ const EMAIL_ERROR = { text: 'Enter an email address in the correct format, for e
 const READ_ONLY_ROLE_EXCLUSIVE_ERROR = { text: "You cannot combine 'Read-only' with any of the other roles" };
 
 describe('a user', () => {
-  let loggedInUser;
+  let aNonAdmin;
+  let anAdmin;
 
   beforeAll(async () => {
     await databaseHelper.wipe(['users']);
-    loggedInUser = await setUpApiTestUser(as);
+    const testUsers = await testUserCache.initialise(app);
+    anAdmin = testUsers().withRole(ADMIN).one();
+    aNonAdmin = testUsers().withoutRole(ADMIN).one();
+    // aNonAdmin = await setUpApiTestUser(as);
   });
 
   beforeEach(async () => {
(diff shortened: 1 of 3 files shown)


portal-api/src/v1/users/routes.js
```diff
@@ -154,32 +155,36 @@ module.exports.findById = (req, res, next) => {
...
+        }
       } else {
-        update(req.params._id, req.body, (updateErr, updatedUser) => {
-          if (updateErr) {
-            next(updateErr);
-          } else {
-            res.status(200).json(sanitizeUser(updatedUser));
-          }
-        });
+        res.status(404).json({});
```
MarRobSoftwire, line 182: This was a 200, but think 404 fits better

portal-api/api-tests/v1/users/users.api-test.js
```diff
@@ -195,75 +196,150 @@ describe('a user', () => {
...
+        const createdUser = response.body.user;
 
-      await as(loggedInUser).put(updatedUserCredentials).to(`/v1/users/${createdUser._id}`);
+        const updatedUserCredentials = {
+          password: 'AbC1234!',
+          passwordConfirm: 'AbC1234!'
+        };
 
-      const { status, body } = await as(loggedInUser).get(`/v1/users/${createdUser._id}`);
+        const { status } = await as(anAdmin).put(updatedUserCredentials).to(`/v1/users/${createdUser._id}`);
```
MarRobSoftwire, line 242: This doesn't verify that the password has been updated, should I add something like
``` 
const { status } = await as().post({ username, password }).to('/v1/login');
expect(status).toEqual(200);
```

portal-api/src/v1/users/routes.js
```diff
@@ -154,32 +155,67 @@ module.exports.findById = (req, res, next) => {
...
-      const errors = applyUpdateRules(user, req.body);
-      if (errors.length) {
-        res.status(400).json({
-          success: false,
-          errors: {
-            count: errors.length,
-            errorList: combineErrors(errors),
-          },
-        });
+  if (req.user.roles.includes(ADMIN)) {
```
JoshBinns2000, line 158: Rather than performing this check in the controller, there is a middleware that can perform this check called `validateUserHasAtLeastOneAllowedRole` in `src/v1/roles`. Your `src/v1/routes.js` file would then have a section like
```
authRouter
  .route('/users/:_id')
  .get(validateUserHasAtLeastOneAllowedRole({ allowedRoles: [ADMIN] }), users.findById)
  ....
```
with similar implementations for the `put` and `delete` functions.

portal-api/src/v1/users/routes.js
```diff
@@ -154,32 +155,67 @@ module.exports.findById = (req, res, next) => {
...
+    findOne(req.params._id, (error, user) => {
+      if (error) {
+        next(error);
```
JoshBinns2000, line 161: Unless you are expecting to call something after `next(error)` from this controller, I would change this to `return next(error);` and then change the `else if` to a plain new `if` below

portal-api/src/v1/users/routes.js
```diff
@@ -154,32 +155,67 @@ module.exports.findById = (req, res, next) => {
...
+      } else if (user) {
```
JoshBinns2000, line 162: This could be changed to a simple guard like `if (!user) { [do something] }` to avoid the nesting below

portal-api/src/v1/users/routes.js
```diff
@@ -154,32 +155,67 @@ module.exports.findById = (req, res, next) => {
...
+        const errors = applyUpdateRules(user, req.body);
+        if (errors.length) {
+          res.status(400).json({
```
JoshBinns2000, line 165: As above, it's better practice to do `return res.status(400)...` etc so that you can avoid nesting. It also ensures that any code below this point will not be executed if this statement is reached, whilst the current implementation implies that it is possible to run the code below your `res.status`

portal-api/src/v1/users/routes.js
JoshBinns2000, line 1: The amount of nesting in the changes to this file makes it quite hard to read. I've made suggestions on how you can make better use of guards and make sure you take advantage of `return` statements to make the code easier to read - feel free to drop me a message to discuss if you have any questions

portal-api/api-tests/v1/users/users.api-test.js
JoshBinns2000: I think performing that test is slightly too granular for an api test as it should only be possible to get a 200 status code if a user has successfully been updated, so you can make the assumption that a user was updated if a 200 is received - the actual update functioning is something I'd expect to check in an e2e test. Not 100% sure what the best practice is though

portal-api/src/v1/users/routes.js
MarRobSoftwire: This wouldn't work because the endpoint is used by non-admins when changing their password.
It could be refactored to separate these use cases though

portal-api/src/v1/users/routes.js
```diff
@@ -154,31 +155,38 @@ module.exports.findById = (req, res, next) => {
 };
 
 module.exports.updateById = (req, res, next) => {
-  findOne(req.params._id, (error, user) => {
+  const userIsAdmin = req.user.roles.includes(ADMIN);
```
ChristianMcc, line 158: Can we add optional chaining here (req.user?.roles?.includes(ADMIN)) and the line below? This will stop exceptions being thrown is any of these are undefined in the request.

Likewise is it worth wrapping this in a try catch block and returning a 500 if any exceptions are thrown? (just makes it easier to debug if we catch errors, log them out then return a 500 to the user so that they have a sensible status code sent back to them)

portal-api/src/v1/users/routes.js
```diff
@@ -154,32 +155,44 @@ module.exports.findById = (req, res, next) => {
...
+    );
+    if (!userIsAdmin && !userIsChangingTheirOwnPassword) {
+      return res.status(403).send();
+    }
+    return findOne(req.params._id, (error, user) => {
+      if (error) {
+        return next(error);
+      }
+      if (!user) {
+        return res.status(404).send();
```
ChristianMcc, line 172: Could we add some logging here of `console.error(`Error updating user: could not find user for id :${req.params.id}`)` or something here? Sorry I'm just getting you to improve existing code that aren't your changes :D

portal-api/src/v1/users/routes.js
```diff
@@ -154,32 +155,44 @@ module.exports.findById = (req, res, next) => {
...
+
+      return update(req.params._id, req.body, (updateErr, updatedUser) => {
+        if (updateErr) {
+          return next(updateErr);
+        }
+        return res.status(200).json(sanitizeUser(updatedUser));
+      });
+    });
+  } catch (e) {
+    console.error(e);
```
ChristianMcc, line 193: Can we do console.error('Error updating user', e) please? makes it easier to read in logs

ChristianMcc (commented):
Sorry little bit of logging to add, rest of the implementation is good though!

ChristianMcc (approved):
Nice one :)

PR #2450, opened 2023-12-15
fix: fix all lint errors

toddym42:
## Introduction :pencil2:
Resolves all lint errors that were surfaced after a merge from `main`.


ChristianMcc (approved):
Thanks for doing this :)

PR #2453, opened 2023-12-18
feat(FN-1613): individual bank report page

JoshBinns2000:
## Notes
Though the current target branch is `main-stbr-tfm-5`, this should be merged into a new feature branch for the `TFM-6` realease.

## Introduction :pencil2:
We want to allow `PDC_READ` and `PDC_RECONCILE` users to see the report data for a specific bank by clicking on that bank name. We want users to be routed to a page containing 4 tabs which are currently unpopulated. The page should have two headings - one for the bank name and one for the current report period.

EDIT: This PR now also addresses the changes required for FN-1603 (where a link to the bank's utilisation report page is only enabled if the bank report status is not `REPORT_NOT_RECEIVED` ie. if a report exists for that bank for the current report period). A screenshot has been added below to show that the link is only enabled if the report status is not `REPORT_NOT_RECEIVED`.

To TFM-UI:
- Adds a new `getUtilisationReportByBankId` controller
- Adds a new `/utilisation-reports/:submissionMonth/bank/:bankId` route
- Adds a new template
- Adds components tests

FN-1613 (cropped)

FN-1603 (cropped)

### DIFF
--- a/trade-finance-manager-ui/server/controllers/utilisation-reports/index.test.ts
+++ b/trade-finance-manager-ui/server/controllers/utilisation-reports/index.test.ts
@@ -5,8 +5,10 @@ import { MOCK_BANK_HOLIDAYS } from '../../test-mocks/mock-bank-holidays';
 import { MOCK_UTILISATION_REPORT_RECONCILIATION_SUMMARY } from '../../test-mocks/mock-utilisation-report-reconciliation-summary';
 import { MOCK_TFM_SESSION_USER } from '../../test-mocks/mock-tfm-session-user';
 import { getReportReconciliationSummariesViewModel } from './helpers';
+import { PRIMARY_NAVIGATION_KEYS } from '../../constants';
 
 jest.mock('../../api');
+jest.mock('express-validator');
 
 console.error = jest.fn();
 
@@ -34,9 +36,8 @@ describe('controllers/utilisation-reports', () => {
       await getUtilisationReports(req, res);
 
       // Assert
-      /* eslint-disable no-underscore-dangle */
+      // eslint-disable-next-line no-underscore-dangle
       expect(res._getRenderView()).toEqual('_partials/problem-with-service.njk');
-      /* eslint-enable no-underscore-dangle */
     });
 
     it('renders the utilisation-reports.njk view with required data', async () => {
@@ -63,7 +64,7 @@ describe('controllers/utilisation-reports', () => {
       /* eslint-disable no-underscore-dangle */
       expect(res._getRenderView()).toEqual('utilisation-reports/utilisation-reports.njk');
       expect(res._getRenderData()).toMatchObject({
-        activePrimaryNavigation: 'utilisation reports',
+        activePrimaryNavigation: PRIMARY_NAVIGATION_KEYS.UTILISATION_REPORTS,
         reportPeriodSummaries: expectedViewModel,
       });
       /* eslint-enable no-underscore-dangle */
--- a/trade-finance-manager-ui/server/controllers/utilisation-reports/utilisation-report-for-bank/index.test.ts
+++ b/trade-finance-manager-ui/server/controllers/utilisation-reports/utilisation-report-for-bank/index.test.ts
@@ -1,0 +1,84 @@
+import httpMocks from 'node-mocks-http';
+import api from '../../../api';
+import { getUtilisationReportByBankId } from '.';
+import { MOCK_TFM_SESSION_USER } from '../../../test-mocks/mock-tfm-session-user';
+import { MOCK_UTILISATION_REPORT_RECONCILIATION_SUMMARY } from '../../../test-mocks/mock-utilisation-report-reconciliation-summary';
+import { PRIMARY_NAVIGATION_KEYS } from '../../../constants';
+import { getIsoMonth } from '../../../helpers/date';
+
+jest.mock('../../../api');
+jest.mock('../../../helpers/date');
+
+console.error = jest.fn();
+
+describe('controllers/utilisation-reports/utilisation-report-for-bank', () => {
+  afterAll(() => {
+    jest.resetAllMocks();
+  });
+
+  describe('getUtilisationReportByBankId', () => {
+    const session = {
+      user: MOCK_TFM_SESSION_USER,
+      userToken: 'user-token',
+    };
+
+    it("redirects to '/not-found' if reconciliation summary does not include the current submission month", async () => {
+      // Arrange
+      const bankId = '123';
+      const { req, res } = httpMocks.createMocks({
+        session, params: { bankId },
+      });
+
+      jest.mocked(api.getUtilisationReportsReconciliationSummary).mockResolvedValue([]);
+
+      // Act
+      await getUtilisationReportByBankId(req, res);
+    
+      // Assert
+      // eslint-disable-next-line no-underscore-dangle
+      expect(res._getRedirectUrl()).toEqual('/not-found');
+    });
+    
+    it("redirects to '/not-found' if reconciliation summary does not contain a report for the requested bank", async () => {
+      // Arrange
+      const bankId = '123';
+      const { req, res } = httpMocks.createMocks({
+        session, params: { bankId },
+      });
+
+      jest.mocked(api.getUtilisationReportsReconciliationSummary).mockResolvedValue(MOCK_UTILISATION_REPORT_RECONCILIATION_SUMMARY);
+
+      // Act
+      await getUtilisationReportByBankId(req, res);
+    
+      // Assert
+      // eslint-disable-next-line no-underscore-dangle
+      expect(res._getRedirectUrl()).toEqual('/not-found');
+    });
+    
+    it("renders the 'utilisation-report-for-bank' page with the correct data", async () => {
+      // Arrange
+      const { submissionMonth } = MOCK_UTILISATION_REPORT_RECONCILIATION_SUMMARY[0];
+      const { bank } = MOCK_UTILISATION_REPORT_RECONCILIATION_SUMMARY[0].items[0];
+      const { req, res } = httpMocks.createMocks({
+        session, params: { bankId: bank.id },
+      });
+
+      jest.mocked(getIsoMonth).mockReturnValue(submissionMonth);
+      jest.mocked(api.getUtilisationReportsReconciliationSummary).mockResolvedValue(MOCK_UTILISATION_REPORT_RECONCILIATION_SUMMARY);
+
+      // Act
+      await getUtilisationReportByBankId(req, res);
+    
+      // Assert
+      /* eslint-disable no-underscore-dangle */
+      expect(res._getRenderView()).toEqual('utilisation-reports/utilisation-report-for-bank.njk');
+      expect(res._getRenderData()).toEqual({
+        user: MOCK_TFM_SESSION_USER,
+        activePrimaryNavigation: PRIMARY_NAVIGATION_KEYS.UTILISATION_REPORTS,
+        bank,
+      });
+      /* eslint-enable no-underscore-dangle */
+    });
+  });
+});
--- a/trade-finance-manager-ui/server/controllers/utilisation-reports/utilisation-report-for-bank/index.ts
+++ b/trade-finance-manager-ui/server/controllers/utilisation-reports/utilisation-report-for-bank/index.ts
@@ -1,0 +1,32 @@
+import { Request, Response } from 'express';
+import api from '../../../api';
+import { getIsoMonth } from '../../../helpers/date';
+import { asUserSession } from '../../../helpers/express-session';
+import { PRIMARY_NAVIGATION_KEYS } from '../../../constants';
+
+export const getUtilisationReportByBankId = async (req: Request, res: Response) => {
+  const { userToken, user } = asUserSession(req.session);
+  const { bankId } = req.params;
+
+  try {
+    const currentPeriodSubmissionMonth = getIsoMonth(new Date());
+    const reconciliationSummaryItems = await api.getUtilisationReportsReconciliationSummary(currentPeriodSubmissionMonth, userToken);
+    const bank = reconciliationSummaryItems
+      .find((summaryItem) => summaryItem.submissionMonth === currentPeriodSubmissionMonth)
+      ?.items.find((item) => item.bank.id === bankId)?.bank;
+    if (!bank) {
+      console.error(`Bank with id ${bankId} not found`);
+      return res.redirect('/not-found');
+    }
+
+    return res.render('utilisation-reports/utilisation-report-for-bank.njk', {
+      user,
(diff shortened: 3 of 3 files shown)


trade-finance-manager-ui/component-tests/utilisation-reports/_macros/report-reconciliation-table.component-test.js
```diff
@@ -35,6 +36,11 @@ describe(component, () => {
 
       wrapper.expectElement(`${rowSelector} th`).toHaveCount(1);
       wrapper.expectElement(`${rowSelector} th:contains("${summaryItem.bank.name}")`).toExist();
+      if (summaryItem.status === 'REPORT_NOT_RECEIVED') {
```
toddym42, line 39: As this is a `.js` file we should use `UTILISATION_REPORT_RECONCILIATION_STATUS.REPORT_NOT_RECEIVED` here.

trade-finance-manager-ui/component-tests/utilisation-reports/_macros/report-reconciliation-table.component-test.js
```diff
@@ -35,6 +36,11 @@ describe(component, () => {
...
+        wrapper.expectLink(`${rowSelector} th > a`).toLinkTo(undefined, summaryItem.bank.name);
```
toddym42, line 40: Is this temporary behaviour or expected for this status? If we don't link anywhere for banks where report isn't received I think this shouldn't be an `a` element.

trade-finance-manager-ui/server/routes/utilisation-reports/index.js
```diff
@@ -1,10 +1,18 @@
...
 const { PDC_TEAM_IDS } = require('../../constants');
 
 const router = express.Router();
 
 router.get('/', validateUserTeam(Object.values(PDC_TEAM_IDS)), getUtilisationReports);
 
+router.get(
+  '/bank/:id',
+  validateUserTeam(Object.values(PDC_TEAM_IDS)),
+  param('id').isString().notEmpty().withMessage('Bank id must be a non-empty string'),
```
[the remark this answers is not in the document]
toddym42: If we use `bankId` as the param name we can just copy the bankId validators from the other projects. Means we don't need to do `{ id: bankId }` in the route handler too.

trade-finance-manager-ui/component-tests/utilisation-reports/_macros/report-reconciliation-table.component-test.js
JoshBinns2000: So I tried to not make it an `a` element, but making it an `a` without an `href` attribute keeps the same text formatting except without the blue underlined colour

JoshBinns2000: Clearly a controversial decision! It turns out you can make sure the text has the same style by wrapping in a `<p class="govuk-body">` tag

trade-finance-manager-ui/component-tests/utilisation-reports/_macros/report-reconciliation-table.component-test.js
```diff
@@ -53,4 +59,23 @@ describe(component, () => {
       }
     });
   });
+
+  it("should link to the correct location when the bank has any report status which isn't 'REPORT_NOT_RECEIVED'", () => {
```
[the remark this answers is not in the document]
JoshBinns2000: I could, but the edge case here is `REPORT_NOT_RECEIVED` as any other case is captured in the `else`. In theory you could pass in an invalid report status and still see the same link, so I don't think there is an advantage to checking all the valid report statuses (if anything the argument would be to check the `REPORT_NOT_RECEIVED` case as this is the only time you get different behaviour, but I primarily just wanted a check to make sure the link did what it is supposed to)

trade-finance-manager-ui/server/controllers/utilisation-reports/index.js
```diff
@@ -29,6 +30,41 @@ const getUtilisationReports = async (req, res) => {
...
+  const { id: bankId } = req.params;
+
+  const errors = validationResult(req);
+  if (!errors.isEmpty()) {
+    console.error(`Error rendering utilisation for bank with id ${bankId}:`, errors.array());
+    return res.render('_partials/problem-with-service.njk', { user });
+  }
+
+  try {
+    const reconciliationSummaryApiResponse = await api.getUtilisationReportsReconciliationSummary({
```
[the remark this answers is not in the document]
JoshBinns2000: Changed it instead to `reconciliationSummaryItems` - it is called `...ApiResponse` I believe as the API response is not quite what the UI considers to be the `reconciliationSummary`, but that doesn't matter as far as my controller is concerned

trade-finance-manager-ui/templates/utilisation-reports/utilisation-report-for-bank.njk
```diff
@@ -0,0 +1,79 @@
+{% from "govuk/components/tabs/macro.njk" import govukTabs %}
+
+{% extends "index.njk" %}
+
+{% set premiumPaymentsHtml %}
+  <div></div> {# TODO: FN-XXXX #}
```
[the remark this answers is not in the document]
JoshBinns2000: I'm not sure which ticket each of them will belong to, but I've added two tickets which seem close and left the other two as I don't think they have tickets yet

trade-finance-manager-ui/server/controllers/utilisation-reports/index.test.ts
```diff
@@ -65,4 +69,96 @@ describe('controllers/utilisation-reports', () => {
...
+        params: { id: bankId },
+      });
+
+      isEmptyMock.mockReturnValueOnce(true);
+
+      const bank = {
+        id: bankId,
+        name: 'Test Bank',
+      };
+      mocked(api.getUtilisationReportsReconciliationSummary).mockResolvedValue([{ bank }] as UtilisationReportReconciliationSummaryItem[]);
```
toddym42, line 143: Can use one of the values from `MOCK_UTILISATION_REPORT_RECONCILIATION_SUMMARY_ITEMS` here to get a more realistic result, rather than casting a partial object.

trade-finance-manager-ui/server/controllers/utilisation-reports/index.ts
```diff
@@ -28,3 +29,35 @@ export const getUtilisationReports = async (req: Request, res: Response) => {
     return res.render('_partials/problem-with-service.njk', { user });
   }
 };
+
+export const getUtilisationReportByBankId = async (req: Request, res: Response) => {
+  const { userToken, user } = req.session;
+  const { bankId } = req.params;
+
+  const errors = validationResult(req);
```
toddym42, line 37: It doesn't look like this will ever return anything, since the validation errors are already handled (with a redirect) by the relevant validation middleware used for this endpoint (`validateUserTeam` and `validateBankId`).

trade-finance-manager-ui/server/controllers/utilisation-reports/index.ts
```diff
@@ -28,3 +29,35 @@ export const getUtilisationReports = async (req: Request, res: Response) => {
...
+    console.error(`Error rendering utilisation for bank with id ${bankId}:`, errors.array());
+    return res.render('_partials/problem-with-service.njk', { user });
+  }
+
+  try {
+    const submissionMonth = getIsoMonth(new Date());
+    const reconciliationSummaryItems = await api.getUtilisationReportsReconciliationSummary(submissionMonth, asString(userToken));
+    const bank = reconciliationSummaryItems.find((summaryItem) => summaryItem.bank.id === bankId)?.bank;
+    if (!bank) {
+      throw new Error(`Bank with id ${bankId} not found`);
```
toddym42, line 48: Would a
```
```
be more appropriate here?

trade-finance-manager-ui/server/controllers/utilisation-reports/index.ts
```diff
@@ -28,3 +29,35 @@ export const getUtilisationReports = async (req: Request, res: Response) => {
...
+    }
+
+    const reportPeriod = getFormattedReportPeriod();
+
+    return res.render('utilisation-reports/utilisation-report-for-bank.njk', {
+      user,
+      activePrimaryNavigation: 'utilisation reports',
```
toddym42, line 55: It might be nice to have these values pulled from a constant (we could just include `'utilisation reports'` initially).

trade-finance-manager-ui/server/middleware/validateBankId/index.test.ts
```diff
@@ -0,0 +1,42 @@
...
+
+console.error = jest.fn();
+
+describe('validateBankId', () => {
+  const notFoundUrl = '/not-found';
+  it.each`
+  description                             | value
+  ${'undefined'}                          | ${undefined}
+  ${'a string with non-digit characters'} | ${'abc'}
+  ${'a string with a non-integer number'} | ${'12.3'}
```
toddym42, line 13: Indenting is off here.

trade-finance-manager-ui/server/middleware/validateBankId/index.test.ts
```diff
@@ -0,0 +1,42 @@
...
+    // Arrange
+    const { req: mockRequest, res: mockResponse } = httpMocks.createMocks<Request<{ bankId: string }>>({
+      params: { bankId: value },
+    });
+    const mockNext = jest.fn();
+
+    // Act
+    validateBankId(mockRequest, mockResponse, mockNext);
+
+    // Assert
```
toddym42, line 24: I would add the following here just to be sure
```
expect(mockNext).not.toHaveBeenCalled();
```

trade-finance-manager-ui/server/middleware/validateBankId/index.ts
```diff
@@ -0,0 +1,15 @@
+import { NextFunction, Request, Response } from 'express';
+
+export const validateBankId = (req: Request, res: Response, next: NextFunction) => {
+  const { bankId } = req.params;
+  if (!bankId) {
+    console.error('No bank id was provided');
+    return res.redirect('/not-found');
+  }
+
+  if (typeof bankId !== 'string' || !bankId.match(/^\d+$/)) {
```
toddym42, line 10: Could use the `!isString(bankId)` helper here.

trade-finance-manager-ui/server/routes/utilisation-reports/index.js
```diff
@@ -1,13 +1,20 @@
...
+const { PDC_TEAM_IDS } = require('../../constants');
 
 const router = express.Router();
 
 router.get('/', validateUserTeam(Object.values(PDC_TEAM_IDS)), getUtilisationReports);
 
 router.get('/:_id/download', validateUserTeam(Object.values(PDC_TEAM_IDS)), validateMongoId, getReportDownload);
 
+router.get(
+  '/bank/:bankId',
```
toddym42, line 14: I wonder if we should make this endpoint specify the submission month (e.g. `/:submissionMonth/bank/:bankId` as its not clear what would be returned from this. (we can steal the `isoMonthValidation` from TFM API to validate it)

I know this is currently doing the same thing as the `/` endpoint by only returning results for the current submission month, but I think that was a mistake and will probably need to change when we add details of previous reports to the "Bank reports" page. What do you think?

trade-finance-manager-ui/component-tests/utilisation-reports/_macros/report-reconciliation-table.component-test.js
```diff
@@ -53,4 +59,23 @@ describe(component, () => {
...
+    const summaryItem = getReportReconciliationSummaryViewModel([
+      {
+        ...MOCK_UTILISATION_REPORT_RECONCILIATION_SUMMARY[0],
+        status: UTILISATION_REPORT_RECONCILIATION_STATUS.PENDING_RECONCILIATION,
+      },
+    ]).at(0);
+    wrapper = render({
+      reportReconciliationSummary: [summaryItem],
```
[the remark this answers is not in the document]
JoshBinns2000: Yeh I'm pretty sure this didn't exist when I first wrote this code, but I saw it on a recent PR - will pull in the change

trade-finance-manager-ui/server/controllers/utilisation-reports/index.ts
JoshBinns2000: I'm not sure why, but when I first wrote this I don't remember that being the case. I also think I was trying to start encouraging the pattern of `res` logic being handled only in the controller, but if we're going to keep the portal pattern then I'm happy to do that

trade-finance-manager-ui/server/routes/utilisation-reports/index.js
JoshBinns2000: I agree that it will be a problem once we start including previous submission months. I will add the functionality at a basic level for this route and rebase on your previous submission months changes once they're in to finish up the logic (under the assumption that will be merged first)

trade-finance-manager-ui/server/controllers/utilisation-reports/index.ts
JoshBinns2000: (Seems that was a lie because I added the validation middleware...)

trade-finance-manager-ui/server/routes/utilisation-reports/index.js
toddym42: I think the previous reports work might still be a while. Is there any information you don't already have to be able to complete this without that new work?

trade-finance-manager-ui/server/controllers/utilisation-reports/index.ts
JoshBinns2000: I have introduced this as part of a separate branch as I may as well introduce this constant across the project - will link the new PR in this comment thread once it is up

JoshBinns2000: It's here: #2552

trade-finance-manager-ui/server/routes/utilisation-reports/index.js
JoshBinns2000: Is there a reason to not use the month and year as query parameters like `/bank/:bankId?sumissionMonth=&submissionYear=`?

toddym42: Yeah, good question. Some reasoning detailed here: https://github.com/UK-Export-Finance/dtfs2/pull/2366#discussion_r1419183543

JoshBinns2000: Reasoning sounds good to me - for my own sake, an example url would then be `/utilisation-reports/2023-12/bank/123`

JoshBinns2000: I have added in this functionality now

PR #2458, opened 2023-12-20
chore(DTFS2-6700): remove tfm search vulnerabilities

MarRobSoftwire:
## Introduction :pencil2:
The search box in TFM shouldn't allow users to attack the database

In `dtfs-central-api` escape regex characters to prevent potential ReDOS attacks

[Attempted other ways attack the database, with no vulnerabilities found](https://ukef-dtfs.atlassian.net/browse/DTFS2-6700)

## Note ⚠️ 
This removes the ability to search the database with regex through the api. Whilst this shouldn't be used anywhere its potentially a breaking change

### DIFF
--- a/dtfs-central-api/src/helpers/escapeRegExp.ts
+++ b/dtfs-central-api/src/helpers/escapeRegExp.ts
@@ -1,1 +1,0 @@
-export const escapeRegExp = (string: string) => string.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
--- a/dtfs-central-api/src/helpers/escapeRegExp.ts
+++ b/dtfs-central-api/src/helpers/escapeRegExp.ts
@@ -1,3 +1,1 @@
-/* eslint-disable import/prefer-default-export */
-
 export const escapeRegExp = (string: string) => string.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
--- a/dtfs-central-api/src/helpers/escapeRegExp.ts
+++ b/dtfs-central-api/src/helpers/escapeRegExp.ts
@@ -1,0 +1,3 @@
+/* eslint-disable import/prefer-default-export */
+
+export const escapeRegExp = (string: string) => string.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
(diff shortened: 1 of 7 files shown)


dtfs-central-api/src/v1/controllers/tfm/deal/tfm-get-deals.controller.js
```diff
@@ -48,6 +48,8 @@ const sortDeals = (deals, sortBy) =>
     return 0;
   });
 
+const escapeRegExp = (string) => string.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
```
MarRobSoftwire, line 51: This is a bit of a hack. 
It would probably be better to use [Text indexes ](https://www.mongodb.com/docs/manual/core/indexes/index-types/index-text/) but this would require modifying the database

dtfs-central-api/src/v1/controllers/tfm/facility/tfm-get-facilities.controller.js
```diff
@@ -18,7 +19,7 @@ exports.getFacilitiesByDealId = async (req, res) => {
 
 exports.getAllFacilities = async (req, res) => {
   const collection = await db.getCollection('tfm-facilities');
-  const searchString = req.body.searchString || '';
+  const searchStringRegExp = escapeRegExp(req.body.searchString || '');
```
abhi-markan, line 22: Can we refactor this statement with optional chaining please?

abhi-markan (changes requested):

dtfs-central-api/src/helpers/escapeRegExp.ts
```diff
@@ -0,0 +1 @@
+export const escapeRegExp = (string: string) => string.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
```
avaitonis, line 1: Do we want to consider using package `escape-string-regexp`? it is already used by about 4 of our dependencies.

Lodash also has this helper, but `dtfs-central-api` is not loading full lodash at the moment.

dtfs-central-api/src/v1/controllers/tfm/deal/tfm-get-deals.controller.js
```diff
@@ -69,23 +70,26 @@ const findDeals = async (searchString, sortBy, fieldQueries, callback) => {
       dateString = String(moment(date).format('DD-MM-YYYY'));
     }
 
+    const searchStringRegex = escapeRegExp(searchString);
```
avaitonis, line 73: Just a suggestion that maybe variable name `searchStringRegex` is not clear, because we don't expect user to enter regex.

Maybe `safeSearchString` or `searchStringEscaped` is better?

abhi-markan (approved):

PR #2461, opened 2023-12-20
chore(DTFS2-6666): remove filter vulnerabilities

MarRobSoftwire:
## Introduction :pencil2:
The keyword filter in portal shouldn't allow users to attack the database

- Use `KEYWORD` when sending a keyword filter to the api
- Convert `KEYWORD` into the `$regex` operator, escaping any regex special characters

## Note ⚠️ 
- This removes the ability to send a `REGEX` request to the api, this shouldn't be used anywhere but is potentially a breaking change

### DIFF
--- a/portal-api/src/v1/helpers/escapeOperators.js
+++ b/portal-api/src/v1/helpers/escapeOperators.js
@@ -1,4 +1,5 @@
 /* eslint-disable no-param-reassign */
+const escapeStringRegexp = require('escape-string-regexp');
 
 /**
  * Objective:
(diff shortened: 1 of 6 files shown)


portal-api/src/v1/helpers/escapeOperators.js
```diff
@@ -18,23 +18,20 @@ Flow:
...
+  The function adds "$eq" to any criteria that is not an array.
   If the input "filter" is not an object or is null, it will be returned as is.
  */
 
-/**
- * This function escapes MongoDB operators from the filter object
- * @param {Object} filter Object comprising of filters
- * @returns {Object} Escaped filter object
- */
+const escapeRegExp = (string) => string.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
```
abhi-markan, line 34: Can the RegEx be referred from a constant please?

abhi-markan (changes requested):
One minor comment to address please.

abhi-markan (approved):

PR #2463, opened 2023-12-20
feat(FN-1483): add AZURE env vars to TFM API disaster recovery scripts

toddym42:


francescastocco (commented):
Nice, I've realised that we need to make the same change for `staging_infrastructure.yml` and `production_infrastructure.yml` too

francescastocco (approved):

PR #2466, opened 2023-12-21
feat(FN-692): infrastructure improvements

francescastocco:
## Introduction :pencil2:
Feature environment infrastructure improvements

- Added what-if to bicep deploy pipeline
- Added a pipeline for linting bicep code when pull request created against `feat/stbr-infra`
- Disabled the function-apps deployment in the feature_deploy pipeline

### DIFF
--- a/.github/workflows/bicep_deploy.yml
+++ b/.github/workflows/bicep_deploy.yml
@@ -127,9 +127,18 @@ jobs:
       resourceGroupName: ${{ inputs.resourceGroupName }}
       action: validate
     secrets: inherit
+    
+  what_if:
+    needs: validate
+    uses: ./.github/workflows/bicep_az_deployment_group.yml
+    with:
+      environment: ${{ inputs.environment }}
+      resourceGroupName: ${{ inputs.resourceGroupName }}
+      action: what-if
+    secrets: inherit
 
   deploy:
-    needs: validate
+    needs: what_if
     uses: ./.github/workflows/bicep_az_deployment_group.yml
     with:
       environment: ${{ inputs.environment }}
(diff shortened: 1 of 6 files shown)


.github/workflows/bicep_lint.yml
```diff
@@ -0,0 +1,30 @@
...
+
+on:
+  pull_request:
+    branches:
+      - feat/stbr-infra
+    paths:
+      - '.github/workflows/bicep_az_deployment_group.yml'
+      - '.github/workflows/bicep_deploy.yml'
+      - '.github/workflows/bicep_workflow_feature.yml'
+      - 'infrastructure/resource_group_level/**'
```
ChristianMcc, line 11: should this just be infrastructure? Right now I assume everything is resource_group_level but would assume if we ever have any bicep that's at a subscription level we would need it here

.github/workflows/bicep_deploy.yml
```diff
@@ -127,9 +127,18 @@ jobs:
...
+    uses: ./.github/workflows/bicep_az_deployment_group.yml
+    with:
+      environment: ${{ inputs.environment }}
+      resourceGroupName: ${{ inputs.resourceGroupName }}
+      action: what-if
+    secrets: inherit
 
   deploy:
-    needs: validate
+    needs: what_if
```
ChristianMcc, line 141: was there a way for this to be manual only for certain environments and not for others?

.github/workflows/bicep_lint.yml
francescastocco: Hm yes, good point

.github/workflows/bicep_deploy.yml
francescastocco: I've left it as automatic for all environments for now as it looked quite complex to add the manual step so will come back and look at it if I get more time

ChristianMcc: Sounds good, will resolve :)

ChristianMcc (approved):

PR #2425, opened 2023-12-05
feat(FN-1618): add tfm endpoint to set report status

JoshBinns2000:
## Introduction :pencil2:
We want to introduce an endpoint for `TFM-API` to communicate with `DTFS-CENTRAL_API` to set a reports status (currently to either `REPORT_NOT_RECEIVED` or `RECONCILIATION_COMPLETED`).

## Resolution :heavy_check_mark:
- Adds the required endpoint
- Adds payload validation for the body of the endpoint
- Adds the relevant unit and api tests
- Adds new `types` to `trade-finance-manager-api`

## Miscellaneous
Updates the `TfmSessionUser` type in `dtfs-central-api` to match the format of `_id` being a `string` rather than an `ObjectId`.

PR #2489, opened 2023-12-28
feat(FN-1536): mark report as done

JoshBinns2000:
## Introduction :pencil2:
We want to add functionality to the TFM utilisation report reconciliation table so that a `PDC_RECONCILE` user can mark a report as either done or not done. The backend for this ticket was split into #2362 and #2425.

Adds the buttons to the table and creates the relevant controller function. The checkboxes that were added do not use the provided `govukCheckboxes` component as this component does not allow you to use individual checkboxes easily. Instead, the HTML source code for these components was used to add the individual checkboxes to the table rows. 

Slight modifications were made to the backend code to account for what it means to mark a report as "done" or "not done":
- If a report had already been uploaded, marking as "not done" sets the status to `PENDING_RECONCILIATION`
- If a report had not been uploaded, and the report had been marked as done by a `PDC_RECONCILE` user, marking as "not done" deletes the report (effectively setting the report status to `NOT_RECEIVED`)

This logic is handled partially by the controller when creating the payload for the request in `trade-finance-manager-ui` and partially by `dtfs-central-api` within the `setToPendingReconciliationOrDeleteReport` function.

Changes were also made to the existing utilisation reports table to prevent users from attempting to download placeholder reports.

## Screenshots 📷 
In the below image, one report is a placeholder and therefore cannot be downloaded and the other is not a placeholder and so can.

FN-1536-most-recent


trade-finance-manager-ui/server/api.js
```diff
@@ -839,6 +839,22 @@ const downloadUtilisationReport = async (userToken, _id) => {
   };
 };
 
+/**
+ * @param {import('./controllers/utilisation-reports').UpdateReportStatusPayload} updateReportStatusPayload - payload to send in request body
+ * @param {string} userToken - token to validate session
+ * @returns {Promise<import('axios').AxiosResponse>}
```
JoshBinns2000, line 845: I don't like this return type, but I'm not sure there is a better option. I'm pretty sure nothing specific gets returned in `response.data` as tfm-api just calls `res.sendStatus(204)`. Open to suggestions for better return types here

trade-finance-manager-ui/server/controllers/utilisation-reports/index.ts
```diff
@@ -1,30 +1,122 @@
...
+      year,
     });
   } catch (error) {
     console.error('Error rendering utilisation reports page', error);
     return res.render('_partials/problem-with-service.njk', { user });
   }
 };
+
+const getReportIdentifiersFromUrl = (body: object, month: number, year: number): ReportIdentifier[] =>
+  Object.keys(body).reduce((reportIdentifiers, key) => {
```
JoshBinns2000, line 58: Just because I anticipate confusion around this function, here is an explanation for why I did it.

The form posts the information from all the inputs in the form as a payload. This includes things like the `_csrfToken` token, but also includes the checkboxes that *have* been ticked (if a checkbox is unticked, it is not sent in the payload, so the value will always be `'on'`). I named the checkboxes `set-status--(bankId|reportId)-(id)` such that `req.body` may look like
```
{
  _csrfToken: abc123,
  set-status--bankId-123: 'on',
  set-status--reportId-abcdef1234: 'on',
}
```
Knowing that it looks like this, the only way to tell which reports the user wants to update is to look through the keys and extract those which match the checkbox format. I then need to check whether or not a bank id or report id is supplied, hence the two groups `idType` and `id` below

trade-finance-manager-ui/server/controllers/utilisation-reports/index.ts
```diff
@@ -1,30 +1,122 @@
...
+  }, [] as ReportIdentifier[]);
+
+export type UpdateReportStatusPayload = {
+  reportsWithStatus: ReportWithStatus[];
+  user: TfmSessionUser;
+};
+
+export const updateUtilisationReportStatus = async (req: Request, res: Response) => {
+  const { user, userToken } = req.session;
+  const { status, month, year } = req.query;
```
JoshBinns2000, line 89: `status` here is set depending on which button is pressed in the client. There is no reason why I can't change this to either `RECONCILIATION_COMPLETED` or `REPORT_NOT_RECEIVED`, but I want to keep it more general as the behaviour when marking as not completed is not consistent

trade-finance-manager-ui/templates/utilisation-reports/_macros/report-reconciliation-table.njk
```diff
@@ -2,56 +2,91 @@
...
-      <tr class="govuk-table__row">
-        <th scope="col" class="govuk-table__header" aria-sort="ascending">Bank</th>
-        <th scope="col" class="govuk-table__header" aria-sort="none">Status</th>
-        <th scope="col" class="govuk-table__header" aria-sort="none">Date report received</th>
-        <th scope="col" class="govuk-table__header" aria-sort="none">Total fees reported</th>
-        <th scope="col" class="govuk-table__header" aria-sort="none">Reported fees left to reconcile</th>
-        <th scope="col" class="govuk-table__header">Download report as CSV</th>
-      </tr>
-    </thead>
+  <form method="POST" data-cy="form">
```
JoshBinns2000, line 9: `method="PUT"` is not supported apparently, so this form needs to be submitted as a post request instead of put

trade-finance-manager-ui/server/routes/utilisation-reports/index.js
```diff
@@ -8,6 +8,8 @@ const router = express.Router();
 
 router.get('/', validateUserTeam(Object.values(PDC_TEAM_IDS)), getUtilisationReports);
 
+router.post('/', validateUserTeam([PDC_TEAM_IDS.PDC_RECONCILE]), updateUtilisationReportStatus);
```
JoshBinns2000, line 11: This needs to be `post` and not `put` due to the html form element not supporting `method="PUT"`

trade-finance-manager-ui/server/controllers/utilisation-reports/index.ts
JoshBinns2000: I could also change the name to `state` to avoid confusion with `status`

trade-finance-manager-ui/server/controllers/utilisation-reports/index.ts
```diff
@@ -1,30 +1,130 @@
...
+
+  try {
+    if (!user) {
+      throw new Error('User is not logged in');
+    }
+
+    const reportIdentifiers = getReportIdentifiersFromUrl(req.body as object, Number(month), Number(year));
+
+    if (status !== 'completed' && status !== 'not-completed') {
+      throw new Error('Supplied status instruction does not match any expected status');
```
JoshBinns2000, line 100: This error message is an example of how using `state` for `completed` and `not-completed` may avoid confusion

trade-finance-manager-ui/server/controllers/utilisation-reports/index.ts
```diff
@@ -1,30 +1,130 @@
...
+    const reportsWithStatus = reportIdentifiers.map((reportIdentifier): ReportWithStatus => {
+      if ('id' in reportIdentifier && statusToSet === UTILISATION_REPORT_RECONCILIATION_STATUS.REPORT_NOT_RECEIVED) {
+        return {
+          status: UTILISATION_REPORT_RECONCILIATION_STATUS.PENDING_RECONCILIATION,
+          report: reportIdentifier,
+        };
+      }
+      if ('bankId' in reportIdentifier && statusToSet === UTILISATION_REPORT_RECONCILIATION_STATUS.REPORT_NOT_RECEIVED) {
+        throw new InvalidReportStatusRequestError("Cannot set non-existing report to 'REPORT_NOT_RECEIVED'");
+      }
```
JoshBinns2000, line 114: I want to discuss this when I'm back, but annoyingly `session.withTransaction` (in `dtfs-central-api`) does not roll back any changes made if an error is thrown inside the function. So, if you try to mark two reports as "not done", one of which is in the `NOT_RECEIVED` state and one in the `RECONCILIATION_COMPLETED` state, an error is thrown and you see the `Problem with service` page. Then, navigating back to "bank reports" shows that the report which was in the `RECONCILIATION_COMPLETED` state is now in the `PENDING_RECONCILIATION` state. To avoid this happening, I catch the attempt to mark a not received report as "not done" in the frontend and show that error on the page (see the screenshot in the PR description)

trade-finance-manager-ui/server/controllers/utilisation-reports/index.ts
toddym42: I think it makes sense to have a different name so you don't confuse it with the `status` in the API request, as you say. I think would be good to have something that identifies how it's different, whereas `state` is just another way of saying the same thing. Maybe it could be something like `formStatus`?

I think we would also have a constant to document the possible values.

trade-finance-manager-ui/server/controllers/utilisation-reports/index.ts
```diff
@@ -1,30 +1,130 @@
...
+      throw new Error('User is not logged in');
+    }
+
+    const reportIdentifiers = getReportIdentifiersFromUrl(req.body as object, Number(month), Number(year));
+
+    if (status !== 'completed' && status !== 'not-completed') {
+      throw new Error('Supplied status instruction does not match any expected status');
+    }
+    const statusToSet =
+      status === 'completed' ? UTILISATION_REPORT_RECONCILIATION_STATUS.RECONCILIATION_COMPLETED : UTILISATION_REPORT_RECONCILIATION_STATUS.REPORT_NOT_RECEIVED;
```
toddym42, line 103: This seems a bit redundant and misleading if it's not actually always used in the API request. We could instead just use the form status (or whatever we choose to call it) to determine the request body (and can also move the invalid status check here too) e.g.
```
const reportsWithStatus = reportIdentifiers.map((reportIdentifier): ReportWithStatus => {
  switch (formStatus) {
    case 'not-completed': {
      if ('id' in reportIdentifier) {
        return {
          status: UTILISATION_REPORT_RECONCILIATION_STATUS.PENDING_RECONCILIATION,
          report: reportIdentifier,
        };
      }
      if ('bankId' in reportIdentifier) {
        throw new InvalidUpdateReportStatusRequestError("Cannot set non-existing report to 'REPORT_NOT_RECEIVED'");
      }
      throw new Error('Invalid report identifier');
    }
    case 'completed': {
      return {
        status: UTILISATION_REPORT_RECONCILIATION_STATUS.RECONCILIATION_COMPLETED,
        report: reportIdentifier,
      };
    }
    default: {
      throw new Error('Supplied status instruction does not match any expected status');
    }
  }
});
```
might be nice to also extract this out to a separate function so the main controller function doesn't contain too much low level logic. e.g.
```
const reportsWithStatus = getReportsWithStatus(formStatus, reportIdentifiers);
```

trade-finance-manager-ui/server/controllers/utilisation-reports/index.ts
```diff
@@ -1,30 +1,130 @@
...
+    });
+
+    const updateReportStatusPayload: UpdateReportStatusPayload = { user, reportsWithStatus };
+    await api.updateUtilisationReportStatus(updateReportStatusPayload, asString(userToken));
+
+    return await getUtilisationReports(req, res);
+  } catch (error) {
+    console.error('Error updating utilisation report statuses:', error);
+    if (error instanceof InvalidUpdateReportStatusRequestError) {
+      res.locals.errorSummary = [{ text: 'Cannot mark "not received" report as "not done"', href: '#' }];
```
toddym42, line 125: `text` - we should use '"not completed"` instead of `"not done"` to match the text of the button. If we're going to use a custom error for this case we could just use the message (or some other custom field if we choose) for the text here.

`href` - I don't think we want an anchor link for this error? In which case we should remove this attribute (see [docs](https://design-system.service.gov.uk/components/error-summary/#options-error-summary-example--error-list))

trade-finance-manager-ui/server/errors/invalid-update-report-status-request.error.test.ts
```diff
@@ -0,0 +1,22 @@
...
+
+  it('exposes the message it was created with', () => {
+    // Act
+    const exception = new InvalidUpdateReportStatusRequestError(message);
+
+    // Assert
+    expect(exception.message).toBe(message);
+  });
+
+  // Test will pass if we ever update to es6
```
toddym42, line 14: I saw in the tsconfig we've now downgraded to es5 from es6 due to some other issues. It would to resolve those issues so we can keep es6 if possible.

trade-finance-manager-ui/templates/utilisation-reports/_macros/report-reconciliation-table.njk
```diff
@@ -2,56 +2,95 @@
...
+        data-module="govuk-button"
+        value="Mark report as completed"
+        data-cy="mark-report-as-completed-button"/>
+      <input
+        formaction="?_csrf={{ csrfToken }}&status=not-completed&month={{ month }}&year={{ year }}"
+        type="submit"
+        class="govuk-button govuk-button--secondary govuk-!-margin-right-1"
+        data-module="govuk-button"
+        value="Mark as not completed"
+        data-cy="mark-as-not-completed-button"/>
```
toddym42, line 93: I think we can simplify these a bit by using `govukButton` (see other uses of `formaction` in the codebase) which uses an underlying `button` element. I think `<input type="button"` or `<input type="submit"` was used historically due to compatibility issues with older browsers (like IE 6), but the default now should just be to use `button` (or in our case we can use the `govukButton` nunjucks template).

trade-finance-manager-ui/templates/utilisation-reports/_macros/report-reconciliation-table.njk
```diff
@@ -2,56 +2,95 @@
...
+
+    <p class="govuk-body">
+      When a report has been reconciled, select it and mark it as completed. If it needs to be returned to its previous state, mark it as not completed.
+    </p>
+
+    <div class="govuk-button-group">
+      <input
+        formaction="?_csrf={{ csrfToken }}&status=completed&month={{ month }}&year={{ year }}"
+        type="submit"
+        class="govuk-button govuk-!-margin-right-1"
```
toddym42, line 83: Looks like the designs have a `govuk-!-margin-right-3` here, though actually you don't need to specify it at all. Gaps between buttons will be automatically added by the surrounding `govuk-button-group`.

trade-finance-manager-ui/tsconfig.json
```diff
@@ -11,7 +11,7 @@
       // "disableReferencedProjectLoad": true,             /* Reduce the number of projects loaded automatically by TypeScript. */
 
       /* Language and Environment */
-      "target": "ES6",                                     /* Set the JavaScript language version for emitted JavaScript and include compatible library declarations. */
+      "target": "es5",                                     /* Set the JavaScript language version for emitted JavaScript and include compatible library declarations. */
```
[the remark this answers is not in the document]
toddym42: What was the issue you encountered? It would be a pain to go backwards as this will throw up other issues and restrict using new functionality.

trade-finance-manager-ui/templates/utilisation-reports/_macros/report-reconciliation-table.njk
```diff
@@ -2,56 +2,95 @@
...
+                Download
+              </a>
+            {% endif %}
+          </td>
+        </tr>
+      {% endfor %}
+      </tbody>
+    </table>
+
+    <p class="govuk-body">
```
toddym42, line 75: When we add the details of previous reports (see [FN-1051](https://ukef-dtfs.atlassian.net/browse/FN-1051)) will there be a set of buttons for each table, or one set of buttons for all the tables? If the latter then these buttons should go outside of this `report-reconciliation-table` macro.

It's not clear from the designs which of the above we should do, so it's something we'll need to find out (if you haven't already). (@ChristianMcc - you may have thoughts on this)

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-reports-reconciliation-summary.controller.js
```diff
@@ -59,6 +59,7 @@ const generateReconciliationSummaryItem = async (bank, submissionMonth) => {
     dateUploaded: report.dateUploaded,
     totalFeesReported: reportData.length,
     reportedFeesLeftToReconcile,
+    isPlaceholderReport: report.azureFileInfo === null,
```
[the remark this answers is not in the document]
JoshBinns2000: I think it should be optional in this case as it neither is nor isn't a placeholder report in the case that there is no report

trade-finance-manager-ui/server/controllers/utilisation-reports/index.ts
```diff
@@ -1,30 +1,130 @@
...
+import { ReportIdentifier, ReportWithStatus } from '../../types/utilisation-reports';
+import { UTILISATION_REPORT_RECONCILIATION_STATUS } from '../../constants';
+import { InvalidUpdateReportStatusRequestError } from '../../errors/invalid-update-report-status-request.error';
+
+/**
+ * Gets the data required to render the /utilisation-reports page
+ */
+const getUtilisationReportsSummaryData = async (userToken: unknown) => {
+  const currentDate = new Date();
+  const oneIndexedMonth = subMonths(currentDate, 1).getMonth() + 1;
```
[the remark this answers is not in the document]
JoshBinns2000: Unless I'm wrong, we want the report period which is the previous month, so this is the way to do it. Variable names could be made clearer though which I will do as part of the comment below

trade-finance-manager-ui/tsconfig.json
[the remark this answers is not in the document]
JoshBinns2000: I think this is more of an issue with my VSCode as, this morning, I'm getting the same issue I was getting then despite using `es5`. I will look into fixing my VSCode setup at some point

trade-finance-manager-ui/templates/utilisation-reports/_macros/report-reconciliation-table.njk
ChristianMcc: Probably going to have to ask Rich/Andy about this one, in my head it was a form and set of buttons per table. Mostly because I was imagining working on one month at a time and if for some reason one bank was behind by a couple of months it could push the buttons off screen and be hard to see what's going on

JoshBinns2000: If the buttons aren't specific to the table being viewed, the checkbox id will need to contain all of the `bankId`, `submissionMonth` and `submissionYear` rather than my current implementation which puts the month and year into the `url`. It is easier to keep the buttons table specific so that we don't need each checkbox to store the month and year in my opinion

trade-finance-manager-ui/templates/utilisation-reports/_macros/report-reconciliation-table.njk
JoshBinns2000: I used this format as `utlisation-report-upload` in `portal` does this rather than using the `govukButton` component, but I will look into it and see if I can find any reason to not use it

JoshBinns2000: Looks like `govukButton` should work fine

trade-finance-manager-ui/server/errors/invalid-update-report-status-request.error.test.ts
JoshBinns2000: Reverted to `es6`

trade-finance-manager-ui/server/controllers/utilisation-reports/index.ts
JoshBinns2000: Removed as there is no specific design for this error case, but as far as my testing showed an `href` is required in order to render the `govukErrorSummary` component

trade-finance-manager-ui/server/controllers/utilisation-reports/index.ts
JoshBinns2000: I changed this to be called `formButton` as that is where the information comes from. There is also a new constant called `FORM_BUTTON_VALUES` which documents each of the possible expected values

dtfs-central-api/src/services/repositories/utilisation-reports-repo/update-report-status.ts
```diff
@@ -87,7 +87,8 @@ export const updateManyUtilisationReportStatuses = async (
         }
         return createReportAndSetAsCompleted(utilisationReportsCollection, filter, uploadedByUserDetails);
       case UTILISATION_REPORT_RECONCILIATION_STATUS.REPORT_NOT_RECEIVED:
-        return setToNotReceivedOrDeleteReport(utilisationReportsCollection, filter);
+      case UTILISATION_REPORT_RECONCILIATION_STATUS.PENDING_RECONCILIATION:
+        return setToPendingReconciliationOrDeleteReport(utilisationReportsCollection, filter);
```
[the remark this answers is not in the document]
JoshBinns2000: I've updated the logic now, but this is kind of the whole reason why this temporary solution is a pain. If you try to mark a report as "not completed", you could theoretically try to set the state to either `REPORT_NOT_RECEIVED` or `PENDING_RECONCILIATION` and what actually happens in the database depends entirely on whether or not a report is a placeholder. However, I've updated the logic now so that you can only try to set a report to `PENDING_RECONCILIATION` and dtfs central api decides whether or not to delete the report or set its status (as far as the frontend is concerned, it can see either of those states after clicking the button and the page refreshing)

trade-finance-manager-api/src/v1/validation/route-validators/route-validators.js
```diff
@@ -33,8 +33,8 @@ const updateReportStatusPayloadValidation = [
   checkSchema({
     'reportsWithStatus.*.status': {
       isIn: {
-        options: `${UTILISATION_REPORT_RECONCILIATION_STATUS.REPORT_NOT_RECEIVED},${UTILISATION_REPORT_RECONCILIATION_STATUS.RECONCILIATION_COMPLETED}`,
-        errorMessage: `Report status must be one of the following: ${UTILISATION_REPORT_RECONCILIATION_STATUS.REPORT_NOT_RECEIVED},${UTILISATION_REPORT_RECONCILIATION_STATUS.RECONCILIATION_COMPLETED}`,
+        options: `${UTILISATION_REPORT_RECONCILIATION_STATUS.REPORT_NOT_RECEIVED},${UTILISATION_REPORT_RECONCILIATION_STATUS.RECONCILIATION_COMPLETED},${UTILISATION_REPORT_RECONCILIATION_STATUS.PENDING_RECONCILIATION}`,
```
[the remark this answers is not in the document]
JoshBinns2000: I originally didn't put it into a seperate file because the rest of the validation chains exist only in this file, but I agree that it should really be in its own - will make the change

trade-finance-manager-api/src/v1/validation/route-validators/route-validators.js
```diff
@@ -33,8 +33,8 @@ const updateReportStatusPayloadValidation = [
...
+        errorMessage: `Report status must be one of the following: ${UTILISATION_REPORT_RECONCILIATION_STATUS.REPORT_NOT_RECEIVED},${UTILISATION_REPORT_RECONCILIATION_STATUS.RECONCILIATION_COMPLETED},${UTILISATION_REPORT_RECONCILIATION_STATUS.PENDING_RECONCILIATION}`,
```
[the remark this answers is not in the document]
JoshBinns2000: Moving this validation to a new file makes it all significantly easier to check as I can define things like `VALID_UPDATE_PAYLOAD_STATUSES` without worrying about making the file too clunky, so good suggestion!

trade-finance-manager-ui/server/controllers/utilisation-reports/index.ts
JoshBinns2000: As we don't actually need to handle this logic, we now just filter out any requests where the user tries to mark a non existing report as "not completed"

trade-finance-manager-ui/templates/utilisation-reports/_macros/report-reconciliation-table.njk
```diff
@@ -2,56 +2,95 @@
...
+            scope="row"
+            class="govuk-table__header govuk-!-font-weight-regular"
+            data-sort-value="{{ summaryItem.bank.name }}"
+          >
+            <div class="govuk-checkboxes govuk-checkboxes--small">
+              <div class="govuk-checkboxes__item" data-cy="checkbox--{{ id }}">
+                <input class="govuk-checkboxes__input govuk-!-top-0" id="{{ checkboxId }}" name="{{ checkboxId }}" type="checkbox"/>
+                <label class="govuk-label govuk-checkboxes__label" for="">
+                  <a class="govuk-link govuk-link--no-visited-state" href="#">{{ summaryItem.bank.name }}</a>
+                </label>
```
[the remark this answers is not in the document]
JoshBinns2000: Thanks for this - I spent a while trying to get the styling to work in the VM but got impatient with it (especially with my home wifi making it very slow to edit in the VM), so I appreciate you figuring it out

trade-finance-manager-ui/server/controllers/utilisation-reports/index.ts
```diff
@@ -1,27 +1,47 @@
...
 import { getFormattedReportDueDate, getFormattedReportPeriod } from '../../services/utilisation-report-service';
 import { getIsoMonth } from '../../helpers/date';
 import { getReportReconciliationSummaryViewModel } from './helpers';
 import { asString } from '../../helpers/validation';
 
+export const getUtilisationReportsSummaryData = async (userToken: unknown) => {
+  const currentDate = new Date();
+  const submissionDate = subMonths(currentDate, 1);
+  const oneIndexedSubmissionMonth = submissionDate.getMonth() + 1;
+  const submissionYear = submissionDate.getFullYear();
```
toddym42, line 13: As discussed, these should refer to the start of the report period, instead of submission.

trade-finance-manager-ui/server/controllers/utilisation-reports/update-utilisation-report-status/index.ts
```diff
@@ -0,0 +1,87 @@
...
+    case FORM_BUTTON_VALUES.NOT_COMPLETED:
+      if ('bankId' in reportIdentifier) {
+        return undefined;
+      }
+      return {
+        status: UTILISATION_REPORT_RECONCILIATION_STATUS.PENDING_RECONCILIATION,
+        report: reportIdentifier,
+      };
+    default:
+      throw new Error('Button query parameter matched no known values');
```
toddym42, line 63: Should add `formButton` somewhere in the error message here so we can see what it tried to do.

trade-finance-manager-ui/server/controllers/utilisation-reports/update-utilisation-report-status/index.ts
```diff
@@ -0,0 +1,87 @@
...
+  switch (formButton) {
+    case FORM_BUTTON_VALUES.COMPLETED:
+      return {
+        status: UTILISATION_REPORT_RECONCILIATION_STATUS.RECONCILIATION_COMPLETED,
+        report: reportIdentifier,
+      };
+    case FORM_BUTTON_VALUES.NOT_COMPLETED:
+      if ('bankId' in reportIdentifier) {
+        return undefined;
+      }
```
toddym42, line 57: It's not intuitive to understand why we'd return `undefined` here. Can we make the reasoning clearer in code? Or if not, add a comment.

trade-finance-manager-ui/server/helpers/validation.js
```diff
@@ -35,13 +35,13 @@ const generateValidationErrors = (fieldId, errorText, count, errors = { errorLis
 
 /**
  * @param {unknown} value - the value to check
- * @param {string} [context] - provides context in the error message if value is not a string. Usually would be the name
+ * @param {string} context - provides context in the error message if value is not a string. Usually would be the name
```
toddym42, line 38: Thanks for doing this 🙏

trade-finance-manager-ui/templates/utilisation-reports/_macros/report-reconciliation-table.njk
```diff
@@ -1,57 +1,99 @@
...
+    </table>
+
+    <p class="govuk-body">
+      When a report has been reconciled, select it and mark it as completed. If it needs to be returned to its previous state, mark it as not completed.
+    </p>
+
+    <div class="govuk-button-group">
+      {{ govukButton({
+        text: "Mark report as completed",
+        classes: "govuk-!-margin-right-1",
```
toddym42, line 83: Looks like this margin needs to be removed again https://github.com/UK-Export-Finance/dtfs2/pull/2489#discussion_r1439575137

trade-finance-manager-ui/templates/utilisation-reports/_macros/report-reconciliation-table.njk
```diff
@@ -1,57 +1,99 @@
...
+        attributes: {
+          formaction: ["?_csrf=", csrfToken, "&form-button=completed&submissionMonth=", submissionMonth, "&submissionYear=", submissionYear] | join,
+          "data-cy": "mark-report-as-completed-button"
+        }
+      }) }}
+      {{ govukButton({
+        text: "Mark as not completed",
+        classes: "govuk-button--secondary govuk-!-margin-right-1",
```
toddym42, line 91: I think we shouldn't need this margin either?

trade-finance-manager-ui/server/controllers/utilisation-reports/update-utilisation-report-status/index.ts
JoshBinns2000: Unfortunately I think this one does need a comment as it is a bit of a hack

trade-finance-manager-api/src/v1/validation/route-validators/update-report-status-payload-validation/index.test.ts
```diff
@@ -0,0 +1,215 @@
...
+      // Arrange
+      const report = { id: '123' };
+      const body = getValidPayloadBody({ report });
+      const req = createRequest({ body });
+
+      // Act
+      const errors = await getUpdateReportStatusPayloadValidationResult(req);
+
+      // Assert
+      expect(errors.length).toEqual(1);
```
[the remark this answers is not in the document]
JoshBinns2000: I did think about this when I first wrote them, but decided not to (largely out of laziness, but also because extracting the error message was a bit of a pain if the tests failed). I will update these tests though now that they are passing

toddym42 (approved):

PR #2491, opened 2023-12-28
fix(DTFS2-6526): focus user to the table header button if it was clicked just now

avaitonis:
## Introduction :pencil2:
Activating any table header button will cause the table column to become filtered by ascending or descending. Keyboard-only users will be repositioned to the beginning of the main content, and this includes screen reader users. This could be problematic for the user as the user will need to re-navigate to the table content.

- Added hidden form input `formId` to detect what form was just activated. Was it sorting, search or pagination. If `formId` is `deals-table-sorting` then we know that sorting event just happened and we can `autofocus` sort column button.
- Added `sortButtonWasClicked` NJK variable to know if this is new sorting event happened.
- Ensured that autofocus doesn't happen for `default` sort and other navigations, like pagination and search.

Simplified how we get active `order` direction, `asc` or `desc`. This allows to have more form inputs controllers/deals/index.js line 61


trade-finance-manager-ui/server/controllers/deals/index.js
```diff
@@ -48,8 +49,7 @@ const queryDeals = async (req, res) => {
   let activeSortByField = '';
   let searchString = '';
   const { userToken } = req.session;
-
-  delete req.body._csrf;
+  const sortButtonWasClicked = (req.body.formId && req.body.formId === 'deals-table-sorting') ? true : false;
```
abhi-markan, line 52: Can be simplified to following statement

```javascript
 const sortButtonWasClicked = req.body?.formId === 'deals-table-sorting'
```

trade-finance-manager-ui/server/controllers/deals/index.js
```diff
@@ -58,13 +58,12 @@ const queryDeals = async (req, res) => {
   const queryParams = { searchString };
 
   if (req.body.descending || req.body.ascending) {
-    const sortBy = Object.getOwnPropertyNames(req.body)[0];
-    const sortByField = req.body[sortBy];
-    activeSortByField = sortByField;
+    const order = req.body.ascending ? 'ascending' : 'descending';
```
abhi-markan, line 61: Might be ideal to use optional chaining

```javascript
  const order = req.body?.ascending ? 'ascending' : 'descending';
```

abhi-markan (changes requested):
Few minor comments to address please.

vamshidharukef (dismissed):
LGTM

trade-finance-manager-ui/server/controllers/deals/index.js
avaitonis: Thanks, simplified.

trade-finance-manager-ui/server/controllers/deals/index.js
avaitonis: Thanks, changed.

trade-finance-manager-ui/server/controllers/deals/index.js
```diff
@@ -48,8 +49,7 @@ const queryDeals = async (req, res) => {
   let activeSortByField = '';
   let searchString = '';
   const { userToken } = req.session;
-
-  delete req.body._csrf;
+  const sortButtonWasClicked = (req.body?.formId === 'deals-table-sorting') ? true : false;
```
abhi-markan, line 52: const sortButtonWasClicked = req.body?.formId

abhi-markan (changes requested):
One minor amend please.

abhi-markan (approved):

trade-finance-manager-ui/server/controllers/deals/index.js
avaitonis: I simplified line 52.

Resolving this comment as PR was approved 1 hour ago.

PR #2492, opened 2023-12-29
fix(DTFS2-6613): fix validation error on loan financial details page

vamshidharukef:
## Introduction :pencil2:
On Loan financial details page when changing currency other than supplier currency there is a validation error. As currency is an object and doesn't save as string as Mongo DB can't change type is the root cause when changing currency and saving.

-  check condition and if empty string 
-  delete the currency property

Added unit test


portal/server/routes/contract/bond/index.js
```diff
@@ -220,8 +220,8 @@ router.post('/contract/:_id/bond/:bondId/financial-details', async (req, res) =>
 });
 
 router.post('/contract/:_id/bond/:bondId/financial-details/save-go-back', provide([BOND]), async (req, res) => {
-  const bondPayload = filterBondFinancialDetailsPayload(req.body);
-  return saveFacilityAndGoBackToDeal(req, res, bondPayload);
+  const sanitizedPayload = filterBondFinancialDetailsPayload(req.body, bondFinancialDetailsPayloadProperties, ['currency']);
```
avaitonis, line 223: Sonarcloud found real issue here.
Function `filterBondFinancialDetailsPayload` expects just one parameter, no need to have change function call at line 223.

portal/server/helpers/constructPayload.js
```diff
@@ -5,16 +5,24 @@
  * @param {Object} body Request body (req.body)
  * @param {Array} properties Interested properties to be added into payload
  * @param {Boolean} csrf Include CSRF token, defaulted to `true`
+ * @param {Array} unsetIfPropertyIsEmpty The condition to determine whether to delete the property
```
avaitonis, line 8: Comment is a bit confusing, not sure if we can call it `condition`.

Maybe this is better `list of property names that can't have empty value as they will have object value later.`

portal/server/helpers/constructPayload.js
```diff
@@ -5,16 +5,24 @@
...
  * @returns {Object} Payload
  */
-const constructPayload = (body, properties, csrf = true) => {
+const constructPayload = (body, properties, unsetIfPropertyIsEmpty = [], csrf = true) => {
   let payload = {};
-
+  const newObj = { ...body };
```
avaitonis, line 13: Using very generic variable names is not good practice, because variable purpose is not clear.

Maybe `bodyCopy` is better? feel free to think for better name with clear purpose.

portal/server/helpers/constructPayload.test.js
```diff
@@ -88,8 +88,25 @@ describe('Unit test cases for constructPayload method', () => {
...
   });
+
+  it('Should delete currency with empty values', () => {
+    const expected = {
+      currentCurrencyValue: '',
+      newCurrencyValue: 'value',
+      anotherCurrenyValue: '',
+    };
+    const unsetIfPropertyIsEmpty = ['currentCurrencyValue', 'newCurrencyValue', 'anotherCurrenyValue'];
+    // const returned = constructPayload(mockBody, payloadProperties, false);
```
avaitonis, line 103: Purpose of this unit test file is to test function `constructPayload`, but you are not calling this function, please uncomment line 103.

You can create new `mockBody`, it is your `input`. Value for `expected` can be object with single value, it is your `expected output`.

```
const mockBody = {
      currentCurrencyValue: '',
      newCurrencyValue: 'value',
      anotherCurrenyValue: '',
};
const expected = {
      newCurrencyValue: 'value'
};

File: portal/server/helpers/constructPayload.js (hunk 2 above)
Comment (abhi-markan) [reply portal/server/helpers/constructPayload.js:8]: * @param should be above Boolean parameter.
* `propertyEmpty` as a variable name will be clear.

File: portal/server/helpers/constructPayload.js [hunk 5]
```diff
@@ -5,16 +5,24 @@
...
   // Return empty payload upon void mandatory arguments
   if (!body || !properties) {
     return payload;
   }
 
+  // Currency will be an object, don't save as empty string, because Mongo DB can't change type.
+  for (const property of unsetIfPropertyIsEmpty) {
```
abhi-markan, line 20: Refactor this if condition please.

abhi-markan: We should prevent an empty property in a payload being added to the database and a validation error should be thrown on the same page preventing a flawed payload being added to the database.

abhi-markan (changes requested):

portal/server/helpers/constructPayload.js
avaitonis: Error exists for BSS and EWCS facilities, for them at the moment we allow skipping some values in forms and coming back to them later. **Also some fields are optional**.

I agree that we could try avoid saving empty values to DB, not sure it will require a lot changes, maybe code is expecting empty values.
@abhi-markan throwing errors would be big change in BSS/EWCS behavior, do we want it? is it bug related?

portal/server/routes/contract/bond/index.js
vamshidharukef: Changed to `const sanitizedPayload = filterBondFinancialDetailsPayload(req.body);`

portal/server/helpers/constructPayload.js
vamshidharukef: - Changed the comment to List of property names that can't have empty value as they will have object value later, defaulted to true
- @param is now boolean `@param {Boolean} propertyEmpty`
- Changed the variable name to `propertyEmpty`

portal/server/helpers/constructPayload.js
vamshidharukef: Changed to `bodyCopy`

portal/server/helpers/constructPayload.test.js
vamshidharukef: Uncommented  line 103 and created new `mockBodyWithEmptyValues`, `payloadPropertiesForEmptyValues`

portal/server/helpers/constructPayload.js
vamshidharukef: Refactored `if` condition

portal/server/helpers/constructPayload.js
```diff
@@ -5,16 +5,24 @@
  * @param {Object} body Request body (req.body)
  * @param {Array} properties Interested properties to be added into payload
  * @param {Boolean} csrf Include CSRF token, defaulted to `true`
+ * @param {Boolean} propertyEmpty List of property names that can't have empty value as they will have object value later, defaulted to True
  * @returns {Object} Payload
  */
-const constructPayload = (body, properties, csrf = true) => {
+const constructPayload = (body, properties, propertyEmpty, csrf = true) => {
```
avaitonis, line 11: Because we are introducing new variable and this function was used before it is tempting to have default value for `propertyEmpty` variable.

Most current functions are passing just 2 variables, so `propertyEmpty` was set to empty (aka `false`) state. If we want to prevent all empty values and do global change, we could set default value to `true`.

portal/server/helpers/constructPayload.js
```diff
@@ -5,16 +5,24 @@
...
   let payload = {};
-
+  const bodyCopy = { ...body };
   // Return empty payload upon void mandatory arguments
   if (!body || !properties) {
     return payload;
   }
 
+  // Currency will be an object, don't save as empty string, because MongoDB can't change the type.
```
avaitonis, line 19: Not sure if we should mention `currency` in comment if we go with generic change. Remove empty properties if flag is set to true. This is because some fields might change type, for example currency field, and MongoDB doesn't allow type change.

portal/server/helpers/constructPayload.js
```diff
@@ -5,16 +5,24 @@
  * @param {Object} body Request body (req.body)
  * @param {Array} properties Interested properties to be added into payload
  * @param {Boolean} csrf Include CSRF token, defaulted to `true`
+ * @param {Boolean} propertyEmpty List of property names that can't have empty value as they will have object value later, defaulted to True
```
abhi-markan, line 8: please re-word to `Whether property can be empty in a payload` and rename the variable to `canPropertyEmpty`.

portal/server/helpers/constructPayload.js
abhi-markan: Agree set default value to `false`

```javascript
canPropertyEmpty = false
```

portal/server/helpers/constructPayload.js
abhi-markan: Since the function is generic in nature please re-write the comment.

portal/server/helpers/constructPayload.js
```diff
@@ -5,16 +5,24 @@
...
  * @returns {Object} Payload
  */
-const constructPayload = (body, properties, csrf = true) => {
+const constructPayload = (body, properties, propertyEmpty, csrf = true) => {
   let payload = {};
-
+  const bodyCopy = { ...body };
```
abhi-markan, line 13: Please rename `bodyCopy` to `payload`.

portal/server/helpers/constructPayload.js
```diff
@@ -5,16 +5,24 @@
...
   // Return empty payload upon void mandatory arguments
   if (!body || !properties) {
     return payload;
   }
 
+  // Currency will be an object, don't save as empty string, because MongoDB can't change the type.
+  for (const property of properties) {
+    if (propertyEmpty && bodyCopy[property] === '') {
```
abhi-markan, line 21:
```javascript
if (canPropertyEmpty && !payload[property]) {
...
}
```

portal/server/routes/contract/bond/index.js
```diff
@@ -189,16 +189,17 @@ const bondFinancialDetailsPayloadProperties = [
...
 
 const filterBondFinancialDetailsPayload = (body) => {
-  const payload = constructPayload(body, bondFinancialDetailsPayloadProperties);
-
-  if (payload.currencySameAsSupplyContractCurrency === 'true') {
-    delete payload.conversionRate;
-    delete payload['conversionRateDate-day'];
-    delete payload['conversionRateDate-month'];
-    delete payload['conversionRateDate-year'];
+  const propertyEmptyFlag = true; // Set to true if you want to validate empty properties
```
abhi-markan, line 192: Rename to `canPropertyEmpty` please.

portal/server/routes/contract/bond/index.js
```diff
@@ -189,16 +189,17 @@ const bondFinancialDetailsPayloadProperties = [
...
+  const sanitizedPayload = constructPayload(body, bondFinancialDetailsPayloadProperties, propertyEmptyFlag);
+
+  if (sanitizedPayload.currencySameAsSupplyContractCurrency === 'true') {
```
abhi-markan, line 195:
```javascript
if (sanitizedPayload.currencySameAsSupplyContractCurrency) {
...
}
```

abhi-markan (changes requested):
Few comments to address.

portal/server/helpers/constructPayload.js
vamshidharukef: Changed to `'Whether property can be empty in a payload' `and renamed variable to `canPropertyBeEmpty`

portal/server/helpers/constructPayload.js
vamshidharukef: set `canPropertyEmpty = false` and re-written comment

portal/server/routes/contract/bond/index.js
vamshidharukef: Renamed to `canPropertyEmpty`

portal/server/helpers/constructPayload.js
abhi-markan: Please rename bodyCopy to payload.

portal/server/helpers/constructPayload.js
```diff
@@ -5,16 +5,24 @@
...
-
+  const bodyCopy = { ...body };
   // Return empty payload upon void mandatory arguments
   if (!body || !properties) {
     return payload;
   }
 
+  // Remove empty properties if flag is set to true, for example currency field, and MongoDB doesn't allow type change.
+  for (const property of properties) {
+    if (canPropertyBeEmpty === false && bodyCopy[property] === '') {
```
abhi-markan, line 21: if (!canPropertyBeEmpty && !bodyCopy[property])

portal/server/helpers/constructPayload.js
```diff
@@ -5,16 +5,24 @@
...
+const constructPayload = (body, properties, canPropertyBeEmpty = false, csrf = true) => {
   let payload = {};
-
+  const bodyCopy = { ...body };
   // Return empty payload upon void mandatory arguments
   if (!body || !properties) {
     return payload;
   }
 
+  // Remove empty properties if flag is set to true, for example currency field, and MongoDB doesn't allow type change.
```
abhi-markan, line 19: flag set to `false` rather?

abhi-markan (changes requested):

portal/server/helpers/constructPayload.js
vamshidharukef: flag set to `false`

portal/server/helpers/constructPayload.js
vamshidharukef: - on line 12 `payload` name is already used. Please suggest another name for `bodyCopy`
- When used `if (!canPropertyBeEmpty && !bodyCopy[property])` unit tests fail so I changed to i`f (!canPropertyBeEmpty && bodyCopy[property] === '')` and unit test are passing.

portal/server/helpers/constructPayload.js
avaitonis: FYI @abhi-markan we already have variable called `payload`, we construct it using list of `properties` so not all data from body is passed.

`bodyCopy` could be renamed to make more sense or code could be changed to avoid need of copying body.

code change 
```
properties
    .filter((property) => property in body)
    .filter((property) => canPropertyBeEmpty || body[[property]] !=='') 
    .forEach((property) => {
      payload = {
        ...payload,
        [property]: body[[property]],
      };
    });
```

vamshidharukef: removed `bodyCopy` and added `.filter((property) => canPropertyBeEmpty || body[[property]] !== '')`

portal/server/routes/contract/bond/index.js
vamshidharukef: Condition `if (sanitizedPayload.currencySameAsSupplyContractCurrency)` is not set to true then the condition is not evaluated.

portal/server/helpers/constructPayload.js
avaitonis: Checked and default value now is false.

portal/server/helpers/constructPayload.js
avaitonis: I checked and because of new minimalistic code comment was removed.

portal/server/helpers/constructPayload.js
```diff
@@ -25,6 +25,7 @@ const constructPayload = (body, properties, csrf = true) => {
   // Property insertion
   properties
     .filter((property) => property in body)
+    .filter((property) => canPropertyBeEmpty || body[[property]] !== '')
```
abhi-markan, line 28: `body[[property]] !== ''` can be `!body[[property]]`

portal/server/helpers/constructPayload.test.js
```diff
@@ -25,6 +25,27 @@ const payloadProperties = [
   'passwordConfirm',
 ];
 
+const mockBodyWithEmptyValues = {
```
abhi-markan, line 28: Can an assertion be added with currency and supplier currency object?

abhi-markan (changes requested):

portal/server/helpers/constructPayload.test.js
vamshidharukef: Added tests for above.

abhi-markan (approved):

abhi-markan (approved):

PR #2548, opened 2024-01-03
fix: bicep refactor

francescastocco:
## Introduction :pencil2:
Refactoring the bicep I wrote when we changed connection strings to environment variables

- Set all tfm-api settings in `trade-finance-manager-api-calculated-variables`, instead of setting it twice
- Rename `tfmApiConnectionStrings` module to `tfmApiCalculatedVariables`
- Add condition to `webapp.bicep` to only set `appSettings` if not empty


OxonianCambion (approved):
Nice.

PR #2552, opened 2024-01-04
feat(FN-1613): introduce primary navigation constant

JoshBinns2000:
## Introduction :pencil2:
Adds a new constant to TFM UI which contains the names of the primary navigation items and updates all occurrences of `activePrimaryNavigation` to use the new constant.


trade-finance-manager-ui/server/constants/primary-navigation-items.ts
```diff
@@ -0,0 +1,5 @@
+export const PRIMARY_NAVIGATION_ITEMS = {
```
toddym42, line 1: I'd lose `_ITEMS` from the name as it doesn't really add anything. Maybe just `PRIMARY_NAVAGATION` or `PRIMARY_NAVIGATION_KEY`? Not a blocker though if you'd prefer to keep as is.

toddym42 (approved):
Thanks for doing this!

PR #2553, opened 2024-01-04
fix(DTFS2-6937): Add prettier check to git hook

abhi-markan:
## Introduction :pencil2:
Ensure `lint-staged` is executed on staged file by hooking into `pre-commit` git hook.

* Introduced `prettier:all` and `prettier:fix:all` npm scripts (root).
* Added `'operator-linebreak': ['error', 'after']` to all `.eslintrc.js` to resolve conflicting Prettier and ESLint rule. 
* Introduced Git hook `pre-commit` for executing `lint-staged` commands on staged files.
* Added prettier check to CI using `npm run prettier:all`.
* Unified `eslintrc.js`, `.eslintignore`, `.prettierrc.json` and `.prettierignore` configuration across all the micro-services.

* Dependencies updates
* Updated SRI for JS files across `portal-ui`, `gef-ui` and `trade-finance-manager-ui`.


azure-functions/acbs-function/.prettierrc.json
```diff
@@ -0,0 +1,18 @@
+{
```
toddym42, line 1: Could we instead remove all the `.prettierignore` and `.prettierrc.json` files from the individual sub-projects, and just have them all fallback to the versions in the root of the main `dtfs2` project? I think this would happen automatically if these are removed, so wouldn't require any extra config. This would ensure our rules are always consistent across projects.

azure-functions/acbs-function/.prettierrc.json
```diff
@@ -0,0 +1,18 @@
...
+  "arrowParens": "always",
+  "bracketSameLine": false,
+  "bracketSpacing": true,
+  "embeddedLanguageFormatting": "auto",
+  "htmlWhitespaceSensitivity": "css",
+  "insertPragma": false,
+  "jsxSingleQuote": false,
+  "printWidth": 160,
```
toddym42, line 9: A `printWidth` of 160 feels quite wide to me, and I often see functions and types that prettier formats to a wider width than I'd like. What your thoughts on adjusting this to something like 120?

abhi-markan: Good suggestion, have adjusted to `120`.

azure-functions/acbs-function/.prettierrc.json
abhi-markan: Good shout, have removed.
Have also removed `.eslintignore` and will look into removing `.eslintrc.js` (where duplicate only)

.eslintrc.js
```diff
@@ -0,0 +1,76 @@
+const baseParserOptions = {
+  ecmaVersion: 2022,
+};
+
+module.exports = {
+  extends: ['airbnb-base', 'prettier'],
```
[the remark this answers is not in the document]
abhi-markan: Have added `plugin:prettier/recommended`

.eslintrc.js
```diff
@@ -0,0 +1,76 @@
...
+    jest: true,
+    browser: true,
+  },
+  root: true,
+  ignorePatterns: ['**/node_modules/**'],
+  parserOptions: baseParserOptions,
+  overrides: [
+    {
+      files: ['*.ts'],
+      extends: ['airbnb-base', 'plugin:@typescript-eslint/recommended-type-checked'],
```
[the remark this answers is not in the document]
abhi-markan: Have added `plugin:prettier/recommended`

external-api/src/database/mongo-client.ts
```diff
@@ -1,6 +1,8 @@
-const { MongoClient } = require('mongodb'); // eslint-disable-line @typescript-eslint/no-var-requires
+// eslint-disable-line @typescript-eslint/no-var-requires
```
avaitonis, line 1: Eslint comment is not required or should be moved to other place.

abhi-markan: Have removed.

PR #2554, opened 2024-01-05
feat(FN-1290): store reporting schedule against banks

francescastocco:
## Introduction :pencil2:
We need to know the reporting schedule for each bank

- Added reporting schedule to mock banks
- Added reporting schedule to db model


dtfs-central-api/src/types/db-models/banks.d.ts
```diff
@@ -1,5 +1,10 @@
 import { WithId } from 'mongodb';
 
+type ReportPeriodSchedule = {
```
JoshBinns2000, line 3: I think this should be `UtilisationReportPeriodSchedule`

JoshBinns2000 (commented):
Looks good code-wise. Only one comment re type name

utils/mock-data-loader/banks/index.js
```diff
@@ -11,6 +11,20 @@ const BANKS = [
...
+      { startMonth: 4, endMonth: 4 },
+      { startMonth: 5, endMonth: 5 },
+      { startMonth: 6, endMonth: 6 },
+      { startMonth: 7, endMonth: 7 },
+      { startMonth: 8, endMonth: 8 },
+      { startMonth: 9, endMonth: 9 },
+      { startMonth: 10, endMonth: 10 },
+      { startMonth: 11, endMonth: 11 },
+      { startMonth: 12, endMonth: 12 },
+    ],
```
toddym42, line 27: Could save this as a `MONTHLY_REPORT_PERIOD_SCHEDULE` or similar to simplify adding this everywhere.

toddym42 (approved):

dtfs-central-api/src/types/db-models/banks.d.ts
francescastocco: I thought about it but if we stored any other reporting schedules they could also be the same type

toddym42 (approved):

PR #2556, opened 2024-01-05
fix(FN-1615): convert repository files to typescript

JoshBinns2000:
## Introduction :pencil2:
We recently introduced repository files to `dtfs-central-api` for database related logic. This happened before we started using typescript, however, and so the existing code has two issues:

1. Some of the types we have are not consistent between files due to the `.js` files not using any type checking (see the changes to `utilisation-data-repo.ts`)
2. When we begin the migration to SQL, the lack of typescript files will make it difficult to see how updating our `db-models` with any new schemas will affect existing functionality

- Converts the files in `services/repositories` to typescript files, including all test files.
- Adds some new types (including constants which are used to derive some of these types)

## Miscellaneous
Fixes a flakey test issue in `services/repositories/utilisation-reports-repo/update-report-status.test.ts` related to the use of `new Date()` and the `portal-api` api test for the `POST /v1/utilisation-reports` endpoint.

### Introduces a new piece of logic for how we handle `exchange rates`:
Previously, if two currencies matched, the `exchange rate` was set to `null`. However, this introduces the need for more complicated logic whenever two values are combined, as you need to check if the `exchange rate` value is `null` before accessing the value. Instead, we now make sure the value is always a `number` by changing the `dtfs-central-api` to set this value to `1` (instead of `null`) if no value is provided (ie. the currency is the same as whatever the `base currency` is).


dtfs-central-api/src/services/repositories/utilisation-reports-repo/utilisation-reports-repo.ts
```diff
@@ -0,0 +1,46 @@
...
+      surname: uploadedByUser.surname,
+    },
+  };
+
+  const utilisationReportDetailsCollection = await db.getCollection(DB_COLLECTIONS.UTILISATION_REPORTS);
+  const savedDetails = await utilisationReportDetailsCollection.insertOne(utilisationReportInfo);
+  return { reportId: savedDetails.insertedId.toString(), dateUploaded: utilisationReportInfo.dateUploaded };
+};
+
+export const getUtilisationReportDetailsByBankIdMonthAndYear = async (bankId: string, month: number, year: number): Promise<UtilisationReport | null> => {
```
JoshBinns2000, line 32: We need to explicitly define the return types for these functions as the `getCollection` returns something like `Promise<Collection<WithoutId<DbCollection>>>`, but the call of `collection.find(...).toArray()` (for example) then returns `WithId<WithoutId<DbCollection>>[]` which is, of course, the same as `DbCollection[]`. In order for typescript to recognise this, however, I need to explicitly provide the return type. I have made this change to almost all the functions that query the database and return something, so any suggestions for a better way to do this will be applied to all those locations

dtfs-central-api/src/types/portal/portal-session-user.d.ts
```diff
@@ -0,0 +1,13 @@
...
+export type PortalSessionUser = Omit<PortalUser,
+  | 'bank'
+  | 'salt'
+  | 'hash'
+  | 'resetPwdToken'
+  | 'resetPwdTimestamp'
+  | 'loginFailureCount'
+  | 'sessionIdentifier'> & {
+  bank: SessionBank;
+};
```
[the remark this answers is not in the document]
JoshBinns2000: (Another thing I need to clarify here is what data we actually use for `user.bank`, as before now I have only ever seen `bank: { id: string; name: string; }` used)

dtfs-central-api/src/types/portal/portal-session-user.d.ts
```diff
@@ -0,0 +1,14 @@
+import { Prettify } from '../types-helper';
+import { PortalUser } from '../db-models/portal-user';
+import { SessionBank } from '../session-bank';
+
+export type PortalSessionUser = Prettify<Omit<PortalUser,
```
toddym42, line 5: I think we should use `Pick` instead of `Omit` to define the type here, because:
* it makes it easier to see what properties are in the type, not requiring you to look at `PortalUser` and subtract these properties
* it's more reflective of how we're actually constructing the portal session user from the portal user
* it will prevent new properties magically getting added to this type if `PortalUser` is ever updated

I'm proposing the same change for `TfmSessionUser` as part of my work for FN-1051.

dtfs-central-api/src/types/types-helper.ts
```diff
@@ -1 +1,5 @@
 export type ValuesOf<T> = T[keyof T];
+
+export type Prettify<T> = {
+  [K in keyof T]: T[K];
+};
```
toddym42, line 5: Do you need to add `& {}` to this type? https://www.totaltypescript.com/concepts/the-prettify-helper

Also, maybe worth adding this link (or wherever you got this from) as it's not immediately obvious how this works.

dtfs-central-api/src/types/portal/portal-session-user.d.ts
toddym42: If you change to `Pick` you might be able to remove `Prettify`? Not 100% sure on that though

JoshBinns2000: Seems like you still need `Prettify`, but `Pick` is much clearer here

dtfs-central-api/src/types/utilisation-reports.d.ts
```diff
@@ -48,3 +49,9 @@ export type UpdateUtilisationReportStatusInstructions = {
   status: UtilisationReportReconciliationStatus;
   filter: ReportFilter;
 };
+
+type UtilisationReportHeader = ValuesOf<typeof UTILISATION_REPORT_HEADERS>;
+
+export type ReportData = {
+  [P in UtilisationReportHeader]: P extends `${string}currency` ? Currency : string;
```
toddym42, line 56: You could change this to match on `${string}currency${string}` which will then cover `currency` being anywhere in the string (including both the start and end, as `${string}` could be the empty string).

dtfs-central-api/src/types/utilisation-reports.d.ts
```diff
@@ -48,3 +49,9 @@ export type UpdateUtilisationReportStatusInstructions = {
   status: UtilisationReportReconciliationStatus;
   filter: ReportFilter;
 };
+
+type UtilisationReportHeader = ValuesOf<typeof UTILISATION_REPORT_HEADERS>;
+
+export type ReportData = {
```
toddym42, line 55: As this type is now in a general types file could we give it a less general name? Maybe `UtilisationReportRawCsvData` to distinguish from other data?

dtfs-central-api/src/services/repositories/banks-repo.ts
```diff
@@ -0,0 +1,14 @@
+import db from '../../drivers/db-client';
+import { DB_COLLECTIONS } from '../../constants';
+import { Bank } from '../../types/db-models/banks';
+
+export const getAllBanks = async (): Promise<Bank[] | null> => {
```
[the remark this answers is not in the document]
JoshBinns2000: I think you are correct. `find()` returns some `FindCursor` thingy which I think will just give an empty array from `toArray()` if it can't find anything

dtfs-central-api/src/services/repositories/utilisation-reports-repo/utilisation-reports-repo.ts
```diff
@@ -0,0 +1,46 @@
...
+  const utilisationReportDetailsCollection = await db.getCollection(DB_COLLECTIONS.UTILISATION_REPORTS);
+  return await utilisationReportDetailsCollection.findOne({ 'bank.id': bankId, month, year });
+};
+
+export const getUtilisationReportDetailsByBankId = async (bankId: string) => {
```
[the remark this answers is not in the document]
JoshBinns2000: Sure - the reason I didn't was that typescript inferred the correct type due to the call of `sortBy`, but no harm in defining it explicitly

dtfs-central-api/src/types/db-models/portal-user.d.ts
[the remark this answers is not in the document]
JoshBinns2000: I agree - this was a mistake of mine when I wrote this file. It should be called `users.d.ts`, but my brain defaulted to the expected name of `portal-users`

dtfs-central-api/src/types/db-models/portal-user.d.ts
```diff
@@ -0,0 +1,25 @@
+import { WithId } from 'mongodb';
+import { Role } from '../portal/role';
+import { Bank } from './banks';
+
+export type PortalUser = WithId<{
+  _id: string;
+  'user-status': string;
+  disabled: string;
```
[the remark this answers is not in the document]
JoshBinns2000: I wasn't sure whether it was a `boolean` or a `string` - this was one of the fields I wanted to clarify :)

dtfs-central-api/src/types/db-models/portal-user.d.ts
```diff
@@ -0,0 +1,25 @@
...
+  username: string;
+  salt: string;
+  hash: string;
+  resetPwdToken?: string;
+  resetPwdTimestamp?: string;
+  lastLogin?: number;
+  loginFailureCount?: number;
+  sessionIdentifier?: string;
+  signInLinkSendDate?: Date;
+  signInLinkSendCount: number;
```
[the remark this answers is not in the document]
JoshBinns2000: I was mainly guessing that these fields were optional, but probably

dtfs-central-api/src/types/utilisation-reports.d.ts
JoshBinns2000: I didn't do that specifically because there could be a key like "X currency exchange rate", which would not be a `Currency`. I thought it was more logical that any string ending with "currency" would probably itself be a `Currency`

dtfs-central-api/src/services/repositories/utilisation-reports-repo/utilisation-reports-repo.ts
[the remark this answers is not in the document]
toddym42: A sad day for WebStorm. It wasn't able to infer it quite as well 😢

dtfs-central-api/src/types/db-models/portal-user.d.ts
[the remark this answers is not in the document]
toddym42: 👍 

Forgot to mention, but I'd preferably still call the type `PortalUser` or at least write a comment to explain it's portal only, just so we still document that.

dtfs-central-api/src/types/utilisation-reports.d.ts
toddym42: Ah, yeah fair point.

dtfs-central-api/src/services/repositories/utilisation-reports-repo/utilisation-reports-repo.ts
JoshBinns2000: I think it only happens because we have been "smart" in including the `_id` field in the `db-models` explicitly via `WithId`, so we then have to remove the `_id` for the collection type definition. But then, the `collection` adds it back once a method on the `collection` is called which returns something. We would get better inferred types for things like `getCollection().method()` if we removed the `WithId` from the `db-models` type, but we would then have to add `WithId` wherever the `db-models` types are used outside of the `collection` definitions, so I think your type definitions in `getCollection` are fine as they are

dtfs-central-api/src/types/utilisation-reports.d.ts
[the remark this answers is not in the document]
JoshBinns2000: Decided to go with `HeaderKey`

dtfs-central-api/src/services/repositories/utilisation-reports-repo/utilisation-reports-repo.ts
```diff
@@ -0,0 +1,85 @@
...
+export const getUtilisationReportDetailsById = async (_id: string): Promise<UtilisationReport | null> => {
+  const collection = await db.getCollection(DB_COLLECTIONS.UTILISATION_REPORTS);
+  return await collection.findOne({ _id: new ObjectId(_id) });
+};
+
+export const getOpenReportsBeforeReportPeriodForBankId = async (reportPeriodStart: ReportPeriodStart, bankId: string): Promise<UtilisationReport[]> => {
+  const collection = await db.getCollection(DB_COLLECTIONS.UTILISATION_REPORTS);
+
+  return await collection
+    .aggregate<UtilisationReport>([
```
JoshBinns2000, line 53: Without passing in this type generic, the return type of `aggregate(...).toArray()` is `Promise<Document[]>` which cannot be assigned to `Promise<UtilisationReport[]>`. Explicitly passing this generic does mean that the return type of `getOpenReportsBeforeReportPeriodForBankId` is inferred correctly, but I still added it explicitly seeing as we have done that for the other functions in this file

toddym42: Sounds good. The other advantage of explicit return types are:

* easier to see the return type just looking at the function
* code will shout at you if you all paths don't return what you expected the return type to be

I know this is a bit of a contentious topic in the TS community (whether to add explicit return types or not), but for what's it's worth, I prefer to add it.

toddym42 (approved):
LGTM. Nice work : )

PR #2558, opened 2024-01-08
feat(FN-1685): fix csv parsing for rows with empty data at end of row

ChristianMcc:
## Introduction :pencil2:
Empty cells at the end of a row would not be processed by the excel library and so we would not get the cell addresses or empty values for these cells when processing

Keep count of how many header columns we have and add in any empty values with cell addresses so we can show descriptive error messages later on in the process.

### DIFF
--- a/portal/server/utils/csv-utils.js
+++ b/portal/server/utils/csv-utils.js
@@ -201,7 +201,7 @@ const extractCsvData = async (file) => {
   let fileBuffer;
 
   try {
-    if (file.mimetype === 'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet') {
+    if (file.mimetype === 'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet' || file.mimetype === 'application/vnd.ms-excel') {
       // Read the .xlsx file using exceljs
       const workbook = new ExcelJS.Workbook();
       await workbook.xlsx.load(file.buffer, { sheetStubs: true }).then(async () => {
--- a/portal/server/utils/csv-utils.js
+++ b/portal/server/utils/csv-utils.js
@@ -13,6 +13,15 @@ const columnIndexToExcelColumn = (index) => {
   return result;
 };
 
+const excelColumnToColumnIndex = (column) => {
+  let result = 0;
+  for (let i = 0; i < column.length; i += 1) {
+    result *= 26;
+    result += column.charCodeAt(i) - 64;
+  }
+  return result - 1;
+};
+
 /**
  * Extracts the value in the cell of an excel cell and removes any new lines or commas so that it doesn't affect parsing as a csv.
  * @param {Object} cell - excel cell.
@@ -21,12 +30,13 @@ const columnIndexToExcelColumn = (index) => {
 const extractCellValue = (cell) => {
   /* eslint-disable-next-line no-underscore-dangle */
   const cellValue = cell.value?.result ?? cell._value?.result ?? cell.value;
-  const cellValueWithoutNewLines = typeof cellValue === 'string'
-    ? cellValue
-      .replace(/\r\n|\r|\n/g, ' ')
-      .replace(/,/g, '')
-      .trim()
-    : cellValue;
+  const cellValueWithoutNewLines =
+    typeof cellValue === 'string'
+      ? cellValue
+          .replace(/\r\n|\r|\n/g, ' ')
+          .replace(/,/g, '')
+          .trim()
+      : cellValue;
   return cellValueWithoutNewLines;
 };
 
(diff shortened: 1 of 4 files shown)


portal/server/utils/csv-utils.js
```diff
@@ -13,6 +13,15 @@ const columnIndexToExcelColumn = (index) => {
   return result;
 };
 
+const excelColumnToColumnIndex = (column) => {
```
toddym42, line 16: Could we add a JSDoc with the type of `column` and the return type.

portal/server/utils/csv-utils.js
```diff
@@ -13,6 +13,15 @@ const columnIndexToExcelColumn = (index) => {
...
+  let result = 0;
+  for (let i = 0; i < column.length; i += 1) {
+    result *= 26;
+    result += column.charCodeAt(i) - 64;
```
toddym42, line 20: Could we give some names to the `26` and `64` constants to explain what they are?

portal/server/utils/csv-utils.js
```diff
@@ -43,11 +56,31 @@ const parseXlsxToCsvArrays = (worksheet) => {
...
         rowDataWithCellAddresses.push(`${cellValue}`);
+        headerCount += 1;
       } else {
         rowDataWithCellAddresses.push(`${cellValue}-${cell.address}`);
+        lastAddress = cell.address;
+        columnCount += 1;
       }
     });
+
+    if (firstRow === false) {
```
toddym42, line 67: Can just be `if (!firstRow)`

portal/server/utils/csv-utils.js
```diff
@@ -43,11 +56,31 @@ const parseXlsxToCsvArrays = (worksheet) => {
...
+      // If the row has no data in the final columns of the row, we need to fill in the empty cells with
+      // empty strings and cell addresses so that the csv parser library can parse the data correctly
+      while (columnCount < headerCount) {
+        rowData.push('');
+        // calculate the cell address here given the last address
+        const lastAddressMatch = lastAddress.match(CELL_ADDRESS_REGEX);
+        const [lastColumn, lastRow] = lastAddressMatch.slice(1);
+        const lastColumnIndex = excelColumnToColumnIndex(lastColumn);
+        const newColumn = columnIndexToExcelColumn(parseInt(lastColumnIndex, 10) + 1);
```
toddym42, line 76: If we give a (JSDoc) return type of `excelColumnToColumnIndex` I think we can remove `parseInt` here?

portal/server/utils/csv-utils.test.js
```diff
@@ -21,6 +22,20 @@ describe('csv-utils', () => {
...
+      const columnIndex = excelColumnToColumnIndex('C');
+
+      expect(columnIndex).toBe(2);
+    });
+
+    it('returns the correct index for a column above 26', async () => {
+      const columnIndex = excelColumnToColumnIndex('AE');
+
+      expect(columnIndex).toBe(30);
+    });
```
toddym42, line 36: What's the significance of 26 here?

portal/server/utils/csv-utils.test.js
```diff
@@ -21,6 +22,20 @@ describe('csv-utils', () => {
...
+  });
+
```
toddym42, line 38: Could we add any tests to check the new logic does what we expect (adding empty strings to empty cells)

portal/server/utils/csv-utils.test.js
ChristianMcc: This just represents wrapping from a single character column code to multiple characters. I can expand the test case name to include this if that's helpful?

portal/server/utils/csv-utils.test.js
ChristianMcc: Mm yes I definitely should have just added these, will add something :)

portal/server/utils/csv-utils.test.js
toddym42: Ah, of course. Probably wouldn't hurt to be a bit more explicit

portal/server/utils/csv-utils.js
ChristianMcc: Added some comments, haven't named as thought comments would do a better job not sure if that's a red flag.
26 = number of characters in alphabet
64 = A is the 65th Unicode character so making A = 1

portal/server/utils/csv-utils.js
```diff
@@ -3,6 +3,18 @@ const csv = require('csv-parser');
 const { Readable } = require('stream');
 const { CELL_ADDRESS_REGEX } = require('../constants/regex');
 
+/**
+ * @typedef {import('exceljs').Worksheet} Worksheet
+ * @typedef {Object} parsedXlsxDataResponse
```
toddym42, line 8: Nit: types would usually start with an upper-case. i.e. `ParsedXlsxDataResponse`

portal/server/utils/csv-utils.js
```diff
@@ -3,6 +3,18 @@ const csv = require('csv-parser');
...
+ * @property {Object} csvData - The index of the object
+ * @property {Object} csvDataWithCellAddresses - The error message of the object
```
toddym42, line 10: Looks like a copy-pasta on the descriptions here.

toddym42 (approved):

PR #2559, opened 2024-01-08
feat(FN-1051): add previous open reports to TFM 'Bank reports' page

toddym42:
## Ticket
[FN-1051](https://ukef-dtfs.atlassian.net/browse/FN-1051)

## Introduction :pencil2:
Add details of previous open reports to the TFM 'Bank reports' screen.

An "open report" is a report that has not been completed.

The existing Central API endpoint
```
utilisation-reports/reconciliation-summary/:submissionMonth
```
not returns details of previous open reports as well as the existing "Current reporting period" reports. The difference is that the existing "Current..." reports include all status, whereas the new previous ones only include open reports.


trade-finance-manager-ui/component-tests/assertions.js
```diff
@@ -1,5 +1,7 @@
 // eslint-disable-next-line no-unused-vars
 const assertions = (wrapper, html, params) => ({
+  html,
+  params,
```
toddym42, line 4: Allows us to access these in tests. I needed the params for one of my tests, but having `html` also makes debugging easier.

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-reports-reconciliation-summary.controller/helpers.ts
```diff
@@ -0,0 +1,169 @@
...
+const addNotReceivedReportsAndMapToSubmissionMonth = (
+  reports: UtilisationReport[],
+  currentSubmissionMonth: IsoMonthStamp,
+): UtilisationReportForSubmissionMonth[] => {
+  const currentReportPeriodStart = getReportPeriodStartForSubmissionMonth(currentSubmissionMonth);
+  const previousReportPeriodStart = getPreviousReportPeriodStart(currentReportPeriodStart);
+
+  const reportsOrderedByReportPeriodStartAscending = orderBy(reports, ['year', 'month'], ['asc', 'asc']);
+
+  return reportsOrderedByReportPeriodStartAscending.reduce((reportsWithSubmissionMonth, report, index) => {
```
JoshBinns2000, line 92: I don't think the `reduce` is appropriate here. You only perform the check for `reportsAreMissing` at the last index, so I think this would be easier to read if written like this:
```
const updatedReportsWithSubmissionMonth = reportsOrderedByReportPeriodStartAscending.map((report) => {
  const reportPeriodStart: ReportPeriodStart = { month: report.month, year: report.year };
  const submissionMonth = getSubmissionMonthForReportPeriodStart(reportPeriodStart);
  return { submissionMonth, report };
});

const mostRecentOpenReport = reportsOrderedByReportPeriodStartAscending.at(-1);
const mostRecentOpenReportPeriodStart: ReportPeriodStart = { month: mostRecentOpenReport.month, year: mostRecentOpenReport.year };

if (isEqualReportPeriodStart(mostRecentOpenReportPeriod, previousReportPeriod)) {
  return updatedReportsWithSubmissionMonth;
}

const mostRecentOpenReportSubmissionMonth = updatedReportsWithSubmissionMonth.at(-1).submissionMonth;
const missingReports = generateMissingReports({
  from: mostRecentOpenReportSubmissionMonth,
  to: currentSubmissionMonth,
});
return [...updatedReportsWithSubmissionMonth, ...missingReports];
```
Haven't tested this, but I think something like this is clearer than `reduce` in this case

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-reports-reconciliation-summary.controller/helpers.ts
```diff
@@ -0,0 +1,169 @@
...
+    return {
+      bank: {
+        id: bank.id,
+        name: bank.name,
+      },
+      status: UTILISATION_REPORT_RECONCILIATION_STATUS.REPORT_NOT_RECEIVED,
+    };
+  }
+
+  const reportData = await getAllUtilisationDataForReport(report);
```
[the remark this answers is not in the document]
JoshBinns2000: Okay, that makes sense

trade-finance-manager-ui/server/helpers/date.js
```diff
@@ -64,6 +64,13 @@ const getBusinessDayOfMonth = (dateInMonth, holidays, businessDay) => {
   return result;
 };
 
+/**
+ * Converts date with index-0 month value to numeric index-1 month
+ * @param {Date} date
+ * @returns {number}
```
JoshBinns2000, line 70: We could change this to now return the `OneIndexedMonth` type

trade-finance-manager-ui/server/helpers/express-session.ts
```diff
@@ -0,0 +1,31 @@
...
+ * (see use of `Partial` {@link https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/express-session/index.d.ts#L17 here})
+ *
+ * This helper function asserts that the `user` and `userToken` properties are
+ * present in the session, or throws an `Error` if not.
+ *
+ * Note: this doesn't assert on the validity of the user data values (which
+ * should have already been authenticated in middleware by this point), only
+ * that the values are present.
+ */
+export const asUserSession = (session: Session): UserSession => {
```
JoshBinns2000, line 19: Is there a reason to use `Object.assign` in the return here rather than the type predicate `Session is UserSession` and return `true` if all the checks pass? It just feels like, when more properties are added to `Session`, we need to both add the check and add the field to `Object.assign`, whereas having a `return true` at the bottom means you only need to add the check

trade-finance-manager-ui/server/test-mocks/mock-tfm-session-user.ts
JoshBinns2000, line 1: I think this file is missing a new line at the bottom

trade-finance-manager-ui/server/helpers/date.js
toddym42: The problem is that with type
```
export type OneIndexedMonth = 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11 | 12;
```
if I have this as the return type it gets upset as `number is not assignable to OneIndexedMonth`. Can probably be ignored here, but in typescript we'd have to do a load of casting to satisfy the type system.

I think I'm going to change the type to
```
export type OneIndexedMonth = number;
```
as the extra type wrangling isn't worth it here, but the alias still gives us some extra documentation without having to rely on a comment.

trade-finance-manager-ui/server/helpers/express-session.ts
toddym42: We could change as you suggest, and then change the function name to `isUserSession`, but then instead of just
```
// do things with user and userToken
```
we'd then have to change things to something like
```
if (isUserSession(req.session)) {
  // do things with user and userToken
} else {
  throw new Error('expected user session...');
}
```
which is a lot more verbose. The likelihood of needing to update this function to add more properties seems quite small, but if we do that seems like less work that having to add the extra code every where this is used.

Let me know though if you had something else in mind.

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-reports-reconciliation-summary.controller/helpers.ts
toddym42: This is much clearer, thanks

trade-finance-manager-ui/server/helpers/express-session.ts
JoshBinns2000: That makes sense to me

trade-finance-manager-ui/server/helpers/date.js
JoshBinns2000: That's a shame, but we do get enough of a benefit from the alias I think so I'm happy with that

JoshBinns2000 (approved):

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-reports-reconciliation-summary.controller/helpers.ts
```diff
@@ -0,0 +1,167 @@
...
+  report: UtilisationReport | null;
+};
+
+type SummaryItemForSubmissionMonth = {
+  submissionMonth: IsoMonthStamp;
+  item: UtilisationReportReconciliationSummaryItem;
+};
+
+const mapToSummaryItem = async (bank: Bank, report: UtilisationReport | null): Promise<UtilisationReportReconciliationSummaryItem> => {
+  // TODO FN-???? - Use actual REPORT_NOT_RECEIVED UtilisationReport when added by job. Throw errors when null.
```
ChristianMcc, line 34: this has a ticket that francesca has made now that we can probably reference just to be complete (as this will be merged into main)

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-reports-reconciliation-summary.controller/helpers.ts
```diff
@@ -0,0 +1,167 @@
...
+        id: bank.id,
+        name: bank.name,
+      },
+      status: UTILISATION_REPORT_RECONCILIATION_STATUS.REPORT_NOT_RECEIVED,
+    };
+  }
+
+  const reportData = await getAllUtilisationDataForReport(report);
+
+  // TODO FN-???? - status to be added to report data to allow us to calculate how
```
ChristianMcc, line 47: Can just put the epic 1398 for this

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-reports-reconciliation-summary.controller/helpers.ts
```diff
@@ -0,0 +1,167 @@
...
+
+  const reportsOrderedByReportPeriodStartAscending = orderBy(reports, ['year', 'month'], ['asc', 'asc']);
+
+  const updatedReportsWithSubmissionMonth = reportsOrderedByReportPeriodStartAscending.map((report) => {
+    const reportPeriodStart = getReportPeriodStartForUtilisationReport(report);
+    const submissionMonth = getSubmissionMonthForReportPeriodStart(reportPeriodStart);
+    return { submissionMonth, report };
+  });
+
+  // TODO FN-???? - when REPORT_NOT_RECEIVED reports are added by a job we can
```
ChristianMcc, line 99: ditto, let's just use FN-1549

ChristianMcc (approved):
LGTM, nice work. Only minor note to quickly add some tickets numbers against TODOs

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-reports-reconciliation-summary.controller/helpers.ts
toddym42: Updated to ref [FN-1949](https://ukef-dtfs.atlassian.net/browse/FN-1949)

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-reports-reconciliation-summary.controller/helpers.ts
toddym42: Have added ref for [FN-1949](https://ukef-dtfs.atlassian.net/browse/FN-1949) (which is for refactoring previous reports)

PR #2561, opened 2024-01-08
Feat/dtfs2 6931/add pagination to tfm and dtfs central backend

oscar-richardson-softwire:
## Introduction :pencil2:
The api doesnt paginate data at all. This should be introduced optionally without breaking existing behaviour, including for external uses of `dtfs-central-api`

* add `page` and `pagesize` parameters to the `GET` tfm deals endpoint. Return a `pagination` object containing metadata.
* pass this data through `trade-finance-manager-api`

## Misc ➕ 
Refactored `trade-finance-manager-api` mock api call implementations


dtfs-central-api/src/v1/controllers/tfm/deal/tfm-get-deals.controller.js
```diff
@@ -2,63 +2,32 @@ const moment = require('moment');
...
-        return -1;
-      }
+const findDeals = async (queryParameters) => {
+  const dealsCollection = await db.getCollection('tfm-deals');
 
-      if (yField > xField) {
-        return 1;
-      }
-    }
+  const {
```
ChristianMcc, line 14: do we want to default pageSize to something as well if it's not provided here? Probably to whatever we're expecting at the moment to be the pagesize

dtfs-central-api/src/v1/controllers/tfm/deal/tfm-get-deals.controller.js
```diff
@@ -2,63 +2,32 @@ const moment = require('moment');
...
+    searchString, sortBy, fieldQueries, pagesize, page = 0
+  } = queryParameters;
 
-    return 0;
-  });
+  let query = {};
```
ChristianMcc, line 18: can we call this databaseQuery to make it clear what this is representing please

dtfs-central-api/src/v1/controllers/tfm/deal/tfm-get-deals.controller.js
```diff
@@ -2,63 +2,32 @@ const moment = require('moment');
...
 
-const findDeals = async (searchString, sortBy, fieldQueries, callback) => {
-  const dealsCollection = await db.getCollection('tfm-deals');
+  const selectedField = {};
+  if (sortBy) {
+    sortBy.bssField = getBSSProperty(sortBy.field);
```
ChristianMcc, line 22: I'm a little confused but should this be selectedField.bssField?
If so could we call selectedField, `fieldToSortOn` or something

ChristianMcc: Can see from later it's not, I wonder if we take out the bssField and field properties into a bssField and a nonBssField variables to make it clear rather than changing the input sortBy parameter

trade-finance-manager-api/src/v1/api.js
```diff
@@ -628,16 +628,14 @@ const updateGefFacility = async (facilityId, facilityUpdate) => {
   }
 };
 
-const queryDeals = async ({ queryParams, start = 0, pagesize = 0 }) => {
+const queryDeals = async ({ queryParams }) => {
```
ChristianMcc, line 631: From the UI or here do we always set say pageSize already (on the queryParams) now? Wondering where we set these values now, related to my comment on the pageSize default in central API

ChristianMcc (commented):
Have only looked at a few of the major files here will give it some more review tomorrow :)

dtfs-central-api/src/v1/controllers/tfm/deal/tfm-get-deals.controller.js
```diff
@@ -2,63 +2,32 @@ const moment = require('moment');
 const escapeStringRegexp = require('escape-string-regexp');
 const db = require('../../../../drivers/db-client');
 const CONSTANTS = require('../../../../constants');
-const getObjectPropertyValueFromStringPath = require('../../../../utils/getObjectPropertyValueFromStringPath');
-const mapDataModel = require('../../../../mapping/mapDataModel');
-const setEmptyIfNull = require('../../../../utils/setEmptyIfNull');
```
oscar-richardson-softwire, line 7: Note to self: is it an issue that we no longer use this? Add a test

dtfs-central-api/src/v1/controllers/tfm/deal/tfm-get-deals.controller.js
```diff
@@ -92,102 +61,122 @@ const findDeals = async (searchString, sortBy, fieldQueries, callback) => {
...
+      currentPage: page,
+      totalPages: 0,
+    };
+    return { deals: [], pagination };
   }
+  const { count, deals } = doc[0];
 
-  if (callback) {
-    callback(deals);
-  }
```
oscar-richardson-softwire, line 161: Note to self: double check that we don't need the option to pass in a callback anymore

dtfs-central-api/api-tests/helpers/set-object-property-value-from-string-path.js
```diff
@@ -0,0 +1,14 @@
+const setObjectPropertyValueFromStringPath = (object, path, value) => {
```
ChristianMcc, line 1: Where is this used? This feels a little off to me that we need to do this so just want to review if we can avoid having this kind of function!

dtfs-central-api/src/v1/controllers/tfm/deal/tfm-get-deals.controller.js
oscar-richardson-softwire: The problem with defaulting `pagesize` to something here is that the requirement of this ticket was to introduce these back-end pagination changes optionally and without breaking existing behaviour. Currently, if `pagesize` is not provided then we do not paginate at all, but if we defaulted `pagesize` here then we would enforce pagination.

dtfs-central-api/src/v1/controllers/tfm/deal/tfm-get-deals.controller.js
oscar-richardson-softwire: Changed

trade-finance-manager-api/src/v1/api.js
oscar-richardson-softwire: At the moment, we don't set the `pagesize` anywhere (so the UI/actual running application behaviour is unchanged) as the idea of this ticket was to introduce the _ability_ to paginate to the back-end without breaking any existing behaviour. We'll work out the best place to set (and the best value for) `pagesize` when we come to add the front-end changes as part of https://ukef-dtfs.atlassian.net/browse/DTFS2-6932

dtfs-central-api/api-tests/helpers/set-object-property-value-from-string-path.js
oscar-richardson-softwire: This is used in `dtfs-central-api\api-tests\v1\tfm\deal\tfm-deals-get-sorting.api-test.js` to set the properties of test deals that will be inserted into the db using the same string path that will then be used in the query parameters to sort these deals. It's to keep the tests DRY and easily extendable.

trade-finance-manager-ui/public/css/styles.css
avaitonis, line 1: I guess you don't need to update css in UI microservice.

trade-finance-manager-ui/public/images/search.svg
avaitonis, line 1: This file was removed, no need to add it.

avaitonis (approved):
LGTM, except minor comments

dtfs-central-api/src/v1/controllers/tfm/deal/tfm-get-deals.controller.js
ChristianMcc: Ah I see, might be something we want to do on frontend then as I'd assume we always want it going forwards once we've implemented it! Sounds good

ChristianMcc (approved):
LGTM :) (remember to remove the test.yml branches update before merging!)

dtfs-central-api/src/v1/controllers/tfm/deal/tfm-get-deals.controller.js
oscar-richardson-softwire: Done

trade-finance-manager-ui/public/images/search.svg
oscar-richardson-softwire: Removed

PR #2563, opened 2024-01-09
feat(FN-1992): update currency validation error message

toddym42:
## Ticket
[FN-1992](https://ukef-dtfs.atlassian.net/browse/FN-1992)

## Introduction :pencil2:
Currently if a report field is a number, but has more than 2 decimal places, it still returns the generic
> <field_name> must be a number

error message. Instead the error message should be

> <field_name> must be a number with a maximum of two decimal places

Update error copy for any columns validating a currency amount
* Total fees accrued for the month
* Fees paid to UKEF for the period
* Facility utilisation

### DIFF
--- a/portal/server/controllers/utilisation-report-service/utilisation-report-upload/utilisation-report-cell-validators/generate-total-fees-accrued-error.js
+++ b/portal/server/controllers/utilisation-report-service/utilisation-report-upload/utilisation-report-cell-validators/generate-total-fees-accrued-error.js
@@ -2,32 +2,25 @@ const { CURRENCY_NUMBER_REGEX } = require('../../../../constants/regex');
 const { FILE_UPLOAD } = require('../../../../constants/file-upload');
 
 const generateTotalFeesAccruedError = (totalFeesAccruedObject, exporterName) => {
-  if (!totalFeesAccruedObject?.value) {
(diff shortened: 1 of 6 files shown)


portal/server/controllers/utilisation-report-service/utilisation-report-upload/utilisation-report-cell-validators/generate-total-fees-accrued-error.js
toddym42, line 1: I didn't need to refactor all this, I initially thought I was adding a new separate validator for number checking, and didn't want to repeat logic again.

By the time I realised that the AC's specify we want to return the same error for both cases I'd already made the changes, and thought no harm leaving in.

francescastocco (approved):
Looks good :) is there a reason you refactored `total-fees-accrued` but not the others?

toddym42:
> Looks good :) is there a reason you refactored `total-fees-accrued` but not the others?

Yes! I'd originally thought the change was going to be more complex, requiring an additional number validator. I'd refactored the code to make adding that more simple, then realised I just needed to update the copy for the existing currency validator. Rather than back it all out I thought it was worth keeping, but didn't feel it was worth the time to refactor everything else at this stage.

PR #2565, opened 2024-01-09
fix(FN-1363): add constant for landing pages

JoshBinns2000:
## Introduction :pencil2:
We want to add a constant to contain the new `\utilisation-report-upload` landing page url.

Introduces the `LANDING_PAGES` constant which contains a `DEFAULT` url and the new `UTILISATION_REPORT_UPLOAD` url. Also updates all existing `/utilisation-report-upload` routes to use the value in this new constant.

Updates the `.eslintrs.js` and `jest.config` files to allow for Typescript development in portal.


portal/server/controllers/utilisation-report-service/utilisation-report-upload/index.js
```diff
@@ -234,7 +235,7 @@ const postReportConfirmAndSend = async (req, res) => {
         ...req.session.utilisationReport,
         paymentOfficerEmail,
       };
-      return res.redirect('/utilisation-report-upload/confirmation');
+      return res.redirect(`${LANDING_PAGES.UTILISATION_REPORT_UPLOAD}/confirmation`);
```
toddym42, line 238: I'm not sure if it's correct to use `LANDING_PAGES` in this context, as the redirect isn't related to going to a landing page, it just happens that the `LANDING_PAGES` constant has what we want.

Maybe instead of `LANDING_PAGES` we make this more general, like
```
const UTILISATION_REPORT_UPLOAD = '/utilisation-report-upload';
const UTILISATION_REPORT_UPLOAD_CONFIRM_AND_SEND = `${UTILISATION_REPORT_UPLOAD}/confirm-and-send`;
const UTILISATION_REPORT_UPLOAD_CONFIRMATION = `${UTILISATION_REPORT_UPLOAD}/confirmation`;

export const ROUTES = {
  DEFAULT_LANDING: '/dashboard/deals/0', <-- maybe construct this from other routes too?
  UTILISATION_REPORT_UPLOAD,
  UTILISATION_REPORT_UPLOAD_CONFIRM_AND_SEND,
  UTILISATION_REPORT_UPLOAD_CONFIRMATION
} as const;
```
What do you think?

portal/server/routes/utilisation-report-service/utilisation-report-upload/index.js
```diff
@@ -9,18 +9,24 @@ const {
...
 const router = express.Router();
 
-const upload = multer({ limits: { fileSize: +MAX_UTILISATION_REPORT_FILE_SIZE_BYTES }, fileFilter: utilisationReportMulterFilter }).single('utilisation-report-file-upload');
+const upload = multer({ limits: { fileSize: +MAX_UTILISATION_REPORT_FILE_SIZE_BYTES }, fileFilter: utilisationReportMulterFilter }).single(
+  'utilisation-report-file-upload',
+);
 
-router.get('/utilisation-report-upload', [validateToken, validateRole({ role: [ROLES.PAYMENT_REPORT_OFFICER] })], (req, res) => getUtilisationReportUpload(req, res));
+router.get(
+  LANDING_PAGES.UTILISATION_REPORT_UPLOAD,
```
toddym42, line 23: Same comment here. It doesn't feel right to be referring to "LANDING_PAGES" in this context.

portal/server/controllers/utilisation-report-service/utilisation-report-upload/index.js
JoshBinns2000: My problem with that is that introducing this constant implies that every route should have a constant, which I don't think it should. To be totally honest I don't agree with the presence of this constant at all as no other route uses a constant, and I only looked at introducing this in the other places to keep the url linked. I would rather remove all the references to `LANDING_PAGES` in locations like this and only keep it in the places where you are routed to a "landing page" (login and the home route)

JoshBinns2000: Actually, I do think that `LANDING_PAGES` is a constant that should only be referenced in those locations, and these other files should use raw strings

toddym42: Yep, that works.

portal/jest.common.config.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
+  preset: 'ts-jest',
+  transform: {
+    '^.+\\.js$': 'babel-jest',
+    '^.+\\.ts$': 'ts-jest',
+  },
+};
```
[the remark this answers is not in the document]
JoshBinns2000: Don't see why not!

toddym42 (approved):
LGTM

PR #2566, opened 2024-01-10
fix(FN-1051): correct open reports DB query

toddym42:
## Ticket
[FN-1051](https://ukef-dtfs.atlassian.net/browse/FN-1051)

## Introduction :pencil2:
Correct the DB query to get open reports (details in file)


dtfs-central-api/src/services/repositories/utilisation-reports-repo/utilisation-reports-repo.ts
```diff
@@ -50,36 +50,22 @@ export const getOpenReportsBeforeReportPeriodForBankId = async (reportPeriodStar
   const collection = await db.getCollection(DB_COLLECTIONS.UTILISATION_REPORTS);
 
   return await collection
```
toddym42, line 52: Summary of changes:
* correct match on `bankId` to `'bank.id'`
* replace `aggregate` with `find` (in a previous version I had a multi-step query, but didn't need it in the end)

JoshBinns2000 (approved):
Nice - especially happy that `aggregate` has been able to be replaced with a more simple `find` :)

PR #2567, opened 2024-01-10
fix(FN-1536): move payload parameters from query to body

JoshBinns2000:
## Introduction :pencil2:
Some of the parameters required to generate the `updateReportStatus` payload were stored in the request `url` as query parameters instead of being hidden in the request body. This meant that manually reloading the page after attempting to mark a report as done or not done would attempt to resubmit the form.

Moves the query parameters to the body and updates the controller to extract these values.

Updates a function which didn't need access to the `userToken`.

### DIFF
--- a/trade-finance-manager-ui/server/controllers/utilisation-reports/update-utilisation-report-status/index.ts
+++ b/trade-finance-manager-ui/server/controllers/utilisation-reports/update-utilisation-report-status/index.ts
@@ -29,7 +29,7 @@ type UpdateUtilisationReportStatusRequestBody = {
   [key: string]: 'on'; // all checkboxes in payload have value 'on'
 };
 
-const getReportIdentifiersFromBody = (body: null | UpdateUtilisationReportStatusRequestBody): ReportIdentifier[] => {
+const getReportIdentifiersFromBody = (body: undefined | UpdateUtilisationReportStatusRequestBody): ReportIdentifier[] => {
   if (!body || typeof body !== 'object') {
     return [];
   }
--- a/trade-finance-manager-ui/server/controllers/utilisation-reports/update-utilisation-report-status/index.ts
+++ b/trade-finance-manager-ui/server/controllers/utilisation-reports/update-utilisation-report-status/index.ts
@@ -5,6 +5,8 @@ import { ReportIdentifier, ReportWithStatus } from '../../../types/utilisation-r
 import { UTILISATION_REPORT_RECONCILIATION_STATUS } from '../../../constants';
 import { getUtilisationReports } from '..';
 import { asUserSession } from '../../../helpers/express-session';
+import { assertValidIsoMonth } from '../../../helpers/date';
+import { getReportPeriodStart } from '../../../services/utilisation-report-service';
 
 const CHECKBOX_PREFIX_REGEX = 'set-status--';
 const CHECKBOX_ID_TYPE_REGEX = 'reportId|bankId';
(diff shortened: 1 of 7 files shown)


trade-finance-manager-ui/server/helpers/date.js
```diff
@@ -85,7 +85,7 @@ const getIsoMonth = (dateInMonth) => {
   return format(dateInMonth, 'yyyy-MM');
 };
 
-const ISO_MONTH_REGEX = /^\d{4}-\d{2}$/;
+const ISO_MONTH_REGEX = /^(?<year>\d{4})-(?<month>\d{2})$/;
```
JoshBinns2000, line 88: I know we don't need this change anymore, but I think it is worth including it as it does make the structure of the regex clearer to read. Thoughts?

trade-finance-manager-ui/server/controllers/utilisation-reports/update-utilisation-report-status/index.ts
```diff
@@ -67,12 +69,27 @@ const getReportWithStatus = (reportIdentifier: ReportIdentifier, formButton: str
...
+  _csrf: string;
+  'form-button': string;
+  'submission-month': string;
+} & {
+  [key: string]: 'on'; // all checkboxes in payload have value 'on'
+};
+
+interface UpdateUtilisationReportStatusRequest extends Request {
+  body: UpdateUtilisationReportStatusRequestBody;
+}
```
toddym42, line 82: This feels like a bit of a hacky way of setting the type of `body` rather than using the generic parameters of the `Request` type. It could be that the `ReqBody` type parameter is used to set types elsewhere within `Request` (I'm not sure if it is, but the point stands).

trade-finance-manager-ui/server/controllers/utilisation-reports/update-utilisation-report-status/index.ts
```diff
@@ -67,12 +69,27 @@ const getReportWithStatus = (reportIdentifier: ReportIdentifier, formButton: str
...
   const { user, userToken } = asUserSession(req.session);
-  const { 'form-button': formButton, reportPeriodStartMonth, reportPeriodYear } = req.query;
 
   try {
-    const reportIdentifiers = getReportIdentifiersFromBody(req.body, Number(reportPeriodStartMonth), Number(reportPeriodYear));
+    const { 'form-button': formButton, 'submission-month': submissionMonth } = req.body;
+    assertValidIsoMonth(submissionMonth);
+    const reportPeriodStart = getReportPeriodStart(submissionMonth);
+
+    const reportIdentifiers = getReportIdentifiersFromBody(req.body, reportPeriodStart.month, reportPeriodStart.year);
```
toddym42, line 92: Now that `submissionMonth` is part of `req.body` can this just be
```
const reportIdentifiers = getReportIdentifiersFromBody(req.body);
```
then let that function work everything out?

trade-finance-manager-ui/server/controllers/utilisation-reports/update-utilisation-report-status/index.ts
JoshBinns2000: I actually think this is the correct way of doing it rather than using the generic. `interface` allows us to extend the type whilst preserving the the original details (primarily required for the `query` type definition), whereas `Request<{}, {}, [Body]>` causes two errors:
- firstly, you can't use `{}` because of some eslint rule
- secondly, changing to `object` works for the type of `req` in this function the same was as `interface`, but when you try to pass that request down to `getUtilisationReports`, you get a new error saying that `"Type 'object' is not assignable to type 'ParamsDictionary'."`

Extending the `interface` instead resolves these issues as only the type we actually modify in `Request` is updated and the rest are kept at their original states

toddym42: Yeah, I get that, but the fact that we're not using the generics provided by the `Request` type means that if `ReqBody` (the third param) is used anywhere other than on the `body` property, then this way means it will be ignored. I see other people are using it, but that doesn't convince me this is good practice.

It is a pain though to have to provide the defaults for the other properties, so I can see why it would be done this way. It's not enough of a concern for me to hold this up and look into further, so happy to go with what you've done.

trade-finance-manager-ui/server/helpers/date.js
toddym42: Yep, agreed 👍

toddym42 (approved):

PR #2579, opened 2024-01-11
fix: update .d.ts files to .ts where applicable

toddym42:
We should only really use `.d.ts` files to declare types for imported libraries that do not provide their own types, or if we want to change a type provided by an imported library (such as where we updated the type of `SessionData` from the `express-validator` library to include our custom `user` and `userToken` properties). When providing types for our own code we should just use `.ts`.

https://www.typescriptlang.org/docs/handbook/2/type-declarations.html


trade-finance-manager-api/src/types/tfm-session-user.ts
```diff
@@ -0,0 +1,5 @@
+import { TfmUser } from './db-models/tfm-users';
+
+export type TfmSessionUser = Omit<TfmUser, 'salt' | 'hash' | 'loginFailureCount' | 'sessionIdentifier' | '_id'> & {
+  _id: string;
```
JoshBinns2000, line 4: Whilst we're here, could we change this to use `Pick` instead of `Omit`

JoshBinns2000 (commented):
Happy to approve after the one comment is resolved :)

JoshBinns2000 (approved):
lgtm

PR #2582, opened 2024-01-12
feat(FN-1946): refactor to store report period on utilisation reports

francescastocco:
## Introduction :pencil2:
DRAFT
On `utilisationReports` collection instead of storing the `month `and `year` of a report we should have a new object called `reportPeriod `which looks like this:
```
reportPeriod: {
  start: {
    month: number,
    year: number,
  }
  end: {
    month: number,
    year: number,
  }
}
```


dtfs-central-api/src/services/repositories/utilisation-reports-repo/utilisation-reports-repo.test.ts
```diff
@@ -78,10 +94,58 @@ describe('utilisation-reports-repo', () => {
     it('sorts the data by year then month', async () => {
       // Arrange
       const bankId = MOCK_UTILISATION_REPORT.bank.id;
-      const report1 = { ...MOCK_UTILISATION_REPORT, month: 2, year: 2022 };
-      const report2 = { ...MOCK_UTILISATION_REPORT, month: 3, year: 2021 };
-      const report3 = { ...MOCK_UTILISATION_REPORT, month: 1, year: 2022 };
-      const report4 = { ...MOCK_UTILISATION_REPORT, month: 2, year: 2021 };
```
toddym42, line 84: Would be nice to keep these on one line as it makes the test a lot easier to reason about. Could we add a function just below
```
describe('getUtilisationReportDetailsByBankId', () => {
```
as follows:
```
const getMockReport = ({ bankId, year, month }: { bankId: string; year: number; month: number }): UtilisationReport => ({
  ...MOCK_UTILISATION_REPORT,
  bank: {
    ...MOCK_UTILISATION_REPORT.bank,
    id: bankId,
  },
  reportPeriod: {
    start: { month, year },
    end: { month, year },
  },
});
```
then I think we can update this to:
```
const report1 = getMockReport({ bankId, month: 2, year: 2022 });
const report2 = getMockReport({ bankId, month: 3, year: 2021 });
const report3 = getMockReport({ bankId, month: 1, year: 2022 });
const report4 = getMockReport({ bankId, month: 2, year: 2021 });
```

dtfs-central-api/src/types/utilisation-reports.ts
```diff
@@ -24,11 +24,18 @@ export type UtilisationReportReconciliationSummary = {
   items: UtilisationReportReconciliationSummaryItem[];
 };
 
-export type ReportPeriodStart = {
+export type ReportPeriodMonthAndYear = {
   month: OneIndexedMonth;
   year: number;
 };
```
toddym42, line 30: I think we can now make this a more generic `MonthAndYear` type and move to `date.ts` adjacent to this file.

dtfs-central-api/src/types/utilisation-reports.ts
```diff
@@ -24,11 +24,18 @@ export type UtilisationReportReconciliationSummary = {
...
 
+export type ReportPeriod = {
+  start: ReportPeriodMonthAndYear;
+  end: ReportPeriodMonthAndYear;
+};
+
+export type ReportPeriodStart = ReportPeriodMonthAndYear;
```
toddym42, line 37: I think we can now remove this and just refer to `MonthAndYear` everywhere. I think what it represents should be clear from context.

dtfs-central-api/src/utils/report-period.ts
```diff
@@ -4,10 +4,7 @@ import { IsoMonthStamp } from '../types/date';
 import { ReportPeriodStart } from '../types/utilisation-reports';
 import { UtilisationReport } from '../types/db-models/utilisation-reports';
 
-export const getReportPeriodStartForUtilisationReport = ({ month, year }: UtilisationReport): ReportPeriodStart => ({
-  month,
-  year,
-});
+export const getReportPeriodStartForUtilisationReport = ({ reportPeriod }: UtilisationReport): ReportPeriodStart => (reportPeriod.start);
```
toddym42, line 7: I think we can probably remove this function now as `reportPeriod.start` is now more concise that calling the function : )

dtfs-central-api/src/v1/swagger-definitions/utilisation-reports/utilisation-report.js
```diff
@@ -8,13 +8,31 @@
...
- *       month:
- *         type: integer
- *         example: 1
- *         description: Month of utilisation report indexed at 1
- *       year:
- *         type: integer
- *         example: 2021
+ *       reportPeriod:
+ *         type: object
+ *         properties:
```
toddym42, line 13: Some extra spaces at the end of the line here to be removed.

dtfs-central-api/src/v1/validation/utilisation-report-service/utilisation-report-validator.js
```diff
@@ -17,33 +17,68 @@ const {
...
+ * @param {unknown} reportPeriod - details of the report period.
+ * @returns {string[]} - an array of errors or an empty array if valid.
+ */
+const validateReportPeriod = (reportPeriod) => {
+  if (!reportPeriod) {
+    return ['Report period is required'];
+  }
+
+  const reportPeriodErrors = [];
+  const { month: startMonth, year: startYear,  } = reportPeriod.start;
```
toddym42, line 60: An extra `, ` to remove after `startYear` here (prettier should take care of this)

dtfs-central-api/src/v1/validation/utilisation-report-service/utilisation-report-validator.js
```diff
@@ -17,33 +17,68 @@ const {
...
+  const { month: endMonth, year: endYear } = reportPeriod.end;
+  const validateStartMonth = validateMonth(startMonth, 'startMonth');
+  if (validateStartMonth !== null) {
+    reportPeriodErrors.push(validateStartMonth)
+  }
```
toddym42, line 65: Since the response of `validateMonth` is an error, I think the following might make this a bit clearer:
```
const startMonthError = validateMonth(startMonth, 'startMonth');
if (startMonthError) {
  reportPeriodErrors.push(startMonthError)
}
```

portal-api/src/scheduler/jobs/helpers/utilisation-report-helpers.js
```diff
@@ -44,13 +44,22 @@ const getReportOverdueChaserDate = async () => {
 };
 
 /**
- * Returns the month (1-indexed) and year of the current report period (i.e. the previous month)
- * @returns {{ month: number, year: number }}
+ * Returns the start and end months (1-indexed) and years of the current report period (i.e. the previous month)
+ * @returns {{ start: {month: number, year: number}, end: {month: number, year: number} }}
```
toddym42, line 48: Could use the following now to keep this consistent with dtfs-central-api
```
* @returns {{ start: MonthAndYear, end: MonthAndYear }}
```

toddym42: Or even
```
* @returns {ReportPeriod}
```
if we define that type too.

portal-api/src/v1/api.js
```diff
@@ -268,29 +267,38 @@ const saveUtilisationReport = async (reportData, month, year, user, fileInfo) =>
...
+
+    if (reportPeriod && !isValidMonth(parseInt(reportPeriod.end.month, 10))) {
+      console.error('Get utilisation reports failed with the following end month %s', reportPeriod.endMonth);
+      throw new Error('Invalid end month provided: %s', reportPeriod.endMonth);
+    }
+
+    if (reportPeriod && !isValidYear(parseInt(reportPeriod.end.year, 10))) {
+      console.error('Get utilisation reports failed with the following end year %s', reportPeriod.endYear);
+      throw new Error('Invalid end year provided: %s', reportPeriod.endYear);
     }
```
toddym42, line 295: Might be nice to extract this validation so we can just do
```
if (!isValidReportPeriod(reportPeriod)) {
  console.error('Get utilisation reports failed with the following report period', reportPeriod);
  throw new Error('Invalid report period provided: %s', reportPeriod);
}
```

portal-api/src/v1/controllers/utilisation-report-service/due-report-dates.controller.js
```diff
@@ -29,19 +29,19 @@ const getNextDueReportDate = (latestReport, currentDueReportDate) => {
...
 
-  const { month: oneIndexedMonth, year } = latestReport;
-  const latestReportDate = new Date(year, oneIndexedMonth - 1);
+  const { month: oneIndexedStartMonth, year: startYear } = latestReport.reportPeriod.start;
+  const latestReportDate = new Date(startYear, oneIndexedStartMonth - 1);
   return addMonths(latestReportDate, 1);
 };
 
 /**
  * @typedef {Object} DueReportDate
```
toddym42, line 38: If we add the generic `MonthAndYear` type to this project we just use that instead and remove this.

portal/server/controllers/utilisation-report-service/utilisation-report-upload/index.js
```diff
@@ -10,22 +10,30 @@ const { getReportAndUserDetails } = require('./utilisation-report-details');
...
  */
 const getDueReportDates = async (userToken, bankId) => {
   const dueReports = await api.getDueReportDatesByBank(userToken, bankId);
   return dueReports.map((dueReport) => {
-    const { month, year } = dueReport;
-    const reportPeriod = format(new Date(year, month - 1), 'MMMM yyyy');
-    return { ...dueReport, reportPeriod };
+    const { startMonth, startYear } = dueReport;
+    const reportPeriodString = format(new Date(startYear, startMonth - 1), 'MMMM yyyy');
+    return { ...dueReport, reportPeriodString };
```
toddym42, line 20: For other formatted dates we've used something like `formattedReportPeriod` elsewhere. I think this also helps differentiate it from something that is just a stringified version (e.g. `JSON.stringify(reportPeriod)`) which I think we have done elsewhere and also named `reportPeriodString`.

trade-finance-manager-api/src/v1/__mocks__/mock-utilisation-report.ts
```diff
@@ -8,14 +8,20 @@ const MOCK_AZURE_FILE_INFO = {
   mimetype: 'text/csv',
 };
 
+const MOCK_REPORT_PERIOD = {
+  startMonth: 11,
+  startYear: 2023,
+  endMonth: 11,
+  endYear: 2023,
+}
```
toddym42, line 16: Should this be
```
const MOCK_REPORT_PERIOD = {
  start: { month: 11, year: 2023 },
  end: { month: 11, year: 2023 },
}
```
?

dtfs-central-api/src/services/repositories/utilisation-reports-repo/utilisation-reports-repo.test.ts
```diff
@@ -25,8 +25,16 @@ describe('utilisation-reports-repo', () => {
       });
       jest.spyOn(db, 'getCollection').mockImplementation(getCollectionMock);
 
-      const mockMonth = 1;
-      const mockYear = 2021;
+      const mockReportPeriod = {
```
JoshBinns2000, line 28: May be worth specifying the type here

dtfs-central-api/src/types/db-models/utilisation-data.ts
```diff
@@ -12,15 +13,9 @@ export type UtilisationData = WithId<{
    */
   bankId: string;
   /**
-   * 1-indexed month of the start of the report period
-   * example: 6 (for June)
+   * Start and end data of the report period
```
JoshBinns2000, line 16: I think it may be worth giving an example here (also should this maybe say `date` instead of `data`?)

dtfs-central-api/src/types/db-models/utilisation-reports.ts
```diff
@@ -17,15 +17,9 @@ export type UtilisationReport = Prettify<
      */
     bank: SessionBank;
     /**
-     * 1-indexed month of the start of the report period
-     * example: 6 (for June)
+     * Start and end data of the report period
```
JoshBinns2000, line 20: Same comment as above

trade-finance-manager-api/src/v1/__mocks__/mock-utilisation-report.ts
JoshBinns2000: I think you should add the type explicitly to `MOCK_REPORT_PERIOD` so that this error in structure gets flagged up for you :)

portal-api/src/scheduler/jobs/helpers/utilisation-report-helpers.js
JoshBinns2000: `portal-api` didn't have much in the way of types, so I've added a new directory for them :)

dtfs-central-api/src/v1/validation/utilisation-report-service/utilisation-report-validator.js
JoshBinns2000: Modified this and all the similar values below

dtfs-central-api/src/services/repositories/utilisation-reports-repo/utilisation-reports-repo.test.ts
JoshBinns2000: It doesn't appear that `bankId` is a field which needs to be passed in as this test uses the same value as `MOCK_UTILISATION_REPORT` anyway

dtfs-central-api/src/v1/swagger-definitions/utilisation-reports/utilisation-report.js
JoshBinns2000: I can't see any empty spaces 😕

JoshBinns2000: Oh I'm guessing they were removed 😆

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-reports.controller.js
```diff
@@ -6,21 +6,19 @@ const {
 const getUtilisationReports = async (req, res) => {
   try {
     const { bankId } = req.params;
+    const { reportPeriod } = req.body;
```
toddym42, line 9: As this is a 'GET' request there shouldn't really be a request body. Also, is it correct that we want to search by the whole report period? If so then the query to the DB should also take into account the end dates.

portal-api/src/v1/validation/validateIds.js
```diff
@@ -60,6 +60,22 @@ const isValidMonth = (month) => Number.isInteger(month) && month >= 1 && month <
  */
 const isValidYear = (year) => Number.isInteger(year) && year >= 2000 && year <= 2100;
 
+/**
+ * isValidReportPeriod
+ * checks if the supplied report period is a valid report period object
+ * @param {import('../../types/utilisation-reports').ReportPeriod} reportPeriod
```
toddym42, line 66: Should type this as `unknown` as a prompt to check that the values exist as is expected of the `ReportPeriod` type.

trade-finance-manager-api/src/types/utilisation-reports.ts
```diff
@@ -4,22 +4,31 @@ import { IsoDateTimeStamp } from './date';
...
   filename: string;
   fullPath: string;
   url: string;
   mimetype: string;
 };
 
+type MonthAndYear = {
+  month: number;
+  year: number;
+};
```
toddym42, line 18: Can this go in a `date.ts` file instead.

dtfs-central-api/src/services/repositories/utilisation-reports-repo/utilisation-reports-repo.ts
```diff
@@ -32,21 +32,21 @@ export const saveUtilisationReportDetails = async (month: number, year: number,
 
 export const getUtilisationReportDetailsByBankIdMonthAndYear = async (bankId: string, month: number, year: number): Promise<UtilisationReport | null> => {
```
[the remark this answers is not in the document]
JoshBinns2000: As discussed, we should still search by a specific `month` and `year` and (once quarterly is introduced) handle the logic to see which report period we are in for the specific bank

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-reports.controller.js
JoshBinns2000: These values now come from `req.query` (just for clarity sake I have upgraded this file to typescript)

toddym42 (dismissed):

portal-api/src/v1/api.js
```diff
@@ -267,14 +267,33 @@ const saveUtilisationReport = async (reportData, reportPeriod, user, fileInfo) =
...
+        method: 'get',
+        url: `${DTFS_CENTRAL_API_URL}/v1/bank/${bankId}/utilisation-reports`,
+        headers: headers.central,
+      });
+      return response.data;
+    }
+
+    if (!isValidReportPeriod(reportPeriod)) {
       console.error('Get utilisation reports failed with the following report period:', reportPeriod);
       throw new Error('Invalid report period provided: %s', reportPeriod);
```
toddym42, line 298: Passing `reportPeriod` as a second argument to populate the `%s` works for `console.error` but not for the `Error` class. Same issue above for
```
throw new Error('Invalid bank ID provided: %s', bankId);
```

portal-api/src/v1/api.js
```diff
@@ -267,14 +267,33 @@ const saveUtilisationReport = async (reportData, reportPeriod, user, fileInfo) =
   }
 };
 
+/**
+ * Gets utilisation reports for a specific bank. If a report
+ * period is not provided, all reports for that bank are
+ * returned. If a report period is provided, the report
+ * submitted for that report period is returned. Returned
+ * reports are ordered by year descending and month
+ * ascending.
```
toddym42, line 276: Looks like both year and month are ascending?
```
return sortBy(filteredUtilisationReports, ['reportPeriod.start.year', 'reportPeriod.start.month']);
```

portal-api/src/v1/api.js
```diff
@@ -268,33 +267,43 @@ const saveUtilisationReport = async (reportData, month, year, user, fileInfo) =>
...
+    const { month, year } = reportPeriod.start;
 
     const response = await axios({
       method: 'get',
-      url: `${DTFS_CENTRAL_API_URL}/v1/bank/${bankId}/utilisation-reports`,
-      params: {
-        month,
-        year,
-      },
+      url: `${DTFS_CENTRAL_API_URL}/v1/bank/${bankId}/utilisation-reports?month=${month}&year=${year}`,
```
toddym42, line 304: When using axios we shouldn't ever need to manually construct query params, but instead pass them as the `params` option. As well as preventing the need to manually construct it, this will also prevent the need for us to URL encode values (if required).

Taking all these changes into account this function can be simplified to
```
const getUtilisationReports = async (bankId, reportPeriod) => {
  try {
    if (!isValidBankId(bankId)) {
      console.error('Get utilisation reports failed with the following bank ID %s', bankId);
      throw new Error(`Invalid bank ID provided: ${bankId}`);
    }

    if (reportPeriod && !isValidReportPeriod(reportPeriod)) {
      console.error('Get utilisation reports failed with the following report period:', reportPeriod);
      throw new Error(`Invalid report period provided: ${reportPeriod}`);
    }

    const response = await axios.get(`${DTFS_CENTRAL_API_URL}/v1/bank/${bankId}/utilisation-reports`, {
      params: reportPeriod ? reportPeriod.start : undefined,
      headers: headers.central,
    });
    return response.data;
  } catch (error) {
    console.error('Unable to get previous utilisation reports %s', error);
    throw error;
  }
};
```

portal-api/src/v1/api.js
JoshBinns2000: I was getting confused with how something else worked (or how it used to work)

portal-api/src/v1/api.js
JoshBinns2000: My reasoning was based on the pattern that `query` values are supposed to act as filters for the data, which is what they would be doing here. I also think you can bypass the URL encoding (to some extent) by storing then in a `query` object and then passing that to the `axios.get` call:
```
axios.get([url], {
  headers: headers.central,
  query: reportPeriod ? reportPeriod.start : undefined,
}
```

toddym42: Perfect, that's what I meant 😅

toddym42 (dismissed):

dtfs-central-api/src/services/repositories/utilisation-reports-repo/utilisation-reports-repo.ts
```diff
@@ -32,13 +31,13 @@
...
   const utilisationReportDetailsCollection = await db.getCollection(DB_COLLECTIONS.UTILISATION_REPORTS);
-  return await utilisationReportDetailsCollection.findOne({ 'bank.id': bankId, month, year });
+  return await utilisationReportDetailsCollection.findOne({ 'bank.id': bankId, 'reportPeriod.start.month': month, 'reportPeriod.start.year': year });
```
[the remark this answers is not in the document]
abhi-markan: Please convert all the usage of `findOne` with with `$eq` operator. (an example below)

```javascript
const maker = await userCollection.findOne({ roles: { $eq: ['maker'] } });
```

dtfs-central-api/src/types/custom-express-request.ts
```diff
@@ -0,0 +1,31 @@
...
+/**
+ * Obtained from the express 'core.Query' type
+ */
+interface RequestQuery {
+  [key: string]: string | string[] | RequestQuery | RequestQuery[] | undefined;
+}
+
+type CustomExpressRequestOptions = {
+  params?: RequestParams;
+  /* eslint-disable @typescript-eslint/no-explicit-any */
```
abhi-markan, line 17: Not a big fan of eslint inline rule disable, anything can be done for an ideal fix?

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-reports.controller.ts
```diff
@@ -0,0 +1,32 @@
...
+  params: {
+    bankId: string;
+  };
+  query: {
+    month: string | undefined;
+    year: string | undefined;
+  };
+}>;
+
+export const getUtilisationReports = async (req: GetUtilisationReportsRequest, res: Response) => {
```
abhi-markan, line 15: Missing docstring, please ensure all new functions have docstrings.

dtfs-central-api/src/v1/validation/utilisation-report-service/utilisation-report-validator.js
```diff
@@ -17,33 +17,70 @@ const {
...
   return null;
 };
 
+/**
+ * Validates the report period for the utilisation report
+ * @param {unknown} reportPeriod - details of the report period.
+ * @returns {string[]} - an array of errors or an empty array if valid.
+ */
+const validateReportPeriod = (reportPeriod) => {
+  if (!reportPeriod || !reportPeriod.start || !reportPeriod.end) {
```
abhi-markan, line 55: Can this be simplified to `!reportPeriod?.start || !reportPeriod?.end`

portal-api/src/v1/api.js
```diff
@@ -268,36 +267,35 @@ const saveUtilisationReport = async (reportData, month, year, user, fileInfo) =>
...
+ * @param {string} bankId
+ * @param {import('../types/utilisation-reports').ReportPeriod} [reportPeriod]
+ */
+const getUtilisationReports = async (bankId, reportPeriod) => {
   try {
     if (!isValidBankId(bankId)) {
-      console.error('Get utilisation reports failed with the following bank ID %s', bankId);
-      throw new Error('Invalid bank ID provided: %s', bankId);
+      console.error('Get utilisation reports failed with the following bank ID: %s', bankId);
+      throw new Error(`Invalid bank ID provided: ${bankId}`);
```
abhi-markan, line 283: String substitution please.

portal-api/src/v1/api.js
```diff
@@ -268,36 +267,35 @@ const saveUtilisationReport = async (reportData, month, year, user, fileInfo) =>
...
     }
 
-    if (month && !isValidMonth(parseInt(month, 10))) {
-      console.error('Get utilisation reports failed with the following month %s', month);
-      throw new Error('Invalid month provided: %s', month);
+    if (reportPeriod && !isValidReportPeriod(reportPeriod)) {
+      console.error('Get utilisation reports failed with the following report period: %s', reportPeriod);
+      throw new Error(`Invalid report period provided: ${reportPeriod}`);
```
abhi-markan, line 288: String substitution please.

portal-api/src/v1/validation/validateIds.js
```diff
@@ -60,6 +60,22 @@ const isValidMonth = (month) => Number.isInteger(month) && month >= 1 && month <
  */
 const isValidYear = (year) => Number.isInteger(year) && year >= 2000 && year <= 2100;
 
+/**
+ * isValidReportPeriod
+ * checks if the supplied report period is a valid report period object
+ * @param {unknown} reportPeriod
+ * @returns {boolean}
+ */
+const isValidReportPeriod = (reportPeriod) => !!(
```
abhi-markan, line 69: !! (Boolean conversation) is deprecated, compound `&&` without brackets will suffice.

portal/server/api.js
```diff
@@ -839,18 +839,17 @@ const getUkefDecisionReport = async (token, payload) => {
...
     formData.append('user', JSON.stringify(uploadingUser));
-    formData.append('month', month);
-    formData.append('year', year);
-    formData.append('reportPeriod', reportPeriod);
+    formData.append('reportPeriod', JSON.stringify(reportPeriod));
+    formData.append('formattedReportPeriod', formattedReportPeriod);
 
     const buffer = Buffer.from(csvFileBuffer);
-    const filename = `${year}_${month}_${FILE_UPLOAD.FILENAME_SUBMITTED_INDICATOR}_${uploadingUser.bank.name}_utilisation_report.csv`;
+    const filename = `${reportPeriod.start.year}_${reportPeriod.start.month}_${FILE_UPLOAD.FILENAME_SUBMITTED_INDICATOR}_${uploadingUser.bank.name}_utilisation_report.csv`;
```
abhi-markan, line 852: Possible constant candidate `_utilisation_report.csv`?

abhi-markan (changes requested):

portal-api/src/v1/validation/validateIds.js
JoshBinns2000: If we want to ensure a `boolean` return type then we either need the `!!` type cast or we will need to change to explicit `=== undefined` checks to make sure the return type is `boolean`. If you just use compound `&&` in this case, the JSDocs return type would need to be updated to say that the function can return `null | undefined | boolean`. Which approach would you prefer?

dtfs-central-api/src/types/custom-express-request.ts
JoshBinns2000: The default type for `import('express').Request.reqBody` and `import('express').Request.resBody` is `any` in the official `@types/express`. This eslint rule was disabled (rather than coming up with a different type) to avoid any unintentional type conflicts between a `CustomExpressRequest` and the default `import('express').Request` type for properties you don't wish to overwrite

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-reports.controller.ts
francescastocco: Do we need to add documentation for typescript functions like these? In this case, the naming and typing of parameters/return values means that adding code does not provide a benefit and adds a small maintenance cost. As we're aiming to write self-documenting code I think additional code would go against DRY principles

portal-api/src/v1/api.js
francescastocco: What's the benefit of using string substitution here? I'm keen to read about this if you have any more info

portal-api/src/v1/validation/validateIds.js
abhi-markan: Can we use `Boolean()`?

portal-api/src/v1/api.js
abhi-markan: https://console.spec.whatwg.org/#formatter

abhi-markan (changes requested):

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-reports.controller.ts
abhi-markan: Docstrings provide documentation for the code, making it easier for others to understand the purpose, usage, and expected behaviour of functions. Additionally, many development tools and IDEs use these comments to provide helpful information and suggestions while coding.

I use this for reference https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html.

abhi-markan (changes requested):

abhi-markan (approved):

PR #2584, opened 2024-01-12
feat(FN-1051): add separator between bank reports tables

toddym42:
## Introduction :pencil2:
After adding previous open reports to the "Bank reports" screen there are now multiple tables on the page, each with their own set of buttons to mark reports as completed or not completed. It may now be confusing for users to understand which set of buttons relate to which table.

current

The solution is to add extra spacing and a separator line between each table to make it a bit clearer:

image


trade-finance-manager-ui/styles/layout.scss
```diff
@@ -56,4 +56,22 @@
...
+    &.ukef-separator-spacing-#{$spacing-unit} {
+      > *:not(:first-child) {
+        @include govuk-responsive-padding($spacing-unit, "top");
+      }
+
+      > *:not(:last-child) {
+        @include govuk-responsive-padding($spacing-unit, "bottom");
+      }
+    }
+  }
```
toddym42, line 76: This allows us to do any of:

* ukef-separator-spacing-1
* ukef-separator-spacing-2
* ukef-separator-spacing-3
* ukef-separator-spacing-4
* ukef-separator-spacing-5
* ukef-separator-spacing-6
* ukef-separator-spacing-7
* ukef-separator-spacing-8
* ukef-separator-spacing-9

without having to create a separate rule for each

ChristianMcc (approved):

PR #2586, opened 2024-01-15
chore(DTFS2-6952): update node and npm versions

ChristianMcc:
## Introduction :pencil2:
All our docker files and the deployed versions of our code are using node 20 rather than node 21. We should match the package.json node versions to the deployed version of node.
Note the azure functions are actually at node 14 not 20 but these are already out of sync and need to be upgraded.
Npm should also be upgraded to the npm version that is coupled with node 20 (10+)

Update package.json files to have engines pointed at node 20 and npm 10

### DIFF
--- a/azure-functions/acbs-function/package.json
+++ b/azure-functions/acbs-function/package.json
@@ -11,7 +11,6 @@
     "axios": "^0.27.2",
     "date-fns": "^2.30.0",
     "dotenv": "^16.3.1",
-    "dtfs": "file:../..",
     "durable-functions": "^1.5.4",
     "eslint": "^8.55.0",
     "eslint-config-airbnb-base": "^15.0.0",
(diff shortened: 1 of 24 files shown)


azure-functions/acbs-function/package.json
```diff
@@ -11,6 +11,7 @@
     "axios": "^0.27.2",
     "date-fns": "^2.30.0",
     "dotenv": "^16.3.1",
+    "dtfs": "file:../..",
```
abhi-markan, line 14: Is this required?

abhi-markan (changes requested):

azure-functions/acbs-function/package.json
ChristianMcc: Ooh these are not, needs to be removed from running the npm install:all on windows, will remove! :)

toddym42: This _should_ stop happening when we move to using npm workspaces :D

abhi-markan (approved):

PR #2588, opened 2024-01-15
feat(FN-2051): allow numeric months in filename

JoshBinns2000:
## Introduction :pencil2:
We want to allow uploaded utilisation reports to use a numeric representation for the report period month that the report is being submitted for. Also want to disallow filenames which included `'SUBMITTED'` in the filename.

Changes the validator regex pattern to also search for the numeric representations and adds a final step to check whether or not the filename contains `'SUBMITTED'`.


portal/server/controllers/utilisation-report-service/utilisation-report-upload/utilisation-report-validator.js
```diff
@@ -104,21 +107,24 @@ const validateCsvData = (csvData) => {
  */
 const validateFilenameContainsReportPeriod = (filename, dueReportPeriod) => {
```
toddym42, line 108: As we're now including other filename validations in this function we could make the name more generic, like `validateUtilisationReportFilename`?

JoshBinns2000: Changed to `validateFilenameFormat`

toddym42 (approved):

PR #2591, opened 2024-01-16
chore: log sign-in link when in development

toddym42:
## Introduction :pencil2:
Automatically log the Portal sign-in link when in development to easily allow sign-in when emails are not actually sent.

### DIFF
--- a/portal-api/src/v1/users/sign-in-link.service.js
+++ b/portal-api/src/v1/users/sign-in-link.service.js
@@ -113,6 +113,10 @@ class SignInLinkService {
   }
 
   async #sendSignInLinkEmail({ userEmail, userFirstName, userLastName, signInLink }) {
+    if (process.env.NODE_ENV === 'development') {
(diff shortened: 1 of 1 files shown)


portal-api/src/v1/users/sign-in-link.service.js
```diff
@@ -113,6 +113,10 @@ class SignInLinkService {
   }
 
   async #sendSignInLinkEmail({ userEmail, userFirstName, userLastName, signInLink }) {
+    if (process.env.NODE_ENV === 'development') {
+      console.info('🔗 signInLink 🔗', signInLink);
```
JoshBinns2000, line 117: Is there a reason to not just `return` after `console.info(...)` so that we don't even bother emailing the sign in link?

toddym42: I didn't want to interfere with the normal flow at all, even if it fails

AlexBramhill (approved):

PR #2593, opened 2024-01-16
feat(FN-2050): handle non-ascii characters in csv files

ChristianMcc:
## Introduction :pencil2:
We need to handle non-ascii characters in csv files in case during the download any of the characters are not ascii characters due to the different encoding used

Replace any non-ascii characters including the 65533 unicode question mark character with spaces. This works for the current mock reports we are using with the allowed headers

### DIFF
--- a/portal/server/utils/csv-utils.js
+++ b/portal/server/utils/csv-utils.js
@@ -166,7 +166,12 @@ const csvBasedCsvToJsonPromise = async (csvBuffer) => {
       csvStream
         .pipe(
           csv({
-            mapHeaders: ({ header }) => header.toLowerCase().replace(/\s/g, ' ').trim(),
+            mapHeaders: ({ header }) =>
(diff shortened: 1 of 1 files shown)


JoshBinns2000 (dismissed):
lgtm

toddym42 (approved):

portal/server/utils/csv-utils.js
```diff
@@ -163,10 +163,16 @@ const csvBasedCsvToJsonPromise = async (csvBuffer) => {
   return new Promise((resolve, reject) => {
     try {
       const csvData = [];
+      const nonPrintableAsciiCharacterRegex = /[^ -~]+/g;
```
JoshBinns2000, line 166: What does this regex pattern actually do?

JoshBinns2000 (approved):

PR #2594, opened 2024-01-16
chore(DTFS2-7023): add typescript to portal routes

MarRobSoftwire:
## Introduction :pencil2:
Portal routes don't use typescript

* convert routes files to TS
* Add types for user & bank (taken from `dtfs-central-api` in main-stbr)


portal/server/generateApp.ts
```diff
@@ -61,7 +69,18 @@ const generateApp = () => {
     };
   }
 
-  const redisClient = redis.createClient(process.env.REDIS_PORT, process.env.REDIS_HOSTNAME, redisOptions);
+  const redisPort = process.env.REDIS_PORT as string;
+
+  // if (!process.env.REDIS_PORT || Number.isNaN(parseInt(process.env.REDIS_PORT, 10))) {
+  //   console.error('Portal UI server - invalid REDIS_PORT', process.env.REDIS_PORT);
+  //   throw new InvalidEnvironmentVariableError('Invalid redis port value.');
+  // }
```
JoshBinns2000, line 77: Is this commented out code supposed to be here?

portal/server/generateApp.ts
```diff
@@ -97,7 +116,7 @@ const generateApp = () => {
   app.use(copyCsrfTokenFromQueryToBody());
   app.use(csrf({
     cookie: {
-      ...cookie,
+      ...cookie as csrf.CookieOptions,
```
JoshBinns2000, line 119: You can remove the need for this explicit cast by setting the type of `cookie` to `session.CookieOptions & csrf.CookieOptions` (I think it's best to avoid explicit casts like this where possible - the most appropriate use case for it that I've found is in test files)

portal/server/generateApp.ts
```diff
@@ -121,21 +140,22 @@ const generateApp = () => {
...
-  // error handler
-  app.use((error, req, res, next) => {
-    if (error.code === 'EBADCSRFTOKEN') {
+  const errorHandler: ErrorRequestHandler = (
+    error: {code: string, statusCode: number},
+    _req,
+    res,
+    next,
+  ) => {
+    if (error?.code === 'EBADCSRFTOKEN') {
```
JoshBinns2000, line 149: Seeing as we're keeping the optional chaining, it might be worthwhile to make the type of `error` be `{ code: string; statusCode: number } | undefined` (also notice the change from a comma to a semi-colon - running prettier in this file should handle that change, but properties in a type definition should be separated by a semi-colon)

portal/server/types/db-models/banks.ts
```diff
@@ -0,0 +1,21 @@
+import { WithId } from 'mongodb';
+
+type ReportPeriodSchedule = {
+  startMonth: number;
+  endMonth: number;
+}
```
JoshBinns2000, line 6: There should be a semi-colon here (prettier will add it)

portal/server/types/date.ts
JoshBinns2000, line 1: I think the only type you use from this file is `UnixTimestamp`, so I would only add that one I think

JoshBinns2000 (commented):
A few issues with code formatting - I would recommend running prettier in all your files (especially with changing from `js` to `ts`). Happy to see a `generateApp` being upgraded :)

portal/server/types/db-models/banks.ts
MarRobSoftwire: Oops, this might be missing on feature branch too

portal/server/generateApp.ts
MarRobSoftwire: Had a comment on previous PR forgot to add

The REDIS_PORT env var is undefined but redis.createClient() doesn't allow this to be passed in.
Assume this variable is used in prod but not sure how to validate it

portal/server/generateApp.ts
```diff
@@ -61,7 +66,11 @@ const generateApp = () => {
     };
   }
 
-  const redisClient = redis.createClient(process.env.REDIS_PORT, process.env.REDIS_HOSTNAME, redisOptions);
+  const redisClient = redis.createClient(
+    parseInt(process.env.REDIS_PORT ?? '6379', 10),
```
MarRobSoftwire, line 70: TS doesn't like `REDIS_PORT`  being undefined here.
Should I add more error handling for missing/invalid environment variables?

JoshBinns2000: In E2ETF, we introduce a helper function called `asString` which uses a type predicate telling typescript to treat the value as a string and throws if the value is not a string:
```
/**
 * @param value - the value to check
 * @param context - provides context in the error message if 'value' is not a
 *  string. Usually would be the name of the variable being passed in
 */
export const asString = (value: unknown, context: string): string => {
  if (!isString(value)) {
    throw new Error(`Expected ${context} to be a string, but was ${typeof value}`);
  }

  return value as string;
};
```

This comes from https://github.com/UK-Export-Finance/dtfs2/blob/main-stbr-tfm-5/trade-finance-manager-api/src/utils/validation.ts

JoshBinns2000: Alternatively, just use `process.env.REDIS_PORT || '6379'` instead of `??`

portal/api-test.jest.config.js
JoshBinns2000, line 1: Under the assumption that E2ETF gets merged before this, you will probably see a lot of conflicts in this file as a lot of this logic is moved to a file called `jest.common.config.js`, so just watch out for that :)

portal/init.sh
JoshBinns2000, line 1: Am working on getting this added as an environment variable, so you can remove the flag being used here (unless this is needed to run the tests?)

portal/server/generateApp.ts
```diff
@@ -121,21 +130,22 @@ const generateApp = () => {
 
   app.get('*', (req, res) => res.render('page-not-found.njk', { user: req.session.user }));
 
-  // error handler
-  app.use((error, req, res, next) => {
-    if (error.code === 'EBADCSRFTOKEN') {
+  const errorHandler: ErrorRequestHandler = (
+    error: {code: string; statusCode: number} | undefined,
```
JoshBinns2000, line 134: Need to run prettier on this

portal/init.sh
MarRobSoftwire: Sadly is for the pipeline

MarRobSoftwire, line 1: This file shouldn't have been recreated

portal/Dockerfile
```diff
@@ -14,4 +14,4 @@ RUN npm cache clean --force
 COPY . .
 
 # Execute Script
-CMD ["npx", "ts-node", "server/index.ts"]
+CMD ["npx", "ts-node", "--transpile-only", "server/index.ts"]
```
AlexBramhill, line 17: Why do we have this here and not elsewhere where we use TS out of curiousity

portal/server/types/db-models/banks.ts
```diff
@@ -0,0 +1,21 @@
+import { WithId } from 'mongodb';
+
+type ReportPeriodSchedule = {
+  startMonth: number;
+  endMonth: number;
+};
+
+export type Bank = WithId<{
```
AlexBramhill, line 8: prettify?

portal/server/types/db-models/users.ts
```diff
@@ -0,0 +1,28 @@
+import { WithId } from 'mongodb';
+import { PortalRole } from '../portal/portal-role';
+import { Bank } from './banks';
+import { UnixTimestamp } from '../date';
+import { Prettify } from '../types-helper';
+
+export type PortalUser = Prettify<
+  WithId<{
```
AlexBramhill, line 8: I'd be tempted to just have `_id` as string in the front end, as thats all we care about really.

@toddym42 do you have an opinion on this?

AlexBramhill: Would mean we dont need mongodb in a frontend service, which really smells to me

AlexBramhill (commented):
Some comments, also maybe something to discuss with Mike

portal/Dockerfile
toddym42: We do this on production with the `TS_NODE_TRANSPILE_ONLY ` env var set to `true`, so this shouldn't be needed here. Leaving it out means we get an extra type check during dev deployments.

portal/init.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
toddym42: This has been deleted on main now. Maybe added back in a merge conflict?

portal/Dockerfile
MarRobSoftwire: The problem seems to come from `tsc` and `ts-node` being slightly different.
IDEs seem to use `tsc`'s type-checking which get the session types from the `express-session.d.ts` file.

As to why this isn't needed in `tfm-ui` I'm not sure

portal/server/types/db-models/banks.ts
```diff
@@ -0,0 +1,21 @@
...
+  emails: string[];
+  companiesHouseNo: string;
+  partyUrn: string;
+  hasGefAccessOnly: boolean;
+  paymentOfficerTeam: {
+    teamName: string;
+    email: string;
+  };
+  utilisationReportPeriodSchedule: ReportPeriodSchedule[];
+}>;
```
toddym42, line 21: Missing a new `isVisibleInTfmUtilisationReports: boolean;` property here (see `dtfs-central-api`)

portal/server/types/db-models/users.ts
toddym42: I think it's technically correct to have `_id: ObjectId` here as this is meant to represent the db-model, not the API response. It doesn't feel quite right to have the DB model here though, as it's not an object that is every actually present in the portal ui project. I think we're defining it here because the `PortalSessionUser` is derived from it, but we lose the benefit of deriving from the actual DB model because this is just a copy and there's no guarantee that this is kept in line with the version in `dtfs-central-api`.

I think then maybe it just makes sense to remove this and explicitly declare all the properties of `PortalSessionUser`. After the workspaces work has been merged to main (hopefully soon!) we could share the DB models in the common library and then have rethink.

In `PortalSessionUser` I would have the `_id` as `string` because it's been serialised at that point when returned by the API.

portal/Dockerfile
JoshBinns2000: It should work without this flag being explicitly used, so I don't think you should add it here. If we are going to add this it should be via the environment variable flag [here](https://typestrong.org/ts-node/docs/options/#transpileonly) :)

portal/server/types/db-models/banks.ts
```diff
@@ -0,0 +1,21 @@
+import { WithId } from 'mongodb';
+
+type ReportPeriodSchedule = {
+  startMonth: number;
```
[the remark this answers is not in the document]
JoshBinns2000: These types are about to be updated (depending on which gets merged first). Look [here](https://github.com/UK-Export-Finance/dtfs2/pull/2751/files#diff-423bf5696447495c86dac637417de60d4ba48f70d44dcfef178a5e997e1e5a95) for the new types

portal/server/types/db-models/users.ts
JoshBinns2000: `_id` will be a string in the frontend as `ObjectId` is serialised when sent as JSON, so `WithId` should be replaced with `_id: string` :)

portal/server/types/db-models/banks.ts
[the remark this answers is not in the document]
MarRobSoftwire: dtfs-central just uses number in the bank db-model. I'm not actually sure whether it's one or zero indexed

portal/server/types/db-models/users.ts
```diff
@@ -0,0 +1,28 @@
...
+    surname: string;
+    roles: PortalRole[];
+    email: string;
+    bank: Bank;
+    username: string;
+    salt: string;
+    hash: string;
+    resetPwdToken?: string;
+    resetPwdTimestamp?: string;
+    lastLogin?: UnixTimestamp;
```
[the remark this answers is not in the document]
JoshBinns2000: The type of `PortalSessionUser` should be based on the result of this `portal-api` function:
```
const sanitizeUser = (user) => ({
  username: user.username,
  firstname: user.firstname,
  surname: user.surname,
  email: user.email,
  roles: user.roles,
  bank: user.bank,
  timezone: user.timezone,
  lastLogin: user.lastLogin,
  'user-status': user['user-status'],
  disabled: user.disabled,
  signInLinkSendDate: user.signInLinkSendDate,
  signInLinkSendCount: user.signInLinkSendCount,
  _id: user._id,
});
```

portal/server/types/db-models/banks.ts
AlexBramhill: I have an issue with us even using these db entries, I think we can get rid of them for now

portal/server/types/db-models/users.ts
AlexBramhill: I think you've summarised my thoughts a lot more coherently than me Mike :D

portal/server/types/db-models/users.ts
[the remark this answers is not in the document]
JoshBinns2000: We have implemented this type in [this PR](https://github.com/UK-Export-Finance/dtfs2/pull/2751/files#diff-d21b8bbac6f76595a45c1579d2b85699409ed7f03ec706d96b3e765e4491dada) :)

portal/server/types/db-models/banks.ts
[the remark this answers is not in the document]
JoshBinns2000: It is a one indexed month - up to you whether or not you want to update it to use that alias :)

portal/server/types/db-models/users.ts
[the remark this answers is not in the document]
AlexBramhill: fwiw as well, the 2fa process doesnt actually store a `user` (good job past me, I hope)
```
    // We do not store this in the user object to avoid existing logic using the existence of a `user` object to draw elements
    req.session.userEmail = userEmail;
```

portal/server/types/db-models/users.ts
```diff
@@ -0,0 +1,28 @@
...
+    loginFailureCount?: number;
+    sessionIdentifier?: string;
+    signInLinkSendDate?: Date;
```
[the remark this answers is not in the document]
MarRobSoftwire: Should be a timestamp but also not stored in session

portal/server/types/portal/portal-session-user.ts
```diff
@@ -0,0 +1,23 @@
+import { Prettify } from '../types-helper';
+import { PortalUser } from '../db-models/users';
+import { SessionBank } from '../session-bank';
+
+export type PortalSessionUser = Prettify<
+  Pick<
+    PortalUser,
+    | 'username'
```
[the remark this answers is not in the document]
MarRobSoftwire: Have checked & updated

portal/server/controllers/utilisation-report-service/utilisation-report-upload/utilisation-report-upload-errors.test.ts
```diff
@@ -1,12 +1,19 @@
...
+import { asLoggedInUserSession } from '../../../helpers/express-session';
 
 jest.mock('./utilisation-report-validator');
+jest.mock('../../../helpers/express-session');
 
 describe('utilisation-report-upload-errors', () => {
   const href = '#utilisation-report-file-upload';
 
+  beforeAll(() => {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any, @typescript-eslint/no-unsafe-return
```
abhi-markan, line 13: Can we avoid inline rule override please?

portal/server/generateApp.ts
```diff
@@ -61,7 +68,7 @@ const generateApp = () => {
     };
   }
 
-  const redisClient = redis.createClient(process.env.REDIS_PORT, process.env.REDIS_HOSTNAME, redisOptions);
+  const redisClient = redis.createClient(parseInt(process.env.REDIS_PORT || '6379', 10), process.env.REDIS_HOSTNAME, redisOptions);
```
abhi-markan, line 71: `REDIS_PORT` will always exists in `env`, for fail-safe purpose if the port must be defined can it be centralised to a constant please and perhaps similar can be done for all the ports env variables.

portal/server/generateApp.ts
```diff
@@ -95,17 +102,21 @@ const generateApp = () => {
...
-    cookie: {
-      ...cookie,
-      maxAge: 43200, // 12 hours
-    },
-  }));
+  app.use(
+    csrf({
+      cookie: {
+        ...cookie,
+        maxAge: 43200, // 12 hours
```
abhi-markan, line 109: Constant please.

portal/server/generateApp.ts
```diff
@@ -95,17 +102,21 @@ const generateApp = () => {
...
+      },
+    }),
+  );
   app.use(csrfToken());
 
-  app.use(morgan('dev', {
-    skip: (req) => req.url.startsWith('/assets') || req.url.startsWith('/main.js'),
-  }));
+  app.use(
+    morgan('dev', {
```
abhi-markan, line 116: Should `dev` not be picked from `NODE_ENV`?

portal/server/generateApp.ts
```diff
@@ -95,17 +102,21 @@ const generateApp = () => {
...
+      skip: (req) => req.url.startsWith('/assets') || req.url.startsWith('/main.js'),
```
abhi-markan, line 117: This can be a middleware to achieve abstraction.

portal/server/generateApp.ts
```diff
@@ -119,23 +130,23 @@ const generateApp = () => {
 
   app.use('/', routes);
 
-  app.get('*', (req, res) => res.render('page-not-found.njk', { user: req.session.user }));
+  app.get('*', (req, res) => {
```
abhi-markan, line 133: Perhaps a little explanation of steps here will be good.

portal/server/generateApp.ts
```diff
@@ -119,23 +130,23 @@ const generateApp = () => {
...
-  // error handler
-  app.use((error, req, res, next) => {
-    if (error.code === 'EBADCSRFTOKEN') {
-      console.error('The user\'s CSRF token is incorrect, redirecting the user to /.');
+  const errorHandler: ErrorRequestHandler = (error: unknown, _req, res, next) => {
+    if (isHttpError(error) && error.code === 'EBADCSRFTOKEN') {
+      console.error("The user's CSRF token is incorrect, redirecting the user to /.");
       // handle CSRF token errors here
-      res.status(error.statusCode || 500);
+      res.status(Number(error.statusCode) || 500);
```
abhi-markan, line 143: `error.statusCode` should always be a number?

portal/server/routes/admin/index.ts
```diff
@@ -0,0 +1,13 @@
+import express from 'express';
+
+import userAdminRoutes from './users';
+
+import { validateToken, validateRole } from '../middleware';
+import { ROLES } from '../../constants';
+
+const router = express.Router();
+
+router.use('/admin/*', [validateToken, validateRole({ role: [ROLES.ADMIN] })]);
```
abhi-markan, line 10: Should be these two lines be one single line?

portal/server/helpers/express-session.ts
```diff
@@ -0,0 +1,101 @@
...
+import { PartiallyLoggedInSessionData, LoggedInSessionData, UserSessionData } from "../types/express-session";
+
+type Session = Request['session'];
+
+type LoggedInUserSession = Session & LoggedInSessionData;
+type PartiallyLoggedInUserSession = Session & PartiallyLoggedInSessionData;
+type UnknownLogInStatusUserSession = Session & Pick<UserSessionData, 'userToken' | 'loginStatus'>;
+
+/**
+ * By default, all session data will be optional
```
abhi-markan, line 11: Good documentation.

portal/server/helpers/express-session.ts
```diff
@@ -0,0 +1,101 @@
...
+ * present in the session, or throws an `Error` if not.
+ *
+ * Note: this doesn't assert on the validity of the user data values (which
+ * should have already been authenticated in middleware by this point), only
+ * that the values are present.
+ */
+export const asLoggedInUserSession = (session: Session): LoggedInUserSession  => {
+  const { user, userToken, loginStatus } = session as LoggedInUserSession;
+
+  if (loginStatus !== 'Valid 2FA' ) {
```
abhi-markan, line 24: `Valid 2FA` has been repeated multiple times in the micro-service, could benefit from a constant value please or perhaps a helper function which ascertain whether value is `Valid 2FA` or not.

abhi-markan: Can you also please remove inline Eslint override.

portal/server/helpers/express-session.ts
```diff
@@ -0,0 +1,101 @@
...
+    // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
+    throw Error(`User is not fully logged in. Log in status ${loginStatus}`)
```
abhi-markan, line 26: Should we throw new `Error()` with `cause`?
[Reference](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/Error)

portal/server/helpers/express-session.ts
```diff
@@ -0,0 +1,101 @@
...
+ * should have already been authenticated in middleware by this point), only
+ * that the values are present.
+ */
+
+
+export const withUnknownLoginStatusUserSession = (session: Session): UnknownLogInStatusUserSession  => {
+  const { userToken, loginStatus } = session as UnknownLogInStatusUserSession;
+
+  if (loginStatus !== 'Valid 2FA' && loginStatus !== 'Valid username and password') {
+    // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
```
abhi-markan, line 92: * `Valid 2FA` try to keep the code DRY please.
* Eslint inline rule.

portal/server/helpers/express-session.ts
```diff
@@ -0,0 +1,101 @@
...
+    throw Error(`User is not fully logged in. Log in status ${loginStatus}`)
```
abhi-markan, line 93: Perhaps use `cause` object?

abhi-markan (changes requested):

portal/server/generateApp.ts
MarRobSoftwire: The other middleware, etc accept an `undefined` port but redis doesn't. I'll add a constant for `REDIS_PORT` but I think leaving the others to handle undefined env vars themselves is ok

portal/server/generateApp.ts
MarRobSoftwire: This seems to be an option on the middleware
https://expressjs.com/en/resources/middleware/morgan.html
> Concise output colored by response status for development use. The :status token will be colored green for success codes, red for server error codes, yellow for client error codes, cyan for redirection codes, and uncolored for information codes.

portal/server/generateApp.ts
MarRobSoftwire: IMO this won't make the code any clearer since it isn't repeated anywhere (within Portal at least).

I guess this is the kind of set up we'd want to commonise across all services so we've got consistent logging

abhi-markan (approved):

PR #2596, opened 2024-01-17
feat(FN-1855): add npm workspaces and common package

toddym42:
## Introduction :pencil2:


README.md
```diff
@@ -48,14 +49,54 @@ This documentation provides a comprehensive overview of the UKEF Digital TradeFi
...
-9. Run `npm install` in the root folder of the repository.
+7. run `npm install` in the root folder of the repository. (note: this will install dependencies for the entire project, including those specified in sub-packages. More details on this in the [npm workspaces](#npm-workspaces) section below)
+8. Start your local environment with `docker-compose -f docker-compose.dev.yml up --build`.
+9. Create mock data by running `npm run load` from the root folder of the repository. This should generate mocks in your database.
 
 Recommended: Install a MongoDB client such as Compass or Robo 3T.
 
 Note: If you're on Windows and experiencing issues with MongoDB, install mongosh for command-line debugging.
 
+## npm workspaces
```
AlexBramhill, line 60: Thank you for adding this :)

docker-compose.dev.yml
```diff
@@ -19,7 +19,9 @@ services:
       - '27017-27019:27017-27019'
 
   dtfs-central-api:
-    build: ./dtfs-central-api
+    build:
+      context: .
+      dockerfile: ./dtfs-central-api/Dockerfile
```
toddym42, line 24: This change is made so that we can build from the root and take into account files from the common library.

This will not result in code from other unrelated packages be added to the docker container, as the Dockerfile in each package only copies across its own package, plus the libs directory which contains the common package.

gef-ui/server/nunjucks-configuration/index.js
```diff
@@ -1,18 +1,19 @@
...
 const configureNunjucks = (opts) => {
   const appViews = [
-    'node_modules/govuk-frontend',
-    'node_modules/@ministryofjustice/frontend',
-    'node_modules/@ministryofjustice/frontend/filters/all',
-    'templates',
+    path.resolve(__dirname, '../../../node_modules/govuk-frontend'),
+    path.resolve(__dirname, '../../../node_modules/@ministryofjustice/frontend'),
+    path.resolve(__dirname, '../../../node_modules/@ministryofjustice/frontend/filters/all'),
+    path.resolve(__dirname, '../../templates'),
```
toddym42, line 16: This was needed due to the `govuk-frontend` and `@ministryofjustice` now being hoisted to the root `node_modules`, and also because of the different contexts in which the docker env and component tests would run.

The docker env would now need:
```
'node_modules/govuk-frontend',
'node_modules/@ministryofjustice/frontend',
'node_modules/@ministryofjustice/frontend/filters/all',
'gef-ui/templates',
```
as its context is the root of the project, but the component tests would need:
```
'../node_modules/govuk-frontend',
'../node_modules/@ministryofjustice/frontend',
'../node_modules/@ministryofjustice/frontend/filters/all',
'templates', 
```
as its context is the root of the `gef-ui` directory.

I played around with changing the jest.config to change the root to the root of the project, but that threw up other issues to do with other packages.

By instead basing the path on `__dirname` (i.e. the directory of this file), the the relative path will always be the same, regardless of the run context.

gef-ui/webpack.common.config.js
```diff
@@ -23,6 +23,12 @@ module.exports = {
       new CssMinimizerPlugin({}),
     ],
   },
+  resolve: {
+    alias: {
+      'node_modules/govuk-frontend': path.resolve(__dirname, '../node_modules/govuk-frontend'),
+      'node_modules/@ministryofjustice': path.resolve(__dirname, '../node_modules/@ministryofjustice'),
+    },
+  },
```
toddym42, line 31: This change is also needed because of the hoisting of dependencies to the root `node_modules`. We can change SCSS files we control from e.g.
```
@import "node_modules/govuk-frontend/govuk/all";
```
to
```
@import "../node_modules/govuk-frontend/govuk/all";
```
but these imports refer to other imports that are part of the package, so not in our control to change. This alias makes the change to all relevant files when building the CSS output.

package.json
```diff
@@ -12,18 +13,31 @@
...
-    "audit:fix:all": "(npm audit --fix); (cd azure-functions/acbs-function && npm audit --fix); (cd azure-functions/number-generator-function && npm audit --fix); (cd dtfs-central-api && npm audit --fix); (cd gef-ui && npm audit --fix); (cd portal && npm audit --fix); (cd portal-api && npm audit --fix); (cd external-api && npm audit --fix); (cd trade-finance-manager-api && npm audit --fix); (cd trade-finance-manager-ui && npm audit --fix); (cd utils/mock-data-loader && npm audit --fix); (cd utils/data-migration && npm audit --fix)",
-    "ci:all": "npm ci && npm ci --prefix azure-functions/acbs-function && npm ci --prefix azure-functions/number-generator-function && npm ci --prefix dtfs-central-api && npm ci --prefix gef-ui && npm ci --prefix portal && npm ci --prefix portal-api && npm ci --prefix external-api && npm ci --prefix trade-finance-manager-api && npm ci --prefix trade-finance-manager-ui && npm ci --prefix utils/mock-data-loader && npm ci --prefix utils/data-migration",
+    "audit:all": "npm audit --if-present",
```
toddym42, line 30: Notes on this `--if-present` flag in the update to the root README.md.

trade-finance-manager-ui/package.json
```diff
@@ -92,7 +94,6 @@
     "file-loader": "^6.2.0",
     "jest": "29.5.0",
     "node-mocks-http": "^1.13.0",
-    "optimize-css-assets-webpack-plugin": "^6.0.1",
```
toddym42, line 95: This package is no longer used. It's been removed in main but looks like it slipped back in when we merged `main` into `main-stbr-tfm-5`. Removing this allows us to not require `--legacy-peer-deps` and do a consistent `npm install` across all the workspace packages.

e2e-tests/gef/cypress/support/commands/loadData.js
```diff
@@ -2,14 +2,11 @@
  * Mock data loader:
  * 1. Clears all the collections
  * 2. Inserts mocked data
-*/
+ */
 
 const loadData = () => {
-  // Ensure project root directory
-  cy.exec('if [ -d "dtfs2" ]; then cd dtfs2; fi');
```
toddym42, line 9: This wasn't doing anything because each `cy.exec` runs in it's own child process (i.e. doesn't affect any other `cy.exec` commands), so even if this navigated it would have no effect on the following `cy.exec` command.

gef-ui/package.json
```diff
@@ -22,14 +22,15 @@
     "api-test": "jest --coverage --verbose --forceExit --config=api-test.jest.config.js",
     "build": "rimraf ./public/js && rimraf ./public/css && webpack --mode production --config webpack.prod.config.js",
     "dev": "webpack --config webpack.dev.config.js --watch",
-    "lint": "eslint ./server ./scripts ./component-tests ./test-helpers ./api-tests",
+    "lint": "eslint ./server ./scripts ./component-tests ./test-helpers ./api-tests && npm run type-check",
```
ChristianMcc, line 25: same thing here won't put it on all, I guess type-check only runs on typescript files so it's fine to be broad?

gef-ui/webpack.common.config.js
```diff
@@ -23,6 +23,13 @@ module.exports = {
       new CssMinimizerPlugin({}),
     ],
   },
+  resolve: {
```
ChristianMcc, line 26: In your docs section noting how hoisting works especially with regard to frontend packages etc. like here would be good

.github/workflows/test.yml
```diff
@@ -145,7 +146,7 @@ jobs:
           npm run load
 
       - name: Execute
-        run: docker-compose exec -T ${{ matrix.services }} npm run api-test
+        run: docker-compose exec -T ${{ matrix.services }} npm run api-test --workspace=${{ matrix.services }}
```
ChristianMcc, line 149: how does this work for services (like libs) that will only have unit tests and not api-tests?

libs/common/tsconfig.json
```diff
@@ -0,0 +1,30 @@
...
+    /* Type Checking */
+    "allowUnreachableCode": false,
+    "noImplicitReturns": true,
+    "noUncheckedIndexedAccess": true,
+    "noUnusedLocals": true,
+    "noUnusedParameters": true,
+    "strict": true,
+
+    /* Modules */
+    "module": "node16",
```
ChristianMcc, line 12: Haven't looked at what everything else is, is node16 just a way to resolve modules as an option and this isn't anything to do with the actual version of node used?

ChristianMcc (commented):
Amazing work, looks good to me! Left a bunch of questions and just needs the documentation section you've mentioned

gef-ui/webpack.common.config.js
toddym42: Separate doc now added [here](https://github.com/UK-Export-Finance/dtfs2/pull/2596/files#diff-0b220ba191203bac20bc8dde2113320ff8bbaeaf016cf66efd7b125c9ab924cd) with some extra notes on the limitations of hoisting.

docs/npm-workspaces.md
```diff
@@ -0,0 +1,72 @@
...
+
+One of the benefits of `npm workspaces` is that common dependencies across all sub-packages will be hoisted to the root `node_modules` with only conflicting dependencies installed in the sub-package `node_modules`. This does however present some challenges. Below are some notes on challenges faced and how we've mitigated them:
+
+Our UI packages (`portal`, `gef-ui` and `trade-finance-manager-ui`) make use of the [govuk-frontend](https://www.npmjs.com/package/govuk-frontend) and [@ministryofjustice/frontend](https://www.npmjs.com/package/@ministryofjustice/frontend) libraries for UI components and styles. When importing styles from these packages into our own Sass `.scss` files the docs advise us to import as follows:
+
+```sass
+@import "node_modules/govuk-frontend/govuk/all";
+```
+
+However, the `govuk-frontend` package has now been hoisted to the root `node_modules` this path is no longer correct. One way we may think to resolve this is to simply change the import path to:
```
toddym42, line 51: Should be
```
... `node_modules`, so this path is no longer correct.
```

docs/npm-workspaces.md
```diff
@@ -0,0 +1,72 @@
...
+  alias: {
+    'node_modules/govuk-frontend':path.resolve(__dirname, '../node_modules/govuk-frontend'),
+    'node_modules/@ministryofjustice':path.resolve(__dirname, '../node_modules/@ministryofjustice'),
+  },
+},
+```
+
+so that all files, regardless if in our control or not, will be updated when the minified CSS files are created. We can now continue to use `@import "node_modules/govuk-frontend/govuk/all";` as suggested in the library docs.
+
+Another way this could have been resolved would be to prevent hoisting of these two packages. Unfortunately though in npm workspaces the choice is to hoist all or nothing (see [install-strategy](https://docs.npmjs.com/cli/v9/commands/npm-install#install-strategy)), however in some workspaces managers there are options to prevent hoisting at a package-by-package level (e.g. see the [nohoist](https://classic.yarnpkg.com/blog/2018/02/15/nohoist/) option provided by Yarn Workspaces). Hopefully this functionality will be implemented in npm workspaces in future.
```
toddym42, line 72: Maybe we _should_ use `"install-strategy": "nested"` to prevent hoisting of common node_modules to the root and prevent issues like the above.

We'd still get the benefits of simplified commands that run on all workspace projects, and auto-symlinking to enabled shared code.

ChristianMcc: After discussing a few things with you in person I think keeping like this for now is good, my biggest thing both here and in general on DTFS would be some tree shaking of modules and we should re-visit this if we ever do the tree shaking

dtfs-central-api/package.json
```diff
@@ -20,14 +20,16 @@
   "scripts": {
     "api-test": "jest --coverage --verbose --forceExit --runInBand --config=api-test.jest.config.js",
     "api-test-dev": "jest --verbose --forceExit --runInBand --config=api-test.jest.config.js",
-    "lint": "eslint ./src ./api-tests",
+    "lint": "eslint ./src ./api-tests && npm run type-check",
     "lint:fix": "eslint ./src ./api-tests --fix",
     "start": "npx ts-node src/index.ts",
     "start:dev": "npx nodemon --exec npm run start",
+    "type-check": "tsc --noEmit",
```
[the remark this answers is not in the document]
ChristianMcc: Nothing too strong, makes sense re auto-generated files. There's a few places that aren't caught by /src and /api-tests etc. so maybe worth a quick review but they seem like 'one off scripts' etc. that maybe could be deprecated anyway!

ChristianMcc (dismissed):
Looks good to me, thanks for the reply to all my comments (from 2 weeks ago oops)

ChristianMcc (dismissed):

libs/common/src/helpers/string.ts
```diff
@@ -0,0 +1,5 @@
+export const isString = (value: unknown): value is string => typeof value === 'string' || value instanceof String;
+
+export const isNullUndefinedOrEmptyString = (value: unknown): boolean => !value || (isString(value) && !value.trim().length);
```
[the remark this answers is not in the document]
AlexBramhill: Names great!

AlexBramhill (approved):

AlexBramhill (dismissed):

docker-compose.dev.yml
```diff
@@ -29,7 +31,8 @@ services:
       - '5005:5005' # http
       - '9223:9223' # debug
     volumes:
-      - ./dtfs-central-api/src:/app/src:ro
+      - ./dtfs-central-api/src:/app/dtfs-central-api/src:ro
```
abhi-markan, line 34: Is there a reason to build up individual containers with path convention `/app/[container]/..`, we follow `/app/..` path convention for building containers.

docs/npm-workspaces.md
```diff
@@ -0,0 +1,72 @@
+# npm workspaces
```
abhi-markan, line 1: Brilliant documentation 👍🏼 
We don't really follow creation documentation inside `docs` directory I would suggest placing this inside `libs/README.md` or append to root `README.md`, though appreciate root README.MD is getting rather large.

dtfs-central-api/Dockerfile
```diff
@@ -9,9 +9,12 @@ ENV GITHUB_SHA=$GITHUB_SHA
...
+COPY libs libs
+COPY dtfs-central-api dtfs-central-api
+
+RUN npm run ci:all
 RUN npm cache clean --force
-COPY . .
 
 # Execute Script
-CMD ["npx", "ts-node", "src/index.ts"]
+CMD ["npx", "ts-node", "dtfs-central-api/src/index.ts"]
```
abhi-markan, line 20: Not entirely convinced with this path convention, when already inside the container Dockerfile.

dtfs-central-api/Dockerfile
```diff
@@ -9,9 +9,12 @@ ENV GITHUB_SHA=$GITHUB_SHA
 WORKDIR /app
 COPY package.json .
 COPY package-lock.json .
-RUN npm ci
+
+COPY libs libs
+COPY dtfs-central-api dtfs-central-api
+
+RUN npm run ci:all
```
abhi-markan, line 16: Should this be not `npm run ci` as I am only clean installing container in question here?

dtfs-central-api/package.json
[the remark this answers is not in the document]
abhi-markan: Makes sense to replace `lint` script with `"lint": "eslint \"src/**/*.{ts,js}\" && npm run type-check",` (as you have suggested above for common lib`

dtfs-central-api/src/v1/validation/create-facility-rules/associated-deal-id.js
```diff
@@ -1,12 +1,12 @@
+const { isNonEmptyString } = require('@ukef/dtfs2-common');
 const { orderNumber } = require('../../../utils/error-list-order-number');
-const { hasValue } = require('../../../utils/string');
 
 // eslint-disable-next-line default-param-last
```
abhi-markan, line 4: Can we remove this, (appreciate its not a new code).

dtfs-central-api/src/v1/validation/create-facility-rules/type.js
```diff
@@ -1,20 +1,20 @@
+const { isNonEmptyString } = require('@ukef/dtfs2-common');
 const { orderNumber } = require('../../../utils/error-list-order-number');
-const { hasValue } = require('../../../utils/string');
 const { FACILITY_TYPE } = require('../../../constants/facilities');
 
 // eslint-disable-next-line default-param-last
```
abhi-markan, line 5: Same as above please.

e2e-tests/gef/cypress/support/commands/loadData.js
```diff
@@ -2,14 +2,11 @@
  * Mock data loader:
  * 1. Clears all the collections
  * 2. Inserts mocked data
-*/
+ */
```
abhi-markan, line 5: Extra space?

e2e-tests/support/tasks.js
```diff
@@ -10,8 +10,7 @@ module.exports = {
 
     return {
       log(message) {
-        // eslint-disable-next-line no-console
-        console.log(message);
+        console.info(message);
```
abhi-markan, line 13: Can you change it to following please, so to ensure its coming from Cypress.

```javascript
console.info('Cypress log: ', message)
```

external-api/package.json
```diff
@@ -19,22 +19,14 @@
   "main": "src/index.ts",
   "scripts": {
     "api-test": "jest --coverage --verbose --forceExit --runInBand --config=api-test.jest.config.js",
-    "lint": "eslint ./src ./api-tests --ext .ts",
+    "lint": "eslint ./src ./api-tests --ext .ts && npm run type-check",
```
abhi-markan, line 22: Same as before

gef-ui/package.json
```diff
@@ -22,14 +22,15 @@
...
     "lint:fix": "eslint ./server ./scripts ./component-tests ./test-helpers ./api-tests --fix",
     "start": "npx ts-node server/index.ts",
     "test": "jest --coverage --verbose --config=jest.config.js",
     "test-dev": "jest --verbose --config=jest.config.js",
     "test-file": "jest --config=jest.config.js --verbose --testMatch",
     "test-quick": "jest --config=jest.config.js",
-    "test-watch": "jest --watchAll --coverage --verbose --config=jest.config.js --detectOpenHandles"
+    "test-watch": "jest --watchAll --coverage --verbose --config=jest.config.js --detectOpenHandles",
```
abhi-markan, line 32: Do we need `test-watch` and `test-quick`, `test-file` and `test-dev` scripts?

e2e-tests/gef/cypress/support/commands/loadData.js
```diff
@@ -2,14 +2,11 @@
...
 
 const loadData = () => {
-  // Ensure project root directory
-  cy.exec('if [ -d "dtfs2" ]; then cd dtfs2; fi');
-
   // Load mock data
-  cy.exec('npm run load');
+  cy.exec('cd ../.. && npm run load');
```
abhi-markan, line 9: Now we have `package.json` in `e2e-tests` should we not add `npm run load` script which then refers to the `root` package `load` script and then we can then directory traversing?

libs/common/package.json
```diff
@@ -0,0 +1,43 @@
...
+  "main": "src/index.ts",
+  "types": "./src/index.ts",
+  "scripts": {
+    "lint": "eslint \"src/**/*.{ts,js}\" && npm run type-check",
+    "lint:fix": "eslint \"src/**/*.{ts,js}\"",
+    "prettier": "prettier --check \"src/**/*.{ts,js}\"",
+    "prettier:fix": "prettier --write \"src/**/*.{ts,js}\"",
+    "type-check": "tsc --noEmit",
+    "unit-test": "jest --coverage --verbose",
+    "unit-test-quick": "jest"
```
abhi-markan, line 29: Do we need this script?

libs/common/src/helpers/string.test.ts
```diff
@@ -0,0 +1,49 @@
...
+    it.each`
+      value        | expected
+      ${null}      | ${true}
+      ${undefined} | ${true}
+      ${''}        | ${true}
+      ${' '}       | ${true}
+      ${'string'}  | ${false}
+      ${' string'} | ${false}
+      ${'string '} | ${false}
+    `('returns $expected when the value is $value', ({ value, expected }: { value: unknown; expected: boolean }) => {
```
abhi-markan, line 30: Shouldn't `value: any` rather `value: unkown`?

libs/common/src/helpers/string.ts
```diff
@@ -0,0 +1,5 @@
+export const isString = (value: unknown): value is string => typeof value === 'string' || value instanceof String;
```
abhi-markan, line 1: Docstrings please.

abhi-markan: Should all the `value: any` here?

abhi-markan: Should we keep the signature to `: string`?

libs/common/src/helpers/validation.ts
```diff
@@ -0,0 +1,14 @@
+import { isString } from './string';
+
+/**
+ * @param value - the value to check
+ * @param context - provides context in the error message if value is not a string. Usually would be the name
+ *  of the variable being passed in
+ */
+export const asString = (value: unknown, context: string): string => {
```
abhi-markan, line 8: `value: any`?

libs/common/src/types/types-helper.ts
```diff
@@ -0,0 +1,7 @@
+export type ValuesOf<T> = T[keyof T];
+
+// https://www.totaltypescript.com/concepts/the-prettify-helper
+export type Prettify<T> = {
+  [K in keyof T]: T[K];
+  // eslint-disable-next-line @typescript-eslint/ban-types
```
abhi-markan, line 6: Can we remove inline script please?

package.json
```diff
@@ -33,29 +47,22 @@
       "cspell lint --gitignore --no-must-find-files --unique --no-progress --show-suggestions --color"
     ]
   },
+  "dependencies": {
+    "axios": "^1.6.7"
```
abhi-markan, line 51: I believe this used by `utils/**` should we create a `package.json`?

portal-api/package.json
```diff
@@ -19,9 +19,10 @@
   "main": "src/index.ts",
   "scripts": {
     "api-test": "jest --coverage --verbose --runInBand --forceExit --clearMocks --config=api-test.jest.config.js",
-    "lint": "eslint ./src ./api-tests ./test-helpers ./__mocks__",
+    "lint": "eslint ./src ./api-tests ./test-helpers ./__mocks__ && npm run type-check",
```
abhi-markan, line 22: Same as above.

abhi-markan (changes requested):
Absolutely brilliant work.
Only few comments to address (Some repeating). Thank you!

libs/common/src/helpers/string.ts
toddym42: > Should all the value: any here?

Using type of `unknown` gives us better type-safety. See my previous comment in the adjacet `string.test.ts` file.

> Should we keep the signature to : string?

The basic return type would be `: boolean`, but by using a [type predicate](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates) of `: value is string` we make the function much more powerful. This tells the TypeScript compiler that if the function returns true then the type of `value` is `string`, so you don't then need to cast it.

e.g. if we just returned `boolean` we would need to add a cast after returning `true` from the `isString` function, even though we now know the `value` is a `string`
```
if (isString(value)) {
  functionThatRequiresAString(value as string);
}
```
but with the `value is string` type predicate the compiler now knows `value` has type of `string` within the `if` block, so can just do
```
if (isString(value)) {
  functionThatRequiresAString(value);
}

File: libs/common/src/helpers/validation.ts (hunk 30 above)
Comment (toddym42) [reply libs/common/src/helpers/validation.ts:8]: See previous comments.

File: libs/common/src/types/types-helper.ts (hunk 31 above)
Comment (toddym42) [reply libs/common/src/types/types-helper.ts:6]: I think this would be a good case for an inline rule because it's an exception rather than a rule we'd want to follow elsewhere. We use `{}` to follow the type mentioned in the in the referenced link, but would want this error to still come up if we tried to use it in our own code elsewhere.

File: package.json (hunk 32 above)
Comment (toddym42) [reply package.json:51]: `utils/data-migration` and `utils/mock-data-loader` each already have their own `package.json`. Did you have something else in mind?

File: external-api/package.json (hunk 24 above)
Comment (toddym42) [reply external-api/package.json:22]: As discussed, will raise a separate PR to review lint scripts to make them more general.

File: gef-ui/package.json (hunk 25 above)
Comment (toddym42) [reply gef-ui/package.json:32]: As discussed, these will be reviewed separately @abhi-markan

File: libs/common/src/types/types-helper.ts (hunk 31 above)
Comment (toddym42) [reply libs/common/src/types/types-helper.ts:6]: As discussed, will remove from here for now and review use on main.

File: package.json (hunk 32 above)
Comment (toddym42) [reply package.json:51]: As discussed, we'll review a common utils package.json separately.

File: e2e-tests/gef/cypress/support/commands/loadData.js [hunk 34]
```diff
@@ -2,14 +2,12 @@
...
+ */
 
 const loadData = () => {
-  // Ensure project root directory
-  cy.exec('if [ -d "dtfs2" ]; then cd dtfs2; fi');
-
   // Load mock data
-  cy.exec('npm run load');
+  // Command context is the 'dtfs2/e2e-tests/gef' directory
+  cy.exec('cd .. && npm run load');
```
abhi-markan, line 10: It should be just `npm run load`

abhi-markan (changes requested):
1 minor comment to address please.

e2e-tests/gef/cypress/support/commands/loadData.js
toddym42: The context of this command is ‘dtfs2/e2e-tests/gef’, not ‘dtfs2/e2e-tests’, so this command is correct, and points to the new ‘load’ script I added to e2e-tests, as you requested. I added a comment above to clarify the context. The tests that use this are all passing.

Please take another look.

ChristianMcc (approved):

abhi-markan (approved):

PR #2598, opened 2024-01-17
fix(DTFS2-6963): Holistic user service approach to block users who have `disabled` `true`

ChristianMcc:
## Introduction
We needed an approach we can use everywhere which validates the status of a user and if they are disabled

## Resolution
Introduce a user service to take care of validating a user's status and disabled state. Add this on the relevant login endpoints during the 'login in progress' part of the 2fa section. Then after this add it to the passport validation so that we now check on every API request if the user is in a valid state, meaning any disabled users don't have access to the system straight away (rather than having a full 12 hour session to do whatever they want with!)


.github/workflows/test.yml
```diff
@@ -8,7 +8,7 @@ run-name: Executing test QA on ${{ github.repository }} 🚀
 
 on:
   pull_request:
-    branches: [main, fix/DTFS2-6910/fix-issue-withh-login-page-redirect]
+    branches: [main, fix/DTFS2-6910/fix-issue-withh-login-page-redirect, fix/DTFS2-6963/patch-user-disabled-feature-to-sign-in-link]
```
AlexBramhill, line 11: reminder to revert

portal-api/src/v1/users/login.controller.js
```diff
@@ -1,9 +1,8 @@
-const { STATUS } = require('../../constants/user');
 const utils = require('../../crypto/utils');
-const { userIsBlocked, userIsDisabled, usernameOrPasswordIncorrect } = require('../../constants/login-results');
+const { usernameOrPasswordIncorrect } = require('../../constants/login-results');
 const { findByUsername, incrementFailedLoginCount, updateSessionIdentifier } = require('./controller');
 
-module.exports.login = (username, password) =>
+module.exports.login = (username, password, userService) =>
```
AlexBramhill, line 5: We decided on this? Happy to have it

portal-api/src/v1/users/passport.js
```diff
@@ -54,13 +64,21 @@ const baseAuthenticationConfiguration = ({ name, passport, additionalValidation,
...
-const loginCompleteAuth = (passport) => {
-  const additionalValidation = (jwtPayload) => jwtPayload.loginStatus === LOGIN_STATUSES.VALID_2FA;
-  baseAuthenticationConfiguration({ name: 'login-complete', passport, additionalValidation });
+const loginCompleteAuth = (passport, userService) => {
+  const name = 'login-complete';
+  const additionalPayloadValidation = (jwtPayload) => jwtPayload.loginStatus === LOGIN_STATUSES.VALID_2FA;
+  const additionalUserValidation = (user, jwtPayload) => {
+    if (userService.isUserBlockedOrDisabled(user)) {
+      console.error("User with username %s is blocked or disabled for '%s' strategy", jwtPayload.username, name);
+      return true;
```
AlexBramhill, line 73: Surprising -- Could we add in a VALIDATION_RESULT const that is either Passed or Failed

ChristianMcc:
I think I'd like to add an E2E test which has a user logged in with a session, who then gets disabled, and test that we are routed to the login page and session is destroyed. I think Oscar added a nice cypress command that can help with that so will get to it in the morning!

AlexBramhill:
>E2E test which has a user logged in with a session, who then gets disabled, and test that we are routed to the login page and session is destroyed 

We might already have this! :) I think I found oscar's current test in e2e does this (logs in, disables user, clicks on link).

e2e-tests/portal/cypress/e2e/journeys/login/navigating-using-sign-in-link.spec.js
```diff
@@ -140,6 +140,21 @@ context('navigating using sign in link', () => {
       signInLink.visit({ token: NOT_EXPIRED_SIGN_IN_TOKEN.signInTokenFromLink, userId: bank1Maker1Id }, { failOnStatusCode: false });
       signInLink.shouldDisplayAccountSuspended();
     });
+
+    it('Opening a valid sign in link and logging in but then disabling the user and visiting the landing page will redirect the user to the login page', () => {
```
ChristianMcc, line 144: this tests disabling a user once they are logged in happily and making sure they get logged out

portal-api/api-tests/v1/users/login-with-sign-in-link.api-test.js
```diff
@@ -300,15 +300,15 @@ describe('POST /users/:userId/sign-in-link/:signInToken/login', () => {
             });
           });
 
-          it('returns a user blocked 403 error', async () => {
+          it('returns a user disabled 403 error', async () => {
```
ChristianMcc, line 303: I updated the API to send the correct disabled or blocked error for the user, the frontend as per initial requirements still shows the blocked page for either a disabled or blocked user

portal-api/src/v1/users/controller.test.js
```diff
@@ -48,4 +48,32 @@ describe('user controller', () => {
       expect(mockCallback).toHaveBeenCalledTimes(1);
     });
   });
+
+  describe('createPasswordToken', () => {
```
ChristianMcc, line 52: added tests to check when trying to reset password that the user isn't blocked or disabled, if they are then we don't return a token

portal-api/src/v1/users/passport.js
```diff
@@ -39,28 +49,45 @@ const baseAuthenticationConfiguration = ({ name, passport, additionalValidation,
...
-          : user && user.sessionIdentifier === jwtPayload.sessionIdentifier;
+        if (sessionIdentifierValidation(user, jwtPayload) === PASSPORT_VALIDATION_RESULTS.FAILED) {
+          console.error("Failed JWT validation for '%s' strategy", name);
+          return done(null, false);
+        }
 
-        if (validation) {
-          const additionalFields = getAdditionalReturnedFields(user);
-          return done(null, { ...sanitize(user), ...additionalFields });
+        if (additionalUserValidation && additionalUserValidation(user, jwtPayload) === PASSPORT_VALIDATION_RESULTS.FAILED) {
```
ChristianMcc, line 57: updated validation in passport so that if we provide extra user validation we provide a clear console log and don't authenticate if they are blocked or disabled once they are on the login-complete passport flow

avaitonis (dismissed):
LGTM

portal-api/src/v1/errors/user-disabled.error.js
```diff
@@ -0,0 +1,9 @@
+class UserDisabledError extends Error {
```
abhi-markan, line 1: Can we please add some documentation on the class, appreciate other classes do not have it either.
Happy to be a separate PR if required.

portal-api/src/v1/errors/user-disabled.error.test.js
```diff
@@ -0,0 +1,17 @@
+const UserDisabledError = require('./user-disabled.error');
+
+describe('UserDisabledError', () => {
+  const userId = 'exampleUserId';
```
abhi-markan, line 4: Will there ever be a scenario, where the `userId` will be empty?

portal-api/src/v1/users/controller.test.js
```diff
@@ -48,4 +48,32 @@ describe('user controller', () => {
...
+    let userService;
+
+    it('if the user is blocked or disabled then no token is returned', async () => {
+      const mockFindOne = jest.fn().mockResolvedValue(TEST_USER);
+      when(db.getCollection).calledWith('users').mockResolvedValue({ findOne: mockFindOne });
+      userService = {
+        isUserBlockedOrDisabled: jest.fn().mockImplementationOnce(() => true),
+      };
+
+      const token = await createPasswordToken('mockEmail', userService);
```
abhi-markan, line 62: `mockEmail` from a pre-defined constant please.

portal-api/src/v1/users/login.controller.js
```diff
@@ -21,12 +22,16 @@ module.exports.login = (username, password) =>
...
-      if (user.disabled) {
-        return resolve({ error: userIsDisabled });
-      }
-
-      if (user['user-status'] === STATUS.BLOCKED) {
-        return resolve({ error: userIsBlocked });
+      try {
+        userService.validateUserIsActiveAndNotDisabled(user);
+      } catch (e) {
+        console.error(e);
```
abhi-markan, line 28: Please standardise `console.error`.

portal/server/routes/middleware/validatePartialAuthToken/index.test.js
```diff
@@ -62,15 +61,32 @@ describe('validatePartialAuthToken', () => {
...
-          .mockRejectedValueOnce(new Error('test error'));
+        when(api.validatePartialAuthToken).calledWith(userToken).mockRejectedValueOnce(new Error('test error'));
       });
 
       itDoesNotCallNext();
       itDestroysTheSession();
       itRedirectsToLoginPage();
     });
+
+    describe('when portal api successfully validates the token as a login complete', () => {
```
abhi-markan, line 72: Should we add assertions where we check following 2FA specific properties values:

* signInLinkSendCount
* signInToken

portal-api/src/v1/users/user.service.js
```diff
@@ -0,0 +1,26 @@
+const { STATUS } = require('../../constants/user');
+const UserBlockedError = require('../errors/user-blocked.error');
+const UserDisabledError = require('../errors/user-disabled.error');
+
+class UserService {
```
abhi-markan, line 5: Please add some documentation.

portal-api/src/v1/users/user.service.js
```diff
@@ -0,0 +1,26 @@
...
+      throw new UserDisabledError(user._id);
+    }
+
+    if (isUserBlocked) {
+      throw new UserBlockedError(user._id);
+    }
+  }
+
+  isUserBlockedOrDisabled(user) {
+    return user['user-status'] === STATUS.BLOCKED || user.disabled === true;
```
abhi-markan, line 20: Not all the users will have `disabled` property (deprecated property) can this please be converted to optional chaining `user?.disabled`

abhi-markan: Missing function documentation.

portal-api/src/v1/users/user.service.test.js
```diff
@@ -0,0 +1,83 @@
...
+
+  describe('when the user is blocked', () => {
+    let blockedUser;
+    beforeEach(() => {
+      blockedUser = produce(testUser, (draft) => {
+        draft['user-status'] = STATUS.BLOCKED;
+      });
+    });
+
+    it('validating user throws a UserBlockedError', async () => {
```
abhi-markan, line 27: Do we need an assertion where `disabled` property does not exists and `user-status: active`?

abhi-markan (changes requested):

portal-api/src/v1/errors/user-disabled.error.test.js
ChristianMcc: Nope, we shouldn't ever throw this error as it references a user being blocked (and we need the userId to know the user)

portal-api/src/v1/users/controller.test.js
abhi-markan: Test name makes sense but any textual explanation is always helpful when being visited in fresh.

portal-api/src/v1/users/login.controller.js
```diff
@@ -21,12 +22,16 @@ module.exports.login = (username, password) =>
...
+        if (e instanceof UserBlockedError) {
+          return resolve({ error: userIsBlocked });
+        }
+        if (e instanceof UserDisabledError) {
+          return resolve({ error: userIsDisabled });
```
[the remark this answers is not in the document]
abhi-markan: Makes sense. Thanks

portal-api/src/v1/users/user.service.js
abhi-markan: Thanks

abhi-markan (approved):

abhi-markan (approved):

PR #2603, opened 2024-01-22
feat(FN-2074): remove dividers when no buttons on 'Bank Reports' page

toddym42:
## Introduction :pencil2:
Currently on the TFM 'Bank Reports' page there are dividing lines between the reporting period summaries. When there are no buttons (i.e. when viewing as a `PDC_READ` user) this can look like there is an extra empty row at the bottom of the table

Only render the divider line when there are buttons at the bottom of each table (i.e. when viewing as a `PDC_RECONCILE` user)


trade-finance-manager-ui/styles/layout.scss
```diff
@@ -58,13 +58,15 @@
...
+      border-bottom: 1px solid $govuk-border-colour;
+    }
   }
 
   @for $spacing-unit from 1 through 9 {
-    &.ukef-separator-spacing-#{$spacing-unit} {
+    &.ukef-stack-spacing-#{$spacing-unit} {
       > *:not(:first-child) {
         @include govuk-responsive-padding($spacing-unit, 'top');
       }
```
toddym42, line 72: "stack" seems to be a more common name for a component that adds spacing and dividers between its children (e.g. [MUI Stack](https://mui.com/material-ui/react-stack/) and [Bootstrap Stack](https://react-bootstrap.netlify.app/docs/layout/stack/))

trade-finance-manager-ui/templates/utilisation-reports/utilisation-reports.njk
```diff
@@ -16,7 +16,18 @@
...
+        govuk-grid-column-full
+        ukef-stack
+        {% if user | userIsInTeam(["PDC_RECONCILE"]) %}
+          ukef-stack-divider
+          ukef-stack-spacing-7
+        {% else %}
+          ukef-stack-spacing-2
+        {% endif %}
+      "
+    >
```
toddym42, line 30: PDC_RECONCILE:

Buttons

PDC_READ:

No_Buttons

PR #2604, opened 2024-01-22
fix(FN-2075): highlight 'Users' nav item on user admin pages in Portal

toddym42:
## Introduction :pencil2:
Currently when on any of the user admin pages in Portal the "Users" nav item is not highlighted:

image

* add `primaryNav: 'users'` to all user admin pages so that the "Users" nav item is highlighted
* add a `PRIMARY_NAV_KEY` constant to use instead of hard-coded strings

### DIFF
--- a/portal/server/routes/admin/users.js
+++ b/portal/server/routes/admin/users.js
@@ -7,7 +7,7 @@ const {
   generateErrorSummary,
   constructPayload,
 } = require('../../helpers');
-const { ALL_BANKS_ID } = require('../../constants');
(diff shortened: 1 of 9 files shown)


portal/server/routes/admin/users.js
toddym42, line 1: Updates here are where the new functionality is added (highlighting the "Users" nav item on user admin pages). Everything else is refactoring.

francescastocco (approved):
Looks good to me :)

PR #2605, opened 2024-01-22
feat(FN-2077): change how `PDC` users behave on login

JoshBinns2000:
## Tickets 🎫 
[FN-2070](https://ukef-dtfs.atlassian.net/browse/FN-2070), [FN-2076](https://ukef-dtfs.atlassian.net/browse/FN-2076), [FN-2077](https://ukef-dtfs.atlassian.net/browse/FN-2077)

## Introduction :pencil2:
We want users who are _only_ in the `PDC_READ` or `PDC_RECONCILE` teams to not be able to see the 'All deals' and 'All facilities' navigation items.

For users in multiple teams (where one of the teams is `PDC_READ` or `PDC_RECONCILE`) all three navigation items should be visible, with 'Bank reports' being on the far right of the available items and the user landing on the 'All deals' page.

Makes the changes and adds tests.

Upgrades the `home` controller to typescript.


trade-finance-manager-ui/templates/_partials/primary-navigation.njk
```diff
@@ -29,10 +29,14 @@
...
-{% elif user %}
-  {% set navItems = [allDealsNavItem, allFacilitiesNavItem] %}
+{% if user %}
+  {% if user | userIsOnlyInTeam(['PDC_READ']) or user | userIsOnlyInTeam(['PDC_RECONCILE']) %}
+    {% set navItems = [bankReportsNavItem] %}
+  {% elif user | userIsInTeam(['PDC_READ', 'PDC_RECONCILE']) %}
+    {% set navItems = [allDealsNavItem, allFacilitiesNavItem, bankReportsNavItem] %}
+  {% else %}
+    {% set navItems = [allDealsNavItem, allFacilitiesNavItem] %}
+  {% endif %}
```
JoshBinns2000, line 39: I'm not a huge fan of all this nesting as it is fairly complicated, but I can't think of a good way of shortening this in pure nunjucks. Any suggestions are appreciated

toddym42: I think this is fine - it's not nesting too deeply, and our helper functions make it quite readable.

What would happen in the case where the user has both the `PDC_READ` and `PDC_RECONCILE` permissions but nothing else? I think this would fall through to the second case where they view everything again? I don't think we would expect this but should probably account for it.

francescastocco: I think although it's quite long it wasn't too complex to understand

francescastocco (dismissed):
Looks good :)

trade-finance-manager-ui/server/controllers/home/index.test.ts
```diff
@@ -0,0 +1,63 @@
...
+  it("should redirect to login ('/') if the user is not logged in", () => {
+    // Arrange
+    const { req, res } = httpMocks.createMocks({ session: {} });
+
+    // Act
+    homeController.getUserHomepage(req, res);
+
+    // Assert
+    // eslint-disable-next-line no-underscore-dangle
+    expect(res._getRedirectUrl()).toEqual('/');
```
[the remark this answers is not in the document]
JoshBinns2000: Sounds good - worth adding as part of the big prettier PR maybe... 👀

trade-finance-manager-ui/templates/_partials/primary-navigation.njk
JoshBinns2000: @toddym42 That's a good point. Just to capture that edge case, I added an extra
```

toddym42 (approved):

PR #2606, opened 2024-01-23
feat(FN-2081): add "visible in TFM" flag to banks

JoshBinns2000:
## Introduction :pencil2:
We want to add a flag to banks which signals whether or not they are viewable in `TFM` (ie. whether or not they submit utilisation reports via `portal`).

Adds the `isUtilisationReportVisibleInTfm` flag to `Banks` in `dtfs-central-api` and `mock-data-loader`. This flag is not optional as we want every banks "opt-in" status to be explicit. We also add this flag to the `portal-api` payload validation as `mock-data-loader` sends the bank payload through this api.

Corrects the `CustomExpressRequest` type helper and fixes some swagger docs issues.

### DIFF
--- a/dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-reports-reconciliation-summary.controller/helpers.ts
+++ b/dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-reports-reconciliation-summary.controller/helpers.ts
@@ -159,9 +159,10 @@ export const generateReconciliationSummaries = async (currentSubmissionMonth: Is
   // TODO FN-1949 - when banks' reporting schedules are added to the `bank` collection, add new repo method to only
   //  fetch banks that are due to submit this month
   const banks = await getAllBanks();
+  const banksVisibleInTfm = banks.filter((bank) => bank.isUtilisationReportVisibleInTfm);
 
-  const allReportsForCurrentSubmissionMonth = await getAllReportsForSubmissionMonth(banks, currentSubmissionMonth);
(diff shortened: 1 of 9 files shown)


dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-reports-reconciliation-summary.controller/helpers.ts
```diff
@@ -159,9 +159,10 @@ export const generateReconciliationSummaries = async (currentSubmissionMonth: Is
   // TODO FN-1949 - when banks' reporting schedules are added to the `bank` collection, add new repo method to only
   //  fetch banks that are due to submit this month
   const banks = await getAllBanks();
+  const banksVisibleInTfm = banks.filter((bank) => bank.isUtilisationReportVisibleInTfm);
```
francescastocco, line 162: If `bank.isUtilisationReportVisibleInTfm` is not defined does it just default to false?

francescastocco (commented):
Looks good, I think `isVisibleInTfmUtilisationReports` makes a bit more sense for naming though

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-reports-reconciliation-summary.controller/helpers.ts
JoshBinns2000: It is expected to always be defined...*but*, if it isn't defined, this will treat it the same as `false`

francescastocco (approved):

PR #2607, opened 2024-01-23
fix(DTFS2-7022): restrict sign in link expired path

AlexBramhill:
## Introduction :pencil2:
The `sign-in-link-expired` path in `portal` was not restricted based on a user having a present 2fa token. The user could not resend a sign in link, so this is not an issue. However, it should be fixed

Added token restriction to endpoint


AlexBramhill:
New PR: https://github.com/UK-Export-Finance/dtfs2/pull/2750

PR #2656, opened 2024-01-24
chore(DTFS2-6993): replace moment.js in portal

MarRobSoftwire:
## Introduction :pencil2:
moment.js is a deprecated package - we should remove it from portal.
Tests are changed by #2690, to make sure behaviour is preserved. 

* Replaced any `moment()` with, where possible, a vanilla `new Date()`. If needed used functions from `date-fns` to modify/compare dates
* Use `date-fns-tz` to handle timezones
* Updated date formatting strings in nunjucks files for `date-fns` system (see note below)

## Note ⚠️ 
* `date-fns-tz` is incompatible with `date-fns` v3 so I've restricted to use `v2.30.0` . This is the same as used in #2595 and shouldn't cause any conflicts
* `date-fns` uses Unicode tokens to format dates, but Moment.js uses its own system. The most important difference is between `DD`, `YYYY` and `dd`, `yyyy`. See [here ](https://github.com/date-fns/date-fns/blob/main/docs/unicodeTokens.md)for details


avaitonis (dismissed):
LGTM

portal/server/routes/contract/bond/index.js
```diff
@@ -468,23 +467,15 @@ router.post('/contract/:_id/bond/:bondId/confirm-requested-cover-start-date', as
...
     } else {
-      const previousCoverStartDate = moment().set({
+      const previousCoverStartDate = set(new Date(), {
         date: Number(bondToRender['requestedCoverStartDate-day']),
         month: Number(bondToRender['requestedCoverStartDate-month']) - 1, // months are zero indexed
         year: Number(bondToRender['requestedCoverStartDate-year']),
       });
 
-      const previousCoverStartDateTimestamp = moment(previousCoverStartDate).utc().valueOf().toString();
+      const previousCoverStartDateTimestamp = previousCoverStartDate.valueOf();
```
MarRobSoftwire, line 476: Should be string

portal/server/routes/contract/bond/index.js
```diff
@@ -468,23 +467,15 @@ router.post('/contract/:_id/bond/:bondId/confirm-requested-cover-start-date', as
...
-      const now = moment();
-
-      const dateOfCoverChange = moment().set({
-        date: Number(moment(now).format('DD')),
-        month: Number(moment(now).format('MM')) - 1, // months are zero indexed
-        year: Number(moment(now).format('YYYY')),
-      });
-
-      const dateOfCoverChangeTimestamp = moment(dateOfCoverChange).utc().valueOf().toString();
+      const today = new Date();
```
abhi-markan, line 478: This is where we can use a helper function which can return epoch or not based on my comments on the other related PR.

```javascript
const getNow = (epoch = false) => epoch ? new Date().valueOf() : new Date();
```

portal/server/nunjucks-configuration/filter-localiseTimestamp.ts
```diff
@@ -0,0 +1,20 @@
+import { formatInTimeZone } from 'date-fns-tz';
+
+export const filterLocaliseTimestamp = (
```
abhi-markan, line 3: Missing documentation, please add.

abhi-markan: Missing unit test cases, please add.

portal/server/nunjucks-configuration/filter-localiseTimestamp.ts
```diff
@@ -0,0 +1,20 @@
...
+  utcTimestamp: number | string,
```
abhi-markan, line 4: Since this function is in TS now, might be ideal to define a `type` `utcTimestamp`?

portal/server/nunjucks-configuration/filter-localiseTimestamp.ts
```diff
@@ -0,0 +1,20 @@
...
+  format: string,
+  targetTimezone: string,
+) => {
+  if (!utcTimestamp) {
```
abhi-markan, line 8: Will this ever be a case?

portal/server/nunjucks-configuration/filter-localiseTimestamp.ts
```diff
@@ -0,0 +1,20 @@
...
+    return '';
+  }
+  const timestamp = parseInt(String(utcTimestamp), 10);
```
abhi-markan, line 11: If this is getting converted to `String` should we limit `utcTimestamp` type to be `String` data type only?

abhi-markan (changes requested):

portal/server/routes/contract/bond/index.js
ChristianMcc: FWIW I'd much prefer two functions, one of getNow and one of getNowAsEpoch, I think calling this function and having a boolean to swap between the two return types is probably a little confusing.

In terms of the bigger question of if this should be abstracted to a helper function, what's the use for this specific helper where it would just be 
`const getNow = () => new Date()`
I personally prefer not abstracting away what we're starting here and using variable naming to show what were doing with .valueOf or not, e.g. 
const today = new Date();
or
const todayAsEpoch = newDate().valueOf();

portal/server/nunjucks-configuration/filter-localiseTimestamp.ts
MarRobSoftwire: IMO this would make the code less readable - I think `number | string` is simple enough to be used inline

portal/server/nunjucks-configuration/filter-localiseTimestamp.ts
MarRobSoftwire: It's used in `templates/contract/components/contact-overview-table.njk`, line 57 
```
{{ details.manualInclusionApplicationSubmissionDate | localiseTimestamp("dd/MM/yyyy", user.timezone) | dashIfEmpty | safe }}
```
The timestamp can be an empty string passed into `dashIfEmpty`

portal/server/helpers/date.ts
```diff
@@ -0,0 +1 @@
+export const getNowAsEpoch = () => (Date.now());
```
abhi-markan, line 1: Docstring please.

portal/server/nunjucks-configuration/filter-localiseTimestamp.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  if (Number.isNaN(timestamp)) {
+    return 'Invalid date';
+  }
+
+  const date = new Date(timestamp);
+
+  return formatInTimeZone(date, targetTimezone, format);
```
abhi-markan, line 29: Unsafe return of an `any` type value here.

MarRobSoftwire: My IDE thinks this is ok - could you clarify? Thanks

abhi-markan (approved):
LGTM

PR #2657, opened 2024-01-24
fix(durable-functions): Error reporting for `acbs-functions`

abhi-markan:
## Introduction :pencil2:
Error object is not being sent when an exception is thrown.

* Added `error` object when new error is thrown.

* Added `400` status code to facility guarantee record creation as an acceptable status code.
* Improved documentation.


avaitonis (approved):
LGTM

PR #2684, opened 2024-01-25
feat(FN-1855): add TypeORM integration

toddym42:
## Introduction :pencil2:
* adds new docker-compose configuration to spin up a SQL Server instance for local development
* adds ground work for [TypeORM](https://typeorm.io/) integration, used to interface with the SQL Server database

See new [sql-db.md](https://github.com/UK-Export-Finance/dtfs2/pull/2684/files#diff-c629e3a3bb1a8ff696609a6a351e3eb0c729424cb5e4b7ab14fd9000a4cd3bc4) doc for further implementation details.


libs/common/src/sql-db/entity/utilisation-report.ts
```diff
@@ -0,0 +1,29 @@
+import { Column, Entity, PrimaryGeneratedColumn } from 'typeorm';
+import { UtilisationReportReconciliationStatus } from '../../types';
+import { ReportPeriod } from './partials';
+
+@Entity('UtilisationReport')
```
toddym42, line 5: If we don't enter a value here (i.e.  just `@Entity()` instead if `@Entity('UtilisationReport')`) then the table name will be based off of the class name. I wanted to add `...Entity` as a suffix to the class name though so it's clear what this is when imported into other packages.

This casing of the table name (UpperCamelCase), and the fact that it's singular `UtilisationReport` instead of `UtilisationReports` is based off of the conventions that Microsoft follow themselves (see examples [here](https://github.com/microsoft/sql-server-samples)). This doesn't seem like a required standard though, but instead more of a preference. If we hadn't specified the table name explicitly then TypeORM would have actually named the table with snake_case `utilisation_report_entity`.

I note that Microsoft also uses UpperCamelCase for property names too, but I haven't overridden those (just because I wasn't forced to). Maybe we want to be consistent. I'm guessing use of UpperCamelCase by Microsoft is only used to be consistent with C# conventions, so maybe we don't need to follow that here.

What do we think?

libs/common/src/sql-db/init.sh
toddym42, line 1: If in the docker-compose we had instead just used
```
image: mcr.microsoft.com/mssql/server
```
instead of referring to our own new [Dockerfile](https://github.com/UK-Export-Finance/dtfs2/pull/2684/files#diff-35f76357d0b521bcbab7b7d44cb821ab9a942376a81bcbf1c402fa29ffb1e640) with
```
build: ./libs/common/src/sql-db
```
then we wouldn't have had to add a Dockerfile at all, or this new `init.sh` script to manually spin up our database.

In that case we would use the default [master](https://learn.microsoft.com/en-gb/sql/relational-databases/databases/master-database) database for local development, however, this isn't meant to be used for our own data. This is instead is a system database that is meant to be used to store system level information (details of other databases, login accounts, etc.)

Using the `master` database can be fine if you're just prototyping things quickly in local development, but we run into complications when we want to do things like drop all the tables and re-run migrations (see the `db:reset` script added to the common package.json), as this will also remove all the system tables too.

For this reason it was worth adding the extra config required to add our own database, and this is also more reflective of what will be done for real on deployed environments.

libs/common/src/sql-db/entity/utilisation-report.ts
```diff
@@ -0,0 +1,29 @@
...
+  @Column()
+  bankId!: string;
+
+  @Column(() => ReportPeriod)
+  reportPeriod!: ReportPeriod;
+
+  // TODO FN-1859 - add AzureFileInfoEntity
+  // @OneToOne(() => AzureFileInfoEntity)
+  // @JoinColumn()
+  // readonly azureFileInfo: AzureFileInfoEntity
```
JoshBinns2000, line 19: Does this need to be its own entity or can it just be another embedded entity?

libs/common/src/sql-db/init.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+set -m
+
+/opt/mssql/bin/sqlservr &
+
+# Wait for SQL Server to start
+sleep 30s
```
JoshBinns2000, line 7: Is there not a better way to wait for the server to start? 😅

libs/common/src/sql-db/entity/utilisation-report.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  @Column()
+  status!: UtilisationReportReconciliationStatus;
+
+  @Column({ type: 'datetime', nullable: true })
+  dateUploaded!: Date | null;
```
JoshBinns2000, line 25: Why is this column nullable?

libs/common/src/sql-db/init.sh
toddym42: Yeah, good call. Updated.

libs/common/src/sql-db/entity/utilisation-report.ts
JoshBinns2000: True - I forgot we were adding that in

libs/common/src/sql-db/init.sh
JoshBinns2000, line 1: What is the behaviour of this if `execute_sql_command()` never fully executes without throwing an error (ie. what is the break condition for the while loop if something quietly goes wrong)?

toddym42: I don't really have any control over whether `sqlcmd` returns or not, but as far as I know it should always return a status (either success or fail). If it didn't then this would just hang, regardless if there is a while loop or not. The app won't start up and it'll be very obviously broken.

In any case I've added a bit more logging so we can see progress and added a time limit to the while loop. Let me know what you think.

JoshBinns2000: I'm happy with the time limit and logging - I was expecting something like that :)

libs/common/src/sql-db/migrations/1706118619447-AddUtilisationReportTable.ts
```diff
@@ -0,0 +1,28 @@
...
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`
+      CREATE TABLE "UtilisationReport" (
+        "id" int NOT NULL IDENTITY(1, 1),
+        "bankId" nvarchar(255) NOT NULL,
+        "reportPeriodStartMonth" int NOT NULL,
+        "reportPeriodStartYear" int NOT NULL,
+        "reportPeriodEndMonth" int NOT NULL,
+        "reportPeriodEndYear" int NOT NULL,
+        "status" nvarchar(255) NOT NULL,
```
ChristianMcc, line 15: postgres has enums but looks like MSSQL might not, sad :(

docker-compose.dev.yml
```diff
@@ -6,6 +6,18 @@ services:
     ports:
       - '3310:3310'
 
+  dtfs-submissions-sql:
+    build: ./libs/common/src/sql-db
+    restart: always
+    environment:
+      - ACCEPT_EULA=Y
+      - MSSQL_SA_PASSWORD=AbC!2345
+      - DB_NAME=DTFSSubmissions
```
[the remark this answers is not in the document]
toddym42: Makes sense to me. Similarly I'll change the service name from `dtfs-submissions-sql` to just `dtfs-sql`.

docker-compose.dev.yml
```diff
@@ -46,6 +59,12 @@ services:
       - EXTERNAL_API_KEY
       - NODE_ENV=${NODE_ENV}
       - RATE_LIMIT_THRESHOLD
+      - SQL_DB_HOST
+      - SQL_DB_PORT
+      - SQL_DB_USERNAME
+      - SQL_DB_PASSWORD
+      - SQL_DB_NAME
+      - SQL_DB_LOGGING
```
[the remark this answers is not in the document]
toddym42: Agreed, that's much clearer 👍

docs/sql-db.md
```diff
@@ -0,0 +1,93 @@
...
+```
+
+TypeORM automatically tracks which migrations have already been run in a `migrations` table in the database.
+
+⚠️ When running this command, it **WON'T** just apply any that have not already been run, but instead find the last one run in the `migrations` table and run any ordered after that one. For this reason we must take care to ensure that any new migrations are added in the correct order when merging git branches.
+
+#### - Reverting migrations
+
+```shell
+npm run db:migrate:down
```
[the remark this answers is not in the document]
toddym42: Thought I'd added a note here, whoops! It looks like TypeORM will revert one at a time, so you just need to run multiple times ([docs](https://typeorm.io/migrations#running-and-reverting-migrations))

libs/common/src/sql-db/migrations/1706118619447-AddUtilisationReportTable.ts
toddym42: Yeah, unfortunately not 😢

libs/common/src/sql-db-entity/utilisation-report.ts
```diff
@@ -0,0 +1,57 @@
...
+   */
+  @Column({ type: 'varchar', nullable: true })
+  uploadedByUserId!: string | null;
+
+  /**
+   * Breakdown of utilisation per facility and currency combination
+   * TODO FN-1855 - should this be called `payments`, `feeRecords`, or something else?
+   */
+  @OneToMany(() => UtilisationDataEntity, (data) => data.report)
+  data!: UtilisationDataEntity[];
```
JoshBinns2000, line 56: I'm not sure exactly what behaviour we want here, but I think we should explicitly define some `cascade` behaviour

libs/common/src/sql-db-entity/azure-file-info.ts
```diff
@@ -0,0 +1,50 @@
...
+   * MIME type - indicating the nature and format of the document
+   * @example 'text/csv'
+   */
+  @Column()
+  mimetype!: string;
+
+  /**
+   * The associated utilisation report
+   */
+  @OneToOne(() => UtilisationReportEntity, (report) => report.azureFileInfo)
```
JoshBinns2000, line 47: Do we want this relation to be bi-directional?

ChristianMcc: FWIW I don't think so. Also are bi-directional relationships supported in TFM for one to one? Can imagine they are just only seen them for many to one/many where a join table exists

libs/common/src/sql-db-entity/utilisation-report.ts
```diff
@@ -0,0 +1,57 @@
+import { Column, Entity, OneToMany, OneToOne, PrimaryGeneratedColumn } from 'typeorm';
+import { UtilisationReportReconciliationStatus } from '../types';
+import { ReportPeriodPartial } from './partials';
+import { AzureFileInfoEntity } from './azure-file-info';
+import { UtilisationDataEntity } from './utilisation-data';
+
+@Entity('UtilisationReport')
+export class UtilisationReportEntity {
```
ChristianMcc, line 8: Do we want a @createdAt and @updatedAt column on this as well? The latter would be especially helpful for audit down the line and could set the precedent for them. CreatedAt might be good to audit when the job added it etc. (and date uploaded being the business logic side!)

libs/common/src/sql-db-connection/migrations/1707306956373-AddUtilisationReportTables.ts
```diff
@@ -0,0 +1,95 @@
...
+        "filename" nvarchar(255) NOT NULL,
+        "fullPath" nvarchar(255) NOT NULL,
+        "url" nvarchar(255) NOT NULL,
+        "mimetype" nvarchar(255) NOT NULL,
+        "utilisationReportId" int NOT NULL,
+        CONSTRAINT "PK_ea6ad35b38f9b9e248b432d1a1e" PRIMARY KEY ("id")
+      )
+    `);
+
+    await queryRunner.query(`
```
ChristianMcc, line 20: I should remember this - is this just typeorm creating an index by default because you have a foreign key here? Semi surprised it's added 'where utilisationReportId IS NOT NULL' considering it's a non-nullable foreign key

libs/common/src/sql-db-connection/migrations/1707306956373-AddUtilisationReportTables.ts
```diff
@@ -0,0 +1,95 @@
...
+      CREATE TABLE "UtilisationData" (
+        "id" int NOT NULL IDENTITY(1, 1),
+        "facilityId" nvarchar(255) NOT NULL,
+        "exporter" nvarchar(255) NOT NULL,
+        "baseCurrency" nvarchar(255) NOT NULL,
+        "facilityUtilisation" int NOT NULL,
+        "totalFeesAccruedForTheMonth" int NOT NULL,
+        "totalFeesAccruedForTheMonthCurrency" nvarchar(255) NOT NULL,
```
ChristianMcc, line 33: I'm happy with all of these being not null but will need to make sure in the refactor everything has a correct default value (not quite like that on mongodb right now, think if currency is empty it doesn't default to baseCurrency.
Somewhere we'll need to have it documented all these kind of edge case/defaulting values that we'll get in the DB.
Might be worth us chatting to Rich and Andy about this when we talk about schema of saying hey it's easier for us if we default if it's not necessary

libs/common/src/sql-db-entity/utilisation-report.ts
toddym42: I've added a new abstract `AuditableEntity` class that all our entities can extend.

libs/common/src/sql-db-connection/migrations/1707306956373-AddUtilisationReportTables.ts
ChristianMcc: business logic it is!

ChristianMcc (approved):
Nice one :)

PR #2687, opened 2024-01-25
chore(e2e-tests): fix lint errors

JoshBinns2000:
## Introduction :pencil2:
There are some lint errors that appear in `e2e-tests` due to the `cypress` `unsafe-to-chain-command` rule.

By wrapping the rule-breaking functions in `cy.wrap`, we ensure that a chainable command is returned and therefore this rule is no longer broken.


e2e-tests/tfm/cypress/fixtures/constants.js
```diff
@@ -43,6 +43,34 @@ const DISPLAY_USER_TEAMS = {
   PDC_READ: 'PDC read',
 };
 
+/**
```
ChristianMcc, line 46: This is a very nice improvement! Is it worth moving this more to an E2E/cypress readme and just leaving a comment here saying 'see readme for alias keys' etc.
This feels like something useful for someone to read through as an approach on how to use cypress that's a bit hidden away here

ChristianMcc (commented):
Really nice changes here! Just left a small comment on how we can document to make sure this can be used in future

e2e-tests/tfm/cypress/fixtures/constants.js
JoshBinns2000: I've updated the `README.md` - if you could have a look that would be great :)

e2e-tests/README.md
```diff
@@ -70,4 +70,49 @@ Each test suite shares a similar directory structure:
...
+
+// Accessing the result of that command (in a test)
+cy.insertAndYieldLength(itemToInsert);
+cy.get(`@${ALIAS_KEY.INSERT_AND_YIELD_LENGTH}`)
+  .then((length) => {
+    // do something with length
+  });
+```
+
+There is also have a helper function called [`aliasSelector`](./support/alias-selector.js) which abstracts the logic of putting the `@` before the alias key.
```
toddym42, line 116: Should remove `have` here.

toddym42 (approved):
Great work. My OCD thanks you for removing all the lint warnings 🙏

toddym42 (approved):

PR #2688, opened 2024-01-25
fix(deployments): remove QAT test run prior to deployment

ChristianMcc:
## Introduction :pencil2:
Due to failing tests running only on dev/staging branch that need to be investigated remove QAT step prior to deployment


abhi-markan:
Closing as the issue seems to be with a flaky test when being deployed to `dev`, however `staging` works fine.

PR #2695, opened 2024-01-29
feat(FN-1947): add scheduler to create utilisation reports

JoshBinns2000:
## Introduction :pencil2:
We want to add a job to create utilisation reports in the `REPORT_NOT_RECEIVED` state at 2 am each day to make the logic around the state of utilisation reports in the database easier to write.

Adds the scheduler to `dtfs-central-api` alongside unit tests. Also add some new repo methods and date helpers.


dtfs-central-api/src/types/db-models/utilisation-reports.ts
```diff
@@ -37,6 +37,7 @@ export type UtilisationReport = Prettify<
     /**
      * Details of the user that uploaded the report
      */
-    uploadedBy: UploadedByUserDetails;
+    uploadedBy?: UploadedByUserDetails;
+    // TODO FN-1947 is there a nice way to narrow this type? Tried a lot so far which breaks other functions
```
JoshBinns2000, line 41: I spent quite a bit of time trying to narrow this type so that `uploadedBy` is only `undefined` if the status is `REPORT_NOT_RECEIVED`. I did manage to do it, but it then starts to throw typescript errors with function return types (specifically in the repo files, but I imagine it would be angry elsewhere as well). I think this happens because excessively narrowing a type works quite well when constructing variables to satisfy the type, but trying to get type inference from other places starts to become quite hard for the compiler (`WithoutId<WithId<UtilisationReport>>` no longer exactly matches `UtilisationReport` according to the compiler, for instance). In this case, as much as I would like to have the narrow type, it is easier to just make `uploadedBy` optional. Any thoughts?

dtfs-central-api/src/utils/report-period.ts
```diff
@@ -27,3 +29,45 @@ export const getPreviousReportPeriodStart = ({ month, year }: MonthAndYear): Mon
 
 export const isEqualReportPeriodStart = (reportPeriodStart1: MonthAndYear, reportPeriodStart2: MonthAndYear): boolean =>
   reportPeriodStart1.year === reportPeriodStart2.year && reportPeriodStart1.month === reportPeriodStart2.month;
+
+export const getCurrentReportPeriodForBankSchedule = (bankReportPeriodSchedule: ReportPeriodSchedule[]): ReportPeriod => {
```
JoshBinns2000, line 33: The logic in this function is pretty complicated. I've tried to name the variables such that it is easier to figure out what is going on, but I'm keen to have any suggestions on how to make the `reportPeriodInPreviousYear` logic easier to digest

dtfs-central-api/src/scheduler/jobs/create-utilisation-reports/index.ts
```diff
@@ -0,0 +1,53 @@
...
+    .filter((_, index) => isMissingBankReport[index])
+    .map((bank) => ({
+      id: bank.id,
+      name: bank.name,
+      reportPeriodSchedule: bank.utilisationReportPeriodSchedule,
+    }));
+};
+
+const createUtilisationReportForBanks = async (): Promise<void> => {
+  const bankIdsWithoutReports = await getBanksWithoutReports();
```
francescastocco, line 30: Should this be 
```suggestion
  const banksWithoutReports = await getBanksWithoutReports();
```

.env.sample
```diff
@@ -44,6 +44,7 @@ UTILISATION_REPORT_OVERDUE_CHASER_DATE_BUSINESS_DAYS_FROM_START_OF_MONTH=15
 UTILISATION_REPORT_SUBMISSION_PERIOD_START_EMAIL_SCHEDULE='0 8 1 * *'
 UTILISATION_REPORT_DUE_EMAIL_SCHEDULE='0 8 * * *'
 UTILISATION_REPORT_OVERDUE_EMAIL_SCHEDULE='0 8 * * *'
+CREATE_UTILISATION_REPORT_FOR_BANKS_SCHEDULE='0 14 * * *'
```
JoshBinns2000, line 47: Need to change this to `0 2 * * *` (2 am)

dtfs-central-api/src/services/repositories/utilisation-reports-repo/utilisation-reports-repo.test.ts
```diff
@@ -91,6 +92,59 @@ describe('utilisation-reports-repo', () => {
...
+          },
+        },
+        dateUploaded: expect.any(Date) as Date,
+        azureFileInfo: null,
+        status: UTILISATION_REPORT_RECONCILIATION_STATUS.REPORT_NOT_RECEIVED,
+        uploadedBy: {
+          id: '0',
+          firstname: 'SUPER',
+          surname: 'USER',
+        },
```
JoshBinns2000, line 144: Need to remove this

dtfs-central-api/src/utils/report-period.ts
```diff
@@ -27,3 +29,45 @@ export const getPreviousReportPeriodStart = ({ month, year }: MonthAndYear): Mon
...
+    return {
+      start: {
+        month: reportPeriodInPreviousYear.startMonth,
+        year: currentYear - 1,
+      },
+      end: {
+        month: reportPeriodInPreviousYear.endMonth,
+        year: currentYear,
+      },
+    };
```
francescastocco, line 51: Following my comment about naming, is this logic correct?

francescastocco (commented):
Looks good so far, I haven't had chance to review everything yet

dtfs-central-api/src/scheduler/index.ts
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * 'node-cron' requires the task function to return `void`, but
+ * some of the scheduler functions return `Promise<void>`. This
+ * function takes in any task and uses an IIFE to turn it into
+ * a function with return type `void`.
+ */
+const promiseToVoid = (task: ReturnType<SchedulerJob['init']>['task']) => (now: TaskInput) => {
+  (async () => {
+    await task(now);
+  })();
```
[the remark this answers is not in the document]
JoshBinns2000: It doesn't change any functionality - this just acts as a wrapper for functions which either return `void` or `Promise<void>` and executes them such that the return type is always `void`

dtfs-central-api/src/scheduler/jobs/create-utilisation-reports/index.ts
JoshBinns2000: Yes - this was from an old version 🙃

dtfs-central-api/src/utils/report-period.ts
JoshBinns2000: Not for that specific edge case, no. I will add a new test to cover it

dtfs-central-api/src/utils/report-period.ts
toddym42: `reportPeriodInPreviousYear` implies this could be when both start and end are in the previous year (e.g. 10-12), but I don't think that's what this is used for. I think this is the case when the schedule overlaps the year boundary?

I'm not sure about using `currentYear` as that implies today's year, but this could be last year.

I think we can simplify this a bit by first finding the current relevant schedule, the returning the result based on that. I'm not sure about the naming, but it would be something like this:
```
export const getCurrentReportPeriodForBankSchedule = (bankReportPeriodSchedule: ReportPeriodSchedule[]): ReportPeriod => {
  const previousMonthDate = subMonths(new Date(), 1);
  const targetMonth = getOneIndexedMonth(previousMonthDate);
  const targetYear = previousMonthDate.getFullYear();

  const currentSchedule = bankReportPeriodSchedule.find((schedule) => {
    const isYearOverlapping = schedule.startMonth > schedule.endMonth && (targetMonth >= schedule.startMonth || targetMonth <= schedule.endMonth);
    const isNonYearOverlapping = schedule.startMonth <= targetMonth && targetMonth <= schedule.endMonth;
    return isYearOverlapping || isNonYearOverlapping;
  });

  if (!currentSchedule) {
    throw new Error(`Failed to get current report period start for bank with report schedule: ${currentSchedule}`);
  }

  const yearOverlapping = currentSchedule.startMonth > currentSchedule.endMonth;

  return {
    start: {
      month: currentSchedule.startMonth,
      year: yearOverlapping ? targetYear - 1 : targetYear,
    },
    end: {
      month: currentSchedule.endMonth,
      year: targetYear,
    },
  };
};
```

dtfs-central-api/src/scheduler/jobs/create-utilisation-reports/index.test.ts
```diff
@@ -0,0 +1,113 @@
...
+  beforeEach(() => {
+    process.env = { ...originalProcessEnv };
+    process.env[scheduleEnvVariableName] = testSchedule;
+  });
+
+  afterAll(() => {
+    process.env = { ...originalProcessEnv };
+  });
+
+  it('throws an error when the environment variable is undefined', () => {
```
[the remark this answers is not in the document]
JoshBinns2000: Seeing as I store the variable name in a variable, I may as well use that specifically here

dtfs-central-api/src/types/db-models/banks.ts
```diff
@@ -1,6 +1,6 @@
 import { WithId } from 'mongodb';
 
-type ReportPeriodSchedule = {
+export type ReportPeriodSchedule = {
```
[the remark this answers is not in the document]
JoshBinns2000: Decided instead to create a new `export type BankReportPeriodSchedule = ReportPeriodSchedule[]` instead

dtfs-central-api/src/utils/report-period.ts
JoshBinns2000: As discussed, I've changed this implementation as we also need to worry about whether or not the report period goes into the future. I'll resolve this conversation, but note that it needs to be looked at again :)

dtfs-central-api/src/scheduler/jobs/create-utilisation-reports/index.test.ts
```diff
@@ -0,0 +1,113 @@
...
+  const testSchedule = '* * * * *';
+
+  let originalProcessEnv: NodeJS.ProcessEnv;
+  beforeAll(() => {
+    originalProcessEnv = process.env;
+  });
+
+  beforeEach(() => {
+    process.env = { ...originalProcessEnv };
+    process.env[scheduleEnvVariableName] = testSchedule;
```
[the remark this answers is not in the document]
JoshBinns2000: Now that I use the name in one of the tests, I've decided to keep the variable, but I don't feel strongly about keeping it in so let me know what you think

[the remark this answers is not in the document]
toddym42: Yeah, fair enough. I think the eventual long term ideal solution would be to not access `process.env` directly, but instead access from a validated type safe config file (like [this](https://github.com/UK-Export-Finance/dtfs2/pull/2684/files#diff-a2e738d59efd83fd545df039217eaa4a188b7a8ace66abdabb14626b6c0a5ac6) example), but we don't have that yet so I'm happy with this.

dtfs-central-api/src/scheduler/jobs/create-utilisation-reports/index.test.ts
```diff
@@ -0,0 +1,148 @@
...
+
+    it('only tries to create reports for banks which do not have a report for the current report period', async () => {
+      // Arrange
+      const job = getJob();
+      jest.mocked(getAllBanks).mockResolvedValue(banks);
+      const getCurrentUtilisationReportByBankIdAndReportPeriodSpy = jest
+        .mocked(utilisationReportsRepo.getCurrentUtilisationReportByBankIdAndReportPeriod)
+        .mockImplementation((bankId: string) => {
+          switch (bankId) {
+            case '1':
```
toddym42, line 130: Instead of hard-coding this here I would add the following above:
```
const bankWithoutReport = banks[0]
```
then you can do
```
case bankWithoutReport.id:
```
here, and
```
expect(saveNotReceivedUtilisationReportSpy).toHaveBeenCalledWith(mockReportPeriod, bankWithoutReport);
```
in the assertion below, then it's all self-explanatory.

dtfs-central-api/src/services/repositories/utilisation-reports-repo/utilisation-reports-repo.ts
```diff
@@ -58,14 +72,17 @@ export const getOpenReportsBeforeReportPeriodForBankId = async (reportPeriodStar
...
       ],
     })
     .toArray();
 };
+
+export const getCurrentUtilisationReportByBankIdAndReportPeriod = async (
+  bankId: string,
+  currentReportPeriod: ReportPeriod,
+): Promise<UtilisationReport | null> =>
+  await getUtilisationReportDetailsByBankIdMonthAndYear(bankId, currentReportPeriod.start.month, currentReportPeriod.start.year);
```
toddym42, line 88: This feels a bit redundant now, and having "Current" in the function name isn't guaranteed because it's dependent on a value we pass. We could instead just have this as
```
export const getUtilisationReportByBankIdAndReportPeriod = async (
  bankId: string,
  reportPeriod: ReportPeriod,
): Promise<UtilisationReport | null> =>
  await getUtilisationReportDetailsByBankIdMonthAndYear(bankId, reportPeriod.start.month, reportPeriod.start.year);
```
but then we may as well just call `getUtilisationReportDetailsByBankIdMonthAndYear`. I wouldn't mind if we wanted to keep this as a helper, but I think we should at least remove `Current` from the function and param name.

dtfs-central-api/src/types/db-models/banks.ts
```diff
@@ -3,7 +3,9 @@ import { WithId } from 'mongodb';
 type ReportPeriodSchedule = {
   startMonth: number;
   endMonth: number;
-}
+};
+
+export type BankReportPeriodSchedule = ReportPeriodSchedule[];
```
toddym42, line 8: We now refer to both singular and plural as a schedule. Maybe the singular should be something like `ReportPeriodMonthRange` instead `ReportPeriodSchedule`?

dtfs-central-api/src/types/db-models/utilisation-reports.ts
toddym42, line 41: We can remove this comment now.

toddym42 (commented):
I still need to go over the `getCurrentReportPeriodForBankSchedule` function logic tomorrow, but these are the comments I have so far

dtfs-central-api/src/services/repositories/utilisation-reports-repo/utilisation-reports-repo.ts
JoshBinns2000: I agree that it is redundant now, although I feel that getting reports by report period rather than month and year is the better way to go given our new way of storing reports. I would prefer to replace the 'by month and year' function with this 'by report period', but what do you think?

dtfs-central-api/src/types/db-models/banks.ts
JoshBinns2000: Decided to remove `ReportPeriodSchedule` and just define
```
type BankReportPeriodSchedule = {
  startMonth: number;
  endMonth: number;
}[];
```

dtfs-central-api/src/services/repositories/utilisation-reports-repo/utilisation-reports-repo.ts
JoshBinns2000: I decided instead to further refine how this function works. It not uses all the information from the `reportPeriod` to select the report rather than just using the `month` and `year`

dtfs-central-api/src/utils/report-period.ts
```diff
@@ -27,3 +29,40 @@ export const getPreviousReportPeriodStart = ({ month, year }: MonthAndYear): Mon
...
+  }
+
+  const scheduleOverlappingYear = bankReportPeriodSchedule.filter((schedule) => schedule.startMonth > schedule.endMonth).at(0);
+  if (!scheduleOverlappingYear) {
+    throw new Error('Failed to get a report period');
+  }
+
+  const reportPeriodIsInPreviousYear = targetMonth <= scheduleOverlappingYear.endMonth;
```
toddym42, line 57: While the logic may be correct, I'm not sure that this name is right. In the following scenario where `12` last year is the target
```
   2023 | 2024
11 (12) | 1
```
`reportPeriodIsInPreviousYear` would be false, but the report period _is_ in the previous year. I'm not sure what the correct name would be, and it's a bit hard to follow why this works.

How about this?
```
export const getCurrentReportPeriodForBankSchedule = (bankReportPeriodSchedule: BankReportPeriodSchedule): ReportPeriod => {
  const today = new Date();
  const previousMonthDate = subMonths(today, 1);
  const targetMonth = getOneIndexedMonth(previousMonthDate);
  const targetYear = previousMonthDate.getFullYear();

  const scheduleInCurrentYear = bankReportPeriodSchedule.find((schedule) => targetMonth >= schedule.startMonth && targetMonth <= schedule.endMonth);
  if (scheduleInCurrentYear) {
    return {
      start: {
        month: scheduleInCurrentYear.startMonth,
        year: targetYear,
      },
      end: {
        month: scheduleInCurrentYear.endMonth,
        year: targetYear,
      },
    };
  }

  const scheduleOverlappingYear = bankReportPeriodSchedule.filter((schedule) => schedule.startMonth > schedule.endMonth).at(0);
  if (!scheduleOverlappingYear) {
    throw new Error('Failed to get a report period');
  }

  const currentYear = today.getFullYear();
  const reportPeriodStartsInPreviousYear = targetYear < currentYear || targetMonth <= scheduleOverlappingYear.endMonth;

  return {
    start: {
      month: scheduleOverlappingYear.startMonth,
      year: reportPeriodStartsInPreviousYear ? currentYear - 1 : currentYear,
    },
    end: {
      month: scheduleOverlappingYear.endMonth,
      year: reportPeriodStartsInPreviousYear ? currentYear : currentYear + 1,
    },
  };
};
```

toddym42: Edit: Now I'm not sure if this accounts to the current case, where the entire report period is in the previous year...
```
      2023 | 2024
10 11 (12) |

Comment (JoshBinns2000) [reply dtfs-central-api/src/utils/report-period.ts:57]: And, further, the report period is actually in the _next_ year relative to the target year

Comment (JoshBinns2000) [reply dtfs-central-api/src/utils/report-period.ts:57]: I think it becomes a bit clearer if the variable name is `reportPeriodStartsInPreviousYear`

Comment (toddym42) [reply dtfs-central-api/src/utils/report-period.ts:57]: Ah, I think the "relative to target year" was the logic I was missing. Yeah, I think if we name this `reportPeriodStartsInPreviousYear` then it makes sense to me 👍

**Review by toddym42 — APPROVED** [review]

File: dtfs-central-api/.eslintrc.js [hunk 15]
```diff
@@ -52,7 +52,10 @@ module.exports = {
         project: './tsconfig.eslint.json',
         tsconfigRootDir: __dirname,
       },
-      rules: baseRules,
+      rules: {
+        ...baseRules,
+        '@typescript-eslint/no-floating-promises': ['error', { ignoreIIFE: true }],
```
abhi-markan, line 57: @toddym42 one for your prettier PR to remember.

dtfs-central-api/src/services/repositories/utilisation-reports-repo/utilisation-reports-repo.ts
```diff
@@ -30,6 +31,18 @@ export const saveUtilisationReportDetails = async (reportPeriod: ReportPeriod, a
   return { reportId: savedDetails.insertedId.toString(), dateUploaded: utilisationReportInfo.dateUploaded };
 };
 
+export const saveNotReceivedUtilisationReport = async (reportPeriod: ReportPeriod, bank: SessionBank) => {
```
abhi-markan, line 34: Missing docstring.

dtfs-central-api/src/services/repositories/utilisation-reports-repo/utilisation-reports-repo.ts
```diff
@@ -58,14 +71,22 @@ export const getOpenReportsBeforeReportPeriodForBankId = async (reportPeriodStar
...
+              $and: [{ year: { $eq: reportPeriodStart.year } }, { month: { $lt: reportPeriodStart.month } }],
             },
           ],
         },
       ],
     })
     .toArray();
 };
+
+export const getUtilisationReportDetailsByBankIdAndReportPeriod = async (bankId: string, reportPeriod: ReportPeriod): Promise<UtilisationReport | null> => {
```
abhi-markan, line 83: Missing docstring.

dtfs-central-api/src/utils/report-period.ts
```diff
@@ -27,3 +29,40 @@ export const getPreviousReportPeriodStart = ({ month, year }: MonthAndYear): Mon
 
 export const isEqualReportPeriodStart = (reportPeriodStart1: MonthAndYear, reportPeriodStart2: MonthAndYear): boolean =>
   reportPeriodStart1.year === reportPeriodStart2.year && reportPeriodStart1.month === reportPeriodStart2.month;
+
+export const getCurrentReportPeriodForBankSchedule = (bankReportPeriodSchedule: BankReportPeriodSchedule): ReportPeriod => {
```
abhi-markan, line 33: Please add docstring to all the functions.

dtfs-central-api/src/utils/report-period.ts
```diff
@@ -27,3 +29,40 @@ export const getPreviousReportPeriodStart = ({ month, year }: MonthAndYear): Mon
...
+      end: {
+        month: scheduleInCurrentYear.endMonth,
+        year: targetYear,
+      },
+    };
+  }
+
+  const scheduleOverlappingYear = bankReportPeriodSchedule.find((schedule) => schedule.startMonth > schedule.endMonth);
+  if (!scheduleOverlappingYear) {
+    throw new Error('Failed to get a report period');
```
abhi-markan, line 54: Do we need `console.error` or will the calling function `catch()` block have a `console.error`?

dtfs-central-api/src/scheduler/jobs/create-utilisation-reports/index.ts
```diff
@@ -0,0 +1,45 @@
+import { getUtilisationReportDetailsByBankIdAndReportPeriod, saveNotReceivedUtilisationReport } from '../../../services/repositories/utilisation-reports-repo';
+import { getAllBanks } from '../../../services/repositories/banks-repo';
+import { SchedulerJob } from '../../../types/scheduler-job';
+import { Bank } from '../../../types/db-models/banks';
+import { getCurrentReportPeriodForBankSchedule } from '../../../utils/report-period';
+import { asString } from '../../../helpers/validation';
+
+const currentBankReportIsMissing = async (bank: Bank): Promise<boolean> => {
```
abhi-markan, line 8: Docstrings please on this file.

dtfs-central-api/src/scheduler/jobs/create-utilisation-reports/index.ts
```diff
@@ -0,0 +1,45 @@
...
+  const currentReportPeriod = getCurrentReportPeriodForBankSchedule(bank.utilisationReportPeriodSchedule);
+  const currentUtilisationReportForBank = await getUtilisationReportDetailsByBankIdAndReportPeriod(bank.id, currentReportPeriod);
+  return !currentUtilisationReportForBank;
+};
+
+const getBanksWithoutReports = async () => {
+  const banks = await getAllBanks();
+
+  const isMissingBankReport = await Promise.all(banks.map(currentBankReportIsMissing));
+  return banks.filter((_, index) => isMissingBankReport[index]);
```
abhi-markan, line 18: We dont' use `_` for unused variable, please provide a meaningful name to the variable.

dtfs-central-api/src/scheduler/jobs/create-utilisation-reports/index.ts
```diff
@@ -0,0 +1,45 @@
...
+
+const createUtilisationReportForBanks = async (): Promise<void> => {
+  const banksWithoutReports = await getBanksWithoutReports();
+  if (banksWithoutReports.length === 0) {
+    return;
+  }
+
+  await Promise.all(
+    banksWithoutReports.map(async ({ id, name, utilisationReportPeriodSchedule }) => {
+      console.info(`Attempting to insert report for bank with id ${id}`);
```
abhi-markan, line 29: Please use string substitution on `console.info` rather interpolation.

dtfs-central-api/src/scheduler/jobs/create-utilisation-reports/index.ts
```diff
@@ -0,0 +1,45 @@
...
+const getBanksWithoutReports = async () => {
+  const banks = await getAllBanks();
+
+  const isMissingBankReport = await Promise.all(banks.map(currentBankReportIsMissing));
+  return banks.filter((_, index) => isMissingBankReport[index]);
+};
+
+const createUtilisationReportForBanks = async (): Promise<void> => {
+  const banksWithoutReports = await getBanksWithoutReports();
+  if (banksWithoutReports.length === 0) {
```
abhi-markan, line 23: Can this be `!banksWithoutReports.length`?

dtfs-central-api/src/scheduler/jobs/create-utilisation-reports/index.ts
```diff
@@ -0,0 +1,45 @@
...
+
+      await saveNotReceivedUtilisationReport(reportPeriod, sessionBank);
+      console.info(`Successfully inserted report for bank with id ${id}`);
+    }),
+  );
+};
+
+export const createUtilisationReportForBanksJob: SchedulerJob = {
+  init: () => ({
+    schedule: asString(process.env.UTILISATION_REPORT_CREATION_FOR_BANKS_SCHEDULE, 'UTILISATION_REPORT_CREATION_FOR_BANKS_SCHEDULE'),
```
abhi-markan, line 41: Might be prudent to de-structure `process.env` at the top to keep the coding style consistent?

dtfs-central-api/src/scheduler/index.ts
```diff
@@ -0,0 +1,29 @@
...
+export const initScheduler = () => {
+  jobs.forEach((job) => {
+    const { schedule, message, task } = job.init();
+
+    if (!cron.validate(schedule)) {
+      console.error(`Failed to add job '${message}' due to invalid schedule '${schedule}'`);
+      return;
+    }
+
+    console.info(`Adding scheduled job '${message}'`);
```
abhi-markan, line 26: String substitution across all the `console` statements please.

.env.sample
```diff
@@ -44,6 +44,7 @@ UTILISATION_REPORT_OVERDUE_CHASER_DATE_BUSINESS_DAYS_FROM_START_OF_MONTH=15
 UTILISATION_REPORT_SUBMISSION_PERIOD_START_EMAIL_SCHEDULE='0 8 1 * *'
 UTILISATION_REPORT_DUE_EMAIL_SCHEDULE='0 8 * * *'
 UTILISATION_REPORT_OVERDUE_EMAIL_SCHEDULE='0 8 * * *'
+UTILISATION_REPORT_CREATION_FOR_BANKS_SCHEDULE='0 2 * * *'
```
abhi-markan, line 47: Should this also be added to `test.yml`?

abhi-markan (changes requested):

dtfs-central-api/src/services/repositories/utilisation-reports-repo/utilisation-reports-repo.ts
```diff
@@ -17,7 +18,7 @@ export const saveUtilisationReportDetails = async (reportPeriod: ReportPeriod, a
     reportPeriod,
     dateUploaded: new Date(),
     azureFileInfo,
-    status: UTILISATION_REPORT_RECONCILIATION_STATUS.PENDING_RECONCILIATION,
+    status: 'PENDING_RECONCILIATION',
```
abhi-markan, line 21: This appears to have been moved back from constant, any specific reason?

abhi-markan (changes requested):

dtfs-central-api/src/services/repositories/utilisation-reports-repo/utilisation-reports-repo.ts
JoshBinns2000: Yes - in typescript files, if the type is a `string` literal union, you are only able to set the value to one of the strings specified by the type. As the type is extracted from the constant, we don't need to use the constant explicitly in typescript files as they will always be in sync.

In a typescript first project, you wouldn't need a constant for `string`-enum things like this and would only have the type:
```
type UtilisationReportReconciliationStatus =
  | 'PENDING_RECONCILIATION'
  | 'REPORT_NOT_RECEIVED'; // etc

/**
  * intellisense shows either 'PENDING_RECONCILIATION' or 'REPORT_NOT_RECEIVED' and
  * you get a TS-error if you put anything else
  */
const status: UtilisationReportReconciliationStatus = 
```

I imagine we never would have introduced the `UTILISATION_REPORTS_RECONCILIATION_STATUS` constant if we didn't need to use it in some javascript files, so removing the use of the constant where we can is done with the hope of being able to permanently remove it eventually

abhi-markan: Happy to approve this, however in my opinion repeating status multiple times is not keeping source code DRY.
One for a wider discussion we should look into following links for references:

* https://github.com/Microsoft/TypeScript/pull/6532
* https://www.typescriptlang.org/docs/handbook/enums.html
* https://github.com/UK-Export-Finance/exip

abhi-markan (approved):

dtfs-central-api/src/services/repositories/utilisation-reports-repo/utilisation-reports-repo.ts
toddym42: My opinion on this (FWIW) is that this doesn't go against DRY principles at all. You're either repeating
```
UTILISATION_REPORT_RECONCILIATION_STATUS.PENDING_RECONCILIATION
```
or you're repeating
```
'PENDING_RECONCILIATION'
```
The constant is redundant, and doesn't add any value when using TypeScript and a union type.

If the type of `status` was `string`, or we were writing plain JavaScript then the constant _does_ add value because it limits the values to those set by the constant, and makes refactoring easier.

But that's not the case here. The type of `status` is a union type
```
export type UtilisationReportReconciliationStatus =
  | 'REPORT_NOT_RECEIVED'
  | 'PENDING_RECONCILIATION'
  | 'RECONCILIATION_IN_PROGRESS'
  | 'RECONCILIATION_COMPLETED';
```
so with TypeScript you can only set one of the above values. Your IDE provides you with auto-complete for these values, and if you try to set anything else TypeScript will shout at you and the code will fail to compile.

This is similar to the `boolean` type which is a union of `true | false`. In that case there isn't any need to use a constant (e.g. `BOOLEAN.TRUE` or `BOOLEAN.FALSE`) and the same principle can be applied here.

PR #2698, opened 2024-01-30
feat(DTFS2-6932): add pagination to deals page front end

oscar-richardson-softwire:
## Introduction :pencil2:
This PR introduces pagination (which works alongside searching and sorting) to the TFM deals table. The necessary back-end changes where made as part of [6931](https://github.com/UK-Export-Finance/dtfs2/pull/2561) - this PR adds the front-end changes.

- Refactored the TFM UI deals controller to redirect POST /deals requests sent by searching/sorting the deals table to the same GET /deals endpoint used when the table is first loaded, with appropriate query parameters. This differs from the old approach of POST /deals re-rendering the table itself. The new approach has the following benefits:

1. Allows simultaneous searching and sorting.
2. Makes the code more DRY.
3. Prevents unwanted duplicate requests to POST /deals if the user refreshes the page after searching/sorting the table.

- Copied the pagination approach and macro used in the deals table in Portal UI.
- Manually verified that these changes did not break Audrius' fixes for the accessibility tickets [6472](https://ukef-dtfs.atlassian.net/browse/DTFS2-6472), [6474](https://ukef-dtfs.atlassian.net/browse/DTFS2-6474) and [6526](https://ukef-dtfs.atlassian.net/browse/DTFS2-6526) using NVDA (a screen reader).

- The pagination element (i.e., containing the links to the different pages) may have accessibility issues. I tested it with NVDA; the heading 'pagination' is read out, as well as all of the links (e.g., 'First', 'Next', '1', '2' etc.). However, this is only read out after the entire table contents has been read out. I am not sure if this is desirable. Perhaps @avaitonis would be able to advise.
- The pagination approach and macro used in the deal in the deals table in Portal UI are currently used. It was thought that moving to the GOV UK pagination component might improve maintainability, but I spoke to Tony and he said that they had had problems with GOV UK pagination on EXIP. Instead, EXIP uses an adaptation of the ONS pagination component. It may be worth copying the approach used on EXIP here, to improve maintainability and alignment between projects.


trade-finance-manager-ui/server/controllers/deals/index.js
```diff
@@ -3,20 +3,31 @@ const { generateHeadingText } = require('../helpers');
 const CONSTANTS = require('../../constants');
 
 const getDeals = async (req, res) => {
+  if (req.params.pageNumber < 0) {
+    return res.redirect('/not-found');
+  }
```
oscar-richardson-softwire, line 8: Note to self: investigate this behaviour

trade-finance-manager-ui/server/controllers/deals/index.js
```diff
@@ -3,20 +3,31 @@ const { generateHeadingText } = require('../helpers');
...
   const queryParams = {
     sortBy: CONSTANTS.DEALS.TFM_SORT_BY_DEFAULT,
+    pagesize: CONSTANTS.DEALS.PAGE_SIZE,
+    page: req.params.pageNumber || 0,
   };
+
   const { userToken } = req.session;
 
-  const apiResponse = await api.getDeals(queryParams, userToken);
+  const { deals, pagination } = await api.getDeals(queryParams, userToken);
```
oscar-richardson-softwire, line 18: Note to self: is api.getDeals guaranteed to never return `undefined`?

trade-finance-manager-ui/server/controllers/deals/index.js
```diff
@@ -28,23 +39,32 @@ const getDeals = async (req, res) => {
...
     });
   }
 
   return res.redirect('/not-found');
 };
 
 const queryDeals = async (req, res) => {
+  if (req.params.pageNumber < 0) {
+    return res.redirect('/not-found');
+  }
```
oscar-richardson-softwire, line 66: Note to self: investigate this behaviour

trade-finance-manager-ui/server/controllers/deals/index.js
```diff
@@ -55,7 +75,11 @@ const queryDeals = async (req, res) => {
     searchString = req.body.search;
   }
 
-  const queryParams = { searchString };
+  const queryParams = {
+    searchString,
+    pagesize: CONSTANTS.DEALS.PAGE_SIZE,
+    page: req.params.pageNumber || 0,
```
oscar-richardson-softwire, line 81: Check if `params` is always defined

trade-finance-manager-ui/server/controllers/deals/index.js
```diff
@@ -67,7 +91,11 @@ const queryDeals = async (req, res) => {
     };
   }
 
-  const { deals, count } = await api.getDeals(queryParams, userToken);
+  const { deals, pagination } = await api.getDeals(queryParams, userToken);
+
+  if (req.params.pageNumber >= pagination.totalPages) {
+    return res.redirect('/not-found');
+  }
```
oscar-richardson-softwire, line 98: Note to self: investigate this behaviour

trade-finance-manager-ui/server/routes/deals/index.js
```diff
@@ -3,7 +3,8 @@ const { getDeals, queryDeals } = require('../../controllers/deals');
 
 const router = express.Router();
 
-router.get('/', getDeals);
+router.get('/', (req, res) => res.redirect('/deals/0'));
```
oscar-richardson-softwire, line 6: Check if this line is necessary

trade-finance-manager-ui/templates/_macros/pagination.njk
```diff
@@ -0,0 +1,62 @@
...
+                <a data-cy="Page_{{i}}" href="{{paginationRoute}}/{{i}}{{queryString}}" class="govuk-link">{{ i + 1 }}</a>
+              </li>
+            {% endif %}
+          {%- endfor %}
+
+          {% if totalPages - currentPage > 4 %}
+            <li class="govuk-body govuk-!-margin-bottom-3">
+              ...
+            </li>
+          {% endif %}
```
oscar-richardson-softwire, line 40: This and the similar part above it don't seem symmetrical (double check)

trade-finance-manager-ui/templates/_macros/pagination.njk
```diff
@@ -0,0 +1,62 @@
...
+            <li class="govuk-body govuk-!-margin-bottom-3">
+              <a data-cy="Previous" href="{{paginationRoute}}/{{currentPage-1}}{{queryString}}" class="govuk-link">Previous</a>
+            </li>
+          {% endif %}
+
+          {% if currentPage > 5 %}
+            <li class="govuk-body govuk-!-margin-bottom-3">
+              ...
+            </li>
+          {% endif %}
```
oscar-richardson-softwire, line 26: This and the similar part below it don't seem symmetrical (double check)

trade-finance-manager-ui/server/controllers/deals/index.js
```diff
@@ -3,20 +3,31 @@ const { generateHeadingText } = require('../helpers');
...
 
 const getDeals = async (req, res) => {
+  if (req.params.pageNumber < 0) {
+    return res.redirect('/not-found');
+  }
+
   const queryParams = {
     sortBy: CONSTANTS.DEALS.TFM_SORT_BY_DEFAULT,
+    pagesize: CONSTANTS.DEALS.PAGE_SIZE,
+    page: req.params.pageNumber || 0,
```
oscar-richardson-softwire: Consider changing to `pageNumber`

trade-finance-manager-ui/server/constants/deals.js
```diff
@@ -11,4 +11,5 @@ const TFM_SORT_BY_DEFAULT = {
 module.exports = {
   TFM_SORT_BY,
   TFM_SORT_BY_DEFAULT,
+  PAGE_SIZE: 20,
```
AlexBramhill, line 14: Sorry - sticking with convention:
```suggestion
  TFM_PAGE_SIZE: 20,
```

AlexBramhill (commented):
couple of minor questions, really good job with this. I think this is one of the places we will keep refining as we need in terms of abstraction.

Feel free to merge once you've let me know about the page number and breadcrumb query :)

trade-finance-manager-ui/server/api.js
```diff
@@ -79,12 +85,16 @@ const getDeals = async (queryParams, token) => {
...
+        pagination: response.data.pagination,
       };
     }
     return {
       deals: [],
-      count: 0,
+      pagination: {
+        totalItems: 0,
+        currentPage: queryParams.page,
+        totalPages: 0,
```
ttbarnes, line 96: Could the default `totalItems` and `totalPages` be constants? e.g `DEFAULT_...`

trade-finance-manager-ui/server/controllers/deals/index.js
```diff
@@ -2,79 +2,66 @@ const api = require('../../api');
...
-
-  if (req.body.search) {
-    searchString = req.body.search;
+/**
+ * Makes a request to TFM API to get the deals and renders the deals page with the deals
+ * @param {Object} req The request object
+ * @param {Object} res The response object
+ * @returns {Object} The result from rendering the page
+ */
+const getDeals = async (req, res) => {
```
ttbarnes, line 34: Should this be wrapped in a try/catch? Maybe not for this PR and needs doing in other controllers

trade-finance-manager-ui/templates/deals/deals.njk
```diff
@@ -34,6 +35,14 @@
 
       {{ dealsTable.render(componentData) }}
 
+      {{ pagination.pagination(
```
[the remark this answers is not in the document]
ttbarnes: Minor, could this be `pagination.render` instead of `pagination.pagination`?

trade-finance-manager-ui/server/controllers/deals/index.js
```diff
@@ -2,79 +2,66 @@ const api = require('../../api');
...
+    field: sortfield ?? CONSTANTS.DEALS.TFM_SORT_BY_DEFAULT.field,
+    order: sortorder ?? CONSTANTS.DEALS.TFM_SORT_BY_DEFAULT.order,
+  };
+  const queryParams = {
+    sortBy,
+    pagesize: CONSTANTS.DEALS.PAGE_SIZE,
+    page: pageNumber,
+  };
+  if (search) {
+    queryParams.searchString = search;
```
ttbarnes, line 51: Could this maybe be moved into a helper function? e.g `generateQueryParams`, would be nicer for unit tests and simpler controller

trade-finance-manager-ui/server/controllers/deals/index.js
```diff
@@ -92,18 +79,58 @@ const queryDeals = async (req, res) => {
     });
   }
 
+  const activeSortByField = sortBy.field;
+  const activeSortByOrder = sortBy.order;
+  const sortButtonWasClicked = sortfield ? true : false;
```
ttbarnes, line 84: Seems this could just be `sortButtonWasClicked = sortfield` or maybe rename `sortfield` to `sortButtonWasClicked`

trade-finance-manager-ui/server/controllers/deals/index.js
```diff
@@ -2,79 +2,66 @@ const api = require('../../api');
...
+  if (sortfield && sortorder) {
+    queryParameters.push(`sortfield=${sortfield}&sortorder=${sortorder}`);
   }
 
-  return res.redirect('/not-found');
+  let queryString = '';
+  if (queryParameters.length > 0) {
+    queryString = `?${queryParameters.join('&')}`;
+  }
+  return queryString;
```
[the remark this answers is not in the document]
oscar-richardson-softwire: There's never a space in any of the `sortField` values :)

Ooh I've have at look at `URLSearchParams`. I won't include it as part of this PR but I might include it as part of [DTFS2-6933](https://ukef-dtfs.atlassian.net/browse/DTFS2-6933) when I commonise some of the code between the deals and facilities tables.

trade-finance-manager-ui/server/api.js
oscar-richardson-softwire: Personally I don't think there's much value in this - this is just want we return when there is no response data, so we would never want to change these to be non-zero. I think leaving it inline makes the code more descriptive.

trade-finance-manager-ui/server/controllers/deals/index.js
oscar-richardson-softwire: Done :)

trade-finance-manager-ui/templates/deals/deals.njk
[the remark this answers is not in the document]
oscar-richardson-softwire: I've actually changed it to pagination.render in my other open PR (DTFS2-6935) so I'll leave it here.

trade-finance-manager-ui/server/controllers/deals/index.js
oscar-richardson-softwire: I've done this. Note that the only error we need to handle with anything other than a 'problem with the service' is when the page number is out of bounds.

trade-finance-manager-ui/server/api.js
```diff
@@ -68,28 +69,30 @@ const getFacilities = async (token, searchString = '') => {
   }
 };
 
+/**
+ * Makes a request to the GET /deals TFM API endpoint
+ * and throws an error if the page number is out of bounds
+ * @param {Object} queryParams Query parameters
+ * @param {string} token Authorisation token
+ * @returns {Object} Deals data and pagination metadata
```
AlexBramhill, line 77: `@throws`

AlexBramhill, line 77: `@throws`

AlexBramhill (commented):
lgtm -- one small comment about adding `throws` to docs, but then happy. Will let Tony rereview and approve

trade-finance-manager-ui/server/api.js
oscar-richardson-softwire: Added

trade-finance-manager-ui/server/api.js
```diff
@@ -68,28 +69,31 @@ const getFacilities = async (token, searchString = '') => {
...
-    return {};
+  const response = await axios({
+    method: 'get',
+    url: `${TFM_API_URL}/v1/deals`,
+    headers: generateHeaders(token),
+    params: queryParams,
+  });
+  const { deals, pagination } = response.data;
+
+  if (queryParams.page >= pagination?.totalPages && !(queryParams.page === 0 && pagination?.totalPages === 0)) {
```
ttbarnes, line 89: Found this a little hard to understand, could this just be the following? Not sure why we'd need a check for `!(page === 0 && totalPages === 0)`?

```js
if (page > pagination.totalPages) {
  // throw 
}
```

Should `totalPages` always be defined, save the `pagination?...`?

trade-finance-manager-ui/server/controllers/deals/index.js
```diff
@@ -1,107 +1,151 @@
 const api = require('../../api');
 const { generateHeadingText } = require('../helpers');
 const CONSTANTS = require('../../constants');
-
-const getDeals = async (req, res) => {
+const PageOutOfBoundsError = require('../../errors/page-out-of-bounds.error');
+
```
ttbarnes, line 5: I can't see unit tests for these changes?

ttbarnes (commented):
Left 2 comments, otherwise LGTM

trade-finance-manager-ui/server/api.js
oscar-richardson-softwire: Good spot - the reason for this was that the backend returned totalPages = 0 when there were no deals, but I realised it was better to remove this logic here and instead change the backend to return totalPages = 1 when there are no deals. You are also right, pagination is always defined, so there is no need for the ?

trade-finance-manager-ui/server/controllers/deals/index.js
oscar-richardson-softwire: There is a unit test for PageOutOfBoundsError. If you mean the frontend pagination changes in general, these tests are in a different ticket, 6935

ttbarnes (approved):

PR #2699, opened 2024-01-30
style(FN-1798): standardise prettier config and add to pre-commit commit hook

toddym42:
**Note**: this PR does not yet include running of prettier across the whole codebase, instead only containing the changes required to update the config (as detailed below) and resolve any linting issues that occur as a result of these changes. This makes this PR much easier to review as these are the important changes that represent actual non-formatting updates. Once approved I will separately add prettier to lint staged and run across the whole project (an extra ~1600 files). This shouldn't need a thorough review as it's only formatting changes. Once all tests pass it should then be good to merge.

To add prettier to the commit hook we should update `lint-staged` in the the root `package.json` from
```
"**/*.{js,ts}": [
  "eslint --fix"
],
```
to
```
"**/*.{js,ts}": [
  "eslint --fix",
  "prettier --write"
],
```
Then additionally run
```
prettier:fix:all`
```
 to reformat all files in the project.

## Introduction :pencil2:

Prettier is an opinionated code formatter that can help keep a consistent code formatting style in the project. There are currently the following issues with how it is used:

* The prettier package and config have been added in the root project, but some individual sub-projects (external-api and portal-api) have their own version of prettier and config added. This may lead to inconsistencies if the versions and config diverge
* Sometimes running prettier on files results in conflicts with the existing eslint config, so prettier changes need to be reverted to keep the linter happy
* Running of prettier is not enforced in any way (either with commit hooks or CI)

* Remove the prettier package and config from sub-projects so that formatting across the entire codebase will be based on the root directory prettier installation and config. Updated the README to clarify this point.
* Add [eslint-plugin-prettier](https://github.com/prettier/eslint-plugin-prettier) package to the elint config in all projects - this resolves any conflicts between eslint and prettier so we can ensure that prettier can always be run, and shows prettier formatting issues as elint errors
* Add running of prettier to the commit hook so that this is always run on staged files even if the dev hasn't run the formatter manually
* Move the running of lint-staged to a "pre-commit" hook instead of the "commit-msg" hook. This means that any eslint or prettier fixes that get done as part of the commit hook are done before the commit is submitted, meaning they will be included in commit you are submitting. As lint-staged fixes are currently included in the "commit-msg" hook, this means you you would add a commit, then need to add a further separate commit with any eslint or prettier changes, which may lead to these changes not seen and committed. (Further details of the different types of commit hooks [here](https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks))
* Add a prettier check to the CI pipeline to check that all files are formatted according to our configured prettier formatting style


abhi-markan:
PR to be replaced by https://github.com/UK-Export-Finance/dtfs2/pull/3030

PR #2708, opened 2024-02-01
feat(DTFS2-6621): ensure usernames are email addresses

AlexBramhill:
## Introduction :pencil2:
We need to ensure that usernames are email addresses.

I've added in validation to both the create a user and modify a user endpoints

This scenario is impossible to reach on our front end -- but the inclusion of these checks prevent this should it be added in the future.

I have updated e2e tests to refer to the email constants in the mock data loader. I have not done this for API and unit tests due to this being a large existing tech debt to be tackled in a separate ticket (see this ticket on the board).


dtfs-central-api/api-tests/v1/user/user-get.api-test.js
```diff
@@ -4,7 +4,7 @@ const { DB_COLLECTIONS } = require('../../../src/constants');
 const api = require('../../api')(app);
 
 const newUser = {
-  username: 'BANK1_MAKER1',
+  username: 'maker1@ukexportfinance.gov.uk',
```
AlexBramhill, line 7: Where applicable, I've replaced these with constants -- I've raised another ticket for the tech debt around our user objects in test cases

e2e-tests/portal/cypress/e2e/journeys/checker/cannot-submit-a-deal-with-newly-issued-facilities-that-have-cover-start-dates-before-min-submission-date/deal.js
```diff
@@ -17,7 +18,7 @@ const deal = {
   },
   maker: {
     _id: '5f3ab3f705e6630007dcfb25',
-    username: 'maker1@ukexportfinance.gov.uk',
+    username: BANK1_MAKER1.username,
```
AlexBramhill, line 21: Again -- these can be rationalised more, however this is a large refactor so I've raised another ticket

gef-ui/server/utils/mocks/mock_requests.js
```diff
@@ -4,7 +4,7 @@ const { MAKER, CHECKER } = require('../../constants/roles');
 const chance = new Chance();
 
 const MOCK_REQUEST = {
-  username: 'BANK1_MAKER1',
+  username: 'maker1@ukexportfinance.gov.uk',
```
AlexBramhill, line 7: We don't call from `users.js` in data loader here as these mocks (unit mocks/api mocks) should not/do not interact with the db.

portal-api/api-tests/v1/users/post-create-user.api-test.js
```diff
@@ -0,0 +1,238 @@
+const databaseHelper = require('../../database-helper');
```
AlexBramhill, line 1: New tests added on `loc 115 - loc 154` -- I've just pulled this file out into it's own thing so we don't have such a giant `users.api-test.js`. Naming of file follows existing pattern for api-tests

portal-api/api-tests/v1/users/post-create-user.api-test.js
```diff
@@ -0,0 +1,238 @@
...
+          ...MOCK_USER,
+          email: '',
+        };
+
+        const { status, body } = await createUser(newUser);
+
+        expect(status).toEqual(400);
+        expect(body.success).toEqual(false);
+        expect(body.errors.errorList.email.text).toEqual(EMAIL_ERROR.text);
+      });
```
AlexBramhill, line 140: For simplicity -- as this is never going to be visible on the front end (as we only have an `email` field in `portal`), we only have two rules -- one checking the `email` is an email address, and the other checking the `username` matches. There is no need to check the `username` is an email address -- as it cannot get through otherwise, and the only way someone would get these errors would be if they posted directly to `portal-api` -- in which case we assume a familiarity of `portal-api`. Basically - YAGNI

portal-api/api-tests/v1/users/validate-username-and-email.api-tests.js
```diff
@@ -0,0 +1,74 @@
...
+        expectedErrorForUsernameTests: USERNAME_AND_EMAIL_MUST_MATCH_ERROR.text,
+      },
+      {
+        description: `is an object`,
+        valueToSetField: AN_OBJECT,
+        expectedErrorForEmailTests: EMAIL_ERROR.text,
+        expectedErrorForUsernameTests: USERNAME_AND_EMAIL_MUST_MATCH_ERROR.text,
+      },
+    ];
+    describe('when validating the email', () => {
```
AlexBramhill, line 43: Yes this could technically be refactored into just one "describe.each" but as this is a test case, for simplicity of reading I haven't

portal-api/api-tests/api-test-users.js
```diff
@@ -253,11 +253,11 @@ const finder = () => {
 };
 
 const apiTestUser = {
-  username: 'api-test-user',
+  username: 'api-tester@ukexportfinance.gov.uk',
   password: 'P@ssword1234',
   firstname: 'API',
   surname: 'Test User',
-  email: 'api-tester@example.com',
+  email: 'api-tester@ukexportfinance.gov.uk',
```
ttbarnes, line 260: Could all of the emails in this file come from constants, similar to `BANK1_CHECKER1.email,` above?

portal-api/api-tests/v1/users/login-with-sign-in-link.api-test.js
```diff
@@ -36,7 +36,8 @@ describe('POST /users/:userId/sign-in-link/:signInToken/login', () => {
   let partiallyLoggedInUserId;
   let partiallyLoggedInUserToken;
 
-  const login = ({ userId, signInToken, userToken }) => as({ token: userToken }).post().to(`/v1/users/${userId}/sign-in-link/${signInToken}/login`);
+  const loginWithSignInLink = ({ userId, signInToken, userToken }) =>
+    as({ token: userToken }).post().to(`/v1/users/${userId}/sign-in-link/${signInToken}/login`);
```
ttbarnes, line 40: More for my own understanding, what does "sign in link" mean in this context? Just a token?

Finding the naming "login with sign in link" confusing because "login" and "sign in" mean the same thing.

portal-api/api-tests/v1/users/post-sign-in-link.api-test.js
```diff
@@ -39,8 +39,8 @@ describe('POST /users/me/sign-in-link', () => {
 
   const saltBytes = Buffer.from(saltHexOne, 'hex');
   const signInToken = '0a1b2c3d4e5f67890a1b2c3d4e5f6789';
-  const username = 'TEMPORARY_USER';
-  const userToCreateAsPartiallyLoggedIn = { ...aMaker, username };
+  const temporaryUsernameAndEmail = 'temporary_user@ukexportfinance.gov.uk';
```
ttbarnes, line 42: Could the value be a constant?

portal-api/src/v1/users/validation/rules/email-must-be-valid-email-address.js
```diff
@@ -0,0 +1,19 @@
+const { isValidEmail } = require('../../../../utils/string');
+
+module.exports = (user, change) => {
```
ttbarnes, line 3: Could we add some documentation?

Not sure if this is preferred in DTFS, but I find having a function name at the top and `module.exports` at the bottom nicer  e.g

```js
const emailMustBeValid = () => { };

module.exports = emailMustBeValid;
```

portal-api/api-tests/v1/users/post-sign-in-link.api-test.js
ttbarnes: Noticed some emails are now using constants, but others are hard coded. Should we update all emails to use constants? Maybe another ticket?

portal-api/api-tests/api-test-users.js
```diff
@@ -45,169 +45,169 @@ const testUsers = [
...
-    username: 'no-roles',
+    username: 'no-roles@ukexportfinance.gov.uk',
     email: 'no-roles@ukexportfinance.gov.uk',
     password: 'P@ssword1234',
     roles: [],
     bank: {},
   },
   {
-    username: 'HSBC-maker-1',
+    username: 'hsbc-maker-1@ukexportfinance.gov.uk',
```
ttbarnes, line 55: Should `username` _not_ be an email? Not sure if this has changed since I last saw this 😊

portal-api/src/v1/users/validation/rules/username-and-email-must-match.js
```diff
@@ -0,0 +1,16 @@
...
+      },
+    },
+  ];
+
+  if (change?.username !== change?.email) {
+    return error;
+  }
+
+  return [];
+};
```
ttbarnes, line 16: Not a blocker - probably tech debt. Some documentation would be good here, also - probably should be done in one go, but I find it easier to read/maintain if we have a function name at the top and `module.exports` at the bottom of the file, e.g

```js
module.exports = usernameAndEmailMustMatchRule;
```

ttbarnes (approved):
LGTM - would be good to have all emails as constants, but understand is not for this ticket, tech improvement 👍

portal-api/src/v1/users/validation/rules/username-and-email-must-match.js
AlexBramhill: Agree - will add to next ticket

ttbarnes (approved):

PR #2711, opened 2024-02-01
chore(FN-2153): remove flaky 'dateFuncs' e2e test helpers

toddym42:
## Introduction :pencil2:
Remove hand-rolled date helper functions to prevent flaky test failures. The main culprit is the following function used to add months to today's date:
```
const nowPlusMonths = (numberMonths) => {
  const date = new Date();
  date.setMonth(date.getMonth() + numberMonths);
  return date;
};
```
Because this doesn't take into account days of the month this can end up unexpected results when the current month has more days than the month we are targeting. e.g. when today is 31st January 2024 and February 2024 has 29 days:
```
const nextMonth = nowPlusMonths(1);
```
```
const date = new Date(2024, 0, 31); // 31st January 2024
date.setMonth(date.getMonth() + 1);
// date now equals Date(2024, 1, 31), or 31st February 2024, which is invalid.
// JS instead interprets this as 2nd March 2024.
```
When we instead use a library, like `date-fns` and do
```
const today = new Date(2024, 0, 31); // 31st January 2024
const nextMonth = addMonths(today, 1);
```
the result is 29th February 2024. The library clamps the day within the target month.

The flaky tests occur because we are sometimes using library functions to correctly calculate dates and other times using the flaky `nowPlusMonths` function. This PR removes the hand-rolled functions to ensure consistency.


ChristianMcc (approved):
I have some thoughts about how we use dateConstants almost everywhere and not just generic functions that calculate (properly) based off inputs, but given this is the pattern very happy with this approach and fix!

e2e-tests/support/tasks.js
```diff
@@ -9,6 +9,12 @@ module.exports = {
     const getUsersCollection = () => db.getCollection(usersCollectionName, connectionOptions);
 
     return {
+      log(message) {
+        // eslint-disable-next-line no-console
+        console.log(message);
```
abhi-markan, line 14: Please remove `console.log` Eslint rules will only allow `info, warn and error`.

abhi-markan (commented):
Comment to refer to please.

PR #2747, opened 2024-02-05
fix(DTFS2-7018): NoSQLi mitigation

abhi-markan:
## Introduction :pencil2:
Ensure database query build are prone to NoSQLi.

* Used `$eq` operator so the untrusted value specified is interpreted as a literal value rather an object.


ChristianMcc (dismissed):
LGTM

ChristianMcc (approved):

PR #2751, opened 2024-02-06
fix(FN-2172): Fix UI with upload error

francescastocco:
## Introduction :pencil2:
When we had an unknown file upload error we didn't pass in `dueReportDates` to the nunjucks template so displayed the `No Reports Due`

- Pass due report dates to nunjucks template when we have an unknown file upload error
- Added an e2e test assertion for this
- Added more testing around file upload errors

## Miscellaneous
Removes all direct imports from `'<path>/constants/roles'` and replaces it with the import from `<path>/constants`.


portal/server/controllers/utilisation-report-service/utilisation-report-upload/index.test.js
JoshBinns2000, line 1: File is missing a new line

portal/server/controllers/utilisation-report-service/utilisation-report-upload/index.test.js
```diff
@@ -0,0 +1,136 @@
...
+          },
+        });
+
+      jest.mocked(getDueReportDates).mockReturnValueOnce([]);
+
+      // Act
+      await postUtilisationReportUpload(req, res);
+
+      // Assert
+      /* eslint-disable no-underscore-dangle */
```
JoshBinns2000, line 43: Seeing as this is only one line, we can use `// eslint-disable-next-line no-underscore-dangle`

portal/server/controllers/utilisation-report-service/utilisation-report-upload/index.test.js
```diff
@@ -0,0 +1,136 @@
...
+      expect(res._getRenderView()).toEqual('utilisation-report-service/utilisation-report-upload/check-the-report.njk');
+      /* eslint-enable no-underscore-dangle */
+    });
+
+    it("redirects to the 'confirm-and-send' url if no file errors", async () => {
+      // Arrange
+      const { res, req } = httpMocks.createMocks({
+        session: { userToken: 'user-token', user: MOCK_PORTAL_SESSION_USER },
+        file: { originalname: 'filename'},
```
JoshBinns2000, line 110: Probably need to run prettier on this file - missing space after `'filename'`

portal/server/controllers/utilisation-report-service/utilisation-report-upload/utilisation-report-upload-errors.js
JoshBinns2000, line 1: File missing new line

portal/server/test-mocks/mock-portal-session-user.ts
```diff
@@ -0,0 +1,17 @@
+import { PortalSessionUser } from '../types/portal-session-user';
+import { ROLES } from '../constants';
+
+export const MOCK_PORTAL_SESSION_USER: PortalSessionUser = {
+  _id: '65954cc526d3899694cafff2',
+  username: 'PAYMENT_REPORT_OFFICER',
+  firstname: 'PAYMENT',
+  surname: 'REPORT OFFICER',
+  email: 'test@testing.com',
+  roles: [ROLES.PAYMENT_REPORT_OFFICER],
```
JoshBinns2000, line 10: You should be able to just use `'PAYMENT_REPORT_OFFICER'` here rather than using the constant (it would be good to stop using constants where the types can handle that for us)

portal/server/types/banks.ts
```diff
@@ -0,0 +1,22 @@
+type ReportPeriodSchedule = {
+  startMonth: number;
+  endMonth: number;
+}
```
JoshBinns2000, line 4: This type is going to change in the new update, so could we replace it with this:
```
type BankReportPeriodSchedule = {
  startMonth: number;
  endMonth: number;
}[];
```

portal/server/types/role-id.ts
```diff
@@ -0,0 +1,4 @@
+import { ValuesOf } from './types-helper';
+import { ROLES } from '../constants';
+
+export type RoleId = ValuesOf<typeof ROLES>;
```
JoshBinns2000, line 4: I think this type should be `Role`?

JoshBinns2000 (commented):
I think prettier needs to be run across a few files here :)

portal/server/controllers/utilisation-report-service/utilisation-report-upload/utilisation-report-upload-errors.test.js
```diff
@@ -0,0 +1,114 @@
+const httpMocks = require('node-mocks-http');
```
ChristianMcc, line 1: Nice tests :)

portal/server/controllers/utilisation-report-service/utilisation-report-upload/index.test.js
```diff
@@ -0,0 +1,136 @@
...
+        });
+
+      jest.mocked(getDueReportDates).mockReturnValueOnce([]);
+
+      // Act
+      await postUtilisationReportUpload(req, res);
+
+      // Assert
+      /* eslint-disable no-underscore-dangle */
+      expect(res._getRenderView()).toEqual('utilisation-report-service/utilisation-report-upload/utilisation-report-upload.njk');
```
ChristianMcc, line 70: Is there anyway to check what is passed into the render view as well here? or would you need to mock res.render out to be able to check what it was called with?

ChristianMcc (commented):
Nice new tests! 2 of my comments are on code that isn't yours but one small question about if we can test what is passed into our functions (technically our bug was calling the right render view but with wrong parameters so if we can test for that that would be great)

portal/server/controllers/utilisation-report-service/utilisation-report-upload/index.test.js
JoshBinns2000: I believe you can write
```
expect(res.render).toHaveBeenCalledWith({ ... });
```

JoshBinns2000: Correction: you use `res._getRenderData()`

portal/server/controllers/utilisation-report-service/utilisation-report-upload/utilisation-report-upload-errors.js
```diff
@@ -0,0 +1,45 @@
...
+    const { formattedReportPeriod } = req.session.utilisationReport;
+    const filename = req.file.originalname;
+    const { filenameError } = validateFilenameFormat(filename, formattedReportPeriod);
+    if (filenameError) {
+      const uploadErrorSummary = [{ text: filenameError, href }];
+      const uploadValidationError = { text: filenameError };
+      return { uploadErrorSummary, uploadValidationError };
+    }
+  
+    return {};
```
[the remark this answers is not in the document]
JoshBinns2000: I decided to make the function return either the errors or `null`

ChristianMcc (commented):
Couple of small linting issues but apart from that LGTM. Nice refactor, also prefer the way the error generating function is working

ChristianMcc (approved):

PR #2753, opened 2024-02-07
feat(FN-1856): run sql migrations in deployment pipelines

toddym42:
## Ticket 🎫 
[FN-1856](https://ukef-dtfs.atlassian.net/browse/FN-1856)

## Introduction :pencil2:
Adds running of SQL migrations to the deployment pipelines.

### DIFF
--- a/.github/workflows/deployment.yml
+++ b/.github/workflows/deployment.yml
@@ -8,7 +8,11 @@
 # Sets environment variables for product, environment, and timezone.
 # Outputs these variables for use in subsequent jobs.
 #
-# 2. WebApp
+# 2. SQL Migrations
+# *****************
+# Runs new SQL migrations on the environment SQL database
+#
+# 3. WebApp
 # *********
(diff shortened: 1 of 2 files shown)


.github/workflows/deployment.yml
```diff
@@ -286,3 +289,34 @@ jobs:
             az functionapp deployment slot delete \
             --slot ${{ github.sha }} \
             --name ${{ env.FUNCTION }}
+
+  sql-migrations:
```
ChristianMcc, line 293: Discussed in person:
- move migrations to before code deployment
- on deployment script look at making it an approval needed skippable action in the pipeline

ChristianMcc (approved):
Nice, let's merge this. Remind us to set up the secrets for our feature env once we swap that over to SQL

.github/workflows/deployment.yml
toddym42: As discussed, we'll hold off adding a step specific approval step for now.

PR #2754, opened 2024-02-08
feat(FN-1859): add type-safe entity create function

toddym42:
## Introduction :pencil2:
Some refactoring around file and class names, but the main change is to add a `createNotReceived` static factory method to the `UtilisationReportEntity` to get some additional type-safety when creating new entities. This is the pattern we intend to use going forward for creation on new entities. By default TypeORM uses `DeepPartial` on their `create` methods, which means that all properties are optional and we rely on runtime validation alone to ensure the values passed are valid.


libs/common/src/sql-db-entities/helpers/user-id.helper.ts
```diff
@@ -0,0 +1,8 @@
+type RequestSource = { platform: 'PORTAL' | 'TFM'; userId: string } | { platform: 'SYSTEM' };
+
+export type DbRequestSourceParam = {
+  requestSource: RequestSource;
+};
+
+export const getDbUpdatedByUserId = (requestSource: RequestSource) =>
```
JoshBinns2000, line 7: Could we maybe create a type for `UpdatedByUserId` for the sake of some documentation? It already took me a minute to figure out why this will be different to `uploadedByUserId`, and I think we can extract the type from `RequestSource`, although I also think it would involve some non-trivial typescript. I don't feel too strongly about what kind of documentation we have for it, but I do think it may benefit from some

libs/common/src/sql-db-entities/partial-entities/month-and-year.partial-entity.ts
```diff
@@ -1,6 +1,6 @@
 import { Column } from 'typeorm';
 
-export class MonthAndYearPartial {
+export class MonthAndYearPartialEntity {
```
JoshBinns2000, line 3: How about we keep these as `Partial` instead of `PartialEntity` and make the directory `partials`? Unlike `base-entities`, these do not exist as their own entities, so I think it may be better to remove the `Entity` label from them. What do you think?

libs/common/src/sql-db-entities/utilisation-report/utilisation-report.types.ts
```diff
@@ -0,0 +1,7 @@
+import { ReportPeriodPartialEntity } from '../partial-entities';
+import { DbRequestSourceParam } from '../helpers';
+
+export type CreateNotReceivedUtilisationReportEntity = DbRequestSourceParam & {
```
JoshBinns2000, line 4: How about a name like `CreateNotReceivedUtilisationReportParams`?

libs/common/src/sql-db-entities/helpers/user-id.helper.ts
toddym42: Have add a new `DbAuditUpdatedByUserId` type. I could have gone further and add a Platform type like this
```
type Platform = {
  PORTAL: 'PORTAL',
  TFM: 'TFM',
  SYSTEM: 'SYSTEM',
}
```
and used that where required, but I think that overcomplicated things and obscures the intention. If any of the values between `DbAuditUpdatedByUserId`, `RequestSource` and `getDbAuditUpdatedByUserId` go our of sync then a type error is thrown, so I think we're safe.

Have also renamed the function to `getDbAuditUpdatedByUserId` and added a JSDoc comment to hopefully make the use clearer.

Let me know what you think.

libs/common/src/sql-db-entities/partial-entities/month-and-year.partial-entity.ts
toddym42: Not sure. I think you could make the same argument for base entities, as they should not exist on their own either, but are always extended.

libs/common/src/sql-db-entities/utilisation-report/utilisation-report.types.ts
toddym42: Agreed, that's clearer 👍

libs/common/src/sql-db-entities/helpers/user-id.helper.ts
JoshBinns2000: You could just define the `Platform` type with
```typescript
type Platform = RequestSource['platform'];
```
and get all three of the platforms if you wanted to, but that would still mean your `DbAuditUpdatedByUserId` type would end up looking like
```typescript
type DbAuditUpdatedByUserId = {
    [P in Platform]: P extends 'SYSTEM' ? P : `${P}-${UserId}`;
}[Platform];
```
which I think it too complex for this project. I'm happy with the current implementation you have come up with, so feel free to resolve this comment :)

JoshBinns2000 (approved):
LGTM :)

PR #2755, opened 2024-02-09
feat(FN-1951): refactor due reports

francescastocco:
## Introduction :pencil2:
Refactor due reports & previous reports page to account for `REPORT_NOT_RECEIVED` status


e2e-tests/support/alias-selector.js
JoshBinns2000, line 1: A branch containing this fix should be merged into `main` soon - I'll let you know when I rebase, but thank you for fixing the formatting :)

e2e-tests/tfm/cypress/fixtures/create-mock-utilisation-report-details.js
JoshBinns2000, line 1: This file shouldn't be here anymore - I think it should have been deleted but was kept in by mistake during a rebase :)

e2e-tests/tfm/cypress/support/central-api/getAllBanks.js
JoshBinns2000, line 1: This file should have also been deleted (rebase mistake) :)

e2e-tests/tfm/cypress/support/commands.js
```diff
@@ -24,6 +24,7 @@ Cypress.Commands.add('deleteFacility', require('./central-api/deleteFacility'));
 Cypress.Commands.add('deleteTfmDeals', require('./central-api/deleteTfmDeals'));
 
 Cypress.Commands.add('updateTFMDeal', require('./central-api/updateTFMDeal'));
+Cypress.Commands.add('getAllBanks', require('./central-api/getAllBanks'));
```
JoshBinns2000, line 27: This line should be removed (rebase mistake)

portal-api/api-tests/v1/utilisation-report-service/due-report-dates.api-test.js
```diff
@@ -56,6 +41,7 @@ describe('GET /v1/banks/:bankId/due-report-dates', () => {
       dateUploaded,
       uploadedBy: aPaymentReportOfficer,
       path: 'www.abc.com',
+      status: 'REPORT_NOT_RECEIVED',
```
JoshBinns2000, line 44: As this file is `.js`, we should probably use the constant (or, if you're feeling up for it, converting the file to typescript would be even better)

portal-api/src/scheduler/jobs/helpers/utilisation-report-helpers.js
```diff
@@ -78,7 +78,7 @@ const getFormattedReportPeriod = () => {
  */
 const getIsReportSubmitted = async (bank) => {
   const reportPeriod = getReportPeriod();
-  const reportsResponse = await api.getUtilisationReports(bank.id, reportPeriod);
+  const reportsResponse = await api.getUtilisationReports({bankId: bank.id, reportPeriod});
```
JoshBinns2000, line 81: Missing spaces in input object

portal-api/src/v1/api.js
```diff
@@ -1,5 +1,10 @@
 const axios = require('axios');
-const { isValidMongoId, isValidBankId, isValidReportPeriod } = require('./validation/validateIds');
+const { 
+  isValidMongoId, 
+  isValidBankId, 
+  isValidReportPeriod, 
+  areValidReportStatuses
```
JoshBinns2000, line 6: Missing comma

portal-api/src/v1/api.js
```diff
@@ -267,16 +272,23 @@ const saveUtilisationReport = async (reportData, reportPeriod, user, fileInfo) =
...
- * reports are ordered by year and month ascending.
- * @param {string} bankId
- * @param {import('../types/utilisation-reports').ReportPeriod} [reportPeriod]
+ * submitted for that report period is returned. If an array of report 
+ * statuses are provided, the reports returned are filtered by status.
+ * Returned reports are ordered by year and month ascending.
+ * @param {ReportDetails}
  */
-const getUtilisationReports = async (bankId, reportPeriod) => {
+const getUtilisationReports = async ({ bankId, reportPeriod, reportStatuses }) => {
```
JoshBinns2000, line 291: I think this function should be defined as follows (using typescript so that I can include type information there instead):
```typescript
const getUtilisationReports = async (bankId: string, reportFilterOpts: ReportFilterOptions) => { ... };
```
The main reason I think this is that the `bankId` parameter is always required, which makes it distinct from the other two parameters which are optional (notice the introduction of square brackets in the JSDocs typedef to highlight this). Given this is the case, I think this section of the code can be rewritten like this:
```javascript
/**
* @typedef {Object} ReportFilterOptions
* @property {import('../types/utilisation-reports').ReportPeriod} [reportPeriod]
* @property {string[]} [reportStatuses]
*/

/**
 * Gets utilisation reports for a specific bank. If a report
 * period or statuses are not provided, all reports for that bank are
 * returned. If a report period is provided, the report
 * submitted for that report period is returned. If an array of report 
 * statuses are provided, the reports returned are filtered by status.
 * Returned reports are ordered by year and month ascending.
 * @param {string} bankId
 * @param {ReportFilterOptions}
 */
const getUtilisationReports = async (bankId, { reportPeriod, reportStatuses }) => {
```
What do you think? I also think it may be worthwhile to rename the function to document that the primary thing we are filtering by is the bank id, but I also don't think it's necessary given the docstring explanation 🤷

portal-api/src/v1/api.js
```diff
@@ -287,7 +299,12 @@ const getUtilisationReports = async (bankId, reportPeriod) => {
       console.error('Get utilisation reports failed with the following report period: %s', reportPeriod);
       throw new Error('Invalid report period provided: %s', reportPeriod);
     }
-    const params = reportPeriod ? reportPeriod.start : undefined;
+
+    if (reportStatuses?.length > 0 && !areValidReportStatuses(reportStatuses)) {
```
JoshBinns2000, line 303: I think `areValidReportStatuses` is a bit unnecessary of a function. I'd probably go with the individual `isValidStatus` function and then do `reportStatuses.every(isValidStatus)`

JoshBinns2000 (commented):
Having finished reviewing yet, but thought I'd submit this first round of comments now :)

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-reports.controller/helpers.test.ts
```diff
@@ -0,0 +1,43 @@
...
+  describe('parseReportPeriod', () => {
+    it('returns undefined if the input is undefined', () => {
+      // Act
+      const result = parseReportPeriod(undefined);
+
+      // Assert
+      expect(result).toBeUndefined();
+    });
+
+    it('returns undefined if the parsed report period is not a valid report period', () => {
```
JoshBinns2000, line 13: I think the expected behaviour of this should be:
- if `reportPeriod` is `undefined`, return `undefined`
- if `reportPeriod` is defined but cannot be parsed OR `JSON.parse` fails, it should throw an error

I am planning on adding this change, but won't push until other reviews are finished

portal/server/helpers/formatReportPeriodToString.ts
```diff
@@ -0,0 +1,16 @@
+import { format } from 'date-fns';
+import { ReportPeriod } from "../types/utilisation-reports";
+
+export const formatReportPeriodToString = (reportPeriod: ReportPeriod): string => {
```
JoshBinns2000, line 4: This function needs some documentation

portal/templates/utilisation-report-service/utilisation-report-upload/utilisation-report-upload.njk
```diff
@@ -19,52 +19,58 @@
...
     </h1>
-    {% if dueReportDates.length === 0%}
+    {% if dueReportPeriods.length === 0 %}
         <p class="govuk-body" data-cy="next-due-report-text">
             The {{ nextReportPeriod }} report can be uploaded from {{ nextReportPeriodSubmissionStart }}.
         </p>
+        {% if lastUploadedReportPeriod === undefined %}
+        <p class="govuk-body" data-cy="no-reports">
+            No reports have been submitted.
+        </p>
```
francescastocco, line 35: I think I added this to handle when no reports had been submitted and none were due but if that throws `problem with service` we probably don't need this anymore

portal/server/helpers/getFormattedReportPeriodSpan.ts
```diff
@@ -0,0 +1,45 @@
...
+ * console.log(formattedReportPeriodSpan); // December 2023 to January 2024
+ */
+export const getFormattedReportPeriodSpan = (reportPeriod: ReportPeriod): string => {
+  const startOfReportPeriod = new Date(reportPeriod.start.year, reportPeriod.start.month - 1);
+  const endOfReportPeriod = new Date(reportPeriod.end.year, reportPeriod.end.month - 1);
+  
+  const formattedEndOfPeriod = format(endOfReportPeriod, 'MMMM yyyy');
+  if (reportPeriod.start.year === reportPeriod.end.year && reportPeriod.start.month === reportPeriod.end.month) {
+    return formattedEndOfPeriod;
+  }
```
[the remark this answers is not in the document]
JoshBinns2000: As I wrote this line, I thought the same thing, but decided against it because I didn't want to further bloat this PR. I will add it in though

portal/server/helpers/getFormattedReportPeriodSpan.ts
```diff
@@ -0,0 +1,45 @@
...
+ * @example
+ * const reportPeriod = {
+ *   start: { month: 12, year: 2023 },
+ *   end: { month: 1, year: 2024 },
+ * };
+ *
+ * const formattedReportPeriodSpan = getFormattedReportPeriodSpan(reportPeriod);
+ * console.log(formattedReportPeriodSpan); // December 2023 to January 2024
+ */
+export const getFormattedReportPeriodSpan = (reportPeriod: ReportPeriod): string => {
```
francescastocco, line 33: I've thought a bit more about naming this and I think `period` and `span` mean the same thing, essentially `the length of time` so I think we can get away with `getFormattedReportPeriod` but understand if you think that's harder for someone new to follow

portal-api/src/v1/controllers/utilisation-report-service/utilisation-report-upload.controller.js
```diff
@@ -108,12 +108,6 @@ const uploadReportAndSendNotification = async (req, res) => {
       return res.status(400).send();
     }
 
-    // If a report for this month/year/bank combo already exists we should not overwrite it
-    const existingReports = await api.getUtilisationReports(parsedUser?.bank?.id, parsedReportPeriod);
-    if (existingReports.length > 0) {
-      return res.status(409).send('Report already exists');
-    }
```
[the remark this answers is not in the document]
JoshBinns2000: It seems like the check should be made at the `portal` level rather than the `portal-api` level, but at this point I don't mind the redundancy of having another level of checking here as well so I've added it back (with updated logic)

dtfs-central-api/src/services/repositories/utilisation-reports-repo/utilisation-reports-repo.ts
```diff
@@ -32,15 +32,15 @@ export const saveUtilisationReportDetails = async (
   };
 
   const utilisationReportDetailsCollection = await db.getCollection(DB_COLLECTIONS.UTILISATION_REPORTS);
-  const updatedResult = await utilisationReportDetailsCollection.updateOne(
```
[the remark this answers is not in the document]
JoshBinns2000: I think a better name would be something like `updateUtilisationReportDetailsWithUploadDetails` - thoughts?

Also, we can just pass the fields we want to set rather than the whole object (in fact, it makes more sense to do this as we only want to update the fields related to updating reports rather than overwriting the original properties with the same values!)

dtfs-central-api/src/services/repositories/utilisation-reports-repo/utilisation-reports-repo.ts
```diff
@@ -32,15 +32,15 @@ export const saveUtilisationReportDetails = async (
...
+  await utilisationReportDetailsCollection.updateOne(
     {
       _id: { $eq: reportId },
-      'reportPeriod': { $eq: reportPeriod },
+      reportPeriod: { $eq: reportPeriod },
```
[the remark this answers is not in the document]
JoshBinns2000: Decided to replace this with matching the whole report that we found in the function which calls this method

portal-api/src/v1/api.js
```diff
@@ -287,7 +297,12 @@ const getUtilisationReports = async (bankId, reportPeriod) => {
       console.error('Get utilisation reports failed with the following report period: %s', reportPeriod);
       throw new Error('Invalid report period provided: %s', reportPeriod);
     }
-    const params = reportPeriod ? reportPeriod.start : undefined;
+
+    if (reportStatuses?.length > 0 && !reportStatuses.every(isValidReportStatus)) {
+      console.error('Get utilisation reports failed with the following report statuses: %s', reportStatuses);
+      throw new Error('Invalid report statuses provided: %s', reportStatuses);
+    }
+    const params = { reportPeriod: JSON.stringify(reportPeriod), reportStatuses };
```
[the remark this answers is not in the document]
JoshBinns2000: Bit tricky here - it appears that `axios` does not stringify your query params, but instead sends it as an object. When `express` then receives that `query`, it will parse it to an `object` rather than a JSON string representation of the object. This means that the `parseReportPeriod` helper in `dtfs-central-api` will always fail as the input will be an `object` which, when supplied to `JSON.parse`, is turned in the `string` `[object Object]` which is not valid JSON.

Tl;dr - I can just remove the helper and we don't have to worry about stringified JSON floating around

dtfs-central-api/src/services/repositories/utilisation-reports-repo/utilisation-reports-repo.ts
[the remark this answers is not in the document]
toddym42: Sounds good to me 👍

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-reports.controller/helpers.ts
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * Parses the report period
+ * @param reportPeriod - The report period to parse
+ * @returns The parsed report period
+ * @throws {SyntaxError} If `reportPeriod` is a `string` but `JSON.parse` fails to parse it
+ * @throws If the result of `JSON.parse(reportPeriod)` is not a valid `ReportPeriod` object
+ */
+export const parseReportPeriod = (reportPeriod: string | undefined): ReportPeriod | undefined => {
+  if (!reportPeriod && typeof reportPeriod !== 'string') {
```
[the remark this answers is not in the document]
JoshBinns2000: Implementation has now changed 🙃

dtfs-central-api/src/types/db-models/utilisation-reports.ts
```diff
@@ -40,3 +40,11 @@ export type UtilisationReport = Prettify<
     uploadedBy?: UploadedByUserDetails;
   }>
 >;
+
+export type UtilisationReportUploadDetails = Prettify<
+  Required<
+    Pick<UtilisationReport, 'azureFileInfo' | 'dateUploaded' | 'uploadedBy'> & {
+      status: Extract<UtilisationReportReconciliationStatus, 'PENDING_RECONCILIATION'>;
```
toddym42, line 47: This use of `Extract` doesn't seem to provide any value because the second argument can be anything. I don't get any auto-complete and can type and invalid value without error.

I think we could instead use this?
```
status: typeof UTILISATION_REPORT_RECONCILIATION_STATUS.PENDING_RECONCILIATION;
```

portal-api/api-test.jest.config.js
```diff
@@ -1,8 +1,9 @@
...
 module.exports = {
-  collectCoverageFrom: ['src/**/*.{js,}'],
+  ...commonConfig,
+  collectCoverageFrom: ['src/**/*.{js,ts}'],
   coverageDirectory: 'generated_reports/coverage/api-test',
-  testMatch: ['**/*.api-test.js'],
-  ...commonSettings,
+  testMatch: ['**/*.api-test.{js,ts}'],
+  setupFilesAfterEnv: ['./api-test-setup.jest.config.js'],
 };
```
toddym42, line 9: Do we need to keep
```
testTimeout: 5000,
```
from the original `portal-api/api-test-common.jest.config.js`?

portal-api/jest.common.config.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
+  preset: 'ts-jest/presets/js-with-babel',
+  transform: {
+    '^.+\\.js$': 'babel-jest',
+    '^.+\\.ts$': 'ts-jest',
+  },
+};
```
toddym42, line 7: I _think_ because we use `preset: 'ts-jest/presets/js-with-babel',` here instead of just `preset: 'ts-jest',` then we may be able to remove the explicit `transform`?

Edit: just tried removing `transform` and all the unit tests still pass

portal/server/controllers/utilisation-report-service/utilisation-report-upload/utilisation-report-details.js
```diff
@@ -1,25 +1,26 @@
 const { format, parseISO } = require('date-fns');
+const { getFormattedReportPeriod } = require('../../../helpers');
 
 /**
  * @typedef {Object} ReportAndUserDetails
  * @property {string} uploadedByFullName - The uploaded by users full name with format '{firstname} {surname}'
  * @property {string} formattedDateAndTimeUploaded - The date uploaded formatted as 'd MMMM yyyy at h:mmaaa'
- * @property {string} lastUploadedReportPeriod - The report period of the report formatted as 'MMMM yyyy'
+ * @property {string} lastUploadedReportPeriod - The report period of the report formatted as described in `getFormattedReportPeriod.ts`
```
toddym42, line 8: nit: Can link here using
```
{@link getFormattedReportPeriod}
```

dtfs-central-api/src/services/repositories/utilisation-reports-repo/utilisation-reports-repo.ts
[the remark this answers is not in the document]
toddym42: This seems a bit overkill as the `_id` is already unique to the report, so this will just slow down the query.

I don't think we need to worry about the case where the report has been updated since we originally fetched it, but if for some reason it didn't match the full report I think it would just fail silently anyway? If we did want to think about this case we'd probably take a different approach, like locking the row while we update.

portal/server/controllers/utilisation-report-service/utilisation-report-upload/utilisation-report-details.js
JoshBinns2000: I tried that earlier and it didn't work - maybe my VSCode was just playing up!

dtfs-central-api/src/types/db-models/utilisation-reports.ts
JoshBinns2000: That's strange because I did get autocomplete from using `Extract` _and_ I couldn't type any value that wasn't `'PENDING_RECONCILIATION'` for this field

dtfs-central-api/src/services/repositories/utilisation-reports-repo/utilisation-reports-repo.ts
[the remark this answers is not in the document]
JoshBinns2000: A question for when we refactor to SQL I think 🙃 - I did replace the filter with the original "id only" check though

portal/server/controllers/utilisation-report-service/utilisation-report-upload/utilisation-report-details.js
JoshBinns2000: Tried it and it does work 👍

portal-api/api-test.jest.config.js
JoshBinns2000: All the other `testTimeout` values in the other directories are 80000 😱

dtfs-central-api/src/types/db-models/utilisation-reports.ts
JoshBinns2000: Discussed - seeing as we're using the constant everywhere anyway, we may as well keep using it here

toddym42 (approved):
LGTM 🚀

PR #2762, opened 2024-02-14
feat(FN-2239): update deployment scripts to set workspace context

toddym42:
## Introduction :pencil2:
Following the introduction of [npm workspaces](https://github.com/UK-Export-Finance/dtfs2/pull/2596) the context that we run microservice Dockerfiles has been updated to be the root of the project. This is done because Dockerfiles now pull in shared code from the `libs` directory, with everything being installed using root `package.json` npm workspace scripts.

Update the GitHub Actions `deployment.yml` and `feature_deploy.yml` scripts to update the run context, as mentioned above.

### DIFF
--- a/.github/workflows/deployment.yml
+++ b/.github/workflows/deployment.yml
@@ -11,7 +11,8 @@
 # 2. WebApp
 # *********
 # Runs sequentially for each web app defined in the matrix.webapp list.
-# Builds and pushes Docker images for each web app using the portal directory for portal-ui and the app's name for others.
+# Sets the Dockerfile location based on the microservice (using the portal directory for portal-ui and the app's name for others).
+# Builds and pushes Docker images using the root workspace as the context and the Dockerfile location specified for the microservice.
 # Creates a temporary deployment slot using the FROM tag (latest by default).
 # Swaps the slot to make the new image live.
 # Cleans up the temporary slot.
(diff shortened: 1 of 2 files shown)


.github/workflows/feature_deploy.yml
```diff
@@ -70,7 +70,8 @@ jobs:
           username: ${{ env.acr_username }}
           password: ${{ env.acr_password }}
           repository: ${{ matrix.repository }}
-          path: ${{ matrix.path || matrix.repository }}
+          context: .
+          dockerfile: ${{ matrix.path || matrix.repository }}/Dockerfile
```
toddym42, line 74: This is the main change, with everything else just being auto-formatting.

* use `path` if you want to change both the location of the Dockerfile and context to run it in
* use `context` and `dockerfile` if you want to set the context to run to a different location to the Dockerfile

A successful deployment to the feature environment with this change can be seen [here](https://github.com/UK-Export-Finance/dtfs2/actions/runs/7901475398).

ChristianMcc: what's the difference between matrix.path and matrix.repository? when is one not defined and we need to fallback to the other?

ChristianMcc (approved):
LGTM although have a question on matrix.path vs matrix.repository. I think separate to this we could look at commonising the build and push approach

.github/workflows/feature_deploy.yml
toddym42: I think this is to handle the case where the `portal-ui` project (name in package.json) is different from the directory name of `portal`. The path for this is specified on lines 59-61 above:
```
include:
  - path: portal
     repository: portal-ui
```

ChristianMcc: ahh gotcha, makes sense

.github/workflows/feature_deploy.yml
```diff
@@ -2,18 +2,18 @@ name: Feature environment deployment
 
 on:
```
abhi-markan, line 3: Can we remove this file please, as we should only have one `deployment.yml` to deploy to `feature` environment.

.github/workflows/deployment.yml
```diff
@@ -134,19 +135,19 @@ jobs:
           username: ${{ env.ACR_USER }}
           password: ${{ secrets.ACR_PASSWORD }}
 
-      - name: Directory 📁
+      - name: Dockerfile Location 🐳
         if: ${{ 'portal-ui' == matrix.webapp }}
-        run: echo "DIRECTORY=portal" >> $GITHUB_ENV
+        run: echo "DOCKERFILE_LOCATION=portal/Dockerfile" >> $GITHUB_ENV
```
abhi-markan, line 140: Just a suggestion (Not a blocker), `DOCKERFILE` should do it.

abhi-markan (changes requested):
Nice work! 2 comments to address please.

.github/workflows/feature_deploy.yml
ChristianMcc: Can we leave it in for now? There's quite a few things in here that mean it being separate is much easier to maintain - will list a few below:

1- we don't have a feature branch, we use feature branches to deploy to feature so there's no mapping between branch and environment
2- it uses dev env variables so there's a slight mismatch there that is hard-coded in that file
3- it wants to be deployed automatically on push to a branch rather than post QAT, and while we could add that to the deployment.yml file I'd feel safer if people are editing the feature_deploy.yml file for now so that they can't affect any dev/staging/prod related deployment logic
4- We don't deploy the functionapps right now on the feature environment but wouldn't want that logic in deployment.yml

If we do need to get rid of it I'd suggest we raise a ticket on the E2ETF board as it's not just a simple task

abhi-markan (approved):

PR #2764, opened 2024-02-15
feat(FN-1953): refactor portal-api scheduled emails

JoshBinns2000:
## Introduction :pencil2:
We want to change the way that `portal-api` checks for existing reports before sending chaser emails.

- Updates the `getIsReportSubmitted` function to only look for reports in the `'REPORT_NOT_RECEIVED'` state
- Updates the unit tests to make sure the `api` method is called with the correct values


portal-api/src/scheduler/jobs/helpers/utilisation-report-helpers.js
```diff
@@ -78,7 +79,10 @@ const getFormattedReportPeriod = () => {
  */
 const getIsReportSubmitted = async (bank) => {
   const reportPeriod = getReportPeriod();
-  const reportsResponse = await api.getUtilisationReports(bank.id, { reportPeriod });
+  const nonNotReceivedStatuses = Object.values(UTILISATION_REPORT_RECONCILIATION_STATUS).filter(
+    (status) => status !== UTILISATION_REPORT_RECONCILIATION_STATUS.REPORT_NOT_RECEIVED,
+  );
```
toddym42, line 84: We've needed this set of statuses quite a few times now, so probably makes sense to add a `getUtilisationReportReceivedStatuses` helper, maybe in the same place as the `UTILISATION_REPORT_RECONCILIATION_STATUS` constant.

toddym42 (approved):

portal-api/src/scheduler/jobs/helpers/utilisation-report-helpers.js
JoshBinns2000: I was hoping to add that as a helper in `common` which I want to do in a separate PR after this branch is merged (I think it would be nice to have a commit at the end of this refactor that commonises a lot of the new things we've added)

JoshBinns2000: Also, all statuses which are not `REPORT_NOT_RECEIVED` do not necessarily correspond to a report being received, hence the name `nonNotReceivedStatuses` (unless we decide that 'Received' and 'Uploaded' mean different things)

toddym42: Yeah, I'd thought about that, but then to me `nonNotReceivedStatuses` means the same thing 🤷‍♂️

JoshBinns2000: The other way I can think of writing this to make it clear is to make an `isReportUploaded` filter like
```typescript
const isReportUploaded = (report: UtilisationReport) =>
  report.status !== 'REPORT_NOT_RECEIVED' && !!report.azureFileInfo;
```
and to use it as follows:
```typescript
const uploadedReports = (
  await api.getUtilisationReports(bank.id, { reportPeriod }),
).filter(isReportUploaded);
```
What do you think?

toddym42: Yeah, that looks good to me. Much clearer what the criteria is and more concise. Nice 👍

PR #2768, opened 2024-02-16
feat(FN-1951): refactor query options for utilisation reports endpoint

JoshBinns2000:
## Introduction :pencil2:
The `GET /v1/banks/:bankId/utilisation-reports` endpoint originally accepted `reportStatuses: UtilisationReportReconciliationStatus[]` as one of its query parameters. This was mostly used to get reports which have any status which isn't `REPORT_NOT_RECEIVED` which, by extension, would return reports which have been uploaded. However, this has two main issues:
- As the number of report statuses grows, this query string for this array would grow with it which could eventually causes issues with the length of the url
- A report having a status which isn't in the `REPORT_NOT_RECEIVED` state does not necessarily imply that the report was uploaded due to the "mark report as completed" behaviour in TFM

Replaces this query option with a simple `excludeNotReceived` flag which:
- is either `'true'` or `'false'` (the type is `string` because of the default way that express parses query parameters)
- if set to `'true'`, excludes reports which are in a `REPORT_NOT_RECEIVED` state _or_ have a `null` `azureFileInfo`
- if not set or set to `'false'`, does nothing

## Miscellaneous 👀 
Refactor `previous-reports` to only show links for reports which have been uploaded.


dtfs-central-api/src/services/repositories/utilisation-reports-repo/utilisation-reports-repo.ts
```diff
@@ -56,7 +56,7 @@ export const saveNotReceivedUtilisationReport = async (reportPeriod: ReportPerio
 
 export type GetUtilisationReportDetailsOptions = {
   reportPeriod?: UtilisationReport['reportPeriod'];
-  reportStatuses?: UtilisationReport['status'][];
+  excludeNotUploaded?: 'true' | 'false';
```
toddym42, line 59: Can this be a boolean? Was this because it's a query param? If so then it should be transformed from a string to a bool when passed to the repo as this detail should be abstracted away from the repo. If we do convert then it's probably worth creating a helper in common to do this conversion.

dtfs-central-api/src/services/repositories/utilisation-reports-repo/utilisation-reports-repo.ts
```diff
@@ -73,8 +73,9 @@ const getUtilisationReportDetailsFilterFromOptions = (options?: GetUtilisationRe
   if (options.reportPeriod) {
     utilisationReportDetailsFilter.reportPeriod = { $eq: options.reportPeriod };
   }
-  if (options.reportStatuses) {
-    utilisationReportDetailsFilter.status = { $in: options.reportStatuses };
+  if (options.excludeNotUploaded && options.excludeNotUploaded === 'true') {
```
toddym42, line 76: I think this can just be
```
if (options?.excludeNotUploaded === 'true') {
```
or
```
if (options?.excludeNotUploaded) {
```
if we convert the param to a boolean.

portal-api/src/v1/api.js
```diff
@@ -270,7 +270,7 @@ const saveUtilisationReport = async (reportData, reportPeriod, user, fileInfo) =
 /**
  * @typedef {Object} GetUtilisationReportsOptions
  * @property {import('../types/utilisation-reports').ReportPeriod} [reportPeriod] - a report period to filter reports by
- * @property {import('../types/utilisation-reports').UtilisationReportReconciliationStatus[]} [reportStatuses] - list of report statuses to filter reports by
+ * @property {('true' | 'false')} [excludeNotUploaded] - whether or not to exclude reports which have not been uploaded
```
toddym42, line 273: Any reason not to use a `boolean` here?

portal-api/src/v1/controllers/utilisation-report-service/due-report-periods.controller.js
```diff
@@ -9,7 +9,9 @@ const getDueReportPeriodsByBankId = async (req, res) => {
   try {
     const { bankId } = req.params;
 
-    const notReceivedReports = await api.getUtilisationReports(bankId, { reportStatuses: [UTILISATION_REPORT_RECONCILIATION_STATUS.REPORT_NOT_RECEIVED] });
+    const notReceivedReports = (await api.getUtilisationReports(bankId)).filter(
+      (report) => report.status === UTILISATION_REPORT_RECONCILIATION_STATUS.REPORT_NOT_RECEIVED,
+    );
```
toddym42, line 14: Should we be using the `excludeNotUploaded` param here?

toddym42 (commented):
Few comments around use of a string boolean instead of a boolean. I think we should be using a normal boolean where possible, and immediately converting from a string boolean to a normal boolean as soon as possible after extracting the value from query params.

portal-api/src/v1/api.js
JoshBinns2000: I suppose not. I think my reasoning was that whenever we send it off in a request, it will become the `string` `'true'` or `'false'`, but seeing as this will be used to construct a request there is no good reason to not use the actual `boolean` values

portal-api/src/v1/controllers/utilisation-report-service/due-report-periods.controller.js
toddym42: Wouldn't that case be handled by this filter?
```
utilisationReportDetailsFilter.azureFileInfo = { $not: { $eq: null } };
```

portal-api/src/v1/api.js
toddym42: Yeah, string values of `'true'` and `'false'` is a limitation of query params, but we shouldn't use the stringified values anywhere other than where we first extract the values from `req.query`.

dtfs-central-api/src/services/repositories/utilisation-reports-repo/utilisation-reports-repo.ts
JoshBinns2000: I originally had this as a `boolean`, but changed it back because the `options` object when calling this function looked like
```typescript
const { excludeNotUploaded } = req.query; // undefined, true or false
...
{
  excludeNotUploaded: excludeNotUploaded && excludeNotUploaded === 'true',
}
```

and I didn't like it for some reason, but I am happy to switch it back to this

dtfs-central-api/src/services/repositories/utilisation-reports-repo/utilisation-reports-repo.ts
JoshBinns2000: We don't need the optional chaining as the function already checks whether or not `options` is defined, but yes :)

dtfs-central-api/src/services/repositories/utilisation-reports-repo/utilisation-reports-repo.ts
toddym42: I think it can just be
```
excludeNotUploaded: excludeNotUploaded === 'true',
```
?

dtfs-central-api/src/services/repositories/utilisation-reports-repo/utilisation-reports-repo.ts
```diff
@@ -73,8 +73,9 @@ const getUtilisationReportDetailsFilterFromOptions = (options?: GetUtilisationRe
...
+    utilisationReportDetailsFilter.status = { $not: { $in: [ 'REPORT_NOT_RECEIVED' ] } };
+    utilisationReportDetailsFilter.azureFileInfo = { $not: { $eq: null } };
```
toddym42, line 78: `options` missing from the JSDoc in the `getOneUtilisationReportDetailsByBankId` function below.

dtfs-central-api/src/services/repositories/utilisation-reports-repo/utilisation-reports-repo.ts
JoshBinns2000: It could - my reasoning for having the first part is that if it is `undefined` (ie. if it was not in the query), then we pass in `undefined` to the repo method rather than `false` (I know this isn't true right now, but we may decide that passing in `false` does something different to `undefined` which I wanted to explicitly avoid)

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-reports.controller/index.ts
```diff
@@ -10,7 +9,7 @@ export type GetUtilisationReportsRequest = CustomExpressRequest<{
   };
   query: {
     reportPeriod?: Request['query'];
-    reportStatuses?: UtilisationReportReconciliationStatus[];
+    excludeNotUploaded: 'true' | 'false';
```
toddym42, line 12: Should this be optional?

dtfs-central-api/src/services/repositories/utilisation-reports-repo/utilisation-reports-repo.ts
toddym42: I'd argue YAGNI here. Also I'm not sure how `undefined` could be different from `false` when "excluding" something? If we want false to do something different we'd probably want a different name, and a different parameter type.

portal-api/src/v1/controllers/utilisation-report-service/due-report-periods.controller.js
JoshBinns2000: Actually, we shouldn't be using that query here. `excludeNotUploaded` will remove reports in the `REPORT_NOT_RECEIVED` state, but we only want reports in this state

dtfs-central-api/src/services/repositories/utilisation-reports-repo/utilisation-reports-repo.ts
JoshBinns2000: Yeh, I was thinking YAGNI too. It's more the semantic side where the three possibilities are strictly speaking distinct things:
- 1: *do* exclude based on A
- 2: *do not* exclude based on A
- 3: *do nothing*

In this case I think it's fine to have options 2 and 3 mean the same thing, but my brain doesn't like conflating the two by default 😅

toddym42 (approved):
LGTM

PR #2771, opened 2024-02-19
feat(FN-1947): only upload reports for banks which are visible in TFM

JoshBinns2000:
## Introduction :pencil2:
For the job to upload reports, an error was being thrown in `dtfs-central-api` due to reporting schedules not existing for certain banks.

Any bank which has `isVisibleInTfmUtilisationReports` set to `true` should have a defined reporting schedule, so this PR adds filtering to the banks to make sure that only those banks are checked.

### DIFF
--- a/dtfs-central-api/src/scheduler/jobs/create-utilisation-reports/index.ts
+++ b/dtfs-central-api/src/scheduler/jobs/create-utilisation-reports/index.ts
@@ -19,27 +19,28 @@ const isCurrentBankReportMissing = async (bank: Bank): Promise<boolean> => {
 };
 
 /**
- * Gets the banks that do not have reports
+ * Gets the banks which are visible in TFM utilisation
+ * reports and do not have any report in the database
  * @returns The banks which do not have reports
  */
-const getBanksWithoutReports = async () => {
-  const banks = await getAllBanks();
+const getBanksWithMissingReports = async (): Promise<Bank[]> => {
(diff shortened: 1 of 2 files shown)


dtfs-central-api/src/scheduler/jobs/create-utilisation-reports/index.ts
```diff
@@ -19,27 +19,28 @@ const isCurrentBankReportMissing = async (bank: Bank): Promise<boolean> => {
...
 /**
- * Gets the banks that do not have reports
+ * Gets the banks which are visible in TFM utilisation
+ * reports and do not have any report in the database
  * @returns The banks which do not have reports
  */
-const getBanksWithoutReports = async () => {
-  const banks = await getAllBanks();
+const getBanksWithMissingReports = async (): Promise<Bank[]> => {
+  const banksVisibleInTfm = (await getAllBanks()).filter((bank) => bank.isVisibleInTfmUtilisationReports);
```
toddym42, line 27: Are we doing this filter elsewhere? Maybe worth adding a new repo method and doing the filter at the db level rather than in memory.

toddym42 (approved):

dtfs-central-api/src/scheduler/jobs/create-utilisation-reports/index.ts
JoshBinns2000: We currently only do this specific logic in one other place and (as far as memory is concerned) there are a known and small number of banks each with a known size, so I'm happy leaving this as is for now with the caveat that this will probably be picked up as part of the SQL refactor

PR #2772, opened 2024-02-19
feat(FN-2253): add MongoDB to SQL Server data migrator

toddym42:
## Introduction :pencil2:
Adds util to help migrate `utilisationReports` and `utilisationData` MongoDB collections to SQL Server.

See [README](https://github.com/UK-Export-Finance/dtfs2/pull/2772/files#diff-3b4b6af991682fa1c20572dc36c265d44da282dce540576f4a4d3b04b42650e7) for further details.


libs/common/src/types/mongodb-models/index.ts
JoshBinns2000, line 1: I wondered about this in the changes I am adding - seeing as our SQL stuff is denoted with a separator between `sql` and `db`, should we do the same with mongo? Don't particularly mind, but I think it might be a bit cleared to see that these types are for different databases if we had `sql-db-<thing>` and `mongo-db-<thing>`

libs/common/src/types/mongodb-models/session-bank.ts
JoshBinns2000, line 1: I don't think this type belongs in `mongodb-models` as it isn't a collection itself, and before now we have only put types which match collections into `db-models`

libs/common/src/types/types-helper.ts
```diff
@@ -1 +1,7 @@
 export type ValuesOf<T> = T[keyof T];
+
+// https://www.totaltypescript.com/concepts/the-prettify-helper
+export type Prettify<T> = {
```
JoshBinns2000, line 4: The return of the king 🙏

utils/data-migration/sql/utilisation-reports/helpers/ejson.helper.ts
```diff
@@ -0,0 +1,3 @@
+import { Document, EJSON } from 'bson';
+
+export const deserialiseEJson = <TResult extends object>(ejson: object): TResult => EJSON.deserialize(ejson as Document) as TResult;
```
JoshBinns2000, line 3: I don't think we need it, but a `zod` validator here might be nice

JoshBinns2000 (commented):
Looks cool :)

ChristianMcc (approved):
LGTM, could we double check this locally with some example data in a local mongodb to a local sql db if we haven't already please (should be fine running outside avd as it's just connection strings and the code here)

PR #2773, opened 2024-02-19
feat(FN-2146): add error logging to cron scheduler. commonise logic

toddym42:
## Introduction :pencil2:
Currently if any errors are thrown during running of node-cron scheduled jobs the job will just fail silently. No error messages appear in the logs and the server carries on as if nothing has happened.

The following error handling is already in place:
```
cron.schedule(schedule, task).on('error', (error) => {
  console.error(`An error occurred running job ${message}:`, error);
});
```
but this appears to only catch errors thrown when the job is initially scheduled, not when the task is run.

* add error logging for errors thrown when the tasks are run - this is done centrally with a `taskWithErrorLogging` so that you don't need to remember to do this when creating any new scheduled job task
* commonise node-cron scheduler logic and types into the new `@ukef/dtfs2-common` package.


libs/common/src/cron-scheduler/scheduler.ts
```diff
@@ -0,0 +1,19 @@
+import cron from 'node-cron';
+import { SchedulerJob } from './scheduler.types';
+import { asyncTaskToSyncTask, taskWithErrorLogging } from './scheduler.helpers';
+
+export const initialiseCronScheduler = (jobs: SchedulerJob[]) => {
```
ChristianMcc, line 5: docstring needed

trade-finance-manager-api/src/generateApp.js
```diff
@@ -6,14 +6,14 @@ const mongoSanitise = require('express-mongo-sanitize');
...
-const initScheduler = require('./scheduler');
+const { initialiseScheduler } = require('./scheduler');
 const seo = require('./v1/middleware/headers/seo');
 const security = require('./v1/middleware/headers/security');
 const removeCsrfToken = require('./v1/middleware/remove-csrf-token');
 const configurePassport = require('./v1/controllers/user/passport');
 const createRateLimit = require('./v1/middleware/rateLimit/index');
 
-initScheduler();
+initialiseScheduler();
```
ChristianMcc, line 16: `initialiseNodeCronJobs`?

ChristianMcc (approved):
LGTM :)

PR #2775, opened 2024-02-20
chore: update 'no-underscore-dangle' eslint rule exceptions

toddym42:
## Introduction :pencil2:
Update the eslint [no-underscore-dangle](https://eslint.org/docs/latest/rules/no-underscore-dangle) rule to allow safe functions used by the [node-mocks-http](https://github.com/eugef/node-mocks-http) package so we don't need to suppress manually in tests.

Also removes a couple of redundant suppressions in other files.


ChristianMcc (approved):
LGTM :)

PR #2777, opened 2024-02-20
feat(FN-1865): move mongo-db types and client to common

JoshBinns2000:
## Introduction :pencil2:
In order to get the seeder for `typeorm` to work, we will need to directly extract items from the `mongo` database. Rather than rewriting the `db-client.ts` file again, it would be good to commonise the logic around the `mongo` db client.

- Adds a new `class` for `MongoDbClient` which handles all the `mongo` client related logic in `libs/common`
- Moves the `types` related to `db-models` from `dtfs-central-api` to `libs/common`


external-api/src/cronJobs/eStoreFacilityFolderCreationJob.cron.ts
```diff
@@ -12,7 +12,7 @@ export const eStoreFacilityFolderCreationJob = async (eStoreData: Estore) => {
     const response = await cronJobLogsCollection.findOneAndUpdate(
       { dealId: { $eq: eStoreData.dealId } },
       { $inc: { facilityFolderRetries: 1 } },
-      { returnNewDocument: true, returnDocument: 'after' },
+      { returnDocument: 'after' },
```
JoshBinns2000, line 15: The `returnNewDocument` option does not exist in `mongodb 4.17.0` and typescript was throwing some errors with it, so I have removed it

libs/common/src/types/mongo-db-models/cron-job-logs.ts
JoshBinns2000, line 1: This file was a bit of a hack to make the `tsc --noemit` step in `external-api` happy

utils/mock-data-loader/.eslintrc.js
JoshBinns2000, line 1: I copied over this and the `tsconfig.eslint.json` file from `dtfs-central-api`

utils/mock-data-loader/re-insert-mocks.js
```diff
@@ -33,7 +33,7 @@ const init = async () => {
   await deleteInitialTFMUser(tfmToken);
   await deleteInitialUser(portalToken);
 
-  db.close();
+  await db.close();
```
JoshBinns2000, line 36: Mildly concerning that this wasn't originally marked with `await`, but because there were no further executions after `init()` this task will have been executed before `node re-insert-mocks.js` exited

libs/common/src/types/mongo-db-models/cron-job-logs.ts
JoshBinns2000, line 1: This file was a bit of a hack to make the `tsc --noemit` step in `external-api` happy

utils/mock-data-loader/.eslintrc.js
JoshBinns2000, line 1: I copied over this and the `tsconfig.eslint.json` file from `dtfs-central-api`

libs/common/src/types/mongo-db-models/cron-job-logs.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * This type is not well defined in external-api,
+ * but causes type errors in external-api where
+ * `eStore` related files use the db client
+ */
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type CronJobLog = any;
```
JoshBinns2000, line 8: This type is a bit of a hack to fix some `tsc --noemit` errors in `external-api`

utils/mock-data-loader/.eslintrc.js
```diff
@@ -1,34 +1,61 @@
+const baseParserOptions = {
```
JoshBinns2000, line 1: I copied this and the `tsconfig.eslint.json` file from `dtfs-central-api`

libs/common/src/types/azure-file-info.ts
```diff
@@ -0,0 +1,4 @@
+import { AzureFileInfoEntity } from '../sql-db-entities/azure-file-info';
+import { AuditableBaseEntity } from '../sql-db-entities/base-entities';
+
+export type AzureFileInfo = Omit<AzureFileInfoEntity, keyof AuditableBaseEntity | 'id' | 'utilisationReport'>;
```
ChristianMcc, line 4: can we have a quick chat about this offline for me to check if this is what we might be trying going forwards etc.

libs/common/src/types/mongo-db-models/cron-job-logs.ts
ChristianMcc: This might affect an eStore API PR in review - so just mentioning when we merge main into SQL we might need to revisit. Happy to keep just like this for now though :) explanation makes sense!

libs/common/src/types/mongo-db-models/db-model.ts
```diff
@@ -0,0 +1,25 @@
...
+  ? TfmUser
+  : TCollectionName extends 'tfm-teams'
+  ? TfmTeam
+  : TCollectionName extends 'cron-job-logs'
+  ? CronJobLog
+  : TCollectionName extends 'utilisationReports' // TODO FN-1853 Remove after SQL refactor
+  ? UtilisationReport
+  : TCollectionName extends 'utilisationData'
+  ? UtilisationData
+  : WithId<object>;
```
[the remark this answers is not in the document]
ChristianMcc: will grab you offline for this too if that's alright!

libs/common/src/mongo-db-client/index.ts
```diff
@@ -0,0 +1,71 @@
+import { Collection, Db as DbConnection, MongoClient, WithoutId } from 'mongodb';
```
ChristianMcc, line 1: can you chuck some docstrings on the functions in this file please?

libs/common/src/mongo-db-client/index.ts
```diff
@@ -0,0 +1,71 @@
...
+  private mongoDbClientConnection: MongoDbClientConnection = {
+    isInitialised: false,
+  };
+
+  constructor({ dbName, dbConnectionString }: ConnectionOptions) {
+    this.dbName = dbName;
+    this.dbConnectionString = dbConnectionString;
+  }
+
+  private async initClient(): Promise<MongoDbClientConnection & { isInitialised: true }> {
```
ChristianMcc, line 34: should this be like getOrInitialiseClient()? Trying to think how for rest of file we can show if the client exists we get it otherwise we initialise?

external-api/src/database/mongo-client.ts
```diff
@@ -1,34 +1,11 @@
-const { MongoClient } = require('mongodb'); // eslint-disable-line @typescript-eslint/no-var-requires
+import { MongoDbClient, MongoDbCollectionName } from '@ukef/dtfs2-common';
 import { dbName, url } from '../config';
 
-let client: any;
-let connection: any = null;
+const mongoDbClient = new MongoDbClient({ dbName, dbConnectionString: url });
```
ChristianMcc, line 4: Are we doing this here so that the common package doesn't have to know the .env files? It feels like this whole file is a bit of a duplicate of the common DB and wondering if we can just have the common DB file and get rid of these files in each of the microservices. What are your thoughts?

ChristianMcc (commented):
Nice work, few small changes but mostly just a few questions it would be good to talk through!

dtfs-central-api/src/types/utilisation-reports.ts
```diff
@@ -1,10 +1,10 @@
 import { ObjectId } from 'mongodb';
-import { ValuesOf } from '@ukef/dtfs2-common';
+import { ValuesOf, UtilisationReportReconciliationStatus } from '@ukef/dtfs2-common';
 import { IsoMonthStamp, MonthAndYear } from './date';
-import { UTILISATION_REPORT_RECONCILIATION_STATUS, UTILISATION_REPORT_HEADERS } from '../constants';
+import { UTILISATION_REPORT_HEADERS } from '../constants';
 import { Currency } from './currency';
 
-export type UtilisationReportReconciliationStatus = ValuesOf<typeof UTILISATION_REPORT_RECONCILIATION_STATUS>;
+export { UtilisationReportReconciliationStatus };
```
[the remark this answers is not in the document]
JoshBinns2000: We can just do that, but I wanted to avoid refactoring all of that at once. This PR will enable us to move types in following PRs to directly import from `@ukef/dtfs2-common`, but if I refactored all the imports of the types I moved to common this PR would become significantly bigger

external-api/src/cronJobs/eStoreFacilityFolderCreationJob.cron.ts
JoshBinns2000: I believe it was `5.Y.Z` (can't remember minor version)

external-api/src/database/mongo-client.ts
JoshBinns2000: I originally wanted to do this, but Mike wanted to avoid needing to define the MongoDB connection `.env` vars in common. That being said, I need them anyway for the seeder, so I might centralise the initialisation of the `MongoDbClient`

external-api/src/cronJobs/eStoreFacilityFolderCreationJob.cron.ts
JoshBinns2000: [This link](https://mongodb.github.io/node-mongodb-native/3.6/api/Collection.html#findOneAndUpdate) shows what options are available, and `returnNewDocument` is not there. It may have been in an old version of MongoDB, but even then (as can be seen in the documentation), it seems like the option to return a document should be entirely controlled by the `returnDocument` parameter (even `returnOriginal` is now deprecated!)

dtfs-central-api/src/scheduler/jobs/create-utilisation-reports/index.test.ts
```diff
@@ -12,7 +12,10 @@ console.info = jest.fn();
 jest.mock('../../../repositories/banks-repo');
 jest.mock('../../../repositories/utilisation-reports-repo');
 jest.mock('../../../utils/report-period');
+
+// eslint-disable-next-line @typescript-eslint/no-unsafe-return
```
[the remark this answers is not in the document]
JoshBinns2000: Discussed offline: leaving this for now, but plan is to add a helper function for this kind of mocking

libs/common/src/types/mongo-db-models/db-model.ts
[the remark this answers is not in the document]
JoshBinns2000: Discussed: the decision has been made to replace `object` with `any` because
1. Using something like `Record<string, unknown>` does fix an issue with properties not being defined on `object`, but causes errors if you don't stringently check the type of the property (feature of `unknown`)
2. Some of the types that `any` is covering here are very complicated to document and will likely never have types created for them

dtfs-central-api/api-tests/wipeDB.js
```diff
@@ -1,4 +1,4 @@
-const db = require('../src/drivers/db-client');
+const db = require('../src/drivers/db-client').default;
```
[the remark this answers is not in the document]
JoshBinns2000: For details on why this happens, [read here](https://stackoverflow.com/questions/33704714/cant-require-default-export-value-in-babel-6-x)

dtfs-central-api/src/types/db-models/db-collection-name.ts
```diff
@@ -1,4 +1,3 @@
-import { ValuesOf } from '@ukef/dtfs2-common';
-import { DB_COLLECTIONS } from '../../constants/db-collections';
+import { MongoDbCollectionName } from '@ukef/dtfs2-common';
 
-export type DbCollectionName = ValuesOf<typeof DB_COLLECTIONS>;
+export type DbCollectionName = MongoDbCollectionName;
```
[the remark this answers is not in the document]
JoshBinns2000: This type isn't needed anymore so I have removed this file :)

ChristianMcc (approved):

PR #2782, opened 2024-02-21
feat(FN-1549): refactor utilisation report report period

toddym42:


.github/workflows/feature_deploy.yml
```diff
@@ -2,7 +2,7 @@ name: Feature environment deployment
 
 on:
   push:
-    branches: [main-stbr-tfm-5]
+    branches: [feat/FN-1549/refactor-report-schedule]
```
toddym42, line 5: Reminder to change this. Do we want this as `main` now?

.github/workflows/test.yml
```diff
@@ -42,6 +42,7 @@ on:
   pull_request:
     branches: 
       - main
+      - feat/FN-1549/refactor-report-schedule
```
toddym42, line 45: Reminder to remove this

.github/workflows/test.yml
```diff
@@ -126,6 +127,7 @@ env:
   UTILISATION_REPORT_SUBMISSION_PERIOD_START_EMAIL_SCHEDULE: ${{ vars.UTILISATION_REPORT_SUBMISSION_PERIOD_START_EMAIL_SCHEDULE }}
   UTILISATION_REPORT_DUE_EMAIL_SCHEDULE: ${{ vars.UTILISATION_REPORT_DUE_EMAIL_SCHEDULE }}
   UTILISATION_REPORT_OVERDUE_EMAIL_SCHEDULE: ${{ vars.UTILISATION_REPORT_OVERDUE_EMAIL_SCHEDULE }}
+  UTILISATION_REPORT_CREATION_FOR_BANKS_SCHEDULE: ${{ vars.UTILISATION_REPORT_CREATION_FOR_BANKS_SCHEDULE }}
```
ChristianMcc, line 130: Could we add this to the deployment scripts as well please? should just be a simple copy paste across development_infrastructure, staging_infrastructure and production_infrastructure

dtfs-central-api/api-tests/v1/utilisation-reports/get-utilisation-reports.api-test.ts
```diff
@@ -0,0 +1,62 @@
...
+const getUrl = (bankId: string) => `/v1/bank/${bankId}/utilisation-reports`;
+
+interface CustomResponse extends Response {
+  body:
+    | {
+        errors: {
+          msg: string;
+        }[];
+      }
+    | ({ errors?: undefined } & UtilisationReport);
```
ChristianMcc, line 21: Just trying to read this type - if we have a UtilisationReport would we not expect errors to not exist? What makes it not:

```
{
        errors: {
          msg: string;
        }[];
      }
    | UtilisationReport
    ```

Comment (ChristianMcc) [reply dtfs-central-api/api-tests/v1/utilisation-reports/get-utilisation-reports.api-test.ts:21]: Is it just so errors always has a chance of existing when looking at the object in this file?

File: dtfs-central-api/src/scheduler/jobs/create-utilisation-reports/index.ts [hunk 5]
```diff
@@ -0,0 +1,60 @@
...
+  return !currentUtilisationReportForBank;
+};
+
+/**
+ * Gets the banks which are visible in TFM utilisation
+ * reports and do not have any report in the database
+ * @returns The banks which do not have reports
+ */
+const getBanksWithMissingReports = async (): Promise<Bank[]> => {
+  const banksVisibleInTfm = (await getAllBanks()).filter((bank) => bank.isVisibleInTfmUtilisationReports);
```
ChristianMcc, line 27: A little worried we named this feature flag as just 'is visible in TFM' and now it's 'is opted into reporting', I think it's fine for now.

ChristianMcc (commented):
Only glossed over this as its all been through multiple rounds of review, no big changes needed but have left a few comments from where I was looking through :)

ChristianMcc (dismissed):

.github/workflows/test.yml
ChristianMcc: let's do this now

.github/workflows/test.yml
ChristianMcc: adding this as a todo on our technical release plan as there's a few we need to update

.github/workflows/feature_deploy.yml
ChristianMcc: yeah in a commit to fix the below may as well do main here for now

dtfs-central-api/api-tests/v1/utilisation-reports/get-utilisation-reports.api-test.ts
toddym42: Yeah exactly, but if there is a utilisation report then errors will be undefined.

dtfs-central-api/src/scheduler/jobs/create-utilisation-reports/index.ts
toddym42: Yeah, agreed. Make sense for this to be more general. Probably wouldn't be a big effort if we did want to change it?

dtfs-central-api/api-tests/v1/utilisation-reports/get-utilisation-reports.api-test.ts
ChristianMcc: Cool, I think the ? is obsolete then if its defnitely undefined

dtfs-central-api/src/scheduler/jobs/create-utilisation-reports/index.ts
ChristianMcc: Yeha just requires a DB change in prod when it gets there which is as much effort as the change. Might be worth us doing anyway and I'll raise a separate ticket for that

ChristianMcc (approved):

ChristianMcc (approved):

dtfs-central-api/api-tests/v1/utilisation-reports/get-utilisation-reports.api-test.ts
toddym42: Yeah, you're right - it is obsolete in the case. 

This is a pattern I follow for these type of unions so that you don't need to specify the value at all when manually constructing (i.e. you are not forced to add `errors: undefined`), but in this case where it just represents the response then it doesn't make any difference.

PR #2794, opened 2024-02-23
feat(FN-1861): download report sql refactor

RLavender98:
## Introduction :pencil2:
When downloading a report, retrieve the data from the SQL db instead of the mongo db.

- In the central api fetch the report from the sql db instead of the mongo db.

- Move common report types to the common library.
- Create a dtfs-central-api-response-types folder in the common library so that the tfm api and portal api can use the same response type.

## Remaining work
- I am going to raise a separate PR for fetching the list of previous reports to display on the previous reports page from the sql db.
- I am yet to run the api tests and e2e tests so I can imagine they might need further changes once I am able to run them.


dtfs-central-api/src/types/db-models/utilisation-data.ts
JoshBinns2000, line 1: Haven't reviewed the whole PR, but we decided that we should probably store mongo related types in a directory called `mongo-db-models` rather than just `db-models` (see [here](https://github.com/UK-Export-Finance/dtfs2/pull/2772/files/fd44433a14153436f3ad3436524bed0f7f089719#r1494566571))

libs/common/src/dtfs-central-api-response-types/utilisation-reports.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export type UtilisationReportResponseBody = {
+  id: number;
+  bankId: string;
+  reportPeriod: ReportPeriod;
+  dateUploaded: IsoDateTimeStamp;
+  azureFileInfo: AzureFileInfo | null;
+  status: UtilisationReportReconciliationStatus;
+  uploadedByUserId: string | null;
+};
```
JoshBinns2000, line 19: In order to keep a link between this type and `UtilisationReportEntity`, I think we should use something like this

```suggestion
type UtilisationReportResponseBody2 = Pick<UtilisationReportEntity, 'id' | 'bankId' | 'reportPeriod' | 'status' | 'uploadedByUserId'> & {
  azureFileInfo: AzureFileInfo | null;
  dateUploaded: IsoDateTimeStamp;
};
```

libs/common/src/index.ts
```diff
@@ -4,3 +4,4 @@ export * from './helpers';
 export * from './sql-db-entities';
 export * from './test-helpers';
 export * from './types';
+export * from './dtfs-central-api-response-types';
```
JoshBinns2000, line 7: Is the idea that these will also be used in the other services that call `dtfs-central-api` endpoints?

portal-api/src/v1/api.js
```diff
@@ -316,19 +316,23 @@ const getUtilisationReports = async (bankId, options) => {
   }
 };
 
-const getUtilisationReportById = async (_id) => {
+/**
+ * @param {number} id
+ * @returns {Promise<import('@ukef/dtfs2-common/').UtilisationReportResponseBody>}
```
JoshBinns2000, line 321: Should be `import('@ukef/dtfs2-common')` I think

portal-api/src/v1/api.js
```diff
@@ -316,19 +316,23 @@ const getUtilisationReports = async (bankId, options) => {
...
+ */
+const getUtilisationReportById = async (id) => {
   try {
-    if (!isValidMongoId(_id)) {
-      throw new Error(`Invalid MongoDB _id provided: '${_id}'`);
+    if (!Number.isInteger(id)) {
```
JoshBinns2000, line 325: I imagine this may be used elsewhere, so could we create a helper like `isValidMongoId` but for the SQL ids? It will probably just be a one liner which returns whether or not the value is an integer like this, but could increase the readability for why we're doing this check

portal-api/src/v1/controllers/utilisation-report-service/utilisation-report-download.controller.js
```diff
@@ -5,30 +5,36 @@ const { FILESHARES } = require('../../../constants/file-upload');
 
 /**
  * Fetches metadata about the utilisation report file associated with the
- * specified MongoDB _id
- * @param _id {string} - the MongoDB _id of the report
+ * specified id and bankId
+ * @param bankId {string} - the id of the bank
+ * @param id {number} - the SQL id of the report
  * @returns {Promise<{ filename: string; mimetype: string }>}
```
JoshBinns2000, line 11: I believe `JSDoc` object properties should be separated by a comma

portal-api/src/v1/controllers/utilisation-report-service/utilisation-report-download.controller.js
```diff
@@ -5,30 +5,36 @@ const { FILESHARES } = require('../../../constants/file-upload');
...
  */
-const getUtilisationReportFileMetadata = async (_id) => {
-  const { azureFileInfo } = await api.getUtilisationReportById(_id);
+const getUtilisationReportFileMetadata = async (bankId, id) => {
+  const response = await api.getUtilisationReportById(id);
+  const { azureFileInfo, bankId: reportBankId } = response;
+
+  if (reportBankId !== bankId) {
+    throw new Error(`Failed to find utilisation report with bankId ${bankId} and report id ${id}`)
+  }
```
JoshBinns2000, line 19: What was the reason for adding this check?

portal-api/src/v1/routes.js
```diff
@@ -289,11 +289,10 @@ authRouter
   );
 
 authRouter
-  .route('/banks/:bankId/utilisation-report-download/:_id')
+  .route('/banks/:bankId/utilisation-report-download/:id')
   .get(
     validateUserHasAtLeastOneAllowedRole({ allowedRoles: [PAYMENT_REPORT_OFFICER] }),
     bankIdValidation,
-    mongoIdValidation,
```
JoshBinns2000, line 296: Could we replace this with an `sqlIdValidation` middleware?

portal/server/routes/utilisation-report-service/utilisation-report-download/index.js
```diff
@@ -1,15 +1,14 @@
...
 const { getReportDownload } = require('../../../controllers/utilisation-report-service');
 
 const router = express.Router();
 
 router.get(
-  '/banks/:bankId/utilisation-report-download/:_id',
+  '/banks/:bankId/utilisation-report-download/:id',
   [validateToken, validateRole({ role: [PAYMENT_REPORT_OFFICER] })],
   validateBankIdForUser,
-  validateMongoId,
```
JoshBinns2000, line 12: As above, would be good to replace this with another middleware for the sql id

trade-finance-manager-api/src/types/utilisation-reports.ts
```diff
@@ -1,38 +1,4 @@
...
-  dateUploaded: IsoDateTimeStamp;
-  azureFileInfo: AzureFileInfo | null;
-  status: UtilisationReportReconciliationStatus;
-  uploadedBy: {
-    id: string;
-    firstname: string;
-    surname: string;
-  };
-};
+import { UtilisationReportReconciliationStatus } from "@ukef/dtfs2-common";
```
JoshBinns2000, line 1: Should be single quotes

trade-finance-manager-api/src/v1/api.js
```diff
@@ -1204,15 +1204,15 @@ const getUtilisationReportsReconciliationSummary = async (submissionMonth) => {
 };
 
 /**
- * @param {string} _id
- * @returns {Promise<import('../types/utilisation-reports').UtilisationReportResponseBody>}
+ * @param {string} id
+ * @returns {Promise<import('@ukef/dtfs2-common/').UtilisationReportResponseBody>}
```
JoshBinns2000, line 1208: As above, I think this should be `import('@ukef/dtfs2-common')`

trade-finance-manager-api/src/types/utilisation-reports.ts
JoshBinns2000, line 1: What was the reason to remove all these exported types? Do we not need them anymore? My first instinct is that the `UtilisationReportResponseBody` type is doing the heavy lifting with replacing these types, but not 100% sure 😅

trade-finance-manager-api/src/v1/__mocks__/mock-utilisation-report.ts
```diff
@@ -1,4 +1,5 @@
-import { AzureFileInfo, ReportPeriod, UtilisationReportResponseBody } from '../../types/utilisation-reports';
+import { AzureFileInfo, ReportPeriod, UtilisationReportResponseBody } from "@ukef/dtfs2-common";
```
JoshBinns2000, line 1: Should be single quotes

trade-finance-manager-api/src/v1/api.js
```diff
@@ -1204,15 +1204,15 @@ const getUtilisationReportsReconciliationSummary = async (submissionMonth) => {
...
  */
-const getUtilisationReportById = async (_id) => {
-  if (!isValidMongoId(_id)) {
-    throw new Error(`Invalid MongoDB _id provided: '${_id}'`);
+const getUtilisationReportById = async (id) => {
+  if (!Number.isInteger(id)) {
```
JoshBinns2000, line 1211: As mentioned above, might be nice to wrap the `isInteger` check in an `isValidSqlId` helper :)

trade-finance-manager-api/src/v1/routes.js
```diff
@@ -134,9 +134,8 @@ authRouter
   );
 
 authRouter
-  .route('/utilisation-reports/:_id/download')
+  .route('/utilisation-reports/:id/download')
   .get(
-    validation.mongoIdValidation,
```
JoshBinns2000, line 139: As above: could we add a middleware for `sqlIdValidation`

trade-finance-manager-api/src/types/utilisation-reports.ts
RLavender98: Yeah I moved all the types in here to the common library

portal-api/src/v1/routes.js
RLavender98: Yeah I'm kind of debating the point of this in the first place, and checking if the id passed in is a number. I would hope our ORM and typing would prevent anything malicious making it's way to the db call. Either way we certainly don't need to check it is an integer and in the controller and the middleware, so perhaps I will just move the check here if we do think it is important?

portal-api/src/v1/controllers/utilisation-report-service/utilisation-report-download.controller.js
RLavender98: So from my first read of the code I had thought it was possible to use an id for a report for a different bank and successfully download it, which would of course be a security issue. Mike pointed out that the folder name being the bank id prevents against that, but he suggested we add this check anyway. Doing it explicitly does a couple of things:
1. Makes the code more readable and future proof by making it clear that it should fail if a user from one bank tries to access another bank's reports.
2. Avoids us bothering to try to fetch a document we know won't be found and allows us to differentiate this failure from other document not found errors (usefulness probably dependent on log readability)

libs/common/src/index.ts
RLavender98: Yes, lots of endpoints are called by tfm api and portal api so this made sense a place to put the source of truth for response types.

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-report.controller.ts
```diff
@@ -1,26 +1,26 @@
...
   };
 }>;
 
 export const getUtilisationReportById = async (req: GetUtilisationReportByIdRequest, res: Response) => {
-  const { _id } = req.params;
+  const { id } = req.params;
 
   try {
-    const utilisationReport = await getUtilisationReportDetailsById(_id);
+    const utilisationReport = await UtilisationReportRepo.findOneBy({id: Number(id)});
```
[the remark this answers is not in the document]
RLavender98: Hmm I left as is because I felt this is the sort of basic query we can use the underlying repo for (note if I didn't need to cast to Number it would be simply `UtilisationReportRepo.findOneBy({id});`) so I felt it would be adding more complexity than removing if we create a method and perhaps setting precedence for a pattern that doesn't fully utilise the ORM?

libs/common/src/dtfs-central-api-response-types/utilisation-reports.ts
```diff
@@ -0,0 +1,19 @@
+import { UtilisationReportReconciliationStatus, ReportPeriod, IsoDateTimeStamp } from '..';
+
+export type AzureFileInfo = {
```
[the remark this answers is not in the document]
RLavender98: Nice I will copy what you have done, hopefully no merge conflicts but will just need to be aware whoever merges second

libs/common/src/dtfs-central-api-response-types/utilisation-reports.ts
RLavender98: I have never really made much use of Omit and Pick before... Seems like a good idea, though when I over on it in vscode then it is rather unclear what the types of the individual fields are. I don't know if that is a good enough reason to not do it.
I guess another reason is that in some respects it is a coincidence they match rather than a cause, in that we might want our db to store stuff differently to how we want to return it from the api and so in tying them together we are perhaps making a stronger link than there really should be?

JoshBinns2000: We have a `Prettify` utility type to get around the clarity issue when hovering on the type. That being said, I think the simplicity of the type in this case means that hovering on the type using `Pick` I've created above does show the "prettified" version anyway.

I don't think it is a coincidence that the types match up. For example, `status` should always like up with the type of `UtilisationReportEntity['status']`. I think it makes it clearer if we use a `Pick` and an intersection like I have done to make it obvious which parameters have been messed with or changed and which parameters have been left as they are in the database

portal-api/src/v1/controllers/utilisation-report-service/utilisation-report-download.controller.js
JoshBinns2000: It does feel a bit like there is some redundancy in checking this like you've said which was why I asked about it, but keeping this in as a form of documentation and future-proofing is fine I think :)

portal-api/src/v1/controllers/utilisation-report-service/utilisation-report-download.controller.test.js
```diff
@@ -17,20 +17,20 @@ describe('controllers/utilisation-report-service/utilisation-report-download', (
...
   describe('getReportDownload', () => {
     const mockBankIdParam = '956';
-    const mockReportMongoId = '5099803df3f4948bd2f98391';
+    const mockReportId = 10;
 
     const getHttpMocks = () =>
       httpMocks.createMocks(
         {
-          params: { bankId: mockBankIdParam, _id: mockReportMongoId },
+          params: { bankId: mockBankIdParam, _id: mockReportId },
```
JoshBinns2000, line 25: Should this now be `id` instead of `_id`?

portal-api/src/v1/controllers/utilisation-report-service/utilisation-report-download.controller.test.js
```diff
@@ -17,20 +17,20 @@ describe('controllers/utilisation-report-service/utilisation-report-download', (
...
         },
         { eventEmitter: events.EventEmitter },
       );
 
     it('returns an error response when the reports details do not contain the filename', async () => {
       // Arrange
       const { req, res } = getHttpMocks();
-      api.getUtilisationReportById.mockResolvedValue({});
+      api.getUtilisationReportById.mockResolvedValue({ bankId: mockBankIdParam});
```
JoshBinns2000, line 33: Missing space before closing curly bracket

portal-api/tsconfig.json
```diff
@@ -1,28 +1,28 @@
...
       // "disableReferencedProjectLoad": true,             /* Reduce the number of projects loaded automatically by TypeScript. */
-  
+
       /* Language and Environment */
       "target": "ES6",                                     /* Set the JavaScript language version for emitted JavaScript and include compatible library declarations. */
       "lib": ["dom", "esnext"],                                      /* Specify a set of bundled library declaration files that describe the target runtime environment. */
       // "jsx": "preserve",                                /* Specify what JSX code is generated. */
-      // "experimentalDecorators": true,                   /* Enable experimental support for TC39 stage 2 draft decorators. */
+      "experimentalDecorators": true,                   /* Enable experimental support for TC39 stage 2 draft decorators. */
       // "emitDecoratorMetadata": true,                    /* Emit design-type metadata for decorated declarations in source files. */
```
JoshBinns2000, line 18: I think this option also need to be set to `true` (see the `tsconfig.json` file [here](https://github.com/UK-Export-Finance/dtfs2/pull/2684/files#diff-5b159449142f44f7a41c4d51121444bc51364a3331a07405bb0588c60531048e))

portal/tsconfig.json
```diff
@@ -14,7 +14,7 @@
     "target": "es6",                                     /* Set the JavaScript language version for emitted JavaScript and include compatible library declarations. */
     "lib": ["dom", "esnext"],                                      /* Specify a set of bundled library declaration files that describe the target runtime environment. */
     // "jsx": "preserve",                                /* Specify what JSX code is generated. */
-    // "experimentalDecorators": true,                   /* Enable experimental support for TC39 stage 2 draft decorators. */
+    "experimentalDecorators": true,                   /* Enable experimental support for TC39 stage 2 draft decorators. */
     // "emitDecoratorMetadata": true,                    /* Emit design-type metadata for decorated declarations in source files. */
```
JoshBinns2000, line 18: Same as above for enabling this :)

trade-finance-manager-api/tsconfig.json
```diff
@@ -14,7 +14,7 @@
     "target": "es5",                                     /* Set the JavaScript language version for emitted JavaScript and include compatible library declarations. */
     "lib": ["dom", "esnext"],                                      /* Specify a set of bundled library declaration files that describe the target runtime environment. */
     // "jsx": "preserve",                                /* Specify what JSX code is generated. */
-    // "experimentalDecorators": true,                   /* Enable experimental support for TC39 stage 2 draft decorators. */
+    "experimentalDecorators": true,                   /* Enable experimental support for TC39 stage 2 draft decorators. */
     // "emitDecoratorMetadata": true,                    /* Emit design-type metadata for decorated declarations in source files. */
```
JoshBinns2000, line 18: Same as above for enabling this option :)

trade-finance-manager-ui/tsconfig.json
```diff
@@ -14,7 +14,7 @@
       "target": "es6",                                     /* Set the JavaScript language version for emitted JavaScript and include compatible library declarations. */
       "lib": ["dom", "esnext"],                                      /* Specify a set of bundled library declaration files that describe the target runtime environment. */
       // "jsx": "preserve",                                /* Specify what JSX code is generated. */
-      // "experimentalDecorators": true,                   /* Enable experimental support for TC39 stage 2 draft decorators. */
+      "experimentalDecorators": true,                   /* Enable experimental support for TC39 stage 2 draft decorators. */
       // "emitDecoratorMetadata": true,                    /* Emit design-type metadata for decorated declarations in source files. */
```
JoshBinns2000, line 18: Same as above for enabling this option :)

portal-api/tsconfig.json
RLavender98: Ah whoops I hadn't pushed that commit yet 😅

libs/common/src/dtfs-central-api-response-types/utilisation-reports.ts
ChristianMcc: FWIW I find this conversation quite interesting.
On one hand if we use omit then it means any new fields that will naturally be in both types get added automatically.
On the other hand if there is some random database type we add that is generated and we don't want it on this type then we have to add it to an omit to not break stuff.
In both those cases it means that there's pros and cons so it's hard to know the best choice - for some things like an API response where we mostly just return database like types does that mean we should go for omit?
In other APIs such as graphql the API response style types are normally pretty separate from say your database types and they don't overlap so much

PR #2798, opened 2024-02-26
feat(FN-1865): add script to display all sql tables in terminal

JoshBinns2000:
## Introduction :pencil2:
Whilst waiting for the AVDs to have SSMS installed, it would be nice to be able to view the data inside the SQL tables.

- Adds a hidden helper to `libs/common` which displays all tables
- Provides functionality for custom formatting to the data in the tables
- Adds a command to `libs/common/package.json` to run the commands

A screenshot of the output you get from running this command (after seeding some data) is shown below.
FN-1865-display-tables

### DIFF
--- a/libs/common/package.json
+++ b/libs/common/package.json
@@ -28,6 +28,7 @@
     "db:migrate": "npm run typeorm -- migration:run -d src/sql-db-connection/data-source.ts",
     "db:migrate:down": "npm run typeorm -- migration:revert -d src/sql-db-connection/data-source.ts",
     "db:reset": "npm run db:drop-all && npm run db:migrate",
+    "db:view-tables": "npx ts-node src/__helpers__/view-all-tables.ts",
     "lint": "eslint \"src/**/*.{ts,js}\" && npm run type-check",
     "lint:fix": "eslint \"src/**/*.{ts,js}\"",
     "prettier": "prettier --check \"src/**/*.{ts,js}\"",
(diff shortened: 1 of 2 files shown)


libs/common/package.json
```diff
@@ -28,6 +28,7 @@
     "db:migrate": "npm run typeorm -- migration:run -d src/sql-db-connection/data-source.ts",
     "db:migrate:down": "npm run typeorm -- migration:revert -d src/sql-db-connection/data-source.ts",
     "db:reset": "npm run db:drop-all && npm run db:migrate",
```
abhi-markan, line 30: I am not sure how this has passed the PR, but we simply cannot have NPM commands which allow DB to be dropped and displayed as this will introduce a vulnerability.

abhi-markan (changes requested):

JoshBinns2000:
This was an old PR that I introduced because I thought we would need the ability to view tables without SSMS, but this is not required anymore

libs/common/package.json
JoshBinns2000: This is a command that can only run locally. I believe that the user we set up for github QAT pipeline for example does not have sufficient permissions to drop the db. Where are you concerned that a vulnerability could be introduced by this npm command?

abhi-markan: A script injected into the system which can trigger this command (highly unlikely) but still should not have such commands.

JoshBinns2000: I am fairly confident that running that command even directly would not drop the db due to the user permissions. For example, in the `test.yml`, we have to use `npm run db:migrate` directly rather than running `npm run db:reset` as the latter will fail to execute due to the `DTFS` user not having sufficient permissions

PR #2748, opened 2024-02-05
feat(DTFS2-6188): remove number generator micro-service

abhi-markan:
## Introduction :pencil2:
Replace number generator Azure functions call with a POST APIM MDM call.

## Resolution :heavy_check_mark:
* Replace `number-generator` microservice with APIM MDM API call.
* Improved test coverage.
* DRY principled code.
* Converted JS to TS where applicable.
* Removed `number-generator` references and start-up commands.
* Deal submission to TFM behavious has been updated to `on-submission` rather than `on-update`, which means when a deal is submitted to TFM after numbers have been granted to deal and its associated facilities.

## Miscellaneous :heavy_plus_sign:
* Dependencies updates
* Documentation updates

PR #2803, opened 2024-02-28
feat(FN-1861): refactor downloading of report to use sql

RLavender98:
## Introduction :pencil2:
Refactor download of previous report journeys to fetch report from SQL database rather than mongodb.

 - Use SQL db for fetching previous report when downloading a report.
 - Introduce dtfs-central-api-response-types in the common library for defining response types used in more than one service.

## Remaining Work
- ~~Ensure api and e2e tests are still working~~. (dtfs-central-api api-tests still failing depend on sql seeding PR)
- I will raise a separate PR for the remaining work on FN-1861 (fetching the list of previous reports to display on previous reports page).


trade-finance-manager-api/src/v1/api.js
```diff
@@ -4,14 +4,8 @@ const { isValidMongoId, isValidPartyUrn, isValidNumericId, isValidCurrencyCode,
...
-const {
-  DTFS_CENTRAL_API_URL,
-  EXTERNAL_API_URL,
-  DTFS_CENTRAL_API_KEY,
-  EXTERNAL_API_KEY,
-  AZURE_ACBS_FUNCTION_URL,
-  AZURE_NUMBER_GENERATOR_FUNCTION_URL
-} = process.env;
+const { DTFS_CENTRAL_API_URL, EXTERNAL_API_URL, DTFS_CENTRAL_API_KEY, EXTERNAL_API_KEY, AZURE_ACBS_FUNCTION_URL, AZURE_NUMBER_GENERATOR_FUNCTION_URL } =
+  process.env;
```
JoshBinns2000, line 8: I'm guessing the formatter did this? Surprising that it re-formatted it in this way 😅

dtfs-central-api/api-tests/v1/utilisation-reports/get-utilisation-report.api-test.js
```diff
@@ -1,47 +1,26 @@
...
     it('gets a utilisation report', async () => {
-      // Arrange
-      const collection = await db.getCollection(DB_COLLECTIONS.UTILISATION_REPORTS);
-      const { insertedId } = await collection.insertOne(MOCK_UTILISATION_REPORT);
-      const _id = insertedId.toString();
-      const dateUploadedAsISOString = MOCK_UTILISATION_REPORT.dateUploaded.toISOString();
-
       // Act
-      const { body, status } = await api.get(getUrl(_id));
+      const { status } = await api.get(getUrl(MOCK_BANKS.BARCLAYS.id));
```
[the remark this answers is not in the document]
JoshBinns2000: I think this could be as simple as mocking out all the repo methods for all api-tests so that they all resolve but don't actually run (probably in the `api-test-setup.jest.config.js` file). This would also meet our criteria of wanting to limit the scope of api-tests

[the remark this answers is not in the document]
AlexBramhill: I think there are a few options here, lots of good thoughts Josh! Sorry if this is irrelevant as I've only read the x lines of new code in this file and nothing else on this review yet!

My hot takes regarding databases and tests

Data in api-tests should stand alone -- ie it should not matter what happens before or after running an api test. To the point that we should definitely not care about `npm run load` at all, or whether or not any test has previously edited the database.

This pattern is used in a few places, notably around user creation. We create the data in the database in (or before) the test we use it. We revert (or delete) these changes afterwards. After all, in a dream world, we shouldn't need to do this, as each api test should start on an empty database. We actually clear the database in a lot of user based api tests.

In terms of mocking repo methods -- I'd argue its simply less time to insert the data into a clean database. We do this elsewhere on the project. It also means we don't have to worry about "what does this complex query need to be mocked with" and allows us to have a more end to end api test.

In terms of _how_ we insert this data in the database there are two ways -- one is by api-calls, the other is by direct insert statements.

I did not like api-calls at the start of this project, however, considering the complex way we modify data it allows us to have an accurate representation of what the data in the database will actually be like. In terms of testing pyramid, this should be sufficient for api testing mostly. I think they work on this project well for this reason. A downside is that we are also somewhat "testing" these endpoints we use to populate dummy data.

The second way, direct inserts, is what I would chose on a relational database (so could be useful here), but is somewhat at risk of falling out of date on non relational databases. For this -- I usually like to create a builder function so that these can be easily populated in the database from a base configuration (immer might be a good shout for this to avoid heavy builder functions). This can be useful for relational databases, but has the downside of us needing to keep these up-to-date if our schemas change (or in the case of non-relational databases, if the B.E changes how these are saved to the db -- and this can cause our tests to invisibly get out of sync). 

I think the two choices above warrant a wider discussion. I'd prefer the second on a new project with a relational database, but I'm aware existing patterns etc.

Thank you for coming to my todd talk. Very sorry if this is tangential

libs/common/src/dtfs-central-api-response-types/utilisation-reports.ts
```diff
@@ -0,0 +1,9 @@
+import { ReportPeriod, IsoDateTimeStamp, AzureFileInfo, UtilisationReportEntity, Prettify } from '..';
+
+export type UtilisationReportResponseBody = Prettify<
```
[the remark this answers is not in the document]
AlexBramhill: I'm team no link, I think, but could be swayed to a convincing argument. My thought process is mainly about avoiding accidental exposure of new data on endpoints:

I'd expect our api-tests to validate the received response against a hardcoded object so we could pick up if there were changes to the entity type that was affecting our response type. That being said, we should also not rely on this.

So the question is -- what's the benefit of linking dtos to our db entities (assuming ive scanned this right)?

Well, for one, we should not use `omit` for the reasons you've stated -- the entity type can grow, and we can't rely on our tests to be written consistently enough pick this up.

We can use `Pick` relatively fine. However, that's not entirely correct. if `Pick` picks a nested type, we are in the exact same position. Therefore `Pick` can only be used on basic types.

So why would we even include picking from an entity, and not just type it manually ourselves to avoid potential accidents here (its easier to see **never** linking to an entity rather than **sometimes** its okay to)?

However -- where do we stop here -- does everything, such as `AzureFileInfo`, have the scope to grow past what we want our response to contain?  Do we draw the line with not linking `entities` to `responses`, or do we go further and insist that each `response` must only use primitives and other `response` objects?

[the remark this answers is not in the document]
AlexBramhill: Also -- A side point as I've not had eyes on this area, are we ensuring that at runtime we are stripping out fields that we don't want to reveal outside the service (because if we aren't -- depending on how we create the response object, we might be returning things we dont want to get out anyway)?

libs/common/src/types/date.ts
```diff
@@ -16,4 +16,9 @@ import { MonthAndYearPartialEntity } from '../sql-db-entities/partial-entities';
  */
 export type UnixTimestamp = number;
 
+/**
+ * ISO 8601 date time string in format 'yyyy-MM-ddThh:mm:ssZ'
+ */
+export type IsoDateTimeStamp = string;
```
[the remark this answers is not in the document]
AlexBramhill: cc @MarRobSoftwire

portal-api/src/v1/api.js
```diff
@@ -316,19 +316,19 @@ const getUtilisationReports = async (bankId, options) => {
...
-
-    const response = await axios.get(`${DTFS_CENTRAL_API_URL}/v1/utilisation-reports/${_id}`, {
+    const response = await axios.get(`${DTFS_CENTRAL_API_URL}/v1/utilisation-reports/${id}`, {
       headers: headers.central,
     });
 
     return response.data;
   } catch (error) {
-    console.error(`Unable to get utilisation report with MongoDB _id '${_id}'`, error);
+    console.error(`Unable to get utilisation report with id '${id}'`, error);
```
AlexBramhill, line 331: Unsure how much we care for this right now -- but if we throw this error up the chain, as appose to catching it and logging it here, further up the chain we can catch this AxiosError and assign it to the `cause` of an error up the chain. We can then keep applying this strategy as we go up to the controller level, where we can catch, log, and then handle these errors if needed to apply -- see `signInLinkController` I believe for a good example of this :) 

 For example, here, you could throw either an `ApiError` or `UtilisationReportNotFoundError` error depending on the response. We could then catch these errors at the controller level of this service, and if its a `UtilisationReportNotFoundError`, we can return a 404 (or bad req, or whatever status). If it's an `ApiError` we could let it default to a 500 by just rethrowing at the controller level.

AlexBramhill: If this is a refactor thats wide in scope -- which it might be looking at the amount of files we've already have touched  -- please just leave this and resolve this comment :) Very happy not to worry about this right now

portal-api/src/v1/controllers/utilisation-report-service/utilisation-report-download.controller.js
```diff
@@ -5,30 +5,36 @@ const { FILESHARES } = require('../../../constants/file-upload');
...
+  const response = await api.getUtilisationReportById(id);
+  const { azureFileInfo, bankId: reportBankId } = response;
+
+  if (reportBankId !== bankId) {
+    throw new Error(`Failed to find utilisation report with bankId ${bankId} and report id ${id}`);
+  }
 
   if (!azureFileInfo?.filename) {
-    throw new Error(`Failed to get filename for utilisation report with _id '${_id}'`);
+    throw new Error(`Failed to get filename for utilisation report with id '${id}'`);
```
AlexBramhill, line 22: Out of curiosity, is there a situation where we would have a 200 response from `getUtilisationReportById` where `azureFileInfo` would be missing either `filename` or `mimetype`? I feel if not, we can just check the response from `api.getUtilisationReportById(id);` and throw a `UtilisationReportNotFound` error at the `api` level, and can bin off a lot of extra code!

portal-api/src/v1/controllers/utilisation-report-service/utilisation-report-download.controller.js
```diff
@@ -37,7 +43,7 @@ const getReportDownload = async (req, res) => {
     });
 
     if (bufferedFile.error) {
-      const errorMessage = `Failed to get utilisation report for download with _id '${_id}' from Azure Storage`;
+      const errorMessage = `Failed to get utilisation report for download with id '${id}' from Azure Storage`;
       console.error(errorMessage, bufferedFile.error);
       return res.status(500).send(errorMessage);
```
AlexBramhill, line 48: Could we throw a custom error here instead of doing a console error and returning?
```
if (bufferedFile.error) {
      throw BlaError(id)
}
```
and then at the bottom we can keep our current handling, but add in a `console.error` log :) 

I quite like having all our return "at the bottom" and all our surprising returns (ie not 2xx or 3xx) in a catch block and handling them like errors -- what do you think (also very open to discuss other patterns!)?

portal-api/src/v1/controllers/utilisation-report-service/utilisation-report-download.controller.js
```diff
@@ -50,7 +56,7 @@ const getReportDownload = async (req, res) => {
 
     return readStream.pipe(res);
   } catch (error) {
-    const errorMessage = `Failed to get utilisation report for download with _id '${_id}'`;
+    const errorMessage = `Failed to get utilisation report for download with id '${id}'`;
     console.error(errorMessage, error);
     return res.status(error.response?.status ?? 500).send({ message: errorMessage });
```
AlexBramhill, line 61: Out of interest, do we want to use `.json` here? It might be worth seeing if any of the sca we have picks up on this as a risk

trade-finance-manager-api/src/v1/validation/route-validators/route-validators.js
```diff
@@ -21,13 +21,13 @@ const bankIdValidation = param('bankId')
 
 const mongoIdValidation = param('_id').isMongoId().withMessage("Invalid MongoDB '_id' path param provided");
 
+const sqlIdValidation = param('id').isInt().withMessage("Invalid 'id' path param provided");
```
AlexBramhill, line 24: Are we expecting these to always be positive? If so we could do something like
```suggestion
const sqlIdValidation = param('id').isInt({min: 0}).withMessage("Invalid 'id' path param provided");
```
just to tie it down even further

AlexBramhill (commented):
Most my comments are in response to Josh :) 

Looking really good!

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-report.controller.ts
```diff
@@ -1,26 +1,26 @@
...
 export const getUtilisationReportById = async (req: GetUtilisationReportByIdRequest, res: Response) => {
-  const { _id } = req.params;
+  const { id } = req.params;
 
   try {
-    const utilisationReport = await getUtilisationReportDetailsById(_id);
+    const utilisationReport = await UtilisationReportRepo.findOneBy({ id: Number(id) });
 
     if (!utilisationReport) {
       return res.status(404).send();
```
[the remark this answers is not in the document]
AlexBramhill: Ruby -- What do you think about throwing a `UtilisationReportNotFoundError` error in `UtilisationReportRepo.findOneBy` if the utilisation report is not found, and catching it and handling `UtilisationReporNotFoundError` specifically to return a 404 in our catch block inside this controller?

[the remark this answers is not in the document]
JoshBinns2000: `UtilisationReportRepo.findOneBy` is a built in method, so we could only throw a custom error like that if we make a custom repo method. There is also the `findOneByOrFail` method, but I think we would have to assume the `OrFail` here is only hit if the report doesn't exist

libs/common/src/dtfs-central-api-response-types/utilisation-reports.ts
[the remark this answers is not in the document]
RLavender98: Happy to change it back to how I previously wrote it if the consensus is we are happy with no linking :) I do personally prefer it as a general pattern.

@AlexBramhill I don't believe we are stripping anything out in dtfs-central-api but we are only actually using a few fields in portal-api and tfm api. I don't see there being any particular danger in things getting out, but I can double check this type captures all the returned fields.

dtfs-central-api/api-tests/v1/utilisation-reports/get-utilisation-report.api-test.js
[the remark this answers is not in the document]
JoshBinns2000: This may be worth scheduling a meeting for offline as it would be good to get input from everyone I think. Does that sound like something we can do @ChristianMcc ?

portal-api/src/v1/api.js
RLavender98: The scope of this ticket is just to refactor the endpoint to hit the SQL db rather than mongodb and has already been quite a bit more work then estimated. I'm all for making improvements as we go, and your suggestions all seem sensible, but I also want to be wary of scope creep. @ChristianMcc Do you want me to review and refactor the error response handling for the flows or shall we leave it as is?

dtfs-central-api/api-tests/v1/utilisation-reports/get-utilisation-report.api-test.js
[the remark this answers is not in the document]
RLavender98: Generally I'm very much in the camp of preferring a fresh db for each integration tests which is obviously not the existing pattern and not something to embark on in this ticket. 

Josh did point out some of the existing tests rely on seeding so given that was going to be simplest I went for that pattern, although I did fail to remember we might have reports not uploaded as I am still trying to get my head around all the domain 😅 Happy to embark on setting up adding SQL inserts for the api tests, what sort of discussion is generally had on topics like this and should I be waiting for that to happen to work on this further? I'm keen for us to try and avoid wasted effort.

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-report.controller.ts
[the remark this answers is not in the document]
AlexBramhill: Hm, is handling what happens when it returns nothing not something we would always want to do when doing a `findOne`?

portal-api/src/v1/api.js
AlexBramhill: Leave as is :)

portal-api/src/v1/controllers/utilisation-report-service/utilisation-report-download.controller.js
AlexBramhill: happy to leave as is as this ticket has scope creeped

libs/common/src/dtfs-central-api-response-types/utilisation-reports.ts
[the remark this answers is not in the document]
AlexBramhill: have a quick check that @ChristianMcc is happy with my mad ramblings :) 

Ignore the comment about what we are returning -- i've raised this offline to discuss if we want any endpoint specific middlewear for it

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-report.controller.ts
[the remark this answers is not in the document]
JoshBinns2000: That is handled by `findOneOrFail`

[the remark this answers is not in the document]
AlexBramhill: (in which case we want to find out what error that throws when it fails and catch it at the controller level?)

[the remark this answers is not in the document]
ChristianMcc: Agreed that if we use findOneOrFail in the catch segment we'll need to differentiate between a 404 and 500, I'm not familiar with the exceptions for it myself right now. Probably a good use for it if there are different exceptions thrown for no entity found vs a 500 style internal error!

dtfs-central-api/src/v1/routes/utilisation-reports-routes.js
```diff
@@ -136,10 +136,6 @@ utilisationReportsRouter
  *       500:
  *         description: Internal Server Error
  */
-utilisationReportsRouter
-  .route('/set-status')
-  .put(
-    putUtilisationReportStatusController.putUtilisationReportStatus,
-  );
+utilisationReportsRouter.route('/set-status').put(putUtilisationReportStatusController.putUtilisationReportStatus);
```
ChristianMcc, line 139: Does this not have an ID in it right now because we support bank ID and report period @JoshBinns2000 ?

dtfs-central-api/src/v1/validation/route-validators/route-validators.js
```diff
@@ -9,6 +9,10 @@ const mongoIdValidation = param('_id').isMongoId().withMessage("Invalid MongoDB
 
 exports.mongoIdValidation = [mongoIdValidation];
 
+const sqlIdValidation = param('id').isInt().withMessage("Invalid 'id' path param provided");
```
ChristianMcc, line 12: Hello I'm dumb and haven't looked up what our IDs are, are we using an incremental ID for our utilisation reports atm?

libs/common/src/types/date.ts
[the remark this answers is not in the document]
ChristianMcc: (can we keep a note of stuff to refactor we can make tickets for and not do it as part of this)

portal-api/api-tests/v1/utilisation-report-service/utilisation-report-download.api-test.js
```diff
@@ -6,10 +6,7 @@ const { DB_COLLECTIONS } = require('../../fixtures/constants');
 const { withClientAuthenticationTests } = require('../../common-tests/client-authentication-tests');
 const { withRoleAuthorisationTests } = require('../../common-tests/role-authorisation-tests');
 const { PAYMENT_REPORT_OFFICER } = require('../../../src/v1/roles/roles');
-
-jest.mock('../../../src/v1/api', () => ({
-  getUtilisationReportById: jest.fn().mockResolvedValue({ azureFileInfo: { filename: 'test-file.csv', mimetype: 'text/csv' } }),
-}));
+const api = require('../../../src/v1/api');
```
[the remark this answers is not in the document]
ChristianMcc: ditto to my earlier comment sorry

libs/common/src/dtfs-central-api-response-types/utilisation-reports.ts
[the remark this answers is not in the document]
ChristianMcc: I feel like i've flipped flopped between both approaches multiple times. I think I'll lean towards no overlapping similar again because we're the ones that are controlling both sides of the API so that's not too much of an issue in terms of missing parameters as our stuff will break if we don't provide the right information! 
It's not an area i've done a tonne of reading about so I'm basing that more on gut feel however than 'this is best practice'

portal-api/src/v1/api.js
ChristianMcc: Leave as is for now please as it's something for 90% of functions in this file :) completely i agree i'd prefer to not catch errors in this file

ChristianMcc, line 331: A weird one to ask but can we do 
```suggestion
    console.error(`Unable to get utilisation report with id '%s': %o`, id, error);
```
We've got a pattern of this in other places for console.info/errors

portal-api/src/v1/controllers/utilisation-report-service/utilisation-report-download.controller.js
ChristianMcc: Fwiw I'm used to just using send for everything as it equates to the same thing, it just makes the code more explicit to do .json but it end result is the same I think. Not sure why it would be a risk has it been flagged before?

portal/server/routes/utilisation-report-service/utilisation-report-download/index.js
```diff
@@ -1,15 +1,16 @@
...
+const {
+  ROLES: { PAYMENT_REPORT_OFFICER },
+} = require('../../../constants');
 const { getReportDownload } = require('../../../controllers/utilisation-report-service');
 
 const router = express.Router();
 
 router.get(
-  '/banks/:bankId/utilisation-report-download/:_id',
+  '/banks/:bankId/utilisation-report-download/:id',
```
ChristianMcc, line 11: Can we validate the id here as it's a user input? Just the sql id validation would be enough, just as we've lost the validateMongoId now

ChristianMcc (commented):
A few comments, a couple of nit picks and mostly just responses to some big comment chains :D

dtfs-central-api/src/v1/validation/route-validators/route-validators.js
```diff
@@ -9,6 +9,10 @@ const mongoIdValidation = param('_id').isMongoId().withMessage("Invalid MongoDB
...
+
+exports.sqlIdValidation = [sqlIdValidation];
```
AlexBramhill, line 14: nice to see this extracted as a custom validator btw!

dtfs-central-api/src/v1/validation/route-validators/route-validators.js
RLavender98: The SQL entity is using the `@PrimaryGeneratedColumn` decorator on the id which by my understanding defaults to auto incrementing IDs

libs/common/src/dtfs-central-api-response-types/utilisation-reports.ts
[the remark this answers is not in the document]
JoshBinns2000: To further bloat this comment thread, I'm not sure we should have these types in `libs/common`. I think response types should be defined in each services `api.js` file as a `JSDoc` which can then be imported in typescript files as they are needed (or in the controller depending on which side of the request the service is on). If we define them here, things may get quite confusing when responses are passed through the various apis, but defining them at the service level next to the actual `api.js` function itself will make it obvious that this response type refers specifically to this api endpoint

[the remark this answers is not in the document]
RLavender98: Hmm we have plenty of stuff in common that is meant to be consumed by only several services. The idea was since portal-api and trade-finance-manager-api both call the same endpoint by utilising the common library we only have one type we need to keep in line with what dtfs returns instead of two and them get out of sync, but happy to duplicate the type across the apis if I am slightly abusing the use of the common library here. Still trying to get my head around the desired patterns with regards to this common library 😄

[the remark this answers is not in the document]
JoshBinns2000: My main worry with it is that we often don't do anything extra to the response in `portal-api` or `tfm-api` (at least with the work we do), so we would have to define the type multiple times no matter what we do. It might just be a nicer pattern to try and keep type definitions for things like responses as close as possible to where the function is called I think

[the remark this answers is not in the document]
ChristianMcc: Hadn't even seen these newer replies when I started the call earlier, I now see why we started talking about this part more and more :D
i'm a little against defining JSDoc types and then importing into typescript, I think it's just one extra quite fiddly thing to do and we have so much inside these api.js files already that's it's going to take us so long. I think I'd just prefer us to work with types in typescript where we can and leave simple JSDocs on new functions where approrpriate in javascript. 

In terms of 'common library or individual microservice' I'd mostly treat it as:
if sql entity, common library
if type is generic to DTFS rather than an individual microservice, common library
if type is only to be used in single microservice, put it in the micorservice or common library if there is a good place for it (I'm saying use your best judgement here/it doesn't matter, in typescript we'd import our types nicely enough that you can always find the type you're looking for where it's used so it doesn't matter too much)
Personally I think it's more microservice level models than response types that fit this kind of bill

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-report.controller.ts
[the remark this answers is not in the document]
RLavender98: Sadly `error instanceof EntityNotFoundError` gives `false` when it should give `true` and we therefore can't safely differentiate the error thrown by `findOneByOrFail` from any other type of error. We believe it is similar to [this issue](https://github.com/UK-Export-Finance/dtfs2/pull/2748/files#r1509088880) but given this is not our own custom error we can't fix it (at least not easily perhaps not at all) so going to leave this as is.

portal/server/routes/middleware/validateSqlId/index.ts
```diff
@@ -0,0 +1,13 @@
+import { NextFunction, Request, Response } from 'express';
+import { isValidSqlId } from '../../../validation/validate-ids';
+
+export const validateSqlId = (req: Request, res: Response, next: NextFunction) => {
```
JoshBinns2000, line 4: Not really sure which of these I prefer, but another way to define middlewares in typescript is to use the `RequestHandler` type:
```typescript
import { RequestHandler } from 'express';
...

export const validateSqlId: RequestHandler = (req, res, next) => { ... };
```
I don't think it needs to be changed here, but just something interesting that we might want to start doing 🤷

portal/server/validation/validate-ids.js
```diff
@@ -74,6 +74,14 @@ const isValidFileName = (value) => FILE_NAME_REGEX.test(value);
  */
 const isValidBankId = (bankId) => isValidRegex(/^\d+$/, bankId);
 
+/**
+ * isValidSqlId
+ * checks if id conforms to regex
+ * @param {String} id
+ * @returns {Boolean} asserts if regex is matched
+ */
+const isValidSqlId = (id) => isValidRegex(/^\d+$/, id);
```
JoshBinns2000, line 83: Not for this PR (and not linked to any work you've done!), but `isValidRegex` is not a very useful helper, isn't typed correctly and has a misleading name - another piece of tech debt rears its head!

trade-finance-manager-api/src/v1/api.js
```diff
@@ -1204,15 +1194,12 @@ const getUtilisationReportsReconciliationSummary = async (submissionMonth) => {
 };
 
 /**
- * @param {string} _id
- * @returns {Promise<import('../types/utilisation-reports').UtilisationReportResponseBody>}
+ * Get utilisation report by id
+ * @param {string} id
+ * @returns {Promise<import('./api-response-types/UtilisationReportResponseBody').UtilisationReportResponseBody>}
```
JoshBinns2000, line 1199: Big fan of this pattern!

trade-finance-manager-ui/server/helpers/validateIds.js
```diff
@@ -26,6 +26,14 @@ const isValidGroupId = (groupId) => isValidRegex(REGEX.INT, groupId);
  */
 const isValidTaskId = (taskId) => isValidRegex(REGEX.INT, taskId);
 
+/**
+  Validates if a value is a valid sql id
+
+ * @param id - the value to validate
+ * @returns Boolean - true if valid, false if not
+ */
```
JoshBinns2000, line 34:
```suggestion
/**
 * Validates if a value is a valid sql id
 * @param {string} id - the value to validate
 * @returns {boolean} `true` if valid, `false` if not
 */
```

trade-finance-manager-ui/server/middleware/validateSqlId/index.ts
```diff
@@ -0,0 +1,13 @@
...
+export const validateSqlId = (req: Request, res: Response, next: NextFunction) => {
+  const { id } = req.params;
+
+  if (!isValidSqlId(id)) {
+    console.error(`Invalid SQL 'id' param provided: '${id}'`);
+    return res.redirect('/not-found');
+  }
+
+  return next();
+};
```
JoshBinns2000, line 13: Missing new line at end of file :)

JoshBinns2000 (commented):
Looking good! Only some minor comments and I really like the new `api-response-types` directory and pattern you've introduced - should make adding new api functions in the future much easier to integrate with our typescript stuff

trade-finance-manager-ui/server/middleware/validateSqlId/index.ts
```diff
@@ -0,0 +1,13 @@
+import { NextFunction, Request, Response } from 'express';
+import { isValidSqlId } from '../../helpers/validateIds';
+
+export const validateSqlId = (req: Request, res: Response, next: NextFunction) => {
```
ChristianMcc, line 4: can you add a quick docstring at the top:
`/**
Middleware to validate that the 'id' field found as a path param is a valid SQL id
*/`

portal/server/routes/middleware/validateSqlId/index.ts
ChristianMcc, line 4: ditto on my other comment for a short docstring :)

dtfs-central-api/api-tests/mocks/entities/azure-file-info-entity.ts
```diff
@@ -0,0 +1,12 @@
+import { AzureFileInfoEntity } from "@ukef/dtfs2-common";
+
+export const anAzureFileInfoEntity = (): AzureFileInfoEntity => {
```
ChristianMcc, line 3: I've seen this kind of naming pattern before which tbh I quite like. It's a little different from the previous pattern of just calling stuff MOCK_AZURE... etc. although there didn't seem to be a pattern before that anyway and it was quite random. It's also nice this is a function than an object to use! Wouldn't mind moving to this

dtfs-central-api/api-tests/v1/utilisation-reports/get-utilisation-report.api-test.ts
```diff
@@ -0,0 +1,63 @@
...
+import { SqlDbDataSource } from '@ukef/dtfs2-common/sql-db-connection';
+import { AzureFileInfo, IsoDateTimeStamp, ReportPeriod, UtilisationReportEntity, UtilisationReportReconciliationStatus } from '@ukef/dtfs2-common';
+import app from '../../../src/createApp';
+import apiModule from '../../api';
+import { anUploadedUtilisationReportEntity } from '../../mocks/entities/utilisation-report-entity';
+
+const api = apiModule(app);
+const getUrl = (id: string) => `/v1/utilisation-reports/${id}`;
+
+const givenUploadedReport = async (): Promise<string> => {
```
ChristianMcc, line 11: Is this naming trying to follow the like given when then approach and why it's framed like this? A pattern like this makes sense, wondering if it's slightly less clear that it's uploading a report and returning one if you don't know the pattern - thoughts?

ChristianMcc (approved):
nice work! have approved - left a couple of comments that wouldn't mind discussing for naming approach though :) feel free to resolve any comments still on the PR

dtfs-central-api/api-tests/mocks/entities/azure-file-info-entity.ts
RLavender98: I recognise this is quite different to current patterns, but yeah I really like using functions for getting test data so there is less possibility of dependency between tests and confusing test failures where one test modifies a mock object by accident and causes other tests to fail!

dtfs-central-api/api-tests/v1/utilisation-reports/get-utilisation-report.api-test.ts
RLavender98: Yeah, the idea behind it is to name the function in a way that makes the test as readable as possible, leaving out the technical details. However I see your point, and think it is perfectly fine if we would prefer more of the detail in the test to not diverge too far from current tests, especially as we are returning the id which is non-obvious unless it is a pattern you know and are familiar with.

Given this is only used once I will perhaps just move it into the test rather than write a more descriptive function name and then on tests where there is reusability advantage to a helper function like this go with a slightly more descriptive name?

dtfs-central-api/api-tests/mocks/entities/azure-file-info-entity.ts
JoshBinns2000: I think the way we've normally done this is to treat `MOCK_<thing>` as a fully defined object for `<thing>` and then modify the fields that we want to for the specific test case, therefore removing the dependency on the shape of the mock object with respect to fields that we don't care about. For example, if we wanted to specifically test the `dateUploaded` property of a utilisation report, we would have a step like
```typescript
it('does something', () => {
  // Arrange
  const dateUploaded = new Date('2024-01-01');
  
  // by using MOCK_UTILISATION_REPORT, we remove any type errors without having to explicitly define all properties
  const reportWithUploadedDate: UtilisationReport = {
    ...MOCK_UTILISATION_REPORT,
    dateUploaded,
  };
  
  // Act
  ...
  
  // Assert
  expect(<someResult>).toEqual(expect.objectContaining({ dateUploaded }));
});
```
A pattern like this seems to just abstract away the "overwriting the fields we need and keeping everything else with valid but arbitrary values" part of the `reportWithUploadedDate` above, so I don't really think it's a different pattern so much as a more explicit way of doing what we've usually been doing. All that to say I don't think there are any risks in introducing this as a way of generating mock data for tests

JoshBinns2000: That being said, a safer pattern that is currently being added in FN-1859 is to have a class in `libs/common` called `UtilisationReportEntityMockBuilder` which creates and returns a `UtilisationReportEntity` with specified fields, defaulting to arbitrary values when those fields are net set in the mock creation. I think we should try and go with that pattern, but I also don't think we need to introduce that as part of this PR

dtfs-central-api/api-tests/v1/utilisation-reports/get-utilisation-report.api-test.ts
ChristianMcc: Yeah sounds good to me :)

dtfs-central-api/api-tests/mocks/entities/azure-file-info-entity.ts
ChristianMcc: Yeah I think I just like this pattern above because it gets rid of the risk of someone going `ooh a mock, i'll use that directly` while doing a test that doesn't really need it too much and moving that spread pattern described essentially out to the function. 
In terms of a mock builder in libs/common rather than anywhere else, I think fitting those would work to be used inside these individual givenUploadedReport/aUtilisationReport, or are you thinking of also moving all helpers that say 'insert a utilisation report into the DB for an api-test' into libs/common as well and having a lot of test function helpers in there?

Worth noting we should be careful as to what we build as part of the libs/common package that gets built in all other services - right now we only build `./src` in the different apps and the api-tests etc. dont make it into deployed environments, and so if we put loads of testing logic in libs/common we'll need to do similar

ChristianMcc: (fwiw i think this should be merged now independent of this, and we build on this approach in future PRs similar to what was just suggested, dont let this discussion block the merging :D )

PR #2805, opened 2024-02-28
chore(DTFS2-6996): remove moment from dtfs-central-api

MarRobSoftwire:
## Introduction :pencil2:
Moment js is deprecated, it should be replaced.

* Replace moment with date-fns to convert a search string into a date when querying `dateReceived`
  * This covers code originally written under [DTFS2-3065](https://ukef-dtfs.atlassian.net/browse/DTFS2-3065). We've had to agree new ACs for this behaviour but have tried to maintain, within reason, the same behaviour


dtfs-central-api/package-lock.json
MarRobSoftwire, line 1: The individual package-lock.json files are redundant and should be deleted. They're no longer updated when running `npm i`
See [here ](https://docs.npmjs.com/cli/v7/using-npm/workspaces) underneath "Below is a post npm install example, given that same previous example structure of files and folders" 
```
+-- node_modules
|  `-- workspace-a -> ../workspace-a
+-- package-lock.json
+-- package.json
`-- workspace-a
   `-- package.json
```

dtfs-central-api/src/helpers/getDateFromSearchString.test.ts
```diff
@@ -0,0 +1,53 @@
...
+    expectedDate: mockDate
+  }));
+  const testDataWithoutYear = acceptedFormatsWithoutYear.map((formatString) => ({
+    formatString,
+    mockString: format(mockDate, formatString),
+    expectedDate: mockDateWithCurrentYear
+  }))
+
+
+  it.each([...testDataWithYear, ...testDataWithoutYear])('should correctly parse a date in the format $formatString', ({ mockString, expectedDate }) => {
```
AlexBramhill, line 48: can we document behaviour if it cant pass the date please :)

AlexBramhill (approved):

PR #2807, opened 2024-02-28
feat(FN-1862): refactor TFM reports page

JoshBinns2000:
## Note 📓 
This PR handles updating the `insertUtilisationReportDetailsIntoDb` and `removeAllUtilisationReportDetailsFromDb` tasks to work the with SQL database instead of mongo and fixes the TFM `navigating-to-utilisation-reports.spec.js` and `mark-report-as-done.spec.js`* tests to insert data with the correct schema. Existing e2e tests which use these cypress tasks will likely fail to insert the reports until those files are updated to insert reports with the correct schema.

*even though this file has been updated to correctly insert reports, the "mark report as completed" functionality has not yet been refactored to SQL which is why this test still fails

## Introduction :pencil2:
We want to refactor the TFM reports page to fetch data from the SQL database.

- Introduces a `UtilisationDataRepo` to `dtfs-central-api`
- Adds a shared `MOCK_UTILISATION_DATA_ENTITY`
- Adds a `MOCK_AUDITABLE_BASE_ENTITY` to `libs/common` for easier creation of other entity mocks
- Updates existing tests for the `utilisation-reports/reconciliation-summary/:submissionMonth` related endpoints

## Miscellaneous
- Updates `e2e-tests` `test.yml` to use the same node version as our other tests


dtfs-central-api/src/repositories/utilisation-reports-repo/utilisation-report-sql.repo.ts
```diff
@@ -1,11 +1,31 @@
...
+            { reportPeriodStart },
+          ),
+      )
+      .getMany();
+    // TODO FN-1862 should we install the typeorm dependency to use that instead of query builder?
+    // return await this.find({
+    //   where: {
+    //     bankId: Equal
+    //   }
+    // })
```
JoshBinns2000, line 29: There are things exported by `typeorm` which can do this sort of logic. However, we don't have `typeorm` as a dependency in `dtfs-central-api`. I assumed this was done for a reason, so made this query using the `createQueryBuilder` instead. If we are happy to install `typeorm` as a dependency in `dtfs-central-api`, then I can write this query in a more javascript-y way

RLavender98: I obviously can't speak for the set up that has already happened, but my assumption would be it hasn't been added because it simply hasn't been needed yet and we should 100% utilise typeorm to make our db interactions as simple and tidy as possible 😄

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-reports-reconciliation-summary.controller/helpers.ts
```diff
@@ -17,29 +17,32 @@ import { isEqualMonthAndYear } from '../../../../utils/date';
...
 type SummaryItemForSubmissionMonth = {
   submissionMonth: IsoMonthStamp;
   item: UtilisationReportReconciliationSummaryItem;
 };
 
-const mapToSummaryItem = async (bank: Bank, report: UtilisationReport): Promise<UtilisationReportReconciliationSummaryItem> => {
-  const reportData = await getAllUtilisationDataForReport(report);
+const mapToSummaryItem = async (bank: Bank, report: UtilisationReportEntity): Promise<UtilisationReportReconciliationSummaryItem> => {
+  const reportData = await UtilisationDataRepo.findByReport(report);
+  if (!reportData) {
```
RLavender98, line 30: Is it possible for this to return a falsy value? My guess would be this either throws an error ( e.g. because it can't get connect to the db) or returns an array, which are always truthy in javascript.

RLavender98 (changes requested):
Looks good, I agree we should be making the most of typeorm's offerings so if you can use them in place of the query builder I think that would be great 😄

dtfs-central-api/src/repositories/utilisation-data-repo/utilisation-data-sql.repo.ts
```diff
@@ -0,0 +1,10 @@
+// TODO FN-1853 - rename this to `utilisation-data.repo.ts` when all repo
+//  methods have been migrated from MongoDB to SQL
+import { SqlDbDataSource } from '@ukef/dtfs2-common/sql-db-connection';
+import { UtilisationDataEntity, UtilisationReportEntity } from '@ukef/dtfs2-common';
+
+export const UtilisationDataRepo = SqlDbDataSource.getRepository(UtilisationDataEntity).extend({
+  async findByReport(report: UtilisationReportEntity): Promise<UtilisationDataEntity[] | null> {
+    return await this.findBy({ report });
+  },
+});
```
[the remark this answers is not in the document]
JoshBinns2000: Funnily enough I already had a commit where I removed this locally and forgot to push it 😅 - good catch either way!

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-reports-reconciliation-summary.controller/helpers.ts
JoshBinns2000: This was a mistake from copying other stuff I think. The return type I put for `.findByReport` was `Promise<UtilisationDataEntity[] | null>`, but it will always be `Promise<UtilisationDataEntity[]>`, so the thing that we would actually want to check for is `reportData.length` being zero. However, it's fine for there to be no report data, so this check should be removed entirely. Thanks for pointing it out!

dtfs-central-api/api-tests/mocks/utilisation-reports/utilisation-data.ts
```diff
@@ -24,10 +25,8 @@ export const MOCK_UTILISATION_DATA: UtilisationData = {
   payments: null,
 };
 
-export const getMockUtilisationDataForReport = (report: UtilisationReport, overrides?: Partial<UtilisationData>): UtilisationData => ({
-  ...MOCK_UTILISATION_DATA,
-  reportId: report._id.toString(),
-  bankId: report.bank.id,
-  reportPeriod: report.reportPeriod,
+export const getMockUtilisationDataForReport = (report: UtilisationReportEntity, overrides?: Partial<UtilisationDataEntity>): UtilisationDataEntity => ({
```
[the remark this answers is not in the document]
JoshBinns2000: It is only used in a unit test, but we don't have any unit test mocks, so all of our test mocks usually end up sitting in these api-test mocks 😅

dtfs-central-api/src/repositories/utilisation-reports-repo/utilisation-report-sql.repo.ts
JoshBinns2000: I have updated the way this query works now to not use the query builder. I'm not convinced it's easier to read, but let me know what you think!

dtfs-central-api/src/repositories/utilisation-reports-repo/utilisation-report-sql.repo.ts
```diff
@@ -2,9 +2,40 @@
...
+          month: LessThan(reportPeriodStart.month),
+        },
+      },
+    };
+
+    return await this.find({
+      where: [
+        { ...bankIdAndStatusFindOptions, ...previousYearFindOptions },
+        { ...bankIdAndStatusFindOptions, ...sameYearPreviousMonthsFindOptions },
+      ],
```
JoshBinns2000, line 38: An array of `where` options functions as an SQL `OR` statement

e2e-tests/support/tasks.js
```diff
@@ -81,14 +87,22 @@ module.exports = {
...
+       * Inserts utilisation report details to the SQL database
+       * @param {Partial<import('@ukef/dtfs2-common').UtilisationReportEntity[]>} utilisationReportDetails
+       * @returns {import('@ukef/dtfs2-common').UtilisationReportEntity[]} The inserted reports
+       */
       async insertUtilisationReportDetailsIntoDb(utilisationReportDetails) {
-        const utilisationReports = await db.getCollection(DB_COLLECTIONS.UTILISATION_REPORTS);
-        return utilisationReports.insertMany(utilisationReportDetails);
+        const utilisationReportRepo = SqlDbDataSource.getRepository(UtilisationReportEntity);
+        const saveOperations = utilisationReportDetails.map((utilisationReport) => utilisationReportRepo.save(utilisationReport));
+        return await Promise.all(saveOperations);
```
ChristianMcc, line 98: Is this again because of the different types of reports we can save at once?

e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/pdc-reconcile/mark-report-as-done.spec.js
```diff
@@ -59,47 +55,40 @@ context(`${PDC_TEAMS.PDC_RECONCILE} users can mark reports as done and not done`
...
       const status = statusWithBankId.at(index)?.status;
 
       if (!status || status === UTILISATION_REPORT_RECONCILIATION_STATUS.REPORT_NOT_RECEIVED) {
-        mockUtilisationReportDetailsWithoutId.push(
-          createNotReceivedReportDetails(reportBank, {
-            reportPeriod,
-          }),
-        );
+        const mockNotReceivedReportDetails = UtilisationReportEntity.createNotReceived({
+          bankId: bank.id,
```
ChristianMcc, line 66: we've pulled out bankId so may as well use that

e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/pdc-reconcile/mark-report-as-done.spec.js
```diff
@@ -59,47 +55,40 @@ context(`${PDC_TEAMS.PDC_RECONCILE} users can mark reports as done and not done`
...
+      const reportId = bankId;
       const status = statusWithBankId.at(index)?.status;
 
       if (!status || status === UTILISATION_REPORT_RECONCILIATION_STATUS.REPORT_NOT_RECEIVED) {
-        mockUtilisationReportDetailsWithoutId.push(
-          createNotReceivedReportDetails(reportBank, {
-            reportPeriod,
-          }),
-        );
+        const mockNotReceivedReportDetails = UtilisationReportEntity.createNotReceived({
```
ChristianMcc, line 65: Coming back to your pain you mentioned of reportDetails being the same for both single and plural. Now that in SQL we just have UtilisationReportEntity could we just call these utilisationReports in code and variable names? Quite a few timees we refer to it so not sure if it's just too big a refactor for now/if we want it

libs/common/src/sql-db-connection/migrations/1709731032407-AddOnDeleteCascadeToAzureFileInfo.ts
```diff
@@ -0,0 +1,137 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class AddOnDeleteCascadeToAzureFileInfo1709731032407 implements MigrationInterface {
+  name = 'AddOnDeleteCascadeToAzureFileInfo1709731032407';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`
+            ALTER TABLE "AzureFileInfo" DROP CONSTRAINT "FK_4370b01aed42cea044ee2b1ce6f"
+        `);
+    await queryRunner.query(`
```
ChristianMcc, line 10: Why are all these extra fields getting changed? have we changed their types somewhere?

libs/common/src/sql-db-entities/base-entities/auditable.base-entity.ts
```diff
@@ -8,6 +8,6 @@ export abstract class AuditableBaseEntity {
   @UpdateDateColumn()
   updatedDate!: Date;
 
-  @Column()
+  @Column({ type: 'varchar' })
```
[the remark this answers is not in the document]
ChristianMcc: is this saving to the database it was struggling with you mean? (Also nice this is where the changes on the migration were)

libs/common/src/test-helpers/mock-data/utilisation-data.entity.mock.ts
```diff
@@ -0,0 +1,20 @@
+import { UtilisationDataEntity } from '../../sql-db-entities';
+import { MOCK_UTILISATION_REPORT_ENTITY } from './utilisation-report.entity.mock';
+import { MOCK_AUDITABLE_BASE_ENTITY } from './auditable.base-entity.mock';
+
+export const MOCK_UTILISATION_DATA_ENTITY: UtilisationDataEntity = {
+  id: 1,
+  facilityId: '123',
+  report: MOCK_UTILISATION_REPORT_ENTITY,
+  exporter: 'potato person',
```
ChristianMcc, line 9: test exporter please :D

ChristianMcc (commented):
Nice work, few small comments/questions :)

.github/workflows/test.yml
```diff
@@ -557,7 +573,13 @@ jobs:
       cancel-in-progress: true
 
     steps:
-      - uses: actions/checkout@v4
+      - name: Repository
+        uses: actions/checkout@v4
+      
+      - name: Node
```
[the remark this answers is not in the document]
JoshBinns2000: No - this was something discussed between @AlexBramhill and I about the node version in e2e tests. All these steps do is ensure that we are using the same node version here as we are for the rest of our tests rather than using the bundled version from cypress

dtfs-central-api/src/repositories/utilisation-reports-repo/utilisation-report-sql.repo.ts
```diff
@@ -2,9 +2,40 @@
 //  methods have been migrated from MongoDB to SQL
 import { SqlDbDataSource } from '@ukef/dtfs2-common/sql-db-connection';
 import { UtilisationReportEntity, ReportPeriod } from '@ukef/dtfs2-common';
+import { Not, Equal, FindOptionsWhere, LessThan } from 'typeorm';
 
 export const UtilisationReportRepo = SqlDbDataSource.getRepository(UtilisationReportEntity).extend({
   async findOneByBankIdAndReportPeriod(bankId: string, reportPeriod: ReportPeriod): Promise<UtilisationReportEntity | null> {
     return await this.findOneBy({ bankId, reportPeriod });
   },
+  async findOpenReportsBeforeReportPeriodStartForBankId(bankId: string, reportPeriodStart: ReportPeriod['start']): Promise<UtilisationReportEntity[]> {
```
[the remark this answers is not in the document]
JoshBinns2000: This method was originally designed to get all open (not in the `'RECONCILIATION_COMPLETED'` state) reports for report periods before the current report period for the specific bank. I'm not fully sure what your question is asking as this function is not really linked to the "uploaded" state of the report

e2e-tests/support/tasks.js
JoshBinns2000: Yup

libs/common/src/sql-db-entities/base-entities/auditable.base-entity.ts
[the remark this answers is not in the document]
JoshBinns2000: No - this was just because of a cypress issue

libs/common/src/sql-db-connection/migrations/1709731032407-AddOnDeleteCascadeToAzureFileInfo.ts
JoshBinns2000: For future readers, see [here](https://github.com/UK-Export-Finance/dtfs2/pull/2807/files#r1514714219)

e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/pdc-reconcile/mark-report-as-done.spec.js
JoshBinns2000: I have refactored it everywhere because why not 🚀

dtfs-central-api/src/repositories/utilisation-reports-repo/utilisation-report-sql.repo.ts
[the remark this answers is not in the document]
ChristianMcc: Long story short I don't want any changes :D 

I was toying with the idea of what do we need an endpoint like this that takes a reportPeriodStart in for when we could just be looking for open reports all the time (as none will ever be in the future as they won't have been made by the job yet) and we might have been able to simplify it massively. But feels irrelevant to think about given we've written out the logic :)

ChristianMcc (approved):
Nice work! Let's get this in :D

PR #2853, opened 2024-03-06
feat(FN-1861): refactor fetching previous reports to use sql

RLavender98:
## Introduction :pencil2:
Refactor fetching previous reports to fetch from the sql database instead of mongo.

- When fetching previous reports to display on the previous reports page fetch from SQL db.
- When fetching reports for specific report period to check for whether a report has already been uploading when uploading a report, fetch from SQL db.


dtfs-central-api/src/repositories/utilisation-reports-repo/utilisation-report-sql.repo.ts
```diff
@@ -1,10 +1,35 @@
...
+  reportPeriod?: ReportPeriod;
+  excludeNotUploaded?: boolean;
+};
 
 export const UtilisationReportRepo = SqlDbDataSource.getRepository(UtilisationReportEntity).extend({
   async findOneByBankIdAndReportPeriod(bankId: string, reportPeriod: ReportPeriod): Promise<UtilisationReportEntity | null> {
     return await this.findOneBy({ bankId, reportPeriod });
   },
+
+  async findAllByBankId(bankId: string, options?: GetUtilisationReportDetailsOptions): Promise<UtilisationReportEntity[]> {
```
RLavender98, line 16: This method is quite involved and I am aware I have not written any tests for it, whilst I have deleted the unit tests relating to the previous mongo implementation. I'm not sure what unit tests if any we want on SQL repositories like this or if we want to cover with api/e2e tests since we have the db connection?

dtfs-central-api/src/repositories/utilisation-reports-repo/utilisation-report-sql.repo.ts
```diff
@@ -1,10 +1,35 @@
...
+        .andWhere('azureFileInfo.utilisationReportId IS NOT NULL')
+        .andWhere('report.status <> :status', { status: UTILISATION_REPORT_RECONCILIATION_STATUS.REPORT_NOT_RECEIVED });
+    }
+
+    if (options?.reportPeriod) {
+      query = query
+        .andWhere('report.reportPeriodStartMonth = :startMonth', { startMonth: options.reportPeriod.start.month })
+        .andWhere('report.reportPeriodEndMonth = :endMonth', { endMonth: options.reportPeriod.end.month })
+        .andWhere('report.reportPeriodStartYear = :startYear', { startYear: options.reportPeriod.start.year })
+        .andWhere('report.reportPeriodEndYear = :endYear', { endYear: options.reportPeriod.end.year });
```
RLavender98, line 30: `reportPeriodStartMonth` etc are the actual db column names, which are generated by typeORM from the nested structure we have in our entity type. I don't think there is a way around using them explicitly but it did feel a bit funky to do so, similarly with using the `azureFileInfo.utilisationReportId` above. I'm thinking this might just be what it is like working with the query builder and I perhaps just need to get used to it, but just wanting to highlight in case I am missing any tricks here!

dtfs-central-api/src/repositories/utilisation-reports-repo/utilisation-report-sql.repo.ts
ChristianMcc: Good question, I'm sure we could figure out how to write some tests where we mock the createQueryBuilder and check it's made in the right way but can also imagine it might be a little grim. If we have the coverage of the different if loops in API tests which I've seen you've added we should be fine without :)

ChristianMcc (commented):
Nice work, couple of minor points :)

dtfs-central-api/src/repositories/utilisation-reports-repo/utilisation-report-sql.repo.ts
```diff
@@ -1,10 +1,35 @@
...
 export const UtilisationReportRepo = SqlDbDataSource.getRepository(UtilisationReportEntity).extend({
   async findOneByBankIdAndReportPeriod(bankId: string, reportPeriod: ReportPeriod): Promise<UtilisationReportEntity | null> {
     return await this.findOneBy({ bankId, reportPeriod });
   },
+
+  async findAllByBankId(bankId: string, options?: GetUtilisationReportDetailsOptions): Promise<UtilisationReportEntity[]> {
+    let query = this.createQueryBuilder('report').where('report.bankId = :bankId', { bankId });
+    if (options?.excludeNotUploaded) {
+      query = query
+        .leftJoinAndSelect('report.azureFileInfo', 'azureFileInfo')
```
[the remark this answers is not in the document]
RLavender98: So I think I need the left join so I can reference the azure file info fields in the next line? 
We can't not check for azure file info at all, because for the completed status, the bank may have uploaded a file or they may have not uploaded a file, so we can't just use status to know if a report was actually uploaded.

dtfs-central-api/api-tests/v1/utilisation-reports/get-utilisation-reports.api-test.ts
```diff
@@ -1,61 +1,121 @@
...
+import { wipeAllUtilisationReports } from '../../test-helpers/wipe-sql-db';
 
 const api = createApi(app);
 
 const getUrl = (bankId: string) => `/v1/bank/${bankId}/utilisation-reports`;
 
+const saveReportToDatabase = async (report: UtilisationReportEntity): Promise<UtilisationReportEntity> =>
+  await SqlDbDataSource.getRepository(UtilisationReportEntity).save(report);
+
+type UtilisationReportResponseBody = {
```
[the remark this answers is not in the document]
ChristianMcc: Ah fair enough, bit annoying we need a full return type in a test but such are api tests, ignore me then :)

dtfs-central-api/src/repositories/utilisation-reports-repo/utilisation-report-sql.repo.ts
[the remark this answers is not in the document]
RLavender98: As discussed offline, I have now changed this to not filter on azure file info and leave that to be filtered by portal-api for the previous reports page specifically.

As a side note, if I had left this as using `createQueryBuilder` I actually would have still had to join on the azure file info table to retrieve it or use an option like the following: `.setFindOptions({loadEagerRelations: true})` since it seems by default it does not load in joint entities even if they are noted as eager, which is perhaps us worth keeping in mind if we do use query builder for future methods!

dtfs-central-api/api-tests/v1/utilisation-reports/get-utilisation-reports.api-test.ts
```diff
@@ -1,61 +1,121 @@
...
+
+    // Act
+    const urlWithQueryParams = axios.getUri({ url: getUrl(bankId), params: { reportPeriod, excludeNotUploaded: true } });
+    const response: CustomResponse = await api.get(urlWithQueryParams);
+
+    // Assert
+    expect(response.status).toEqual(200);
+    expect((response.body as UtilisationReportResponseBody[]).length).toEqual(1);
+    expect((response.body as UtilisationReportResponseBody[])[0].id).toEqual(idOfReportForReportPeriod);
+  }, 10000);
```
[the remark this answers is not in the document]
RLavender98: Woops! Removed 😄

dtfs-central-api/api-tests/mocks/entities/utilisation-report-entity.ts
```diff
@@ -11,4 +11,26 @@ export const anUploadedUtilisationReportEntity = (): UtilisationReportEntity =>
...
+};
+
+export const aNonUploadedMarkedReconciledUtilisationReportEntity = (): UtilisationReportEntity => {
+  const mock = new UtilisationReportEntity();
+  mock.bankId = '123';
+  mock.azureFileInfo = undefined;
+  mock.reportPeriod = aReportPeriodPartialEntity();
+  mock.status = UTILISATION_REPORT_RECONCILIATION_STATUS.RECONCILIATION_COMPLETED;
+  mock.data = [];
+  mock.updatedByUserId = 'SYSTEM';
```
JoshBinns2000, line 34: A report which has been marked as completed would have `updatedByUserId` be equal to `'TFM-<userId>'`

dtfs-central-api/api-tests/v1/utilisation-reports/get-utilisation-reports.api-test.ts
```diff
@@ -1,61 +1,123 @@
...
+
 interface CustomResponse extends Response {
   body:
     | {
         errors: {
           msg: string;
         }[];
       }
-    | ({ errors?: undefined } & UtilisationReport);
+    | ({ errors?: undefined } & UtilisationReportResponseBody[]);
```
JoshBinns2000, line 38: I think we probably want to change this type now seeing as we're losing its benefits by needing all the `as` type assertions below. My idea would be that we have two functions which assert type correct type and two response types: a `success` and `error` response:
```typescript
interface CustomErrorResponse extends Response {
  body: {
    errors: {  msg: string }[];
  };
}

interface CustomResponse extends Response {
  body: UtilisationReportResponseBody;
}

// inside `describe` - needs to use `function` instead of `() => {}` to use `asserts <> is <>`
function assertIsErrorResponse(response: Response): asserts response is CustomErrorResponse {
  if (!('errors' in response.body)) {
    throw new Error("Expected error response to contain 'response.body.errors'");
  }
  expect(Array.isArray(response.body.errors)).toBe(true);
  response.body.errors.forEach((error) => {
    if (!('msg' in error)) {
      throw new Error("Expected 'response.body.errors' array items to all contain 'msg' property");
    }
    expect(typeof error.msg === 'string').toBe(true);
  );
}
```
with a similar function for `assertIsSuccessResponse`. This would give us the type safety we want as well as giving us some testing on what the endpoint returns for free

dtfs-central-api/api-tests/v1/utilisation-reports/get-utilisation-reports.api-test.ts
```diff
@@ -1,61 +1,123 @@
...
+  });
+
+  afterEach(async () => {
+    await wipeAllUtilisationReports();
   });
 
   it('returns 400 when an invalid bank id is provided', async () => {
     // Act
-    const response: CustomResponse = await api.get(getUrl('invalid-mongo-id'));
+    const response: CustomResponse = await api.get(getUrl('invalid-id'));
```
JoshBinns2000, line 53: If we introduce the stuff I've commented above, I think the `: CustomResponse` type annotation won't be needed as we will be asserting type with the helper functions which are themselves just expected a `Response` (which I think `api.<fn>` will return by default)

JoshBinns2000 (commented):
Just a few comments on the `supertest` type stuff as I've only looked at that file

dtfs-central-api/api-tests/v1/utilisation-reports/get-utilisation-reports.api-test.ts
```diff
@@ -1,61 +1,123 @@
...
+
+type UtilisationReportResponseBody = {
+  id: number;
+  bankId: string;
+  status: UtilisationReportReconciliationStatus;
+  uploadedByUserId: string;
+  reportPeriod: ReportPeriod;
+  azureFileInfo: AzureFileInfo | null;
+  dateUploaded: IsoDateTimeStamp;
+};
```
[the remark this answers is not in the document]
RLavender98: Agree it is missing info, I was perhaps being a bit cheeky returning more than we really should... I'll tidy up with a mapping function.
As for the second comment, I don't really see what you mean. Either way it gives the same the same overall type no?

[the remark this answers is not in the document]
JoshBinns2000: It does give the same type, but it's just a semantic thing - `response.body` will be an array, so the `ResponseBody` type itself should be an array

dtfs-central-api/api-tests/v1/utilisation-reports/get-utilisation-reports.api-test.ts
RLavender98: Ok so I have gone with the two types as you suggest which I agree is much nicer, but then I have perhaps been a bit sneaky and just typed the response as whichever one we expect it to be. I think this is ok, because assertions will fail anyway if it returns the incorrect type, and the status type is in some respects doing the type check for us. It also prevents us having to write complex type checking functions which I'm not sure are generally worth the effort? Happy to discuss further if you do think this is indeed a bit sneaky, I'm just keen to keep tests as easy to write as possible!

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-report.test.ts
```diff
@@ -0,0 +1,53 @@
...
+    const { req, res } = getHttpMocks();
+
+    // Act
+    await getUtilisationReportById(req, res);
+
+    // Assert
+    expect(res.statusCode).toEqual(200);
+    // eslint-disable-next-line no-underscore-dangle
+    const responseData = res._getData() as GetUtilisationReportResponse;
+    expect(responseData.id).toEqual(mockUtilisationReport.id);
```
ChristianMcc, line 38: can we do anything interesting like:
expect(expect.ObjectContaining(responseData)).toEqual(mockUtilisationReport)? Assuming this is just mockUtilisationReport has more properties than response data. or do you think that makes it harder to maintian in the future?

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-reports.controller/index.test.ts
```diff
@@ -135,24 +140,75 @@ describe('getUtilisationReports', () => {
...
+    expect(findAllByBankIdMock).toHaveBeenCalledWith(bankId, {
+      reportPeriod: validReportPeriod,
+      excludeNotReceived: true,
+    });
+
+    expect(res.statusCode).toEqual(200);
+    // eslint-disable-next-line no-underscore-dangle
+    const responseData = res._getData() as GetUtilisationReportResponse[];
+    expect(responseData.length).toEqual(1);
+    expect(responseData[0].id).toEqual(mockUtilisationReports[0].id);
```
ChristianMcc, line 199: ditto to other comment, although I'd imagine it this might just be the best way and is quite simple

libs/common/src/test-helpers/mock-data/utilisation-report.entity.mock-builder.ts
```diff
@@ -66,7 +66,7 @@ export class UtilisationReportEntityMockBuilder {
     return this;
   }
 
-  public withAzureFileInfo(azureFileInfo: AzureFileInfoEntity): UtilisationReportEntityMockBuilder {
+  public withAzureFileInfo(azureFileInfo: AzureFileInfoEntity | undefined): UtilisationReportEntityMockBuilder {
```
ChristianMcc, line 69: is this for like 'marked as completed' reports? wondering when we would want to call a test with no azure file info aka why it accepts undefined - probably more for my own understanding

portal-api/api-tests/v1/utilisation-report-service/last-uploaded.api-test.js
```diff
@@ -36,47 +30,36 @@ describe('GET /v1/banks/:bankId/utilisation-reports/last-uploaded', () => {
...
-            month: lastUploadedReportPeriodMonth + 1,
-            year,
-          },
-          end: {
-            month: lastUploadedReportPeriodMonth + 1,
-            year,
-          },
+    lastUploadedReport.bankId = bank.id;
+    const notReceivedReport = UtilisationReportEntityMockBuilder.forStatus(UTILISATION_REPORT_RECONCILIATION_STATUS.REPORT_NOT_RECEIVED)
+      .withId(6)
```
ChristianMcc, line 47: Is this just us needing to add an id? would we want to just let the DB generate this or are we trying to override so that we can refer to it later?

ChristianMcc (commented):
Lovely changes and some nice improvements to tests! Mostly just left some questions that probably dont require action :)

libs/common/src/test-helpers/mock-data/utilisation-report.entity.mock-builder.ts
JoshBinns2000: This is specifically for "marked as completed" reports (replied because I've added this in my PR as well)

RLavender98: Yeah that's right, the mock builder by default adds azure file info if the status you give it is anything other than REPORT_NOT_RECEIVED. So if you ever want to test anything without azure file info and a status other than REPORT_NOT_RECEIVED you need to explicitly set it to undefined.

portal-api/api-tests/v1/utilisation-report-service/last-uploaded.api-test.js
RLavender98: So, the mock builder adds the id as 1, so if i am writing a test with more than one entity, I need to set the ids for each entity to different values, otherwise when I save the entities after the first one it doesn't add a second entity but it also didn't throw as big an error as I would have liked it to 😅

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-report.test.ts
RLavender98: The other way around like so: `expect(mockUtilisationReport).toEqual(expect.objectContaining(responseData));` gets close to working, however it doesn't deal with the nested objects also just being subsets, which is the case for the azure file info, so I think we would have to do something even more complex at which point I quite like the simplicity of each of these lines even though it is more lines (which one might forget to add to in the future if adding fields, which is a concern I guess but I don't think it is too big of one).

ChristianMcc (approved):

PR #2881, opened 2024-03-08
feat(DTFS2-6965): update api call and docker compose

ChristianMcc:
## Introduction :pencil2:
The replica set was causing intermittent issues for developers over time so should be moved from the default. We also needed to update the content-type to text/plain as per the data team's request

Create a new docker-compose.replica-set.yml file. I looked into switching the replica set on/off based off an environment variable but it was not something that can be done without it getting messy as docker-compose don't support top level if statements


dtfs-central-api/src/services/changeStream/changeStreamApi.ts
```diff
@@ -13,17 +13,15 @@ export const postAuditDetails: (changeStreamDocument: ChangeStreamDocument) => P
...
-  ) {
+  // TODO DTFS2-6967 enable delete functionality
+  if (changeStreamDocument.operationType !== 'insert' && changeStreamDocument.operationType !== 'update' && changeStreamDocument.operationType !== 'replace') {
     console.info('Change stream document is not suitable event for audit API, skipping');
     return;
   }
 
-  const fullDocument = changeStreamDocument.operationType !== 'delete' ? changeStreamDocument.fullDocument : null;
+  // TODO DTFS2-6967 enable delete functionality
+  // const fullDocument = changeStreamDocument.operationType !== 'delete' ? changeStreamDocument.fullDocument : null;
```
abhi-markan, line 23: Can we remove the commented code and move it to the ticket please.

docker-compose.dev.yml
```diff
@@ -7,22 +7,14 @@ services:
       - '3310:3310'
```
abhi-markan, line 7: Nice update, I would suggest removing `docker-compose.dev.yml` and moving it to `docker-compose.yml`.

abhi-markan (changes requested):

docker-compose.dev.yml
ChristianMcc: Sure, you happy for me to remove all the resource constraints? will update readme with this too

abhi-markan (approved):

PR #2793, opened 2024-02-23
feat(FN-1859): refactor upload report logic to SQL

toddym42:
## Introduction :pencil2:
We want to refactor the uploading report logic to use the SQL database.

## Resolution :heavy_check_mark:
- Updates the state machine event logic to upload reports
- Adds a `UtilisationReportEntityMockBuilder` class
- Adds new and updates existing tests

PR #2883, opened 2024-03-11
feat(FN-1862): refactor TFM mark report as completed

JoshBinns2000:
## Introduction :pencil2:
We need to update the mark report as completed functionality to use the SQL database.

- Introduces a `UtilisationDataMockEntityBuilder` to `libs/common`
- Introduces an `SqlDbHelper` (`sql-db-helper.ts`) api test helper to `dtfs-central-api`
- Adds more controller tests for `put-utilisation-report-status.controller`


libs/common/src/test-helpers/mock-data/utilisation-data.entity.mock-builder.ts
```diff
@@ -0,0 +1,99 @@
...
+    this.data.facilityUtilisation = facilityUtilisation;
+    return this;
+  }
+
+  public withTotalFeesAccruedForTheMonth(totalFeesAccruedForTheMonth: number): UtilisationDataEntityMockBuilder {
+    this.data.totalFeesAccruedForTheMonth = totalFeesAccruedForTheMonth;
+    return this;
+  }
+
+  public withTotalFeesAccruedForTheMonthCurrency(totalFeesAccruedForTheMonthCurrency: Currency): UtilisationDataEntityMockBuilder {
```
JoshBinns2000, line 66: Update this and the below to just have the input be `currency` or `exchangeRate`

.github/workflows/test.yml
```diff
@@ -43,6 +43,7 @@ on:
     branches:
       - main
       - feat/e2etf-sql
+      - feat/FN-1859/upload-report-sql
```
JoshBinns2000, line 46: This will be removed after #2793 has been merged

dtfs-central-api/api-tests/sql-db-helper.ts
```diff
@@ -0,0 +1,63 @@
+import { AzureFileInfoEntity, UtilisationDataEntity, UtilisationReportEntity } from '@ukef/dtfs2-common';
+import { SqlDbDataSource } from '@ukef/dtfs2-common/sql-db-connection';
+
+const initialize = async () => await SqlDbDataSource.initialize();
+
+type SqlTableName = 'UtilisationReport' | 'UtilisationData' | 'AzureFileInfo';
```
JoshBinns2000, line 6: There is a lot of logic that typeorm handles for us in terms of cascade updates, deletes, upserts etc. I have defined the helper functions in here to work on one entity at a time so that we don't try to "outsmart" typeorm and let it handle most of the logic

dtfs-central-api/src/errors/transaction-failed.error.ts
```diff
@@ -0,0 +1,11 @@
+import { HttpStatusCode } from 'axios';
+import { ApiError } from '.';
+
+export class TransactionFailedError extends ApiError {
+  constructor(message: string = 'Transaction failed') {
+    super({
+      message,
+      status: HttpStatusCode.InternalServerError,
```
JoshBinns2000, line 8: Because the transaction could have failed for a variety of reasons, I think a `500` is appropriate here as, hopefully, any issues related to the shape of the request have already been validated before the transaction starts. Any thoughts?

dtfs-central-api/src/v1/controllers/utilisation-report-service/put-utilisation-report-status.controller/index.test.ts
```diff
@@ -0,0 +1,302 @@
...
+
+      existingReports.forEach((report) => {
+        report.setAsNotCompleted({ requestSource });
+        expect(mockRepository.save).toHaveBeenCalledWith(report);
+      });
+    });
+
+    const reportStatusesToTest: UtilisationReportReconciliationStatus[] = ['PENDING_RECONCILIATION', 'REPORT_NOT_RECEIVED'];
+
+    it.each(reportStatusesToTest)("responds with an error if trying to mark a report with status '%s' as not completed", async (reportStatus) => {
```
RLavender98, line 269: These statuses are very tied to this test case so could we just inline them in the each?

dtfs-central-api/src/services/state-machines/utilisation-report/utilisation-report.state-machine.test.ts
RLavender98, line 1: I feel we should probably add unit tests on this level checking for invalid combinations of report status and event type and asserting they get rejected.

libs/common/src/sql-db-entities/utilisation-report/utilisation-report.entity.ts
```diff
@@ -79,4 +79,14 @@ export class UtilisationReportEntity extends AuditableBaseEntity {
...
     this.updatedByUserId = getDbAuditUpdatedByUserId(requestSource);
   }
+
+  public setAsCompleted({ requestSource }: { requestSource: DbRequestSource }): void {
+    this.status = 'RECONCILIATION_COMPLETED';
+    this.updatedByUserId = getDbAuditUpdatedByUserId(requestSource);
+  }
+
+  public setAsNotCompleted({ requestSource }: { requestSource: DbRequestSource }): void {
+    this.status = this.azureFileInfo ? 'PENDING_RECONCILIATION' : 'REPORT_NOT_RECEIVED';
```
RLavender98, line 89: I think it would be better for this choice of status to be handled explicitly in the state machine, and keep the methods on the entity class to `setAsState` or `setStatus` which accepts a status parameter. This feels slightly hidden to me and I think it would make sense for the state machine event handler that gets called to handle the business logic around states, especially as we build out logic around further state transitions with more complex interactions.

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/manually-set-incomplete/manually-set-incomplete.event-handler.ts
```diff
@@ -1,12 +1,18 @@
...
 
-export const handleUtilisationReportManuallySetIncompleteEvent = (
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+export const handleUtilisationReportManuallySetIncompleteEvent = async (
   report: UtilisationReportEntity,
+  { requestSource, transactionEntityManager }: ManuallySetIncompleteEventPayload,
 ): Promise<UtilisationReportEntity> => {
-  throw new NotImplementedError('TODO FN-1862');
+  report.setAsNotCompleted({ requestSource });
+  return await transactionEntityManager.getRepository(UtilisationReportEntity).save(report);
```
RLavender98, line 17: Hmm is there a way we can call our custom repo methods within the context of the transaction? If so could we use that repo here so we aren't calling `getRepository` directly other than in repos and in tests?

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/manually-set-incomplete/manually-set-incomplete.event-handler.test.ts
```diff
@@ -1,14 +1,70 @@
...
+  };
+  const updatedByUserId = getDbAuditUpdatedByUserId(requestSource);
+
+  const mockSave = jest.fn();
+
+  const mockRepository = {
+    save: mockSave,
+  };
+
+  it(`calls the correct repo methods and sets the report to '${UTILISATION_REPORT_RECONCILIATION_STATUS.PENDING_RECONCILIATION}' when the report was previously uploaded`, async () => {
```
RLavender98, line 26: It is generally best to avoid words like 'correct' in test names and instead explicitly spell out what we mean by correct because if someone is not familiar with the context 'correct' doesn't really tell them anything. I think here for example `sets the report to '${UTILISATION_REPORT_RECONCILIATION_STATUS.PENDING_RECONCILIATION}' when the report was previously uploaded and saves using the entity manager` would perhaps be a better name.

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/manually-set-completed/manually-set-completed.event-handler.test.ts
```diff
@@ -1,14 +1,45 @@
...
+  };
+  const updatedByUserId = getDbAuditUpdatedByUserId(requestSource);
+
+  const mockSave = jest.fn();
+
+  const mockRepository = {
+    save: mockSave,
+  };
+
+  it('calls the correct repo methods and updates the report', async () => {
```
RLavender98, line 24: Again same comment re correct, but also here we should also spell out what we mean by 'updates'.

RLavender98 (changes requested):
Looks good, just a few things around where logic lives/test writing and would be good to get @ChristianMcc's eyes on the transaction handling.

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/manually-set-incomplete/manually-set-incomplete.event-handler.ts
JoshBinns2000: Unfortunately not. I'm going to leave a long reply comment below regarding the options we have with respect to transactions in typeorm, but the "query runner" approach requires you to use the entity manager supplied by the query runner instance - if we used `UtilisationReportRepo` here, the query would run outside the context of the transaction

dtfs-central-api/src/services/state-machines/utilisation-report/utilisation-report.state-machine.test.ts
JoshBinns2000: I'm pretty sure that test is performed somewhere in this file (or at least near here) - I'll find the PR where those tests were added later on and link that here :)

dtfs-central-api/src/v1/controllers/utilisation-report-service/put-utilisation-report-status.controller/index.test.ts
```diff
@@ -0,0 +1,302 @@
...
+        status: 'REPORT_NOT_RECEIVED',
+      };
+
+      const { req, res } = getHttpMocks();
+      req.body.reportsWithStatus = [reportWithStatus];
+
+      const existingReport = UtilisationReportEntityMockBuilder.forStatus(reportStatus).withId(reportWithStatus.reportId).build();
+
+      const existingReportStateMachine = UtilisationReportStateMachine.forReport(existingReport);
+      jest.spyOn(UtilisationReportStateMachine, 'forReportId').mockImplementation(() => Promise.resolve(existingReportStateMachine));
```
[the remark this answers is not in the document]
JoshBinns2000: The reason I think it is better here to directly mock the return of the state machine is that it is easier to trace what you are mocking by looking at the controller. If you mock the `UtilisationReportRepo.fineOneByOrFail` function, it isn't immediately obvious that this is going to affect that call to `UtilisationReportStateMachine.forReportId`. That being said, I don't have a strong opinion on this so would be good to see what you think as well @ChristianMcc

libs/common/src/test-helpers/mock-data/utilisation-data.entity.mock-builder.ts
```diff
@@ -0,0 +1,99 @@
...
+import { Currency } from '../../types';
+
+export class UtilisationDataEntityMockBuilder {
+  private readonly data: UtilisationDataEntity;
+
+  private constructor(data: UtilisationDataEntity) {
+    this.data = data;
+  }
+
+  public static forReport(report: UtilisationReportEntity): UtilisationDataEntityMockBuilder {
```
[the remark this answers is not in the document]
JoshBinns2000: A row in the utilisation data (soon to be fee record) table will always have a foreign key linking it to a report, so we will always need a report attached to the utilisation data entity. Considering that, I don't think there will be a case where we are performing any logic that only mocks utilisation data without needing to link it to a specific report, which was my thinking when creating this mock builder

dtfs-central-api/src/v1/controllers/utilisation-report-service/put-utilisation-report-status.controller/index.test.ts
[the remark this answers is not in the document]
JoshBinns2000: This is a unit test

dtfs-central-api/src/v1/controllers/utilisation-report-service/put-utilisation-report-status.controller/index.ts
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    return res.sendStatus(HttpStatusCode.Ok);
+  } catch (error) {
+    console.error('Error updating utilisation report statuses:', error);
+    if (error instanceof ApiError) {
+      return res.status(error.status).send(`Failed to update utilisation report statuses: ${error.message}`);
+    }
+    return res.status(HttpStatusCode.InternalServerError).send(`Failed to update utilisation report statuses`);
+  }
```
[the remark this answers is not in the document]
RLavender98: > I don't think there is a way for us to get around this, so it is unfortunately something we just need to accept whenever we perform transactions

I think we will probably want to have quite a lot of stuff using transactions and we need a solution that works with our repo objects, even if that means adapting our repo objects. As I say I'm not an expert on this by any means but I do think it should be possible to come up with a pattern that works nicely with transactions or I'm sure typeORM would not be as popular as it is.

This [article](https://medium.com/@dev.muhammet.ozen/advanced-transaction-management-with-nestjs-typeorm-43a839363491) is sadly using nest.js so the code has to be viewed in that context, it obviously won't look identical for us at all, but looking at the start of the advanced section under the repositories header, they actually pass a dataSource to their repo constructor. Perhaps we can take some inspiration from here to adapt our repo implementation?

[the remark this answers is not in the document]
JoshBinns2000: The way we define our repository is just to extend the repository which exists on our `SqlDbDataSource` connection, so you can do the same thing with the transaction entity manager:

```typescript
const UtilisationReportRepo = transactionEntityManager.manager.getRepository(UtilisationReportEntity).extend({ ... });
```

However, as we don't need it right now, I'd argue YAGNI may apply as we may not need our more complex repo methods for any transactions that we perform

libs/common/src/test-helpers/mock-data/utilisation-report.entity.mock-builder.ts
```diff
@@ -71,6 +71,11 @@ export class UtilisationReportEntityMockBuilder {
     return this;
   }
 
+  public withoutAzureFileInfo(): UtilisationReportEntityMockBuilder {
+    this.report.azureFileInfo = undefined;
+    return this;
+  }
```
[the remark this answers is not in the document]
JoshBinns2000: My main problem with `.withAzureFileInfo(azureFileInfo: AzureFileInfoEntity | undefined)` is that calling `.withAzureFileInfo()` will remove the azure file info entity currently attached to the object, but we have done that by calling a method which says "with". It would be like asking for a burger with ketchup, being served a burger without ketchup and being told "you didn't specify how much ketchup, so I assumed you meant 0" 😅 

@ChristianMcc another one where it may just be better to get your opinion. The two options we have come up with are (inside `UtilisationReportEntityMockBuilder`)
```typescript
withAzureFileInfo(azureFileInfo: AzureFileInfoEntity | undefined) {
  this.report.azureFileInfo = azureFileInfo;
}
```
or
```typescript
withAzureFileInfo(azureFileInfo: AzureFileInfoEntity) {
  this.report.azureFileInfo = azureFileInfo;
}

withoutAzureFileInfo() {
  this.report.azureFileInfo = undefined;
}
```

e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/pdc-reconcile/mark-report-as-done.spec.js
```diff
@@ -149,7 +149,10 @@ context(`${PDC_TEAMS.PDC_RECONCILE} users can mark reports as done and not done`
     pages.utilisationReportsPage.clickMarkReportAsCompletedButton(submissionMonth);
 
     cy.get(aliasSelector(utilisationReportsAlias)).each((utilisationReport) => {
-      const { id, bankId } = utilisationReport;
+      const { bankId, id } = utilisationReport;
+      if (!statusWithBankId.find(({ bankId: bankIdToMatch }) => bankId === bankIdToMatch)) {
+        return;
+      }
```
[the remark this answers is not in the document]
JoshBinns2000: `return` will just move to the next iteration of the loop here.

This is done as a bit of a hack - I wanted to only test 3 reports as there are only three possibilities when we mark reports as completed or not completed (at the moment). I have an array defined at the top of the test called `statusWithBankId` which has the three specific bank ids with status that I want to test. However, this loop iterates over all the banks, so I check to see whether or not the current bank is one of the banks I want to test. If it isn't, I just return early

libs/common/src/test-helpers/mock-data/utilisation-report.entity.mock-builder.ts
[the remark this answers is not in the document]
RLavender98: Just to be clear:
Screenshot 2024-03-14 at 15 47 49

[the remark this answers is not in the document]
JoshBinns2000: Ah nice - I didn't realise typescript would do that. In that case, let's go with `withAzureFileInfo` and I'll update that in this PR

dtfs-central-api/src/services/state-machines/utilisation-report/utilisation-report.state-machine.test.ts
JoshBinns2000: If you go to the bottom of each `describe('when the report is in <status>` blocks in `utilisation-report.state-machine.test.ts` you see that the invalid combinations of report status and event type are tested

dtfs-central-api/src/v1/controllers/utilisation-report-service/put-utilisation-report-status.controller/index.test.ts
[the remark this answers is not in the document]
JoshBinns2000: Discussed: keeping the mocking to only the database interactions is going to be easier to maintain and keep track of, and it also gives us some integration testing of the state machine for free

dtfs-central-api/src/v1/controllers/utilisation-report-service/put-utilisation-report-status.controller/index.test.ts
```diff
@@ -113,10 +100,19 @@ describe('put-utilisation-report-status.controller', () => {
     manager: mockTransactionManager,
   } as unknown as QueryRunner;
 
+  const utilisationReportRepoFindOneByOrFailSpy = jest.spyOn(UtilisationReportRepo, 'findOneByOrFail');
+
+  const mockFindOneByOrFail = (reports: UtilisationReportEntity[]) => (where: Parameters<typeof UtilisationReportRepo.findOneByOrFail>[0]) => {
```
RLavender98, line 105: Could we put this and the consts above at the top of their describe block before any tests? It's nice to keep them all in one place.

dtfs-central-api/src/v1/controllers/utilisation-report-service/put-utilisation-report-status.controller/index.ts
```diff
@@ -0,0 +1,97 @@
...
+}>;
+
+const assertReportWithStatusIsPopulated = (reportWithStatus: ReportWithStatus) => {
+  const { status, reportId } = reportWithStatus;
+  if (!status || !reportId) {
+    throw new InvalidPayloadError("Request body item 'reportsWithStatus' supplied does not match required format");
+  }
+};
+
+const executeEventHandlerWithQueryRunner = async (
```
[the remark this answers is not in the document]
JoshBinns2000: Waiting to see what happens with the below transaction code before updating this, but I agree that this function doesn't need the `WithQueryRunner` part (this function used to take a `queryRunner` as an input argument hence the name)

dtfs-central-api/src/v1/controllers/utilisation-report-service/put-utilisation-report-status.controller/index.test.ts
JoshBinns2000: I usually prefer to define them as they are needed as I think it makes it easier to read

RLavender98: I actually find it harder to read 😅. I think it is more likely to cause confusion written this way. I rarely read a test file in order except when reviewing a freshly written test file and I would expect to see the test set up and mocking at the top of the relevant describe block rather than find it sprinkled throughout.

RLavender98: I actually wonder if we want to assert in the above tests that we don't actually call the state machine and update the statuses?

Also, if the failure case test breaks and the invalid request is not caught correctly, we don't actually want to then be calling all the way through to the database. In this specific case nothing bad will happen, but I think that is at least part of the reason the convention is to put one's set up at the top.

JoshBinns2000: I think this is just a difference in opinion then as I almost always read test files in order. The compromise here I think it to wrap this part in another `describe('when the payload is valid', ...)` block which distinguishes it from the other tests, so then these mocks can be defined at the top of that block and it doesn't appear to be randomly spread out

RLavender98: This isn't a difference in opinion, it is actually best practice to put these mocks at the top. In fact I've just been thinking about this more and the mocks you set up halfway down the tests are actually being set up before the test runs in most likeliness, so reading the top of this file you would have no idea the first two tests have mocks in them but they do!

If you run a test file with the following (as I have just done to make sure I am not in fact wrong 😅 ) you should get one, then three, then two printed to the console.

```
console.log('ONE');

it('fake test', () => {
  console.log('TWO');
})

console.log('BANANA THREE');
```
If you really do want to go with the multiple describe blocks (which I think is unnecessary) then please put the invalid payload ones inside an 'when the payload is invalid' describe block also.

JoshBinns2000: Yes, `jest` runs tests asynchronously (technically, concurrently).

I haven't seen this listed anywhere as a best-practice, and the concern about the later-defined values being available is mitigated somewhat by the `"no-use-before-define"` eslint rule. However, if we are going to have every single mock defined at the top, then the tests should really be using those mocks. That's not a problem in this case as we can just test the the query runner is never initialised

dtfs-central-api/src/v1/controllers/utilisation-report-service/put-utilisation-report-status.controller/index.ts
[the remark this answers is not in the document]
ChristianMcc: Originally when I looked at this I mostly thought 'this is a one off transaction thats going to go in the bin soon' hence why I was quite relaxed to just use a queryRunner and not worry too much. I still can't tell in TFM if we are going to have a tonne or almost no transactions aha - although this final part you're describing Josh, is this just literally creating a new repo extension on the transactions version of the UtilisationReportEntity that adds all our existing ones on?
Have you had a chance to test if this works etc.?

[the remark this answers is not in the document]
JoshBinns2000: Yes - this is just creating a new repo extension, but using all our methods that we have already defined on the original `UtilisationReportRepo`. I haven't tested it at runtime, but typescript sees all the methods defined on our `UtilisationReportRepo` in `TransactionRepo` (as defined above), so I don't see why it wouldn't work at runtime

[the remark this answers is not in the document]
ChristianMcc: Interesting! For this place it would be nice to:
- Pull all the transaction related stuff into a separate function, where that lives might be a bit hard, I'd expect it to be in something like a utilisation-report-service but we don't really have services.
- That function would start the transaction and then extend to have the extra repo classes
- Calls whatever repo methods it needs

I'm assuming that we can't have a constant transaction manager that is extended in methods because each 'transactionManager' will be a different one that needs to be extended individually.

We could have  amethod as you said in the UtilisationReportRepo that takes in a transactionEntityManager, extends it, and returns the new transactionManager which might be nice!

Thoughts from anyone?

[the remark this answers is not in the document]
ChristianMcc: Had a small discussion, for now can we just extract the transaction style logic to a separate function and we can revisit if we end up having a lot of transactions coming in in TFM release 6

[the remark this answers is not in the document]
JoshBinns2000: Just as a closing comment, I've also added some docstrings to all the functions in this file

ChristianMcc (approved):
LGTM (stealing Ruby's green tile while she's off project but looks like all conversations are resolved!)

PR #2888, opened 2024-03-12
feat(FN-1853): run api tests outside container

JoshBinns2000:
## Introduction :pencil2:
We want to see if we can run api tests outside the docker container.

- Updates `test.yml` to run api tests outside the docker container.


abhi-markan (changes requested):
PR description does not give any explanation, why we need to run the `api-tests` outside the container?

PR #2889, opened 2024-03-12
chore(DTFS2): setup teams message on push to main

ChristianMcc:
## Introduction :pencil2:
For release management we want to track changes made to main to make it easier for non technical users to see when updates to main are made.

Add a new github action that uses a microsoft teams webhook to message a teams channel with a simple payload of commit details


.github/workflows/notify_ms_teams.yml
```diff
@@ -0,0 +1,72 @@
+name: Notify Teams on Push
```
abhi-markan, line 1: Can we follow the stands from `test.yml` or `sca.yml` please:
* File name one word `notify.yml`.
* Consistent step names from `test.yml` please.
* Ensure timezone, environment is setup in the similar manner please.

.github/workflows/notify_ms_teams.yml
```diff
@@ -0,0 +1,72 @@
...
+
+      - name: Extract Commit Information
+        id: commit_info
+        run: |
+          COMMIT_MESSAGE=$(git log --format=%B -n 1 ${{ github.sha }})
+          COMMIT_SHA=$(git log --format=%H --abbrev=7 -n 1 ${{ github.sha }})
+          COMMIT_TIMESTAMP=$(git log --format=%ci -n 1 ${{ github.sha }})
+          COMMIT_AUTHOR=${{ github.actor }}
+
+          echo "Commit Message: $COMMIT_MESSAGE"
```
abhi-markan, line 28: Do we need to echo the response,  provided they are being stored in env variable?

.github/workflows/notify_ms_teams.yml
```diff
@@ -0,0 +1,72 @@
...
+          echo "Commit SHA: $COMMIT_SHA"
+          echo "Commit Author: $COMMIT_AUTHOR"
+          echo "Commit Timestamp: $COMMIT_TIMESTAMP"
+
+          echo "COMMIT_MESSAGE=$COMMIT_MESSAGE" >> $GITHUB_ENV
```
abhi-markan, line 33: Not a blocker, would it help to reduce this to single step which then will not require any env storage?

.github/workflows/notify_ms_teams.yml
```diff
@@ -0,0 +1,72 @@
...
+          echo "COMMIT_SHA=$COMMIT_SHA" >> $GITHUB_ENV
+          echo "COMMIT_AUTHOR=$COMMIT_AUTHOR" >> $GITHUB_ENV
+          echo "COMMIT_TIMESTAMP=$COMMIT_TIMESTAMP" >> $GITHUB_ENV
+
+      - name: Notify Teams
+        env:
+          MICROSOFT_TEAMS_WEBHOOK: ${{ secrets.MICROSOFT_TEAMS_WEBHOOK  }}
+        run: |
+          curl -g -X POST -H 'Content-Type: application/json' -d '{
```
abhi-markan, line 42: Nice native approach.

abhi-markan (changes requested):

.github/workflows/notify_ms_teams.yml
ChristianMcc: I can remove this I just thought this would be helpful logging to have within this action

.github/workflows/notify_ms_teams.yml
ChristianMcc: I wanted to split up the steps into readable sections here (even if it meant using env storage!)

.github/workflows/notify_ms_teams.yml
ChristianMcc: I will update some to be repository etc.

For the custom `extract commit information` I could just rename to extract but it definitely takes away some of the context, are we okay with having multiple words in a step name?

.github/workflows/notify_ms_teams.yml
abhi-markan: Makes sense, thanks.

.github/workflows/notify_ms_teams.yml
abhi-markan: Yes please, if we can remove it to keep the file concise.

.github/workflows/notify_ms_teams.yml
abhi-markan: Yes please one word, can add some documentation about steps for context.
Reason I ask is when the pipeline is observed on VSCode multi words are concatenated.

abhi-markan (approved):

abhi-markan (dismissed):

AlexBramhill (approved):

AlexBramhill:
@ChristianMcc Merged as all tests passing, and didn't want to risk having to rerun again.

PR #2890, opened 2024-03-13
feat(FN-2510): refactor schedule job to use SQL

JoshBinns2000:
## Introduction :pencil2:
We want the job to create reports to query the SQL database instead of mongo.

- Updates the job to query the SQL database
- Updates the unit tests


dtfs-central-api/src/scheduler/jobs/create-utilisation-reports/index.test.ts
```diff
@@ -146,26 +155,33 @@ describe('scheduler/jobs/create-utilisation-reports', () => {
...
+      expect(saveUtilisationReportSpy).toHaveBeenCalledTimes(1);
+
+      const newReportForBankWithoutReport = UtilisationReportEntity.createNotReceived({
+        bankId: bankWithoutReport.id,
+        reportPeriod: mockReportPeriod,
+        requestSource: {
+          platform: 'SYSTEM',
+        },
+      });
+      expect(saveUtilisationReportSpy).toHaveBeenCalledWith(newReportForBankWithoutReport);
```
RLavender98, line 184: I'm slightly surprised this works as I was under the impression this checked for the same object instance being passed in rather than doing field by field comparison... Do you know why this works?

I'm also just wondering if we have coverage anywhere that this `createNotReceived` method does what we expect? This would possibly be the place I would expect us to assert those things if it is not otherwise covered? (e.g. checking the status of the created entity is the `REPORT_NOT_RECEIVED` status rather than just using the mainline code creation method which will therefore always match)

JoshBinns2000: #2883 has added some test coverage for methods like `UtilisationReportEntity.createNotReceived`, but you are right that it does not exist at present

That is a good point though about the test passing with comparing different class instances. I'll have a bit of a play around with that and see if I can find what jest is doing here (I know that `.toEqual` does quite a bit under to hood which is why it gets used in place of `.toBe` which I _think_ functions more like `===`)

RLavender98: Great, I'll take a look at that review next 😄 

It's probably not worth sinking too much time into looking into it so if the answer doesn't come easy don't worry about it, but do let me know if you find out!

JoshBinns2000: Just to link this to `.toHaveBeenCalledWith`, the [docs](https://jestjs.io/docs/expect#tohavebeencalledwitharg1-arg2-) say it uses the same algorithm as `.toEqual` which was why I was talking about that 😄

RLavender98 (approved):
Approving, under the expectation a test is added for `createNotReceived` on the FN-1862 review

PR #2893, opened 2024-03-13
feat(DTFS2-6965): iterate over collections in database to set up changestream

ChristianMcc:
## Introduction :pencil2:
On the azure cosmosdb it's not running a version of mongo that can watch the whole database

Update the code to iterate over collections in the database and watch them individually

### DIFF
--- a/.github/workflows/feature_deploy.yml
+++ b/.github/workflows/feature_deploy.yml
@@ -52,13 +52,13 @@ jobs:
       max-parallel: 2
       matrix:
         repository:
-          - portal-ui
-          - portal-api
-          - gef-ui
-          - trade-finance-manager-ui
-          - trade-finance-manager-api
+          # - portal-ui
+          # - portal-api
+          # - gef-ui
+          # - trade-finance-manager-ui
+          # - trade-finance-manager-api
           - dtfs-central-api
-          - external-api
+          # - external-api
         include:
           - path: portal
             repository: portal-ui
--- a/.github/workflows/feature_deploy.yml
+++ b/.github/workflows/feature_deploy.yml
@@ -6,7 +6,6 @@ on:
   push:
     branches:
       - feat/e2etf-sql
-      - feat/DTFS2-6965/test-change-stream-in-azure
     paths:
       - "portal/**"
       - "portal-api/**"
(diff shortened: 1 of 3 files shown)


dtfs-central-api/src/services/changeStream/setupChangeStream.ts
```diff
@@ -1,19 +1,43 @@
...
  */
-export const setupChangeStream = async () => {
-  console.info('Setting up mongodb change stream');
-  const database = await getClient();
-  const changeStream = database.watch([], { fullDocument: 'updateLookup' });
+const setupChangeStreamForCollection = async (collectionName: string) => {
+  console.info('Setting up change stream for collection', collectionName);
+  const databaseConnection = await getConnection();
+  const changeStream = (databaseConnection.collection(collectionName)).watch(
+    [{ $match: { operationType: { $in: ['insert', 'update', 'replace'] } } }, { $project: { _id: 1, fullDocument: 1, ns: 1, documentKey: 1 } }],
```
MarRobSoftwire, line 14: Do we want to track `'delete'` too?
Might need handling differently in the `$project` if the `fullDocument` no longer exists

dtfs-central-api/src/services/changeStream/setupChangeStream.ts
```diff
@@ -1,19 +1,43 @@
...
     postAuditDetails(changeStreamDocument).catch((error) => {
       console.error('Error sending change stream update to API', error);
     });
   });
-  console.info('Mongodb change stream initialised');
+};
+
+/**
+ * Sets up a change stream on the mongodb database and sends any changes to the audit API
+ * @returns
```
MarRobSoftwire, line 27: Is this line needed?

.github/workflows/feature_deploy.yml
```diff
@@ -48,16 +49,16 @@ jobs:
...
-      max-parallel: 4
+      max-parallel: 2
       matrix:
         repository:
-          - portal-ui
-          - portal-api
-          - gef-ui
-          - trade-finance-manager-ui
-          - trade-finance-manager-api
+          # - portal-ui
```
MarRobSoftwire, line 55: Not entirely sure what this is doing but I'm not too familar with github actions yaml

dtfs-central-api/src/services/changeStream/setupChangeStream.ts
ChristianMcc: Yeah I've removed for now, we'll probably handle delete with updating with a TTL but I've got a separate ticket now for how we can handle delete functions
Biggest issue is if we delete normally we can't have the document which will have the details of who deleted it :(

.github/workflows/feature_deploy.yml
```diff
@@ -48,7 +48,7 @@ jobs:
     # environment:
     #   name: ${{ needs.deployment-environment.outputs.environment }}
     strategy:
-      max-parallel: 4
+      max-parallel: 2
```
ChristianMcc, line 51: This is because it gets sad if you try to many at the same time so I've reduced it to 2 for now

MarRobSoftwire (approved):

PR #2896, opened 2024-03-14
feat(DTFS2-7053): include user audit details in TFM api

MarRobSoftwire:
## Introduction :pencil2:
All endpoints in tfm-api should add `auditDetails` when inserting/updating data.

- Add audit details to each database change

## Misc ➕ 
- Remove obsolete gef controller


trade-finance-manager-api/src/v1/controllers/feedback-controller.js
```diff
@@ -38,7 +38,11 @@ exports.create = async (req, res) => {
     howCanWeImprove,
     emailAddress,
     submittedBy,
-    created: getUnixTime(new Date())
+    created: getUnixTime(new Date()),
+    auditDetails: {
```
MarRobSoftwire, line 42: There might not be any user logged in when submitting feedback, I have a few ideas on what to do here:
- set lastUpdatedBy: null
- set `lastUpdatedBySystem: true`
 - Have an exemption on this collection to exclude it from the change stream

trade-finance-manager-api/src/v1/controllers/gef.controller.js
MarRobSoftwire, line 1: This file was never used

trade-finance-manager-api/src/v1/controllers/acbs.controller.js
MarRobSoftwire, line 1: `clearACBSLog` is only used in tests - should this be moved to a helper?

trade-finance-manager-api/src/v1/controllers/user/user.controller.js
```diff
@@ -28,11 +28,18 @@ exports.findByUsername = async (username, callback) => {
...
+ * @param {import('../../../types/tfm-session-user').TfmSessionUser | undefined} sessionUser logged in user
+ * @param {(error: string | null, createdUser: object) => void} callback
+ * @returns 
+ */
+exports.create = async (user, sessionUser, callback) => {
   const collection = await db.getCollection('tfm-users');
   const tfmUser = {
     ...user,
     status: USER.STATUS.ACTIVE,
+    auditDetails: sessionUser?._id ? generateTfmUserAuditDetails(sessionUser._id) : null,
```
MarRobSoftwire: If no user is logged in I've set this to null - not entirely happy this is the best option

trade-finance-manager-api/src/constants/payloads.js
MarRobSoftwire, line 1: This file feels like it could be so much better if it was Typescript (alongside the corresponding payloads.js that does the validation)

trade-finance-manager-api/src/v1/controllers/user/user.controller.js
```diff
@@ -54,12 +62,20 @@ exports.create = async (user, callback) => {
...
+ * @param {import('../../../types/tfm-session-user').TfmSessionUser | undefined} sessionUser logged in user
+ * @param {(error: string | null, updatedUser: object) => void} callback
+ */
+exports.update = async (_id, update, sessionUser, callback) => {
   if (!ObjectId.isValid(_id)) {
     throw new Error('Invalid User Id');
   }
 
-  const userUpdate = { ...update };
+  const userUpdate = produce(update, (draftUpdate) => {
```
MarRobSoftwire, line 76: I don't know why this was cloned, but if it does it should be done properly

trade-finance-manager-api/src/v1/controllers/feedback-controller.js
ChristianMcc: Do we have access to the user if they are logged in and leave feedback? Wonder if we should do our best to populate if they are logged in and it's possible, and then resort to the `noUserLoggedIn` solution if not.
I think we'll need to include this in the changeStream just for audit purposes so can't exclude it :(

trade-finance-manager-api/src/v1/controllers/user/user.controller.js
ChristianMcc: Can you be not logged in for this one? :( that strikes me as odd is it like that for tests or is it just api key only or smt?

trade-finance-manager-api/src/v1/controllers/user/user.routes.js
```diff
@@ -46,7 +46,8 @@ module.exports.createTfmUser = (req, res, next) => {
 
   const newUser = { ...userToCreate, salt, hash };
 
-  return create(newUser, (error, user) => {
+  // This is called on the open and auth router ('v1/user' and 'v1/users') endpoints so req.user may be undefined
```
ChristianMcc, line 49: Oh I see, fair enough I understand now

trade-finance-manager-api/src/v1/controllers/user/user.controller.js
ChristianMcc: I read the comment lower and I'm with you, let's leave like this for now because I don't see anyone actually hitting the not logged in endpoint in production etc.

ChristianMcc (commented):
Nice work :)

libs/common/src/helpers/changeStream/generateAuditDetails.ts
```diff
@@ -1,18 +1,20 @@
...
   lastUpdatedAt: Date;
-  lastUpdatedByPortalUserId?: string;
-  lastUpdatedByTfmUserId?: string;
+  lastUpdatedByPortalUserId?: ObjectId;
+  lastUpdatedByTfmUserId?: ObjectId;
   lastUpdatedByIsSystem?: boolean;
 };
 
-export const generatePortalUserAuditDetails = (userId: string): AuditDetails => ({
+export const generatePortalUserAuditDetails = (userId: string | ObjectId): AuditDetails => ({
```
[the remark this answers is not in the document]
MarRobSoftwire: I think it does, I've just had a play with `flatten` and 
```
flatten({
  foo: {
    string: 'string',
    bar: null,
  },
  foobar: null,
})
```
is converted to
```
{ '$set': { 'foo.string': 'string', 'foo.bar': null, foobar: null } }
```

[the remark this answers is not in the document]
MarRobSoftwire: I think making them explicitly null sounds like a good idea

trade-finance-manager-api/src/v1/controllers/user/user.controller.js
MarRobSoftwire: Yeah it's a weird thing - I'll add a comment explaining that the audit logs will be null in development only

libs/common/src/helpers/changeStream/generateAuditDetails.ts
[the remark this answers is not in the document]
ChristianMcc: excellent news!

trade-finance-manager-api/src/v1/controllers/acbs.controller.js
ChristianMcc: yeah it should, if it's only going to take 10-15 mins then go for it

ChristianMcc (dismissed):
Nice work, if you want to move the helper that's probably useful if it's quick and I'll reapprove after :)

ChristianMcc (approved):
lovely stuff

PR #2907, opened 2024-03-18
fix(dtfs2-7063): facility id record are missing in acbs

AlexBramhill:
## Introduction :pencil2:
Azure functions were failing to create deals. This was due to an issue with dates, and how we parse them.

When we create a deal in azure functions we pass `submissionDate`, which is an `epoch`, through `formatDate` returning an `iso date time as string`. Later in the deal creation process in azure functions, we call the same `formatDate` function on this newly created `iso date time as string`. This caused an `invalidDate` to be created, and lead to the deal failing to create.

We would expect this `iso date time as string` to return a `formatted date` when parsed through `formatDate`

We added handling to allow `formatDate` to parse `iso date time as sting`s

We've added test cases to ensure this is fixed, as well as running the data from the bug report locally to ensure it returns expected values.

### DIFF
--- a/azure-functions/acbs-function/helpers/date.js
+++ b/azure-functions/acbs-function/helpers/date.js
@@ -179,13 +179,13 @@ const getYearAndMmdd = (date) => {
 
   return isValid(parsedDate)
     ? {
-        mmdd: format(parsedDate, 'MM-dd'),
-        year: format(parsedDate, 'yyyy'),
-      }
+      mmdd: format(parsedDate, 'MM-dd'),
+      year: format(parsedDate, 'yyyy'),
+    }
     : {
-        mmdd: 'Invalid date',
-        year: 'Invalid date',
-      };
+      mmdd: 'Invalid date',
+      year: 'Invalid date',
+    };
 };
 
 module.exports = {
--- a/azure-functions/acbs-function/helpers/date.js
+++ b/azure-functions/acbs-function/helpers/date.js
@@ -176,13 +176,13 @@ const getYearAndMmdd = (date) => {
 
   return isValid(parsedDate)
     ? {
-        mmdd: format(parsedDate, 'MM-dd'),
-        year: format(parsedDate, 'yyyy'),
-      }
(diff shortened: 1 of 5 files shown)


azure-functions/acbs-function/helpers/date.js
```diff
@@ -150,13 +152,7 @@ const getDateStringFromYearMonthDay = (year, month, day) => {
...
   }
 
-  return formatOrReturnInvalidDate(
-    new Date(
-      Number(formatYear(year)),
-      Number(month) - 1,
-      Number(day),
-    ),
-  );
+  return formatOrReturnInvalidDate(new Date(Number(formatYear(year)), Number(month) - 1, Number(day)));
```
abhi-markan, line 155: Can the argument computation be moved to `variable` or a `function` to improve readbility?

abhi-markan (changes requested):

azure-functions/acbs-function/helpers/date.js
AlexBramhill: added -- sorry for rebase

azure-functions/acbs-function/helpers/date.js
```diff
@@ -144,19 +146,20 @@ const getDateStringFromYearMonthDay = (year, month, day) => {
...
 
-  const daysInMonth = getDaysInMonth(new Date(Number(formatYear(year)), Number(month) - 1));
+  const yearAsNumber = Number(formatYear(year));
+  const monthAsZeroIndexedNumber = Number(month) - 1;
+  const dayAsNumber = Number(day);
+  const startOfMonth = new Date(yearAsNumber, monthAsZeroIndexedNumber);
+
+  const daysInMonth = getDaysInMonth(startOfMonth);
 
   if (daysInMonth < Number(day)) {
```
ChristianMcc, line 156: This could technically be dayAsNumber

ChristianMcc (commented):
lil nit pick

ChristianMcc (approved):
a green tile for me

abhi-markan (approved):

PR #2908, opened 2024-03-18
feat(FN-1853): fix portal e2e test mock reports

JoshBinns2000:
## Introduction :pencil2:
Portal e2e tests are currently failing due to the reports that are being inserted having the wrong schema.

- Updates the generated mocks in `e2e-tests/portal` to use the `UtilisationReportEntityMockBuilder`
- Improves test coverage for some e2e tests

### DIFF
--- a/e2e-tests/portal/cypress/e2e/journeys/payment-report-officer/utilisation-report-upload/confirm-and-send.spec.js
+++ b/e2e-tests/portal/cypress/e2e/journeys/payment-report-officer/utilisation-report-upload/confirm-and-send.spec.js
@@ -23,6 +23,7 @@ context('Confirm and send', () => {
 
       confirmAndSend.mainHeading().should('not.exist');
       confirmAndSend.currentUrl().should('contain', '/utilisation-report-upload');
+      confirmAndSend.currentUrl().should('not.contain', '/confirm-and-send');
     });
 
     it('Should route to the Upload Report page when the change button is selected', () => {
@@ -30,6 +31,7 @@ context('Confirm and send', () => {
 
       confirmAndSend.mainHeading().should('not.exist');
       confirmAndSend.currentUrl().should('contain', '/utilisation-report-upload');
+      confirmAndSend.currentUrl().should('not.contain', '/confirm-and-send');
     });
 
     it('Should route to the Confirmation page when the Confirm and Send button is selected', () => {
@@ -46,6 +48,7 @@ context('Confirm and send', () => {
       cy.visit(relativeURL('/utilisation-report-upload/confirm-and-send'));
 
       confirmAndSend.currentUrl().should('contain', '/utilisation-report-upload');
+      confirmAndSend.currentUrl().should('not.contain', '/confirm-and-send');
     });
   });
 });
--- a/e2e-tests/support/tasks.js
+++ b/e2e-tests/support/tasks.js
@@ -1,7 +1,7 @@
 const crypto = require('node:crypto');
 const { MongoDbClient } = require('@ukef/dtfs2-common/mongo-db-client');
 const { SqlDbDataSource } = require('@ukef/dtfs2-common/sql-db-connection');
-const { UtilisationReportEntity, UtilisationDataEntity } = require('@ukef/dtfs2-common');
+const { UtilisationReportEntity, FeeRecordEntity } = require('@ukef/dtfs2-common');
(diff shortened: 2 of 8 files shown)


e2e-tests/portal/cypress/e2e/journeys/payment-report-officer/utilisation-report-upload/confirm-and-send.spec.js
```diff
@@ -23,13 +23,15 @@ context('Confirm and send', () => {
 
       confirmAndSend.mainHeading().should('not.exist');
       confirmAndSend.currentUrl().should('contain', '/utilisation-report-upload');
+      confirmAndSend.currentUrl().should('not.contain', '/confirm-and-send');
```
JoshBinns2000, line 26: I've added this because our tests which want to navigate to the `Upload Report page` will pass even if we navigate to `/utilisation-report-upload/confirm-and-send` as that url does contain `/utilisation-report-upload`

e2e-tests/support/tasks.js
```diff
@@ -102,6 +102,7 @@ module.exports = {
        * Deletes all the rows from the utilisation report and azure file info tables
        */
       async removeAllUtilisationReportsFromDb() {
+        await SqlDbDataSource.manager.getRepository(UtilisationDataEntity).delete({});
```
JoshBinns2000, line 105: This is required as we don't have cascade deletes for `UtilisationDataEntity`, so we need to delete those first

portal-api/src/v1/controllers/utilisation-report-service/utilisation-report-upload.controller.js
```diff
@@ -112,7 +113,6 @@ const uploadReportAndSendNotification = async (req, res) => {
 
     const existingReports = await api.getUtilisationReports(bankId, {
       reportPeriod: parsedReportPeriod,
-      excludeNotReceived: true,
```
JoshBinns2000, line 115: We should be checking for the existence of one report in the `REPORT_NOT_RECEIVED` state - `excludeNotReceived` will specifically remove reports in this state which we do not want

portal-api/src/v1/controllers/utilisation-report-service/utilisation-report-upload.controller.js
```diff
@@ -121,6 +121,14 @@ const uploadReportAndSendNotification = async (req, res) => {
 
     const existingReport = existingReports[0];
 
+    if (existingReport.status !== UTILISATION_REPORT_RECONCILIATION_STATUS.REPORT_NOT_RECEIVED) {
+      return res
+        .status(500)
```
JoshBinns2000, line 126: Should this status be a `403`?

RLavender98 (approved):
Looks good!

e2e-tests/portal/cypress/e2e/journeys/payment-report-officer/utilisation-report-upload/confirm-and-send.spec.js
JoshBinns2000: I imagine we would need to use a regex to do that, as our url can sometimes end with our `?_csrf=` query

portal-api/src/v1/controllers/utilisation-report-service/utilisation-report-upload.controller.js
JoshBinns2000: The user will probably see the "Problem with service" or "Something went wrong" page independent of which error is thrown (...probably). It's more of a semantic thing for `portal-api` where a `500` suggests there was an internal server error, whereas this is not an internal server error and more of a "data exists in the wrong format" error (which shouldn't be a `403` either in this case so I'm not sure why I suggested that 😅 ). Either way, it's not like we handle anything differently depending on the error code, so there isn't much point of changing this - it was more of a curiosity thing!

RLavender98: Yeah that makes sense, and I'm actually not sure semantically which it is! I think other than a user trying to access the endpoint directly when they shouldn't (defo a 4 something case!), the most likely (although probably not that likely) case for this is if a user loads the page for upload but then before uploading a different user uploads a report for the period, in which case I'm actually not sure what the correct response is. 500 doesn't seem too unreasonable, but perhaps 409 might technically be the correct choice?

PR #2910, opened 2024-03-18
feat(FN-2334): accessibility update labels on previous reports page

francescastocco:
## Introduction :pencil2:
Tester has observed when viewing the code behind the page, labels for user interface components do not match the visually displayed name on the page. It is expected that the displayed name of user interface components begins with or is the same as the component label or 'Accessible Name' within the HTML.

It has been observed that on the 'Previous GEF report' page the 'January 2024' CTA has a label that starts with "Download...". This does not match the visually displayed text and may cause issues for speech recognition software users.

- Update headers to say `Download`
- Update link text and aria-label to have the same value

- Designer suggested to also change the name of downloaded file to have the same value as the link text and label. This is still outstanding, I have left a comment for clarification in Jira.


RLavender98 (approved):
Nice! 😄

PR #2917, opened 2024-03-19
chore(FN): fix teams notification

ChristianMcc:
## Introduction :pencil2:
Correctly assign dev environment for notify job


.github/workflows/notify.yml
```diff
@@ -13,8 +13,7 @@ jobs:
   notify-teams:
     runs-on: ubuntu-latest
     environment:
-      environment: ${{ env.environment }}
-      timezone: ${{ env.timezone }}
+      name: dev
```
abhi-markan, line 16: environment:
      name: ${{ needs.setup.outputs.environment }}
      
Can we please follow similar pattern i.e. `sca.yml` 
image

abhi-markan (changes requested):

.github/workflows/notify.yml
ChristianMcc: I originally had something similar to this setup but realised it was a redundant step that doesn't give us anything. Happy to add it back in and will now just wanted to reduce number of steps if it wasn't needed!

ChristianMcc: Done :)

abhi-markan (approved):

abhi-markan (approved):

abhi-markan (approved):

PR #2918, opened 2024-03-19
feat(FN-1864): update activity logs columns

JoshBinns2000:
## Introduction :pencil2:
We want to update our `AuditableBaseEntity` to use the columns specified by the data team.

- Updates the columns of `AuditableBaseEntity` to:
  - `lastUpdatedAt: Date` - `datetime2`
  - `lastUpdatedByPortalUserId: string | null` - `nvarchar`
  - `lastUpdatedByTfmUserId: string | null` - `nvarchar`
  - `lastUpdatedByIsSystemUser: boolean` - `bit`
- Add a `updateLastUpdatedBy` method to `AuditableBaseEntity` which handles the logic around updating these fields
- Adds tests for the above method
- Adds a migration to update all the fields


libs/common/src/sql-db-entities/base-entities/auditable.base-entity.ts
```diff
@@ -1,13 +1,39 @@
...
+
+  @Column({ type: 'nvarchar', nullable: true })
+  lastUpdatedByTfmUserId!: string | null;
+
+  @Column()
+  lastUpdatedByIsSystemUser!: boolean;
 
-  @Column({ type: 'nvarchar' })
-  updatedByUserId!: DbAuditUpdatedByUserId;
+  public updateActivityDetails(requestSource: DbRequestSource): void {
```
RLavender98, line 17: It's not a big deal, but I'm not the biggest fan of 'activity details' for the wording. I think it would be nicer if the name linked in some ways to the class name or the db field names. Perhaps 'updateLastUpdatedBy'?

RLavender98 (approved):
Looks good, just a minor nitpick on wording 😄

libs/common/src/sql-db-entities/base-entities/auditable.base-entity.ts
```diff
@@ -1,13 +1,39 @@
...
+    const { platform } = requestSource;
+    switch (platform) {
```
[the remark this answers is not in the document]
JoshBinns2000: You need to destructure it if you want to avoid a typescript error in the `default` case.

Originally, I wanted to have an error message that says "I don't recognise this platform: ${platform}". You get a tsc error if you do that using `requestSource.platform`, because it thinks `requestSource` is of type `never` which means it has no properties on it. If you destructure at the start, and then use `platform` in the `default` case, you only get a lint error which we have turned off in other services. In the implementation I currently have you are right that there is no point to doing that, but I'm going to update the error message and then there will be :)

libs/common/src/sql-db-entities/base-entities/auditable.base-entity.ts
JoshBinns2000: I used that name hoping someone would come up with a better suggestion, and I think `updateLastUpdatedBy` is good so I'll go with that. Thanks!

libs/common/src/sql-db-entities/base-entities/auditable.base-entity.test.ts
```diff
@@ -0,0 +1,81 @@
...
+
+// We cannot create instances of abstract classes, so we define a test class below
+class TestAuditableBaseEntity extends AuditableBaseEntity {}
+
+describe('AuditableBaseEntity', () => {
+  describe('updateLastUpdatedBy', () => {
+    describe("when the request source platform is 'PORTAL'", () => {
+      const userId = 'abc123';
+
+      it.each`
```
ChristianMcc, line 12: I normally view it.each tests as being different inputs triggering different outputs rather than here where you are checking one update of an auditableBaseEntity having different values set correctly. Unless there's a benefit I'm missing can we change this to just be a single test asserting the whole object matches the object object we expect?
partly I also want the test to check that's all we're adding here and nothing else  which this method misses out on

libs/common/src/sql-db-entities/base-entities/auditable.base-entity.test.ts
```diff
@@ -0,0 +1,81 @@
...
+
+        // Assert
+        expect(testAuditableBaseEntity[field]).toBe(value);
+      });
+    });
+
+    describe("when the request source platform is 'TFM'", () => {
+      const userId = 'abc123';
+
+      it.each`
```
ChristianMcc, line 32: ditto

libs/common/src/sql-db-entities/base-entities/auditable.base-entity.test.ts
```diff
@@ -0,0 +1,81 @@
...
+        // Act
+        testAuditableBaseEntity.updateLastUpdatedBy({ platform: 'TFM', userId });
+
+        // Assert
+        expect(testAuditableBaseEntity[field]).toBe(value);
+      });
+    });
+
+    describe("when the request source platform is 'SYSTEM'", () => {
+      it.each`
```
ChristianMcc, line 50: ditto

ChristianMcc (commented):
LGTM :) nit pick on a few of the tests

ChristianMcc (approved):
LGTM

PR #2920, opened 2024-03-20
feat(FN-1859): fix upload report page

JoshBinns2000:
## Introduction :pencil2:
There is a bug currently with `portal` on the "Upload a report" screen when a report has already been uploaded. This bug is due to `portal` expecting to see an `uploadedByUser` object, but instead only saw the `uploadedByUserId` string.

- Updated `dtfs-central-api` to map the `uploadedByUserId` to the `UploadedByUser: UploadedByUserDetails` object that `portal` is expecting
- Updated associated tests
- Improved e2e test coverage to make sure this behaviour is caught in future pull requests

- Adds a `problemWithService` page to the `portal` e2e pages
- Updates the `/previous-reports` related controllers to typescript*
- Fixes e2e tests (except for `portal` `payment-report-officer` tests linked to filename validation)

*the type of `response.body` in `/bank/:bankId/utilisation-reports` is subtly different to what is returned by `/bank/:bankId/utilisation-reports/:id` but this is not documented anywhere, making it a consistent pain point. This is a good opportunity to fix this by switching files to typescript and adding a new `api-response-type` to `portal` and `portal-api`


utils/sql-db-seeder/src/utilisation-report/utilisation-report.helper.ts
```diff
@@ -78,8 +78,8 @@ export const createUploadedReport = (
   utilisationReport.reportPeriod = reportPeriod;
   utilisationReport.status = status;
   utilisationReport.dateUploaded = new Date();
-  utilisationReport.uploadedByUserId = null;
-  utilisationReport.updatedByUserId = getDbAuditUpdatedByUserId({ platform: 'TFM', userId: portalUser._id.toString() });
+  utilisationReport.uploadedByUserId = portalUser._id.toString();
+  utilisationReport.updatedByUserId = getDbAuditUpdatedByUserId({ platform: 'PORTAL', userId: portalUser._id.toString() });
```
JoshBinns2000, line 82: Noticed from demo that this seed helper generated incorrect data for the uploaded report, so decided to update it here

portal/server/controllers/utilisation-report-service/previous-reports/index.ts
```diff
@@ -0,0 +1,58 @@
...
+import { PRIMARY_NAV_KEY } from '../../../constants';
+import { asLoggedInUserSession } from '../../../helpers/express-session';
+
+type GetPreviousReportsRequest = Request & {
+  query: {
+    targetYear?: string;
+  };
+};
+
+export const getPreviousReports = async (req: GetPreviousReportsRequest, res: Response) => {
```
JoshBinns2000, line 13: Our original previous reports controller has always been a bit of a pain point for me as the return type of `api.getPreviousUtilisationReports...` in both `portal-api` and `dtfs-central-api` is different to what we would expect. This felt like a good opportunity to refactor this controller to typescript, make it a bit more readable and add types to document what gets returned

portal-api/src/v1/controllers/utilisation-report-service/previous-reports.controller.ts
```diff
@@ -0,0 +1,89 @@
...
+export const groupAndSortReports = (dbReports: UtilisationReportResponseBody[]): GroupedReport[] => {
+  const years = getYears(dbReports);
+  const groupedReports = groupReportsByStartYear(years, dbReports);
+  const groupedReportsByStartYear = populateOmittedYears(groupedReports, years);
+  return orderBy(groupedReportsByStartYear, ['year'], ['desc']);
+};
+
+type GetPreviousReportsByBankIdRequest = Request<{ bankId: string }>;
+
+export const getPreviousReportsByBankId = async (req: GetPreviousReportsByBankIdRequest, res: Response) => {
```
JoshBinns2000, line 71: The return type of this controller has always been a pain for me to figure out each time I have to look at it, so I took this as an opportunity to convert this file to typescript and document what gets returned

dtfs-central-api/src/types/utilisation-reports.ts
```diff
@@ -15,11 +16,19 @@ export type GetUtilisationReportResponse = {
...
+      uploadedByUser: UploadedByUserDetails;
+      azureFileInfo: AzureFileInfo;
+      dateUploaded: Date;
+    }
+  | {
+      uploadedByUser: null;
+      azureFileInfo: null;
+      dateUploaded: null;
+    }
+);
```
JoshBinns2000, line 31: This feels like a good use-case for a more narrow type as our mapping function used by the controllers can ensure that we are either in an "uploaded" or "not uploaded" state

dtfs-central-api/src/repositories/utilisation-reports-repo/utilisation-report-sql.repo.ts
```diff
@@ -44,6 +44,8 @@ export const UtilisationReportRepo = SqlDbDataSource.getRepository(UtilisationRe
 
     if (options?.excludeNotReceived) {
       findByOptionsWhere.status = Not('REPORT_NOT_RECEIVED');
+      const notReceivedReports = await this.findBy(findByOptionsWhere);
+      return notReceivedReports.filter((report) => report.azureFileInfo);
```
JoshBinns2000, line 48: Because our join column exists on the `AzureFileInfo` table, we can't query against the `UtilisationReport` table for it. Therefore, we must either use a query builder (which would be quite complex in this case due to the difference combinations of `options`) or just filter once we get all reports from the `UtilisationReport` table. I've implemented the latter here

dtfs-central-api/src/types/utilisation-reports.ts
RLavender98: Oh nice, I didn't consider this when I wrote it but perhaps should have done! That said, whilst I agree this typing portrays the returned object better and we should keep it now you have gone to the effort of doing it, I do just want to highlight how I see some of the trade offs, to ensure we weigh up the benefits against the dev effort when doing refactors of this kind in the future:

- With this change, I would say we now *have* to assert it falls into one of the two types when mapping to this, when we already know it will be one of the two because we wrote the code that gets reports into these states, except now we have more code to maintain. (and actually because here this is just being returned as the response immediately after being mapped to, and we already knew it would be one of the two shapes, we could have left it and still safely parse the response in say portal api with the stricter type without bothering to do so here (although I will accept that could look confusing))
- Narrowing types can sometimes be good for documentation as code. In this specific case, I think it was already fairly clear which fields would and would not be populated when the file is or isn't uploaded, but this does make it concrete for anyone less familiar with the domain!
- Anywhere objects are of this type, if we want to access azureFileInfo, will still either have to check it has a value or not, or assert which of the two possible types the main object falls into, which is not necessarily simpler. This can have a lot of value if we are filtering to one of the subtypes and then we don't have to be continually checking for field existence before using it, but I'm not sure we have any cases like that? (e.g. if we had left it we could still introduce a tighter type in say portal if we ever needed to to benefit from knowledge of their combined existence)

That all said, let's keep it! But, I do wonder if we can make the type slightly more readable by maybe extracting the types we are doing the operations on? One option which is perhaps a bit lengthy but prevent us having to define weird partial types that don't have individual domain meaning would be to have the underlying type that is as was before but not exportable, and then have a derived type where the fields are required and another where they are null, and then our endpoint actually just returns one | the other? What do you think?

dtfs-central-api/src/mapping/mapUtilisationReport.ts
```diff
@@ -20,12 +10,48 @@ const mapAzureFileInfoEntityToAzureFileInfo = (fileInfoEntity: AzureFileInfoEnti
...
+  uploadedByUserId: string;
+} {
+  const allUploadFieldsArePopulated: boolean =
+    uploadedReportEntity.dateUploaded !== null && uploadedReportEntity.azureFileInfo !== undefined && uploadedReportEntity.uploadedByUserId !== null;
+  if (!allUploadFieldsArePopulated) {
+    throw new Error('Failed to map data - report seems to have been uploaded but is missing some required fields');
+  }
+}
+
+export const mapUtilisationReportEntityToGetUtilisationReportResponse = async (
```
RLavender98, line 25: I was perhaps being a bit cheeky not adding unit tests for this specifically the first time around, but now that this contains a fair amount more logic I think we should introduce unit tests

utils/sql-db-seeder/src/utilisation-report/utilisation-report.helper.ts
RLavender98: Nice, thanks!

RLavender98 (changes requested):
Nice job, thanks for fixing! Mostly just a question around the database query change really 😄

dtfs-central-api/api-tests/sql-db-helper.ts
```diff
@@ -8,6 +8,8 @@ type SqlTableName = 'UtilisationReport' | 'FeeRecord' | 'AzureFileInfo';
 const deleteAllEntries = async (tableName: SqlTableName): Promise<void> => {
   switch (tableName) {
     case 'UtilisationReport':
+      // Cascade deletes are not enabled for the 'FeeRecord' table
+      await SqlDbDataSource.manager.delete(FeeRecordEntity, {});
```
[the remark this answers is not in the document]
JoshBinns2000: @ChristianMcc I can't fully remember what the reasoning was, but there was definitely a reason 😅

dtfs-central-api/src/repositories/utilisation-reports-repo/utilisation-report-sql.repo.ts
JoshBinns2000: Discussed with @ChristianMcc: the intended behaviour of this query was updated without my knowledge, and I didn't realise it was a change we wanted to prefer to query on status than the not uploaded flag. I'll revert this and the controllers which require a check for `azureFileInfo` being defined will be updated as well

dtfs-central-api/api-tests/sql-db-helper.ts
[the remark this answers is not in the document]
JoshBinns2000: This originally came from Christian worrying about deleting production data and not wanting a cascade delete, but considering that we should only ever be soft deleting utilisation reports in production having a cascade delete here should be fine. I just ran it by him on the huddle and it's fine to add. I'll implement this in an upcoming PR with a new migration

PR #2924, opened 2024-03-22
Fix/dtfs2 7062 regression portal bss ewcs clone eligibility supporting documentation page save go back button display http error 502

abhi-markan:
## Introduction :pencil2:
A `502` error is being thrown (due to `portal` micro-service crashing) when the property `securityDetails.exporter` is being access for a `BSS/EWCS` deal where EC have not been fully filled and `supportingInformation` property in the database is incomplete.

When saving and going back, the function tries to access a property which is saved in the database and thus causes the container to crash.

* Optional chaining
* Quick return for an invalid argument.

* Improved unit test coverage for the function `submittedDocumentationMatchesOriginalData`.


AlexBramhill (approved):
great documention in these tests! allowed me to quickly understand behaviour

lgtm

PR #2926, opened 2024-03-22
feat(FN-1862): remove mark not received as completed

JoshBinns2000:
## Introduction :pencil2:
We want to remove the ability of `PDC_RECONCILE` users to mark reports in the `REPORT_NOT_RECEIVED` state as completed. This change will now mean that any report not in the `REPORT_NOT_RECEIVED` state has been uploaded.

- Updates the state machine to no longer allow reports in the `REPORT_NOT_RECEIVED` state to be marked as completed
- Updates TFM-UI to not display the checkbox for reports in the `REPORT_NOT_RECEIVED` state
- Updates unit, api and e2e* tests to reflect these changes
- Update the `UtilisationReportEntityMockBuilder` to use a generic (see [this comment](https://github.com/UK-Export-Finance/dtfs2/pull/2926#discussion_r1537473223))
- Updates `tfm-ui` so that you can mark `'PENDING_RECONCILIATION'` and `'RECONCILIATION_COMPLETED'` reports as not completed and completed respectively (see [this comment](https://github.com/UK-Export-Finance/dtfs2/pull/2926#discussion_r1537463613))

*some of the TFM e2e tests rely on testing quarterly banks which has not been implemented yet. The mock bank with a quarterly schedule has `id` `'10'`, so these have been filtered for and a `TODO FN-1601` comment has been left in the places where this filtering is performed

### DIFF
--- a/e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/pdc-reconcile/mark-report-as-done.spec.js
+++ b/e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/pdc-reconcile/mark-report-as-done.spec.js
@@ -18,7 +18,8 @@ context(`${PDC_TEAMS.PDC_RECONCILE} users can mark reports as done and not done`
   const utilisationReportsAlias = 'utilisationReportsAlias';
 
   const displayStatusSelector = 'td > strong[data-cy="utilisation-report-reconciliation-status"]';
-  const tableCellCheckboxSelector = (reportId) => `th > div > div > input[data-cy="table-cell-checkbox--set-status--reportId-${reportId}"]`;
+  const tableCellCheckboxSelector = (reportId, status) =>
+    `th > div > div > input[data-cy="table-cell-checkbox--set-status--reportId-${reportId}-currentStatus-${status}"]`;
 
   const getDisplayStatus = (utilisationReportReconciliationStatus) => {
     switch (utilisationReportReconciliationStatus) {
@@ -88,13 +89,13 @@ context(`${PDC_TEAMS.PDC_RECONCILE} users can mark reports as done and not done`
      */
     const notReceivedUtilisationReports = [];
     cy.get(aliasSelector(utilisationReportsAlias)).each((utilisationReport) => {
-      const { bankId, id, reportPeriod } = utilisationReport;
+      const { bankId, id, reportPeriod, status } = utilisationReport;
 
       pages.utilisationReportsPage
         .tableRowSelector(bankId, submissionMonth)
         .should('exist')
         .within(($tableRow) => {
-          cy.wrap($tableRow).get(tableCellCheckboxSelector(id)).should('exist');
+          cy.wrap($tableRow).get(tableCellCheckboxSelector(id, status)).should('exist');
         });
 
       const notReceivedUtilisationReport = UtilisationReportEntityMockBuilder.forStatus('REPORT_NOT_RECEIVED')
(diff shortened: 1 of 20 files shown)


trade-finance-manager-ui/server/controllers/utilisation-reports/update-utilisation-report-status/index.ts
```diff
@@ -22,24 +25,26 @@ const FORM_BUTTON_VALUES = {
...
+const getReportIdsAndStatusesFromBody = (
+  body: undefined | UpdateUtilisationReportStatusRequestBody,
+): { id: string; status: UtilisationReportReconciliationStatus }[] => {
   if (!body || typeof body !== 'object') {
-    throw new Error('Expected request body to be an object');  
+    throw new Error('Expected request body to be an object');
   }
 
   return Object.keys(body)
     .filter((key) => key.match(CHECKBOX_PATTERN.WITHOUT_GROUPS))
```
ChristianMcc, line 39: what's the point of this without groups filter? is it just a pre filter or cna we just match against the whole checkbox pattern and if it fails to match the reportId and currentStatus we throw it away like we currently do?

e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/pdc-reconcile/mark-report-as-done.spec.js
```diff
@@ -1,33 +1,25 @@
...
-      bankId: undefined,
-      status: UTILISATION_REPORT_RECONCILIATION_STATUS.PENDING_RECONCILIATION,
-    },
-    {
-      bankId: undefined,
-      status: UTILISATION_REPORT_RECONCILIATION_STATUS.RECONCILIATION_COMPLETED,
-    },
-  ];
+  const tableCellCheckboxSelector = (reportId, status) =>
+    `th > div > div > input[data-cy="table-cell-checkbox--set-status--reportId-${reportId}-currentStatus-${status}"]`;
```
ChristianMcc, line 22: fwiw given the data-cy tag is so explicit I'm surprised to see that before we had a th >div > div selector rather than just on a data-cy check but doesn't really matter either way just giving some thoughts :D it makes the selector a little more brittle than just having a data-cy selector that should always be unique

ChristianMcc (commented):
LGTM, couple of minor comments. Just grab one of Ruby or francesca to sense check any replies to my comments and it should be good to merge after that :)

portal-api/api-tests/v1/utilisation-report-service/previous-reports.api-test.js
```diff
@@ -40,9 +40,10 @@ describe('GET /v1/banks/:bankId/utilisation-reports', () => {
           year,
         },
       })
-      .withDateUploaded(new Date(year, 0))
+      .withDateUploaded(new Date(`${year}-01`))
```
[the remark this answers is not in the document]
JoshBinns2000: I find it to be more readable as it removes the need to remember that javascript `Date` months are zero indexed

trade-finance-manager-ui/server/controllers/utilisation-reports/update-utilisation-report-status/index.ts
francescastocco: I think Christian was asking the 1st question - makes sense to me :)

francescastocco (approved):
LGTM 👍 I've sense-checked your replies to Christian

PR #2933, opened 2024-03-25
fix(DTFS2-7061): regression portal proceed next button is missing after issuing the ain facilities

abhi-markan:
## Introduction :pencil2:
Facility cover start date confirmation flag `coverStartConfirmed` was not being set for Acknowledged facility upon submission to the checker.

* Ensured `coverStartDate` flag is set to `true` upon facility stage update when pre-defined conditions are met.

* Documentation update
* Improved unit test coverage.
* Improved API test coverage.
* integrity attribute syntax fix.


portal-api/src/v1/controllers/deal-status/update-facility-cover-start-dates.js
```diff
@@ -1,42 +1,56 @@
...
-  const modifiedDeal = deal;
+  try {
+    const modifiedDeal = { ...deal };
 
-  if (modifiedDeal.facilities.length) {
-    modifiedDeal.facilities.forEach(async (facilityId) => {
-      const facility = await facilitiesController.findOne(facilityId);
+    if (!modifiedDeal.facilities || !modifiedDeal.facilities?.length) {
+      console.error('No facilities found in deal %s', modifiedDeal._id);
+      return modifiedDeal;
```
ChristianMcc, line 17: I don't know where this is called, but should this ever happen in the system or should we throw an error here that can be caught?

portal-api/src/v1/controllers/deal-status/update-facility-cover-start-dates.js
```diff
@@ -1,42 +1,56 @@
...
+    }
 
+    for (const facilityId of modifiedDeal.facilities) {
```
ChristianMcc, line 20: for (const x of y) is much better than an async foreach so definitely happen with this - can we use a .map and do this at the same time instead of do we need it to be sequential? Can imagine facilitiesController.update() might overwrite some deal details and not be able to be in parallel so if that's the case can ignore me

portal-api/src/v1/controllers/deal-status/update-facility-cover-start-dates.js
abhi-markan: Good shout, have replaced with `throw new Error(...)` and have also subseuqently updated the unit test cases.

portal-api/src/v1/controllers/deal-status/update-facility-cover-start-dates.js
abhi-markan: Moving to `for(...)` was done to allow sequential execution of the facilities update to avoid building up of any race condition, using `map` will not be advisable due to its ability to simuletaneously amend all facilities at the same time as you have rightly mentioned.

ChristianMcc: Okay great, thanks for confirming will resolve this comment!

portal-api/src/v1/helpers/date.test.ts
```diff
@@ -13,7 +13,7 @@ describe('getNowAsEpoch', () => {
     expect(typeof result).toBe('number');
   });
 
-  it('should return a value greater than or equal to the previous value', () => {
+  it('should return EPOCH from getNowAsEpoch function call, whose subsequent value will be either equal or greater than the previous call', () => {
```
ChristianMcc, line 16: Much more eloquently put than my suggestion, nice :D

ChristianMcc (approved):
LGTM :)

PR #2940, opened 2024-03-26
feat(FN-1864): add default value for lastUpdatedByIsSystemUser column

JoshBinns2000:
## Introduction :pencil2:
The previous #2918 pull request caused the deployment pipeline to fail with migrations due to the `lastUpdatedByIsSystemUser` not having a default value.

Removes the old migration and generates a new one which includes the default value.


ChristianMcc (approved):
Thanks for fixing!

ChristianMcc (approved):

PR #2941, opened 2024-03-26
feat(FN-2335): accessibility fix for utilisation report upload buttons

JoshBinns2000:
## Introduction :pencil2:
The utilisation report upload button labels did not include information about the expected file type, meaning users who use a screen reader would not be provided that information when focusing on the button.

Moves the `div.govuk-hint` element inside of the `label`.

## Screenshots 📷 

Below are screenshots from using the google chrome accessibility dev tools to show that the label for the button now include the file type information.

"Utilisation report upload"

FN-2335-upload-report-page

"Check the report"

FN-2335-check-the-report-1

FN-2335-check-the-report-2

### DIFF
--- a/portal/component-tests/utilisation-report-service/utilisation-report-upload.component-test.js
+++ b/portal/component-tests/utilisation-report-service/utilisation-report-upload.component-test.js
@@ -81,7 +81,8 @@ describe(page, () => {
     });
 
     it('should state which report the page is expecting to be uploaded', () => {
-      wrapper.expectText('[data-cy="upload-report-text"]').toRead('Upload December 2022 report');
+      wrapper.expectText('[data-cy="upload-report-text"]').toMatch(/[Upload December 2022 report]/);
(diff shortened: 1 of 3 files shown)


portal/component-tests/utilisation-report-service/utilisation-report-upload.component-test.js
```diff
@@ -113,7 +114,8 @@ describe(page, () => {
     });
 
     it('should state which report the page is expecting to be uploaded', () => {
-      wrapper.expectText('[data-cy="upload-report-text"]').toRead('Upload January 2023 report');
+      wrapper.expectText('[data-cy="upload-report-text"]').toMatch(/(Upload January 2023 report)/);
```
AlexBramhill, line 117: Out of curiosity do we need the `(` and `)`? I've not checked out this branch to check how `expect` works in jest, but regex wise, it seems we could lose it.

AlexBramhill (approved):

PR #2943, opened 2024-03-27
feat(FN-1853): squash migrations

JoshBinns2000:
## Introduction :pencil2:
We want to squash our migrations.

Squashes all current migrations into a single migration.


.github/workflows/feature_deploy.yml
```diff
@@ -15,6 +15,7 @@ on:
       - "dtfs-central-api/**"
       - "external-api/**"
       - "azure-functions/acbs-function/**"
+      - "libs/common/**"
```
JoshBinns2000, line 18: This change will mean that merging this branch to trigger a feature environment deployment. Without it, as this PR only changes files in `libs/common`, merging this PR will not trigger a deployment

PR #2949, opened 2024-03-28
feat(FN-1268): send email prompts for quarterly periods

RLavender98:
## Introduction :pencil2:
Send emails to prompt banks with quarterly reporting schedule to upload reports.

- Email is sent to bank with quarterly reporting schedule when submission window opens for quarterly reporting period. (1st working day of month)
- Email is sent to bank with quarterly reporting schedule when submission window is closing for quarterly reporting period. (10th business day of month)
- Email is sent to bank with quarterly reporting schedule when report is overdue for quarterly reporting period. (20th business day of month)
- Banks with quarterly reporting schedule are not sent any reminder emails for months which are not submission months for their quarters.
- Quarterly emails display report period in emails as '<StartMonth> to <EndMonth> <Year>' when quarter in within one year
- Quarterly emails display report period in emails as '<StartMonth> <StartYear> to <EndMonth> <EndYear>' when quarter spans two years

- Refactor logic around whether report is due to use existence of report in 'REPORT_NOT_RECEIVED' state
- Filter out banks not visible in TFM Utilisation Reports (rather than rely on logic that checks for a payment officer team email and logs if not found)


portal-api/src/cron-scheduler-jobs/helpers/utilisation-report-helpers.test.ts
```diff
@@ -0,0 +1,637 @@
...
+        },
+        end: {
+          month: 1,
+          year: 2098,
+        },
+      };
+      jest.mocked(api.getUtilisationReports).mockResolvedValue([]);
+
+      // Act + Assert
+      await expect(async () => await getIsReportDue(BANK_ID, reportPeriod)).rejects.toThrow(
```
JoshBinns2000, line 196: After some testing, I think we don't need to mark the function inside `expect` with `async-await` because we have used the `rejects` option. I don't think there is necessarily any value in removing the `async-await` part though - just something I found interesting!

portal-api/src/cron-scheduler-jobs/helpers/utilisation-report-helpers.test.ts
```diff
@@ -0,0 +1,637 @@
...
+    });
+
+    it('sends emails using the default team name when bank does not have one set', async () => {
+      // Arrange
+      jest.mocked(externalApi.bankHolidays).getBankHolidayDatesForRegion.mockResolvedValue([]);
+
+      const validEmail = 'valid-email@example.com';
+      const bank = aBank();
+      bank.isVisibleInTfmUtilisationReports = true;
+      bank.paymentOfficerTeam = { emails: [validEmail] } as PaymentOfficerTeam;
```
JoshBinns2000, line 413: Is there any reason why we can't do `bank.paymentOfficerTeam.emails = [validEmail]` to avoid the type assertion?

portal-api/src/cron-scheduler-jobs/helpers/utilisation-report-helpers.test.ts
```diff
@@ -0,0 +1,637 @@
...
+      await sendEmailToAllBanksWhereReportNotReceived({
+        emailDescription: '',
+        sendEmailCallback,
+      });
+
+      // Assert
+      expect(sendEmailCallback).toHaveBeenCalledTimes(3);
+      expect(sendEmailCallback).toHaveBeenCalledWith(validBarclaysEmail, validBarclaysTeamName, 'October 2023');
+      expect(sendEmailCallback).toHaveBeenCalledWith(validHsbcEmail, validHsbcTeamName, 'October 2023');
+      expect(sendEmailCallback).toHaveBeenCalledWith(otherValidHsbcEmail, validHsbcTeamName, 'October 2023');
```
JoshBinns2000, line 510: It might be nice to link these values to some array which we can extract from the bank `paymentOfficerTeam`s defined above, but I also think that might be a lot of work for little gain so I don't feel strongly about it

portal-api/src/cron-scheduler-jobs/helpers/utilisation-report-helpers.test.ts
```diff
@@ -0,0 +1,637 @@
...
+      await sendEmailToAllBanksWhereReportNotReceived({
+        emailDescription: '',
+        sendEmailCallback,
+      });
+
+      // Assert
+      expect(sendEmailCallback).not.toHaveBeenCalled();
+      expect(console.error).toHaveBeenCalledWith(expect.stringContaining('Failed to send'), expect.any(Error));
+    });
+  });
```
JoshBinns2000, line 636: Very nice tests!

portal-api/src/cron-scheduler-jobs/helpers/utilisation-report-helpers.ts
```diff
@@ -0,0 +1,203 @@
...
+ *   The callback is provided the following argument `{ emailAddress: string; recipient: string }`,
+ *   extracted from the provided bank.
+ * @returns {Promise<void>}
+ */
+const sendEmailToBankIfReportNotReceived = async (bank: BankResponse, emailDescription: string, sendEmailCallback: SendEmailCallback): Promise<void> => {
+  const { name: bankName, id: bankId, utilisationReportPeriodSchedule: schedule } = bank;
+  const currentReportingPeriodForBank = getCurrentReportPeriodForBankSchedule(schedule);
+  const formattedReportPeriod = getFormattedReportPeriod(currentReportingPeriodForBank);
+
+  if (isPreviousCalendarMonth(currentReportingPeriodForBank.end.month)) {
```
JoshBinns2000, line 160: Could we swap around the `if-else` here to return early if `isPreviousCalendarMonth` returns `false`?

portal-api/src/cron-scheduler-jobs/helpers/utilisation-report-helpers.ts
```diff
@@ -0,0 +1,203 @@
...
+    console.info(
+      `Not sending '${emailDescription}' email to '${bankName}' (bank ID: ${bankId}) - report is not due this month, current reporting period for bank is ${formattedReportPeriod}`,
+    );
+  }
+};
+
+/**
+ * For all banks, checks if the current utilisation report period for bank is in it's submission period and
+ * if so if utilisation report has been received for the current reporting period
+ * then, where not yet received, attempts to call the provided callback function to send the required email
```
JoshBinns2000, line 180: I think this description is a little tricky to read. Is there any way we could re-phrase it to make it a bit clearer?

portal-api/src/v1/api.js
```diff
@@ -366,6 +366,10 @@ const getBankById = async (bankId) => {
   }
 };
 
+/**
+ * Call the central API to get all banks
+ * @returns {Promise<import('../../src/v1/api-response-types').BankResponse[]>} response of API call or wrapped error response
```
JoshBinns2000, line 371: I think this path should be `'./api-response-types'`

portal-api/src/cron-scheduler-jobs/helpers/utilisation-report-helpers.ts
```diff
@@ -0,0 +1,203 @@
+import { format, subMonths } from 'date-fns';
+import { MonthAndYear, OneIndexedMonth, PaymentOfficerTeam, ReportPeriod, getCurrentReportPeriodForBankSchedule, getOneIndexedMonth } from '@ukef/dtfs2-common';
+import externalApi from '../../external-api/api';
+import api from '../../v1/api';
+import { getBusinessDayOfMonth } from '../../utils/date';
+import { isValidEmail } from '../../utils/string';
+import { BANK_HOLIDAY_REGION } from '../../constants/bank-holiday-region';
+import { BankResponse } from '../../v1/api-response-types';
+
+export type SendEmailCallback = (emailAddress: string, recipient: string, formattedReportPeriod: string) => Promise<void>;
```
JoshBinns2000, line 10: I noticed that the old parameter type for this callback used to be `({ emailAddress, recipient, formattedReportPeriod }: { emailAddress: string; recipient: string; formattedReportPeriod: string })` - just wondering what the reason for this change was :)

JoshBinns2000 (commented):
Left a few small comments, but this is looking good especially with respect to tests (they were a pleasure to read through in this PR 😄)

portal-api/src/cron-scheduler-jobs/helpers/utilisation-report-helpers.ts
RLavender98: My assumption was it was written that way due to it being javascript to make it clearer what the parameters were, since we did have a comment detailing the type but it wasn't attached to the usages very clearly so hard to see when doing dev. Since now it is in typescript it was easier to determine parameters and their types and I think it is more readable to not bundle the parameters into an object only to immediately deconstruct them.

portal-api/src/cron-scheduler-jobs/helpers/utilisation-report-helpers.ts
```diff
@@ -0,0 +1,203 @@
...
+    );
+  }
+
+  return reportsInReportPeriod[0].status === 'REPORT_NOT_RECEIVED';
+};
+
+/**
+ * Get the email recipient from the bank specific paymentOfficerTeam or fall back to a generic default
+ */
+export const getEmailRecipient = (paymentOfficerTeam: PaymentOfficerTeam, bankName: string): string => {
```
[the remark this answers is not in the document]
RLavender98: Everywhere this is used, typescript believes this parameter we pass in is of type `PaymentOfficerTeam`. My presumption is we expect every bank to have this unless there has been human error during manually adding a bank to the database, so I'm not really sure how we want to type that. I could just change the signature of this function but that feels weird given it is not typed optionally on the Bank type. Unless actually the Bank type has been typed wrong and this should be optional?

portal-api/src/cron-scheduler-jobs/helpers/utilisation-report-helpers.ts
RLavender98: Good idea! Originally I couldn't because this was part of the function below but now it is extracted that is a good shout

portal-api/src/cron-scheduler-jobs/helpers/utilisation-report-helpers.test.ts
RLavender98: Ahh nice thanks, I got very confused trying to get this to work in a way that be assert an error was thrown 😅

portal-api/src/cron-scheduler-jobs/helpers/utilisation-report-helpers.test.ts
RLavender98: This line is also ensuring there is no team name provided, which is why we need the type assertion. It's a bit tricky to try and type that explicitly in a way I can trick typescript into thinking the bank is a Bank, so I will add a comment noting that is intentional.

portal-api/src/cron-scheduler-jobs/helpers/utilisation-report-helpers.ts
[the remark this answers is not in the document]
RLavender98: Hm you're right this looks weird but I'm not sure what the best solution is. 

Everywhere this is used, typescript believes this parameter we pass in is of type PaymentOfficerTeam. My presumption is we expect every bank to have this unless there has been human error during manually adding a bank to the database, so I'm not really sure how we want to type that. I could just change the signature of this function but that feels weird given it is not typed optionally on the Bank type., but perhaps that is ok.

Unless actually the Bank type has been typed wrong and this should just be optional?

portal-api/src/cron-scheduler-jobs/helpers/utilisation-report-helpers.test.ts
JoshBinns2000: Ah I see. In that case, I generally prefer to use a `// @ts-expect-error <explanantion>` comment to highlight why typescript doesn't like it as you both get to avoid the `as` type assertion and get a comment explaining why it seems wrong

portal-api/src/cron-scheduler-jobs/helpers/utilisation-report-helpers.ts
RLavender98: I think the state machine is quite different, we have a lot of callbacks that we want to nominally have the same shape, either they do or don't take a report, and then they have a bunch of other stuff that differs per event which we wrap up in a payload, so here it makes sense to have an object so that we can support that nicely. I think if we had a few more parameters here then we might want to group them up in a semantic way into one or more objects to keep the code nice and readable, like we do with the state machine

portal-api/src/cron-scheduler-jobs/helpers/utilisation-report-helpers.ts
```diff
@@ -0,0 +1,200 @@
...
+ */
+const formatMonth = (month: OneIndexedMonth): string => format(new Date(`2023-${month}`), 'MMMM');
+
+/**
+ * Returns the given report period as a string.
+ * For a single month this is in 'MMMM yyyy' format.
+ * For a period of more than one month where the start and end month and in the same year this is in the 'MMMM to MMMM yyyy' format.
+ * For a period of more than one month where the start and end months are in different years this is in the 'MMMM yyyy to MMMM yyyy' format.
+ */
+export const getFormattedReportPeriod = (reportPeriod: ReportPeriod): string => {
```
JoshBinns2000, line 67: This is commonised in #2985 so we need to remember to update this instance of it too (this will get merged first)

portal-api/src/cron-scheduler-jobs/helpers/utilisation-report-helpers.ts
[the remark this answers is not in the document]
JoshBinns2000: I'm not sure. I'm surprised eslint isn't upset by this, but it's not worth digging into I think - happy to leave it as it is

JoshBinns2000 (approved):
LGTM (with the one comment about the report period formatting being commonised in #2985)

PR #2979, opened 2024-04-02
feat(FN-2561): update error text

francescastocco:
## Introduction :pencil2:
Error copy was incorrectly referencing base and accrual currency

- Update error text & tests


JoshBinns2000 (approved):
LGTM

PR #2980, opened 2024-04-02
feat(DTFS2-NONE): update logo to tudor crown

AlexBramhill:
## Introduction :pencil2:
We need to update the crown to be a tudor crown

Kept the existing pattern for the header pages, and updated the svg following html guides here:
https://github.com/alphagov/govuk-frontend/releases/tag/v4.8.0

The current setup is most likely due to the requirement to have many more additional class tags on the header element, so I've kept the file the same for now.


AlexBramhill:
closed as ticket now exists

PR #2982, opened 2024-04-02
feat(dtfs2 7093): Update to Tudor crown

AlexBramhill:
## Introduction :pencil2:
We need to update the crown to be a tudor crown

Kept the existing pattern for the header pages, and updated the svg following html guides here:
https://github.com/alphagov/govuk-frontend/releases/tag/v4.8.0

The current setup is most likely due to the requirement to have many more additional class tags on the header element, so I've kept the file the same for now.

I've not been able to test favicon is updated -- we should make sure this is the case in (and remember that favicons cache oddly)


abhi-markan (approved):

AlexBramhill:
closed to merge into r1.5 (needed to rebuild public folder)

PR #2985, opened 2024-04-03
feat(FN-2353): move get formatted report period helper to common

JoshBinns2000:
## Introduction :pencil2:
We want to move the `getFormattedReportPeriod` helper to `libs/common` so that it can be reused in `TFM` in upcoming work.

- Moves tests and other relevant date helpers to `libs/common`
- Updates extraneous imports


francescastocco (approved):
LGTM

PR #2988, opened 2024-04-03
fix(DTFS2-6119): fix duplicate emails being sent

AlexBramhill:
## Introduction :pencil2:
This is half a front end fix for ensuring that we do not resubmit forms.

This PR tackles buttons that currently use `govukButton`.

Added the built in method `preventDoubleClick: true` to all `govukButton`.

There is additional work to extract other buttons into `govukButton`s but this will be a different PR


abhi-markan (changes requested):
Can `preventDoubleClick` be applied at a global level so all the future buttons are automatically adhere to such rules. I believe EXIP repo has implemented at a global level.

AlexBramhill:
This is not the correct fix -- more to follow.

PR #2991, opened 2024-04-04
feat(FN-1613): bank report details page and TFM-6 feature flag

JoshBinns2000:
## Tickets 🎫 

[FN-1613](https://ukef-dtfs.atlassian.net/browse/FN-1613)
[FN-1603](https://ukef-dtfs.atlassian.net/browse/FN-1603)

## Introduction :pencil2:
This PR involves two key parts - the introduction of the utilisation report reconciliation page for a specific bank and a feature flag for TFM payment reconciliation related work.

For reports on the `<TFM-URL>/utilisation-reports` page that are not in the `REPORT_NOT_RECEIVED` state, the bank name should be a link which directs the user to the `<TFM-URL>/utilisation-reports/:reportId` page. This page contains the bank name, the report period and four tab items which are currently unpopulated. It should be noted here that this link will only appear if the below feature flag is enabled. Otherwise, the bank name will always be plain text.

The new feature flag proposed in this PR is called `FF_TFM_PAYMENT_RECONCILIATION_ENABLED` and should have the value `'true'` or `'false'` in the root `process.env`. This can be seen in the `zod` schema used to parse the feature flags in `libs/common`. Rather than using the feature flag directly in the microservices, a helper is exported from `libs/common` with the name `isTfmPaymentReconciliationEnabled` which simply returns whether or not the flag is enabled, abstracting away the environment variable from the actual implementation. A middleware is created in `trade-finance-manager-ui` which uses this helper to redirect to `not-found` depending on whether or not the feature flag is enabled.

## Resolution ✔️ 
- Adds a new feature flag system in `libs/common`
- Adds a new `/utilisation-reports/reconciliation-details/:reportId` endpoint
- Adds a new `/utilisation-reports/:reportId` route to `tfm-ui`
- Updates api, unit and component tests

Below are some screenshots of the code when it executes successfully (slightly outdated).

FN-1613 (cropped)

FN-1603 (cropped)

- Converts some files to typescript


dtfs-central-api/src/v1/routes/utilisation-reports-routes.js
```diff
@@ -141,4 +141,8 @@ utilisationReportsRouter
  */
 utilisationReportsRouter.route('/set-status').put(putUtilisationReportStatusController.putUtilisationReportStatus);
 
+utilisationReportsRouter
```
JoshBinns2000, line 144: Need to add swagger docs

trade-finance-manager-api/src/v1/api.js
```diff
@@ -1209,6 +1208,19 @@ const updateUtilisationReportStatus = async (reportsWithStatus, user) => {
   return response.data;
 };
 
+/**
+ * Gets the utilisation report reconciliation data by report id
+ * @param {string} reportId - The report id
+ * @returns {Promise<import('./api-response-types').UtilisationReportReconciliationBody>}
```
JoshBinns2000, line 1214: This name needs to change

trade-finance-manager-api/src/v1/controllers/utilisation-reports/get-utilisation-report-reconciliation-by-id.controller/index.ts
```diff
@@ -0,0 +1,29 @@
...
+  try {
+    const utilisationReportReconciliationData = await api.getUtilisationReportReconciliationById(reportId);
+
+    return res.status(HttpStatusCode.Ok).send(utilisationReportReconciliationData);
+  } catch (error) {
+    console.error('Failed to get report by id');
+    if (error instanceof AxiosError) {
+      return res.status(error.response?.status ?? HttpStatusCode.InternalServerError).send(`Failed to get utilisation report with id ${reportId}: ${error.message}`);
+    }
+    return res.status(HttpStatusCode.InternalServerError).send(`Failed to get utilisation report with id ${reportId}`);
```
JoshBinns2000, line 27: Need to update error messages

trade-finance-manager-api/src/v1/routes.js
```diff
@@ -131,10 +131,18 @@ authRouter
...
-  .get(validation.sqlIdValidation, handleExpressValidatorResult, utilisationReportsController.getUtilisationReportDownload);
+  .get(validation.sqlIdValidation('id'), handleExpressValidatorResult, utilisationReportsController.getUtilisationReportDownload);
 
 authRouter
   .route('/utilisation-reports/set-status')
   .put(validation.updateReportStatusPayloadValidation, handleExpressValidatorResult, utilisationReportsController.updateUtilisationReportStatus);
 
+authRouter
+  .route('/utilisation-reports/:id')
+  .get(validation.sqlIdValidation('id'), handleExpressValidatorResult, utilisationReportsController.getUtilisationReportById);
```
JoshBinns2000, line 142: Need to remove this

trade-finance-manager-ui/component-tests/utilisation-reports/utilisation-report-for-bank.component-test.js
```diff
@@ -0,0 +1,43 @@
+const { PRIMARY_NAVIGATION_KEYS } = require('../../server/constants');
+const pageRenderer = require('../pageRenderer');
+
+const page = '../templates/utilisation-reports/utilisation-report-for-bank.njk';
```
JoshBinns2000, line 4: Change name of template to include reconciliation

trade-finance-manager-ui/server/middleware/validateSqlId/index.ts
```diff
@@ -2,8 +2,8 @@ import { NextFunction, Request, Response } from 'express';
 import { isValidSqlId } from '../../helpers/validateIds';
 
 /** Middleware to validate that the 'id' field found as a path param is a valid SQL id */
-export const validateSqlId = (req: Request, res: Response, next: NextFunction) => {
-  const { id } = req.params;
+export const validateSqlId = (key: string) => (req: Request, res: Response, next: NextFunction) => {
```
JoshBinns2000, line 5: Should say `paramName` not `key`

trade-finance-manager-ui/server/middleware/validateUserTeam/index.test.ts
```diff
@@ -0,0 +1,68 @@
+import httpMocks from 'node-mocks-http';
+import { TeamId, TEAM_IDS } from '@ukef/dtfs2-common';
+import { validateUserTeam } from '.';
```
JoshBinns2000, line 3: I thought this was done in a refactor...

trade-finance-manager-ui/templates/utilisation-reports/_macros/report-reconciliation-table.njk
```diff
@@ -7,6 +7,7 @@
   {% set summaryItems = params.summaryItems %}
   {% set submissionMonth = params.submissionMonth %}
   {% set csrfToken = params.csrfToken %}
+  {% set FF_TFM_6 = params.FF_TFM_6 %}
```
JoshBinns2000, line 10: Need to fix this is say `isTfm6FeatureFlagEnabled`

trade-finance-manager-ui/server/middleware/validateUserTeam/index.test.ts
JoshBinns2000: It was not

trade-finance-manager-ui/.eslintrc.js
```diff
@@ -59,7 +59,10 @@ module.exports = {
         project: './tsconfig.eslint.json',
         tsconfigRootDir: __dirname,
       },
-      rules: baseRules,
+      rules: {
+        ...baseRules,
+        '@typescript-eslint/no-misused-promises': 'off',
```
RLavender98, line 64: Same again

trade-finance-manager-ui/component-tests/utilisation-reports/_macros/report-reconciliation-table.component-test.js
```diff
@@ -57,6 +58,8 @@ describe(component, () => {
       wrapper.expectElement(`${rowSelector} th`).toHaveCount(1);
       wrapper.expectElement(`${rowSelector} th:contains("${summaryItem.bank.name}")`).toExist();
 
+      wrapper.expectText(`${rowSelector} th > p`).toRead(summaryItem.bank.name);
```
RLavender98, line 61: Have you checked if this fails if the bank name is a link, e.g. if you set the feature flag to true at the top of this test and run it does it fail?

trade-finance-manager-ui/component-tests/utilisation-reports/utilisation-report-reconciliation-for-bank.component-test.js
```diff
@@ -0,0 +1,43 @@
...
+  });
+
+  it.each`
+    tabName               | dataCy
+    ${'Premium payments'} | ${'bank-report-tab-premium-payments'}
+    ${'Keying sheet'}     | ${'bank-report-tab-keying-sheet'}
+    ${'Payment details'}  | ${'bank-report-tab-payment-details'}
+    ${'Utilisation'}      | ${'bank-report-tab-utilisation'}
+  `("should render the '$tabName' tab", ({ dataCy }) => {
+    wrapper.expectElement(`[data-cy="${dataCy}"]`).toExist();
```
RLavender98, line 41: I can see we use lots of `data-cy` on the project for our component tests. There is an issue with this approach: we are just testing that we have put certain test-helper attributes in place and not actually testing anything of real substance.
In this case, we obviously don't want to be testing the gov uk component so I think find to still use the same selector, but maybe we could check the text content of the element returned contains the tab name so that we are testing we have set the tab names correctly?

trade-finance-manager-ui/templates/utilisation-reports/_macros/report-reconciliation-table.njk
```diff
@@ -24,20 +25,30 @@
...
-          <tr
-            class="govuk-table__row"
-            data-cy="utilisation-report-reconciliation-table-row-bank-{{ summaryItem.bank.id }}-submission-month-{{ submissionMonth
-              }}">
-            <th scope="row" class="govuk-table__header govuk-!-font-weight-regular" data-sort-value="{{ summaryItem.bank.name }}">
-            {# TODO FN-1603 - link to individual bank report page (replace `div` with `a`) #}
-            {# <a class="govuk-link govuk-link--no-visited-state" href="#">#}
+      {% for summaryItem in summaryItems %}
+        {% set checkboxId = ['set-status--reportId-', summaryItem.reportId, '-currentStatus-', summaryItem.status] | join %}
+        {% set checkboxLabel %}
```
RLavender98, line 30: Possibly not directly related to your changes but I'm not sure it makes sense to call this the checkbox label given it is quite a bit more than that. I also actually can't tell but is this used a label for the checkbox but have we checked keyboard nav and screenreader usage both still make sense for following the link and also checking the check box now this can be a link?

trade-finance-manager-ui/templates/utilisation-reports/utilisation-report-reconciliation-for-bank.njk
```diff
@@ -0,0 +1,79 @@
...
+{% set keyingSheetHtml %}
+  <div></div> {# TODO: FN-1722 #}
+{% endset %}
+
+{% set paymentDetailsHtml %}
+  <div></div> {# TODO: No tickets created yet #}
+{% endset %}
+
+{% set utilisationHtml %}
+  <div></div> {# TODO: No tickets created yet #}
```
RLavender98, line 18: Similar to my other comment with regards to TODOs, but even more so let's please not have any TODOs without ticket numbers.
I can see the ACs do say we should all four tabs but also I'm questioning that slightly since Christian mentioned we may not end up doing the utilisation tab. It would make more sense for tickets populating tabs to add their own tabs past the first one. @ChristianMcc should/could we descope any of these tabs from this ticket?

trade-finance-manager-ui/server/controllers/utilisation-reports/utilisation-report-reconciliation-for-bank/index.ts
```diff
@@ -0,0 +1,28 @@
...
+
+    const formattedReportPeriod = getFormattedReportPeriodWithLongMonth(utilisationReportReconciliation.reportPeriod);
+
+    // TODO FN-1615 update when feeRecord information is displayed on the page
+    return res.render('utilisation-reports/utilisation-report-reconciliation-for-bank.njk', {
+      user,
+      activePrimaryNavigation: PRIMARY_NAVIGATION_KEYS.UTILISATION_REPORTS,
+      bank: utilisationReportReconciliation.bank,
+      formattedReportPeriod,
+      // feeRecords: utilisationReportReconciliation.feeRecords,
```
RLavender98, line 22: Again, please can we just not have the TODO and the commented out feeRecords property

trade-finance-manager-ui/server/api-response-types/utilisation-report-response-body.ts
```diff
@@ -0,0 +1,19 @@
+import { AzureFileInfo, IsoDateTimeStamp, ReportPeriod, UploadedByUserDetails, UtilisationReportReconciliationStatus } from '@ukef/dtfs2-common';
+
+export type UtilisationReportResponseBody = {
```
RLavender98, line 3: Is this used anywhere?

RLavender98 (changes requested):
Nice job! Mostly fairly minor comments 😄

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-report-reconciliation-by-id.controller/index.ts
```diff
@@ -0,0 +1,85 @@
...
+      name: bankName,
+    },
+    status,
+    reportPeriod,
+    dateUploaded,
+    feeRecords: mappedFeeRecords,
+  };
+};
+
+export const getUtilisationReportReconciliationById = async (req: GetUtilisationReportReconciliationByIdRequest, res: Response<ResponseBody>) => {
```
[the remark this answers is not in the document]
JoshBinns2000: I was struggling with naming this, but I prefer `ReconciliationDetails` so I'll go with that

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-report-reconciliation-by-id.controller/index.ts
```diff
@@ -0,0 +1,85 @@
...
+}>;
+
+type ResponseBody = UtilisationReportReconciliationItem | string;
+
+/**
+ * TODO FN-1615 this function should extract the relevant details from the fee record entity that are needed for the premium payments screen
+ * @param feeRecord - The fee record entity
+ * @returns The mapped fee record
+ */
+const mapFeeRecordEntityToReconciliationItemFeeRecord = (feeRecord: FeeRecordEntity): UtilisationReportReconciliationItem['feeRecords'][number] => ({
```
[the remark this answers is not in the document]
JoshBinns2000: My assumption was that the mapping folder was intended for shared mapping functions rather than any and all mapping functions and, seeing as these functions are specific only to this endpoint (at least, for now), I wanted to keep them near the controller. I based this approach on the reconciliation summary stuff, where a `helper.ts` and `helper.test.ts` contain the mapping functions, but maybe those also belong in the mapping folder. My guess is that the mapping folder is preferable because you can get more granular testing, which a `helper.ts` file would also achieve, but are there any other reasons you have for preferring it over that approach?

dtfs-central-api/src/types/utilisation-reports.ts
```diff
@@ -61,3 +61,18 @@ export type UtilisationReportUploadDetails = Prettify<
...
+  bank: {
+    id: string;
+    name: string;
+  };
+  status: UtilisationReportReconciliationStatus;
+  reportPeriod: ReportPeriod;
+  dateUploaded: Date;
+  feeRecords: {
+    facilityId: string;
+    // TODO FN-1615 add extra required details of fee records
```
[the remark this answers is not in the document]
JoshBinns2000: I left these in as a reminder for myself for the work I'm doing on 1615 to make it easier as I went along, but considering that I've basically finished that work I can remove all these TODOs. I agree in general about not having these comments in the code :)

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-report-reconciliation-by-id.controller/index.test.ts
```diff
@@ -0,0 +1,149 @@
...
+      const feeRecords = [
+        FeeRecordEntityMockBuilder.forReport(reconciliationInProgressReport).withFacilityId('12345678').build(),
+        FeeRecordEntityMockBuilder.forReport(reconciliationInProgressReport).withFacilityId('87654321').build(),
+      ];
+
+      reconciliationInProgressReport.feeRecords = feeRecords;
+
+      findOneSpy.mockResolvedValue(reconciliationInProgressReport);
+
+      const mappedFeeRecords = feeRecords.map((feeRecord) => ({ facilityId: feeRecord.facilityId }));
```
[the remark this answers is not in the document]
JoshBinns2000: In truth I was just being lazy when I wrote this as I agree that you should be manually constructing these 😅

trade-finance-manager-ui/component-tests/utilisation-reports/_macros/report-reconciliation-table.component-test.js
JoshBinns2000: It does indeed pass even if the feature flag is set to true. I added an extra line to check that the `<a>` element does not exist and then it works as intended

trade-finance-manager-ui/component-tests/utilisation-reports/utilisation-report-reconciliation-for-bank.component-test.js
JoshBinns2000: I think this was one of the things I added last December hence it lacking any real substance. I updated this test in FN-1615 with the same ideas you've suggested, but I'll move those updates to this PR :)

trade-finance-manager-ui/server/api-response-types/utilisation-report-response-body.ts
JoshBinns2000: Nope 😅 - will remove it

trade-finance-manager-ui/templates/utilisation-reports/utilisation-report-reconciliation-for-bank.njk
JoshBinns2000: For now, I'll just remove the TODOs. In terms of keeping the four tabs, I don't see much downside in meeting the ACs here and removing specific tabs later if we decide not to keep them, but I'll do whatever @ChristianMcc says :)

trade-finance-manager-api/.eslintrc.js
```diff
@@ -58,7 +58,10 @@ module.exports = {
         project: './tsconfig.eslint.json',
         tsconfigRootDir: __dirname,
       },
-      rules: baseRules,
+      rules: {
+        ...baseRules,
+        '@typescript-eslint/no-misused-promises': 'off',
```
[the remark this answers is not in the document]
JoshBinns2000: This one isn't needed

dtfs-central-api/.eslintrc.js
```diff
@@ -59,6 +59,7 @@ module.exports = {
         ...baseRules,
         '@typescript-eslint/no-floating-promises': ['error', { ignoreIIFE: true }],
         '@typescript-eslint/restrict-template-expressions': ['error', { allowNever: true }],
+        '@typescript-eslint/no-misused-promises': 'off',
```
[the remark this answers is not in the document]
JoshBinns2000: This one isn't needed

trade-finance-manager-ui/.eslintrc.js
JoshBinns2000: This one is actually needed as a result of changes the utilisation reports routes file to typescript. Our controllers technically do not match the definition of what express expects the controllers to be: express expects a function with the `(req, res, next?)` parameters and returning `void`, but our functions are usually `async`, meaning they return `Promise<void>` (sometimes they return other things, but this isn't really a problem as we could just move the `return` statement to the next line and everything would work the same). Typescript type-checking isn't annoyed by this, but the typescript linter is as it regards functions returning `Promise<void>` where `void` was expected as a misused promise (hence the name)

trade-finance-manager-ui/templates/utilisation-reports/_macros/report-reconciliation-table.njk
JoshBinns2000: This was originally passed to the checkbox macro as a label for the checkbox, so it can be called something else now. I'm not fully sure what the second part of your question means, but, if I understand correctly, then that won't be a problem now that the checkbox has been moved to the end of the table row

trade-finance-manager-ui/.eslintrc.js
RLavender98: Hmm interesting... I wonder if we should actually disable this for this specific case, whether that is doing inline `// eslint-disable-line @typescript-eslint/no-misused-promises` or if we are able to do something clever in the eslint config to apply a specific rule to route files. It just seems like a rule that is otherwise generally useful so feels a shame to turn it off otherwise?

Is it right that it is ok having async controllers returning promises because our controllers are the last thing our routes call to?

trade-finance-manager-ui/templates/utilisation-reports/_macros/report-reconciliation-table.njk
RLavender98: Sorry, I just mean have you checked the table is sensibly navigable by keyboard, and then separately if you have tried with a screenreader. I think most interesting is to check if what is announced when landing on a checkbox is meaningful and clear given the text content presumably telling you which bank it belongs to is now in a link.

trade-finance-manager-ui/.eslintrc.js
JoshBinns2000: I'm not sure it is necessarily worthwhile to try to keep the eslint rule enabled because I'm not sure that it provides anything useful. Its main utility seems to be for situations where you are wrapping everything with promises when you don't need to and, even though that code will run, there is an overhead to building code like that. Do you have any situations in mind where this rule could catch something?

Also, I believe it's fine for both middleware and the controller to return `Promise<void>` instead of `void` because everything in express seems to execute asynchronously (I will look into this now and leave a follow up comment)

trade-finance-manager-ui/server/controllers/utilisation-reports/utilisation-report-reconciliation-for-bank/index.ts
RLavender98, line 1: This file also needs it's name updating, since it is actually on a report basis not a bank basis

RLavender98 (commented):
Just a few final name change requests!

trade-finance-manager-ui/.eslintrc.js
RLavender98: From [the docs](https://typescript-eslint.io/rules/no-misused-promises/) it seems it helps you spot places you have forgotten an await e.g. if you try and spread the result of an api call into a new object without awaiting then it will give you the no-misused-promises warning, or if you try to use a Promise<boolean> inside an if statement having forgotten to await it.

RLavender98: I've had a read also and agree their docs do suggest the actual returning of the promise is fine in this case, not sure why express has typed it as such...

trade-finance-manager-ui/templates/utilisation-reports/utilisation-report-reconciliation-for-bank.njk
ChristianMcc: I reckon remove TODOs, leave in tabs (because it's probably easier to raise a new PR than to go through changing ACs and while I'm unsure if the last tab will exist I'm not confident enough to base a decision on that just yet :D

trade-finance-manager-ui/.eslintrc.js
JoshBinns2000: I think that side of the benefits from `no-misused-promises` is also captured by the [`no-floating-promises`](https://typescript-eslint.io/rules/no-floating-promises/) rule. Can you think of any cases where `no-floating-promises` will miss something that `no-misused-promises` would catch?

RLavender98: This is obviously rather manufactured but if you turn of no-misused-promises this doesn't show any error:
```
const reportsForPeriodPromise = api.getUtilisationReports(bankId, {
    reportPeriod,
  });

  const obj = {
    ...reportsForPeriodPromise
  };

  const reportsForPeriod = await reportsForPeriodPromise;
```

JoshBinns2000: Yeh that is interesting - the message you get there is `Expected a non-Promise value to be spreaded in an object` which is a strange one.

Do you think disabling this rule only for files in the `server/routes` directory would be a good compromise here?

RLavender98 (approved):

trade-finance-manager-ui/templates/utilisation-reports/_macros/report-reconciliation-table.njk
JoshBinns2000: I think the checkbox doesn't have any good accessibility with it, but that probably needs to be logged as a new ticket and fixed in that. I'm going to resolve this conversation for now as I can't merge with unresolved conversations, but definitely something we should make a ticket for I think

PR #2994, opened 2024-04-04
fix(FN-1862): remove extra db queries in reconciliation summaries

JoshBinns2000:
## Introduction :pencil2:
The reconciliation summaries controller in `dtfs-central-api` was querying the `FeeRecord` table unnecessarily as typeorm supports performing the joins to get all `FeeRecord`s linked to the found `UtilisationReport` in a single query.

- Updates the controller helpers
- Updates the unit tests
- Fixes an issue with the api test where the mock bank was not being properly inserted

### DIFF
--- a/dtfs-central-api/src/v1/routes/utilisation-reports-routes.js
+++ b/dtfs-central-api/src/v1/routes/utilisation-reports-routes.js
@@ -10,7 +10,6 @@ const {
   getUtilisationReportsReconciliationSummary,
 } = require('../controllers/utilisation-report-service/get-utilisation-reports-reconciliation-summary.controller');
 const putUtilisationReportStatusController = require('../controllers/utilisation-report-service/put-utilisation-report-status.controller');
-const { sqlIdValidation } = require('../validation/route-validators/route-validators');
 
 const utilisationReportsRouter = express.Router();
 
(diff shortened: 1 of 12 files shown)


dtfs-central-api/api-tests/v1/utilisation-reports/utilisation-reports-get-reconciliation-summary.api-test.js
```diff
@@ -1,36 +0,0 @@
...
-      const submissionMonth = '2023-11';
-
-      // Act
-      const { status } = await api.get(`/v1/utilisation-reports/reconciliation-summary/${submissionMonth}`);
-
-      // Assert
-      expect(status).toEqual(200);
-    });
-
-    it('returns a 400 response when the submissionMonth is not a valid ISO month', async () => {
```
RLavender98, line 25: This test doesn't seem to have been carried over is that intentional?

RLavender98 (approved):
I think this is a good change, nice spot! It's going to conflict with my branch so if you could give me a heads up when you merge it please 😅

dtfs-central-api/api-tests/v1/utilisation-reports/utilisation-reports-get-reconciliation-summary.api-test.js
JoshBinns2000: Because it returns a `400`, I didn't see much value in adding any checks for the response body. Do you think we should add a check for the error message?

PR #3023, opened 2024-04-11
fix(DTFS2-7091): Handle invalid TFM-UI user session

abhi-markan:
## Introduction :pencil2:
Ensure TFM-UI invalid user session does not crash the container

* Deal and user session validation, if void user will be re-directed to problem with service page.

* Added few logs for troubleshooting.
* Improved test coverage.
* Lint fixes.

### DIFF
--- a/portal-api/src/v1/controllers/deal-status.controller.js
+++ b/portal-api/src/v1/controllers/deal-status.controller.js
@@ -66,6 +66,8 @@ exports.update = async (req, res) => {
 
     const currentStatus = deal.status;
 
+        console.info('Updating portal deal %s status to %s from %s', dealId, newStatus, currentStatus);
+
     if (newStatus !== CONSTANTS.DEAL.DEAL_STATUS.READY_FOR_APPROVAL && newStatus !== CONSTANTS.DEAL.DEAL_STATUS.ABANDONED) {
       if (!userCanSubmitDeal(deal, user)) {
         return res.status(HttpStatusCode.Unauthorized).send();
--- a/trade-finance-manager-ui/server/controllers/case/underwriting/index.js
+++ b/trade-finance-manager-ui/server/controllers/case/underwriting/index.js
@@ -1,4 +1,5 @@
 const { format, fromUnixTime } = require('date-fns');
+const { isEmpty } = require('lodash');
 const api = require('../../../api');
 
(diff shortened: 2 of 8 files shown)


portal-api/src/v1/controllers/deal-status.controller.js
```diff
@@ -66,6 +66,8 @@ exports.update = async (req, res) => {
 
     const currentStatus = deal.status;
 
+        console.info('Updating portal deal %s status to %s from %s', dealId, newStatus, currentStatus);
```
AlexBramhill, line 69: fwiw - over indented?

AlexBramhill (dismissed):

portal-api/src/v1/controllers/deal-status.controller.js
abhi-markan: For troubleshooting with race condition.

trade-finance-manager-ui/server/controllers/case/underwriting/index.js
```diff
@@ -20,11 +21,21 @@ const { hasAmendmentInProgressDealStage, amendmentsInProgressByDeal } = require(
  * renders underwriting page with deal, tfm, dealId, user, leadUnderwriter, pricingAndRisk, underwriterManagersDecision
  */
 const getUnderwriterPage = async (req, res) => {
-  const dealId = req.params._id;
+  const dealId = req?.params?._id;
   const { user, userToken } = req.session;
 
+  if (!dealId || !user || !userToken) {
```
AlexBramhill, line 27: Does our tfm middleware not handle having a session already (am assuming it's like portal) :) (so wouldn't need user token or user check here?)?

trade-finance-manager-ui/server/controllers/case/underwriting/lead-underwriter/index.js
```diff
@@ -8,7 +8,12 @@ const getLeadUnderwriter = async (deal, user, token) => {
   let currentLeadUnderWriter;
   let currentLeadUnderWriterUserId;
 
-  if (deal.tfm.leadUnderwriter) {
+  if (!deal || !user || !token) {
```
AlexBramhill, line 11: same re user and token :)

PR #3026, opened 2024-04-15
feat(FN-2333): mobile screenreader error link focus order

RLavender98:
## Introduction :pencil2:
Ensure mobile screenreader anchor link gives focus to targeted element

- added a tabIndex of -1 to anchor link target which fixes the issue of the mobile screenreader losing the focus order upon clicking the anchor link


portal/templates/utilisation-report-service/utilisation-report-upload/check-the-report.njk
```diff
@@ -18,7 +18,7 @@
     </h1>
     {% if validationErrors %}
         <div class="govuk-form-group govuk-form-group--error govuk-!-margin-top-7">
-            <h2 class="govuk-heading-m" id="validation-errors-table">Errors found in {{ filename }}</h2>
+            <h2 class="govuk-heading-m" id="validation-errors-table" tabIndex="-1">Errors found in {{ filename }}</h2>
```
francescastocco, line 21: Ooh is it `tabIndex` or `tabindex`?
https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex suggests lowercase `i` but I don't know if that's important?

francescastocco (commented):
Considering the 2nd Note on [this page](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex) uses error summary as an explicit example of when to use `tabindex` I think this is a good solution.

And as you said in slack the issue raised by a gov dev [here](https://axesslab.com/skip-links/) looks to be referring to using `tabindex` on the main element which we aren't doing.

portal/templates/utilisation-report-service/utilisation-report-upload/check-the-report.njk
RLavender98: Ah whoops I'm still living in a react world 😅

RLavender98:
Converting to draft as this is now going to be in release-2.1 so we shouldn't merge it into the SQL branch

francescastocco (approved):

PR #3028, opened 2024-04-15
fix(DTFS2-7019): get API call race condition build-up

abhi-markan:
## Introduction :pencil2:
A sporadic issue has been identified where when a MIA deal is submitted to TFM on the second submission (to be coverted to a MIN), the deal does not successfully converts to MIN. After a prologues RCA following was identified:

* `submissionCount` is set to `1` when `TFM-API` fetches deal from `deal.submit.controller.js`, however `Portal-API` updated the submission count to `2` before submitting the deal to TFM (before TFM fetches the portal deal for further processing).
* The issue is sporadic in nature and seems to suggest either a delay in portal deal update in `deals` collection or an ambigious race condition being build up.

* Added an artificial `200ms` delay as the RCA has identified `53ms` delay in `lastUpdatedAt` timestamp property field when fetching the `BSS/EWCS` deal since portal deal last update from `portal-api`.

* Addec complimentary unit test cases.

### DIFF
--- a/trade-finance-manager-api/src/v1/controllers/deal.submit.controller.js
+++ b/trade-finance-manager-api/src/v1/controllers/deal.submit.controller.js
@@ -1,4 +1,3 @@
-const { delay } = require('lodash');
 const { findOneTfmDeal, findOnePortalDeal, findOneGefDeal } = require('./deal.controller');
 const { addPartyUrns } = require('./deal.party-db');
 const { createDealTasks } = require('./deal.tasks');
@@ -20,6 +19,7 @@ const { updatePortalDealFromMIAtoMIN } = require('./update-portal-deal-from-MIA-
 const { sendDealSubmitEmails, sendAinMinAcknowledgement } = require('./send-deal-submit-emails');
 const mapSubmittedDeal = require('../mappings/map-submitted-deal');
 const { dealHasAllUkefIds, dealHasAllValidUkefIds } = require('../helpers/dealHasAllUkefIds');
+const { delay } = require('../helpers/delay');
 
 /**
  * Retrieves a deal from the portal based on the provided deal ID and deal type.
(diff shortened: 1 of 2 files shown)


AlexBramhill (approved):
happy to have this to test this is where the issue is coming from

trade-finance-manager-api/src/v1/controllers/deal.submit.controller.js
```diff
@@ -29,6 +30,14 @@ const { dealHasAllUkefIds, dealHasAllValidUkefIds } = require('../helpers/dealHa
 const getPortalDeal = async (dealId, dealType) => {
   let deal;
 
+  /**
+   * 0.2s (200ms) delay
+   * Below is added to avoid any race condition being build up,
+   * where portal deal updates `deals` collection will not be
+   * fetched when updated on close proximity calls.
+   */
+  delay(200);
```
avaitonis, line 39: Do we need `await` here?

trade-finance-manager-api/src/v1/helpers/delay.js
```diff
@@ -0,0 +1,11 @@
+/**
+ * Returns a promise that resolves after a specified delay in milliseconds.
+ * @param {number} ms - The delay in milliseconds. Default value is 1000.
+ * @returns {Promise} - A promise that resolves after the specified delay.
+ */
+const delay = (ms = 1000) =>
```
avaitonis, line 6: Found alternative solution online https://www.npmjs.com/package/delay

```If you target Node.js 16 or later, you can use import {setTimeout} from 'node:timers/promises'; await setTimeout(1000); instead```

File: trade-finance-manager-api/src/v1/controllers/deal.submit.controller.js [hunk 3]
```diff
@@ -46,6 +47,14 @@ const getPortalDeal = async (dealId, dealType) => {
  * Azure function
  */
 const submitDealAfterUkefIds = async (dealId, dealType, checker) => {
+  /**
+   * 0.2s (200ms) delay
+   * Below is added to avoid any race condition being build up,
+   * where portal deal updates `deals` collection will not be
+   * fetched when updated on close proximity calls.
+   */
+  await delay(() => {}, 200);
```
avaitonis, line 56: Lodash will delay execution of `() => {}`, but not main thread.

lodash.delay returns timerId, so execution of `() => {}` can be cancelled, but this is not promise. I don't think await will wait timer to execute.

avaitonis (approved):
This quick workaround looks good.

PR #3033, opened 2024-04-16
fix(dtfs2 6119): update prevent double click occuring

AlexBramhill:
## Introduction :pencil2:
Emails are sometimes being resent due to double clicking submit buttons. 
We currently have `preventDoubleClick` on in all our ui services, but this only provides a second of double click prevention. Due to long submit times on some of our forms, we occasionally receive double submissions on these forms.

Add a js modification to disable the button used for submission on click. This only is done on the last form in case of multiple forms on page (not present currently, but exist in other gov projects).

 - I've kept the `preventDoubleClick` logic for buttons outside of forms


libs/common/src/ui-scripts/disableFormOnSubmit.js
```diff
@@ -0,0 +1,32 @@
+// Prevent multiple form submissions by pressing enter or clicking the submit button
+// The govukButton component has a preventDoubleClick property, but this only debounces in a one-second window,
+// and is therefore not sufficient for our needs as multiple forms take longer than a second to submit.
+
+const addDisableFormOnSubmit = () => {
+  const forms = document.querySelectorAll('form');
+  // only take the last form on the page, as it is foreseeable to have multiple forms on a page,
+  // and the 'continue' button should always be in the last form on a page.
+  const lastForm = forms[forms.length - 1];
```
abhi-markan, line 9: Need to ensure `forms.length` is not zero.

libs/common/src/ui-scripts/disableFormOnSubmit.js
```diff
@@ -0,0 +1,32 @@
...
+  lastForm.addEventListener('submit', (e) => {
+    if (hasSubmitted) {
+      // prevent multiple form submissions by keyboard
+      e.preventDefault();
+    }
+    hasSubmitted = true;
+
+    const buttons = document.querySelectorAll('.govuk-button');
+    buttons.forEach((button) => {
+      button.setAttribute('disabled', '');
```
abhi-markan, line 26: Will this also mark the button style to `greyed out`, should we add appropriate class to disabled button for enhanced user experience?

libs/common/src/ui-scripts/disableFormOnSubmit.js
```diff
@@ -0,0 +1,32 @@
...
+  let hasSubmitted = false;
+
+  lastForm.addEventListener('submit', (e) => {
+    if (hasSubmitted) {
+      // prevent multiple form submissions by keyboard
+      e.preventDefault();
+    }
+    hasSubmitted = true;
+
+    const buttons = document.querySelectorAll('.govuk-button');
```
abhi-markan, line 24: Should this only target `submit` button only?

abhi-markan (commented):

libs/common/src/ui-scripts/disableFormOnSubmit.js
AlexBramhill: Sounds good, I'll do this so it's clearer (as if `querySelectorAll` returns an empty array, `  const lastForm = forms[forms.length - 1];` would equal `undefined` that would be picked up on line 11, but I agree this can be clearer.

libs/common/src/ui-scripts/disableFormOnSubmit.js
AlexBramhill: This does style it as greyed out already as the css selector targets `disabled`  (this is also standard html behaviour, but its good to know were also accessing our own style sheets
```
.govuk-button--disabled,
.govuk-button[disabled='disabled'],
.govuk-button[disabled] {
  opacity: 0.5;
}
```

libs/common/src/ui-scripts/disableFormOnSubmit.js
AlexBramhill: I'd love to do this, but it appears that our form buttons arent always `submit` -- prime example being our login form button.

abhi-markan (approved):

abhi-markan (dismissed):

libs/common/src/ui-scripts/disable-form-submit-on-submission.js
```diff
@@ -0,0 +1,120 @@
...
+
+  if (!lastForm) {
+    return null;
+  }
+
+  return lastForm;
+};
+
+const validateAndDeleteElement = (elementToDelete) => {
+  if (elementToDelete instanceof HTMLElement) {
```
AlexBramhill, line 24: These validations are to make eslint happy

libs/common/src/ui-scripts/disable-form-submit-on-submission.js
```diff
@@ -0,0 +1,116 @@
...
+ * This hidden input copies any name or value attributes that the submission submitter has,
+ * as when we disable the initial submitter, the disabled submitter will not submit
+ * this information to the server.
+ * @param {HTMLElement} submitter
+ */
+const createHiddenInputOfSubmitter = (submitter) => {
+  const hiddenInput = document.createElement('input');
+
+  hiddenInput.setAttribute('id', 'resubmit-prevention-hidden-input');
+  hiddenInput.setAttribute('type', 'hidden');
```
MarRobSoftwire, line 54: Think this is ok accessibility-wise but just want to double check we dont need `aria-hidden` or similar

libs/common/src/ui-scripts/disable-form-submit-on-submission.js
```diff
@@ -0,0 +1,116 @@
...
+      return;
+    }
+
+    const hasPreviouslySubmitted = !!priorSubmitter;
+    if (hasPreviouslySubmitted) {
+      const isSameSubmitterAsPrevious = submitter === priorSubmitter;
+      if (isSameSubmitterAsPrevious) {
+        // prevents a duplicate resubmission if a user resubmits the form not using buttons
+        // ie from using keyboard navigation to submit the form
+        event.preventDefault();
```
MarRobSoftwire, line 101: This is scary but understand its needed

libs/common/src/ui-scripts/disable-form-submit-on-submission.js
```diff
@@ -0,0 +1,116 @@
...
+  lastForm.addEventListener('submit', (event) => {
+    const { submitter } = event;
+
+    if (!(submitter instanceof HTMLElement)) {
+      return;
+    }
+
+    const hasPreviouslySubmitted = !!priorSubmitter;
+    if (hasPreviouslySubmitted) {
+      const isSameSubmitterAsPrevious = submitter === priorSubmitter;
```
MarRobSoftwire, line 97: Is there any risk that this could be called with:
1. I submit a form with invalid values
2. I correct the validation errors and try to submit again, but this is called

Have had a check myself & it seems to work ok

portal/public/css/styles.css
MarRobSoftwire, line 1: What change updated this file? Or was it just out of date on main?

portal/public/js/govukFrontend.js.map
MarRobSoftwire, line 1: Also surprised to see this here!

gef-ui/webpack.common.config.js
```diff
@@ -8,6 +8,7 @@ module.exports = {
     govukFrontend: './scripts/govuk-frontend.js',
     jsEnabled: './scripts/js-enabled.js',
     correspondenceAddress: './scripts/correspondence-address.js',
+    disableFormSubmitOnSubmission: '../libs/common/src/ui-scripts/disable-form-submit-on-submission.js',
```
MarRobSoftwire, line 11: Just want to double check: Webpack is only ever called in the workspaces directory set up right - this should never be called within a service when it wont have this to reference

MarRobSoftwire (commented):
LGTM, just a couple questions that probably dont need action

portal/public/js/govukFrontend.js.map
AlexBramhill: Me too... Apparently needed though

portal/public/css/styles.css
AlexBramhill: Think out of date on main

libs/common/src/ui-scripts/disable-form-submit-on-submission.js
AlexBramhill: So you'd only get the validation errors after a `post` request, so this would not run as `hasPreviouslySubmitted` would be `false` :) 

This prevents clicking `submit` and while it is still loading, changing data in an input field and hitting `enter`, causing another submit to fire dispite the button being disabled

libs/common/src/ui-scripts/disable-form-submit-on-submission.js
AlexBramhill: Yeah this is hopefully not that scary -- it's to stop people being hard locked on a screen and also avoiding any previous hidden input values being submitted incorectly --

we sometimes use values on buttons to send values to the server to let us know what action was called on a post request. When we `disable` a button, these no longer get sent so we have to add a hidden input for them.

say you click on a submit button with value `submit`, this creates a hidden input with `submit` value
then you quickly click cancel with value `cancel`, creating a hidden input with the `cancel` value

if we dont clean up the first hidden input, depending on how our server code is written, we might submit the deal anyway

libs/common/src/ui-scripts/disable-form-submit-on-submission.js
AlexBramhill: `hidden` should cover this

gef-ui/webpack.common.config.js
AlexBramhill: This should be fine as we only call webpack config manually (via `npm run build` for example). We don't call this when we build automagically in ci/cd as we need to update the SHA to allow the js to run :)

PR #2948, opened 2024-03-28
feat(DTFS2-7082): add audit logs to update deal snapshot endpoint

MarRobSoftwire:
## Introduction :pencil2:
All endpoints in dtfs-central-api should add `auditDetails` when inserting/updating data. This PR covers the update deal snapshot endpoint

## Resolution :heavy_check_mark:
- Add audit details to database changes for update deal snapshot endpoint

## Note ⚠️ 
This is a breaking change to the endpoint - the data used to be given in the form
```json
{
  "a-new-field": "data"
}
```
But now is of the form
```json
{
  "updateSnapshot": {
    "a-new-field": "data"
  },
  "user": {
    "_id": "example-id",
  }
}
```

PR #3034, opened 2024-04-17
feat(DTFS2-7087): add audit logs to create user endpoint

MarRobSoftwire:
## Introduction :pencil2:
All endpoints in dtfs-central-api should add `auditDetails` when inserting/updating data. This PR covers the create user endpoint

- Add audit details to database changes for create user endpoint

## Note ⚠️ 
AFAIK this endpoint is not used anywhere, either this should be deleted, or the tfm-api create users endpoint call this instead


dtfs-central-api/src/v1/routes/tfm-routes.js
```diff
@@ -671,6 +671,14 @@ tfmRouter.route('/users').get(
  *         application/json:
  *           schema:
  *             $ref: '#/definitions/TFMUser'
+ *             auditDetails:
```
ChristianMcc, line 674: One to think about, if this is the same everywhere we might want to make a definition of it and just reference that everywhere

ChristianMcc (approved):
LGTM :) one thing to think about for these PRs

dtfs-central-api/src/v1/routes/tfm-routes.js
MarRobSoftwire: Its nearly the same, just has different enum values - not sure if there's a nice way to handle this without repetition

MarRobSoftwire: Have decided to just do this in a verbose way as I think its neater - added to https://github.com/UK-Export-Finance/dtfs2/pull/2948 though, not here

PR #3035, opened 2024-04-17
fix(DTFS2-7105): consistent ignore files

abhi-markan:
## Introduction :pencil2:
Ensure consistent `.*ignore` files content

* Single root `.prettierignore` and `.eslintignore` files.
* Removed micro-service specific files, since they will not be required during build process.

* Disabled Jest generated reports as per https://github.com/jthodge/til/blob/main/javascript/disable-jests-auto-generated-coverage-reports.md guide.
* `tsc --noEmit` to be executed on `*.ts` staged files only.
* Few lint fixes.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,1 +1,19 @@
+# Root
+node_modules/
+dist/
+build/
+public/
 generated_reports/
+coverage/
+package*.json
+.cache
+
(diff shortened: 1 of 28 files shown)


avaitonis (approved):
No issues from me

PR #3036, opened 2024-04-17
feat(FN-2293): display report status in portal

RLavender98:
## Introduction :pencil2:
We want to let bank users see the status of reports as they are reconciled.

- Display the report status of previously uploaded reports on the previous reports page in portal

image


RLavender98:
Converting to draft because needs to go into a release-3 branch and not into feat/e2etf-sql

francescastocco (commented):
Looks good!

portal/component-tests/utilisation-report-service/_macros/report-reconciliation-status.component-test.js
JoshBinns2000, line 1: Not something for this PR, but if we start sharing more components between services it may be worthwhile to add a components export to `libs/common` so that they remain consistent

portal/server/types/view-models/previous-reports.ts
JoshBinns2000, line 1: Now that the `lint:fix` command should work (after merging the latest version of `main` in), I think that needs to be run for `portal` to fix formatting errors in this file (and others if there are any)

portal/server/controllers/utilisation-report-service/previous-reports/helpers/previous-reports-view-model-mapper.test.ts
```diff
@@ -0,0 +1,250 @@
...
+        { year: 2024, reports: [] },
+        { year: 2023, reports: [] },
+      ];
+
+      // Act
+      const result = mapToPreviousReportsViewModel('2023', user, responseBody);
+
+      // Assert
+      expect(result.navItems).toHaveLength(2);
+      expect(result.navItems[0]).toEqual({
```
JoshBinns2000, line 73: Does this work if you write `expect(result.navItems).toEqual([{ <navItem1> }, { <navItem2> }]);`?

portal/test-helpers/test-data/bank-report-period-schedule.ts
```diff
@@ -0,0 +1,16 @@
+import { BankReportPeriodSchedule } from '@ukef/dtfs2-common';
+
+export const aMonthlyBankReportPeriodSchedule = (): BankReportPeriodSchedule => [
```
JoshBinns2000, line 3: I'm not so sure about this test helper - I thought the `aThing()` pattern was good for getting a base object which can have it's properties overwritten or modified to match your test case whilst still preserving everything else you would expect it to have. In this case, you would never want to modify anything on the return value of this function, so it doesn't fit that criteria (with the caveat that this criteria could just be something my brain has come up with from our current usage of this pattern).

It's also a bit of a semantic thing in the `aMonthlyBankReportPeriodSchedule` uses the indefinite article, but there is only one monthly bank report period schedule. In this context it can also mean "a monthly bank report period schedule *object*" , which is fine with respect to the indefinite article, so if it is meant in this way then I think it's fine.

This is mainly a concern to me because of our current usage of this pattern, as it may lead some unsuspecting dev to write something like
```typescript
const scheduleWithJanuaryToMarch: BankReportPeriodSchedule = [
  ...aMonthlyBankReportPeriodSchedule(),
  { startMonth: 1, endMonth: 3 },
];
```
which wouldn't work as they expect. I doubt someone would end up doing this so I don't feel strongly about it, but what do you think?

JoshBinns2000 (changes requested):
After the latest changes to `main` get merged in, I think a `npm run lint:fix` in `portal` needs to be run as there are a few formatting issues. Other than that, it looks pretty good and I like the new previous reports work

portal/server/controllers/utilisation-report-service/previous-reports/helpers/previous-reports-view-model-mapper.test.ts
RLavender98: Yeah, I think to me it doesn't make much difference, generally I like try and keep my assertions as specific as possible (although this is covering quite a lot anyway I suppose actually). I can switch it if you think asserting the whole array at once is better?

portal/test-helpers/test-data/bank-report-period-schedule.ts
RLavender98: I actually think the `aThing` helpers are really useful for the tests where we don't want to change anything, just gives us what we need straight up, with zero chance of cross-test interference.
And yeah for me it is the "object"-ness of this that justifies the grammar, I do agreed does feel slightly weirder than some of the other uses.
That example usage would actually be quite a nice way for making an invalid schedule to test the schedule validator for example, and a case where if instead someone accidentally edited a variable we used across tests to achieve it would cause lots of other tests to fail!

JoshBinns2000: That all makes sense to me :)

portal/server/controllers/utilisation-report-service/previous-reports/index.ts
```diff
@@ -1,60 +1,27 @@
...
 
 type GetPreviousReportsRequest = Request & {
   query: {
     targetYear?: string;
   };
 };
 
+const renderPreviousReportsPage = (res: Response, context: PreviousReportsViewModel) => {
+  res.render('utilisation-report-service/previous-reports/previous-reports.njk', context);
+};
```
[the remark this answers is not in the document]
JoshBinns2000: Okay - I think we should leave this for now, but I wonder if there will be a way of writing a declaration file which adds a generic for `res.render` like
```typescript
res.render<RenderOptions extends object>(view: string, options?: RenderOptions, ...)
```
but that's not for this PR

portal/server/controllers/utilisation-report-service/previous-reports/helpers/previous-reports-view-model-mapper.test.ts
JoshBinns2000: I think in this case asserting the whole array would allow you to capture both the length and item assertions in one line, but I don't particularly mind which one we go with

RLavender98: I would say I don't think keeping assertions to as few lines as possible is important, in fact I would almost go the other way, in trying to have each way a test can break as it's own assertion so you immediately see which way the test broke when it breaks. It also gives you a good way of seeing if you are testing too much in one test, if you end up with lots of assertions you might actually be best off splitting the test rather than combining the assertions down into as few assertions as possible.

JoshBinns2000 (approved):
LGTM - if possible, could we add a screenshot of what this adds to the PR description so that, if people come back to this PR, they know what was added in terms of the frontend?

portal/server/controllers/utilisation-report-service/previous-reports/helpers/previous-reports-view-model-mapper.test.ts
JoshBinns2000: I think jest is pretty good at showing you where the test has gone wrong even if you only do one object assertion, but I suppose it depends on how you read tests and what specifically goes wrong - I have seen cases where more assertions makes it harder to figure out where things went wrong and cases where fewer assertions do the same thing. I don't think it makes much difference either way in this particular case though and agree with the general principle that if you're testing a large object which has some mapped values, the tests themselves should probably be split up rather than just the assertions

PR #3037, opened 2024-04-17
fix(DTFS2-7091): validation before updating details deal object

abhi-markan:
## Introduction :pencil2:
During deal update in `deals` collection for `BSS/EWCS` deal numerous deal update API calls are invoked namily (when submitting deal second time to TFM):
* Submission count (deal.details)
* Facilities update timestamp (deal.facilitiesUpdated}

The second call does not update the `deal.details` object, which causes the `deal.details` object to be reverted back to its previous state as only the update is sent to `updateDeal()` function.

* `deal.details` and `deal.eligibility` object validation prior to update.

* Removed non-production code:
  * Console logs
  * Artificial delay
* Various lint fixes


dtfs-central-api/src/v1/controllers/portal/deal/update-deal.controller.js
```diff
@@ -91,47 +77,43 @@ const updateDeal = async (dealId, dealChanges, user, existingDeal, routePath) =>
       }
 
       let originalDealDetails;
+      let originalDealEligibility;
+
+      const update = {
```
avaitonis, line 82: Can we explain this change in code comment? what processes are modifying deal at same time and how we avoid issue.
Things like multiple `updatedAt` updates can happen in future too.

abhi-markan: Good shout, have added.

avaitonis (approved):
LGTM, great catch.

This reduces risk of overriding Deal values by multiple threads, but other global solutions could be considered.

PR #3043, opened 2024-04-19
fix(FN-1859): batch save fee records in uploading report 

ChristianMcc:
## Introduction :pencil2:
We were hitting issues when saving large utilisation reports in the form of a parameter limit being hit in SQL server (2100) which was stopping us uploading these files

I've split up the saving into a transaction that saves the report and then chunks through the saving of the fee records to get around this issue


dtfs-central-api/src/helpers/fee-record.helper.test.ts
```diff
@@ -43,6 +45,7 @@ describe('fee-record.helper', () => {
           lastUpdatedByIsSystemUser: false,
           lastUpdatedByPortalUserId: requestSource.userId,
           lastUpdatedByTfmUserId: null,
+          report: report,
```
ChristianMcc, line 48: oops this should be shorthand

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/report-uploaded/report-uploaded.event-handler.ts
```diff
@@ -19,11 +21,28 @@ export type UtilisationReportReportUploadedEvent = BaseUtilisationReportEvent<'R
...
+    feeRecordCsvRowToSqlEntity({
+      dataEntry,
+      requestSource,
+      report,
+    }),
+  );
+  await transactionEntityManager.save(FeeRecordEntity, feeRecordEntities, { chunk: 100 });
+
+  report.feeRecords = feeRecordEntities;
+  return report;
```
JoshuaB-Coding, line 47: I'm surprised the linter didn't get angry with this bit as well because we're modifying a function parameter. Because we have annotated the return type of this function, I think it would be fine to remove the error by returning `{ ...report, feeRecords: feeRecordEntities };`, but that's only if we think this error is worth addressing

JoshuaB-Coding (commented):
Nice way of addressing the chunking issue - just left a few small comments

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/report-uploaded/report-uploaded.event-handler.ts
JoshuaB-Coding: Just checked and it seems that the linter does get angry with this as well 😆

dtfs-central-api/src/v1/controllers/utilisation-report-service/post-upload-utilisation-report.controller/index.ts
```diff
@@ -73,8 +78,33 @@ export const postUploadUtilisationReport = async (req: PostUploadUtilisationRepo
...
     });
+    await queryRunner.commitTransaction();
+    return updatedReport;
+  } catch (error) {
+    await queryRunner.rollbackTransaction();
+    if (error instanceof ApiError) {
+      throw error;
+    } else {
+      throw new TransactionFailedError();
+    }
```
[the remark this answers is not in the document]
ChristianMcc: Will update the try-catch, the api error comes inside the eventHandler if it's an invalid event and can propagate up to here

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/report-uploaded/report-uploaded.event-handler.ts
ChristianMcc: I'm just going to remove line 46, right now it has no use anyway

dtfs-central-api/src/v1/controllers/utilisation-report-service/post-upload-utilisation-report.controller/index.ts
```diff
@@ -51,16 +52,20 @@ type PostUploadUtilisationReportRequest = CustomExpressRequest<{
...
-/**
- * Controller for the utilisation report upload
- * @param req - The request object
- * @param res - The response object
- */
-export const postUploadUtilisationReport = async (req: PostUploadUtilisationReportRequest, res: Response) => {
-  try {
-    const { reportId, reportData, user, fileInfo } = req.body;
-    const uploadedByUserId = user._id.toString();
+export const uploadReportInTransaction = async (
```
francescastocco, line 55: Does this need a docstring?

dtfs-central-api/src/v1/controllers/utilisation-report-service/post-upload-utilisation-report.controller/index.ts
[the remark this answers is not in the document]
JoshuaB-Coding: I forgot but the `TransactionFailedError` does extend `ApiError`, so it's just a specific case of `ApiError` with a message that looks similar to the one you put (and a 500). So I think you would only want to throw the `TransactionFailedError` rather than having this `if-else`

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/report-uploaded/report-uploaded.event-handler.ts
JoshuaB-Coding: I think it does do something. Without that line, `report.feeRecords` would be `undefined` (or `[]` - I'm not sure which by default but I think `undefined` probably)

ChristianMcc: Ye I suppose I was being lazy and meaning we don't use those feeRecords everywhere. Cause of typing issues I'll just add an `updateWithFeeRecords` method to the utilisationReportEntity to fix this

PR #3044, opened 2024-04-19
fix(DTFS2-6655)!: validate users on portal-api 

vamshidharukef:
## Introduction :pencil2:
There’s no role validation on the users routes in portal-api. n some cases that makes sense (e.g. the PUT endpoint needs to be open to everyone so they can update their password, and the user needs to give their current password in the request body to do this so it's secure).
GET ‘/users' = this endpoint is only used in admin routes in Portal UI and by admin users in data-migration and mock-data-loader, so can be 'admin' only.
GET ‘/users/:_id' = this endpoint is used by admin roles in Portal UI, but also if a Portal UI user wants to view their own profile. Hence, we should validate if the user’s _id in req.user matches that in the URL, or if they have the 'admin' role.
PUT ‘/users/:_id' = this endpoint is used by admin routes in Portal UI, but also if a Portal UI user wants to change their own password. Hence, as above, we should validate if the user’s _id in req.user matches that in the URL, or if they have the 'admin' role.
DELETE '/users/:_id' = this endpoint is only used by admin users in data-migration and mock-data-loader.
DELETE '/users/:_id/disable' = this endpoint appears to be unused and should be removed. There is a test for it in the Portal API API tests but it does not appear anywhere else in the code base. There is a GET '/users/disable/:_id' route in Portal UI which renders user-disable.njk, but there is no corresponding POST route, hence both the route and the Nunjucks file  removed.

Updated tests


portal-api/api-tests/v1/users/post-create-user.api-test.js
```diff
@@ -120,8 +120,19 @@ describe('a user', () => {
...
+          const user1 = body.users.find((user) => user.username === MOCK_USER.username);
+          if(user1){
+            expect(user1.roles).toStrictEqual([READ_ONLY, READ_ONLY]);
+          } else {
+            throw new Error(`User ${MOCK_USER.username} not found`);
+          }
+        } else {
+          expect(body.users).toBeUndefined();
+        }
+        // expect(body.users.find((user) => user.username === MOCK_USER.username).roles).toStrictEqual([READ_ONLY, READ_ONLY]);
```
oscar-richardson-softwire, line 135: As this test is just about checking that a user can be created with the `READ_ONLY` role repeated, it might be better to revert the `expect` statements to how they were before and instead change line 121 to make the `GET` request `as(anAdmin)` (because you've changed `GET /users` to be admin-only). This is because we don't really care about the fact that making a `GET users` request as a non-admin returns a `403` for the purposes of this test, as this test is about `POST /users`.

portal-api/api-tests/v1/users/post-create-user.api-test.js
```diff
@@ -133,8 +144,19 @@ describe('a user', () => {
...
+          const user1 = body.users.find((user) => user.username === MOCK_USER.username);
+          if(user1){
+            expect(user1.roles).toStrictEqual([READ_ONLY, READ_ONLY]);
+          } else {
+            throw new Error(`User ${MOCK_USER.username} not found`);
+          }
+        } else {
+          expect(body.users).toBeUndefined();
+        }
+        // expect(body.users.find((user) => user.username === MOCK_USER.username).roles).toStrictEqual([READ_ONLY]);
```
oscar-richardson-softwire, line 159: Same comment as the test above - this test is also testing `POST /users` and doesn't really care about the fact that `GET /users` returns a `403` for non-admin users.

portal-api/api-tests/v1/users/post-create-user.api-test.js
```diff
@@ -88,9 +88,9 @@ describe('a user', () => {
     describe('it creates the user', () => {
       it('it creates the user if all provided data is valid', async () => {
         await createUser(MOCK_USER);
-        const { status, body } = await as(aNonAdmin).get(BASE_URL);
+        const { status, body } = await as(ADMIN).get(BASE_URL);
 
-        expect(status).toEqual(200);
+        expect(status).toEqual(401);
```
oscar-richardson-softwire, line 93:
```suggestion
        const { status, body } = await as(anAdmin).get(BASE_URL);

        expect(status).toEqual(200);
```

oscar-richardson-softwire: Apologies for not noticing this on our call - this should've been `anAdmin`, not `ADMIN`, as the latter is just a constant for the string `'admin'`!

portal-api/api-tests/v1/users/put-update-user.api-test.js
```diff
@@ -164,7 +164,7 @@ describe('a user', () => {
 
         await as(createdUser).put(updatedUserCredentials).to(`/v1/users/${createdUser._id}`);
 
-        const { status } = await as(aNonAdmin).get(`/v1/users/${createdUser._id}`);
+        const { status } = await as(createdUser).get(`/v1/users/${createdUser._id}`);
 
         expect(status).toEqual(200);
```
oscar-richardson-softwire, line 169: This test actually never checked what it was supposed to in the first place. It makes a request to `PUT users/:id` to update the user's password but then checks the status of a request to `GET users/:id`, which will be `200` regardless of the outcome of the request to update the password. The password update was actually failing with `401` because the `createdUser` was not being set up with a bearer token correctly. This test should instead check that the update password request succeeds with a `200`, that the user can't log in with their old password and that they can log in with their new password.

oscar-richardson-softwire: The test should look something like this:
`it('a user can update their own password', async () => {
        const updatedPassword = 'updatedPassword1!';
        const updatedUserCredentials = {
          currentPassword: initialPassword,
          password: updatedPassword,
          passwordConfirm: updatedPassword,
        };
        const { status: updatePasswordStatus } = await as(createdUser).put(updatedUserCredentials).to(`/v1/users/${createdUser._id}`);
        expect(updatePasswordStatus).toEqual(200);
        const { status: loginWithInitalPasswordStatus } = await as().post({ username: createdUser.username, password: initialPassword }).to('/v1/login');
        expect(loginWithInitalPasswordStatus).toEqual(401);
        const { status: loginWithUpdatedPasswordStatus } = await as().post({ username: createdUser.username, password: updatedPassword }).to('/v1/login');
        expect(loginWithUpdatedPasswordStatus).toEqual(200);
});`

oscar-richardson-softwire: Similarly, the `as admin` => `a user's password can be updated` test should be changed to follow this pattern.

oscar-richardson-softwire: In order to make this work, you'll need to set up the bearer token correctly for the `createdUser` by adding the following lines at the end of the `beforeEach` block:

```
const initialUserCredentials = {
      password: initialPassword,
      passwordConfirm: initialPassword,
    };
await as(anAdmin).put(initialUserCredentials).to(`/v1/users/${createdUser._id}`);
const { token } = await createLoggedInUserSession(createdUser);
createdUser.token = token;
```

portal-api/src/v1/users/routes.js
```diff
@@ -26,7 +26,27 @@ const userService = new UserService();
...
+  if (req.user && req.user.roles.includes('admin')) {
+    // User is an admin, allow access to the route
+    next();
+  } else {
+    // User is not an admin, send a 403 Forbidden response
+    res.status(403).json({ error: 'Unauthorized' });
+  }
+};
+
+const isUserOrAdmin = (req, res, next) => {
```
AlexBramhill, line 39: As mentioned in a call -- I think it might be nicer if we kept everything in our routes: `.\portal-api\src\v1\routes.js` and use our existing middleware `validateUserHasAtLeastOneAllowedRole({ allowedRoles: [ADMIN] })` for when we need to check that a user is admin.

We'll need to write some new middleware following the same pattern as `validateUserHasAtLeastOneAllowedRole` to add in validation for `validateUserIsAdminOrIsActingOnSelf`. For now, we could try ``validateUserIsActingOnSelfToChangePasswordOrIsAdmin`. Ideally, we could later abstract that to be `validateUserIsAdminOrIsActingOnSelf({actingOnSelfRules: changingPasswordRule})` which would be great and reusable.

We should look to add tests for the middleware `validateUserIsAdminOrIsActingOnSelf` in much the same way as `validateUserHasAtLeastOneAllowedRole` already has.

We could also look at adding `withValidateUserHasAtLeastOneAllowedRole` tests to our api tests, much like we have `withValidateUsernameAndEmailMatchTests`, but I think that might be outside the scope of this ticket for now

portal-api/api-tests/v1/users/post-create-user.api-test.js
vamshidharukef: Changed line 121 to make the `GET` request `as(ADMIN)`

portal-api/api-tests/v1/users/post-create-user.api-test.js
vamshidharukef: Changed as above comment

portal-api/api-tests/v1/users/put-update-user.api-test.js
vamshidharukef: I had added `initialPassword` and imported `createLoggedInUserSession`

portal-api/api-tests/v1/users/with-validate-password.api-tests.js
```diff
@@ -114,8 +119,9 @@ function itShouldReturnAnErrorIf(testCases) {
     async ({
       makeRequestWithModifiedPayload,
       expectResponseErrorToMatchExpected = (body) => expect(body.errors.errorList.password).toEqual(PASSWORD_ERROR),
+      middleware,
     }) => {
-      const { status, body } = await makeRequestWithModifiedPayload();
+      const { status, body } = await makeRequestWithModifiedPayload(middleware);
```
abhi-markan, line 124: Why is the middleware being passed as an argument, can a standard pattern be followed please.

portal-api/src/v1/roles/validate-user-has-at-least-one-allowed-role.js
```diff
@@ -9,9 +9,14 @@ const { userHasAtLeastOneAllowedRole } = require('./user-has-at-least-one-allowe
 const validateUserHasAtLeastOneAllowedRole = ({
   allowedRoles,
 }) => (req, res, next) => {
+
   const { user } = req;
 
-  if (!userHasAtLeastOneAllowedRole({ user, allowedRoles })) {
+  if (!user) {
+    throw new Error(`Failed to create user session with status ${allowedRoles}: user is null`);
```
abhi-markan, line 16: Please add `console.error`.

portal-api/src/v1/roles/validate-user-is-acting-on-self-to-change-password-or-is-admin.js
```diff
@@ -0,0 +1,23 @@
+// const { userHasAtLeastOneAllowedRole } = require('./user-has-at-least-one-allowed-role');
```
abhi-markan, line 1: Commented out code.

portal-api/src/v1/roles/validate-user-is-acting-on-self-to-change-password-or-is-admin.js
```diff
@@ -0,0 +1,23 @@
...
+ * old password have at least one of the allowedRoles, and calls the next handler otherwise.
+ * @param {{ allowedRoles: string[] }}
+ * @returns {(req, res, next) => void}
+ */
+const validateUserIsActingOnSelfToChangePasswordOrIsAdmin = (req, res, next) => {
+  if (!req.user) {
+    return res.status(401).send('Unauthorized');
+  }
+
+  if (req.user._id !== req.params._id && !req.user.roles.includes('admin')) {
```
abhi-markan, line 14: Optional chaining please.

portal-api/src/v1/roles/validate-user-is-acting-on-self-to-change-password-or-is-admin.js
```diff
@@ -0,0 +1,23 @@
+// const { userHasAtLeastOneAllowedRole } = require('./user-has-at-least-one-allowed-role');
+
+/**
+ * Creates a middleware that returns a 401 error if the current user logs with
+ * old password have at least one of the allowedRoles, and calls the next handler otherwise.
+ * @param {{ allowedRoles: string[] }}
+ * @returns {(req, res, next) => void}
+ */
+const validateUserIsActingOnSelfToChangePasswordOrIsAdmin = (req, res, next) => {
```
abhi-markan, line 9: Perhpas shorted function name.

portal-api/src/v1/routes.js
```diff
@@ -103,9 +103,10 @@ authRouter
   .get(mandatoryCriteria.findOne)
   .delete(validateUserHasAtLeastOneAllowedRole({ allowedRoles: [ADMIN] }), mandatoryCriteria.delete);
 
-authRouter.route('/users').get(users.list).post(users.create);
-authRouter.route('/users/:_id').get(users.findById).put(users.updateById).delete(users.remove);
-authRouter.route('/users/:_id/disable').delete(users.disable);
+authRouter.route('/users').get(users.list).post(validateUserHasAtLeastOneAllowedRole({ allowedRoles: [ADMIN] }), users.create);
+authRouter.route('/users/:_id').get(validateUserIsActingOnSelfToChangePasswordOrIsAdmin, users.list);
+authRouter.route('/users/:_id').get(users.findById).put(users.updateById).delete(validateUserHasAtLeastOneAllowedRole({ allowedRoles: [ADMIN] }), users.remove);
```
abhi-markan, line 108: Is `disabled` endpoint deprecated?

abhi-markan: If yes then please update the swagger documentation and mark this PR as breaking.

portal-api/src/v1/users/routes.js
```diff
@@ -31,6 +31,26 @@ const userService = new UserService();
 const signInLinkService = new SignInLinkService(randomGenerator, hasher, userRepository, userService);
 const signInLinkController = new SignInLinkController(signInLinkService);
 
+// const isAdmin = (req, res, next) => {
```
abhi-markan, line 34: Please ensure no commented code is included in the PR for review.

abhi-markan (changes requested):

portal-api/api-tests/api-test-users.js
```diff
@@ -259,7 +259,7 @@ const apiTestUser = {
   surname: 'Test User',
   email: 'api-tester@ukexportfinance.gov.uk',
   timezone: 'Europe/London',
-  roles: [MAKER],
+  roles: [ADMIN],
```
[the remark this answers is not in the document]
vamshidharukef: Added new test user with `admin` role

portal-api/api-tests/v1/users/post-create-user.api-test.js
```diff
@@ -87,28 +87,25 @@ describe('a user', () => {
 
     describe('it creates the user', () => {
       it('it creates the user if all provided data is valid', async () => {
-        await createUser(MOCK_USER);
-        const { status, body } = await as(aNonAdmin).get(BASE_URL);
+        // await createUser(MOCK_USER);
```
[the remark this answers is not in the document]
vamshidharukef: Removed comment

portal-api/api-tests/v1/users/post-create-user.api-test.js
```diff
@@ -87,28 +87,25 @@ describe('a user', () => {
 
     describe('it creates the user', () => {
       it('it creates the user if all provided data is valid', async () => {
```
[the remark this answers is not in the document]
vamshidharukef: Added assertion for Maker and Checker.

portal-api/api-tests/v1/users/put-update-user.api-test.js
```diff
@@ -156,17 +166,18 @@ describe('a user', () => {
...
         };
-
-        await as(createdUser).put(updatedUserCredentials).to(`/v1/users/${createdUser._id}`);
-
-        const { status } = await as(aNonAdmin).get(`/v1/users/${createdUser._id}`);
-
-        expect(status).toEqual(200);
+        const { status: updatePasswordStatus } = await as(createdUser).put(updatedUserCredentials).to(`/v1/users/${createdUser._id}`);
+        expect(updatePasswordStatus).toEqual(200);
+        const { status: loginWithInitalPasswordStatus } = await as().post({ username: createdUser.username, password: initialPassword }).to('/v1/login');
```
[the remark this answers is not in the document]
vamshidharukef: Added assertion for `maker` and `checker`.

portal-api/api-tests/v1/users/users.api-test.js
```diff
@@ -250,6 +252,6 @@ describe('a user', () => {
   });
 
   async function createUser(userToCreate) {
-    return as(aNonAdmin).post(userToCreate).to('/v1/users');
+    return as(anAdmin).post(userToCreate).to('/v1/users');
```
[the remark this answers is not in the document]
vamshidharukef: Added assertion for `maker` and `checker`.

portal-api/api-tests/v1/users/with-validate-password.api-tests.js
vamshidharukef: Removed argument.

portal-api/src/v1/roles/validate-user-has-at-least-one-allowed-role.js
vamshidharukef: Added `console.error`.

portal-api/src/v1/roles/validate-user-is-acting-on-self-to-change-password-or-is-admin.js
vamshidharukef: Removed comment

portal-api/src/v1/roles/validate-user-is-acting-on-self-to-change-password-or-is-admin.js
vamshidharukef: Added optional chaining.

portal-api/src/v1/routes.js
```diff
@@ -103,9 +103,10 @@ authRouter
   .get(mandatoryCriteria.findOne)
   .delete(validateUserHasAtLeastOneAllowedRole({ allowedRoles: [ADMIN] }), mandatoryCriteria.delete);
 
-authRouter.route('/users').get(users.list).post(users.create);
-authRouter.route('/users/:_id').get(users.findById).put(users.updateById).delete(users.remove);
-authRouter.route('/users/:_id/disable').delete(users.disable);
+authRouter.route('/users').get(users.list).post(validateUserHasAtLeastOneAllowedRole({ allowedRoles: [ADMIN] }), users.create);
+authRouter.route('/users/:_id').get(validateUserIsActingOnSelfToChangePasswordOrIsAdmin, users.list);
```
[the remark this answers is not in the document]
vamshidharukef: It fetches all the users with `admin`.

portal-api/src/v1/users/routes.js
vamshidharukef: Removed comment

portal-api/src/v1/routes.js
vamshidharukef: According to point 9 https://ukef-dtfs.atlassian.net/browse/DTFS2-6655 `disabled` endpoint is no longer used. Marked PR as breaking

portal-api/src/v1/roles/validate-user-has-at-least-one-allowed-role.js
abhi-markan: Console.error to use string substitituion.

portal-api/src/v1/roles/validate-user-is-acting-on-self-to-change-password-or-is-admin.js
```diff
@@ -0,0 +1,21 @@
+/**
+ * Creates a middleware that returns a 401 error if the current user logs with
+ * old password have at least one of the allowedRoles, and calls the next handler otherwise.
+ * @param {{ allowedRoles: string[] }}
+ * @returns {(req, res, next) => void}
+ */
+const validateSelfOrAdmin = (req, res, next) => {
+  if (!req.user) {
+    return res.status(401).send('Unauthorized');
```
abhi-markan, line 9: Please use `HttpStatusCode` from `axios`.

abhi-markan (changes requested):

portal-api/src/v1/routes.js
abhi-markan: Has the swagger documentation been updated or does it not exists? Thanks

portal-api/src/v1/roles/validate-user-is-acting-on-self-to-change-password-or-is-admin.js
```diff
@@ -0,0 +1,21 @@
...
+  }
+
+  if (req?.user?._id !== req?.params?._id && !req?.user?.roles?.includes('admin')) {
```
abhi-markan, line 12: ADMIN from constant please.

abhi-markan (changes requested):

portal-api/src/v1/roles/validate-user-is-acting-on-self-to-change-password-or-is-admin.js
vamshidharukef: Used `HttpStatusCode` from `axios`.

portal-api/src/v1/roles/validate-user-is-acting-on-self-to-change-password-or-is-admin.js
vamshidharukef: Imported `constant`.

portal-api/src/v1/routes.js
vamshidharukef: No, I don't think it exists.

portal-api/src/v1/roles/validate-user-is-acting-on-self-to-change-password-or-is-admin.js
```diff
@@ -0,0 +1,23 @@
...
+const CONSTANT = require('../../constants');
+/**
+ * Creates a middleware that returns a 401 error if the current user logs with
+ * old password have at least one of the allowedRoles, and calls the next handler otherwise.
+ * @param {{ allowedRoles: string[] }}
+ * @returns {(req, res, next) => void}
+ */
+const validateSelfOrAdmin = (req, res, next) => {
+  if (!req.user) {
+    return res.status(axios.HttpStatus.UNAUTHORIZED).send('Unauthorized');
```
abhi-markan, line 11: Can this please be `HttpStatusCode.Unauthorized` to keep it consistent.

portal-api/src/v1/roles/validate-user-is-acting-on-self-to-change-password-or-is-admin.js
```diff
@@ -0,0 +1,23 @@
...
+  }
+
+  if (req?.user?._id !== req?.params?._id && !req?.user?.roles?.includes(CONSTANT.ADMIN)) {
+    return res.status(axios.HttpStatus.FORBIDDEN).send('Forbidden');
```
abhi-markan, line 15: Same as above.

portal-api/src/v1/routes.js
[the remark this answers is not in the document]
abhi-markan: Will a more appropriate name be wise, something like `fetchAllAdmin` or does this do more than that?

abhi-markan (changes requested):

portal-api/src/v1/roles/validate-user-is-acting-on-self-to-change-password-or-is-admin.js
vamshidharukef: Changed to `HttpStatusCode.Unauthorized`

portal-api/src/v1/roles/validate-user-is-acting-on-self-to-change-password-or-is-admin.js
vamshidharukef: Changed to `HttpStatusCode.Forbidden`

abhi-markan:
@vamshidharukef still have few failing tests.

e2e-tests/portal/cypress/e2e/journeys/maker/dashboard/deals/dashboard-table-data.spec.js
```diff
@@ -179,16 +177,18 @@ context('View dashboard deals as a maker', () => {
...
       .totalItems()
       .invoke('text')
       .then((text) => {
-        expect(text.trim()).equal(`(${ALL_BANK1_DEALS.length} items)`);
+        expect(text.trim()).equal(`(${ALL_DEALS.length} items)`);
       });
 
-    cy.get('table tr').find(`[data-cy="deal__status--${ALL_BANK2_DEALS[0]._id}"]`).should('not.exist');
+    ALL_BANK2_DEALS.forEach((deal) => {
+      cy.get('table tr').find(`[data-cy="facility__name--link--${deal._id}"]`).should('not.exist');
```
ttbarnes, line 191: Is there a reason for replacing the `deal__status` assertion with `facility__name`?

portal-api/api-tests/api-test-users.js
```diff
@@ -279,6 +279,37 @@ const setUpApiTestUser = async (as) => {
...
+  password: 'NewP@ssword1234',
+  firstname: 'NewAdmin',
+  surname: 'Test User',
+  email: 'new-admin-tester@ukexportfinance.gov.uk',
+  timezone: 'Europe/London',
+  roles: [ADMIN],
+  bank: banks.any,
+};
+
+const setUpNewAdminTestUser = async (as) => {
```
ttbarnes, line 293: Should this be called `setUpAdminTestUser`? Since the other function is called `setUpApiTestUser`, rather than `setUpNewApiTestUser`

e2e-tests/portal/cypress/e2e/journeys/maker/dashboard/deals/dashboard-table-data.spec.js
vamshidharukef: There are no deals displayed when searching for deals so the assertion fails hence replaced with facility_name.

portal-api/api-tests/api-test-users.js
vamshidharukef: This is already called as setUpAdminTestUser. Not sure if I misunderstood.

e2e-tests/gef/cypress/e2e/application/application-activities-AIN.spec.js
```diff
@@ -70,7 +70,7 @@ context('Submit AIN deal and check portalActivities', () => {
   describe('check portalActivity Page', () => {
     beforeEach(() => {
       cy.saveSession();
-      cy.login(BANK1_MAKER1);
+      cy.login(ADMIN);
```
abhi-markan, line 73: This should be `BANK1_MAKER1` please.

e2e-tests/portal/cypress/e2e/journeys/admin/dashboard/dashboard-table-data.spec.js
```diff
@@ -5,7 +5,7 @@ const relative = require('../../../relativeURL');
 
 const { GEF_DEAL_DRAFT, GEF_FACILITY_CASH, GEF_FACILITY_CONTINGENT } = require('../fixtures');
 
-const { BANK1_MAKER1, ADMIN } = MOCK_USERS;
+const { ADMIN } = MOCK_USERS;
```
abhi-markan, line 8: This cannot be `ADMIN` either please.

portal-api/api-tests/api-test-users.js
abhi-markan: setUpAdminTestUser please.

portal-api/src/v1/roles/validate-user-is-acting-on-self-to-change-password-or-is-admin.js
```diff
@@ -0,0 +1,22 @@
...
+ * Creates a middleware that returns a 401 error if the current user logs as admin or other user, and calls the next handler otherwise.
+ * @param {{ allowedRoles: string[] }}
+ * @returns {(req, res, next) => void}
+ */
+const validateUserIsActingOnSelfToChangePasswordOrIsAdmin = (req, res, next) => {
+  if (!req.user) {
+    return res.status(HttpStatusCode.Unauthorized).send('Unauthorized');
+  }
+
+  if (req?.user?._id !== req?.params?._id && !req?.user?.roles?.includes(CONSTANT.ADMIN) && !req?.user?.roles?.includes(CONSTANT.MAKER)) {
```
abhi-markan, line 13: Simplify the compound condition.

abhi-markan: Will this prevent current logged in `maker` and `admin` from changing their own password?

abhi-markan (changes requested):

portal-api/src/v1/roles/validate-user-has-at-least-one-allowed-role.js
vamshidharukef: Added string substitituion to console.error

portal-api/src/v1/routes.js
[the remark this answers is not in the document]
vamshidharukef: renamed to validateUserPermission

portal-api/api-tests/api-test-users.js
vamshidharukef: Changed to setUpAdminTestUser.

e2e-tests/portal/cypress/e2e/journeys/admin/dashboard/dashboard-table-data.spec.js
vamshidharukef: Added back `BANK1_MAKER1`

portal-api/src/v1/roles/validate-user-is-acting-on-self-to-change-password-or-is-admin.js
vamshidharukef: Simplified the compound condition. Yes, `maker` and `admin` can change their own password.

e2e-tests/portal/cypress/e2e/journeys/admin/activate-deactivate-user/activate-deactivate-user.spec.js
```diff
@@ -12,7 +12,7 @@ context('Admin user updates an existing user', () => {
     firstname: 'first',
     surname: 'last',
     bank: 'Barclays Bank',
-    roles: ['maker'],
+    roles: ['admin', 'maker'],
```
ttbarnes, line 15: Should this be reverted? I'm not familiar with how `userToUpdate` is being used in this test exactly, but it reads like it's defining the user that an admin user will update 

Therefore should the test not be that a pure admin with only admin roles, updates a user that does _not_ have admin roles?

At the moment it looks like an admin user is updating another user that also has an admin role?

e2e-tests/portal/cypress/e2e/journeys/checker/dashboard/dashboard.spec.js
```diff
@@ -60,13 +60,13 @@ context('View dashboard deals as a checker', () => {
...
 
-    cy.insertOneGefApplication(GEF_DEALS.READY_FOR_CHECK, BANK1_MAKER1).then((gefDeal) => {
+    cy.insertOneGefApplication(GEF_DEALS.READY_FOR_CHECK, ADMIN).then((gefDeal) => {
       cy.setGefApplicationStatus(gefDeal._id, GEF_DEALS.READY_FOR_CHECK.status, BANK1_MAKER1).then((updatedGefDeal) => {
         BANK1_DEALS.push(updatedGefDeal.body);
       });
     });
 
-    cy.insertOneDeal(BSS_DEALS.READY_FOR_CHECK, BANK2_MAKER2);
+    cy.insertOneDeal(BSS_DEALS.READY_FOR_CHECK, BANK1_MAKER2);
```
ttbarnes, line 69: What's the reason for changing this from `BANK2_MAKER2` to `BANK1_MAKER2`?

e2e-tests/portal/cypress/e2e/journeys/checker/dashboard/dashboard.spec.js
```diff
@@ -60,13 +60,13 @@ context('View dashboard deals as a checker', () => {
       BANK1_DEALS.push(createdDeal);
     });
 
-    cy.insertOneGefApplication(GEF_DEALS.READY_FOR_CHECK, BANK1_MAKER1).then((gefDeal) => {
+    cy.insertOneGefApplication(GEF_DEALS.READY_FOR_CHECK, ADMIN).then((gefDeal) => {
```
ttbarnes, line 63: Seems that this should be `BANK2_MAKER2` instead of `ADMIN`? I don't see why `ADMIN` is required here, unrelated to the test description, _"View dashboard deals as a checker"_?

e2e-tests/portal/cypress/e2e/journeys/admin/activate-deactivate-user/activate-deactivate-user.spec.js
vamshidharukef: Removed `admin`' from roles.

e2e-tests/portal/cypress/e2e/journeys/checker/dashboard/dashboard.spec.js
vamshidharukef: Changed back to `BANK2_MAKER2`

e2e-tests/portal/cypress/e2e/journeys/checker/dashboard/dashboard.spec.js
```diff
@@ -84,23 +84,20 @@ context('View dashboard deals as a checker', () => {
 
     const { exporter, bankRef, product, status, type, updated, link } = dashboardDeals.row;
 
-    // should only see 2 deals - ready for check and in Bank 1
-    const EXPECTED_DEALS = BANK1_DEALS.filter((deal) => deal.status === CONSTANTS.DEALS.DEAL_STATUS.READY_FOR_APPROVAL);
```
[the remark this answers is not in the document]
vamshidharukef: Reverted back

e2e-tests/portal/cypress/e2e/journeys/checker/dashboard/dashboard.spec.js
vamshidharukef: With `BANK1_MAKER1` the test is unable to create deal so changed to `ADMIN` but when testing it logs as `BANK1_MAKER1`.
  
Capture

ttbarnes: What does `signInLinkResponse` return? Bit odd that `insertGefApplication`  (that calls the `/v1/gef/application`) endpoint is not working with `BANK1_MAKER1`.

I can't see how this could happen with these changes, as in the API only the `/user`routes have changed, unless i'm missing something

e2e-tests/portal/cypress/e2e/journeys/maker/dashboard/deals/dashboard-table-data.spec.js
```diff
@@ -179,16 +177,18 @@ context('View dashboard deals as a maker', () => {
 
   // TODO: DTFS2-5372 - fix.
   it('should not show deals created by other banks', () => {
-    cy.login(BANK1_MAKER1);
+    cy.login(ADMIN);
```
ttbarnes, line 180: I think this should be `BANK1_MAKER1` as it's not an admin dashboard test?

e2e-tests/portal/cypress/e2e/journeys/maker/dashboard/deals/dashboard-table-data.spec.js
ttbarnes: But the test here is going to dashboard _deals_? `dashboardDeals.visit();` 

The test says: 

`it('should not show deals created by other banks', () => { ...`

But this change is testing that a facility with a particular deal ID does not exist, whereas before it tested that a deal (that belongs to another bank) does not exist?

portal-api/api-tests/v1/users/put-update-user.api-test.js
```diff
@@ -182,10 +178,15 @@ describe('a user', () => {
...
 
-        expect(status).toEqual(200);
-        expect(body.roles).toEqual(MOCK_USER.roles);
+        expect(status).toEqual(401);
+        // Check if body is not empty before accessing roles
+        if (Object.keys(body).length !== 0) {
+          expect(body.roles).toEqual(MOCK_USER.roles);
+        } else {
+          console.error('No user data returned in body');
+        }
```
ttbarnes, line 189: Should avoid conditionals and logs in tests as much as possible - could it just be the following instead?

```js
expect(body.roles).toEqual(MOCK_USER.roles);
```

portal-api/src/v1/roles/validate-user-is-acting-on-self-to-change-password-or-is-admin.js
```diff
@@ -0,0 +1,26 @@
...
+  if (!isActingOnSelf && !isAdmin && !isMaker) {
+    return res.status(HttpStatusCode.Forbidden).send('Forbidden');
+  }
+
+  return next();
+};
+
+module.exports = {
+  validateUserPermission,
+};
```
ttbarnes, line 26: Do we have some unit tests for this?

Can we remove `to-change-password` from the file name as the middleware doesn't have any mentions of password functionality.

Afterwards would be good to rename the middleware const (`validateUserPermission`) to align with the file name, like the existing `userHasAtLeastOneAllowedRole` middleware.

e2e-tests/portal/cypress/e2e/journeys/maker/dashboard/deals/dashboard-table-data.spec.js
vamshidharukef: With `BANK1_MAKER1` user assertion fails as number of the deals returned are 3 instead of 4.

Capture

portal-api/api-tests/v1/users/put-update-user.api-test.js
vamshidharukef: Removed conditionals

portal-api/src/v1/roles/validate-user-is-acting-on-self-to-change-password-or-is-admin.js
vamshidharukef: Added unit test and renamed middleware to` validate-user-is-acting-on-self-or-is-admin.js`. Also renamed `const { validateUserPermission } = require('./roles/validate-user-is-acting-on-self-or-is-admin');`

e2e-tests/gef/cypress/support/commands/loadData.js
```diff
@@ -5,8 +5,6 @@
  */
 
 const loadData = () => {
-  // Ensure project root directory
```
abhi-markan, line 8: This shouldn't be removed.

e2e-tests/portal/cypress/e2e/journeys/admin/dashboard/dashboard-table-data.spec.js
```diff
@@ -27,7 +27,7 @@ context('Admin dashboard', () => {
       deal = insertedDeal;
     });
 
-    cy.insertOneGefApplication(GEF_DEAL_DRAFT, BANK1_MAKER1).then((dealGef) => {
```
abhi-markan, line 30: Please ensure applications are inserted as `BANK1_MAKER1` as an `ADMIN` will never create an application in a real scenario.

abhi-markan (changes requested):

e2e-tests/portal/cypress/e2e/journeys/checker/dashboard/dashboard.spec.js
vamshidharukef: `signInLinkResponse` returns user details as `maker1@ukexportfinance.gov.uk` with `maker` role. With `maker` role user is trying to access end point `http://localhost:5001/v1/users` but only `admin` have access to this so the response received is `401: Unauthorised You do not have access to this page`. Hence the test fails.
Capture

ttbarnes (commented):
I've been thinking about this. Quick summary:

- API protection has been put in place so a non-admin user cannot call the API directly, basically.
- `cy.insertOneGefApplication` creates an application directly in the API.
- Therefore, if the API call is made with any user that is not admin, it will be rejected.
- With this in mind, it makes sense that the command has been updated to use `ADMIN`.
- However, this is not good for E2E tests because it is not a true user journey.

I can think of 3x approaches:

1) Update the API to allow Makers to create deals via the API if the environment is a dev/test environment.

2) Refactor the E2E tests to go through a true user journey in the UI, instead of calling the API. Quick/bare minimum example:

```
cy.createGefApplicationReadyForCheck({});

// instead of `cy.insertOneGefApplication()` etc.
```

3) Temporarily go with the current `ADMIN` approach until we have no.2 in place.

No options are quick/easy or ideal in the interests of getting this PR merged.

No.2 would be the cleanest solution, but this would probably be a lot of work and will increase the time it takes to run E2E tests.

What do you think @abhi-markan ?

abhi-markan:
> * been put in place s

Hey @toddym42 thanks for the explanation I would prefer option 2 as it represents true E2E user journey as any other option is a work around and will add to our tech-debt. @vamshidharukef  are you okay to update the E2E tests and work closely with @ttbarnes on this please. Appreciate this is a long task.

ttbarnes:
Hey @vamshidharukef @abhi-markan - just thinking about the E2E changes @vamshidharukef is working on and this PR

I think we should create a fresh PR from `main`, purely for the E2E changes - they're going to be a big changes.

Otherwise, this PR, which is about some API auth changes, will become huge and have both fundamental E2E changes and API auth changes. I suggest doing the following instead:


1. 1x new PR to update E2E tests to create individual GEF applications through the UI, instead of API.
2. 1x new PR to update E2E tests to create individual BSS/EWCS applications through the UI, instead of API.
3. 1x new PR to update Portal Dashboard E2E tests (for multiple applications, after no.1 and 2). It will probably be wise to create a specific test environment endpoint for such dashboard/pagination tests, but this another topic we should discuss.
4. Once the above is merged, this API auth changes PR can become much simpler with no need to modify the E2E tests, should become a pure API PR 😎 

Any thoughts please?

ttbarnes:
I suggest we close this PR (open for ~5-6 months) and open a new one after some E2E changes are made in other PRs.

It will be easier than fixing the conflicts. Also, the E2E test changes in this PR shouldn't be necessary (other PRs will resolve this).

There will only be more conflicts as more PRs are merged and this current PR will be open for probably at least another 2x weeks.

ttbarnes:
Closing this as per comment above. Once DTFS-7431 and DTFS2-7434 are complete, we can create a new simple PR with pure API changes.

PR #3050, opened 2024-04-24
feat(FN-1615): add seeder for fee records

JoshBinns2000:
## Introduction :pencil2:
We need a way of seeding fee record data to view parts of TFM 6.

Adds a new seeder for a fee record which:
- has matching currencies
- has differening currencies*

*This will have an affect on the `Reported fees` and `Reported payments` columns of the TFM premium payments table.


utils/sql-db-seeder/src/utilisation-report/utilisation-report.seed.ts
```diff
@@ -27,6 +27,9 @@ export default class UtilisationReportSeeder implements Seeder {
     const uploadedReportReportPeriod = getCurrentReportPeriodForBankSchedule(paymentReportOfficerBank.utilisationReportPeriodSchedule);
     const uploadedReport = createUploadedReport(paymentReportOfficer, uploadedReportReportPeriod, 'PENDING_RECONCILIATION');
 
+    // The reports need to be seeded either before or with the fee records
+    appendFeeRecordsToUtilisationReport(uploadedReport);
```
AlexBramhill, line 31: Not a comment for a change, just a general comment re side effects :) was there a reason we did it this way (with side effects)?

I tend to dislike functions with side effects, even if we aren't strictly functional programming, as my default assumption is we don't have side effects to functions. I had to do a check here to make sure that we were modifying this object -- although that's probably just me being used to patterns without side effects ie:
`const uploadedReportWithFeeRecords = {...uploadedReport, feeRecords: createFeeRecordsForUtilisationReport(uploadedReport)}`
(or even, just combining this inside `createUploadReport`)

JoshBinns2000: Because this is a class, we can't just do `{ ...uploadedReport, feeRecords: foo }` as it will lose the methods that exist on the prototype. The way to do it without side effects would be to do
```typescript
const uploadedReportFeeRecords = createFeeRecords({ ... });
uploadedReport.feeRecords = uploadedReportFeeRecords;
```
which I am happy to do

utils/sql-db-seeder/src/utilisation-report/utilisation-report.seed.ts
```diff
@@ -27,6 +27,10 @@ export default class UtilisationReportSeeder implements Seeder {
     const uploadedReportReportPeriod = getCurrentReportPeriodForBankSchedule(paymentReportOfficerBank.utilisationReportPeriodSchedule);
     const uploadedReport = createUploadedReport(paymentReportOfficer, uploadedReportReportPeriod, 'PENDING_RECONCILIATION');
 
+    // The reports need to be seeded either before or with the fee records
+    const { feeRecordWithMatchingCurrencies, feeRecordWithDifferingCurrencies } = createFeeRecordsForReport();
+    uploadedReport.feeRecords = [feeRecordWithMatchingCurrencies, feeRecordWithDifferingCurrencies];
```
AlexBramhill, line 32: `uploadedReport.updateWithFeeRecords`?

AlexBramhill, line 32: `uploadedReport.updateWithFeeRecords(feeRecords: createFeeRecordsForReport())` maybe?

JoshBinns2000: In actual production code we almost never interact with properties directly and use more of this object-oriented style. I didn't mind as much here because it is just for seeding, but I suppose for consistency's sake we should use that method

AlexBramhill (approved):
i realised i commented and have reviewed it all :) happy for you to wait for someone else on your team to review it, but this all is good work

utils/sql-db-seeder/src/utilisation-report/utilisation-report.helper.ts
```diff
@@ -99,3 +101,28 @@ export const createUploadedReport = (
 
   return utilisationReport;
 };
+
+export const createFeeRecordsForReport = (): FeeRecordEntity[] => {
+  const feeRecordWithMatchingCurrencies = createFeeRecord({
+    facilityId: '12345678',
```
RLavender98, line 107: This might be a bit complicated, but have we thought about matching these to actual facilities so that when we get to adding the keying sheet we can fetch data about the facility?

JoshBinns2000: If I match it to an actual facility, I would probably then need to worry about properly mapping the data as well which I'd rather not do at this stage. For now, I just want some simple data in there so that I can actually view the page and the table, and we can update the seeder as we advance the keying sheet functionality

RLavender98: Ok we can leave to when we add keying sheets, just worth being aware we are likely to make changes to this when we are writing tests since they run the seeding job

ChristianMcc (commented):
everyone beat me to it so im submitting this single tiny comment instead of finishing it

ChristianMcc (approved):

PR #3056, opened 2024-04-26
feat(FN-2734): add batch save to migrator

JoshBinns2000:
## Introduction :pencil2:
If there are too many fee records, trying to save everything all at once causes an error (see #3043)

Updates the migrator to save reports first and then batch save fee records afterwards.


ChristianMcc (approved):
LGTM but can you check linting/formatting on this against the codacy?

ChristianMcc (changes requested):
Just need a lint fix

ChristianMcc (dismissed):

ChristianMcc (approved):

PR #3058, opened 2024-04-29
feat(dtfs2-7069): update dtfs helpdesk hyperlink

AlexBramhill:
## Introduction :pencil2:
update email to use new helpdesk hyperlink, update accessibility to use env var for email address

use env vars to determine whether to display helpdesk hyperlink


trade-finance-manager-ui/templates/_partials/footer.njk
```diff
@@ -29,4 +33,4 @@
       </div>
     </div>
   </div>
-</footer>
+</footer>
```
MarRobSoftwire, line 36: Should the new line have been deleted?

trade-finance-manager-ui/templates/_partials/footer.njk
```diff
@@ -4,14 +4,18 @@
       <div class="govuk-footer__meta">
         <div class="govuk-footer__meta-item govuk-footer__meta-item--grow">
           <h2 data-cy="contact-us-footer">Contact us</h2>
-
           <dl class="contact-details">
-            <dt>Email:</dt>
-            <dd>DigitalService.TradeFinance@ukexportfinance.gov.uk</dd>
+            {% if CONTACT_US_SELF_SERVICE_PORTAL_URL %}
```
MarRobSoftwire, line 8: Is the plan to phase this out eventually? Should there be a ticket & `TODO` for this?

MarRobSoftwire (commented):
Couple minor Qs but looks good

trade-finance-manager-ui/templates/_partials/footer.njk
AlexBramhill: So...it could technically be phased out later. I thought about it and maybe think it's better to keep in because:
- We're going to have different env vars for non dev and prod
- Therefore we are always going to have an env var for this
- Therefore there is a chance this isn't present
- Therefore we should allow a fallback to email address if not present
- As part of another ticket, we've updated the email addresses, so I've added in the same pattern here (with fallback, just in case!)

Does that make sense to you? Or should we make a ticket for it to scream if this env var is missing?

AlexBramhill: I've added a ticket for this and for a review of whether we want to remove this and added comments

MarRobSoftwire: Yeah think it feels weird to fail silently but makes sense to worry about that once Prod has the relevant env updated

MarRobSoftwire (dismissed):

trade-finance-manager-ui/server/nunjucks-configuration/index.js
```diff
@@ -24,6 +29,13 @@ const configureNunjucks = (opts) => {
 
   const nunjucksEnvironment = nunjucks.configure(appViews, opts);
 
+  // TODO: DTFS2-7138 remove contact us env vars
+  nunjucksEnvironment.addGlobal('CONTACT_US_SELF_SERVICE_PORTAL_URL', CONTACT_US_SELF_SERVICE_PORTAL_URL);
+  nunjucksEnvironment.addGlobal(
+    'CONTACT_US_EMAIL_ADDRESS',
+    CONTACT_US_EMAIL_ADDRESS || 'DigitalService.TradeFinance@ukexportfinance.gov.uk',
```
abhi-markan, line 36: Can we remove `'DigitalService.TradeFinance@ukexportfinance.gov.uk'` please?

abhi-markan (changes requested):

trade-finance-manager-ui/server/nunjucks-configuration/index.js
AlexBramhill: sorry other comment got lost -- we should keep this tohave a fall back in case env vars are missing. we have a ticket to remove this once we've migrated over

abhi-markan (dismissed):

abhi-markan (approved):

PR #3062, opened 2024-05-01
chore(FN): commonise and fix sql db seeder

JoshBinns2000:
## Introduction :pencil2:
We want to commonise the `sql-db-seeder` into the `utils` package.

Commonising causes the `typeorm-extension` node modules to be installed at the project root and not in `utils/sql-db-seeder`, likely due to version with other packages conflicts. As `typeorm-extension` requires you to directly run the `cli.cjs` script from the node modules if you do not build the typescript files, we need to somehow get access to the node modules in the `utils/sql-db-seeder` directory.

This PR achieves this by adding a `"predb:seed"` script to `utils/package.json` which removes `utils/sql-db-seeder/node_modules` and then copies the root `node_modules/typeorm-extension` into `utils/sql-db-seeder`. The rest of the commands then work as they previously did.


ChristianMcc (approved):

utils/package.json
```diff
@@ -1,13 +1,15 @@
 {
-    "name": "utilities-mock-data-loader",
+    "name": "dtfs-utilities",
     "version": "1.0.0",
     "description": "DTFS Utilities",
     "license": "MIT",
     "author": "UK Export Finance",
     "scripts": {
+        "predb:seed": "rm -rf sql-db-seeder/node_modules && (cd sql-db-seeder && mkdir node_modules && cp -r ../../node_modules/typeorm-extension node_modules)",
+        "db:seed": "npx ts-node sql-db-seeder/node_modules/typeorm-extension/bin/cli.cjs seed:run -r sql-db-seeder -d src/seeding-data-source.ts",
```
RLavender98, line 9: Please can we update the docs to mention you need to run `predb:seed` and maybe also a short explanation of this weirdness?

PR #3064, opened 2024-05-01
feat(FN-2420): ukef utilisation report email notifications to multiple addresses

RLavender98:
## Introduction :pencil2:
Add support for sending utilisation report upload emails to multiple email addresses.

- Use a list of emails in the env var for UKEF report email notification recipients
- Send utilisation report uploaded notifications to all emails in the list

- Refactor the upload controller to extract some of the logic into separate services
- Add some unit tests for the controller and created services
- Don't fail upload request if sending notification email to UKEF fails


portal-api/src/v1/services/utilisation-report/email-service.ts
```diff
@@ -0,0 +1,82 @@
...
+import { EMAIL_TEMPLATE_IDS } from '../../../constants';
+import { formatDateForEmail } from '../../helpers/formatDateForEmail';
+import api from '../../api';
+import { InvalidEnvironmentVariableError } from '../../errors';
+
+const getUkefGefReportingEmailRecipients = () => {
+  const EmailsSchema = array(string());
+
+  try {
+    const emails =  EmailsSchema.parse(JSON.parse(asString(process.env.UKEF_GEF_REPORTING_EMAIL_RECIPIENT, 'UKEF_GEF_REPORTING_EMAIL_RECIPIENT')));
```
RLavender98, line 13: This requires emails to be provided as: 
`UKEF_GEF_REPORTING_EMAIL_RECIPIENT=["email1", "email2"]`
The env var is just read as a string so we can change this to a different form if we could like, just anything we can then parse and turn into an array

portal-api/src/v1/controllers/utilisation-report-service/utilisation-report-upload.controller.js
```diff
@@ -1,161 +1,52 @@
...
-    if (existingReport.status !== UTILISATION_REPORT_RECONCILIATION_STATUS.REPORT_NOT_RECEIVED) {
-      return res
-        .status(500)
-        .send(
-          `Expected report to be in '${UTILISATION_REPORT_RECONCILIATION_STATUS.REPORT_NOT_RECEIVED}' state (was actually in '${existingReport.status}' state)`,
-        );
+    try {
+      await sendUtilisationReportUploadNotificationEmailToUkefGefReportingTeam(parsedUser?.bank?.name, formattedReportPeriod);
+    } catch(error) {
+      console.error('Failed to send report upload notification to ukef gef reporting team %o', error);
```
RLavender98, line 31: This catching here is a change to functionality, it just felt weird to be showing the user an error page when this emails fails to send given they are not told it was sending it in the first place and otherwise the upload went fine. But if we don't want this I can change this back. I thought about changing the other email to also fail quietly from the user's perspective but left it since we display the email addresses on the page so that would have required more thinking.

portal-api/src/v1/services/utilisation-report/azure-file-service.js
RLavender98, line 1: I wasn't too sure on the best place to put this and the email-service and indeed their names. Settled for a services folder, with a `utilisation-reports` folder then `-service` file names. Keen to know if there is a better pattern I could use.

portal-api/src/v1/services/utilisation-report/email-service.ts
```diff
@@ -0,0 +1,82 @@
...
+        bankName,
+        reportPeriod,
+      }),
+    ),
+  );
+};
+
+/**
+ * Calls the DTFS Central API to get bank details by bank ID and
+ * returns only the payment officer team
```
JoshBinns2000, line 39: Missing `@param`

portal-api/src/v1/services/utilisation-report/email-service.ts
```diff
@@ -0,0 +1,82 @@
...
+    return bank.paymentOfficerTeam;
+  } catch (error) {
+    console.error('Unable to get bank payment officer team details by ID %o', error);
+    throw error;
+  }
+};
+
+/**
+ * Sends notification email to bank payment officer team that a utilisation report has been
+ * received and return the payment officer team email address.
```
JoshBinns2000, line 53: Missing `@param`s

portal-api/src/v1/validation/utilisation-report/utilisation-report-upload-validator.ts
```diff
@@ -0,0 +1,21 @@
+import { ReportPeriod, UTILISATION_REPORT_RECONCILIATION_STATUS, getFormattedReportPeriodWithLongMonth } from '@ukef/dtfs2-common';
+import { InvalidReportStatusError } from '../../errors';
+import api from '../../api';
+
+export const validateReportForPeriodIsInReportNotReceivedStateAndReturnId = async (bankId: string, reportPeriod: ReportPeriod): Promise<number> => {
```
JoshBinns2000, line 5: This feels to me like it is doing too much for one function 🤔

portal-api/src/v1/services/utilisation-report/azure-file-service.js
JoshBinns2000: This looks fine to me

portal-api/src/v1/controllers/utilisation-report-service/utilisation-report-upload.controller.js
JoshBinns2000: Personally, I think the failing silently is something we want to add to the function being called rather than in the controller (a try-catch inside the try-catch feels a bit strange to read). My preferred method would be to have two methods exported from the emails service - one which can throw and has the name `sendUtilisation...OrThrow` and one which doesn't, but I'm not sure if that's the best solution here

JoshBinns2000 (changes requested):
I think there are some places where prettier needs to be run, but otherwise looks pretty good

portal-api/src/v1/errors/invalid-report-status.error.ts
```diff
@@ -0,0 +1,6 @@
+export class InvalidReportStatusError extends Error {
```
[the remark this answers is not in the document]
RLavender98: In the interest of this PR not taking on too many refactors I'm tempted to push against this for now

portal-api/src/v1/services/utilisation-report/email-service.ts
```diff
@@ -0,0 +1,82 @@
...
+  } catch (error) {
+    console.error('Failed to parse UKEF_GEF_REPORTING_EMAIL_RECIPIENTS ', error);
+    throw new InvalidEnvironmentVariableError('Failed to parse UKEF_GEF_REPORTING_EMAIL_RECIPIENTS');
+  }
+};
+
+/**
+ * Sends notification email to UKEF GEF reporting email recipients that a utilisation report has been submitted
+ * @param {string} bankName - name of the bank
+ * @param {string} reportPeriod - period for which the report covers as a string, eg. June 2023
```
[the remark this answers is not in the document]
RLavender98: That was pre-existing, we get the formatted report period in the request from portal so that has been used I assume for simplicity over reformatting the report period object?

portal-api/src/v1/validation/utilisation-report/utilisation-report-upload-validator.ts
RLavender98: Yeah i struggled with this one a bit, I think I should perhaps move most of it back to the controller

portal-api/src/v1/controllers/utilisation-report-service/utilisation-report-upload.controller.js
RLavender98: Originally I had it in the downstream method but the issue I felt was then it doesn't match the behaviour of the other send email method, and also both send multiple emails and could throw sending one and not the others as well so 'OrThrow' seems a little misleading in and of itself if I add it to the other method...

portal-api/src/v1/api.js
```diff
@@ -346,11 +346,14 @@ const getUtilisationReportById = async (id) => {
...
+/**
+ * Call the central API to get a bank
+ * @returns {Promise<import('./api-response-types').BankResponse>} response of API call
+ */
 const getBankById = async (bankId) => {
   try {
     if (!isValidBankId(bankId)) {
-      console.error('Get bank failed with the following bank ID %s', bankId);
-      return false;
+      throw new Error(`Invalid bank id: ${bankId}`);
```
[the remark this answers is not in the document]
JoshBinns2000: It looks like there may now be some indenting which is slightly off after removing the try-catch :)

portal-api/src/v1/controllers/utilisation-report-service/utilisation-report-upload.controller.js
```diff
@@ -1,161 +1,80 @@
...
+const { InvalidReportStatusError } = require('../../errors');
 
 /**
- * Calls the DTFS Central API to get bank details by bank ID and
- * returns only the payment officer team name and email
- * @param {string} bankId - the bank ID
- * @returns {Promise} payment officer team name and email
+ * Fetches report for bank for given report period
+ * @param {string} bankId
+ * @param {ReportPeriod} reportPeriod
```
JoshBinns2000, line 16: I think this may need to be `{import('@ukef/dtfs2-common').ReportPeriod}`?

portal-api/src/v1/controllers/utilisation-report-service/utilisation-report-upload.controller.js
```diff
@@ -1,161 +1,80 @@
...
-    const existingReport = existingReports[0];
-
-    if (existingReport.status !== UTILISATION_REPORT_RECONCILIATION_STATUS.REPORT_NOT_RECEIVED) {
-      return res
-        .status(500)
-        .send(
-          `Expected report to be in '${UTILISATION_REPORT_RECONCILIATION_STATUS.REPORT_NOT_RECEIVED}' state (was actually in '${existingReport.status}' state)`,
-        );
+    const report = await getReportForPeriod(bankId, parsedReportPeriod);
+    validateReportIsInReportNotReceivedState(report);
```
JoshBinns2000, line 49: I think this is better than the function we had before - nice change!

portal-api/src/v1/controllers/utilisation-report-service/utilisation-report-upload.controller.js
```diff
@@ -1,161 +1,80 @@
...
+    const azureFileInfo = await saveUtilisationReportFileToAzure(file, bankId);
+    const { dateUploaded } = await api.saveUtilisationReport(report.reportId, parsedReportData, parsedUser, azureFileInfo);
+
+    try {
+      await sendUtilisationReportUploadNotificationEmailToUkefGefReportingTeam(
+        parsedUser?.bank?.name,
+        formattedReportPeriod,
+      );
+    } catch (error) {
+      console.error('Failed to send report upload notification to ukef gef reporting team %o', error);
```
JoshBinns2000, line 59: Do you think it is worthwhile for us to add a comment above this try-catch to explain why we don't want everything to fail if this throws?

portal-api/src/v1/controllers/utilisation-report-service/utilisation-report-upload.controller.js
```diff
@@ -1,161 +1,80 @@
...
-      parsedUser,
+      new Date(dateUploaded),
+      parsedUser.firstname,
+      parsedUser.surname,
     );
     return res.status(201).send({ paymentOfficerEmails });
   } catch (error) {
-    console.error('Failed to save utilisation report %o', error);
+    if (error instanceof InvalidReportStatusError) {
+      return res.status(500).send(error.message);
```
JoshBinns2000, line 71: Could we update `InvalidReportStatusError` to have its own error code rather than hard-coding the `500` here?

portal-api/src/v1/services/utilisation-report/email-service.test.ts
```diff
@@ -0,0 +1,106 @@
...
+    });
+
+    it('should throw an error if email recipients not provided in correct format', async () => {
+      // Arrange
+      process.env.UKEF_GEF_REPORTING_EMAIL_RECIPIENT = 'email1@ukexportfinance.gov.uk,email2@ukexportfinance.gov.uk';
+
+      // Act + Assert
+      await expect(
+        sendUtilisationReportUploadNotificationEmailToUkefGefReportingTeam('My Bank', 'June 2026'),
+      ).rejects.toThrow(Error);
```
JoshBinns2000, line 52: Do you think there is any benefit here to asserting that the error thrown is exactly what we expect rather than just checking that `Error` is thrown? I only ever include the `Error` constructor if I actually want to assert against the body of the error, like
```typescript
await expect(...).rejects.toThrow(new Error('my error message');
```
Otherwise, I tend to just use
```typescript
await expect(...).rejects.toThrow();
```
which I think does the same as the assertion we have here. Is there a benefit I'm missing with providing the plain `Error` object as an argument to `toThrow`?

portal-api/src/v1/validation/utilisation-report/utilisation-report-validator.test.ts
```diff
@@ -0,0 +1,29 @@
...
+      Object.values(UTILISATION_REPORT_RECONCILIATION_STATUS).filter((status) => status !== 'REPORT_NOT_RECEIVED'),
+    )('throws InvalidReportStatusError if report has status %s', (status: UtilisationReportReconciliationStatus) => {
+      // Arrange
+      const report = { ...aUtilisationReportResponse(), status };
+
+      // Act + Assert
+      expect(() => validateReportIsInReportNotReceivedState(report)).toThrow(InvalidReportStatusError);
+    });
+
+    it('returns report id if report has REPORT_NOT_RECEIVED status', () => {
```
JoshBinns2000, line 21: Should this be `'does not throw if the report has REPORT_NOT_RECEIVED status'`?

portal-api/src/v1/errors/invalid-report-status.error.ts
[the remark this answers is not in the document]
JoshBinns2000: That's fair. I left a comment about this elsewhere, but I think we should still include the specific http error code in this class as if it were using the `ApiError` class so that the controller doesn't need to hard code the `500` itself

portal-api/src/v1/services/utilisation-report/email-service.ts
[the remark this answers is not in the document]
JoshBinns2000: That makes sense to me

portal-api/src/v1/controllers/utilisation-report-service/utilisation-report-upload.controller.js
RLavender98: I guess I didn't add one because to me it feels fairly clear what the intention is and I am quite used to this being the pattern around downstream secondary actions. It is covered by a test case that you would have to intentionally remove to change this. That said, if this looks confusing to you as a reader then I will add a comment as that would be the only reason for a comment and as the author it is perhaps not down to me to solely judge that.

JoshBinns2000: It doesn't look confusing to me - I just wanted to know whether you thought it would be worth including a comment or not. More than happy to not include one as I do think it is quite clear as is :)

portal-api/src/v1/controllers/utilisation-report-service/utilisation-report-upload.controller.js
```diff
@@ -1,161 +1,80 @@
...
-
-/**
- * Sends notification email to PDC Inputters that a utilisation report has been submitted
- * @param {string} bankName - name of the bank
- * @param {string} reportPeriod - period for which the report covers as a string, eg. June 2023
- */
-const sendEmailToPdcInputtersEmail = async (bankName, reportPeriod) => {
-  await sendEmail(EMAIL_TEMPLATE_IDS.UTILISATION_REPORT_NOTIFICATION, PDC_INPUTTERS_EMAIL_RECIPIENT, {
-    bankName,
+const getReportForPeriod = async (bankId, reportPeriod) => {
```
JoshBinns2000, line 18: Could this be `getReportByBankIdAndReportPeriod`?

portal-api/src/v1/controllers/utilisation-report-service/utilisation-report-upload.controller.js
```diff
@@ -1,161 +1,80 @@
...
-
-    console.info(`Successfully saved utilisation report to Azure for bank: ${bankId}`);
-    return fileInfo;
-  } catch (error) {
-    console.error('Failed to save utilisation report to Azure ', error);
-    throw error;
-  }
-};
+  return reportsForPeriod[0];
+}
```
JoshBinns2000, line 32: Missing semi-colon I think (now that the prettier PR has been merged, it should be possible to run `npm run prettier:fix` to fix any and all of these formatting things)

portal-api/src/v1/services/utilisation-report/email-service.test.ts
```diff
@@ -0,0 +1,115 @@
+import { BankResponse } from 'src/v1/api-response-types';
```
JoshBinns2000, line 1: This should import from `'../../api-response-types'` I think

JoshBinns2000 (commented):
Only a couple of small comments - happy to approve once they're resolved and main is merged in

JoshBinns2000 (dismissed):
LGTM

JoshBinns2000 (dismissed):

JoshBinns2000 (approved):

PR #3091, opened 2024-05-02
feat(FN-1488): add find reports by year page

RLavender98:
## Introduction :pencil2:
Add find reports by year page to TFM

- Add a link to the find reports by year page to the banks reports screen
- Add new page "Find report by year" where the user can select a bank and a year to search for reports

Screenshot 2024-05-03 at 10 42 00

Screenshot 2024-05-03 at 10 42 18


trade-finance-manager-ui/server/controllers/utilisation-reports/find-reports-by-year/search-input-validator.test.ts
```diff
@@ -0,0 +1,126 @@
...
+describe('controllers/utilisation-reports/find-reports-by-year/search-input-validator', () => {
+  const BANK_ID_ONE = '1';
+  const BANK_ID_TWO = '2';
+  const BANK_ID_THREE = '3';
+  const BANK_IDS: string[] = [BANK_ID_ONE, BANK_ID_TWO, BANK_ID_THREE];
+
+  describe('search-input-validator', () => {
+    it('returns empty errors if no bank and year', () => {
+      // Arrange
+
```
JoshBinns2000, line 13: I'm guessing empty `// Arrange` is here to make it clear that you are not including a bank or year?

trade-finance-manager-ui/server/controllers/utilisation-reports/find-reports-by-year/search-input-validator.ts
```diff
@@ -0,0 +1,60 @@
...
+const isBefore2024 = (year: number): boolean => year < 2024;
+
+const isInTheFuture = (year: number): boolean => year > new Date().getFullYear();
+
+const validateYearInputAndReturnErrorMessageIfInvalid = (year: unknown): string | undefined => {
+  const validYearRegex = new RegExp(REGEX.YEAR);
+  if (!year || !isNonEmptyString(year) || !validYearRegex.test(year)) {
+    return 'Enter a valid year';
+  }
+  const yearParsed = Number(year);
```
JoshBinns2000, line 24: Should this be `parsedYear`?

trade-finance-manager-ui/server/controllers/utilisation-reports/find-reports-by-year/search-input-validator.ts
```diff
@@ -0,0 +1,60 @@
...
+  bankError: string | undefined;
+  yearError: string | undefined;
+};
+
+const isBefore2024 = (year: number): boolean => year < 2024;
+
+const isInTheFuture = (year: number): boolean => year > new Date().getFullYear();
+
+const validateYearInputAndReturnErrorMessageIfInvalid = (year: unknown): string | undefined => {
+  const validYearRegex = new RegExp(REGEX.YEAR);
```
JoshBinns2000, line 20: This creation of a `RegExp` object isn't needed - a regex literal can be used in the same was as a `RegExp` object: `REGEX.YEAR.test(year)`

trade-finance-manager-ui/server/controllers/utilisation-reports/find-reports-by-year/search-input-validator.ts
```diff
@@ -0,0 +1,60 @@
...
+  if (!bankId || !isNonEmptyString(bankId) || !allBankIds.includes(bankId)) {
+    return 'Select a bank';
+  }
+  return undefined;
+};
+
+export const validateSearchInput = (
+  bankId: unknown,
+  year: unknown,
+  allBankIds: string[],
```
JoshBinns2000, line 41: I think these parameters should be wrapped up in an object rather than being defined one after another in this case

trade-finance-manager-ui/server/controllers/utilisation-reports/find-reports-by-year/search-input-validator.ts
```diff
@@ -0,0 +1,60 @@
...
+): FindUtilisationReportByYearValidationErrors => {
+  if (!bankId && !year) {
+    return { errorSummary: [], bankError: undefined, yearError: undefined };
+  }
+
+  const errorSummary: Error[] = [];
+  const bankError = validateBankInputAndReturnErrorMessageIfInvalid(bankId, allBankIds);
```
JoshBinns2000, line 48: This `validate<Thing>AndReturn<Thing>IfInvalid` seems a bit bloated to me. What do you think about instead calling this function `getValidationErrorForBankInput`?

trade-finance-manager-ui/server/test-mocks/mock-banks.ts
```diff
@@ -13,7 +13,7 @@ const MONTHLY_REPORT_PERIOD_SCHEDULE = [
   { startMonth: 12, endMonth: 12 },
 ];
 
-const MOCK_BANKS = {
+export const MOCK_BANKS = {
```
JoshBinns2000, line 16: Seeing as you've added the `Bank` type below, could we add the `MOCK_BANKS: Bank[]` type assertion here?

trade-finance-manager-ui/server/controllers/utilisation-reports/find-reports-by-year/search-input-validator.ts
```diff
@@ -0,0 +1,60 @@
+import { isNonEmptyString } from '@ukef/dtfs2-common';
+import { REGEX } from '../../../constants';
+
+type Error = {
```
JoshBinns2000, line 4: This probably shouldn't be called `Error` as the `Error` type definition is already taken by the javascript `Error` class

trade-finance-manager-ui/templates/utilisation-reports/find-utilisation-reports-by-year.njk
```diff
@@ -0,0 +1,75 @@
...
+                text: yearError,
+                attributes: {
+                  'data-cy': 'year-error-message'
+                }
+              },
+              value: selectedYear
+            }) }}
+          {{ govukButton ({
+            text: "Continue",
+            attributes: {'data-cy': 'continue-button'}
```
JoshBinns2000, line 69: Should this be on a new line (like line 62)?

trade-finance-manager-ui/server/controllers/utilisation-reports/find-reports-by-year/search-input-validator.test.ts
```diff
@@ -0,0 +1,126 @@
...
+      // Assert
+      expect(errorSummary).toEqual(expectedErrorSummary);
+      expect(bankError).toEqual(undefined);
+      expect(yearError).toEqual(expectedYearError);
+    });
+
+    it('returns year error if year param is after current year', () => {
+      // Arrange
+      const now = new Date();
+      const dateInNextYear = add(now, { years: 1 });
```
JoshBinns2000, line 97: Could we use `addYears` here instead?

JoshBinns2000 (dismissed):
Most of this looks good. My main concern at the moment is that we are using query parameters to decide which template to render, whereas I think it is probably more appropriate to use a different route with path parameters when a successful search is done so that the `render the previous reports` controller is separate from the `find reports by bank and year` controller

trade-finance-manager-ui/server/controllers/utilisation-reports/find-reports-by-year/search-input-validator.ts
JoshBinns2000: Changed to `ErrorSummaryItem`

trade-finance-manager-api/api-tests/v1/banks/get-all-banks.api-test.ts
```diff
@@ -0,0 +1,27 @@
...
+    it('gets banks for authenticated user', async () => {
+      // Arrange
+      const user = await testUserCache.initialise(app);
+
+      // Act
+      const response: CustomResponse = await as(user).get('/v1/banks');
+
+      // Assert
+      expect(response.status).toEqual(200);
+      expect(response.body).not.toBeNull();
```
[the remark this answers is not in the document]
ChristianMcc: I reckon not worth as this test isn't populating a bank then testing it's the right banks it's just relying on data from mocker-data-loader, I don't like coupling those too tightly even though it's done in other places

trade-finance-manager-ui/templates/utilisation-reports/find-utilisation-reports-by-year.njk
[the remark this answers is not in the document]
ChristianMcc: Rename sounds good, i don't mind too much if it doesn't change

ChristianMcc (dismissed):
LGTM :)

ChristianMcc (approved):

PR #3092, opened 2024-05-03
feat(FN-1615): add generic for express render options

JoshBinns2000:
## Introduction :pencil2:
It would be nice to have a way of defining the type we want to use for the express `res.render` method.

- Adds a new declaration file to merge the declaration for the express `core.Response.render` function
- Updates `res.render` in the controller for `getUtilisationReportReconciliationForReport` in TFM-UI

### DIFF
--- a/trade-finance-manager-ui/server/controllers/utilisation-reports/utilisation-report-reconciliation-for-report/index.ts
+++ b/trade-finance-manager-ui/server/controllers/utilisation-reports/utilisation-report-reconciliation-for-report/index.ts
@@ -1,16 +1,8 @@
 import { Request, Response } from 'express';
-import { SessionBank, ValuesOf, getFormattedReportPeriodWithLongMonth } from '@ukef/dtfs2-common';
+import { getFormattedReportPeriodWithLongMonth } from '@ukef/dtfs2-common';
 import api from '../../../api';
 import { asUserSession } from '../../../helpers/express-session';
 import { PRIMARY_NAVIGATION_KEYS } from '../../../constants';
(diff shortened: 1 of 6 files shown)


trade-finance-manager-ui/server/controllers/utilisation-reports/utilisation-report-reconciliation-for-report/index.ts
```diff
@@ -9,9 +9,14 @@ export const getUtilisationReportReconciliationByReportId = async (req: Request,
   const { reportId } = req.params;
 
   try {
-    const utilisationReportReconciliationDetails = await api.getUtilisationReportReconciliationDetailsById(reportId, userToken);
+    const utilisationReportReconciliationDetails = await api.getUtilisationReportReconciliationDetailsById(
+      reportId,
+      userToken,
```
abhi-markan, line 14: Various GitHub alerts on type any.

trade-finance-manager-ui/server/types/express-serve-static-core.d.ts
```diff
@@ -0,0 +1,17 @@
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
```
abhi-markan, line 1: Please remove inline rule override.

abhi-markan (changes requested):
If this approach is must then I would prefer being applied across other UI microservices for consistency.

abhi-markan:
@JoshBinns2000 is this PR still required or can we close this?

JoshBinns2000:
The pattern of adding a `render<View>` function at the controller level has been added in preference to this typescript-first approach

PR #3098, opened 2024-05-03
Feat/dtfs2-7075/refactor-and-commonise-validation-of-payloads

AlexBramhill:
## Introduction :pencil2:
As part of the work adding the `isTrusted` role, I've found the need to have partial validation on our payloads.

To fix this I looked at commonising our existing validation of payloads into common lib. I also took the time to convert the file to typescript and utilise zod, which makes our checking more precise.

I also commonised the schemas.

added `areAllPropertiesRequired` to allow for checking of updates


libs/common/src/schemas/payload.ts
```diff
@@ -0,0 +1,73 @@
+import z from 'zod';
+
+export const BANK = z.object({
+  id: z.string(),
+  name: z.string(),
+  mga: z.array(z.string()),
+  emails: z.array(z.string()),
```
JoshBinns2000, line 7: zod has inbuilt email validation too: `z.string().email()`

PR #3109, opened 2024-05-13
feat(FN-1615): fix migration adding status to work with ledger tables

JoshBinns2000:
## Introduction :pencil2:
Ledger tables require you new columns added to be nullable even if they have a default value. However, once they are populated with the default value, they can then be altered to be non-nullable. 

- Updates the migration to first add a nullable `status` column with a default value, and then alters the column to be non-nullable


ChristianMcc (approved):
LGTM :)

PR #3122, opened 2024-05-15
feat(FN-1615): seed more random fee records

JoshBinns2000:
## Introduction :pencil2:
In order to get more realistic tests for the premium payments table, we want to seed many more random fee records.

- Uses `@faker-js/faker` to generate random fee record ids, exporters and amounts
- Uses a batch save for fee records to save 200 randomly generated fee records
- Wraps the seeder `save` calls in a transaction so that the seeder can be re-ran if it fails

- Removes the `REPORT_COMPLETED` status from the `FEE_RECORD_STATUS` constant


RLavender98 (approved):
Looks good, thanks for adding 😄

PR #3123, opened 2024-05-15
feat(FN-1692): add select all checkbox and hidden js elements

JoshBinns2000:
## Introduction :pencil2:
We want to add a "select all" checkbox which can toggle all the checkboxes inside the premium payments table. We want this checkbox to only be visible if javascript is enabled and hidden otherwise (`visibility: hidden` is apparently better for accessibility than `display: none`, hence using this css property for the `.hide-if-js-not-enabled` class).

- Adds a new script which adds an event listener to the select all checkbox if it is present
- Adds a new script which removes the `.hide-if-js-not-enabled` class from all elements where it finds it when the document loads
- Adds a new `select-all-table-cell-checkbox.njk` component with component tests
- Updates the `premium-payments-table.njk` table to have use the above component in the table header row

The select all checkbox behaves as follows:
- if javascript is not enabled, it is not visible on the screen
- when checked, sets all the other checkboxes to checked
- when unchecked, sets all the other checkboxes to unchecked
- if all the other checkboxes are checked, it automatically toggles to checked
- if any of the other checkboxes are unchecked, it automatically toggles to unchecked

image


trade-finance-manager-ui/scripts/enable-select-all-table-checkbox.js
```diff
@@ -0,0 +1,55 @@
+const selectAllTableCellCheckboxSelector = 'table.govuk-table input[type="checkbox"]#select-all-checkbox';
+const selectAllTableCellCheckbox = document.querySelector(selectAllTableCellCheckboxSelector);
+
+const allOtherTableCellCheckboxesSelector = 'table.govuk-table input[type="checkbox"]:not(#select-all-checkbox)';
+const allOtherTableCellCheckboxes = document.querySelectorAll(allOtherTableCellCheckboxesSelector);
```
JoshBinns2000, line 5: `querySelectorAll` returns a `NodeList`. In chrome, it is possible to extract an array from this type using `document.querySelectorAll(...).entries().toArray()`. However, this `.toArray` method does not appear in my intellisense in VSCode, which leads me to believe this behaviour may not be supported in all browsers. To be on the safe side, I went with the `Array` prototype methods below which I believe have very good browser support, but I'm not sure if this is something we necessarily want to do

trade-finance-manager-ui/styles/hide-if-js-not-enabled.scss
```diff
@@ -0,0 +1,3 @@
+.hide-if-js-not-enabled {
+  visibility: hidden;
+}
```
RLavender98, line 3: I think we should be able to make use of the js-enabled script by using the following:
```
.hide-if-js-not-enabled {
  visibility: hidden;
}

.js-enabled .hide-if-js-not-enabled {
  visibility: visible;
}
```
And then we shouldn't need the script that removes all the instances of this class

trade-finance-manager-ui/templates/utilisation-reports/_macros/premium-payments-table.njk
```diff
@@ -28,9 +31,11 @@
         data-cy="premium-payments-table-row--feeRecordId-{{ feeRecord.id }}"
       >
         <td class="govuk-table__cell">
-          {{ tableCellCheckbox.render({
-            checkboxId: checkboxId
-          }) }}
+          {% if feeRecord.status === 'TO_DO' or feeRecord.status === 'DOES_NOT_MATCH' %}
+            {{ tableCellCheckbox.render({
+              checkboxId: checkboxId
+            }) }}
```
RLavender98, line 37: Would it be possible to have a component test for this?

trade-finance-manager-ui/scripts/enable-select-all-table-checkbox.js
RLavender98: If this works I think it is fine, it is pretty neat and sensible to think about the browser support. We could also just use a `forEach` over the `NodeList` with some additional logic instead of the array `all` and `some` methods but I don't know that it matters too much?

trade-finance-manager-ui/scripts/enable-select-all-table-checkbox.js
RLavender98, line 1: I'm guessing this is complicated to write tests for other than in e2e-tests? Is there a plan to add some for the page?

RLavender98 (changes requested):
A couple of suggestions and questions on tests but otherwise looks great!

trade-finance-manager-ui/styles/hide-if-js-not-enabled.scss
JoshBinns2000: I updated the styles near the table instead of in this (the `js-hidden` technically already does the same thing as this, but with `display: none` which I wanted to avoid)

trade-finance-manager-ui/scripts/enable-select-all-table-checkbox.js
JoshBinns2000: I tend to prefer using `some` and `every` over a `forEach` because (as far as I know) you can't get it to stop early without wrapping it in a try-catch and making it throw, whereas those two will stop early if they don't meet the criteria

trade-finance-manager-ui/scripts/enable-select-all-table-checkbox.js
RLavender98: Ok let's just try and make sure we at least try to test via e2e-tests when we add them for the page 👍

trade-finance-manager-ui/component-tests/utilisation-reports/_macros/premium-payments-table.component-test.js
```diff
@@ -83,4 +89,26 @@ describe(component, () => {
...
+      status: feeRecordStatus,
+    };
+    const wrapper = render({ feeRecords: [feeRecordItem] });
+
+    const rowSelector = `[data-cy="premium-payments-table-row--feeRecordId-${feeRecordItem.id}"]`;
+    if (checkboxShouldExist) {
+      wrapper.expectElement(`${rowSelector} input#feeRecordId-${feeRecordItem.id}`).toExist();
+    } else {
+      wrapper.expectElement(`${rowSelector} input#feeRecordId-${feeRecordItem.id}`).notToExist();
+    }
```
RLavender98, line 112: Could we split this into two tests, one for the 'should' cases and one for the 'should not' cases, to me on the readability vs commonisation scale this feels one step too commonised

RLavender98 (approved):

PR #3125, opened 2024-05-15
feat(FN-1615): update the formatting for the currency and amount string

JoshBinns2000:
## Introduction :pencil2:
We want to use proper formatting for currency and amount strings.

- Updates the currency helper to use a formatter
- Adds unit tests for the currency helpers


RLavender98 (dismissed):

RLavender98 (approved):
Nice tests 😄

PR #3135, opened 2024-05-17
feat(FN-1697): add a payment button

JoshBinns2000:
## Introduction :pencil2:
We want to introduce the add payment route to TFM-UI. This route should be `/utilisation-reports/:reportId/add-payment` and should be hidden behind the `FF_TFM_PAYMENT_RECONCILIATION_ENABLED` feature flag and redirect back to `/utilisation-reports/:reportId` with a specific error being displayed and persisting the checkboxes which were checked by the user before submitting the form.

- Adds a new `POST` route*
- Adds a new middleware which validates the request body for this route
- Add functionality to store the checked checkbox ids and add payment error key in `req.session` before the redirect, process it in the `getUtilisationReconciliationByReportId` controller with a new helper and clear the data stored in `req.session` after it has been processed in the controller
- Updates the `table-cell-checkbox.njk` template to accept an `isChecked` parameter to set whether or not the checkbox should already be checked on render
- Adds a helper to get the specific error summary view model based on the add payment error key stored in `req.session`
- Updates the `utilisation-report-reconciliation-for-report.njk` template to render the page with the error if an error is supplied
- Adds and updates many tests
- Updates the `enable-select-all-table-cell-checkbox.js` client-side script to check the select all checkbox if all the checkboxes are already checked on page load

*a `POST` route is used so that the checkbox ids which need to be processed are stored in the request body rather than in the request query

- Adds a view model type for the `ErrorSummaryViewModel`


trade-finance-manager-ui/server/types/express-session.d.ts
```diff
@@ -7,5 +8,8 @@ export type UserSessionData = {
 
 // https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/express-session/index.d.ts#L199-L211
 declare module 'express-session' {
-  interface SessionData extends UserSessionData {}
+  interface SessionData extends UserSessionData {
+    addPaymentErrorKey: AddPaymentErrorKey;
+    checkedCheckboxIds: Record<string, true | undefined>;
+  }
```
RLavender98, line 14: I'm not sure if these fields should be on the base type as that would apply for all endpoints?

trade-finance-manager-ui/server/middleware/validatePostAddPaymentRequestBody/index.ts
```diff
@@ -0,0 +1,73 @@
...
+  reportId: string,
+  addPaymentError: AddPaymentErrorKey,
+  checkedCheckboxIds: Record<string, true | undefined> = {},
+) => {
+  req.session.addPaymentErrorKey = addPaymentError;
+  req.session.checkedCheckboxIds = checkedCheckboxIds;
+  return res.redirect(`/utilisation-reports/${reportId}`);
+};
+
+export const validatePostAddPaymentRequestBody = (req: Request, res: Response, next: NextFunction) => {
```
RLavender98, line 28: This is quite a long function so I am going to see if I can find a way to extract some parts of it.

trade-finance-manager-ui/server/controllers/utilisation-reports/utilisation-report-reconciliation-for-report/index.ts
```diff
@@ -14,18 +15,22 @@ export const getUtilisationReportReconciliationByReportId = async (req: Request,
   const { reportId } = req.params;
 
   try {
+    const { addPaymentErrorSummary, isCheckboxChecked } = getAndClearRedirectSessionData(req);
```
RLavender98, line 18: We could so with a controller test for the error handling

trade-finance-manager-ui/server/controllers/utilisation-reports/utilisation-report-reconciliation-for-report/get-and-clear-redirect-session-data.ts
```diff
@@ -0,0 +1,39 @@
+import { Request } from 'express';
+import { ErrorSummaryViewModel } from '../../../types/view-models';
+import { getAddPaymentError } from '../helpers/get-add-payment-error-helper';
+
+const clearRedirectSessionData = (req: Request): void => {
+  delete req.session.addPaymentErrorKey;
+  delete req.session.checkedCheckboxIds;
+};
+
+export const getAndClearRedirectSessionData = (
```
RLavender98, line 10: I feel this would be best named in a way that is related to the page, I know it lives in the folder but it otherwise reads very generically but isn't which could be a source of confusion

trade-finance-manager-ui/templates/_macros/table-cell-checkbox.njk
```diff
@@ -1,16 +1,28 @@
...
+          checked
+        />
+      {% else %}
+        <input
+          class="govuk-checkboxes__input govuk-!-top-0 table-cell-checkbox-input"
+          id="{{ checkboxId }}"
+          data-cy="table-cell-checkbox--{{ checkboxId }}"
+          name="{{ checkboxId }}"
+          type="checkbox"
+        />
```
RLavender98, line 24: I'm hoping we can do this without the else, by doing `checked="{{checked}}"` but I need to check if that works

trade-finance-manager-ui/server/types/express-session.d.ts
RLavender98: Fields here don't need to be optional because the type of the request session is typed as: `session.Session & Partial<session.SessionData>;`

francescastocco (approved):
LGTM :)

PR #3145, opened 2024-05-22
chore(dtfs2-none): fix typo

natdeanlewissoftwire:
## Introduction :pencil2:
Generic copy updates that I found as part of my poking around the service post rampup

Makes the example 8 digit companies house actually be 8 digits

Adding spaces, punctuation, etc


AlexBramhill:
Really good spots -- I am surprised nothing is kicking up a fuss regarding the branch name... Would you mind changing it to
`chore/dtfs2-none/fix-typo` and the pr name to `chore(dtfs2-none): fix typo` please? these are our standard patterns :) in terms of syntax (and I think this is more a chore than a feature, and I'm not sure how relevant `rampup` is to the ticket name as this makes it into change logs).

PR #3149, opened 2024-05-23
fix(dtfs2-none): fix boolean coersion for sql config

AlexBramhill:
## Introduction :pencil2:
There is a surprising way that zod coerces booleans (https://zod.dev/?id=coercion-for-primitives). This means that `"false"` would parse as `true`

I've made the coercion a lot more restrictive. We could have simply checked and done `typeof val === 'string' && val.toLowerCase() === 'false' ? false : val`, however I'm not sure we want to coerse all things truthy and falsy.

### DIFF
--- a/libs/common/src/helpers/schemas/z-boolean-coerce.test.ts
+++ b/libs/common/src/helpers/schemas/z-boolean-coerce.test.ts
@@ -1,0 +1,59 @@
+import { zBooleanCoerce } from './z-boolean-coerce';
+
+describe('zBooleanCoerce', () => {
+  describe('when parsing a value with zBooleanCoerce', () => {
+    describe('when the provided value can be coerced to a boolean', () => {
+      it.each(getSuccessTestCases())('should return $result when the provided value is $description', ({ value, result }) => {
+        expect(zBooleanCoerce.parse(value)).toBe(result);
+      });
+    });
+
+    describe('when the provided value cannot be coerced to a boolean', () => {
+      it.each(getFailureTestCases())('should throw an error when the provided value is $description', ({ value }) => {
+        expect(() => zBooleanCoerce.parse(value)).toThrow();
+      });
+    });
+  });
+
+  describe('when parsing a value with zBooleanCoerce and additional chaining', () => {
+    describe('when chaining with .optional()', () => {
+      it('should return undefined when the provided value is undefined', () => {
+        expect(zBooleanCoerce.optional().parse(undefined)).toBe(undefined);
+      });
+    });
+
+    describe('when chaining with .nullable()', () => {
+      it('should return null when the provided value is null', () => {
+        expect(zBooleanCoerce.nullable().parse(null)).toBe(null);
+      });
+    });
+  });
+
+  function getSuccessTestCases() {
+    return [
+      { value: true, result: true, description: 'true' },
+      { value: false, result: false, description: 'false' },
+      { value: 'true', result: true, description: '"true"' },
+      { value: 'false', result: false, description: '"false"' },
+      { value: 'TRUE', result: true, description: '"TRUE"' },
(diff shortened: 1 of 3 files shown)


libs/common/src/helpers/schemas/z-boolean-coerce.test.ts
```diff
@@ -0,0 +1,59 @@
...
+  describe('when parsing a value with zBooleanCoerce', () => {
+    describe('when the provided value can be coerced to a boolean', () => {
+      it.each(getSuccessTestCases())('should return $result when the provided value is $description', ({ value, result }) => {
+        expect(zBooleanCoerce.parse(value)).toBe(result);
+      });
+    });
+
+    describe('when the provided value cannot be coerced to a boolean', () => {
+      it.each(getFailureTestCases())('should throw an error when the provided value is $description', ({ value }) => {
+        expect(() => zBooleanCoerce.parse(value)).toThrow();
```
MarRobSoftwire, line 13: fwiw could ensure we're not getting an error for another reason
```suggestion
        expect(() => zBooleanCoerce.parse(value)).toThrow(ZodError);
```

MarRobSoftwire, line 13: fwiw could ensure we're not getting an error for another reason
```suggestion
        expect(() => zBooleanCoerce.parse(value)).toThrow(ZodError);
```

MarRobSoftwire (approved):

libs/common/src/helpers/schema/z-boolean-strict-coerce.ts
```diff
@@ -0,0 +1,24 @@
+import z from 'zod';
+
+const booleanLike = z.union([
```
abhi-markan, line 3: Do we even need `zod` package, can we not do this natively?

abhi-markan (commented):
Good find and solution, however can we achieve this natively wihtout using `zod`?

libs/common/src/helpers/schema/z-boolean-strict-coerce.ts
AlexBramhill: I'd say zod offers a lot more than us hand rolling validation ourselves (our existing validation has a few blind spots -- which zod covers nicely).

It also provides good error msg reporting, and is easy to use and read.

It also allows for type inference from schemas, meaning once we add validation to something, we also can generate a typescript type from it (one source of truth):
```
const settingsSchema = z.object({
  darkMode: z.boolean().default(false),
  notifications: z.boolean().default(true),
});

type Settings = z.infer<typeof settingsSchema>;
```

tldr - I think it offers a massive timesave into us rolling our own validation, with some great benefits like type inference.

libs/common/src/helpers/schema/z-boolean-strict-coerce.ts
```diff
@@ -0,0 +1,24 @@
...
+ * This addresses an issue where 'false' would be coerced to true -- causing issues in env vars.
+ */
+export const zBooleanStrictCoerce = booleanLike.transform((val) => {
+  if (typeof val === 'string') {
+    return ['true', '1'].includes(val.toLowerCase());
+  }
+  if (typeof val === 'number') {
+    return val === 1;
+  }
+  return val;
```
MarRobSoftwire, line 23: fwiw I think we can cast the value to string, since we've reduced it to one of the above literals
```suggestion
return ['true', '1'].includes(String(val).toLowerCase());
```

MarRobSoftwire (approved):
Another minor comment but LGTM

libs/common/src/helpers/schema/z-boolean-strict-coerce.ts
AlexBramhill: great spot

PR #3150, opened 2024-05-23
feat(FN-1696): add add payment page with selected fee record details displayed

RLavender98:
## Introduction :pencil2:
- Add 'Add a payment' page
- Fetch and display selected fee records details on add payment page

- Resolves [FN-1696](https://ukef-dtfs.atlassian.net/browse/FN-1696)

## Misc ➕
- Add page titles to the other utilisation report pages (shows up as name of browser tab when on page)


dtfs-central-api/src/v1/routes/utilisation-reports-routes.js
```diff
@@ -169,4 +170,37 @@ utilisationReportsRouter
...
+ *       400:
+ *         description: Bad request
+ *       404:
+ *         description: Not Found (if the report with matching id cannot be found)
+ *       500:
+ *         description: Internal Server Error
+ */
+utilisationReportsRouter
+  .route('/:id/selected-fee-records-details')
+  .get(validation.sqlIdValidation('id'), handleExpressValidatorResult, getSelectedFeeRecordDetails);
```
JoshBinns2000, line 204: I think it might be fair in this case for the selected fee record ids to be stored in the query rather than the `GET` request body. My reasoning is that this will only ever be an array of numbers, so it's probably a lot safer for us to assume that we won't reach the URL length limit even if we select many fee records than it would be if we were using an object, for instance. I also think the documentation above needs to be updated to either include the request body if we decide to keep that or the query if we move to that :)

dtfs-central-api/src/v1/validation/utilisation-report-service/selected-fee-record-validator.test.ts
```diff
@@ -0,0 +1,47 @@
...
+      // Act + Assert
+      expect(() => validateSelectedFeeRecordsAllHaveSamePaymentCurrency(selectedFeeRecords)).toThrow(InvalidPayloadError);
+    });
+  });
+
+  function aFeeRecordWithPaymentCurrency(paymentCurrency: Currency) {
+    return FeeRecordEntityMockBuilder.forReport(UtilisationReportEntityMockBuilder.forStatus('RECONCILIATION_IN_PROGRESS').build())
+      .withPaymentCurrency(paymentCurrency)
+      .build();
+  }
```
JoshBinns2000, line 46: Could this be moved to the top and could we use an arrow function for it?

trade-finance-manager-api/src/v1/controllers/utilisation-reports/get-selected-fee-records-details.controller.ts
```diff
@@ -0,0 +1,16 @@
...
+export const getSelectedFeeRecordsDetails = async (req: Request<{ id: number }, object, { feeRecordIds: number[] }>, res: Response) => {
+  try {
+    const { id } = req.params;
+    const { feeRecordIds } = req.body;
+    const SelectedFeeRecordsDetails = await api.getSelectedFeeRecordsDetails(id, feeRecordIds);
+    res.status(200).send(SelectedFeeRecordsDetails);
+  } catch (error) {
+    const errorMessage = 'Failed to get selected fee records details';
+    console.error(errorMessage, error);
+    res.status((error as AxiosError).response?.status ?? 500).send(errorMessage);
```
JoshBinns2000, line 14: I'm not sure we can always assume this is an `AxiosError`, so this might want to be `error?.response?.status ?? 500` (it probably will be an axios error in prod, but when working on a new feature we may get a `TypeError` for example here which would cause the container to crash unexpectedly due to `response` not existing on `error` in this case, which might be quite hard to debug)

trade-finance-manager-ui/server/controllers/utilisation-reports/add-payment/index.test.ts
```diff
@@ -0,0 +1,67 @@
...
+console.error = jest.fn();
+
+describe('controllers/utilisation-reports/:id/add-payment', () => {
+  const userToken = 'user-token';
+  const requestSession = {
+    userToken,
+    user: MOCK_TFM_SESSION_USER,
+  };
+  describe('when initial post request with valid body is received from premium payments page', () => {
+    const requestBodyForPostFromPremiumPaymentsPage: object = {
```
JoshBinns2000, line 18: Was there a reason to type this as `object`?

trade-finance-manager-ui/server/controllers/utilisation-reports/add-payment/index.ts
```diff
@@ -0,0 +1,49 @@
...
+import api from '../../../api';
+import { asUserSession } from '../../../helpers/express-session';
+import {
+  getFeeRecordIdFromPremiumPaymentsCheckboxId,
+  getPremiumPaymentsCheckboxIdsFromObjectKeys,
+} from '../../../helpers/premium-payments-table-checkbox-id-helper';
+import { CustomExpressRequest } from '../../../types/custom-express-request';
+
+export type AddPaymentRequest = CustomExpressRequest<{
+  reqBody: object;
```
JoshBinns2000, line 13: I think this type should probably be `Record<FeeRecordViewModelCheckboxId, 'on'>` (`object` as a type is a bit too broad I think)

trade-finance-manager-ui/server/helpers/premium-payments-table-checkbox-id-helper.ts
```diff
@@ -0,0 +1,29 @@
...
+const FEE_RECORD_STATUS_REGEX = new RegExp(FEE_RECORD_STATUS_REGEX_GROUP);
+
+const PREMIUM_PAYMENTS_TABLE_CHECKBOX_ID_REGEX = new RegExp(
+  `feeRecordId-\\d+-reportedPaymentsCurrency-${CURRENCY_REGEX_GROUP}-status-${FEE_RECORD_STATUS_REGEX_GROUP}`,
+);
+
+export const getPremiumPaymentsCheckboxIdsFromObjectKeys = (object: object): string[] =>
+  Object.keys(object).filter((key) => PREMIUM_PAYMENTS_TABLE_CHECKBOX_ID_REGEX.test(key));
+
+export const getFeeRecordStatusFromPremiumPaymentsCheckboxId = (checkboxId: string): FeeRecordStatus => {
```
JoshBinns2000, line 16: As we are using the non-null assertion on the result of `.match`, we should maybe think about narrowing the type of the input parameter to the `FeeRecordViewModelCheckboxId` type - otherwise, we may want to consider adding some error handling. What do you think?

trade-finance-manager-ui/templates/utilisation-reports/_macros/selected-reported-fees-details-table.njk
```diff
@@ -0,0 +1,36 @@
...
+  <table class="govuk-table"
+         data-module="moj-sortable-table"
+         data-cy="selected-reported-fees-details-table">
+    <caption class="govuk-table__caption govuk-table__caption--m">Selected reported fees details</caption>
+    <thead class="govuk-table__head">
+      <tr class="govuk-table__row">
+        <th scope="col" class="govuk-table__header">Facility ID</th>
+        <th scope="col" class="govuk-table__header">Exporter</th>
+        <th scope="col" class="govuk-table__header govuk-table__header--numeric">Reported fee</th>
+        <th scope="col" class="govuk-table__header govuk-table__header--numeric">Reported payments</th>
```
JoshBinns2000, line 13: Shouldn't this be `Reported payment`? Or should the above be `Reported fees`?

trade-finance-manager-ui/templates/utilisation-reports/add-payment.njk
```diff
@@ -0,0 +1,17 @@
+{% import "./_macros/selected-reported-fees-details-table.njk" as selectedReportedFeesDetailsTable %}
+
+{% extends "index.njk" %}
+
+{% block pageTitle %}Add a payment{% endblock%}
+{% block content %}
+  <div class="govuk-grid-row govuk-!-padding-top-7">
+    <div class="govuk-grid-column-one-half">
+      <span class="govuk-caption-l" data-cy="add-a-payment-heading-caption">{{ bank.name }}, {{ reportPeriod }}</span>
```
JoshBinns2000, line 9: Could we maybe call `reportPeriod` `formattedReportPeriod` to be consistent with the other templates?

trade-finance-manager-ui/test-helpers/test-data/add-payment-view-model.ts
```diff
@@ -0,0 +1,10 @@
+import { AddPaymentViewModel } from '../../server/types/view-models';
+
+export const anAddPaymentViewModel = (): AddPaymentViewModel => ({
+  bank: { name: 'Test bank ' },
+  reportPeriod: 'Some reporting period',
+  reportedFeeDetails: {
+    totalReportedPayments: 'GBP 200',
+    feeRecords: [],
```
JoshBinns2000, line 8: Shouldn't matter as this is just mock data, but do you think we could get any unexpected results from having a value for `totalReportedPayments` when `feeRecords` is empty?

JoshBinns2000 (changes requested):
Looking good - mostly minor comments

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-selected-fee-records-details.controller/helpers.test.ts
```diff
@@ -0,0 +1,125 @@
...
+      const result = await mapToSelectedFeeRecordDetails('123', aReportPeriod(), [feeRecordWithDifferingCurrencies, feeRecordWithMatchingCurrencies]);
+
+      // Assert
+      expect(result.totalReportedPayments).toEqual<CurrencyAndAmount>({ amount: 3000, currency: 'GBP' });
+    });
+  });
+
+  function aFeeRecord() {
+    return FeeRecordEntityMockBuilder.forReport(UtilisationReportEntityMockBuilder.forStatus('RECONCILIATION_IN_PROGRESS').build()).build();
+  }
```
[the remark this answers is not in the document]
RLavender98: So I wrote this as a function over an arrow function so I could write it at the bottom, and I want it at the bottom because the idea is it is fairly irrelevant what it does, so it's nice to not have it cluttering the top of the file.

trade-finance-manager-ui/templates/utilisation-reports/_macros/selected-reported-fees-details-table.njk
RLavender98: That is what the screenshot of the design on my ticket has as the headers

dtfs-central-api/src/v1/routes/utilisation-reports-routes.js
RLavender98: Hmm I hadn't really considered quite how much shorter this would be than the original page request, and agree it will be a fair bit shorter. I do however still think it is safer to do it this way, and has also now already been implemented this way 😅 so I am not too keen to change it unless you think there is a good reason to do so, especially as it is less safe

trade-finance-manager-ui/test-helpers/test-data/add-payment-view-model.ts
RLavender98: I do think mock data should be a representation of a realistic state so I will add in one, I think it just felt like effort at the time when this did satisfy the type 😅

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-selected-fee-records-details.controller/helpers.test.ts
```diff
@@ -0,0 +1,125 @@
...
+        .withPaymentExchangeRate(1.1)
+        .build();
+
+      const feeRecordWithMatchingCurrencies = FeeRecordEntityMockBuilder.forReport(
+        UtilisationReportEntityMockBuilder.forStatus('PENDING_RECONCILIATION').build(),
+      )
+        .withFeesPaidToUkefForThePeriod(1000)
+        .withFeesPaidToUkefForThePeriodCurrency('GBP')
+        .withPaymentCurrency('GBP')
+        .build();
```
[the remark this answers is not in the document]
RLavender98: I've changed it, I can't decide if it matters since the method doesn't use that fact at all and shouldn't, but it does match the real use case data closer

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-selected-fee-records-details.controller/helpers.ts
```diff
@@ -0,0 +1,37 @@
...
+
+const getTotalReportedPayments = (feeRecords: SelectedFeeRecordDetails[]): CurrencyAndAmount => ({
+  currency: feeRecords[0]?.reportedPayment.currency,
+  amount: feeRecords.map((feeRecord) => feeRecord.reportedPayment.amount).reduce((total, currentAmount) => total + currentAmount),
+});
+
+export const mapToSelectedFeeRecordDetails = async (
+  bankId: string,
+  reportPeriod: ReportPeriod,
+  selectedFeeRecordEntities: FeeRecordEntity[],
```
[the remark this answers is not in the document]
RLavender98: That's an interesting point, since it was just test readability in this case I have updated the tests to be more readable as I don't generally like packaging up parameters just to immediately unpack them, but that is an interesting case

trade-finance-manager-api/src/v1/controllers/utilisation-reports/get-selected-fee-records-details.controller.ts
RLavender98: Ah fairs, I did just blindly copy this 😬 `error?.response?.status ?? 500` sadly doesn't play nice but found another controller with a solution and copied that

trade-finance-manager-ui/server/helpers/premium-payments-table-checkbox-id-helper.ts
RLavender98: Yeah that makes sense, I have moved and renamed the type as now used fairly extensively outside the view model

dtfs-central-api/src/v1/routes/utilisation-reports-routes.js
JoshBinns2000: From looking online there seems to be many hot takes from both the `GET request shouldn't use body` and `GET request should use body` sides, so I'm happy to go with it. An interesting concern I read about was that you may end up with less security as HTML servers typically aren't set up to handle `GET` requests with a body, so things may sneak through checks that would normally get picked up in query or path parameters. I don't know how relevant this is for modern servers though, so I'm happy to keep the `GET` request body here

trade-finance-manager-api/src/v1/controllers/utilisation-reports/get-selected-fee-records-details.controller.ts
JoshBinns2000: Nice solution :)

dtfs-central-api/src/v1/validation/utilisation-report-service/selected-fee-record-validator.test.ts
JoshBinns2000: Replied to in another thread: it's kept at the bottom as the way this function works isn't really important for the test

JoshBinns2000 (approved):
LGTM :)

PR #3153, opened 2024-05-24
feat(FN-2588): search reports by year results tfm

francescastocco:
## Introduction :pencil2:
When a user searches for utilisation reports by bank and year we need to display tthe results

- new route in dtfs-central-api to get utilisation reports by bank ID and year
- new function in the bank repo to get bank by ID
- common validation for year
- new route in tfm-api to get utilisation reports by bank ID and year
- use govuk table component to display results

- add prettier ignore path all `prettier:fix` commands in the `package.json` files


trade-finance-manager-ui/templates/utilisation-reports/_macros/search-reports-table.njk
```diff
@@ -0,0 +1,52 @@
...
+    {{- reportRows.push([
+      {html: reportPeriodHtml }, 
+      {text: reconciliationStatus.render({
+        status: report.status,
+        displayStatus: report.displayStatus
+      }) }, 
+      {text: report.formattedDateUploaded},
+      {text: report.totalFeesReported},
+      {text: report.reportedFeesLeftToReconcile}
+    ]) | reject() }}
```
francescastocco, line 26: This `| reject()` feels a bit of a hacky way to not display the output of the `push()` method but couldn't get it to work without it, any suggestions welcome

trade-finance-manager-ui/templates/utilisation-reports/_macros/search-reports-table.njk
```diff
@@ -0,0 +1,52 @@
...
+  {% set user = params.user %}
+  {% set reports = params.reports %}
+  {% set isTfmPaymentReconciliationFeatureFlagEnabled = params.isTfmPaymentReconciliationFeatureFlagEnabled %}
+  {% set reportRows = [] %}
+
+  {% for report in reports %}
+    {% set reportPeriodHtml =
+      '<p class="govuk-body govuk-!-margin-0">' + report.formattedReportPeriod + '</p>' 
+      if not isTfmPaymentReconciliationFeatureFlagEnabled 
+      else '<a class="govuk-link govuk-link--no-visited-state" href="/utilisation-reports/' + report.reportId + '">' + report.formattedReportPeriod + '</a>'
```
francescastocco, line 14: Not keen on this syntax for concatenating strings, is there a better way to do it?

trade-finance-manager-ui/server/controllers/utilisation-reports/find-reports-by-year/index.ts
```diff
@@ -70,7 +71,18 @@ export const getFindReportsByYear = async (req: Request, res: Response) => {
       });
     }
 
-    return renderPreviousBankReportsByYear(res, { user });
+    const { bankName, year, reports } = await api.getReportsByBankAndYear(userToken, bankIdQuery!, yearQuery!);
```
francescastocco, line 74: I'm not sure about using the non-null assertion operator here, thoughts on alternatives?

trade-finance-manager-ui/server/api.js
```diff
@@ -895,6 +896,29 @@ const getAllBanks = async (userToken) => {
...
+    isValidBankId(bankId);
+    assertValidIsoYear(year);
+
+    const { data } = await axios.get(`${TFM_API_URL}/v1/utilisation-reports/${bankId}/${year}`, {
+      headers: generateHeaders(userToken),
+    });
+
+    return data;
+  } catch (error) {
+    console.error('Failed to get previous utilisation reports', error);
```
francescastocco, line 917: Need to update this error text

trade-finance-manager-ui/component-tests/utilisation-reports/_macros/search-reports-table.component-test.js
```diff
@@ -0,0 +1,83 @@
...
+  });
+
+  it('should render the table data with links to the utilisation report reconciliation for bank page when the feature is enabled', async () => {
+    const wrapper = await getWrapper({ isTfmPaymentReconciliationFeatureFlagEnabled: true });
+    const { reports } = wrapper.params;
+
+    reports.forEach((report) => {
+      wrapper.expectElement(`td:contains("${report.formattedReportPeriod}")`).toExist();
+
+      wrapper.expectLink('td > a').toLinkTo(`/utilisation-reports/${report.reportId}`, report.formattedReportPeriod);
```
francescastocco, line 69: This test was failing locally with 
```
Expected: "November 2023"
Received: "November 2023November 2023"
```
but when I printed the html to the console it was like this which is what I was expecting so not sure why this is failing
image

e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/find-utilisation-reports-by-bank-and-year.spec.js
francescastocco, line 1: Could I pair with someone on this? I've hard-coded the bank IDs in the test which I don't really like and wanted to only insert data against one bank but couldn't figure it out

trade-finance-manager-ui/templates/utilisation-reports/_macros/search-reports-table.njk
RLavender98: In the `dashboard-deals-table.njk` they set a html fragment as follows:
```
{% set linkHtml %}
        <a
          href="{{ url }}{{ deal._id }}"
          class="govuk-link"
          data-cy="deal__link--{{ deal._id }}"
        >
          {{ deal.bankInternalRefName }}
        </a>
      {% endset %}

```
So perhaps could try something along those lines?

trade-finance-manager-ui/server/controllers/utilisation-reports/find-reports-by-year/index.ts
RLavender98: Ahh this is interesting... Perhaps we could have the `validateSearchInput` function return the values once validated as `string`s and use those? That way if ever the validation changes to not check it will cause a type error on this line?

dtfs-central-api/api-tests/v1/utilisation-reports/get-utilisation-reports.api-test.ts
```diff
@@ -185,3 +185,173 @@ describe('GET /v1/bank/:bankId/utilisation-reports', () => {
...
 });
+
+describe('GET /v1/utilisation-reports/:bankId/year', () => {
+  const getUrl = (bankId: string, year: string) => `/v1/utilisation-reports/${bankId}/${year}`;
+
+  const portalUser = {
+    _id: new ObjectId(),
+    firstname: 'Test',
+    surname: 'User',
+  } as PortalUser;
```
RLavender98, line 196: I'm not a fan of `as`ing this, I know we do it in a fair few places but there is a `aPortalUser` test helper already defined which we can call to get a portal user and then just set the relevant fields on if they need to be set to specific values for the tests.

trade-finance-manager-ui/server/controllers/utilisation-reports/helpers/find-reports-by-year-helper.ts
```diff
@@ -0,0 +1,21 @@
...
+import { UtilisationReportReconciliationSummaryItem } from '../../../types/utilisation-reports';
+import { reconciliationStatusCodeToDisplayStatus } from './reconciliation-summary-helper';
+
+type ReportViewModel = UtilisationReportReconciliationSummaryItem & {
+  formattedReportPeriod: string;
+  displayStatus: string;
+  formattedDateUploaded?: string;
+};
+
+export const getReportViewModel = (apiItem: UtilisationReportReconciliationSummaryItem): ReportViewModel => {
```
RLavender98, line 12: I feel like the name of this method and the type it returns is a little vague, we display reports in different ways in different granularities so if we could make this more specific that would be better.

I'm also not very clear on how this is lining up correctly with the 'UtilisationReportsByBankAndYearViewModel' type?

trade-finance-manager-ui/server/controllers/utilisation-reports/find-reports-by-year/index.test.ts
```diff
@@ -262,12 +262,16 @@ describe('controllers/utilisation-reports/find-reports-by-year', () => {
...
 
       // Act
       await getFindReportsByYear(req, res);
 
       // Assert
-      expect(res._getRenderView()).toEqual('utilisation-reports/previous-bank-reports-by-year.njk');
+      expect(res._getRenderView()).toEqual('utilisation-reports/utilisation-reports-by-bank-and-year-results.njk');
+      expect((res._getRenderData() as UtilisationReportsByBankAndYearViewModel)?.bankName).toBe(BANK_NAME_ONE);
+      expect((res._getRenderData() as UtilisationReportsByBankAndYearViewModel)?.year).toBe(yearQuery);
+      expect((res._getRenderData() as UtilisationReportsByBankAndYearViewModel)?.reports).toStrictEqual([]);
```
RLavender98, line 274: I do think it would be good to check that if reports are returned they get mapped over correctly if it isn't too much of a faff to do so

dtfs-central-api/src/v1/routes/utilisation-reports-routes.js
```diff
@@ -169,4 +170,35 @@ utilisationReportsRouter
...
+ *               items:
+ *                 $ref: '#/definitions/UtilisationReportReportReconciliationSummaryItem'
+ *       400:
+ *         description: Bad request
+ *       500:
+ *         description: Internal Server Error
+ */
+utilisationReportsRouter
+  .route('/:bankId/:year')
+  .get(validation.bankIdValidation, validation.yearValidation('year'), handleExpressValidatorResult, getUtilisationReportsByBankIdAndYear);
```
RLavender98, line 202: In a similar vein of thought to the comment on the path, I think it would be good if the controller name indicated that we are returning reconciliation summaries rather than just reports as I made false assumptions on my first reading that it was returning full utilisation report objects.

RLavender98 (commented):
Looks good! Mostly just a couple of comments around clarity of naming 😄

trade-finance-manager-ui/templates/utilisation-reports/_macros/search-reports-table.njk
francescastocco: Oooh yes, this is what I was looking for

trade-finance-manager-ui/templates/utilisation-reports/_macros/search-reports-table.njk
francescastocco: This was where I got the ` | reject()` from, just missed this example somehow

dtfs-central-api/src/v1/routes/utilisation-reports-routes.js
```diff
@@ -169,4 +170,35 @@ utilisationReportsRouter
...
+ *               type: array
+ *               items:
+ *                 $ref: '#/definitions/UtilisationReportReportReconciliationSummaryItem'
+ *       400:
+ *         description: Bad request
+ *       500:
+ *         description: Internal Server Error
+ */
+utilisationReportsRouter
+  .route('/:bankId/:year')
```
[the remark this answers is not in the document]
francescastocco: Yep, this makes sense & is a good point, I hadn't given much thought to the path

trade-finance-manager-ui/server/controllers/utilisation-reports/helpers/find-reports-by-year-helper.ts
francescastocco: Oh yes good point, I've updated them

dtfs-central-api/src/v1/routes/utilisation-reports-routes.js
francescastocco: Moving to `bank-routes`

RLavender98 (commented):
Looks good, just two very minor comments!

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-reports.controller/index.ts
```diff
@@ -4,9 +4,9 @@ import { CustomExpressRequest } from '../../../../types/custom-express-request';
 import { UtilisationReportRepo } from '../../../../repositories/utilisation-reports-repo';
 import { parseReportPeriod } from '../../../../utils/report-period';
 import { mapUtilisationReportEntityToGetUtilisationReportResponse } from '../../../../mapping/mapUtilisationReport';
-import { mapReportToSummaryItem } from '../helper';
-import { getBankById } from '../../../../repositories/banks-repo';
-import { NotFoundError } from '../../../../errors';
+// import { mapReportToSummaryItem } from '../helper';
+// import { getBankById } from '../../../../repositories/banks-repo';
+// import { NotFoundError } from '../../../../errors';
```
RLavender98, line 9: Seems to be some commented code left in by accident

RLavender98 (approved):

PR #3158, opened 2024-05-28
feat(FN-2882): update premium payments table alignment

JoshBinns2000:
## Introduction :pencil2:
We want to update the alignment on columns in the premium payments table which contain numeric values.

- Updates the template to use the `govuk-table__header--numeric` and `govuk-table__cell--numeric` classes
- Updates the component tests to check these classes are used where they are needed


francescastocco (approved):
LGTM

PR #3159, opened 2024-05-29
feat(FN-2900): update local SQL to use ledger tables

JoshBinns2000:
## Introduction :pencil2:
The database we use in deployed environments uses ledger tables, but our local setup does not. This means that it is possible for migrations to work locally and on the test pipeline, but fail once deployed. To get around this, we want to update the initial migration to create all tables with ledger enabled. This results in the original `db:drop-all` command we have to fail as we cannot drop tables in the same way, so we also need to update this command to properly reset the database.

- Updates the initial migration without changing the filename*
- Adds a new `reset-database.sh` script to `libs/common/src/sql-db-connection`

*we don't need to run this migration as a new migration in deployed environments, so there is no need to generate a new migration for it


ChristianMcc (approved):
Nice one :)

PR #3168, opened 2024-05-30
fix(DTFS2-7159): fix ts-node type checking in all services

JoshBinns2000:
## Introduction :pencil2:
The `ts-node` type checking rules differ slightly from the standard typescript compiler. This can lead to situations where `tsc --noEmit` will pass with no errors, but the container will fail to run due to a typescript type-checking error.

After some reading, a simple solution to this is to include the below option in `tsconfig.json`. This tells `ts-node` to not use its default `include`/`exclude` options when type checking, but instead use the `tsconfig.json` `include`/`exclude` options.

```json
{
  "ts-node": {
    "files": true
  },
  "include": [...],
  "exclude": [...]
}
```

For further reading, refer to the below links:
- https://stackoverflow.com/questions/67436289/typescripts-declaration-merging-not-working-as-expected-using-ts-node
- https://github.com/TypeStrong/ts-node/issues/1132#issuecomment-716642560


e2e-tests/tsconfig.json
```diff
@@ -98,6 +98,9 @@
       // "skipDefaultLibCheck": true,                      /* Skip type checking .d.ts files that are included with TypeScript. */
       "skipLibCheck": true                                 /* Skip type checking all .d.ts files. */
     },
+    "ts-node": {
+      "files": true
+    },
     "include": ["**/*.ts"],
     "exclude": ["./node_modules/*", "node_modules"]
   }
```
MarRobSoftwire, line 106: legacy but fwiw no new line at end of file

libs/common/tsconfig.json
```diff
@@ -28,5 +28,9 @@
     /* Completeness */
     "skipLibCheck": true
   },
-  "include": ["**/*.ts"]
+  "ts-node": {
+    "files": true
+  },
+  "include": ["**/*.ts"],
+  "exclude": []
```
MarRobSoftwire, line 35: Do we want to explicitly ignore `node_modules` as in other places?

e2e-tests/tsconfig.json
MarRobSoftwire: Also in a couple other places

MarRobSoftwire (dismissed):

libs/common/tsconfig.json
JoshBinns2000: I'm not sure...I think for consistency we probably should, so I'll add that change and see if everything still works

MarRobSoftwire (approved):

PR #3170, opened 2024-05-30
fix(FN-NONE): update npm start scripts

JoshBinns2000:
## Introduction :pencil2:
The current `npm start` script always tries to rebuild all container images defined in `docker-compose.yml`. This can be very slow especially if you already have some images cached. Typically, `docker compose up` is sufficient as this will build only the images you don't have but doesn't try to rebuild ones that have already been built successfully.

- Define `npm start` to run `docker compose up`
- Add a `npm run start:build` script which runs `docker compose up --build`
- Define the `:build` script variants for both


ChristianMcc (approved):

package.json
```diff
@@ -41,7 +41,9 @@
     "prettier:all": "npm run prettier --workspaces --if-present",
     "prettier:fix:all": "npm run prettier:fix --workspaces --if-present",
     "spellcheck": "cspell lint --gitignore --no-must-find-files --unique --no-progress --show-suggestions --color '**/*'",
-    "start": "docker compose up --build",
+    "start": "docker compose up",
+    "start:build": "docker compose up --build",
```
MarRobSoftwire, line 45:
```suggestion
    "start:build": "npom run start -- --build",
```

JoshBinns2000: Sure - why not

ChristianMcc (approved):

PR #3027, opened 2024-04-15
feat(FN-2183): reorder table columns

francescastocco:
## Introduction :pencil2:
Auditable Base Entity columns should be moved to the end of the AzureFileInfo, FeeRecord and UtilisationReport tables

## Resolution :heavy_check_mark:
- Updated migration

PR #3171, opened 2024-05-30
feat(FN-1739): add payment table

JoshBinns2000:
## Introduction :pencil2:
We want to add a new SQL `Payment` table.

- Adds the new entity
- Generates the new migration (with ledger enabled)
- Adds the new `ManyToMany` relation to the fee record entity
- Adds a new entity mock builder for the `Payment`
- Adds new methods to the fee record entity mock builder to allow for defining payments

### DIFF
--- a/libs/common/src/sql-db-connection/migrations/1716990167763-AddPaymentTable.ts
+++ b/libs/common/src/sql-db-connection/migrations/1716990167763-AddPaymentTable.ts
@@ -6,15 +6,15 @@ export class AddPaymentTable1716990167763 implements MigrationInterface {
   public async up(queryRunner: QueryRunner): Promise<void> {
     await queryRunner.query(`
             CREATE TABLE "Payment" (
-                "lastUpdatedAt" datetime2 NOT NULL CONSTRAINT "DF_3c7c91b94fc3c2cb358a90fa629" DEFAULT getdate(),
-                "lastUpdatedByPortalUserId" nvarchar(255),
-                "lastUpdatedByTfmUserId" nvarchar(255),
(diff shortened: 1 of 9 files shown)


libs/common/src/sql-db-connection/migrations/1716990167763-AddPaymentTable.ts
```diff
@@ -0,0 +1,67 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class AddPaymentTable1716990167763 implements MigrationInterface {
+  name = 'AddPaymentTable1716990167763';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`
+            CREATE TABLE "Payment" (
+                "lastUpdatedAt" datetime2 NOT NULL CONSTRAINT "DF_3c7c91b94fc3c2cb358a90fa629" DEFAULT getdate(),
```
ChristianMcc, line 9: I think we looked at putting these as the last columns instead of first columns before, I think francesca did this before but I'm not sure if it was a manual migration change or something else with the entities, @francescastocco do you remember?

ChristianMcc (commented):
Nice one, LGTM apart from small comment about ordering of columns in SQL

libs/common/src/sql-db-connection/migrations/1716990167763-AddPaymentTable.ts
JoshBinns2000: It was something Francesca did manually (#3027) - I'll make that fix now

ChristianMcc (approved):
LGTM, did we just manually change the migration and we can't do anything about it entity wise? (I think that might have been what I remember from the past but sad if so)

PR #3172, opened 2024-05-30
feat(FN-1848): check fee records match payment method

JoshBinns2000:
## Introduction :pencil2:
We want to add a method to the `PaymentEntity` class which returns whether or not the payment matches the fee records.

- Adds the method
- Adds unit tests
- Adds a `PAYMENT_MATCHING_CURRENCY_TO_TOLERANCE_MAP` constant with all values currently set to zero


libs/common/src/sql-db-entities/payment/payment.entity.ts
```diff
@@ -51,4 +54,26 @@ export class PaymentEntity extends AuditableBaseEntity {
...
   }
+
+  public feeRecordsMatchPayment(): boolean {
+    if (this.feeRecords.length === 0) {
+      throw new PaymentHasNoFeeRecordsError();
+    }
+
+    if (this.feeRecords.some(({ paymentCurrency }) => paymentCurrency !== this.currency)) {
+      throw new PaymentAndFeeRecordCurrencyDoesNotMatchError();
+    }
```
RLavender98, line 65: I feel a bit conflicted about these errors, it's a bit hard to tell without seeing the intended usage, but I would say we should never have a payment not attached to any fee records, and we should never have a payment that doesn't match the fee record payment currency. Any request to that effect should be rejected before we ever create an entity no? As such, it seems bizarre to be checking for that and maintaining specific error codes for those cases that are currently only getting thrown from here?

RLavender98: I've read the ticket and see why these checks have been added. That said, I don't really like that the ticket this comes from has no ACs, I think it would make more sense for these checks to be added down the line, if and when it becomes possible for these situations to occur?

libs/common/src/sql-db-entities/payment/payment.entity.ts
```diff
@@ -51,4 +54,26 @@ export class PaymentEntity extends AuditableBaseEntity {
...
+    const totalFeesPaidToUkefForThePeriodInPaymentCurrency = this.feeRecords
+      .map((feeRecord) => feeRecord.getFeesPaidToUkefForThePeriodInThePaymentCurrency())
+      .reduce((total, feesPaid) => total.add(feesPaid), new Big(0));
+
+    const difference = totalFeesPaidToUkefForThePeriodInPaymentCurrency.minus(this.amountReceived).abs();
+    if (difference.gt(tolerance)) {
+      return false;
+    }
+    return true;
+  }
```
RLavender98, line 78: How will this work and be used in the many to many case, e.g. payment 1 and payment 2 are both against fee record 1, fee record 2 and fee record 3. Then neither payment will match the total individually, they will need to be added to each other and then compared to the total of the reported fees. I'm tempted to say we should scrap this ticket and PR and include this within a PR of the wider context?

JoshBinns2000: I didn't think about that case, but you are right and that probably means we would need to have a separate function rather that a method on the entity. Unlike with the entity, we can't be sure that the `feeRecords` and `payments` that get passed to this hypothetical function are linked, so that would also require some additional checks.

Either way, I'm going to close this PR and just get this in with the `ADD_PAYMENT` event handler

PR #3174, opened 2024-05-30
feat(FN-1488): handle find reports by year form submission with empty params

francescastocco:
## Introduction :pencil2:
When the `find reports by year` form is submitted with empty params we should display errors to the user

- Check whether original URL contains a `?`


trade-finance-manager-ui/server/controllers/utilisation-reports/find-reports-by-year/index.test.ts
```diff
@@ -144,13 +144,41 @@ describe('controllers/utilisation-reports/find-reports-by-year', () => {
...
+    });
+
     it("renders the 'find-utilisation-reports-by-year.njk' view with required data when there are query params with errors", async () => {
       // Arrange
       const yearQuery = '20';
 
       const { res, req } = httpMocks.createMocks({
         session: requestSession,
         query: { year: yearQuery },
+        originalUrl: '/utilisation-reports/find-reports-by-year?',
```
JoshBinns2000, line 181: May be worth storing this in a variable instead of repeating it in each test setup?

JoshBinns2000 (commented):
Looks good - only one comment about the `originalUrl` maybe being stored in a variable rather than being repeated

JoshBinns2000 (approved):
LGTM

PR #3178, opened 2024-05-30
fix(FN-NONE): enable dockerignore by adding it to root

JoshBinns2000:
## Introduction :pencil2:
Because of the project monorepo structure, the `context` in the `docker-compose*.yml` files is always set to be the project root. This means that, unlike with standard projects, docker expects the `.dockerignore` file to be at the same level as the `context` rather than the level of the `Dockerfile`. As such, the `.dockerignore` files have been ignored before now.

- Consolidates all `.dockerignore` files into a root level `.dockerignore`
- Adds directives to ignore `api-tests`, `test-helpers` and `__mocks__` directories
- Adds directives to ignore `.jest.config.js` files inside directories*

*there seems to be an issue with `.dockerignore` files ignore patterns with respect to recursive matching based on file extension. For example, `**/*.jest.config.js` will not match anything, but `dtfs-central-api/*.jest.config.js` will match all files at the `dtfs-central-api` level which have the extension `.jest.config.js`. This also means that it is not possible for us to ignore `*.test.js` files without explicitly finding every place where these files exist. However, for files such as `.jest.config.js` which exist at only one level, it is easy enough to include these in the `.dockerignore` file.


generate-dockerignore.gha.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
```
abhi-markan, line 1: I am not sure if this script is required?

.base-gha.dockerignore
```diff
@@ -0,0 +1,50 @@
+**/node_modules
```
abhi-markan, line 1: Would prefer one singular `.dockerignore` file at the root please.

generate-dockerignore.local.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
```
abhi-markan, line 1: What is the purpose of this script please?

abhi-markan (changes requested):

generate-dockerignore.gha.sh
JoshBinns2000: This is a draft PR, so this isn't final. Either way, the reason for this script was so that we could have separate `.dockerignore` files for local dev and github actions. This is required because the local `.dockerignore` files attempt to ignore `.jest.config.js` files, which stops the api-test QAT pipelines from running

JoshBinns2000:
The way dockerignore works in deployments is apparently different to the way we do it for local development, so this PR might fix local development but adversely affect deployment

PR #3206, opened 2024-06-03
feat(FN-2900): update local SQL to use ledger tables (#3159)

JoshBinns2000:
## Introduction :pencil2:
We want to enable ledger tables locally for `release-2.0`.


ChristianMcc (dismissed):
LGTM :)

.github/workflows/deployment.yml
```diff
@@ -72,18 +72,24 @@ jobs:
...
         run: echo "Timezone set to ${{ env.TIMEZONE }}"
 
-  sql-migrations:
-    name: SQL Migrations 🗄️
-    if: ${{ github.event.workflow_run.conclusion == 'success' }}
-    needs: setup
-    runs-on: [self-hosted, DTFS, deployment]
+  database:
+    name: MSSQL 🗄️
+    if: ${{ github.event.workflow_run.conclusion == 'success' && vars.DATABASE == '1' }}
```
ChristianMcc, line 89: Could we move this now @abhi-markan ? everyhing is on SQL so we always want to run migrations now

.github/workflows/deployment.yml
```diff
@@ -72,18 +72,24 @@ jobs:
       environment: ${{ env.ENVIRONMENT }}
       timezone: ${{ env.TIMEZONE }}
     steps:
+      - name: Configure 🔨
+        if:  ${{ !contains(fromJson('["feature", "staging", "prod"]'), github.event.workflow_run.head_branch) }}
+        run: |
+          echo "ENVIRONMENT=dev" >> $GITHUB_ENV
```
ChristianMcc, line 78: So this just overrides it from head_branch to dev if it's not feature/staging/prod? is this just a bit of a plaster fix for getting release-2.0 to go to dev but no where else?

ChristianMcc (approved):

.github/workflows/deployment.yml
abhi-markan: This is to avoid running on `prod`

.github/workflows/deployment.yml
abhi-markan: Correct, this is for all future `release-*` to use `dev` environment.

.github/workflows/deployment.yml
ChristianMcc: Ah I guess if release-1.6 beats us there then we won't want it to run
Also jsut realised- these changes are on deployment.yml when we only care about them on main

PR #3207, opened 2024-06-03
chore(FN): update SQL documentation for CDC changes

ChristianMcc:
## Introduction :pencil2:
I wanted to add some short documentation regarding CDC (change data capture) which is enabled on staging and production

### DIFF
--- a/doc/sql-db.md
+++ b/doc/sql-db.md
@@ -186,3 +186,8 @@ SQL_DB_LOGGING_ENABLED
 Note: in contrast to other common functionality which is imported from `'@ukef/dtfs2-common'`, the `SqlDbDataSource` is imported from a separate `'@ukef/dtfs2-common/sql-db-connection'` sub-directory. This means that only those packages that need to connect to the DB and import from this sub-directory will need to be provided with access to the required environment variables.
 
 [//]: # 'TODO FN-1859 - add details on how to use repos with `extend` and use of Data Mapper pattern'
+
+## CDC restrictions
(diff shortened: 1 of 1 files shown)


doc/sql-db.md
```diff
@@ -186,3 +186,8 @@ SQL_DB_LOGGING_ENABLED
 Note: in contrast to other common functionality which is imported from `'@ukef/dtfs2-common'`, the `SqlDbDataSource` is imported from a separate `'@ukef/dtfs2-common/sql-db-connection'` sub-directory. This means that only those packages that need to connect to the DB and import from this sub-directory will need to be provided with access to the required environment variables.
 
 [//]: # 'TODO FN-1859 - add details on how to use repos with `extend` and use of Data Mapper pattern'
+
+## CDC restrictions
+CDC (Change Data Capture) is enabled on staging and prod environments which will block certain actions which cause a table to be dropped. This is not limited to just dropping tables but also queries which behind the scenes SQL will drop the old table and make a new one, e.g. re-ordering columns in a table.
```
JoshBinns2000, line 191:
```suggestion
CDC (Change Data Capture) is enabled on staging and prod environments which will block certain actions which cause a table to be dropped. This means that any operation which will result in a table being dropped, directly or otherwise, is not permitted. For example, reordering columns in SQL Server can only be achieved by dropping the old table and creating a new one, which will be blocked by CDC.
```

JoshBinns2000 (commented):
Looks good - just one suggested change to make one sentence a little clearer

PR #3209, opened 2024-06-03
feat(FN-1713): add sorting to selected fee record details table

RLavender98:
## Introduction :pencil2:
Make the selected reported fee details table on the add payments page sortable.

- By default sorts by facility id
- Exporter column is sortable alphabetically
- Reported fee and reported payments columns are sortable first by currency (alphabetically) then by value (increasing)


JoshBinns2000 (approved):
LGTM

PR #3211, opened 2024-06-04
feat(FN-1739): add new payment to utilisation report

JoshBinns2000:
## Introduction :pencil2:
We want to add the backend for adding a payment to a utilisation report.

- Renames the `UtilisationReportStateMachine` and `FeeRecordStateMachine` event
- Adds a middleware to validate the `add-payment` payload using `zod`
- Adds extensive unit and api testing
- Adds swagger docs for the new endpoint

- Adds a new `getMockQueryRunner` test helper to create a mock query runner to avoid repeating the same setup in each unit test file where transactions need to be tested


dtfs-central-api/src/services/state-machines/utilisation-report/utilisation-report.state-machine.ts
```diff
@@ -24,7 +24,10 @@ export class UtilisationReportStateMachine {
   }
 
   public static async forReportId(id: number): Promise<UtilisationReportStateMachine> {
-    const report = await UtilisationReportRepo.findOneByOrFail({ id });
+    const report = await UtilisationReportRepo.findOneBy({ id });
+    if (!report) {
+      throw new NotFoundError(`Failed to find report with id ${id}`);
+    }
```
JoshBinns2000, line 30: This change allows us to more easily use the `ApiError` in controllers

dtfs-central-api/src/v1/controllers/utilisation-report-service/post-add-payment.controller/index.test.ts
```diff
@@ -0,0 +1,158 @@
...
+import { PostAddPaymentRequest, postAddPayment } from '.';
+import { TfmSessionUser } from '../../../../types/tfm/tfm-session-user';
+import { aTfmSessionUser } from '../../../../../test-helpers/test-data/tfm-session-user';
+import { addPaymentToUtilisationReport } from './helpers';
+import { PostAddPaymentPayload } from '../../../routes/middleware/payload-validation/validate-post-add-payment-payload';
+import { NewPaymentDetails } from '../../../../types/utilisation-reports';
+
+jest.mock('./helpers');
+
+class TestApiError extends ApiError {
```
JoshBinns2000, line 14: Not sure if this is the best approach to do this, but this allows us to test that errors which extend `ApiError` are handled correctly rather than needing the rely on a specific class which could be removed down the line

dtfs-central-api/src/v1/controllers/utilisation-report-service/post-add-payment.controller/helpers.test.ts
```diff
@@ -0,0 +1,206 @@
...
+      expect(utilisationReportStateMachineConstructorSpy).toHaveBeenCalledWith(reportId);
+    });
+
+    it("calls the 'FeeRecordRepo.findBy' method with the supplied list of fee record ids", async () => {
+      // Act
+      await addPaymentToUtilisationReport(reportId, feeRecordIds, tfmUser, newPaymentDetails);
+
+      // Assert
+      expect(feeRecordFindBySpy).toHaveBeenCalledWith({ id: In(feeRecordIds) });
+    });
```
RLavender98, line 72: Test names which essentially say 'calls the function named blah' are very easy to get out of date if we update the function name or if actually end up fetching the required data a different way. Test names and assertions should focus on what the function is actually intending to do. This could be 'fetches the fee records corresponding to the payment' but I think perhaps more interesting to assert the parameters in a test that is checking we make use of the returned fee records correctly, or make use of the returned state machine correctly. 

One option also is that there is actually a neat way to achieve checking we call with the correct a parameters without having to explicitly assert it is to use this sort of pattern:
```
const fn = jest.fn();
when(fn).calledWith(1).mockReturnValue('yay!');
```
(you do then have to be careful the same response won't be sent for other parameter values though when you set up your mocks)

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/add-a-payment/add-a-payment.event-handler.ts
```diff
@@ -0,0 +1,50 @@
...
+      stateMachine.handleEvent({
+        type: 'PAYMENT_ADDED',
+        payload: {
+          transactionEntityManager,
+          status: feeRecordStatusToSet,
+          requestSource,
+        },
+      }),
+    ),
+  );
```
RLavender98, line 42: It feels out of place for the utilisation report state machine to decide the fee record status to set, I'd almost be tempted to not bother having a fee record state machine and let it set the statuses itself if this logic is already here. Instead, could our fee record state machine perhaps accept a list of fee records and then the logic for determining the status to set can live there? This is probably one we should also get @ChristianMcc 's thoughts on.

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/add-a-payment/helpers.test.ts
RLavender98, line 1: Please can we also add a test where the 'feesPaidToUkefForThePeriodCurrency' differs from the 'paymentCurrency' just to have some protection that we calling the method that coverts rather than just using the 'feesPaidToUkefForThePeriod' straight up?

dtfs-central-api/src/v1/controllers/utilisation-report-service/post-add-payment.controller/helpers.test.ts
```diff
@@ -0,0 +1,206 @@
...
+      // Arrange
+      const feeRecordsWithGBPPaymentCurrency = feeRecordIds.map((feeRecordId) =>
+        FeeRecordEntityMockBuilder.forReport(utilisationReport).withId(feeRecordId).withPaymentCurrency('GBP').build(),
+      );
+      feeRecordFindBySpy.mockResolvedValue(feeRecordsWithGBPPaymentCurrency);
+
+      const newPaymentDetailsWithEURCurrency: NewPaymentDetails = {
+        ...newPaymentDetails,
+        currency: 'EUR',
+      };
```
RLavender98, line 92: Can we add a duplicate of this test in which all but one of the fee records matches the payment currency?

RLavender98 (changes requested):
Nice job! A lot of comments about tests but a lot of them are around just one repeated issue.

dtfs-central-api/api-tests/mocks/test-users/mock-tfm-user.ts
```diff
@@ -9,5 +9,4 @@ export const MOCK_TFM_USER: TfmSessionUser = {
   firstName: 'Test',
   lastName: 'User',
   status: 'active',
-  lastLogin: 0,
```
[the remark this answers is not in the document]
JoshBinns2000: I missed it off the zod schema - have added it back

dtfs-central-api/src/v1/routes/utilisation-reports-routes.js
```diff
@@ -215,4 +217,56 @@ utilisationReportsRouter
...
+ *         description: OK
+ *       400:
+ *         description: Bad request
+ *       404:
+ *         description: Not Found
+ *       500:
+ *         description: Internal Server Error
+ */
+utilisationReportsRouter
+  .route('/:reportId/add-payment')
```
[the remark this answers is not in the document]
JoshBinns2000: Sure

dtfs-central-api/src/v1/routes/middleware/payload-validation/schemas/tfm-team.schema.ts
```diff
@@ -0,0 +1,4 @@
+import { TEAM_IDS, TeamId } from '@ukef/dtfs2-common';
+import z from 'zod';
+
+export const TfmTeamSchema = z.enum(Object.values(TEAM_IDS) as [TeamId, ...TeamId[]]);
```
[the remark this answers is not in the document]
JoshBinns2000: This is because of how `zod` works - when you create a `zod` `enum`, it expects an array with at least one element (the type is `[string, ...string[]]`. `Object.values(TEAM_IDS)` returns `('PDC_RECONCILE' | 'PDC_READ' | ...)[]` which could technically be an empty array, therefore throwing a type error

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/add-a-payment/add-a-payment.event-handler.ts
JoshBinns2000: This reply is just for @ChristianMcc to have some context.

Somewhat implicitly, the use of the fee record state machine here ensures that the fee record is in the appropriate state to add a payment. For instance, if the fee record is in `MATCH` state and you try to add another payment, that error would be caught here. The computation of which status to set is done at the utilisation report state machine level mostly for performance reasons, but also because there should really only be one fee record attached to one fee record state machine. The `'PAYMENT_ADDED'` event accepting the new status to set makes sense to me logically, as we are telling the fee record state machine that a new payment has been added and your payments now match or do not match. The state machine then says either "no, you can't add a new payment to this fee record" or "okay, I'll update the status"

RLavender98: Ooh that is an interesting perspective, I had kind of forgotten about that state check on the state machine. 
As a side comment on that note, do we want to prevent them adding another payment if the fee record has match status? I can't see that AC on any ticket

RLavender98: I was kind of in the mind that we should just treat fee records as groups but I have just realised that once we key they can be updated separately to done so I guess we shouldn't entertain that idea...
We could instead just tell the fee record state machine whether the payments match the reported payments via a boolean and then let the fee record state machine own the knowledge of what status to set? It's a shame it would be really inefficient to calculate it for each fee record inside the state machine...

JoshBinns2000: > As a side comment on that note, do we want to prevent them adding another payment if the fee record has match status? I can't see that AC on any ticket

This is less an "ought" comment and more an "is" from me - at present, the state machine does only permit adding a payment if you are in the `TO_DO` status. I am going to update it so that you can add a payment if the status is `TO_DO`, `DOES_NOT_MATCH` and `MATCH`

dtfs-central-api/src/v1/controllers/utilisation-report-service/post-add-payment.controller/helpers.ts
```diff
@@ -0,0 +1,76 @@
...
+    await queryRunner.commitTransaction();
+  } catch (error) {
+    await queryRunner.rollbackTransaction();
+    if (error instanceof ApiError) {
+      throw new TransactionFailedError(error.message);
+    }
+    throw new TransactionFailedError('Unknown error');
+  } finally {
+    await queryRunner.release();
+  }
```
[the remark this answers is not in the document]
JoshBinns2000: I will work on getting a PR adding a commonised helper for this once this is merged in - the other hidden benefit of this is that we would no longer have to worry about testing the transaction working correctly in each individual function as we would only need to test it once where the helper is defined (with respect to starting transactions, committing, etc)

dtfs-central-api/src/v1/controllers/utilisation-report-service/post-add-payment.controller/helpers.test.ts
JoshBinns2000: I suppose the use of `jest-when` implicitly performs this test, like you say. I'm tempted to try using that for future tests, but seeing as I have included quite extensive explicit checks here I'd prefer to keep them this time and just update the test names (I may try using `when` for the state machine changes you've suggested below though)

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/add-a-payment/add-a-payment.event-handler.test.ts
```diff
@@ -0,0 +1,278 @@
...
+      feeRecords,
+      paymentDetails,
+      requestSource,
+    });
+
+    // Assert
+    expect(mockEntityManager.save).toHaveBeenCalledWith(PaymentEntity, newPaymentEntity);
+  });
+
+  it("calls the 'feeRecordsMatchAttachedPayments' function with the supplied fee records and the new payment", async () => {
```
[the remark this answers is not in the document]
JoshBinns2000: Yeh, this test is a bit unnecessary considering that the tests below check the change in the event handler payload based on what we mock this function to return. I'll just remove this test

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/add-a-payment/add-a-payment.event-handler.test.ts
```diff
@@ -0,0 +1,278 @@
...
+    jest.restoreAllMocks();
+  });
+
+  it('creates and saves the new payment entity using the supplied entity manager', async () => {
+    // Arrange
+    const utilisationReport = UtilisationReportEntityMockBuilder.forStatus('PENDING_RECONCILIATION').build();
+
+    const feeRecords = aListOfFeeRecordsForReport(utilisationReport);
+    const feeRecordStateMatchines = getFeeRecordStateMachinesForFeeRecords(feeRecords);
+    mockFeeRecordStateMachineConstructorImplementation(feeRecords, feeRecordStateMatchines);
```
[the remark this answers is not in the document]
JoshBinns2000: I didn't do this originally because I was concerned about how tricky it would be to mock the state machine, but it's easy enough to just include the following at the bottom of the file

```typescript
function aMockFeeRecordStateMachine(): FeeRecordStateMachine {
  return {
    handleEvent: jest.fn(),
  } as unknown as FeeRecordStateMachine;
}
```

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/add-a-payment/add-a-payment.event-handler.test.ts
```diff
@@ -0,0 +1,278 @@
...
+    jest.resetAllMocks();
+    jest.restoreAllMocks();
+  });
+
+  it('creates and saves the new payment entity using the supplied entity manager', async () => {
+    // Arrange
+    const utilisationReport = UtilisationReportEntityMockBuilder.forStatus('PENDING_RECONCILIATION').build();
+
+    const feeRecords = aListOfFeeRecordsForReport(utilisationReport);
+    const feeRecordStateMatchines = getFeeRecordStateMachinesForFeeRecords(feeRecords);
```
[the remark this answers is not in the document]
JoshBinns2000: Nice - who doesn't like a good typo every now and again 🙃

dtfs-central-api/.eslintrc.js
```diff
@@ -90,5 +90,22 @@ module.exports = {
...
+      parser: '@typescript-eslint/parser',
+      parserOptions: {
+        ...baseParserOptions,
+        project: './tsconfig.eslint.json',
+        tsconfigRootDir: __dirname,
+      },
+      rules: {
+        ...baseRules,
+        '@typescript-eslint/unbound-method': 'off',
+        'no-param-reassign': 'off',
```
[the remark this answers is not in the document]
JoshBinns2000: The `no-param-reassign` was to remove an inline eslint disable, but considering it is just one place I'd rather keep it inline.

The `unbound-method` comes from the new `mockQueryRunner` I am passing around - each function defined on it (because I am using the `as unknown as QueryRunner` type assertion) is defined as a class method, meaning it is bound to whatever `this` is. Typescript eslint does not like this and prefers to use arrow functions where the context of `this` is not clear. An example of where this error appears is `expect(mockQueryRunner.connect).toHaveBeenCalled()` - `.connect` is considered to be unbound.

The solution to this is to temporarily modify `mock-query-runner.ts` to export each inidividual mock as well as the query runner. Considering that the PR after this will introduce a commonised transaction handler which removes the need for this, I will remove the `.eslintrc.js` changes and update `mock-query-runner.ts`

libs/common/src/sql-db-entities/fee-record/fee-record.entity.ts
```diff
@@ -103,6 +103,7 @@ export class FeeRecordEntity extends AuditableBaseEntity {
    */
   @ManyToMany(() => PaymentEntity, (payment) => payment.feeRecords, {
     cascade: ['insert'],
+    onDelete: 'CASCADE',
```
RLavender98, line 106: Just to check have we tested that deleting a payment deletes the connection to the fee record but no the fee record itself and vice versa?

RLavender98 (dismissed):
Looks good! Just one question about the cascade deletes but otherwise happy for you to merge if @ChristianMcc is too!

libs/common/src/sql-db-entities/fee-record/fee-record.entity.ts
JoshBinns2000: I have confirmed that, at least in terms of running SQL commands, the cascade delete only deletes the join table reference

ChristianMcc:
Just FYI leaving this here - we can merge this now but I'm going to have a look post fact so I'm not blocking and if anything comes up will just need another PR to fix if that's okay :)

RLavender98 (approved):

PR #3214, opened 2024-06-06
feat(DTFS2-7055): update docker-compose.gha.yml to use replica set

MarRobSoftwire:
## Introduction :pencil2:
Change streams & transactions need mongodb to use a replica set. In order to test these features the GHA should run on one.

- Update the docker-compose file

## Misc
It's unclear if this change affects the flakiness of the tests. To check this I've run the pipeline 6 times & found:
- 16 test suites failed flakily
- 7 of these were genuine test failures, the remainder `write EPIPE`, docker failures, MDL failures, etc.
- The `dtfs-central-api` utilisisation report tests seem to be the most flakey, accounting for 4 of these 7 and failing more often than not.
Aside from the utilisisation report, this doesnt seem to be a notable increase in flakiness over a normal PR (at a glance)

### DIFF
--- a/docker-compose.gha.yml
+++ b/docker-compose.gha.yml
@@ -21,14 +21,22 @@ services:
       - ./dtfs-sql-volume-dtfs-submissions:/data/db
 
   dtfs-submissions-data:
-    image: mongo:5.0.5
-    restart: always
+    build:
+      context: ./mongodb
+      dockerfile: Dockerfile
     environment:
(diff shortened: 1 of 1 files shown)


abhi-markan (approved):
LGTM

* However this has been probelamatic in the past with MongoDB not booting up properly and has been sporadic, is that not the issue anymore?
* Can we remove replica set docker compose file after this?

PR #3215, opened 2024-06-06
feat(FN-1698): add some e2e tests for add payment journey

RLavender98:
## Introduction :pencil2:
Add some starting e2e-tests for the add payment journey

- Test that fee record details get fetched and displayed
- Test for an error scenario

- Add a helper command to cypress for getting an input by it's label text


e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/add-payment.spec.js
```diff
@@ -0,0 +1,92 @@
...
+    pages.utilisationReportAddPaymentPage.selectedReportedFeesDetailsTable().contains('22222222');
+    pages.utilisationReportAddPaymentPage.selectedReportedFeesDetailsTable().contains('Exporter 2');
+    pages.utilisationReportAddPaymentPage.selectedReportedFeesDetailsTable().contains('EUR 200');
+    pages.utilisationReportAddPaymentPage.selectedReportedFeesDetailsTable().contains('GBP 400');
+
+    pages.utilisationReportAddPaymentPage.selectedReportedFeesDetailsTable().contains('Total reported payments GBP 450');
+  });
+
+  it('should display errors when form submitted with invalid values', () => {
+    cy.getInputByLabelText('Amount received').type('100');
```
RLavender98, line 72: I recognise this is not the most traditional cypress test pattern, I generally prefer writing tests from the perspective of if a user where to visit the page how do they identify landmarks and interact over using lots of test ids, because then the tests are going to be more realistic and close to user interactions. But, if we would prefer to stick to the 'data-cy' selection pattern then I am (somewhat reluctantly) willing to change this to that as I do appreciate that is what [cypress suggest themselves](https://docs.cypress.io/guides/references/best-practices).

RLavender98: Although I do find it interesting that they recommend 'data-cy' as the best method for element selection and then [recommend the testing library extension](https://docs.cypress.io/guides/references/best-practices#Cypress-and-Testing-Library) which is the anti-pattern using user first element selection...

JoshBinns2000 (commented):
I think this needs to have main merged in as some files appear to have been repeated, but the e2e tests look good so far

e2e-tests/tfm/cypress/support/utils/getInputByLabelText.js
```diff
@@ -0,0 +1,6 @@
+export default (label) => {
+  return cy
+    .contains('label', label)
+    .invoke('attr', 'for')
+    .then((id) => cy.get(`#${id}`));
```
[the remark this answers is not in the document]
RLavender98: I have changed this to your suggestion but then have changed it back because there is a slight difference, `.contains('selector', 'text')` gets an element of that selector type with the text where as `.get('selector').contains('text')`, gets all types of that selector and then finds an element within one of then containing that text, so you are not guaranteed that the returned element actually is a label, but to be able to assume it has a 'for' attribute we really do need it to be a html 'label' and not say a span within the label that contains the text we are trying to fetch by

JoshBinns2000 (approved):
LGTM

PR #3216, opened 2024-06-06
chore(FN): fix type-check:all script

JoshBinns2000:
## Introduction :pencil2:
The `npm run type-check:all` script, when run locally, would show a lot of errors due to using the root level `tsconfig.json` file.

The root level `.eslintrc.js` was only used for `utils`, which causes issues if you move the root level `tsconfig.json` to `utils`.

- Removes the root level `tsconfig.json` file
- Removes extra `tsconfig.json` files in `utils/**`
- Adds `tsconfig.json` files where they are needed
- Removes the `type-check` script from `acbs-functions` (this directory has no typescript files, so the `type-check` always shows an error due to it not being able to find any files matching `**/*.ts`)
- Removes the root level `.eslintrc.js`


ChristianMcc (approved):
LGTM :)

PR #3218, opened 2024-06-07
feat(FN-1739): update table column names and data types

JoshBinns2000:
## Introduction :pencil2:
We want to rename some of the Payment table columns and update one of the data types to match our other monetary value columns.

- Updates the `PaymentEntity` class
- Updates references to the old column names

- Makes the `reset-database.sh` script more robust

### DIFF
--- a/libs/common/src/sql-db-connection/migrations/1717746522387-PaymentTableUpdateAmountColumns.ts
+++ b/libs/common/src/sql-db-connection/migrations/1717746522387-PaymentTableUpdateAmountColumns.ts
@@ -1,0 +1,69 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+/**
+ * This migration has been manually modified in order to work
+ * with ledger tables.
+ */
+export class PaymentTableUpdateAmountColumns1717746522387 implements MigrationInterface {
+  name = 'PaymentTableUpdateAmountColumns1717746522387';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    // Create the new columns with the required data-type and set to nullable
+    await queryRunner.query(`
+            ALTER TABLE "Payment"
+            ADD "amount" decimal(14, 2) NULL
+        `);
+    await queryRunner.query(`
+            ALTER TABLE "Payment"
+            ADD "reference" nvarchar(255) NULL
+        `);
+
+    // Copy all the values from the old columns to the new columns
+    await queryRunner.query(`
+            UPDATE "Payment" SET "amount" = "amountReceived"
+        `);
+    await queryRunner.query(`
+            UPDATE "Payment" SET "reference" = "paymentReference"
+        `);
+
+    // Delete the old columns
+    await queryRunner.query(`
+            ALTER TABLE "Payment"
+            DROP COLUMN "amountReceived"
+        `);
+    await queryRunner.query(`
+            ALTER TABLE "Payment"
+            DROP COLUMN "paymentReference"
+        `);
+  }
+
+  public async down(queryRunner: QueryRunner): Promise<void> {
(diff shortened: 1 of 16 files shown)


libs/common/src/sql-db-connection/migrations/1717746522387-PaymentTableUpdateAmountColumns.ts
```diff
@@ -0,0 +1,69 @@
...
+            ADD "amountReceived" int NULL
+        `);
+    await queryRunner.query(`
+            ALTER TABLE "Payment"
+            ADD "paymentReference" nvarchar(255) NULL
+        `);
+
+    // Copy all the values from the old columns to the new columns
+    await queryRunner.query(`
+            UPDATE "Payment" SET "amountReceived" = "amount"
```
ChristianMcc, line 53: This technically might not work going from a decimal to a non decimal but I think leaving this in is fine - we don't expect to run the down migrations much on a non-empty database that has had new data put into it/we could manually get around this. Just noting this for anyone looking at the PR in the future!

libs/common/src/sql-db-connection/reset-database.sh
```diff
@@ -1,10 +1,26 @@
 #!/bin/bash
```
ChristianMcc, line 1: Do you think it's worth having a short explanation anywhere of what this actually does, even if it's a couple of sentences of why we need it in the ledger DB documentation?

ChristianMcc (approved):
LGTM, left a small comment about documentation that might be worth thinking about

libs/common/src/sql-db-entities/payment/payment.entity.ts
```diff
@@ -42,12 +43,12 @@ export class PaymentEntity extends AuditableBaseEntity {
   })
   feeRecords!: FeeRecordEntity[];
 
-  static create({ currency, amountReceived, dateReceived, paymentReference, feeRecords, requestSource }: CreatePaymentParams): PaymentEntity {
+  static create({ currency, amount, dateReceived, reference, feeRecords, requestSource }: CreatePaymentParams): PaymentEntity {
     const payment = new PaymentEntity();
     payment.currency = currency;
-    payment.amountReceived = amountReceived;
+    payment.amount = amount;
     payment.dateReceived = dateReceived;
```
AlexBramhill, line 50: Am guessing we're leaving `dateRecieived` as `dateReceived` to signal that it's not an outstanding payment for whatever reason

AlexBramhill (commented):
lgtm -- could you consider adding the documentation Christian talked about :)

AlexBramhill (approved):
lgtm -- could you consider adding the documentation Christian talked about :)

libs/common/src/sql-db-connection/reset-database.sh
JoshBinns2000: I'm already planning on updating the ledger docs because they're technically out of date now that we know we can't really add non-nullable columns

libs/common/src/sql-db-entities/payment/payment.entity.ts
JoshBinns2000: `dateReceived` feels like it is the correct name as it is the payment the date was received - not the date the payment was added to the table. Leaving it as `date` seems like it would be more ambiguous to me

libs/common/src/sql-db-connection/migrations/1717746522387-PaymentTableUpdateAmountColumns.ts
JoshBinns2000: I've confirmed that the migration does work in the other direction even with entries with decimal amounts in the table - the amount is just converted to an integer

ChristianMcc (approved):

PR #3219, opened 2024-06-07
feat(FN-1739): link add payment frontend to backend

JoshBinns2000:
## Introduction :pencil2:
We want to link up the adding a payment frontend to the backend.

This PR also resolves the [FN-1742](https://ukef-dtfs.atlassian.net/browse/FN-1742) ticket.

- Adds a new route to `TFM-API`
- Adds tests
- Updates the `TFM-UI` controller to allow for adding a new payment

- Fixes the `TFM-API` `custom-express-request.ts` file

### DIFF
--- a/trade-finance-manager-api/src/v1/controllers/utilisation-reports/post-payment.controller.ts
+++ b/trade-finance-manager-api/src/v1/controllers/utilisation-reports/post-payment.controller.ts
@@ -1,0 +1,42 @@
+import { isAxiosError, HttpStatusCode } from 'axios';
+import { Response } from 'express';
+import { Currency } from '@ukef/dtfs2-common';
+import api from '../../api';
+import { CustomExpressRequest } from '../../../types/custom-express-request';
+import { TfmSessionUser } from '../../../types/tfm-session-user';
+
+export type PostPaymentRequestBody = {
+  feeRecordIds: number[];
+  paymentCurrency: Currency;
+  paymentAmount: number;
+  datePaymentReceived: string;
+  paymentReference?: string;
+  user: TfmSessionUser;
+};
+
+export type PostPaymentRequestParams = {
+  reportId: string;
+};
+
+export type PostPaymentRequest = CustomExpressRequest<{
+  reqBody: PostPaymentRequestBody;
+  params: PostPaymentRequestParams;
+}>;
+
+export const postPayment = async (req: PostPaymentRequest, res: Response) => {
+  try {
+    const { reportId } = req.params;
+    const { feeRecordIds, paymentCurrency, paymentAmount, datePaymentReceived, paymentReference, user } = req.body;
+
+    await api.addPaymentToFeeRecords(reportId, feeRecordIds, user, paymentCurrency, paymentAmount, datePaymentReceived, paymentReference);
+    return res.sendStatus(HttpStatusCode.Ok);
+  } catch (error) {
+    const errorMessage = 'Failed to add payment';
+    if (isAxiosError(error)) {
+      console.error(`${errorMessage}: ${error.message}`);
+      return res.status(error.response?.status || HttpStatusCode.InternalServerError).send(`${errorMessage}: ${error.message}`);
+    }
+    console.error(errorMessage);
+    return res.status(HttpStatusCode.InternalServerError).send(errorMessage);
+  }
+};
--- a/trade-finance-manager-api/src/v1/controllers/utilisation-reports/post-payment.controller.ts
+++ b/trade-finance-manager-api/src/v1/controllers/utilisation-reports/post-payment.controller.ts
@@ -32,11 +32,8 @@ export const postPayment = async (req: PostPaymentRequest, res: Response) => {
     return res.sendStatus(HttpStatusCode.Ok);
   } catch (error) {
     const errorMessage = 'Failed to add payment';
-    if (isAxiosError(error)) {
-      console.error(`${errorMessage}: ${error.message}`);
-      return res.status(error.response?.status || HttpStatusCode.InternalServerError).send(`${errorMessage}: ${error.message}`);
-    }
-    console.error(errorMessage);
-    return res.status(HttpStatusCode.InternalServerError).send(errorMessage);
+    const errorStatus = (isAxiosError(error) && error.response?.status) || HttpStatusCode.InternalServerError;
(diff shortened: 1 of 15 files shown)


trade-finance-manager-ui/server/controllers/utilisation-reports/add-payment/add-payment-form-values-validator.ts
```diff
@@ -9,10 +9,20 @@ const isCurrencyNumberOptionallyWithThousandsSeparators = (value: string) => {
   return currencyNumberWithOptionalThousandsSeparatorsRegex.test(value);
 };
 
-const isPaymentCurrencyValid = (paymentCurrency: string | undefined): boolean => {
+const isPaymentCurrencyValid = (paymentCurrency: string | undefined): paymentCurrency is Currency => {
```
ChristianMcc, line 12: hadn't actually seen a type predicate like this before but it's nice!

trade-finance-manager-ui/server/controllers/utilisation-reports/add-payment/index.ts
```diff
@@ -87,21 +99,27 @@ export const addPayment = async (req: AddPaymentRequest, res: Response) => {
...
+    const selectedFeeRecordDetails = await api.getSelectedFeeRecordsDetails(reportId, feeRecordIds, userToken);
     return renderAddPaymentPage(res, {
       user,
       activePrimaryNavigation: PRIMARY_NAVIGATION_KEYS.UTILISATION_REPORTS,
-      reportId: Number(reportId),
+      reportId,
       selectedFeeRecordCheckboxIds: checkedCheckboxIds,
       errors,
-      formValues,
+      formValues: canSubmitFormValues ? EMPTY_ADD_PAYMENT_FORM_VALUES : formValues,
```
ChristianMcc, line 122: What's the scenario for this again? is this supporting clearing the page if we are adding a second payment?

trade-finance-manager-ui/server/controllers/utilisation-reports/add-payment/index.ts
```diff
@@ -87,21 +99,27 @@ export const addPayment = async (req: AddPaymentRequest, res: Response) => {
     const { reportId } = req.params;
     const checkedCheckboxIds = getPremiumPaymentsCheckboxIdsFromObjectKeys(req.body);
     const feeRecordIds = checkedCheckboxIds.map((checkboxId) => getFeeRecordIdFromPremiumPaymentsCheckboxId(checkboxId));
-    const { errors, formValues, paymentNumber } = extractAddPaymentFormValuesAndValidateIfPresent(req.body);
+    const feeRecordPaymentCurrency = getFeeRecordPaymentCurrencyFromPremiumPaymentsCheckboxId(checkedCheckboxIds[0]);
+
+    const { isAddingPayment, errors, formValues, paymentNumber } = extractAddPaymentFormValuesAndValidateIfPresent(req.body, feeRecordPaymentCurrency);
+    const canSubmitFormValues = errors.errorSummary.length === 0;
```
ChristianMcc, line 105: Might be worth just calling this something to do with errors, confused me a little with having a true/false of can we submit form values, when it feels like it's more do we have errors that we are displaying back to the user and therefore keeping data on the page

Even just formHasErrors or something

trade-finance-manager-ui/server/controllers/utilisation-reports/add-payment/index.ts
```diff
@@ -87,21 +99,27 @@ export const addPayment = async (req: AddPaymentRequest, res: Response) => {
...
 
-    if ('addPaymentFormSubmission' in req.body && errors.errorSummary.length === 0) {
-      // TODO FN-1739 save data upon valid submission
-      return res.status(501).send('Your request was valid, but we have not finished building the feature');
+    if (isAddingPayment && canSubmitFormValues) {
+      const { addAnotherPayment, ...paymentFormValues } = formValues;
+      await api.addPaymentToFeeRecords(reportId, paymentFormValues, feeRecordIds, user, userToken);
+      if (addAnotherPayment && addAnotherPayment !== 'true') {
```
ChristianMcc, line 110: is the radio allowed to be unticked and still submit the form? Assumed you'd have to pick yes or no and not have it as undefined

ChristianMcc (commented):
LGTM! couple of small comments on the different errors side of this

trade-finance-manager-ui/server/controllers/utilisation-reports/add-payment/index.ts
JoshBinns2000: Exactly - without this, the second payment form would be pre-populated with the data from the first payment

trade-finance-manager-api/src/v1/controllers/utilisation-reports/post-payment.controller.ts
```diff
@@ -0,0 +1,42 @@
...
+    const { reportId } = req.params;
+    const { feeRecordIds, paymentCurrency, paymentAmount, datePaymentReceived, paymentReference, user } = req.body;
+
+    await api.addPaymentToFeeRecords(reportId, feeRecordIds, user, paymentCurrency, paymentAmount, datePaymentReceived, paymentReference);
+    return res.sendStatus(HttpStatusCode.Ok);
+  } catch (error) {
+    const errorMessage = 'Failed to add payment';
+    if (isAxiosError(error)) {
+      console.error(`${errorMessage}: ${error.message}`);
+      return res.status(error.response?.status || HttpStatusCode.InternalServerError).send(`${errorMessage}: ${error.message}`);
```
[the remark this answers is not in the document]
JoshBinns2000: Sure - I'll just make sure we return the specific error code but not the message

trade-finance-manager-ui/server/controllers/utilisation-reports/add-payment/index.ts
JoshBinns2000: This is more of a typescript thing as the type at this point in the code has been asserted to be `'true' | 'false'`, but it would be hard to make it so that typescript picks that up. I have added a `ValidatedAddPaymentFormValues` type so that I can make this check more clear - let me know what you think

RLavender98 (approved):
LGTM!

PR #3221, opened 2024-06-07
fix(FN): fix swagger docs and open route

JoshBinns2000:
## Introduction :pencil2:
We want to open the route for `/api-docs` locally in `dtfs-central-api`. After opening the route, it was discovered that the swagger docs were not displaying properly.

- Opens the route
- Fixes the swagger docs `api` definition so that they docs are visible at the `/v1/api-docs` route

- Puts some extraneous routes into the `Utilisation Report` tag

### DIFF
--- a/dtfs-central-api/src/generateApp.js
+++ b/dtfs-central-api/src/generateApp.js
@@ -18,6 +18,8 @@ SqlDbDataSource.initialize()
 const generateApp = () => {
   const app = express();
 
+  app.use(`/v1/${SWAGGER_ROUTE}`, swaggerRoutes);
+
   app.use(seo);
   app.use(security);
   app.use(healthcheck);
@@ -39,7 +41,6 @@ const generateApp = () => {
   app.use(`/v1/${TFM_ROUTE}`, tfmRoutes);
   app.use(`/v1/${USER_ROUTE}`, userRoutes);
   app.use(`/v1/${UTILISATION_REPORTS_ROUTE}`, utilisationReportsRoutes);
-  app.use(`/v1/${SWAGGER_ROUTE}`, swaggerRoutes);
 
   // Return 200 on get to / to confirm to Azure that
   // the container has started successfully:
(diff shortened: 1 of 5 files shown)


dtfs-central-api/src/v1/routes/utilisation-reports-routes.js
```diff
@@ -148,10 +148,10 @@ utilisationReportsRouter.route('/set-status').put(putUtilisationReportStatusCont
 
 /**
  * @openapi
- * /utilisation-reports/reconciliation-details/:reportId:
+ * /:utilisation-reports/reconciliation-details/:reportId:
```
MarRobSoftwire, line 151: Surprised to see this when the other routes dont have a `:` - what does this do?

dtfs-central-api/src/v1/swagger.js
```diff
@@ -28,11 +29,16 @@ const swaggerDefinition = {
       name: 'User',
       description: 'Get and create Portal (BSS/GEF) users. This is only used in the central API.',
```
MarRobSoftwire, line 30: Legacy but do you know what `This is only used in the central API` mean? Is it that these endpoints are only used in tests?

dtfs-central-api/src/generateApp.js
```diff
@@ -18,6 +18,8 @@ SqlDbDataSource.initialize()
 const generateApp = () => {
   const app = express();
 
+  app.use(`/v1/${SWAGGER_ROUTE}`, swaggerRoutes);
```
MarRobSoftwire, line 21: Is this moved so the endpoints aren't behind the API key?

dtfs-central-api/src/v1/routes/utilisation-reports-routes.js
JoshBinns2000: It was just a typo 🙃

dtfs-central-api/src/v1/swagger.js
JoshBinns2000: I have no idea

MarRobSoftwire (approved):

PR #3223, opened 2024-06-07
feat(FN-2902): disable sorting when payments columns are empty

JoshBinns2000:
## Introduction :pencil2:
We want to disable sorting on the premium payments table "Payments received" and "Total payments received" column when the fee record table `paymentsReceived` column contains no values.

- Updates the controller to supply a `enablePaymentsReceivedSorting` flag
- Updates associated tests

### DIFF
--- a/trade-finance-manager-ui/server/controllers/utilisation-reports/utilisation-report-reconciliation-for-report/index.ts
+++ b/trade-finance-manager-ui/server/controllers/utilisation-reports/utilisation-report-reconciliation-for-report/index.ts
@@ -6,6 +6,9 @@ import { PRIMARY_NAVIGATION_KEYS } from '../../../constants';
 import { mapFeeRecordItemsToFeeRecordViewModelItems } from '../helpers';
 import { UtilisationReportReconciliationForReportViewModel } from '../../../types/view-models';
 import { getAndClearAddPaymentFieldsFromRedirectSessionData } from './get-and-clear-add-payment-fields-from-redirect-session-data';
+import { FeeRecordItem } from '../../../api-response-types';
+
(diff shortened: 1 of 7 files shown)


trade-finance-manager-ui/server/controllers/utilisation-reports/utilisation-report-reconciliation-for-report/index.ts
```diff
@@ -6,6 +6,9 @@ import { PRIMARY_NAVIGATION_KEYS } from '../../../constants';
 import { mapFeeRecordItemsToFeeRecordViewModelItems } from '../helpers';
 import { UtilisationReportReconciliationForReportViewModel } from '../../../types/view-models';
 import { getAndClearAddPaymentFieldsFromRedirectSessionData } from './get-and-clear-add-payment-fields-from-redirect-session-data';
+import { FeeRecordItem } from '../../../api-response-types';
+
+const isAtLeastOnePaymentReceivedNonNull = (feeRecords: FeeRecordItem[]): boolean => feeRecords.some(({ paymentsReceived }) => paymentsReceived !== null);
```
ChristianMcc, line 11: Nitpick but can we name this something like:
`feeRecordsHaveAtLeastOnePaymentReceived`

The current name make it sounds a bit like paymentsReceived can be null and that is a payment received rather than us just checking one exists (by checking it's not null)

JoshBinns2000: I'm not sure if I'm misunderstanding what you mean, but each `paymentsReceived` is of type `CurrencyAndAmount | null`, so it can be `null`. Either way, I do think updating the function name makes sense, but the check for it existing would be `!== undefined`, so the choice of `is<...>NonNull` was intentional

ChristianMcc (approved):
LGTM :)

PR #3227, opened 2024-06-10
feat(FN-1617): add facility filter button to premium payments page

francescastocco:
## Introduction :pencil2:
Allow PDC Users to filter fee records by a partial facility ID

- Display an input field, submit & clear button on premium payment page
- On submit pass query param to dtfs-central-api
- When fetching utilisation reports from DB find using `LIKE` facility ID match on fee records when facility ID query param present
- Display errors

Since last review:
- Have added in returning the full group if a fee record matching the filter is part of a group
- The empty state has not been added yet as waiting for design input. Will add as a follow up PR, currently just displays the table with no rows

Screenshots (differ slightly to design to avoid nested forms):
image
image


trade-finance-manager-ui/server/types/view-models/utilisation-report-reconciliation-for-report-view-model.ts
```diff
@@ -30,5 +30,6 @@ export type UtilisationReportReconciliationForReportViewModel = BaseViewModel &
   formattedReportPeriod: string;
   reportId: number;
   feeRecords: FeeRecordViewModelItem[];
-  errorSummary: [ErrorSummaryViewModel] | undefined;
+  addPaymentError: [ErrorSummaryViewModel] | undefined;
+  facilityIdQueryError: [ErrorSummaryViewModel] | undefined;
```
RLavender98, line 34: Rather than make these arrays of only one item could we make them just e.g.: `addPaymentError: ErrorSummaryViewModel | undefined;`?
It made slightly more sense as `errorSummary` as that is an array in all other view models, so might have caused confusion if it were not an array but here I feel it is the other way around.

trade-finance-manager-ui/server/controllers/utilisation-reports/utilisation-report-reconciliation-for-report/index.ts
```diff
@@ -15,9 +16,11 @@ export const getUtilisationReportReconciliationByReportId = async (req: Request,
   const { reportId } = req.params;
 
   try {
-    const { addPaymentErrorSummary, isCheckboxChecked } = getAndClearAddPaymentFieldsFromRedirectSessionData(req);
+    const { addPaymentError, isCheckboxChecked } = getAndClearAddPaymentFieldsFromRedirectSessionData(req);
```
RLavender98, line 19: Have we tested what happens if the user applies a filter, selects some fee records with differing payment currencies and or statuses and then clicks add a payment? Interested to know if the filter query params get persisted or not.

trade-finance-manager-ui/component-tests/utilisation-reports/utilisation-report-reconciliation-for-report.component-test.js
```diff
@@ -79,4 +79,16 @@ describe(page, () => {
...
+    wrapper.expectElement('[data-cy="facility-filter-input"]').toExist();
+  });
+
+  it('should render the facility ID filter submit button', () => {
+    wrapper.expectElement('[data-cy="facility-filter-submit-button"]').toExist();
+  });
+
+  it('should render the facility ID filter clear button', () => {
+    wrapper.expectElement('[data-cy="facility-filter-clear-button"]').toExist();
+  });
```
RLavender98, line 93: I'm not a big fan of tests that just test for existence of a data-cy, fine to select elements using data-cy as long as we are then using them for something or asserting their content etc, e.g. could get the filter input by data-cy and then validate that the value is set to the passed in query param/not initiated when there is no param.

I think also the `params` above actually need updating as the view model type has changed, and I think we could also possibly add some extra tests around the errors now there is more logic happening there inside the template itself.

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-report-reconciliation-details-by-id.controller/index.ts
```diff
@@ -1,36 +1,61 @@
...
 export const getUtilisationReportReconciliationDetailsById = async (req: GetUtilisationReportReconciliationDetailsByIdRequest, res: Response<ResponseBody>) => {
   const { reportId } = req.params;
+  const { facilityIdQuery } = req.query;
 
   try {
+    const feeRecordFindOptions: FindOptionsWhere<UtilisationReportEntity> =
+      facilityIdQuery && REGEX.UKEF_PARTIAL_FACILITY_ID_REGEX.test(facilityIdQuery)
+        ? {
+            feeRecords: {
+              facilityId: Like(`%${facilityIdQuery}%`),
```
RLavender98, line 32: Now that this is even more complicated I think it warrants a custom repo method so the controller itself doesn't need to know so much about the db interaction

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-report-reconciliation-details-by-id.controller/index.ts
```diff
@@ -1,36 +1,61 @@
...
+        ...feeRecordFindOptions,
+      },
     });
+
     if (!utilisationReport) {
-      throw new NotFoundError(`Failed to find a report with id '${reportId}'`);
+      const errorMessage = facilityIdQuery
+        ? `Failed to find a report with id '${reportId}' & facility id containing '${facilityIdQuery}'`
+        : `Failed to find a report with id '${reportId}'`;
+      throw new NotFoundError(errorMessage);
```
RLavender98, line 49: What happens if the report exists but has no fee records matching the facility id search term?

I was expecting something along the lines of the page to still load but the table of fee records to be replaced with a message about there not being any matching the search term, but I can't see that in the template, although I do appreciate there is not an actual design for this attached to the ticket.

RLavender98 (changes requested):
A question around empty state handling and otherwise mostly small comments 😄

trade-finance-manager-ui/templates/utilisation-reports/utilisation-report-reconciliation-for-report.njk
```diff
@@ -10,31 +11,76 @@
 {% endblock %}
 
 {% set premiumPaymentsHtml %}
-  {% if errorSummary %}
+  {% if premiumPaymentFormError or facilityIdQueryError %}
     {{ govukErrorSummary({
       titleText: "There is a problem",
-      errorList: errorSummary,
+      errorList: [premiumPaymentFormError, facilityIdQueryError],
```
JoshBinns2000, line 17: Does nothing strange happen when one of these errors is `undefined`?

JoshBinns2000 (commented):
Looks good so far - just a few small comments

trade-finance-manager-ui/server/middleware/validatePostAddPaymentRequestBody/index.ts
```diff
@@ -11,6 +11,21 @@ import { PremiumPaymentsTableCheckboxId } from '../../types/premium-payments-tab
 
 const isRequestBodyAnObject = (body: unknown): body is object => !body || typeof body === 'object';
 
+const FACILITY_ID_QUERY_REGEX = /facilityIdQuery=(?<facilityIdQuery>\d+)/;
+
+const getFacilityIdQueryFromReferer = (req: Request): string | undefined => {
+  const { referer } = req.headers;
```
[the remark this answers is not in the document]
RLavender98: Ooh I didn't know that would be the url we are coming from, I tried `req.originalUrl` thinking that was the referer but it was the url it was directing to, not the url is came from, I'll give that a go!

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-report-reconciliation-details-by-id.controller/helpers.test.ts
```diff
@@ -9,380 +10,895 @@ import {
...
-
-      jest.mocked(getBankNameById).mockResolvedValue(undefined);
-
-      // Act / Assert
-      await expect(mapUtilisationReportEntityToReconciliationDetails(uploadedReport)).rejects.toThrow(
-        new NotFoundError(`Failed to find a bank with id '${bankId}'`),
-      );
-      expect(getBankNameById).toHaveBeenCalledWith(bankId);
-    });
+    describe('without filters', () => {
```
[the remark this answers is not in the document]
RLavender98: I was referring specifically to the fact that the page allows filtering but I guess that is perhaps not so relevant to the central api endpoint so will update

trade-finance-manager-ui/server/controllers/utilisation-reports/utilisation-report-reconciliation-for-report/validate-facility-id-query.ts
```diff
@@ -0,0 +1,28 @@
+import { Request } from 'express';
+import { asString, isNonEmptyString } from '@ukef/dtfs2-common';
+import { ErrorSummaryViewModel } from '../../../types/view-models';
+import { REGEX } from '../../../constants';
+
+export const validateFacilityIdQuery = (
```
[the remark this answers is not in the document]
RLavender98: I didn't write this, but I see what you mean, will update!

trade-finance-manager-ui/templates/utilisation-reports/utilisation-report-reconciliation-for-report.njk
RLavender98: It doesn't seem to but I would thought something weird might happen also:
Screenshot 2024-07-03 at 11 27 11

trade-finance-manager-ui/server/middleware/validatePostAddPaymentRequestBody/index.ts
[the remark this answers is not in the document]
RLavender98: Ah so when I console log `req.url` it logs `/9/add-payment` which is the url we are headed to not the url we came from

trade-finance-manager-ui/server/controllers/utilisation-reports/utilisation-report-reconciliation-for-report/validate-facility-id-query.ts
```diff
@@ -1,28 +1,18 @@
...
-): {
-  validatedFacilityIdQuery: string | undefined;
-  facilityIdQueryError: ErrorSummaryViewModel | undefined;
-} => {
-  const { facilityIdQuery } = req.query;
-  const { originalUrl } = req;
-  const facilityIdQueryAsString = facilityIdQuery ? asString(facilityIdQuery, 'facilityIdQuery') : undefined;
-  let facilityIdQueryError: ErrorSummaryViewModel | undefined;
-
+export const validateFacilityIdQuery = (facilityIdQuery: string | undefined, originalUrl: string): ErrorSummaryViewModel | undefined => {
```
JoshBinns2000, line 5: The use of `originalUrl` in this function still feels a bit redundant to me. The `originalUrl.includes('?') && originalUrl.includes('facilityIdQuery')` should be captured by `req.query.facilityIdQuery === undefined` because of the way that the request object is generated, so checking the actual `url` shouldn't be required

trade-finance-manager-ui/server/controllers/utilisation-reports/utilisation-report-reconciliation-for-report/index.ts
```diff
@@ -20,12 +20,14 @@ export const getUtilisationReportReconciliationByReportId = async (req: Request,
   const { reportId } = req.params;
 
   try {
-    const { validatedFacilityIdQuery, facilityIdQueryError } = validateFacilityIdQuery(req);
+    const { facilityIdQuery } = req.query;
```
JoshBinns2000, line 23: This can be moved outside of the `try-catch` (just to match how other controllers currently do this)

trade-finance-manager-ui/server/middleware/validatePostAddPaymentRequestBody/index.ts
```diff
@@ -20,6 +35,10 @@ const redirectWithError = (
 ) => {
   req.session.addPaymentErrorKey = addPaymentError;
   req.session.checkedCheckboxIds = checkedCheckboxIds;
+  const facilityIdQuery = getFacilityIdQueryFromReferer(req);
+  if (facilityIdQuery) {
+    return res.redirect(`/utilisation-reports/${reportId}?facilityIdQuery=${facilityIdQuery}`);
+  }
```
JoshBinns2000, line 41: To avoid this `if` block, could we use `axios.getUri` to construct this url? I just checked and, if `facilityIdQuery` is `undefined`, `getUri` will not include it in the url which matches the behaviour we want

JoshBinns2000 (commented):
Just a couple more small comments

trade-finance-manager-ui/server/controllers/utilisation-reports/utilisation-report-reconciliation-for-report/validate-facility-id-query.ts
RLavender98: It's not redundant because if they click filter without typing anything the query param is tacked onto the url but when you read the value it is undefined and in that case we want to display an error message, but not in the case where it is undefined because there is no query param at all because they didn't just click the filter button.

JoshBinns2000 (approved):

PR #3239, opened 2024-06-13
feat(FN-1750): add recorded payments table to add payments page

RLavender98:
## Introduction :pencil2:
Add recorded payments table to add payments page

- Add collapsable recorded payments table to add payments page detailing the date, amount and reference of any previously recorded payments for the selected fee records

## Out of scope ➖ 
I'm going to do one or two follow up PRs covering:
- setting the payment number in the heading above the form
- adding in a check for the payments table in the e2e-tests

## Screenshots
Screenshot 2024-06-13 at 17 38 06
Screenshot 2024-06-13 at 17 38 16


dtfs-central-api/src/repositories/utilisation-reports-repo/utilisation-report.repo.ts
```diff
@@ -154,15 +154,15 @@ export const UtilisationReportRepo = SqlDbDataSource.getRepository(UtilisationRe
    * @param feeRecordIds - The fee record ids to include
    * @returns The utilisation report with the attached fee records
    */
-  async findOneByIdWithFeeRecordsFilteredById(reportId: number, feeRecordIds: number[]): Promise<UtilisationReportEntity | null> {
+  async findOneByIdWithFeeRecordsFilteredByIdWithPayments(reportId: number, feeRecordIds: number[]): Promise<UtilisationReportEntity | null> {
```
JoshBinns2000, line 157: Could we update the docstring for this method?

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-selected-fee-records-details.controller/helpers.ts
```diff
@@ -26,12 +40,16 @@ export const mapToSelectedFeeRecordDetails = async (
...
+  const distinctPaymentsForFeeRecords = selectedFeeRecordEntities[0].payments;
+  const recordedPaymentDetails = distinctPaymentsForFeeRecords.map((paymentEntity) => mapPaymentEntityToSelectedFeeRecordsPaymentDetails(paymentEntity));
   const selectedFeeRecordDetails = selectedFeeRecordEntities.map((feeRecordEntity) => mapFeeRecordEntityToSelectedFeeRecordDetails(feeRecordEntity));
 
   return {
     bank: { name: bankName },
     reportPeriod,
     totalReportedPayments: getTotalReportedPayments(selectedFeeRecordDetails),
     feeRecords: selectedFeeRecordDetails,
+    payments: recordedPaymentDetails ?? [],
```
JoshBinns2000, line 53: I _think_ `recordedPaymentDetails` will always be an array - if you enable the `payments` relation on the query and there are no related payments, then `payments` will be an empty array, so I think `recordedPaymentDetails` will always be an array, so the nullish coalesce operator isn't necessary

libs/common/src/types/selected-fee-records-details.ts
```diff
@@ -1,6 +1,12 @@
 import { CurrencyAndAmount } from './currency';
 import { ReportPeriod } from './utilisation-reports';
 
+export type SelectedFeeRecordsPaymentDetails = {
+  value: CurrencyAndAmount;
```
JoshBinns2000, line 5: Why don't we store the `currency` and `amount` at the top level instead of inside this `value` field?

JoshBinns2000: I think I can see why - it's so that you can call `getFormattedCurrencyAndAmount` directly on the `value` field, right?

trade-finance-manager-ui/component-tests/utilisation-reports/_macros/recorded-payments-for-selected-fee-records-table.component-test.js
```diff
@@ -0,0 +1,52 @@
...
+
+  it('should render the table headings', () => {
+    const wrapper = getWrapper();
+    wrapper.expectElement(`thead th`).toHaveCount(3);
+    wrapper.expectElement(`thead th:contains("Date received")`).toExist();
+    wrapper.expectElement(`thead th:contains("Amount received")`).toExist();
+    wrapper.expectElement(`thead th:contains("Payment reference")`).toExist();
+  });
+
+  it('should render one row per fee record displaying the date received, value of payment and reference if provided', () => {
```
JoshBinns2000, line 44: Could we also add an assertion that the number of rows rendered is correct (the below only asserts that rows 1 and 2 have the correct data; not that there are only two which are rendered)

trade-finance-manager-ui/server/controllers/utilisation-reports/add-payment/index.ts
```diff
@@ -49,7 +56,15 @@ const extractFormValuesFromRequestBody = (requestBody: AddPaymentRequestBody): A
 
 const renderAddPaymentPage = (res: Response, context: AddPaymentViewModel) => res.render('utilisation-reports/add-payment.njk', context);
 
-const mapToSelectedReportedFeesDetailsViewModel = (selectedFeeRecordData: SelectedFeeRecordsDetails): SelectedReportedFeesDetailsViewModel => {
+const mapToRecordedPaymentDetailsViewModel = (payment: SelectedFeeRecordsPaymentDetailsResponse): RecordedPaymentDetailsViewModel => {
+  return {
+    reference: payment.reference,
+    value: getFormattedCurrencyAndAmount(payment.value),
```
JoshBinns2000, line 62: Could we call this `formattedCurrencyAndAmount` to match other places where we use this?

trade-finance-manager-ui/templates/utilisation-reports/_macros/recorded-payments-for-selected-fee-records-table.njk
```diff
@@ -0,0 +1,41 @@
+{% from "govuk/components/accordion/macro.njk" import govukAccordion %}
+{% macro render(params) %}
```
JoshBinns2000, line 2: Other files have an empty line between imports and the macro, so could we add that here too please?

trade-finance-manager-ui/styles/accordion.scss
```diff
@@ -0,0 +1,5 @@
+.accordion-without-show-all {
+  .govuk-accordion__controls {
+    display: none
```
JoshBinns2000, line 3: Should there be a semi-colon here?

JoshBinns2000 (commented):
Nice job - just some minor comments

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-selected-fee-records-details.controller/helpers.ts
RLavender98: Ah yes nice, I think I added it at one point to try and get rid of a bug but clearly wasn't needed or things changed 😄

libs/common/src/types/selected-fee-records-details.ts
RLavender98: I think I just liked it, in my head they are very linked and so this just made sense. I do see now that I consider it more that perhaps it isn't quite as clear cut as I thought so I actually don't think it matters too much either way, but it does make it nice for the formatting as you point out.

JoshBinns2000: My main concern with the word `value` was semantic as it feels a bit strange to me to access the "amount" by using `details.value.amount`, but I also think it can make sense as is. With respect to formatting, I did a bit of experimenting and found that any type which overlaps with the `CurrencyAndAmount` type can be used in `getFormattedCurrencyAndAmount`. For example, the following is valid:

```typescript
type SomeDetails = {
  currency: Currency;
  amount: number;
  id: number;
  ... other fields
};

const details: SomeDetails = {
  currency: 'GBP',
  amount: 100,
  id: 1,
  ...
};

const formattedCurrencyAndAmount = getFormattedCurrencyAndAmount(someDetails); // works fine
```

Just to try and avoid making things more confusing that they already are with all these fee record and payment types, I think it may be better to keep the `currency` and `amount` fields at the top level of this object, but I'm not sure whether or not it does actually make things clearer

JoshBinns2000 (approved):
LGTM

PR #3243, opened 2024-06-17
feat(FN-1714 & FN-1721): add check keying data page

JoshBinns2000:
## Introduction :pencil2:
We want to add the following functions to the "Generate keying data" button:
- when clicked
  - if there are matching fee record payment groups
    - go to `/utilisation-reports/:reportId/check-keying-data`
    - display matching fee record payment groups
    - display button to generate keying sheet data
    - after clicking, redirects to `/utilisation-reports/:reportId#keying-sheet`
  - if there are no matching fee record payment groups
    - display the error

The actual generation of the keying data is beyond the scope of these tickets.

- Updates the button
- Adds the new template
- Adds the new routes with controllers
- Adds the error handling through the request session when the button is pressed
- Adds a skeleton endpoint to tfm-api and dtfs-central-api which is called but does not actually generate keying data
- Adds tests

image
image


trade-finance-manager-ui/templates/utilisation-reports/check-keying-data.njk
```diff
@@ -0,0 +1,59 @@
...
+
+    <table class="govuk-table premium-payments-table" data-module="moj-sortable-table" data-cy="check-keying-data-table">
+      <thead class="govuk-table__head">
+        <tr class="govuk-table__row">
+          <th scope="col" class="govuk-table__header">Facility ID</th>
+          <th scope="col" class="govuk-table__header">Exporter</th>
+          <th scope="col" class="govuk-table__header govuk-table__header--numeric">Reported fees</th>
+          <th scope="col" class="govuk-table__header govuk-table__header--numeric">Reported payments</th>
+          <th scope="col" class="govuk-table__header govuk-table__header--numeric">Payments received</th>
+          <th scope="col" class="govuk-table__header">Status</th>
```
RLavender98, line 32: The design shows these as sortable, possibly need to raise on the ticket what 'standard' sorting means payments received column which can have multiple values? (see my other comments re grouping which obviously impacts the ability to sort other columns)

trade-finance-manager-ui/templates/utilisation-reports/_macros/check-keying-data-table-row.njk
```diff
@@ -0,0 +1,56 @@
...
+        {% endif %}
+      </td>
+
+      <td class="govuk-table__cell{{ " no-border" if not isLastFeeRecordInGroup }}" data-sort-value="{{ displayStatus }}">
+        {{ feeRecordStatus.render({
+          status: status,
+          displayStatus: displayStatus
+        }) if isFirstFeeRecordInGroup else "" }}
+      </td>
+    </tr>
```
RLavender98, line 54: I think the ticket is pretty clear that in the payments received bullet that we shouldn't be grouping fee records into multiple rows? Has there been a change to this that has not been recorded on the ticket?

trade-finance-manager-ui/server/controllers/utilisation-reports/check-keying-data/index.ts
```diff
@@ -0,0 +1,47 @@
...
+
+const getNumberOfFacilities = (feeRecordPaymentGroups: FeeRecordPaymentGroup[]): number =>
+  feeRecordPaymentGroups.reduce((total, { feeRecords }) => feeRecords.length + total, 0);
+
+export const postCheckKeyingData = async (req: Request, res: Response) => {
+  const { userToken, user } = asUserSession(req.session);
+  const { reportId } = req.params;
+
+  try {
+    const { bank, reportPeriod, feeRecordPaymentGroups } = await api.getUtilisationReportReconciliationDetailsById(reportId, userToken);
```
RLavender98, line 19: See my other comment re grouping but I think we might want to consider a separate endpoint for this fetching specifically the matched fee records in a shape closer to what the front end needs

trade-finance-manager-ui/templates/utilisation-reports/check-keying-data.njk
JoshBinns2000: I've made every column sortable aside from payments received (due to them being listed in specific cases)

trade-finance-manager-ui/server/controllers/utilisation-reports/check-keying-data/index.ts
JoshBinns2000: I've added the new endpoint which I think may have doubled the size of this PR 🙃

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-fee-records-to-key.controller/helpers.ts
```diff
@@ -0,0 +1,53 @@
...
+  return feeRecordPaymentEntityGroups.reduce((feeRecordsToKey, group) => {
+    const { feeRecords, payments } = group;
+    if (feeRecords.length === 1) {
+      // If one fee record to many payments, list all the payments
+      return [...feeRecordsToKey, getFeeRecordsToKeyWithPayments(feeRecords[0], payments)];
+    }
+    // Otherwise, fee record payments received should equal fee record reported payments
+    return [...feeRecordsToKey, ...feeRecords.map((feeRecord) => getFeeRecordToKey(feeRecord))];
+  }, [] as FeeRecordToKey[]);
+};
```
RLavender98, line 53: I think we could do with some better naming here, was rather confusing to read at first. I appreciate the attempt to use the key word 'map' less, but here I think it does apply quite well, since we pass in one object, manipulate it's fields into a relatively similar shape and return it. I would suggest something along the lines:
- `getFeeRecordsToKeyFromFeeRecordEntities` -> `mapToFeeRecordsToKey`
- `getFeeRecordsToKeyWithPayments` -> `mapToFeeRecordToKeyWithAllPaymentsListed`
- `getFeeRecordToKey` -> `mapToFeeRecordToKeyWithReportedPaymentsAsSolePayment`

And could have a method named `mapToFeeRecordToKey` which takes a fee record entity and a payments currency and amount array, and we do the deciding of what to set for payments separately outside. (this common method could either be called be the separate methods for setting the full list or setting the reported payments or we could do that directly in the main function and not have two separate ones.)

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-fee-records-to-key.controller/index.ts
```diff
@@ -0,0 +1,46 @@
...
+
+    return res.status(HttpStatusCode.Ok).send({
+      id,
+      bank: {
+        id: bankId,
+        name: bankName,
+      },
+      reportPeriod,
+      feeRecords: feeRecordsToKey,
+    });
```
RLavender98, line 37: Is this return type getting typed checked against the expected response shape?

trade-finance-manager-api/src/v1/api-response-types/utilisation-report-reconciliation-details-response-body.ts
```diff
@@ -1,6 +1,6 @@
 import { IsoDateTimeStamp, ReportPeriod, UtilisationReportReconciliationStatus, FeeRecordStatus, CurrencyAndAmount } from '@ukef/dtfs2-common';
 
-type FeeRecordItem = {
+export type FeeRecordItem = {
```
RLavender98, line 3: Can we move this to a common file since it is used in multiple responses now please?

trade-finance-manager-ui/component-tests/utilisation-reports/_macros/check-keying-data-table-row.component-test.js
```diff
@@ -0,0 +1,160 @@
...
+    wrapper.expectElement(`${rowSelector} td:has(ul.payments-list)`).toExist();
+    wrapper.expectElement(`${rowSelector} td:has(ul.payments-list)`).hasClass(govukNumericCellClassName);
+
+    wrapper.expectElement(`${rowSelector} ul.payments-list li`).toHaveCount(3);
+    wrapper.expectElement(`${rowSelector} li:contains("GBP 90.91")`).toExist();
+    wrapper.expectElement(`${rowSelector} li:contains("GBP 50.10")`).toExist();
+    wrapper.expectElement(`${rowSelector} li:contains("GBP 314.59")`).toExist();
+  });
+
+  it('renders the display status badge in the first table row', () => {
```
RLavender98, line 146: Given this returns a single row now this test can be renamed

RLavender98 (commented):
Looks good just a couple of small comments 😄

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-fee-records-to-key.controller/index.ts
JoshBinns2000: Yes - the first argument of the `Response` generic is the `ResponseBody`

trade-finance-manager-api/src/v1/api-response-types/utilisation-report-reconciliation-details-response-body.ts
JoshBinns2000: I'm also going to rename this to `FeeRecord` as the `Item` part feels a bit redundant now that we have more clear definitions around things like fee record payment groups

dtfs-central-api/src/types/utilisation-reports.ts
```diff
@@ -108,11 +83,6 @@ export type NewPaymentDetails = {
   reference?: string;
 };
 
-export type FeeRecordToKey = FeeRecordItem & {
-  paymentsReceived: CurrencyAndAmount[];
-  status: FeeRecordStatus;
-};
-
 export type UtilisationReportWithFeeRecordsToKey = {
```
RLavender98, line 86: I've only just processed this name sorry for not noting earlier, but I would be tempted to name this just 'FeeRecordsToKey' or something similar with the 'FeeRecordsToKey' being the key "thing", and then renaming the id field to 'reportId'. The reason being that this type isn't really a report with something added, it's mostly just some fee records mapped into an interesting shape with a few bonus pieces of information about the report really.  'With' names can get quite confusing as you get more of them and have to keep checking which one you should be using at any given time and what the difference is between them.

RLavender98 (commented):
Just one minor comment I noticed upon second reading but otherwise all looks good 😄

RLavender98 (approved):

PR #3244, opened 2024-06-18
feat(FN-2731): remove sql data migrator

JoshBinns2000:
## Introduction :pencil2:
We want to remove the sql data migrator.

- Removes the `utils/data-migration/sql` directory


ChristianMcc:
In the rejigging of the package.json files within data-migration etc. have we just lost the mongo-sql-migrator:utilisation-reports script somewhere? thought it was missing here but looks liek it might just not exist aha

ChristianMcc (dismissed):

ChristianMcc (approved):
LGTM :)

PR #3247, opened 2024-06-19
feat(FN-1720): delete payment endpoint

RLavender98:
## Introduction :pencil2:
Add an endpoint to dtfs central api for deleting a payment

- Adds a new endpoint `DELETE utilisation-reports/:reportId/payments/:paymentId`
- Payment can only be deleted if utilisation report is in RECONCILIATION_IN_PROGRESS status
- Payment can only be deleted if the fee records it is attached to are in the MATCH or DOES_NOT_MATCH status
- Checks and sets the fee record statuses after deletion


JoshBinns2000 (commented):
Looks good - comments for me to resolve

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/delete-payment/delete-payment.event-handler.ts
```diff
@@ -0,0 +1,55 @@
...
+  { paymentId, transactionEntityManager, requestSource }: DeletePaymentEventPayload,
+): Promise<UtilisationReportEntity> => {
+  const payment = await transactionEntityManager.findOne(PaymentEntity, { where: { id: paymentId }, relations: { feeRecords: true } });
+
+  if (!payment) {
+    throw new NotFoundError(`Failed to find a payment with id: ${paymentId}`);
+  }
+
+  const linkedFeeRecords = payment.feeRecords;
+  if (linkedFeeRecords.length === 0) {
```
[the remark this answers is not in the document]
JoshBinns2000: Discussed with @ChristianMcc - although we probably won't encounter this situation, there isn't really any harm in allowing for the deletion of an unlinked payment

dtfs-central-api/src/helpers/fee-record-matching.test.ts
```diff
@@ -0,0 +1,119 @@
+import { Currency, FeeRecordEntityMockBuilder, PaymentEntityMockBuilder, UtilisationReportEntityMockBuilder } from '@ukef/dtfs2-common';
+import { feeRecordsAndPaymentsMatch } from './fee-record-matching';
+
+describe('fee-record-matching', () => {
+  describe('feeRecordsAndPaymentsMatch', () => {
+    it('returns true when the payments attached to the fee records have the same total payments', () => {
```
ChristianMcc, line 6: I think this test name is slightly lacking (only referencing total payments not total fees paid to ukef) but then when I typed out a more in depth one it also didn't read well so I've resigned myself to leave this comment and just say this is fine

dtfs-central-api/src/v1/routes/utilisation-reports-routes.js
```diff
@@ -217,6 +219,44 @@ utilisationReportsRouter
...
+ *       200:
+ *         description: OK
+ *       404:
+ *         description: Not Found
+ *       500:
+ *         description: Internal Server Error
+ */
+utilisationReportsRouter
+  .route('/:reportId/payments/:paymentId')
+  .delete(
```
ChristianMcc, line 252: some lovely API validation in the middleware here, not the first time but just a shout out that's it's great we've introduced more of this!

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/delete-payment/delete-payment.event-handler.test.ts
```diff
@@ -0,0 +1,321 @@
...
+    await expect(
+      handleUtilisationReportDeletePaymentEvent(report, {
+        transactionEntityManager: mockEntityManager,
+        paymentId: 1,
+        requestSource: aDbRequestSource(),
+      }),
+    ).rejects.toThrow(NotFoundError);
+  });
+
+  it('removes payment if there are no linked fee records', async () => {
```
ChristianMcc, line 51: I spoke to Josh briefly about this, I agree with having this in but it did make me think:
The ticket that has us editing what fee records go with a payment - we should check if payments can be 'orphaned' and unattached from all fee records and how we would then find them (e.g. the payments tab if it's going to exist on the individual bank report page)

ChristianMcc (approved):
Really nice work! Left a few comments but none requiring action just some thoughts :)

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/delete-payment/delete-payment.event-handler.test.ts
RLavender98: Ooh interesting, yeah I wonder if we also want to link payments directly to a report...

JoshBinns2000 (approved):

PR #3249, opened 2024-06-20
feat(FN-1720): add unit tests for zod schemas

JoshBinns2000:
## Introduction :pencil2:
We want to add some test coverage for the newly added zod schemas.

- Adds unit tests
- Updates the `MongoObjectIdSchema` schema


AlexBramhill (approved):
Great work -- happy for this to be merged. I think as we increase schemas we can look to commoise some of these tests before these test files get too big/many!

PR #3253, opened 2024-06-21
feat(dtfs2-6966): add audit logs to update deal status endpoint

AlexBramhill:
## Introduction :pencil2:
We need to add audit details to portal endpoints

This PR captures a missed update deal status endpoint

### DIFF
--- a/dtfs-central-api/src/v1/controllers/portal/deal/update-deal-status.controller.js
+++ b/dtfs-central-api/src/v1/controllers/portal/deal/update-deal-status.controller.js
@@ -6,9 +6,6 @@ const { findOneDeal } = require('./get-deal.controller');
 const { mongoDbClient: db } = require('../../../../drivers/db-client');
 
 const updateDealStatus = async ({ dealId, status, existingDeal, auditDetails }) => {
-  if (!ObjectId.isValid(dealId)) {
-    return { status: 400, message: 'Invalid Deal Id' };
-  }
   const dealsCollection = await db.getCollection(MONGO_DB_COLLECTIONS.DEALS);
 
   const previousStatus = existingDeal.status;
@@ -34,8 +31,6 @@ const updateDealStatus = async ({ dealId, status, existingDeal, auditDetails }) 
 
   return findAndUpdateResponse.value;
 };
-
-exports.updateDealStatus = updateDealStatus;
 
 exports.updateDealStatusPut = async (req, res) => {
   const {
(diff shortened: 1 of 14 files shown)


trade-finance-manager-api/src/v1/controllers/update-portal-deal-status.js
```diff
@@ -31,7 +31,7 @@ const updatePortalDealStatus = async (deal) => {
   }
 
   if (dealType === CONSTANTS.DEALS.DEAL_TYPE.BSS_EWCS) {
-    await api.updatePortalBssDealStatus(_id, status);
+    await api.updatePortalBssDealStatus({ dealId: _id, status, auditDetails });
   } else if (dealType === CONSTANTS.DEALS.DEAL_TYPE.GEF) {
     await api.updatePortalGefDealStatus(_id, status);
```
ChristianMcc, line 36: did I miss this in a different PR or is this missing?

ChristianMcc: Or is it just not technically called here, I feel like maybe we should add it in here anyway even if this route is unused as it's a bit weird otherwise

ChristianMcc (commented):
This one is a bit weird, we only update bss deals etc. but there is that weird code path in the middle which is unused but I think we should cover it anyway

trade-finance-manager-api/src/v1/controllers/update-portal-deal-status.js
AlexBramhill: Yeah -- this actually goes off to GEF. Will raise another PR for this

ChristianMcc (dismissed):
LGTM, thanks! :)

dtfs-central-api/src/v1/controllers/portal/deal/update-deal-status.controller.js
```diff
@@ -1,58 +1,67 @@
...
+  const {
+    params: { id: dealId },
+    body: { status, auditDetails },
+  } = req;
+
+  if (!ObjectId.isValid(dealId)) {
+    return res.status(400).send({ status: 400, message: 'Invalid Deal Id' });
+  }
+
+  try {
```
ChristianMcc, line 45: Not saying to do this here at all*
how long do you think it would take to turn this into middleware that just runs this code? Is it just probably copy pasting a middleware onto every route that has this then deleeting this everywhere?

ChristianMcc (approved):
LGTM! a question but nothing to action on this PR

dtfs-central-api/src/v1/controllers/portal/deal/update-deal-status.controller.js
AlexBramhill: I really think it would be nice to
1. create audit details on entering portal-api/tfm-api in middleware
2. create a validate audit details/transform into audit record in dtfs-central-api
Then it should just be a job of passing things around.

Realistically, I would put bullet 2 at around 5 points, but I wouldn't be surprised if it were a 3.
I think bullet 1 above at around the same, as we should already be constructing these at `res/req` functions but this is not always the case (sometimes we construct the audit details in tfm-api after receiving a request from portal-api, which im not keen on)

PR #3257, opened 2024-06-25
feat(FN-1714): update query when getting reports

JoshBinns2000:
## Introduction :pencil2:
The original query would return `null` if there are no fee records matching the supplied status. This PR makes it so that the `feeRecords` array will instead be empty.

- Updates the `UtilisationReportRepo`

### DIFF
--- a/dtfs-central-api/src/repositories/utilisation-reports-repo/utilisation-report.repo.ts
+++ b/dtfs-central-api/src/repositories/utilisation-reports-repo/utilisation-report.repo.ts
@@ -1,6 +1,7 @@
 import { SqlDbDataSource } from '@ukef/dtfs2-common/sql-db-connection';
 import { UtilisationReportEntity, ReportPeriod, FeeRecordStatus } from '@ukef/dtfs2-common';
 import { Not, Equal, FindOptionsWhere, LessThan, In } from 'typeorm';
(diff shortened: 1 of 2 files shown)


dtfs-central-api/src/repositories/utilisation-reports-repo/utilisation-report.repo.ts
```diff
@@ -174,15 +175,21 @@ export const UtilisationReportRepo = SqlDbDataSource.getRepository(UtilisationRe
...
-    return await UtilisationReportRepo.findOne({
+    const report = await this.findOne({
+      where: { id: reportId },
+    });
+
+    if (!report) {
+      return null;
+    }
+
+    report.feeRecords = await FeeRecordRepo.find({
```
RLavender98, line 186: I see that this works but it feels a little suboptimal as we have to be careful everywhere. I wonder if setting the entity to default to an empty array would work and is a better solution. e.g. I haven't tried it but I'm hoping this could work:
```
@OneToMany(() => FeeRecordEntity, (feeRecord) => feeRecord.report, {
   cascade: ['insert', 'update'],
})
feeRecords: FeeRecordEntity[] = [];
```
Would you mind trying that if you have't already?

PR #3262, opened 2024-07-01
feat(FN-1720): add confirm delete payment page

JoshBinns2000:
## Introduction :pencil2:
We want to add a page to allow the user to confirm that they want to delete the selected payment.

- Adds the page
- Adds a new query to the `GET /v1/utilisation-reports/:reportId/payment/:paymentId` endpoint which decides whether or not to include the fee record related fields in the response
  - This is added as a payment can be attached to many fee records, and sending all that information in the response body when it isn't needed is unnecessary
- Updates api tests to test for the new query paramater

image

### DIFF
--- a/trade-finance-manager-ui/component-tests/utilisation-reports/confirm-delete-payment.component-test.js
+++ b/trade-finance-manager-ui/component-tests/utilisation-reports/confirm-delete-payment.component-test.js
@@ -1,0 +1,52 @@
+const pageRenderer = require('../pageRenderer');
+
+const page = '../templates/utilisation-reports/confirm-delete-payment.njk';
+const render = pageRenderer(page);
+
+describe(page, () => {
+  const getListRow = (key, value) => ({ key: { text: key }, value: { text: value } });
+
+  const getWrapper = ({ paymentSummaryListRows } = { paymentSummaryListRows: [] }) => render({ paymentSummaryListRows });
+
+  it('renders the page heading', () => {
+    const wrapper = getWrapper();
+
+    wrapper.expectText('h1').toRead('Are you sure you want to delete the payment?');
+  });
+
+  it('renders a summary list item for each supplied row', () => {
+    const paymentSummaryListRows = [
+      getListRow('First item', 'First value'),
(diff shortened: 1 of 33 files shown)


trade-finance-manager-ui/component-tests/utilisation-reports/confirm-delete-payment.component-test.js
```diff
@@ -0,0 +1,52 @@
...
+
+    wrapper.expectElement('div.govuk-radios__item').toHaveCount(2);
+
+    const yesRadioLabelSelector = 'label.govuk-radios__label:contains("Yes")';
+    wrapper.expectElement(yesRadioLabelSelector).toExist();
+    wrapper.expectInput(`div.govuk-radios__item:has(${yesRadioLabelSelector}) input`).toHaveValue('yes');
+
+    const noRadioLabelSelector = 'label.govuk-radios__label:contains("No")';
+    wrapper.expectElement(noRadioLabelSelector).toExist();
+    wrapper.expectInput(`div.govuk-radios__item:has(${noRadioLabelSelector}) input`).toHaveValue('no');
```
RLavender98, line 44: I think I have a slight preference for this style of selector for the radios: `input[type="radio"][value="no"]` as seen in the add-payment component tests, simply because we aren't relying on gov uk classes, and more on the expected resulting html. Neither is perfect and either could break if the library changes their implementation so I don't mind if you prefer and want to keep it like this, it just feels slightly weird to me generally to depend on class names.

RLavender98 (approved):
Looks good, nice job 😄

trade-finance-manager-ui/component-tests/utilisation-reports/confirm-delete-payment.component-test.js
JoshBinns2000: I did consider doing the `input[type="radio"][value="no"]` check, but I wanted to check that the input was inside the same div as the matching label. Using the class name was just a way for me to do that without adding `data-cy` attributes, but I do agree in general that using class names is not a great way of selecting components

PR #3265, opened 2024-07-01
feat(FN-NONE): enable tfm-ui typescript component tests

JoshBinns2000:
## Introduction :pencil2:
As we start to add more types for template render view models, it would be nice to be able to write component tests using those more strict view model types. A pre-requisite for this is updating the rendering helper functions and assertions to be typescript.

- Updates the `pageRenderer`, `componentRenderer` and `assertions.js` to typescript
- Updates all existing imports


trade-finance-manager-ui/component-tests/pageRenderer.ts
```diff
@@ -0,0 +1,11 @@
+import { load } from 'cheerio';
+import { assertions } from './assertions';
+import configureNunjucks from '../server/nunjucks-configuration';
+
+const nunjucks = configureNunjucks({}) as { render: (pageLocation: string, params: object) => string };
```
JoshBinns2000, line 5: I looked around for any declaration files for the `nunjucks` package, but I couldn't find anything. As a result, putting in this explicit type assertion seems like the best way to do it

RLavender98: From a quick google this came up: https://www.npmjs.com/package/@types/nunjucks did you consider this/can you tell if this is accurate and might do we what we want?

Otherwise if we leave this as is, then I think we should at least add a comment explaining this and also maybe linking to wherever would be the most useful place to look if we ever wanted to extend this type to cover more of the returned utilities?

RLavender98 (commented):
Thanks for doing this, will be really useful!

trade-finance-manager-ui/component-tests/pageRenderer.ts
JoshBinns2000: Not sure why that didn't appear when I looked up the types, but those are what we need - thanks!

RLavender98 (approved):

RLavender98 (approved):

RLavender98 (approved):

PR #3283, opened 2024-07-01
feat(FN-1718): add edit payment event handler

JoshBinns2000:
## Introduction :pencil2:
We want to add the functionality to edit a payment.

- Adds a new `PATCH /v1/utilisation-reports/:reportId/payment/:paymentId` endpoint
- Adds a new payload validation helper
- Adds tests


dtfs-central-api/src/v1/routes/utilisation-reports-routes.js
```diff
@@ -399,11 +399,55 @@ utilisationReportsRouter.route('/:reportId/fee-records-to-key').get(validation.s
...
+ *             type: object
+ *             properties:
+ *               paymentAmount:
+ *                 type: number
+ *               datePaymentReceived:
+ *                 type: string
+ *                 format: date
+ *               paymentReference:
+ *                 type: string
+ *                 required: false
```
RLavender98, line 433: This is interesting, because when not provided it updates the field to null from a value if one already exists, which might not be the default expectation for a PATCH when a field is not required, what do you think? But then given how JSON works if undefined it just won't be in the body right so I do think this type is correct? Can we add a description or something to make it clear that not providing one unsets it?

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/edit-payment/edit-payment.event-handler.ts
```diff
@@ -0,0 +1,49 @@
...
+): Promise<UtilisationReportEntity> => {
+  payment.update({
+    amount: paymentAmount,
+    dateReceived: datePaymentReceived,
+    reference: paymentReference,
+    requestSource,
+  });
+  await transactionEntityManager.save(PaymentEntity, payment);
+
+  const feeRecordsAndPaymentsMatch = await feeRecordsMatchAttachedPayments(feeRecords, transactionEntityManager);
```
RLavender98, line 31: Have we tested manually that these fee records have the new payment amount value and not a reference to the old value? (I remember having to be careful with deleting to fetch the array of payments again after deleting, although this could be different as is a value changing not a full payment no longer existing!)

RLavender98: I feel like I might be missing something because I also don't see how the fee records have attached payments or where we are calling the new repo method that has been added to the utilisation report repo...

dtfs-central-api/src/repositories/utilisation-reports-repo/utilisation-report.repo.ts
```diff
@@ -219,4 +219,26 @@ export const UtilisationReportRepo = SqlDbDataSource.getRepository(UtilisationRe
...
+
+  /**
+   * Finds a utilisation report with the supplied id with the
+   * attached fee records which are attached to the payment
+   * with the supplied id
+   * @param reportId - The report id
+   * @param paymentId - The payment id
+   * @returns The utilisation report with attached fee records
+   */
+  async findOneByIdWithFeeRecordsFilteredByPaymentId(reportId: number, paymentId: number): Promise<UtilisationReportEntity | null> {
```
RLavender98, line 231: Where is this used?

dtfs-central-api/api-tests/v1/utilisation-reports/patch-payment.api-test.ts
```diff
@@ -0,0 +1,164 @@
...
+
+    expect(paymentEntity.lastUpdatedByTfmUserId).toBe(tfmUserId);
+    expect(paymentEntity.lastUpdatedByPortalUserId).toBeNull();
+    expect(paymentEntity.lastUpdatedByIsSystemUser).toBe(false);
+
+    const reportEntity = (await SqlDbHelper.manager.findOneBy(UtilisationReportEntity, { id: reportId }))!;
+
+    expect(reportEntity.lastUpdatedByTfmUserId).toBe(tfmUserId);
+    expect(reportEntity.lastUpdatedByPortalUserId).toBeNull();
+    expect(reportEntity.lastUpdatedByIsSystemUser).toBe(false);
```
RLavender98, line 162: As we are checking db fields could we also add checks for the fee record status changing by perhaps doing a case where we go from match to does not match or the other way around? That check existing and passing would probably answer a couple of my other comments 😄

RLavender98 (commented):
Mostly just a few questions around understanding how we are correctly updating the fee record status

dtfs-central-api/src/v1/routes/utilisation-reports-routes.js
JoshBinns2000: After a small but of thinking, I reckon we should make this field required and have it be either `null` or `string`. If it is `null`, we set the field to `undefined` (which can be handled by the zod schema). Doing that probably makes the intention clearer

trade-finance-manager-ui/server/controllers/utilisation-reports/edit-payment/index.test.ts
```diff
@@ -444,7 +444,48 @@ describe('controllers/utilisation-reports/edit-payment', () => {
...
+        req.body['paymentDate-year'] = '2024';
+
+        const paymentReference = 'A payment reference';
+        req.body.paymentReference = paymentReference;
+
+        const parsedFormValues: ParsedEditPaymentFormValues = {
+          paymentAmount,
+          datePaymentReceived,
+          paymentReference,
+        };
```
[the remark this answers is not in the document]
JoshBinns2000: I used to prefix things with `expected` and then stopped because I started to think it was redundant in many cases. In this case though I agree that `expected` makes sense to use

dtfs-central-api/src/v1/routes/utilisation-reports-routes.js
JoshBinns2000: As a side note for future readers, there is a reason why we used `undefined` instead of `null` for the typeorm entities - SQL Server processes the javascript `null` value as an `object`, which can't be inserted into SQL Server tables, therefore throwing an error. If you want to set a column to `NULL`, you need to use `undefined` in SQL Server

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/edit-payment/edit-payment.event-handler.ts
JoshBinns2000: I had tested that this works correctly locally, but you are right that it could do with some test coverage in the api test (and in e2e tests in an upcoming PR). The `feeRecordsMatchAttachedPayments` function loads the attached payments when it is called (hence the `await`), so the payments it checks against are the up to date once

dtfs-central-api/src/repositories/utilisation-reports-repo/utilisation-report.repo.ts
JoshBinns2000: Not sure - will remove

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/edit-payment/edit-payment.event-handler.ts
RLavender98: Ahh nice, I failed to realise that helper was fetching the payments fresh and not just using the ones currently attached. That makes sense!

RLavender98 (approved):
LGTM!

PR #3285, opened 2024-07-02
feat(FN-3011): remove sql ledger with migration scripts

JoshBinns2000:
## Introduction :pencil2:
We want to remove ledger tables.

- Adds two migration scripts - one for feature/dev environment where all migrations have been run and one for prod/staging environments where only the first migration has been run
- Updates the `libs/common` scripts
  - Adds a `db:remove-ledger` script which should be removed later
  - Reverts the change made to the `db:reset` command
- Updates existing migrations and entities*

*the original way that these migrations/entities were written were done as a compromise to work with the ledger tables. As ledger tables are now gone, these have been rewritten


libs/common/src/sql-db-connection/migrations/1717746522387-PaymentTableUpdateAmountColumns.ts
```diff
@@ -1,69 +1,35 @@
...
   public async up(queryRunner: QueryRunner): Promise<void> {
-    // Create the new columns with the required data-type and set to nullable
     await queryRunner.query(`
-            ALTER TABLE "Payment"
-            ADD "amount" decimal(14, 2) NULL
-        `);
-    await queryRunner.query(`
-            ALTER TABLE "Payment"
-            ADD "reference" nvarchar(255) NULL
+            EXEC sp_rename "[Payment].[amountReceived]", "amount", "COLUMN"
```
ChristianMcc, line 8: Twice I've thought about leaving a comment saying can we do this with a new migration, but both times I've gone to delete. I think this is fine to change this one as we can't release. I think we just need to keep this PR open until we know what the next release will be - what we don't want to do is release old code which has this old migration and then we won't run these new changes because the migration has technically been run.
later this week i should know when we can run these scripts and if it can be end of next week

libs/common/src/sql-db-entities/payment/payment.entity.ts
```diff
@@ -13,19 +13,19 @@ export class PaymentEntity extends AuditableBaseEntity {
   /**
    * The currency the payment was made in
    */
-  @Column({ type: 'nvarchar' })
+  @Column({ nullable: false, type: 'nvarchar' })
```
ChristianMcc, line 16: jsut to double check with your changes to the migration, if you run all migrations from a fresh db we don't have any pending migration changes now?

utils/data-migration/remove-sql-ledger/copy-and-remove-ledger.feature.sql
```diff
@@ -0,0 +1,360 @@
+-- ENABLE ROLLBACK TRANSACTION ON ERROR
```
ChristianMcc, line 1: I think I'd prefer to reset feature and run migrations, as that's closer to what will happen on staging + prod (minimal changes to a migration that exists then a new migration). Happy to leave this in for now anyway

ChristianMcc (commented):
Good stuff thank you for this! As discussed in person you've already actioned the checking migrations are in the right state from a fresh DB :)

ChristianMcc (dismissed):
LGTM if we can merge this asap

ChristianMcc (approved):

PR #3288, opened 2024-07-03
feat(FN-1718): add edit payment e2e tests

JoshBinns2000:
## Introduction :pencil2:
We want to add some e2e tests for the TFM edit payment journey.

- Adds the e2e test

- Adds some extra assertions to the `add-payment.spec.js` test

### DIFF
--- a/e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/add-payment.spec.js
+++ b/e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/add-payment.spec.js
@@ -1,4 +1,5 @@
 import {
+  FEE_RECORD_STATUS,
   FeeRecordEntityMockBuilder,
   PaymentEntityMockBuilder,
   UTILISATION_REPORT_RECONCILIATION_STATUS,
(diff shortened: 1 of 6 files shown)


e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/add-payment.spec.js
```diff
@@ -68,23 +68,23 @@ context('PDC_RECONCILE users can add a payment to a report', () => {
   });
 
   it('should render the selected fee record details', () => {
-    pages.utilisationReportAddPaymentPage.selectedReportedFeesDetailsTable().contains('11111111');
-    pages.utilisationReportAddPaymentPage.selectedReportedFeesDetailsTable().contains('Exporter 1');
-    pages.utilisationReportAddPaymentPage.selectedReportedFeesDetailsTable().contains('JPY 100');
-    pages.utilisationReportAddPaymentPage.selectedReportedFeesDetailsTable().contains('GBP 50');
+    pages.utilisationReportAddPaymentPage.selectedReportedFeesDetailsTable().contains('11111111').should('exist');
```
RLavender98, line 71: On reading the difference between these, I found that we might actually be able to do `pages.utilisationReportAddPaymentPage.selectedReportedFeesDetailsTable().should('contain', '11111111')` which I think I like the most?

RLavender98 (dismissed):

e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/add-payment.spec.js
JoshBinns2000: I tried to find that in the assertions sections of the cypress documentation, but couldn't seem to. On a second look, it seems like it is at the bottom of the list of assertions which is why I missed it 🙃

RLavender98 (approved):

PR #3291, opened 2024-07-04
feat(FN-3052): add screenreader labels to premium payments table checkboxes

RLavender98:
- Added aria-labels to the premium payments table checkboxes

### DIFF
--- a/trade-finance-manager-ui/server/controllers/utilisation-reports/helpers/reconciliation-for-report-helper.ts
+++ b/trade-finance-manager-ui/server/controllers/utilisation-reports/helpers/reconciliation-for-report-helper.ts
@@ -57,6 +57,11 @@ const getCheckboxIdForFeeRecordsAndStatus = (feeRecords: FeeRecord[], status: Fe
   return `feeRecordIds-${feeRecordIdList}-reportedPaymentsCurrency-${reportedPaymentsCurrency}-status-${status}`;
 };
 
+const getCheckboxAriaLabel = (feeRecords: FeeRecord[]): string => {
+  const feeRecordFacilityIdList = feeRecords.map(({ facilityId }) => facilityId).join(' ');
+  return `Select ${feeRecordFacilityIdList}`;
+};
+
 export const mapFeeRecordPaymentGroupsToFeeRecordPaymentGroupViewModelItems = (
   feeRecordPaymentGroups: FeeRecordPaymentGroup[],
   isCheckboxChecked: (checkboxId: string) => boolean = () => false,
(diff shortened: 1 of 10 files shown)


trade-finance-manager-ui/templates/utilisation-reports/_macros/premium-payments-table.njk
```diff
@@ -38,9 +38,9 @@
         {{ tableHeader({ headerText: 'Total payments received', isNumericColumn: true, enableSorting: enablePaymentsReceivedSorting, ariaSort: 'none' }) }}
         {{ tableHeader({ headerText: 'Status', isNumericColumn: false, enableSorting: true, ariaSort: 'none' }) }}
         {% if userCanEdit %}
-          <th scope="col" class="govuk-table__header">
+          <td scope="col" class="govuk-table__header">
```
RLavender98, line 41: If this is a `th` the select all checkbox aria-label gets reader out for all the items in the column, but 'Select all' describes the specific checkbox not each item in the column

trade-finance-manager-ui/server/controllers/utilisation-reports/helpers/reconciliation-for-report-helper.ts
```diff
@@ -57,6 +57,11 @@ const getCheckboxIdForFeeRecordsAndStatus = (feeRecords: FeeRecord[], status: Fe
   return `feeRecordIds-${feeRecordIdList}-reportedPaymentsCurrency-${reportedPaymentsCurrency}-status-${status}`;
 };
 
+const getCheckboxAriaLabel = (feeRecords: FeeRecord[]): string => {
+  const feeRecordFacilityIdList = feeRecords.map(({ facilityId }) => facilityId).join(' ');
```
sw-mattw, line 61: Not sure on this but if we're joining facility ID's like this, is easily distinguishable in the screen reader audio output that they are separate numbers? As in if we have facility ID's 123 and 456, is it clear that these are not a single facility ID?

RLavender98: It pauses between the two as if they are separate words, when I added columns it started reading them as 'one million eight hundred and thirty thousand.....' and so on which was I guess also ok, except it was inconsistent between ones where there is only one id which it decided to read as 'one eight three zero' so for consistency removed the commas.

sw-mattw: Cool thanks for clarifying

sw-mattw (approved):

PR #3293, opened 2024-07-05
feat(FN-1722): add generate keying data event handler

JoshBinns2000:
## Introduction :pencil2:
We want to add the event handler to actually generate keying data for `MATCH` fee records.

### Note
Generating keying data results in all fee records which are at the `MATCH` state to be converted to the `READY_TO_KEY`. There are three amounts which are to be computed when keying data is generated called
- `fixedFeeAdjustment`
- `premiumAccrualBalanceAdjustment`
- `principalBalanceAdjustment`

These amounts are nullable and are only to be calculated if all fee records with a specific facility ID are at the `MATCH` state. For example, if there are three fee records with facility ID `"12345678"`, two of which are at `MATCH` status and the other being at `TO_DO` and we generate keying data, the `MATCH` fee records should have their status updated to `READY_TO_KEY` and the amounts listed above should *not* be computed (should be `NULL`). If the `TO_DO` fee record is then moved to the `MATCH` status and keying data is generated again, all three fee records should now have those fields be non-null.

This example case is covered by the `post-keying-data.api-test.ts` api test.

- Adds the event handlers
- Adds new nullable columns to the `FeeRecord` table related to the keying sheet*
- Adds unit and api tests

*these values are always set to 10 for now by the event handler as computing these values is the responsibility of other tickets. We choose 10 instead of 0 or `null` because we want to be able to see which fee records have had values generated in the frontend (see the next PR).

- Makes the fee record seeder only set the fee record status to `TO_DO`


dtfs-central-api/src/v1/controllers/utilisation-report-service/post-keying-data.controller/index.ts
```diff
@@ -1,24 +1,50 @@
...
+  reqBody: PostKeyingDataPayload;
+}>;
+
+export const postKeyingData = async (req: PostKeyingDataRequest, res: Response) => {
   const { reportId } = req.params;
+  const { user } = req.body;
 
   try {
-    const utilisationReport = await UtilisationReportRepo.findOneByIdWithFeeRecordsFilteredByStatus(Number(reportId), ['MATCH']);
+    const feeRecordsWhichCanBeKeyed = await FeeRecordRepo.findByReportIdAndStatusesWithReport(Number(reportId), ['MATCH', 'READY_TO_KEY']);
```
JoshBinns2000, line 20: This variable name is misleading and I am planning on updating it

dtfs-central-api/src/v1/controllers/utilisation-report-service/post-keying-data.controller/index.ts
```diff
@@ -1,24 +1,50 @@
...
 
-    const { feeRecords: matchingFeeRecords } = utilisationReport;
-    if (matchingFeeRecords.length === 0) {
-      throw new NoMatchingFeeRecordsError();
+    const utilisationReport = feeRecordsWhichCanBeKeyed[0].report;
+    if (!utilisationReport) {
+      throw new Error('The report on the found fee records is not defined');
     }
 
+    const generateKeyingDataDetails = await getGenerateKeyingDataDetails(feeRecordsWhichCanBeKeyed);
```
RLavender98, line 31: I feel like this is a little opaque in terms of the naming since `GenerateKeyingDataDetails` doesn't really tell us what it is until we dig down into, I tried to think of a better name but failed but realised I almost expect this to happen inside the event handler? Then we could just pass the utilisationReportStateMachine a list of FeeRecords and at least contain this slightly odd type to where it is crucial?

dtfs-central-api/src/v1/controllers/utilisation-report-service/post-keying-data.controller/index.ts
RLavender98, line 1: This is a bit of an edge case so hoping we are happy with it, but if between loading the check keying data screen and clicking generate, another user updates another fee record to match status by adding a payment say, then this will move that fee record to ready to key even though it wasn't checked. @ChristianMcc do you know if is that would be acceptable?

dtfs-central-api/src/v1/controllers/utilisation-report-service/post-keying-data.controller/helpers.test.ts
```diff
@@ -0,0 +1,116 @@
...
+
+    beforeEach(() => {
+      findByStatusesSpy.mockRejectedValue(new Error('Some error'));
+    });
+
+    afterEach(() => {
+      jest.resetAllMocks();
+    });
+
+    it('returns each supplied fee record with generateKeyingData set to true when there are no facility ids which are not at the TO_DO or DOES_NOT_MATCH status', async () => {
```
RLavender98, line 26: Too many negatives, should be: `when there are no facility ids which are at the TO_DO or DOES_NOT_MATCH status`

dtfs-central-api/src/v1/controllers/utilisation-report-service/post-keying-data.controller/helpers.test.ts
```diff
@@ -0,0 +1,116 @@
...
+          case secondFacilityId:
+            expect(generateKeyingData).toBe(false);
+            return;
+          default:
+            expect(generateKeyingData).toBe(true);
+        }
+      });
+    });
+
+    it('only sets generateKeyingData to true for one of the fee records at the MATCH status', async () => {
```
RLavender98, line 100: `with the same facilityId`

dtfs-central-api/src/services/state-machines/fee-record/event-handlers/keying-data-generated/keying-data-generated.event-handler.ts
```diff
@@ -0,0 +1,23 @@
...
+  generateKeyingData: boolean;
+  requestSource: DbRequestSource;
+};
+
+export type FeeRecordKeyingDataGeneratedEvent = BaseFeeRecordEvent<'KEYING_DATA_GENERATED', KeyingDataGeneratedEventPayload>;
+
+export const handleFeeRecordKeyingDataGeneratedEvent = async (
+  feeRecord: FeeRecordEntity,
+  { transactionEntityManager, generateKeyingData, requestSource }: KeyingDataGeneratedEventPayload,
+): Promise<FeeRecordEntity> => {
```
RLavender98, line 16: I wonder if this should actually be named `GENERATE_KEYING_DATA` since it is this handler that does that actual generation, it's not just some side-effect status update like `PAYMENT_DELETED` is? That does make it a bit tricky for the payload property, `generateKeyingData` but I almost wonder if that parameter should actually just be named `isLastFeeRecordForFacility` and then this function owns knowing that being the last fee record for a facility or not corresponds to whether we calculate those specific fields or not?

RLavender98 (changes requested):
A few changes needed but overall a nice job given the complexity!

dtfs-central-api/src/services/state-machines/fee-record/event-handlers/keying-data-generated/keying-data-generated.event-handler.ts
RLavender98: Whilst the computation is complex, the concept is simple, so I do think `isLastFeeRecordForFacility` will work as it is just renaming a parameter no?

dtfs-central-api/src/v1/controllers/utilisation-report-service/post-keying-data.controller/helpers.ts
```diff
@@ -0,0 +1,34 @@
...
+import { FeeRecordRepo } from '../../../../repositories/fee-record-repo';
+
+export type GenerateKeyingDataDetails = {
+  feeRecord: FeeRecordEntity;
+  generateKeyingData: boolean;
+}[];
+
+const getFeeRecordFacilityIdsWhichCannotBeKeyed = async (): Promise<Set<string>> => {
+  const feeRecordsWhichCannotBeKeyed = await FeeRecordRepo.findByStatuses(['TO_DO', 'DOES_NOT_MATCH']);
+  return feeRecordsWhichCannotBeKeyed.reduce((facilityIds, { facilityId }) => facilityIds.add(facilityId), new Set<string>());
```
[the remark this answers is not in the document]
JoshBinns2000: I think this was just an oversight as I was doing this - we should only be checking fee records on the actual report as (at least, in theory), there could exist a fee record for a different report with the same facility id which has not been moved to `MATCH` yet, which would cause this to behave incorrectly

dtfs-central-api/src/services/state-machines/fee-record/event-handlers/keying-data-generated/keying-data-generated.event-handler.ts
JoshBinns2000: Discussed offline: to remove any implication about ordering but make the intention more clear, we have decided to go with `isFinalFeeRecordForFacility`

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/generate-keying-data/generate-keying-data.event-handler.ts
```diff
@@ -1,29 +1,55 @@
...
+      if (processedFacilityIds.has(facilityId)) {
+        return stateMachine.handleEvent({
+          type: 'GENERATE_KEYING_DATA',
+          payload: {
+            transactionEntityManager,
+            isFinalFeeRecordForFacility: false,
+            requestSource,
+          },
+        });
+      }
```
RLavender98, line 45: I think this would be slightly cleaner if you kept the mapping fee records to objects pairing them with the isFinalFeeRecordForFacility and then mapped through that to call the fee record state machine. It is another loop through, but I think it would be more readable, and I think having a type for that pairing is less opaque and confusing when it would just be in this file and not exported? I'm not too set on this though so I'll leave it up to you.

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/generate-keying-data/generate-keying-data.event-handler.test.ts
```diff
@@ -37,50 +36,168 @@ describe('handleUtilisationReportGenerateKeyingDataEvent', () => {
...
+          requestSource,
+        },
+      });
+    });
+  });
+
+  describe('when there are multiple fee records with the same facility id', () => {
+    const facilityId = '12345678';
+
+    it('calculates keying data only for the first fee record in the payload', async () => {
```
RLavender98, line 87: This might be a tad pedantic, but there is no requirement for it to be the first, that is an implementation choice, so (leaving the body of the test the same) could we rename this to 'calculates keying data for only one of the fee records in the payload with that facility id' please?

RLavender98 (commented):
Just a few rename requests, otherwise all looking good!

dtfs-central-api/src/v1/controllers/utilisation-report-service/post-keying-data.controller/index.ts
ChristianMcc: Yep let's ignore that edge case , in general only a single person is working on a single report at a time so this shouldn't ever happen

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/generate-keying-data/generate-keying-data.event-handler.ts
```diff
@@ -1,29 +1,55 @@
...
+
+      processedFacilityIds.add(facilityId);
       return stateMachine.handleEvent({
-        type: 'KEYING_DATA_GENERATED',
+        type: 'GENERATE_KEYING_DATA',
         payload: {
           transactionEntityManager,
-          generateKeyingData,
+          isFinalFeeRecordForFacility: !facilityIdsWhereKeyingDataCannotBeCalculated.has(facilityId),
```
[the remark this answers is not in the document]
JoshBinns2000: I had a different idea for how we can do this to fit with the new idea of a "final" fee record - let me know what you think

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/generate-keying-data/generate-keying-data.event-handler.ts
JoshBinns2000: Sure - I try to avoid looping through twice, but if you think it makes it clearer then I don't mind

RLavender98 (approved):
LGTM!

PR #3294, opened 2024-07-05
feat(FN-1811): changes to bank reports page

RLavender98:
## Introduction :pencil2:
When the feature flag for the payment reconciliation is enabled we don't want users to mark reports as done manually.

- Add total facilities reported column when feature flag is on
- Remove download column when feature flag is on
- Remove mark as buttons and selection checkboxes when feature flag is on

## Screenshots
Feature flag off:
Screenshot 2024-07-05 at 17 39 41
Screenshot 2024-07-05 at 17 39 57

Feature flag on:
Screenshot 2024-07-05 at 17 39 19
Screenshot 2024-07-05 at 17 41 42


dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-reports-reconciliation-summary.controller/reconciliation-summary-item-mapper.ts
```diff
@@ -0,0 +1,32 @@
...
+const getCountOfDistinctFacilities = (feeRecords: FeeRecordEntity[]): number => {
+  const distinctFacilityIds = new Set(feeRecords.map((feeRecord) => feeRecord.facilityId));
+  return distinctFacilityIds.size;
+};
+
+export const mapReportToSummaryItem = (bank: Bank, report: UtilisationReportEntity): UtilisationReportReconciliationSummaryItem => {
+  const reportReceived = report.status !== 'REPORT_NOT_RECEIVED';
+  const totalFacilitiesReported = getCountOfDistinctFacilities(report.feeRecords);
+  const totalFeesReported = report.feeRecords.length;
+  const reportedFeesLeftToReconcile = totalFeesReported - getCountOfReconciledFeeRecords(report.feeRecords);
```
RLavender98, line 17: It has just occurred to me that this will need to be behind the feature flag since we don't update the fee record statuses on the 'mark as' actions so I will add that in on Monday

trade-finance-manager-ui/templates/utilisation-reports/_macros/report-reconciliation-table-manual-reconciliation.njk
```diff
@@ -0,0 +1,107 @@
...
+            </td>
+            <td class="govuk-table__cell" data-sort-value="{{ summaryItem.totalFeesReported }}">
+              {{ summaryItem.totalFeesReported }}
+            </td>
+            <td class="govuk-table__cell" data-sort-value="{{ summaryItem.reportedFeesLeftToReconcile }}">
+              {% if summaryItem.status === 'RECONCILIATION_COMPLETED' %}
+                0
+              {% else %}
+                {{ summaryItem.totalFeesReported }}
+              {% endif %}
```
RLavender98, line 59: This might be an actual terms change but it felt wrong for a report to be allowed to be complete and have fees left to reconcile? That said, I'm not sure what we are going to do for old reports when we turn the feature flag on, @ChristianMcc do you know if we have a ticket to cover not allowing users to go to premium reports page etc for reports that were set to reconciled manually from before we had the reconciliation journey?

RLavender98: (Also I haven't written tests for this yet but will do so on Monday)

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-reports-reconciliation-summary.controller/reconciliation-summary-item-mapper.ts
RLavender98, line 17: For reports which are marked as complete this will always be the full number of reported fees, which matches the current behaviour, but it seems weird to me that we would display a 'reconciled' report as having fees left to reconcile so I was think we could add a check for if it is reconciled and then just set to 0 (also saves us the calculation for actual reconciled reports i.e. not manual).

@ChristianMcc this has made me think, what are we doing going forwards about reports that were marked as completed? Will be want them to be links to the individual report page in the find reports by year? (and when we first turn the FF on possibly on the bank reports page also) I think currently they will be linked and then actions will show as being possible except they will fail due to the state machine checking report status.

ChristianMcc: Good point I hadn't thought about this before! Going to be quite hard to identify them as we don't have any kind of flag on the report itself, just the state of it's fee records.
In an ideal world I think mark as done behaviour would be updated to set everything as reconciled as well as the utilisation report status (then we would have to run a migration on the old ones that don't have this behaviour), however this feels like way too much overhead for something we won't care about much.

The only place we will be able to get to them is the previous reports/search reports page right?

RLavender98: Yes (except for the remainder of the month we turn the feature on if earlier in the month they marked a report as complete) (and then a user could also directly navigate to the individual bank report page if they knew the id and put it in the url directly but I don't think we should worry about this given we expect the api to not accept any calls?)

dtfs-central-api/src/v1/swagger-definitions/utilisation-reports/utilisation-reports-reconciliation-summary.js
```diff
@@ -24,6 +24,8 @@
  *       dateUploaded:
  *         type: string
  *         example: 2021-01-01T00:00:00.000Z
+ *       totalFacilitiesReported:
+ *         type: number
```
JoshBinns2000, line 28: I think the optional fields on the `UtilisationReportReconciliationSummaryItem` type need to be updated in these docs, but I'm not sure what they should be updated to. I think `required` tag is only for request values, and `nullable: true` doesn't exactly work as the field wouldn't be `null` and would instead just not be present. What do you think?

trade-finance-manager-ui/server/types/view-models/utilisation-reports-view-model.ts
```diff
@@ -0,0 +1,22 @@
+import { IsoMonthStamp } from '@ukef/dtfs2-common';
+import { BaseViewModel } from './base-view-model';
+import { UtilisationReportReconciliationSummaryItem } from '../utilisation-reports';
+
+export type SummaryItemViewModel = UtilisationReportReconciliationSummaryItem & {
```
JoshBinns2000, line 5: Seeing as we are exporting these types, could we make the names a bit more specific than just `Summary<Thing>`?

JoshBinns2000 (commented):
Nice job - only a few tiny comments

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-reports-reconciliation-summary.controller/helpers.ts
```diff
@@ -13,7 +13,7 @@ import {
 import { UtilisationReportReconciliationSummary, UtilisationReportReconciliationSummaryItem } from '../../../../types/utilisation-reports';
 import { UtilisationReportRepo } from '../../../../repositories/utilisation-reports-repo';
 import { getAllBanks } from '../../../../repositories/banks-repo';
-import { mapReportToSummaryItem } from '../helper';
+import { mapReportToSummaryItem } from './reconciliation-summary-item-mapper';
```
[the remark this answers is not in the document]
RLavender98: None of the other controllers have nested helpers folders, so I'm not particularly inclined too unless is there a pre-established pattern elsewhere that we move them to a folder if there are more than one?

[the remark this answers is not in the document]
JoshBinns2000: I am about to add this pattern in the `get-utilisation-report-reconciliation-details-by-id.controller` because of the complexity of that helper, hence the suggestion. The reason I'd prefer that pattern to having this multiple file setup is that the concept of a `helpers.ts` file living at the same level as another file which contains something functioning as a helper seems counter intuitive to me

dtfs-central-api/src/v1/swagger-definitions/utilisation-reports/utilisation-reports-reconciliation-summary.js
RLavender98: So I think the red stars indicate required, so if anything we might need to be putting required on more fields?
Screenshot 2024-07-08 at 15 44 58

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-reports-reconciliation-summary.controller/index.test.ts
```diff
@@ -13,7 +13,7 @@ import {
 } from '@ukef/dtfs2-common';
 import { ObjectId } from 'mongodb';
 import { GetUtilisationReportsByBankIdAndYearRequest, getUtilisationReportsReconciliationSummary, getUtilisationReportSummariesByBankIdAndYear } from './index';
-import { generateReconciliationSummaries } from './helpers';
+import { generateReconciliationSummaries } from './helpers/reconciliation-summary-generator';
```
JoshBinns2000, line 16: I think this should just come from `'./helpers'`

JoshBinns2000 (commented):
Just one comment on an import

dtfs-central-api/src/v1/swagger-definitions/utilisation-reports/utilisation-reports-reconciliation-summary.js
JoshBinns2000: Interesting, although it seems strange to me to tell a user that the field we are returning to them is "required". I guess many of our swagger docs aren't accurate with respect to this optional behaviour anyway, so let's just leave this for now

JoshBinns2000 (dismissed):
LGTM

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-reports-reconciliation-summary.controller/reconciliation-summary-item-mapper.ts
RLavender98: Resolving this thread as PR is ready, will follow up offline on if there are any actions to be taken.

JoshBinns2000 (approved):

PR #3297, opened 2024-07-08
feat(FN-2622): automatically match fee records with zero payments

RLavender98:
## Introduction :pencil2:
Fee records with zero payments should move to match status automatically when report is uploaded.

- When a fee record has 'fees paid to ukef for the period' equal to zero then the status of the created fee record should be 'MATCH'
- When a fee record has 'fees paid to ukef for the period' with a non-zero amount then the status of the created fee record should be 'TO_DO'

image


JoshBinns2000 (approved):
LGTM - nice job

PR #3300, opened 2024-07-09
feat(FN-1849): add table for tolerance data and seed zero threshold tolerances

RLavender98:
## Introduction :pencil2:
We need a way to store tolerances to that we can have a tolerance per currency for when we match payments.

- Add a new Table 'Tolerance'
- Seed zero threshold tolerances for each of the currencies we support


sw-mattw (approved):

ChristianMcc (approved):
Nice :)

PR #3301, opened 2024-07-10
feat(DTFS2-7280): Typescript compilation check during CICD

abhi-markan:
## Introduction :pencil2:

* Added `validate:all` script to `sca.yml`, which will run following NPM sub-scripts:
   * `lint:all`: ESLint check
   * `prettier:all`: Prettier check
   * `typecheck-all`: Typescript compilation check
   * `validate:yml`: Validate all YML file syntax.

* Added `--pretty` flag to all `tsc` check.
* Dependencies updates


.github/workflows/sca.yml
```diff
@@ -119,3 +119,23 @@ jobs:
         uses: fossas/fossa-action@main
         with:
           api-key: ${{ secrets.FOSSA_API_KEY }}
+
+  # 5. Validation
+  validation:
```
ChristianMcc, line 124: Are we doubling up the linting etc. now by doing it both here and in lint.yml as validate:all does all of them now? Not sure if we just bin lint.yml or split them out at all. I guess validation step is a bit vague for all the steps now happening within here? The output might be quite messy so I'm a bit worried about reading the output of errors if it's one huge step

abhi-markan: Good point, have removed the `validate:all` from `sca.yml` and updated with `type-check:all` and `validate:yml` in `lint.yml` GHA file.

ChristianMcc (approved):
LGTM :)

PR #3302, opened 2024-07-10
feat(DTFS2-NONE): reset process.env correctly in tests

MarRobSoftwire:
## Introduction :pencil2:
When modifying the `process.env` file in tests, its not always correctly reset

Shallow clone the original process.env at the start of every test, so it's not reassigned unintentionally

### DIFF
--- a/dtfs-central-api/src/cron-scheduler-jobs/create-utilisation-reports/index.test.ts
+++ b/dtfs-central-api/src/cron-scheduler-jobs/create-utilisation-reports/index.test.ts
@@ -13,7 +13,7 @@ jest.mock('@ukef/dtfs2-common', () => ({
   getCurrentReportPeriodForBankSchedule: jest.fn(),
 }));
 
-const originalProcessEnv = process.env;
+const originalProcessEnv = { ...process.env };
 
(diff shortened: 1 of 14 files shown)


dtfs-central-api/src/cron-scheduler-jobs/create-utilisation-reports/index.test.ts
```diff
@@ -13,7 +13,7 @@ jest.mock('@ukef/dtfs2-common', () => ({
   getCurrentReportPeriodForBankSchedule: jest.fn(),
 }));
 
-const originalProcessEnv = process.env;
+const originalProcessEnv = { ...process.env };
 
 describe('scheduler/jobs/create-utilisation-reports', () => {
   afterEach(() => {
```
BethThomas141, line 19: Is it worth keeping the `afterEach`'s consistent across files with  `process.env = originalEnv` as opposed to the second spread?  (You've done it some places, I think you just missed it in this file and a few others! :))

BethThomas141 (approved):
Great spot, just a small consistency request!

BethThomas141 (approved):
LGTM! 🚀

PR #3304, opened 2024-07-11
fix(acbs-durable-functions): resolve durable functions container startup error

abhi-markan:
## Introduction :pencil2:
Durable functions `acbs` Azure Functions container was failing due to lack of `TS` transpile to `JS`.

* Webpack installation could scope creep the resolution will be time intensive, as a swift resolution constant have been introduced to `acbs-functions` as JS.
* Removed `@ukef/dtfs2-common` package dependency.

* Dockerfile: NPM cache clean has been added post installation.
* All `"@ukef/dtfs2-common` package version has been explicitly set to specific version `1.0.0`.

### DIFF
--- a/azure-functions/acbs-function/Dockerfile
+++ b/azure-functions/acbs-function/Dockerfile
@@ -14,5 +14,6 @@ WORKDIR /home/site/wwwroot
 COPY package.json .
 
 RUN npm i
+RUN npm cache clean --force
 
 COPY . /home/site/wwwroot
--- a/azure-functions/acbs-function/api.js
+++ b/azure-functions/acbs-function/api.js
@@ -12,7 +12,9 @@
 
 require('dotenv').config();
 const axios = require('axios');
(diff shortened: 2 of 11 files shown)


azure-functions/acbs-function/Dockerfile
```diff
@@ -14,5 +14,6 @@ WORKDIR /home/site/wwwroot
 COPY package.json .
 
 RUN npm i
+RUN npm cache clean --force
```
MarRobSoftwire, line 17: Out of interest, why is this needed?

MarRobSoftwire (dismissed):

azure-functions/acbs-function/Dockerfile
abhi-markan: This is to reclaim the disk space, and requires `--force` flag.
Reference: https://docs.npmjs.com/cli/v7/commands/npm-cache

avaitonis (approved):
Approved

PR #3313, opened 2024-07-15
feat(FN-1724): keying sheet principal adjustment

JoshBinns2000:
## Introduction :pencil2:
We want to add the data required to calculate the keying sheet principal adjustment.

- Adds a new `FacilityUtilisationData` table
- Uses the `FeeRecord.facilityId` field to link to the `FacilityUtilisationData.id` primary key as a foreign key
- Updates the api tests to work with the new foreign key requirement
- Updates the utilisation report upload logic to add the requirement that the facility id provided in the report csv data must exist in the facility utilisation data table
- Adds api tests for the utilisation report upload
- Fixes e2e tests

- Updates portal e2e tests to use the `NODE_TASKS` constant rather than the portal commands


libs/common/src/sql-db-connection/migrations/1721040044234-AddFacilityUtilisationDataTable.ts
```diff
@@ -0,0 +1,44 @@
...
+                "lastUpdatedAt" datetime2 NOT NULL CONSTRAINT "DF_5e17cd4417568f3be4c469c70cb" DEFAULT getdate(),
+                "lastUpdatedByPortalUserId" nvarchar(255),
+                "lastUpdatedByTfmUserId" nvarchar(255),
+                "lastUpdatedByIsSystemUser" bit NOT NULL CONSTRAINT "DF_c394215c914eb1852c617c54bbb" DEFAULT 0,
+                CONSTRAINT "PK_789e99214c8f2bf186c167d4dfa" PRIMARY KEY ("id")
+            )
+        `);
+    await queryRunner.query(`
+            INSERT INTO "FacilityUtilisationData" (id)
+            SELECT DISTINCT "facilityId" FROM "FeeRecord"
```
JoshBinns2000, line 20: This line is added to make data backwards compatible. We will be creating a separate script to check that the fee records in staging and prod do marry up to facility ids in the mongo database before running this migration, so it should be fine to run this in all environments by that point

libs/common/src/sql-db-entities/facility-utilisation-data/facility-utilisation-data.entity.ts
```diff
@@ -0,0 +1,25 @@
+import { Entity, OneToMany, PrimaryColumn } from 'typeorm';
+import { AuditableBaseEntity } from '../base-entities';
+import { FeeRecordEntity } from '../fee-record';
+import { MonetaryColumn } from '../custom-columns';
+
+@Entity('FacilityUtilisationData')
+export class FacilityUtilisationDataEntity extends AuditableBaseEntity {
```
ChristianMcc, line 7: After thinking a bit about this, I think it might be useful for us to store report period on this as well so that we know what month/time this data is for. We could assume it's always for the most recent report period behind the one we're doing but it would also make our data in the data lake etc. make more sense downstream.
What do you think?

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/report-uploaded/report-uploaded.event-handler.ts
RLavender98, line 1: I might have missed something, but how are we updating the utilisation so we have the latest reports value to use in the next reports calculation?

dtfs-central-api/api-tests/v1/utilisation-reports/get-selected-fee-record-details.api-test.ts
```diff
@@ -13,25 +14,54 @@ import { testApi } from '../../test-api';
...
 describe('GET /v1/utilisation-reports/:id/selected-fee-records-details', () => {
   const bankId = '123';
   const bank: Bank = { ...aBank(), id: bankId, name: 'Test bank' };
 
   const reportId = 1;
+
+  const facilityUtilisationData = FacilityUtilisationDataEntityMockBuilder.forId('000123').build();
+
   const reportPeriod = aReportPeriod();
+  const utilisationReport = UtilisationReportEntityMockBuilder.forStatus('RECONCILIATION_IN_PROGRESS')
```
RLavender98, line 36: I think things have gotten slightly messed up merging or rebasing here as I think the set up for these was moved into the tests

libs/common/src/sql-db-connection/migrations/1721227404776-AddFacilityUtilisationDataTable.ts
```diff
@@ -0,0 +1,85 @@
...
+            ALTER COLUMN "facilityId" nvarchar(10) NOT NULL
+        `);
+
+    /**
+     * Commands to populate the FacilityUtilisation data with required
+     * data so that the foreign key constraint can be added. This query
+     * sets the FacilityUtilisation report period to the oldest report
+     * period associated with the fee record with the inserted facility
+     * id.
+     */
```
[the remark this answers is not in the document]
RLavender98: I wonder for existing reports do we actually want to set the utilisation to the value of the previous report if it exists?

dtfs-central-api/api-tests/v1/utilisation-reports/delete-payment.api-test.ts
```diff
@@ -21,6 +25,9 @@ describe('DELETE /v1/utilisation-reports/:reportId/payment/:paymentId', () => {
   const tfmUser = aTfmUser();
   const tfmUserId = tfmUser._id.toString();
 
+  const facilityId = '11111111';
+  const facilityUtilisationData = FacilityUtilisationDataEntityMockBuilder.forId(facilityId).build();
```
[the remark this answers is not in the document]
JoshBinns2000: The fee record entity mock builder does create one, but this is just so that the lookup works correctly. We don't want cascade inserts to be enabled for fee records on facility utilisation data as we want to ensure fee records with invalid facility ids don't get uploaded, and part of the safety we get around that comes from the foreign key constraint. However, I do think it will get annoying very quickly that we need to do this, so I am thinking a good compromise here may be to update the `sql-db-helper` so that it handles inserting facility utilisation data where it doesn't exist for the fee records it needs to insert. What do you think?

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/report-uploaded/report-uploaded.event-handler.ts
JoshBinns2000: That comes as part of some later work as far as I know

dtfs-central-api/api-tests/v1/utilisation-reports/put-keying-data-mark-as-to-do.api-test.ts
```diff
@@ -9,16 +15,22 @@ console.error = jest.fn();
 describe('/v1/utilisation-reports/:reportId/keying-data/mark-as-to-do', () => {
   const getUrl = (reportId: number | string) => `/v1/utilisation-reports/${reportId}/keying-data/mark-as-to-do`;
 
+  const facilityUtilisationData = FacilityUtilisationDataEntityMockBuilder.forId('11111111').build();
```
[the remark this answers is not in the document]
JoshBinns2000: I believe this [reply](https://github.com/UK-Export-Finance/dtfs2/pull/3313#discussion_r1684282632) addresses this comment as well

dtfs-central-api/api-tests/v1/utilisation-reports/get-selected-fee-record-details.api-test.ts
JoshBinns2000: Hm very possibly. I'll try to set it back to what is currently on main

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/report-uploaded/report-uploaded.event-handler.ts
RLavender98: Hm the only ticket I can see that feels applicable is [FN-2055](https://ukef-dtfs.atlassian.net/browse/FN-2055) but those ACs just say that there should be a filed for 'current utilisation' and not what that should be and also seems to overlap with some of this work. @ChristianMcc do you have a clear picture of which ticket the work involved in updating the utilisation value should be done under?

dtfs-central-api/api-tests/v1/utilisation-reports/delete-payment.api-test.ts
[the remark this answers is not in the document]
RLavender98: Ah I see... I guess I'm not completely convinced by the need to not have cascade inserts given we are validating existence. And anyway surely if we try and insert a fee record without a facility utilisation and an invalid facility id then the insert will fail and it has nothing to do with a cascade insert?

Also, how are we getting the facility ids in there for new facilities going forwards? I am guessing that insertion just going to happen completely separately from utilisation report uploads?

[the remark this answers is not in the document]
RLavender98: I do accept turning on cascade inserts for test data is possibly a bad idea though. As for saving in sql-db-helper, I'm torn, I think if we do we should be explicit in the method name and have one that does it by default and one that doesn't. On the other hand for most PRs it will hopefully be a small thing to add, like the overhead for creating a report so if we don't want cascade inserts then I think this is not so bad.

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/report-uploaded/report-uploaded.event-handler.ts
ChristianMcc: Excellent call - it's on my to-do list to get this populated as the exact mechanic I haven't got nailed down yet. There was one way Andy had suggested but im not 100% convinced by it yet. I'll make sure this is in place before I go on A/L!

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/report-uploaded/report-uploaded.event-handler.ts
```diff
@@ -34,6 +42,14 @@ export const handleUtilisationReportReportUploadedEvent = async (
   });
   await transactionEntityManager.save(UtilisationReportEntity, report);
 
+  const reportCsvDataFacilityIds = reportCsvData.map(({ 'ukef facility id': facilityId }) => facilityId);
+  const allReportCsvDataFacilityIdsExist = (
+    await Promise.all(reportCsvDataFacilityIds.map((facilityId) => transactionEntityManager.existsBy(FacilityUtilisationDataEntity, { id: facilityId })))
```
JoshBinns2000, line 47: Remove this check (is in another ticket and should check against the tfm collection)

libs/common/src/sql-db-connection/migrations/1721227404776-AddFacilityUtilisationDataTable.ts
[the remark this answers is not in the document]
RLavender98: Resolving this as as discussed on a call:
- on production we will be replacing this data with accurate data from some other source as yet to be determined
- for local dev and test envs we will come up with something soon after merging this to allow us to have a good range of facility ids to upload against

dtfs-central-api/api-tests/v1/utilisation-reports/delete-payment.api-test.ts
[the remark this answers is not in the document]
RLavender98: Discussed on call and agreed it should be safe enough to have cascade inserts and actually depending on future behaviour implementation decision there is a chance it could be necessary in mainline code anyway

dtfs-central-api/src/services/state-machines/fee-record/event-handlers/generate-keying-data/generate-keying-data.event-handler.test.ts
```diff
@@ -124,9 +131,11 @@ describe('handleFeeRecordGenerateKeyingDataEvent', () => {
       expect(feeRecord.premiumAccrualBalanceAdjustment).toBe(10);
     });
 
-    it('sets the fee record principalBalanceAdjustment to 10', async () => {
+    it('sets the fee record principalBalanceAdjustment to the difference between the fee record utilisation and the facility utilisation data utilisation', async () => {
```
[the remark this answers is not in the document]
JoshBinns2000: After moving the logic of calculating this value to its own function, I have added tests for those cases there instead

dtfs-central-api/api-tests/v1/utilisation-reports/post-upload-utilisation-report.api-test.ts
```diff
@@ -0,0 +1,210 @@
...
+    const response = await testApi.post(payload).to(getUrl());
+
+    // Assert
+    expect(response.status).toBe(HttpStatusCode.Created);
+
+    const feeRecordCount = await SqlDbHelper.manager.count(FeeRecordEntity, {});
+    expect(feeRecordCount).toBe(3);
+  });
+
+  it('creates an entry in the FacilityUtilisationData table for each facility id in the report csv data', async () => {
```
RLavender98, line 97: What is the intended difference between this test and `creates a new FacilityUtilisationData row using the report reportPeriod if the report data has a facility id which does not already exist` below?

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/report-uploaded/report-uploaded.event-handler.test.ts
```diff
@@ -53,4 +57,48 @@ describe('handleUtilisationReportReportUploadedEvent', () => {
...
+
+    const azureFileInfoEntity = AzureFileInfoEntity.create({
+      ...azureFileInfo,
+      requestSource,
+    });
+    report.updateWithUploadDetails({
+      azureFileInfo: azureFileInfoEntity,
+      uploadedByUserId,
+      requestSource,
+    });
```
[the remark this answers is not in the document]
JoshBinns2000: No idea - probably came from a copy-paste

dtfs-central-api/api-tests/v1/utilisation-reports/post-upload-utilisation-report.api-test.ts
JoshBinns2000: Have updated the logic and, as a result, this test

PR #3314, opened 2024-07-16
fix(DTFS2-7284): optional chaining on bond parties

abhi-markan:
## Introduction :pencil2:
ACBS prerequisite check has failed due to missing bond related properties.

* Added optional chaining.
* Added mock resolved value to `createACBS`  and `issueAcbsFacilities` functions.
* Fixed CRON expression (every 10 minutes)

### DIFF
--- a/trade-finance-manager-api/api-tests/v1/deals/deals-submit.facilities.api-test.js
+++ b/trade-finance-manager-api/api-tests/v1/deals/deals-submit.facilities.api-test.js
@@ -1,9 +1,4 @@
-jest.mock('../../../src/v1/controllers/acbs.controller', () => ({
-  issueAcbsFacilities: jest.fn(),
-}));
-
 const api = require('../../../src/v1/api');
-const acbsController = require('../../../src/v1/controllers/acbs.controller');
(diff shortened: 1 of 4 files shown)


JoshBinns2000 (approved):
LGTM

trade-finance-manager-api/src/cron-scheduler-jobs/check-azure-acbs-function.js
```diff
@@ -1,4 +1,4 @@
-const defaultSchedule = '*/10 * * * * *';
```
abhi-markan, line 1: There was one extra astric, above resolves to `every 10th minute`.

trade-finance-manager-api/api-tests/v1/deals/deals-submit.facilities.api-test.js
```diff
@@ -1,9 +1,4 @@
-jest.mock('../../../src/v1/controllers/acbs.controller', () => ({
```
abhi-markan, line 1: Standard approach would be to add this to `beforeEach` block, where it now has been moved to.

JoshBinns2000 (approved):

PR #3317, opened 2024-07-17
feat(FN-1725): add tfm facilities repo

JoshBinns2000:
## Introduction :pencil2:
We want a repo for the mongo tfm facilities collection.

- Adds a new `TfmFacilitiesRepo` class
- Updates the existing tfm-facilities controller to be typescript and use the new repo
- Adds static helpers to the `TfmFacilitiesRepo` which parses a found mongo collection item*

*the type of the tfm-facilities collection is not always consistent and is hard to document. In order to reach a compromise between overly strict type checking and no type information at all, we introduce a set of methods on the `TfmFacilitiesRepo` to parse any result from a call to the repo. The parser uses a `zod` schema to validate the information that we expect to exist on the document and, if valid, returns the parsed data. The schema which is used to parse this data should be extended with new fields as and when they are needed.

- Adds generic to audit logs delete many method added by @MarRobSoftwire which allows for type inference on the filter

### DIFF
--- a/dtfs-central-api/src/repositories/tfm-facilities-repo/tfm-facilities.repo.test.ts
+++ b/dtfs-central-api/src/repositories/tfm-facilities-repo/tfm-facilities.repo.test.ts
@@ -1,4 +1,5 @@
 import { ObjectId } from 'mongodb';
+import { ZodError } from 'zod';
 import { ParsedTfmFacility, RawTfmFacility, TfmFacilitiesRepo } from './tfm-facilities.repo';
 
 describe('tfm-facilities-repo', () => {
@@ -33,7 +34,7 @@ describe('tfm-facilities-repo', () => {
       const input = {} as unknown as RawTfmFacility;
 
       // Act / Assert
-      expect(() => TfmFacilitiesRepo.validateAndParseFindOneResult(input)).toThrow();
+      expect(() => TfmFacilitiesRepo.validateAndParseFindOneResult(input)).toThrow(ZodError);
     });
 
     it("throws an error when the '_id' field is not a valid object id", () => {
@@ -41,7 +42,7 @@ describe('tfm-facilities-repo', () => {
       const input = { ...aValidTfmFacilityDocument(), _id: 123 } as unknown as RawTfmFacility;
 
       // Act / Assert
-      expect(() => TfmFacilitiesRepo.validateAndParseFindOneResult(input)).toThrow();
+      expect(() => TfmFacilitiesRepo.validateAndParseFindOneResult(input)).toThrow(ZodError);
     });
 
     it("throws an error when the 'facilitySnapshot' field is undefined", () => {
@@ -49,7 +50,7 @@ describe('tfm-facilities-repo', () => {
       const input = { ...aValidTfmFacilityDocument(), facilitySnapshot: undefined } as unknown as RawTfmFacility;
 
       // Act / Assert
-      expect(() => TfmFacilitiesRepo.validateAndParseFindOneResult(input)).toThrow();
+      expect(() => TfmFacilitiesRepo.validateAndParseFindOneResult(input)).toThrow(ZodError);
     });
 
     const REQUIRED_FACILITY_SNAPSHOT_FIELDS: (keyof ParsedTfmFacility['facilitySnapshot'])[] = [
@@ -82,7 +83,7 @@ describe('tfm-facilities-repo', () => {
       };
 
       // Act / Assert
-      expect(() => TfmFacilitiesRepo.validateAndParseFindOneResult(input)).toThrow();
+      expect(() => TfmFacilitiesRepo.validateAndParseFindOneResult(input)).toThrow(ZodError);
     });
 
     const NULLABLE_FACILITY_SNAPSHOT_FIELDS: (keyof ParsedTfmFacility['facilitySnapshot'])[] = ['coverStartDate', 'coverEndDate', 'issueDate', 'monthsOfCover'];
@@ -132,22 +133,6 @@ describe('tfm-facilities-repo', () => {
     });
 
     const FACILITY_SNAPSHOT_DATE_FIELDS: (keyof ParsedTfmFacility['facilitySnapshot'])[] = ['coverStartDate', 'coverEndDate', 'issueDate'];
-    it.each(FACILITY_SNAPSHOT_DATE_FIELDS)("coerces the 'facilitySnapshot.%s.$date' field to a Date object when the value is an ISO string", (field) => {
-      // Arrange
-      const date = new Date();
-      const input = {
-        ...aValidTfmFacilityDocument(),
-        facilitySnapshot: {
-          ...aValidTfmFacilityDocumentFacilitySnapshotResult(),
-          [field]: { $date: date.toISOString() },
-        },
-      };
-
-      // Act / Assert
-      const result = TfmFacilitiesRepo.validateAndParseFindOneResult(input);
-      expect(result.facilitySnapshot[field]).toEqual({ $date: date });
-    });
-
     it.each(FACILITY_SNAPSHOT_DATE_FIELDS)("coerces the 'facilitySnapshot.%s.$date' field to a Date object when the value is a timestamp", (field) => {
       // Arrange
       const date = new Date();
--- a/dtfs-central-api/src/repositories/tfm-facilities-repo/tfm-facilities.repo.test.ts
+++ b/dtfs-central-api/src/repositories/tfm-facilities-repo/tfm-facilities.repo.test.ts
@@ -1,0 +1,167 @@
+import { ObjectId } from 'mongodb';
+import { ParsedTfmFacility, RawTfmFacility, TfmFacilitiesRepo } from './tfm-facilities.repo';
+
+describe('tfm-facilities-repo', () => {
+  describe('TfmFacilitiesRepo.validateAndParseFindOneResult', () => {
+    const aValidTfmFacilityDocumentFacilitySnapshotResult = () => ({
+      _id: new ObjectId(),
+      dealId: new ObjectId(),
+      type: 'A type',
+      hasBeenIssued: true,
+      name: 'A name',
+      shouldCoverStartOnSubmission: true,
+      coverStartDate: null,
+      coverEndDate: null,
+      issueDate: null,
+      monthsOfCover: 12,
+      details: [],
+      detailsOther: '',
+      value: 9000,
+      coverPercentage: 12,
+      interestPercentage: 6,
+      ukefFacilityId: '12345678',
+      dayCountBasis: 365,
+    });
+
+    const aValidTfmFacilityDocument = () => ({
+      _id: new ObjectId(),
+      facilitySnapshot: aValidTfmFacilityDocumentFacilitySnapshotResult(),
+    });
+
+    it('throws an error when the input is an empty object', () => {
+      // Arrange
+      const input = {} as unknown as RawTfmFacility;
+
+      // Act / Assert
+      expect(() => TfmFacilitiesRepo.validateAndParseFindOneResult(input)).toThrow();
+    });
+
+    it("throws an error when the '_id' field is not a valid object id", () => {
+      // Arrange
+      const input = { ...aValidTfmFacilityDocument(), _id: 123 } as unknown as RawTfmFacility;
+
+      // Act / Assert
+      expect(() => TfmFacilitiesRepo.validateAndParseFindOneResult(input)).toThrow();
+    });
+
+    it("throws an error when the 'facilitySnapshot' field is undefined", () => {
+      // Arrange
+      const input = { ...aValidTfmFacilityDocument(), facilitySnapshot: undefined } as unknown as RawTfmFacility;
+
+      // Act / Assert
+      expect(() => TfmFacilitiesRepo.validateAndParseFindOneResult(input)).toThrow();
+    });
+
+    const REQUIRED_FACILITY_SNAPSHOT_FIELDS: (keyof ParsedTfmFacility['facilitySnapshot'])[] = [
+      '_id',
+      'dealId',
+      'type',
+      'hasBeenIssued',
+      'name',
+      'shouldCoverStartOnSubmission',
+      'coverStartDate',
+      'coverEndDate',
+      'issueDate',
+      'monthsOfCover',
+      'details',
+      'detailsOther',
+      'value',
+      'coverPercentage',
+      'interestPercentage',
+      'ukefFacilityId',
+      'dayCountBasis',
+    ];
+    it.each(REQUIRED_FACILITY_SNAPSHOT_FIELDS)("throws an error when the 'facilitySnapshot.%s' field is missing", (field) => {
+      // Arrange
+      const input = {
+        ...aValidTfmFacilityDocument(),
+        facilitySnapshot: {
+          ...aValidTfmFacilityDocumentFacilitySnapshotResult(),
+          [field]: undefined,
+        },
+      };
+
+      // Act / Assert
+      expect(() => TfmFacilitiesRepo.validateAndParseFindOneResult(input)).toThrow();
+    });
+
+    const NULLABLE_FACILITY_SNAPSHOT_FIELDS: (keyof ParsedTfmFacility['facilitySnapshot'])[] = ['coverStartDate', 'coverEndDate', 'issueDate', 'monthsOfCover'];
+    it.each(NULLABLE_FACILITY_SNAPSHOT_FIELDS)(
+      "does not throw an error when the 'facilitySnapshot.%s' field is null and returns the object with the same field",
+      (field) => {
+        // Arrange
+        const input = {
+          ...aValidTfmFacilityDocument(),
+          facilitySnapshot: {
+            ...aValidTfmFacilityDocumentFacilitySnapshotResult(),
+            [field]: null,
+          },
+        };
+
+        // Act / Assert
+        const result = TfmFacilitiesRepo.validateAndParseFindOneResult(input);
+        expect(result.facilitySnapshot[field]).toBeNull();
+      },
+    );
+
+    it('does not throw an error when the input contains an extra field and does not return than field in the result', () => {
+      // Arrange
+      const input = {
+        ...aValidTfmFacilityDocument(),
+        someOtherField: 'Some other value',
+      };
+
+      // Act / Assert
+      const result = TfmFacilitiesRepo.validateAndParseFindOneResult(input);
+      expect(result).not.toHaveProperty('someOtherField');
+    });
+
+    it("does not throw an error when the 'facilitySnapshot' object contains an extra field and does not return than field in the result", () => {
+      // Arrange
+      const input = {
+        ...aValidTfmFacilityDocument(),
+        facilitySnapshot: {
+          ...aValidTfmFacilityDocumentFacilitySnapshotResult(),
+          someOtherField: 'Some other value',
+        },
+      };
+
+      // Act / Assert
+      const result = TfmFacilitiesRepo.validateAndParseFindOneResult(input);
+      expect(result.facilitySnapshot).not.toHaveProperty('someOtherField');
+    });
+
+    const FACILITY_SNAPSHOT_DATE_FIELDS: (keyof ParsedTfmFacility['facilitySnapshot'])[] = ['coverStartDate', 'coverEndDate', 'issueDate'];
+    it.each(FACILITY_SNAPSHOT_DATE_FIELDS)("coerces the 'facilitySnapshot.%s.$date' field to a Date object when the value is an ISO string", (field) => {
+      // Arrange
+      const date = new Date();
+      const input = {
+        ...aValidTfmFacilityDocument(),
+        facilitySnapshot: {
+          ...aValidTfmFacilityDocumentFacilitySnapshotResult(),
+          [field]: { $date: date.toISOString() },
+        },
+      };
+
+      // Act / Assert
+      const result = TfmFacilitiesRepo.validateAndParseFindOneResult(input);
+      expect(result.facilitySnapshot[field]).toEqual({ $date: date });
+    });
+
+    it.each(FACILITY_SNAPSHOT_DATE_FIELDS)("coerces the 'facilitySnapshot.%s.$date' field to a Date object when the value is a timestamp", (field) => {
+      // Arrange
+      const date = new Date();
+      const input = {
+        ...aValidTfmFacilityDocument(),
+        facilitySnapshot: {
+          ...aValidTfmFacilityDocumentFacilitySnapshotResult(),
+          [field]: { $date: date.getTime() },
+        },
+      };
+
+      // Act / Assert
+      const result = TfmFacilitiesRepo.validateAndParseFindOneResult(input);
+      expect(result.facilitySnapshot[field]).toEqual({ $date: date });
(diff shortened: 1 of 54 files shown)


dtfs-central-api/src/repositories/tfm-facilities-repo/tfm-facilities.repo.ts
```diff
@@ -0,0 +1,80 @@
...
+export type RawTfmFacility = WithId<object>;
+
+export type ParsedTfmFacility = z.infer<typeof TfmFacilitySchema>;
+
+export class TfmFacilitiesRepo {
+  /**
+   * Gets the TFM facilities collection
+   * @returns The TFM facilities collection
+   */
+  public static async getCollection(): Promise<Collection<WithoutId<RawTfmFacility>>> {
```
MarRobSoftwire, line 15: Do we want to move away from this being `public`? I assume it would be a real pain to make this private

dtfs-central-api/src/repositories/tfm-facilities-repo/tfm-facilities.repo.ts
```diff
@@ -0,0 +1,80 @@
...
+    return await mongoDbClient.getCollection('tfm-facilities');
+  }
+
+  /**
+   * Validates and parses the result of a mongo `findOne` operation
+   * @param result - The result of any `findOne` operation
+   * @returns The parsed result
+   * @throws If the result does not match the TFM facilities schema
```
MarRobSoftwire, line 23:
```suggestion
   * @throws {import('zod').ZodError} If the result does not match the TFM facilities schema
```

dtfs-central-api/src/repositories/tfm-facilities-repo/tfm-facilities.repo.ts
```diff
@@ -0,0 +1,80 @@
...
+   */
+  public static validateAndParseFindOneResult(result: RawTfmFacility): ParsedTfmFacility {
+    return TfmFacilitySchema.parse(result);
+  }
+
+  /**
+   * Validates and parses the result of a mongo `find` operation
+   * @param result - The result of any `find` operation
+   * @returns The parsed results
+   * @throws If the each item in the results array does not match the TFM facilities schema
```
MarRobSoftwire, line 33:
```suggestion
   * @throws If any item in the results array does not match the TFM facilities schema
```

dtfs-central-api/src/repositories/tfm-facilities-repo/tfm-facilities.repo.ts
```diff
@@ -0,0 +1,80 @@
...
+  /**
+   * Updates the tfm facility with the supplied id
+   * @param id - The id to search by
+   * @param update - The update to apply
+   * @param [options] - Update options
+   * @returns The updated document
+   */
+  public static async findOneByIdAndUpdate(id: string | ObjectId, update: UpdateFilter<WithoutId<RawTfmFacility>>, options: FindOneAndUpdateOptions = {}) {
+    const collection = await this.getCollection();
+    return await collection.findOneAndUpdate({ _id: { $eq: new ObjectId(id) } }, update, options);
```
MarRobSoftwire, line 68: Obviously its used in many places but `findOneAndUpdate` returns a `ModifyResult` in mongodb v4 & just the document (i think) from v5 onwards. 
Would therefore prefer an `updateOne` unless you have objections. If refactoring places `findOneAndUpdate` is used is too difficult, maybe mark this function as `@deprecated`?

MarRobSoftwire: Either way `@returns The updated document` is a bit of a lie

dtfs-central-api/src/repositories/tfm-facilities-repo/tfm-facilities.repo.test.ts
```diff
@@ -0,0 +1,167 @@
...
+      _id: new ObjectId(),
+      facilitySnapshot: aValidTfmFacilityDocumentFacilitySnapshotResult(),
+    });
+
+    it('throws an error when the input is an empty object', () => {
+      // Arrange
+      const input = {} as unknown as RawTfmFacility;
+
+      // Act / Assert
+      expect(() => TfmFacilitiesRepo.validateAndParseFindOneResult(input)).toThrow();
```
MarRobSoftwire, line 36: FWIW we can make sure we're not getting a different error here
```suggestion
      expect(() => TfmFacilitiesRepo.validateAndParseFindOneResult(input)).toThrow(ZodError);
```

dtfs-central-api/src/repositories/tfm-facilities-repo/tfm-facilities.repo.test.ts
```diff
@@ -0,0 +1,167 @@
...
+        },
+      };
+
+      // Act / Assert
+      const result = TfmFacilitiesRepo.validateAndParseFindOneResult(input);
+      expect(result.facilitySnapshot).not.toHaveProperty('someOtherField');
+    });
+
+    const FACILITY_SNAPSHOT_DATE_FIELDS: (keyof ParsedTfmFacility['facilitySnapshot'])[] = ['coverStartDate', 'coverEndDate', 'issueDate'];
+    it.each(FACILITY_SNAPSHOT_DATE_FIELDS)("coerces the 'facilitySnapshot.%s.$date' field to a Date object when the value is an ISO string", (field) => {
```
MarRobSoftwire, line 135: Do we want to coerce date fields from either an ISO string or a timestamp? Given they'll always be one of the types would it be clearer (and more useful) to only accept one of the types (in each case)

MarRobSoftwire: This comment might not make sense how ive written it

dtfs-central-api/src/repositories/tfm-facilities-repo/tfm-facility.schema.ts
```diff
@@ -0,0 +1,36 @@
...
+ */
+export const TfmFacilitySchema = z.object({
+  _id: MongoObjectIdSchema,
+  facilitySnapshot: z.object({
+    _id: MongoObjectIdSchema,
+    dealId: MongoObjectIdSchema,
+    type: z.string(),
+    hasBeenIssued: z.boolean(),
+    name: z.string(),
+    shouldCoverStartOnSubmission: z.boolean(), // boolean or string?
```
MarRobSoftwire, line 23: `// boolean or string?` indeed

dtfs-central-api/src/repositories/tfm-facilities-repo/tfm-facility.schema.ts
```diff
@@ -0,0 +1,36 @@
+import z from 'zod';
+import { MongoObjectIdSchema } from '../../v1/routes/middleware/payload-validation/schemas';
+
+const TfmFacilityDateFieldSchema = z
+  .object({
+    $date: z.coerce.date(),
```
MarRobSoftwire, line 6: z.coerce.date() just calls `new Date` (if i recall) so does weird stuff like parsing `'1'` as `Mon Jan 01 2001`. Would you be happy to handroll something using `parseISO` and `toDate` from date-fns?

dtfs-central-api/src/v1/controllers/tfm/amendments/tfm-get-amendments.controller.js
```diff
@@ -51,7 +50,7 @@ const findAllAmendmentsByFacilityId = async (facilityId) => {
     if (!ObjectId.isValid(facilityId)) {
       throw new Error('Invalid facility Id');
     }
-    const collection = await db.getCollection(MONGO_DB_COLLECTIONS.TFM_FACILITIES);
+    const collection = await TfmFacilitiesRepo.getCollection();
```
MarRobSoftwire, line 53: I dont think there's a good alternative to using these `aggregate` pipelines but I wish they were also in the repo & all the mongodb stuff could be hidden

MarRobSoftwire: FEATURE ENVY

dtfs-central-api/src/v1/controllers/tfm/amendments/tfm-post-amendments.controller.js
```diff
@@ -55,7 +54,7 @@ exports.postTfmAmendment = async (req, res) => {
   if (latestCompletedAmendmentVersion) {
     amendment.version = latestCompletedAmendmentVersion + 1;
   }
-  const collection = await db.getCollection(MONGO_DB_COLLECTIONS.TFM_FACILITIES);
+  const collection = await TfmFacilitiesRepo.getCollection();
   await collection.updateOne(
```
MarRobSoftwire, line 58: definitely think this could be behind the Repo

dtfs-central-api/src/v1/controllers/tfm/amendments/tfm-put-amendments.controller.js
```diff
@@ -32,7 +32,7 @@ exports.updateTfmAmendment = async (req, res) => {
     return res.status(404).send({ status: 404, message: 'The amendment does not exist' });
   }
 
-  const collection = await db.getCollection(MONGO_DB_COLLECTIONS.TFM_FACILITIES);
+  const collection = await TfmFacilitiesRepo.getCollection();
```
MarRobSoftwire, line 35: Also could use an updateOne on TfmFacilitiesRepo

dtfs-central-api/src/repositories/tfm-facilities-repo/tfm-facility.schema.ts
JoshBinns2000: It does just call `new Date`. I am happy to leave this as a standard `z.coerce.date()` because this is something that should be validated on the way in rather than the way out

dtfs-central-api/src/repositories/tfm-facilities-repo/tfm-facilities.repo.ts
```diff
@@ -30,7 +26,7 @@ export class TfmFacilitiesRepo {
    * Validates and parses the result of a mongo `find` operation
    * @param result - The result of any `find` operation
    * @returns The parsed results
-   * @throws If the each item in the results array does not match the TFM facilities schema
+   * @throws {z.ZodError} If any item in the results array does not match the TFM facilities schema
```
MarRobSoftwire, line 29: I dont think this is resolving properly

dtfs-central-api/src/repositories/tfm-facilities-repo/tfm-facilities.repo.ts
MarRobSoftwire: Hmmm fair point. Could you make the jsdoc something like `a ModifyResult containing the document`

MarRobSoftwire: Or to separate the function from the implementation a bit more, add some handling


```suggestion
    const modifyResult = await collection.findOneAndUpdate({ _id: { $eq: new ObjectId(id) } }, update, options);
    if (!modifyResult.ok) {
      throw new Error();
    }
    return modifyResult.value;
```

dtfs-central-api/src/repositories/tfm-facilities-repo/tfm-facilities.repo.ts
```diff
@@ -0,0 +1,110 @@
...
+    return await collection.find({ _id: { $in: ids } }).toArray();
+  }
+
+  /**
+   * Method to perform a custom operation on the TFM
+   * facilities collection
+   * @param operation - The database operation to perform
+   * @returns The result of the supplied operation
+   */
+  public static async custom<TReturn>(operation: (collection: Collection<WithoutId<RawTfmFacility>>) => TReturn | Promise<TReturn>): Promise<TReturn> {
```
MarRobSoftwire, line 106: I think this is better but still feels a bit envious. Happy to go with for now/see what others think

dtfs-central-api/src/repositories/tfm-facilities-repo/tfm-facilities.repo.ts
JoshBinns2000: The type of `value` is always `TSchema | null` so I'd rather delegate the error handling to whatever controller is using this method, but I am happy to only return the document rather than the entire result

dtfs-central-api/src/repositories/tfm-facilities-repo/tfm-facilities.repo.ts
JoshBinns2000: Have replaced with an `executeAggregate` method so that the `collection` is no longer exposed

dtfs-central-api/src/repositories/tfm-facilities-repo/tfm-facilities.repo.ts
JoshBinns2000: It is in my IDE :shrug:

libs/common/src/change-stream/delete-many.ts
```diff
@@ -68,17 +84,20 @@ const deleteManyWithAuditLogs = async ({ filter, collectionName, db, auditDetail
...
       collectionName,
       db,
       auditDetails,
     });
-
-    return { acknowledged: true };
   }
   const collection = await db.getCollection(collectionName);
-  return collection.deleteMany(filter);
+  return await collection.deleteMany(filter);
```
AlexBramhill, line 102: is this await (and the one above) redundant as we're not inside a try catch

dtfs-central-api/src/repositories/tfm-facilities-repo/tfm-facilities.repo.ts
```diff
@@ -0,0 +1,170 @@
...
+
+  /**
+   * Method to execute an aggregate against the TFM facilities collection
+   * @param pipeline - The aggregate pipeline
+   * @param [options] - The aggregate options
+   * @returns The aggregation cursor and array of aggregate results
+   */
+  public static async executeAggregate(pipeline: Document[], options?: AggregateOptions): Promise<ExecuteAggregateResult> {
+    const collection = await this.getCollection();
+    const aggregationCursor = collection.aggregate(pipeline, options);
```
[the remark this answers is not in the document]
JoshBinns2000: No idea - I suppose we can cite YAGNI here

libs/common/src/change-stream/delete-many.ts
```diff
@@ -50,11 +61,16 @@ const deleteManyWithAuditLogs = async ({ filter, collectionName, db, auditDetail
...
 
-      const deleteResult = await collection.deleteMany({ _id: { $in: documentsToDeleteIds } }, { session });
+      const deleteManyFilter = {
+        _id: { $in: documentsToDeleteIds },
+      } as unknown as Filter<WithoutId<DbModel<CollectionName>>>;
+      const deleteResult = await collection.deleteMany(deleteManyFilter, { session });
       if (!(deleteResult.acknowledged && deleteResult.deletedCount === documentsToDeleteIds.length)) {
         throw new WriteConcernError();
       }
+      transactionDeleteResult = { ...deleteResult };
```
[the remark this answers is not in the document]
JoshBinns2000: Because the mongo transaction doesn't return the value you would expect it to. My assumption was that it should probably return whatever the supplied function returns, but it doesn't, so we need to update the thing we want to return in the outer scope during the transaction

dtfs-central-api/src/v1/controllers/tfm/amendments/tfm-get-amendments.controller.js
```diff
@@ -13,19 +12,16 @@ const CONSTANTS = require('../../../../constants');
...
-      .aggregate([
-        { $project: { _id: false, amendments: '$amendments' } },
-        { $unwind: '$amendments' },
-        { $match: { 'amendments.status': { $eq: status } } },
-        { $group: { _id: '$_id', amendments: { $push: '$amendments' } } },
-        { $project: { _id: false, amendments: true } },
-      ])
-      .toArray();
+    const { resultsArray } = await TfmFacilitiesRepo.executeAggregate([
+      { $project: { _id: false, amendments: '$amendments' } },
```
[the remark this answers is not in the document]
JoshBinns2000: I'll have a think about this. For context, the reason I didn't is that it wasn't necessarily clear what the aggregate pipeline is doing, meaning the name to give to the repo method was not clear to me. Considering that both you and Mark have said this though, I will try to figure out what the pipeline is doing and come up with a descriptive function name for it

libs/common/src/change-stream/delete-many.ts
AlexBramhill: cool, am happy with that, I tend to lean the other way of "its not needed so lets leave it out", but happy for this

dtfs-central-api/src/v1/controllers/tfm/facility/tfm-get-facilities.controller.ts
```diff
@@ -0,0 +1,98 @@
...
+   *        "dealType": "GEF",
+   *        "type": "Cash",
+   *        "value": 1000000,
+   *        "currency": "GBP",
+   *        "coverEndDate": "2021-08-12T00:00:00.000Z",.
+   *        "companyName": "Mock company name",
+   *        "hasBeenIssued": true
+   *     }
+   * }]
+   */
```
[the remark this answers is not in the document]
JoshBinns2000: I'm hesitant about moving them to the repo because we can't be sure the type is correct - I'd rather just remove them outright. What do you think?

dtfs-central-api/src/repositories/tfm-facilities-repo/tfm-facilities.repo.ts
```diff
@@ -0,0 +1,170 @@
...
+    return { aggregationCursor, resultsArray };
+  }
+
+  /**
+   * Delete many TFM facilities by the supplied deal id
+   * @param dealId - The deal id
+   * @param auditDetails - The audit details
+   * @returns The delete result
+   */
+  public static async deleteManyByDealId(dealId: string | ObjectId, auditDetails: AuditDetails): Promise<{ acknowledged: boolean }> {
```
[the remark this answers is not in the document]
JoshBinns2000: I don't think so - why would it be?

dtfs-central-api/src/repositories/tfm-facilities-repo/tfm-facilities.repo.test.ts
```diff
@@ -0,0 +1,152 @@
...
+    ];
+    it.each(REQUIRED_FACILITY_SNAPSHOT_FIELDS)("throws an error when the 'facilitySnapshot.%s' field is missing", (field) => {
+      // Arrange
+      const input = {
+        ...aValidTfmFacilityDocument(),
+        facilitySnapshot: {
+          ...aValidTfmFacilityDocumentFacilitySnapshotResult(),
+          [field]: undefined,
+        },
+      };
```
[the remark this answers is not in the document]
JoshBinns2000: We could, but I prefer the code the way it is as I think it's a bit more explicit about what we're doing

dtfs-central-api/src/v1/controllers/tfm/deal/tfm-delete-deal.controller.ts
```diff
@@ -40,12 +41,7 @@ export const deleteDeal = async (req: CustomExpressRequest<{ reqBody: { auditDet
       auditDetails,
     });
 
-    await deleteMany({
-      filter: { 'facilitySnapshot.dealId': { $eq: deal._id } },
-      collectionName: MONGO_DB_COLLECTIONS.TFM_FACILITIES,
-      db: mongoDbClient,
-      auditDetails,
-    });
+    await TfmFacilitiesRepo.deleteManyByDealId(deal._id, auditDetails);
```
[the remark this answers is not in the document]
JoshBinns2000: Discussed with @MarRobSoftwire : he says yes

e2e-tests/tfm/cypress/e2e/journeys/case-amendments/amendments-version-order.spec.js
```diff
@@ -286,9 +286,9 @@ context('Amendments underwriting - amendments should be in correct order of vers
     amendmentsPage.continueAmendment().click();
     cy.url().should('contain', 'cover-end-date');
 
-    amendmentsPage.amendmentCoverEndDateDayInput().clear().focused().type(dateConstants.threeMonthsDay);
-    amendmentsPage.amendmentCoverEndDateMonthInput().clear().focused().type(dateConstants.threeMonthsMonth);
-    amendmentsPage.amendmentCoverEndDateYearInput().clear().focused().type(dateConstants.threeMonthsYear);
+    amendmentsPage.amendmentCoverEndDateDayInput().clear().focused().type(dateConstants.threeYearsDay);
```
AlexBramhill, line 289: Any reason this is 3 years 3 months const do we know?

AlexBramhill: (also this const is 3 years and 3 months)

e2e-tests/tfm/cypress/e2e/journeys/case-amendments/add-bank-decision-proceed.spec.js
```diff
@@ -20,6 +20,7 @@ context('Amendments underwriting - add banks decision - proceed', () => {
   const todayIsEndOfMonth = add(new Date(), { days: 1 }).getDate() === 1;
   const aMonthFromNowIsEndOfMonth = add(new Date(), { months: 1, days: 1 }).getDate() === 1;
   const facilityTenor = todayIsEndOfMonth && !aMonthFromNowIsEndOfMonth ? '25 months' : '26 months';
+  const updatedFacilityTenor = '25 months';
```
AlexBramhill, line 23: Why is this needed? :)

e2e-tests/tfm/cypress/e2e/journeys/case-amendments/add-amendment-request-check-previous.spec.js
```diff
@@ -230,9 +230,9 @@ context('Amendments - should not allow amendments to have same coverEndDate/valu
     amendmentsPage.continueAmendment().click();
     cy.url().should('contain', 'cover-end-date');
 
-    amendmentsPage.amendmentCoverEndDateDayInput().clear().focused().type(dateConstants.threeDaysDay);
-    amendmentsPage.amendmentCoverEndDateMonthInput().clear().focused().type(dateConstants.threeDaysMonth);
-    amendmentsPage.amendmentCoverEndDateYearInput().clear().focused().type(dateConstants.threeDaysYear);
+    amendmentsPage.amendmentCoverEndDateDayInput().clear().focused().type(dateConstants.threeYearsDay);
```
AlexBramhill, line 233: same q about why this has changed

dtfs-central-api/src/repositories/tfm-facilities-repo/tfm-facilities.repo.ts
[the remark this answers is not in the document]
AlexBramhill: nvm i dug down a bit further and found what i was looking for

AlexBramhill (approved):
The decoupling of mongodb from the controllers is going to have a large positive effect on the codebase. Great job

AlexBramhill (approved):

PR #3322, opened 2024-07-18
feat(FN-1728): display fee record group status correctly for ready to key data

RLavender98:
## Introduction :pencil2:
Fee records in a group which have had their keying sheet generated can have different statuses to each other.

- Display the status 'READY TO KEY' when any fee record in a group has status `READY_TO_KEY`
- Use the status of the first record in the group otherwise (since all fee records will have the same status)
- Change to test set up to allow easily testing with different combinations of statuses

Group in premium payments tab:
Screenshot 2024-07-18 at 15 21 14
Corresponding keying sheet tab rows for some of the fee records in group:
Screenshot 2024-07-18 at 15 21 43
Screenshot 2024-07-18 at 15 21 59


JoshBinns2000 (approved):
LGTM

PR #3324, opened 2024-07-19
feat(FN-1722): generate and display keying sheet rows for auto matched zero payment fee records

RLavender98:
## Introduction :pencil2:
There is a bug wherein you cannot generate keying data if there are zero payment fee records which have been auto matched and no other changes made since in that case the report is still in the `PENDING_RECONCILIATION` status.

Also rows with no payments are not showing up on the keying sheet table.

- Allow keying sheet data to be generated when the report is in status `PENDING_RECONCILIATION`
- Update report status to `RECONCILIATION_IN_PROGRESS` when keying sheet data is generated if it is not already
- Display keying sheet rows with no fee payments with dashes in the payment columns (TBC if dashes are the desired behaviour - awaiting reply on the jira ticket)

Screenshot 2024-07-19 at 18 22 26


dtfs-central-api/api-tests/v1/utilisation-reports/post-keying-data.api-test.ts
```diff
@@ -22,7 +29,9 @@ describe('POST /v1/utilisation-reports/:reportId/keying-data', () => {
   const tfmUser = aTfmUser();
   const tfmUserId = tfmUser._id.toString();
 
-  const anUploadedUtilisationReport = () =>
+  const anUploadedReconciliationInProgressUtilisationReport = () =>
+    UtilisationReportEntityMockBuilder.forStatus('RECONCILIATION_IN_PROGRESS').withId(reportId).withUploadedByUserId(portalUserId).build();
+  const anUploadedPendingReconciliationUtilisationReport = () =>
```
JoshBinns2000, line 34: The name of this doesn't seem to line up with the status used

dtfs-central-api/src/types/fee-records.ts
```diff
@@ -34,7 +34,7 @@ export type KeyingSheetRow = {
   facilityId: string;
   exporter: string;
   feePayments: {
-    dateReceived: Date;
+    dateReceived: Date | undefined;
```
JoshBinns2000, line 37: Just because this is something that gets sent in a response body, could we have it be `Date | null`?

JoshBinns2000 (commented):
Nice work - just two small comments

utils/sql-db-seeder/src/fee-record-payment-group/fee-record-payment-group.seeder.ts
```diff
@@ -75,6 +75,22 @@ export class FeeRecordPaymentGroupSeeder {
     return this;
   }
 
+  private async saveFacilityUtilisationData(dataSource: DataSource): Promise<void> {
```
JoshBinns2000, line 78: It looks to me like this method is already on `main` - has something gone wrong with merging `main` in at some point?

dtfs-central-api/src/types/fee-records.ts
```diff
@@ -34,7 +34,7 @@ export type KeyingSheetRow = {
   facilityId: string;
   exporter: string;
   feePayments: {
-    dateReceived: Date;
+    dateReceived: Date | null;
```
JoshBinns2000, line 37: Could we also update the type in the swagger docs for this route?

trade-finance-manager-ui/server/api-response-types/utilisation-report-reconciliation-details-response-body.ts
```diff
@@ -27,7 +27,7 @@ export type KeyingSheetRow = {
   feePayments: {
     currency: Currency;
     amount: number;
-    dateReceived: IsoDateTimeStamp;
+    dateReceived: IsoDateTimeStamp | undefined;
```
JoshBinns2000, line 30: I think this should now be `IsoDateTimeStamp | null` (probably the same for the tfm-ui response type)

JoshBinns2000 (approved):
LGTM

PR #3326, opened 2024-07-22
feat(dtfs2-6616): automatically amend facility guarantee in acbs

AlexBramhill:
## Introduction :pencil2:
We want to automatically amend facility guarantee in acbs

Adds new `acbs-amend-facility-guarantee-record` suborchestrated function, called in `acbs-amend-facility`
Adds new `activity-update-facility-guarantee`, called in `acbs-amend-guarantee


azure-functions/acbs-function/mappings/facility/facility-guarantee-amend.js
```diff
@@ -0,0 +1,26 @@
+const facilityGuaranteeAmend = (amendment) => {
```
ttbarnes, line 1: Could we add some documentation like the other functions, describing the related fields? e.g

```
/*
  "guaranteedLimit"      Description
  "expirationDate"       Description
*/
```

azure-functions/acbs-function/mappings/facility/facility-guarantee-amend.js
```diff
@@ -0,0 +1,26 @@
...
+  try {
+    const { amount, facilityGuaranteeDates } = amendment;
+    const record = {};
+
+    if (!amount && !facilityGuaranteeDates) {
+      throw new Error('Either amount or facilityGuaranteeDates.guaranteeExpiryDate is required.');
```
ttbarnes, line 7: Should the condition also check for `guaranteeExpiryDate`? As the error message mentions this, but it's currently only checking for `facilityGuaranteeDates`

azure-functions/acbs-function/mappings/facility/facility-guarantee-amend.js
```diff
@@ -0,0 +1,26 @@
...
+
+    if (amount) {
+      record.guaranteedLimit = amount;
+    }
+
+    if (facilityGuaranteeDates) {
+      record.expirationDate = facilityGuaranteeDates.guaranteeExpiryDate;
+    }
+
+    // Return amended FGR
```
ttbarnes, line 18: What is `FGR` ?

azure-functions/acbs-function/mappings/facility/facility-guarantee-amend.test.js
```diff
@@ -0,0 +1,56 @@
...
+  const aValidAmount = 123.456;
+  const aValidAmountExpected = 123.46;
+  const aValidGuaranteeExpiryDate = '2022-01-01';
+  const aValidGuaranteeExpiryDateExpected = '2022-01-01';
+  const aValidAmountRequest = { amount: aValidAmount };
+  const aValidAmountExpectedResult = { guaranteedLimit: aValidAmountExpected };
+  const aValidGuaranteeExpiryDateRequest = { facilityGuaranteeDates: { guaranteeExpiryDate: aValidGuaranteeExpiryDate } };
+  const aValidGuaranteeExpiryDateExpectedResult = { expirationDate: aValidGuaranteeExpiryDateExpected };
+  const aValidAmountAndGuaranteeExpiryDateRequest = { ...aValidAmountRequest, ...aValidGuaranteeExpiryDateRequest };
+  const aValidAmountAndGuaranteeExpiryDateExpectedResult = { ...aValidAmountExpectedResult, ...aValidGuaranteeExpiryDateExpectedResult };
```
ttbarnes, line 13: Finding these hard to read - not sure if this is an established pattern, I think an object structure would be easier to read/maintain

Does it need `aValid`? Can't see any "invalid" scenarios here?

azure-functions/acbs-function/src/functions/acbs-amend-facility.js
```diff
@@ -142,11 +142,20 @@ df.app.orchestration('acbs-amend-facility', function* amendFacility(context) {
 
     yield context.df.Task.all([facilityCovenantRecord]);
 
+    // 4. SOF: Facility Guarantee Record (FCR)
```
ttbarnes, line 145: Is FCR correct - not sure on the ACBS/business terminologies here, but i'd expect the acronym to be `FGR` to match the title?

azure-functions/acbs-function/src/functions/activity-update-facility-guarantee.js
```diff
@@ -0,0 +1,81 @@
...
+      throw new Error('Invalid facility guarantee update payload');
+    }
+
+    const { facilityIdentifier, acbsFacilityGuaranteeInput } = payload;
+
+    // Call create guarantee API
+    const submittedToACBS = getNowAsIsoString();
+    const { status, data } = await api.updateFacilityGuarantee(facilityIdentifier, acbsFacilityGuaranteeInput);
+
+    // Throw error upon an unsuccessful response
```
ttbarnes, line 48: Do we need this comment, feels self explanatory?

azure-functions/acbs-function/mappings/facility/facility-guarantee-amend.js
AlexBramhill: This was a suggested comment -- it's matching patterns from the other files -- Facility Guarantee Record

azure-functions/acbs-function/mappings/facility/facility-guarantee-amend.test.js
AlexBramhill: updated to pattern.

We do have invalid tests below -- however the objects are just used for one test so no need for reusability.

azure-functions/acbs-function/mappings/facility/facility-guarantee-amend.js
AlexBramhill: Great shout -- added

azure-functions/acbs-function/src/functions/acbs-amend-facility.js
AlexBramhill: reat spot

azure-functions/acbs-function/src/functions/activity-update-facility-guarantee.js
AlexBramhill: I agree, removed

azure-functions/acbs-function/mappings/facility/facility-guarantee-amend.js
ttbarnes: Nice one @AlexBramhill , not sure if we should follow the same pattern as other ACBS functions I can see where they have non-jsdoc style documentation at the top, seems the list the ACBS specific fields and what they are?

azure-functions/acbs-function/mappings/facility/facility-guarantee-amend.test.js
ttbarnes: Really not a fan of these variable names they're very long e.g

```js
aValidAmountAndGuaranteeExpiryDateRequest
aValidAmountAndGuaranteeExpiryDateExpectedResult

// ... 

expect(result).toEqual(aValidAmountAndGuaranteeExpiryDateExpectedResult);
```

How about something more like this? Maybe with some kind of "generic" `expected` object

```js
const mockAmount = 123.456

const mockAmendment = {
  amount: 123.456,
  facilityGuaranteeDates: { guaranteeExpiryDate: '2022-01-01' }
};

it('should ...', () => {
  const result = facilityGuaranteeAmend(mockAmendment);

  const expected = {
    targetAmount: expected.targetAmount,
    targetAmount: 123.456,
    expirationDate: '2022-01-01',
  };

});
```

AlexBramhill: updated

azure-functions/acbs-function/mappings/facility/facility-guarantee-amend.js
AlexBramhill: I've added it as JSDocs. I think this is ultimately better and does not impact the code for the worst by introducing any patterns which will cause issue if not implimented everywhere... lmk what you think

azure-functions/acbs-function/.eslintrc.js
```diff
@@ -3,6 +3,7 @@ const baseParserOptions = {
 };
 
 const baseRules = {
+  'no-param-reassign': ['error', { props: true, ignorePropertyModificationsForRegex: ['req', 'res'] }],
```
ttbarnes, line 6: Do we need this? I can't see any new param reassigns with req/res

AlexBramhill: good catch, removed

ttbarnes (dismissed):

azure-functions/acbs-function/mappings/facility/facility-guarantee-amend.js
```diff
@@ -0,0 +1,42 @@
...
+const facilityGuaranteeAmend = (amendment) => {
+  try {
+    const { amount, facilityGuaranteeDates } = amendment;
+    const record = {};
+
+    if (!amount && !facilityGuaranteeDates?.guaranteeExpiryDate) {
+      throw new Error('Invalid argument set provided');
+    }
+
+    if (amount) {
```
abhi-markan, line 26: Can this please be written in following style to match `facility-loan-amend`

```javascript
        record = {
          ...record,
          guaranteedLimit: to2Decimals(amount),
        };
```

azure-functions/acbs-function/mappings/facility/facility-guarantee-amend.js
```diff
@@ -0,0 +1,42 @@
...
+ * @param {object} amendment - The amendment details.
+ * @param {number|string=} amendment.amount - The amount to be amended, can be a number or a string. It is required if guaranteeExpiryDate is not provided.
+ * @param {object=} amendment.facilityGuaranteeDates - The dates related to the facility guarantee.
+ * @param {string=} amendment.facilityGuaranteeDates.guaranteeExpiryDate - The expiry date of the guarantee in 'YYYY-MM-DD' format. It is required if amount is not provided.
+ * @returns {MappedFacilityGuaranteeAmendment | {} } - The amended facility guarantee record, or an empty object if there is an error.
+ */
+const facilityGuaranteeAmend = (amendment) => {
+  try {
+    const { amount, facilityGuaranteeDates } = amendment;
+    const record = {};
```
abhi-markan, line 20: const { guaranteeExpiryDate } = facilityGuaranteeDates;

azure-functions/acbs-function/mappings/facility/facility-guarantee-amend.test.js
```diff
@@ -0,0 +1,46 @@
...
+    });
+  });
+
+  describe('when there is no amendment amount or guarantee expiry date', () => {
+    const amendment = {};
+    const result = facilityGuaranteeAmend(amendment);
+    expect(result).toEqual({});
+  });
+
+  describe('when there are extra fields', () => {
```
abhi-markan, line 35: Can you please add an assertion for following:

* Exceptional handling
* When `facilityGuaranteeDates` has an empty object.
* When the amount is in wrong data type.

azure-functions/acbs-function/src/functions/activity-update-facility-guarantee.js
```diff
@@ -0,0 +1,80 @@
...
+ * @throws {Error} - Throws an error if the payload is invalid, if the API request to generate the guarantee ID fails, or if any other error occurs.
+ */
+const handler = async (payload) => {
+  try {
+    if (!payload) {
+      throw new Error('Invalid facility guarantee update payload');
+    }
+
+    const { facilityIdentifier, acbsFacilityGuaranteeInput } = payload;
+
```
abhi-markan, line 43: Missing mandatory field check, please refer to `activity-update-facility-loan-amount.js`

```javascript
const missingMandatory = findMissingMandatory(acbsFacilityLoanInput, mandatoryFields);
```

azure-functions/acbs-function/src/functions/activity-update-facility-guarantee.js
```diff
@@ -0,0 +1,80 @@
...
+            dataReceived: data,
+            dataSent: acbsFacilityGuaranteeInput,
+          },
+          null,
+          4,
+        ),
+      );
+    }
+
+    return {
```
abhi-markan, line 65: In the following order please

```suggestion
 return {
      status,
      submittedToACBS,
      receivedFromACBS: getNowAsIsoString(),
      dataSent: acbsFacilityGuaranteeInput,
      ...data,
    };
```

azure-functions/acbs-function/src/functions/activity-update-facility-guarantee.js
```diff
@@ -0,0 +1,80 @@
...
+      status,
+      dataSent: acbsFacilityGuaranteeInput,
+      submittedToACBS,
+      receivedFromACBS: getNowAsIsoString(),
+      ...data,
+    };
+  } catch (error) {
+    console.error('Unable to create facility guarantee record %o', error);
```
abhi-markan, line 73:
```suggestion
    console.error('Unable to amend facility guarantee record %o', error);
```

azure-functions/acbs-function/src/functions/activity-update-facility-guarantee.js
```diff
@@ -0,0 +1,80 @@
...
+    throw new Error(`Unable to create facility guarantee record ${error}`);
```
abhi-markan, line 74:
```suggestion
    throw new Error(`Unable to amend facility guarantee record ${error}`);
```

abhi-markan (changes requested):

azure-functions/acbs-function/mappings/facility/facility-guarantee-amend.js
AlexBramhill: We shouldn't look to destruct here as amendment may not have `facilityGuaranteeDates` on, so we might throw if we try and access this by destructuring. We could look a throwing in some defaults when destructing, but this is a pattern I've not seen in acbs

azure-functions/acbs-function/mappings/facility/facility-guarantee-amend.js
AlexBramhill: updated

azure-functions/acbs-function/src/functions/activity-update-facility-guarantee.js
AlexBramhill: Three are no mandatory fields here -- both are optional

package.json
```diff
@@ -32,6 +32,7 @@
     "lint:all": "npm run lint --workspaces --if-present",
     "lint:fix:all": "npm run lint --workspaces --if-present -- --fix",
     "load": "(cd utils && npm run load)",
+    "prepare": "husky",
```
AlexBramhill, line 35: This does not replace any existing commits when run (runs automatically on `npm i`), at least when I've tested this

azure-functions/acbs-function/mappings/facility/facility-guarantee-amend.test.js
AlexBramhill: We check exception handling -- existing pattern is to catch and return an empty object (I've added an `it` to the test which was missing!)

I've added a check when `facilityGuaranteeDates` is an empty object

Regarding `amount` is the wrong datatype -- this is currently not handled at all well (one of the reasons I looked at adding zod!). Technically, as this is a unit test, we should be mocking `to2Decimals` -- I'll update in line with this

azure-functions/acbs-function/src/functions/activity-update-facility-guarantee.js
abhi-markan: Understood, maybe we should do below to ensure we are not making an API call without any fields to get `400` back. What are your thoughts?
let mandatoryFields = [];

```javascript
if (acbsFacilityGuaranteeInput.expirationDate) {
mandatoryFields.push('expirationDate');
}

if (acbsFacilityGuaranteeInput.amount) {
mandatoryFields.push('amount');
}
```

azure-functions/acbs-function/mappings/facility/facility-guarantee-amend.test.js
abhi-markan: Understood regarding above points thanks for that.
Wondering if we can remove this `to2DecimalsResponse` and rely on the actual response?

abhi-markan (commented):
Nice work, minor comments to go through please. Thanks

azure-functions/acbs-function/src/functions/activity-update-facility-guarantee.js
AlexBramhill: We had a similar converstaion for the previous 6616 ticket, where it was decided we'd add this verification into the `facility-covenant-amend` file, so I've added the similar into the `facility-guarantee-amend` file
```
    if (!amount && !facilityGuaranteeDates?.guaranteeExpiryDate) {
      throw new Error('Invalid argument set provided');
    }
```

The suggested code
```
if (acbsFacilityGuaranteeInput.fieldExample) {  
mandatoryFields.push('fieldExample');  
}  
```
wouldn't do anything I think? As mandatoryField function just checks to see if `acbsFacilityGuaranteeInput` contains the fields passed in. As we are driving `mandatoryFields` by what exists in `acbsFacilityGuaranteeInput`, we are saying "Heres a list of fields that I know exist in `acbsFacilityGuaranteeInput` as that is how I've got the list of them, please check they exist in `acbsFacilityGuaranteeInput`

AlexBramhill: There is an issue here with existing patterns and how we handle errors. As both fields are mandatory, the agreed patterns for where we look to validate means that this will happily fail silently. 

I'm going to update this to work for now and try to stick to the existing error handling patterns as much as possible, but this will greatly limit where I can add these checks. I've added jsdoc to clarify the structure of the mapped amendments at any point.

azure-functions/acbs-function/src/functions/activity-update-facility-covenant.js
```diff
@@ -41,6 +41,12 @@ const handler = async (payload) => {
 
     const { facilityIdentifier, acbsFacilityCovenantInput } = payload;
 
+    // As both fields on the payload are optional,
```
abhi-markan, line 44: Can we please remove this as this introduces another mechanism to validate a payload, though a good catch for a payload where both fields can be optional.

azure-functions/acbs-function/src/functions/activity-update-facility-guarantee.js
abhi-markan: Good shout on updating the JSDocs, agree on the limitation.

abhi-markan (changes requested):

azure-functions/acbs-function/src/functions/activity-update-facility-covenant.js
AlexBramhill: updated

abhi-markan (approved):

PR #3328, opened 2024-07-23
chore(dtfs2-none): refactor create facility amendment to improve error handling

AlexBramhill:
## Introduction :pencil2:
There is a bug in remote dev that is hard to pinpoint

This PR reformates the area of the code, as well as adding wider middleware


trade-finance-manager-api/src/v1/controllers/amendment.controller.js
```diff
@@ -179,13 +181,24 @@ const getAllAmendments = async (req, res) => {
...
+      throw new AmendmentNotCreatedError(facilityId);
+    }
+
     return res.status(200).send({ amendmentId });
+  } catch (error) {
+    if (isAxiosError(error)) {
+      return res.status(error?.response?.status || 500).json({ data: 'Failed to create facility amendment' });
+    }
+
+    return next(error);
```
AlexBramhill, line 200: This handles all our apiErrors automatically for us now with global middleware

MarRobSoftwire: Does `throw error` not work?

MarRobSoftwire: Seen your comment elsewhere now this makes sense plz ignore

MarRobSoftwire (dismissed):

MarRobSoftwire (approved):

trade-finance-manager-api/src/v1/api.js
```diff
@@ -340,25 +340,22 @@ const updateFacility = async ({ facilityId, tfmUpdate, auditDetails }) => {
...
 
-      return response.data;
-    } catch (error) {
-      console.error('Error creating facility amendment %o', error);
-      return { status: error?.response?.status || 500, data: 'Failed to create facility amendment' };
-    }
-  } else {
-    console.error('Invalid facilityId provided');
-    return { status: 400, data: 'Invalid facility id' };
+  if (!hasValidUri(DTFS_CENTRAL_API_URL)) {
```
abhi-markan, line 347: Can this be applied to other endpoints as well please, appreciate a new PR will be required for this.

trade-finance-manager-api/src/v1/api.js
```diff
@@ -340,25 +340,22 @@ const updateFacility = async ({ facilityId, tfmUpdate, auditDetails }) => {
 };
 
 const createFacilityAmendment = async (facilityId, auditDetails) => {
```
abhi-markan, line 342: Can we wrap this in `try-catch` block please, in case the API call fails.

portal-api/src/v1/gef/controllers/facilities.controller.js
```diff
@@ -304,8 +304,8 @@ exports.deleteByDealId = async (req, res) => {
     });
     return res.status(200).send(response);
   } catch (error) {
-    // This implimentation of returning 200 if the document is not found
-    // is to preserve existing implimentation
+    // This implementation of returning 200 if the document is not found
+    // is to preserve existing implementation
     if (error instanceof DocumentNotFoundError) {
       return res.sendStatus(200);
```
abhi-markan, line 310: HttpStatusCode please.

libs/common/src/middleware/problem-with-service.middleware.ts
```diff
@@ -0,0 +1,8 @@
+import { ErrorRequestHandler } from 'express';
+
+/**
+ * Frontend middleware that renders problem with service if other middlewares do not handle the error.
+ */
+export const problemWithService: ErrorRequestHandler = (error: { message?: string }, _req, res) => {
```
abhi-markan, line 6: Please remove `_req` variable.

libs/common/src/middleware/problem-with-service.middleware.ts
```diff
@@ -0,0 +1,8 @@
...
+  res.render('partials/problem-with-service.njk', { error });
```
abhi-markan, line 7: Might be good to add `console.error`.

libs/common/src/middleware/api-error-handling.middleware.ts
```diff
@@ -0,0 +1,14 @@
+import { ErrorRequestHandler } from 'express';
+import { ApiError } from '../errors';
+
+/**
+ * Middleware to handle errors of the ApiError type at a global level
+ * Note: Express 4.x requires the route function to be wrapped in a try catch block
+ * and to call `next(error)` to pass the error for this middleware to work
+ */
+export const apiErrorHandling: ErrorRequestHandler = (error, _req, res, next) => {
```
abhi-markan, line 9: Please remove `_req`

libs/common/src/errors/amendment-not-created.error.ts
```diff
@@ -0,0 +1,11 @@
+import { HttpStatusCode } from 'axios';
+import { ApiError } from './api.error';
+
+export class AmendmentNotCreatedError extends ApiError {
+  constructor(facilityId?: string) {
+    const message = facilityId ? `Amendment not created for facility: ${facilityId}` : 'Amendment not created';
```
abhi-markan, line 6: This is a good touch to the exceptional handling, few comments:

* My understanding is `facilityId` should always exists for an amendment to be processed, can it be please made mandatory.
* Can we also specify why it failed, i.e. unable to get amendment ID, API call and etc.

abhi-markan (changes requested):

libs/common/src/middleware/csrf-error-handling.middleware.ts
```diff
@@ -0,0 +1,15 @@
+import { ErrorRequestHandler } from 'express';
+import { isHttpError } from 'http-errors';
+
+/**
+ * Frontend middleware that handles CSRF token errors.
+ */
+export const csrfErrorHandling: ErrorRequestHandler = (error, _req, res, next) => {
```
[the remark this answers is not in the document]
AlexBramhill: I think `_` is always good to show intentionally not using a parameter, but happy to remove

trade-finance-manager-api/src/v1/api.js
AlexBramhill: We don't want to I think -- as we handle `AxiosError` at the controller level (changing the response), and all the `ApiErrors` at the global middleware level -- Catching it here would still mean we'd have to handle these errors at the controller level, and we'd want to keep the grainularity of the errors, so we'd just rethrow

trade-finance-manager-api/src/v1/api.js
AlexBramhill: I'll raise another PR for this, but this validation should take place on application startup, not every time we hit an end point as its a static constant :)

trade-finance-manager-api/src/v1/controllers/amendment.controller.js
```diff
@@ -179,13 +181,24 @@ const getAllAmendments = async (req, res) => {
...
+
+    const { amendmentId } = await api.createFacilityAmendment(facilityId, generateTfmAuditDetails(req.user._id));
+
+    if (!amendmentId) {
+      throw new AmendmentNotCreatedError(facilityId);
+    }
+
     return res.status(200).send({ amendmentId });
+  } catch (error) {
+    if (isAxiosError(error)) {
```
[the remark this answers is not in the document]
AlexBramhill: Yep -- Due to error handling. I've mentioned in another comment -- but we're handling our known errors here and in global middleware (and unknown errors are being treated as unhandled exceptions, and get handled as a 500, which is standard)

trade-finance-manager-api/src/v1/controllers/amendment.controller.js
```diff
@@ -179,13 +181,24 @@ const getAllAmendments = async (req, res) => {
...
 };
 
-const createFacilityAmendment = async (req, res) => {
-  const { facilityId } = req.body;
-  const { amendmentId } = await api.createFacilityAmendment(facilityId, generateTfmAuditDetails(req.user._id));
-  if (amendmentId) {
+const createFacilityAmendment = async (req, res, next) => {
+  try {
+    const { facilityId } = req.body;
+
```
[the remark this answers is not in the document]
AlexBramhill: We do in the api call itelf, but I agree we should validate payload as it enters rather than as it leaves

trade-finance-manager-api/src/v1/api.js
```diff
@@ -340,25 +340,22 @@ const updateFacility = async ({ facilityId, tfmUpdate, auditDetails }) => {
...
   }
+
+  const response = await axios({
+    method: 'post',
+    url: `${DTFS_CENTRAL_API_URL}/v1/tfm/facilities/${facilityId}/amendments`,
+    headers: headers.central,
+    data: { auditDetails },
+  });
+
+  return response.data;
```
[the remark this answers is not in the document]
AlexBramhill: It cant be with axios :)

libs/common/src/middleware/csrf-error-handling.middleware.ts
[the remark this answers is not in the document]
abhi-markan: As discussed earlier DTFS codebase has stayed away from using prefixed `_` variable.

trade-finance-manager-api/src/v1/controllers/amendment.controller.js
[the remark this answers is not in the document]
abhi-markan: Understood thanks for the explanation.

libs/common/src/middleware/api-error-handling.middleware.ts
```diff
@@ -0,0 +1,14 @@
+import { ErrorRequestHandler } from 'express';
+import { ApiError } from '../errors';
+
+/**
+ * Middleware to handle errors of the ApiError type at a global level
+ * Note: Express 4.x requires the route function to be wrapped in a try catch block
+ * and to call `next(error)` to pass the error for this middleware to work
+ */
+export const apiErrorHandling: ErrorRequestHandler = (error, req, res, next) => {
```
abhi-markan, line 9: * Please add types to arguments.
```typescript
import { ErrorRequestHandler, Request, Response, NextFunction } from 'express';

error: unknown, req: Request, res: Response, next: NextFunction
```
* This function doesn't need to explicitly return anything. However, it's good practice to annotate the function correctly for clarity and type safety. i.e. `(error, req, res, next): void`

libs/common/src/middleware/csrf-error-handling.middleware.ts
```diff
@@ -0,0 +1,16 @@
+import { HttpStatusCode } from 'axios';
+import { ErrorRequestHandler } from 'express';
+import { isHttpError } from 'http-errors';
+
+/**
+ * Frontend middleware that handles CSRF token errors.
+ */
+export const csrfErrorHandling: ErrorRequestHandler = (error, req, res, next) => {
```
abhi-markan, line 8: Same as above please.

trade-finance-manager-api/src/v1/controllers/amendment.controller.js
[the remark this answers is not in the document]
abhi-markan: Thanks

libs/common/src/errors/amendment-not-created.error.ts
```diff
@@ -0,0 +1,11 @@
+import { HttpStatusCode } from 'axios';
+import { ApiError } from './api.error';
+
+export class AmendmentNotCreatedError extends ApiError {
```
abhi-markan, line 4: One liner documentation will be a nice touch.

abhi-markan (changes requested):

AlexBramhill:
This PR no longer has priority, and is being put to the side for now.

When picking this up there's some thought to be had about how we integrate this pattern into existing codebase.

PR #3331, opened 2024-07-24
feat(FN-1725): tfm amendments payload validation and types

JoshBinns2000:
## Introduction :pencil2:
We want to add some payload validation to tfm amendments to get some information about the types of the TFM facilties collection. We also want to improve error logging in existing controllers.

- Introduces commonised logic for creating zod payload validators and logging the validation errors
- Adds payload validation for POST and PUT amendments
- Adds partial db-model types to libs/common*
- Upgrades some controller files to typescript
- Updates some constants in `dtfs-central-api` and adds the `AMENDMENT_STATUS` constant to `libs/common` along with an `AmendmentStatus` type
- Removes a lot of the unnecessary logic in the `tfm-get-amendments` central-api controller 
- Updates error handling and logging in controllers to make it easier to see what error occurred

*these types are based on observations about requests that are sent to the endpoints where this validation has been added. As new properties are discovered, they should be added to these types


libs/common/src/types/mongo-db-models/facility.ts
```diff
@@ -0,0 +1,44 @@
...
+ */
+export type Facility = {
+  _id: ObjectId;
+  dealId: ObjectId;
+  type: string;
+  hasBeenIssued: boolean;
+  name: string;
+  shouldCoverStartOnSubmission: boolean;
+  coverStartDate: UnixTimestamp | null;
+  coverEndDate: UnixTimestamp | null;
```
AlexBramhill, line 19: just a reminder as talked about on slack, this is stored as a date, and I reckon it's mongodb storing it as a string that secretly is a date (`$date` field), and it may or may not be parsable as a `number` straight off the bat

dtfs-central-api/src/repositories/tfm-facilities-repo/tfm-facility.schema.ts
```diff
@@ -2,9 +2,8 @@ import z from 'zod';
 import { MongoObjectIdSchema } from '../../v1/routes/middleware/payload-validation/schemas';
 
 const TfmFacilityDateFieldSchema = z
```
AlexBramhill, line 4: Little bit worried about this if it's not tested for `coverEndDate` -- Could we get some tests testing for all methods of how we store date in our database plz? :)

dtfs-central-api/src/v1/routes/middleware/payload-validation/create-validation-middleware-for-schema.ts
```diff
@@ -0,0 +1,18 @@
+import { AnyZodObject, ZodError } from 'zod';
+import { RequestHandler } from 'express';
+import { HttpStatusCode } from 'axios';
+
+const getFormattedZodErrors = (error: ZodError): string[] => error.issues.map(({ path, message, code }) => `${path.join('.')}: ${message} (${code})`);
+
+export const createValidationMiddlewareForSchema =
```
AlexBramhill, line 7: nice pattern

dtfs-central-api/src/v1/routes/middleware/payload-validation/create-validation-middleware-for-schema.ts
```diff
@@ -0,0 +1,18 @@
...
+  <TSchema extends AnyZodObject>(schema: TSchema): RequestHandler =>
+  (req, res, next) => {
+    const { success, error, data } = schema.safeParse(req.body);
+    if (success) {
+      req.body = data;
+      return next();
+    }
+    const formattedErrors = getFormattedZodErrors(error);
+    console.error('Payload validation error occurred:', formattedErrors);
+    return res.status(HttpStatusCode.BadRequest).send(formattedErrors);
```
AlexBramhill, line 17: interested why youve gone for safe parse as apposed to try catch (not saying either is right or wrong) -- for the code simplicity?

libs/common/src/types/mongo-db-models/facility.ts
JoshBinns2000: Turns out it is either a `UnixTimestampString` or a `Date` based on the staging database. Fortunately, any operations which use this date (multiplying it by 1000 usually) work whether the value is a string or a date in this case

dtfs-central-api/src/v1/routes/middleware/payload-validation/create-validation-middleware-for-schema.ts
JoshBinns2000: Mainly for readability. I take the view that try-catch blocks are usually quite tough to read because the flow of the code gets disrupted if something you call throws an error. For example, between the two below blocks of code, I would probably choose the former:

```typescript
const amendment = await TfmFacilityRepo.findAmendmentsByAmendmentId('abc123');
if (!amendment) {
  return res.status(404).send("Failed to find an amendment with id 'abc123'");
}
return res.status(200).send(amendment);
```

```typescript
try {
  const amendment = await TfmFacilityRepo.findAmendmentsByAmendmentId('abc123');
  if (!amendment) {
    throw new NotFoundError("Failed to find an amendment with id 'abc123'");
  }
  return res.status(200).send(amendment);
} catch (error: NotFoundError) { // we know the error type here
  const { status, message } = error;
  return res.status(status).send(message);
}
``` 

I think the utility of try-catch starts to come when you have functions which can throw many different types of errors and you want to put the handling of those errors into one place. At that point, it makes more sense to me to lose the readability of more explicit checks for the benefit of easier to read and maintain error handling. In the case where we only have one check here though, I think having code which can be easily read line-by-line is better

libs/common/src/types/mongo-db-models/facility.ts
```diff
@@ -0,0 +1,44 @@
+import { ObjectId } from 'mongodb';
+import { UnixTimestamp } from '../date';
+import { Currency } from '../currency';
+
+/**
+ * Type of the mongo db "facilities" collection
```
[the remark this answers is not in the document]
JoshBinns2000: I've made the decision to remove the parsing methods from the tfm facilities repo so that we defer to this type as the default from now on and don't have two sources of truth to maintain

dtfs-central-api/src/repositories/tfm-facilities-repo/tfm-facilities.repo.test.ts
```diff
@@ -1,6 +1,7 @@
 import { ObjectId } from 'mongodb';
 import { ZodError } from 'zod';
-import { ParsedTfmFacility, RawTfmFacility, TfmFacilitiesRepo } from './tfm-facilities.repo';
+import { TfmFacility } from '@ukef/dtfs2-common';
+import { ParsedTfmFacility, TfmFacilitiesRepo } from './tfm-facilities.repo';
```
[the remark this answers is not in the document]
JoshBinns2000: See [this comment](https://github.com/UK-Export-Finance/dtfs2/pull/3331#discussion_r1692645394)

dtfs-central-api/src/repositories/tfm-facilities-repo/tfm-facility.schema.ts
JoshBinns2000: See [this comment](https://github.com/UK-Export-Finance/dtfs2/pull/3331#discussion_r1692645394)

dtfs-central-api/api-tests/v1/tfm/facilities/tfm-facilities-get-facilities-for-deal.api-test.js
```diff
@@ -25,9 +25,8 @@ describe('/v1/tfm/deals/:id/facilities', () => {
   describe('GET /v1/tfm/deal/:id/facilities', () => {
     it('returns a 400 error if the id is not a valid mongo id', async () => {
       const invalidMongoId = 'abc';
-      const { status, body } = await testApi.get(`/v1/tfm/deals/${invalidMongoId}/facilities`);
+      const { status } = await testApi.get(`/v1/tfm/deals/${invalidMongoId}/facilities`);
```
AlexBramhill, line 28: any reason for this>

dtfs-central-api/api-tests/helpers/with-validate-audit-details.api-tests.ts
```diff
@@ -28,7 +28,13 @@ export const withValidateAuditDetailsTests = ({ makeRequest, validUserTypes = AU
     it('should return 400 if no auditDetails provided', async () => {
       const { status, body } = await makeRequest();
 
-      expect(body).toEqual({ status: 400, message: 'Invalid auditDetails, Missing property `userType`' });
+      // Error body can either come from payload validation or controller validation
+      if (Array.isArray(body)) {
+        expect(body).toEqual(['auditDetails: Invalid input (invalid_union)']);
```
AlexBramhill, line 33: do you think we should look to return the same thing for both, as they are the same errors

JoshBinns2000: Technically we could, but the way to do that would involve overwriting the existing logic around how we validate audit details as we don't have loads of control over the return type of the zod validator. I will try to make it consistent across services though

dtfs-central-api/api-tests/v1/tfm/facilities/tfm-facilities-get-facilities-for-deal.api-test.js
JoshBinns2000: Because the test only specifies that we return a 400, and the actual response body comes from express-validator should not really matter for the sake of the test (I also really dislike the `res.status(400).send({ status: 400, message: '...' })` pattern)

AlexBramhill: Do you think we should test the error message anyway, in case it's used in future downstream?

dtfs-central-api/api-tests/helpers/with-validate-audit-details.api-tests.ts
AlexBramhill: Thank you -- Even if we just try/catch the zod errors and throw the other error it might be better

dtfs-central-api/api-tests/v1/tfm/facilities/tfm-facilities-get-facilities-for-deal.api-test.js
JoshBinns2000: Not at the api-test level - I think the status communicates more to us than the body of the error at the api-test level, and testing the body of the error would be a test of the express-validator package rather than a test of our implementation

dtfs-central-api/api-tests/helpers/with-validate-audit-details.api-tests.ts
JoshBinns2000: try-catching the zod error won't fix this because it's to do with how I format the error. I also think the `"Invalid auditDetails, Missing property 'userType'"` is misleading because the validation can fail for different reasons. Like I say, I'll try to come up with something that works

dtfs-central-api/api-tests/v1/tfm/facilities/tfm-facilities-get-facilities-for-deal.api-test.js
AlexBramhill: I'd somewhat agree except `400` is generic, and there has been issues on this codebase where tests similar to this are passing but really its an issue with the test configuration.

I get what you mean that it should be tested in the express validator package, I think that is a fair point... I'd still like to see it to document it's working as intended though?

dtfs-central-api/src/v1/validation/route-validators/route-validators.js
```diff
@@ -5,9 +5,14 @@ const bankIdValidation = param('bankId').isString().matches(/^\d+$/).withMessage
 
 exports.bankIdValidation = [bankIdValidation];
 
-const mongoIdValidation = param('_id').isMongoId().withMessage("Invalid MongoDB '_id' path param provided");
+/**
+ * Validator for a path parameter which is a mongo id
+ * @param {string} paramName
+ * @returns {import('express-validator').ValidationChain}
+ */
+const mongoIdValidation = (paramName) => param(paramName).isMongoId().withMessage("Invalid MongoDB '_id' path param provided");
```
AlexBramhill, line 13: We sometimes have multiple mongoIds in one URL. Do we want to add the param name to the message

JoshBinns2000: Ah yes, that was my mistake. I'll update that

dtfs-central-api/src/v1/controllers/cron-jobs/cron-jobs.controller.ts
```diff
@@ -29,6 +19,13 @@ export const deleteAllEstoreLogs = async (req: CustomExpressRequest<{ reqBody: {
 
     return res.status(200).send();
   } catch (error) {
+    if (error instanceof InvalidAuditDetailsError) {
```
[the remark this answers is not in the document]
JoshBinns2000: I think I disagree - having multiple try-catch blocks seems to be against the spirit of what a try-catch is. In this case specifically, the only reason it would be unclear that the error came from the `validateAuditDetails` function is that it is not immediately obvious that that function throws. I prefer to call functions which throw on error something like `assert<Thing>IsValid` or to write

```typescript
const validationError = validateAuditDetails(auditDetails);
if (validationError) {
  return res.status(HttpStatusCode.BadRequest).send(validationError);
}
```

However, in this case, we do have specific error handling around whether or not the error is linked to the audit details and I think that does make it obvious without the tightly coupled try-catch blocks. If we wanted the tightly coupled try-catch blocks as an option, I would prefer to go with the cope snippet above

AlexBramhill (approved):

dtfs-central-api/api-tests/v1/tfm/facilities/tfm-facilities-get-facilities-for-deal.api-test.js
JoshBinns2000: Discussed offline - work to assure the correct errors are thrown will be addressed in a follow-up PR

PR #3335, opened 2024-07-26
feat(FN-1848): take tolerances into account when checking for match status

RLavender98:
## Introduction :pencil2:
Use tolerances when checking if reported payments and received payments match for a group of fee records.

- Fetch the active tolerances from the database when checking for match status
- A group should be considered at match status if the difference between the reported payments and received payments is less than or equal to the tolerance for that currency
-  A group should not be considered at match status if the difference between the reported payments and received payments is greater than the tolerance for that currency


dtfs-central-api/src/services/utilisation-reports/payment-matching-tolerance.service/index.ts
JoshBinns2000, line 1: Why has this been added as a service instead of a repo?

dtfs-central-api/src/services/utilisation-reports/payment-matching-tolerance.service/index.ts
```diff
@@ -0,0 +1,21 @@
...
+import { PaymentMatchingTolerances } from '../../../types/payment-matching-tolerances';
+import { NotFoundError } from '../../../errors';
+
+const mapActivePaymentMatchingToleranceEntitiesToPaymentMatchingTolerances = (
+  activeToleranceEntities: PaymentMatchingToleranceEntity[],
+): PaymentMatchingTolerances => {
+  return Object.values(CURRENCY).reduce((partialTolerances, currency) => {
+    const currencyTolerance = activeToleranceEntities.find((tolerance) => tolerance.currency === currency);
+    if (!currencyTolerance) {
+      throw new NotFoundError(`No tolerance found for currency ${currency}`);
```
JoshBinns2000, line 12: Why do we want to load every tolerance every time we call this method? Based on the implementation, I think we should only ever need one tolerance per currency, so throwing an error for the `EUR` currency when we're looking at `GBP` payments may be confusing

e2e-tests/e2e-fixtures/payment-matching-tolerances.fixture.ts
```diff
@@ -0,0 +1,13 @@
+import { CURRENCY, PaymentMatchingToleranceEntityMockBuilder } from '@ukef/dtfs2-common';
+
+const toleranceIdGenerator = (function* idGenerator(): Generator<number, number, unknown> {
```
JoshBinns2000, line 3: Can't you use an anonymous function here?

JoshBinns2000 (commented):
Looking good! My main concern is around adding the fetching tolerances as a `service` as it feels slightly out of place compared to our existing patterns

dtfs-central-api/src/services/utilisation-reports/payment-matching-tolerance.service/index.ts
RLavender98: Aha I actually did first add it as a repo but then I wasn't sure if I liked it because it is quite different to our other SQL repos which all use extended typeorm repos and return entities whereas this using transaction entity manager and maps out of entity form before returning and my thought was down the line when we want to actually allow users to edit tolerances etc we might want a repo of that style. That said if you think this would be better named as and living with the repos then I can move it?

dtfs-central-api/src/services/utilisation-reports/payment-matching-tolerance.service/index.ts
RLavender98: That's a good point. I think originally I wrote it as passing in the tolerances and thought we had cases where we might call this many times in one endpoint so wanted to write it like that so we only fetch the tolerances once. I then realised actually that many check case doesn't actually exist so moved it in so I think that is a good refactor suggestion, I'll change it!

dtfs-central-api/src/services/utilisation-reports/payment-matching-tolerance.service/index.ts
JoshBinns2000: If we're now making it so that we only query for a specific currency, then I think a repo method does work better as we shouldn't need the mapping logic anymore. And if we want to use the entity manager, I think the one-line `entityManager.findOneBy(PaymentMatchingToleranceEntity, { isActive: true, currency: <currency> })` rather than a repo might itself be fine.

This has also given me an idea for how we may want to implement repo methods with transactions which I'll message you about offline

dtfs-central-api/src/helpers/fee-record-matching.test.ts
```diff
@@ -615,12 +634,9 @@ describe('fee-record-matching', () => {
 
             const payments = [PaymentEntityMockBuilder.forCurrency(paymentCurrency).withAmount(paymentAmount).build()];
 
-            jest.mocked(getActivePaymentMatchingTolerances).mockResolvedValue({
-              USD: 0,
-              JPY: 0,
-              EUR: 0,
-              GBP: tolerance,
-            });
+            mockFindOneByCurrencyAndIsActiveTrue.mockResolvedValue(
```
JoshBinns2000, line 637: Just out of curiosity, do you think there is a benefit to using `jest-when` here? I haven't been able to draw a line in my head yet where we should use `jest-when` and where it is excessive, so I'm curious to hear your thoughts

JoshBinns2000 (approved):
LGTM - the one comment is just a question, so going to approve anyway

e2e-tests/e2e-fixtures/payment-matching-tolerances.fixture.ts
JoshBinns2000: Yeh I thought that might be the case - oh well!

dtfs-central-api/src/helpers/fee-record-matching.test.ts
RLavender98: So I don't think it really matters? I think my opinion is that if I was asserting the call in every test then it would be better to use `jest-when`, instead to remove all those additional assertions, instead I have one test that specifically fails if we call it with the wrong parameters and then the others will only fail for separate reasons which I like since it aligns with the test only failing for one reason (or as little reasons as possible) which we strive for in unit tests.

JoshBinns2000: That sounds along the same lines of thinking I have, so sounds good to me!

PR #3336, opened 2024-07-26
fix(DTFS2-7292): downgraded MSSQL to 2022

abhi-markan:
## Introduction :pencil2:
A recent update on Microsoft MSSQL version especially `2022` have path and driver updates.
This has broken the pipeline due to absence of any health check for `dtfs-sql` container pipeline debugging has been inaccurate.

This PR aims to mitigate both the issues.

* Downgraded MSSQL docker image to explicit version `mcr.microsoft.com/mssql/server:2022-CU13-ubuntu-22.04`
* Refactored `init.sh` for MSSQL server boot.
* Added health check to `dtfs-sql` container.

* Renamed `dtfs-submissions-data` to `dtfs-mongo` for consistent naming convention.

### DIFF
--- a/libs/common/src/sql-db-connection/Dockerfile
+++ b/libs/common/src/sql-db-connection/Dockerfile
@@ -1,5 +1,5 @@
 # Base image
-FROM mcr.microsoft.com/mssql/server:2019-latest
+FROM mcr.microsoft.com/mssql/server:2022-CU13-ubuntu-22.04
 
 # Setup
 COPY init.sh /
--- a/libs/common/src/sql-db-connection/Dockerfile
+++ b/libs/common/src/sql-db-connection/Dockerfile
@@ -1,5 +1,5 @@
 # Base image
-FROM mcr.microsoft.com/mssql/server
+FROM mcr.microsoft.com/mssql/server:2019-latest
 
 # Setup
 COPY init.sh /
--- a/libs/common/src/sql-db-connection/init.sh
+++ b/libs/common/src/sql-db-connection/init.sh
@@ -1,54 +1,42 @@
 #!/bin/bash
 set -m
-
-execute_sql_command() {
-  /opt/mssql-tools/bin/sqlcmd -S localhost -U sa -P "$MSSQL_SA_PASSWORD" -Q "$1" > /dev/null
-  return $?
-}
-
-get_now_unix_timestamp() {
-  echo $(date +%s)
-}
-
 /opt/mssql/bin/sqlservr &
-
-readonly time_limit_seconds=60
-readonly interval_seconds=5
-readonly start_time=$(get_now_unix_timestamp)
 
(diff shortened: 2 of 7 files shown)


libs/common/src/sql-db-connection/init.sh
JoshBinns2000, line 1: I don't think we should change this file. What was the reason to remove the start-up time limit?

libs/common/src/sql-db-connection/Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Base image
-FROM mcr.microsoft.com/mssql/server
+FROM mcr.microsoft.com/mssql/server:2019-latest
```
JoshBinns2000, line 2: Is there any reason to not use the more recent `mcr.microsoft.com/mssql/server:2022-CU13-ubuntu-22.04` image?

libs/common/src/sql-db-connection/init.sh
```diff
@@ -1,54 +1,42 @@
...
-do
-  current_time=$(get_now_unix_timestamp)
-  elapsed_seconds=$((current_time - start_time))
-
-  if [ $elapsed_seconds -gt $time_limit_seconds ]; then
-    echo "❌ Error: Failed to create database '$SQL_DB_NAME' - SQL Server did not start within $time_limit_seconds seconds."
-    exit 1
-  fi
+# Executes SQL query
+query() {
```
JoshBinns2000, line 8: If we want to change `command` to `query`, I think calling the function `execute_sql_query` is probably more clear

libs/common/src/sql-db-connection/init.sh
```diff
@@ -1,54 +1,42 @@
...
-  echo "❌ MSSQL failed to start after $elapsed_seconds seconds. Waiting for $interval_seconds more seconds."
-  sleep $interval_seconds
-done
+# Initial DB setup
+initialise() {
+  # 1. Database
+  query "CREATE DATABASE [$SQL_DB_NAME];"
+  query "USE [$SQL_DB_NAME];"
 
-echo "✅ MSSQL initialised, creating database '$SQL_DB_NAME'..."
```
JoshBinns2000, line 36: I think this logging is quite helpful when looking inside the container as the other logs aren't very descriptive. Is there a reason to remove them?

libs/common/src/sql-db-connection/init.sh
abhi-markan: We discussed this during our conversation.

libs/common/src/sql-db-connection/init.sh
abhi-markan: I have kept the logging in place and added an extra line to distinguish between when the MSSQL service is running and when it has been initialised.

JoshBinns2000 (approved):

PR #3337, opened 2024-07-29
feat(FN-3195): add script to mark fee records of complete reports as reconciled

RLavender98:
## Introduction :pencil2:
We will have reports that were manually reconciled when we turn on the new reconciliation pages, and if the fee records are still at TO_DO status then this will cause errors if a user tries to view the page.

- Add SQL script to update fee records of reconciled reports to reconciled that we can run when we turn on the reconciliation journey

- Seed a manually completed report with the statuses we expect after running the script so that we have mimic old data for local dev going forwards


JoshBinns2000 (approved):
LGTM!

PR #3339, opened 2024-07-30
feat(FN-NONE): fix tfm scripts build

JoshBinns2000:
## Important ⚠️ 
This PR looks like it removes a large amount of code. What has actually happened is that when webpack builds the scripts, it outputs a file with a single line of code. This line can be very long, though, and prettier converted it into a much longer file. Re-running `npm run build` without formatting therefore appears to remove a massive number of lines despite not changing any of the code.

## Introduction :pencil2:
When we run the `npm run build` command in `trade-finance-manager-ui`, it rebuilds files that we previously formatted in the prettier PR. This PR rebuilds those files without formatting and updates the integrity validations so that running `npm run build` in the future will not re-update these files.

- Updates the files
- Updates the integrity validations

- Fixes some e2e test which failed due to the date being 31/07/2024


AlexBramhill (approved):

AlexBramhill (approved):

e2e-tests/tfm/cypress/e2e/journeys/case-amendments/add-bank-decision-proceed.spec.js
```diff
@@ -101,8 +101,8 @@ context('Amendments underwriting - add banks decision - proceed', () => {
     cy.url().should('contain', 'cover-end-date');
 
     amendmentsPage.amendmentCoverEndDateDayInput().clear().focused().type(dateConstants.tomorrowDay);
-    amendmentsPage.amendmentCoverEndDateMonthInput().clear().focused().type(dateConstants.todayMonth);
-    amendmentsPage.amendmentCoverEndDateYearInput().clear().focused().type(dateConstants.todayYear);
+    amendmentsPage.amendmentCoverEndDateMonthInput().clear().focused().type(dateConstants.tomorrowMonth);
```
abhi-markan, line 104: Will this fail when running tomorrow?
I think I have seen something similar before where an E2E was failing on a specific date.

abhi-markan (changes requested):
Good spot on the prettier formatted `JS` files, as part of this PR could you also please:
* Check `gef-ui` and `portal` please.
* Add `/public` directory to `.prettierignore` and ESLint ignore path, as I fear running `lint:all` locally will pretty fix these files again.

e2e-tests/tfm/cypress/e2e/journeys/case-amendments/add-bank-decision-proceed.spec.js
JoshBinns2000: No - line 103 uses the `tomorrowDay` constant, but the previous 104 and 105 were using `todayMonth` and `todayYear` which explains why it had been passing every day up to now

AlexBramhill (approved):

e2e-tests/gef/cypress/e2e/facility/about-facility.spec.js
```diff
@@ -217,13 +217,12 @@ context('About Facility Page', () => {
       cy.visit(relative(`/gef/application-details/${application.id}/facilities/${facilityId}/about-facility`));
       aboutFacility.facilityName().should('have.value', 'Name');
       aboutFacility.shouldCoverStartOnSubmissionNo().should('be.checked');
-      aboutFacility.coverStartDateDay().should('have.value', now.getDate());
-      aboutFacility.coverStartDateMonth().should('have.value', now.getMonth() + 1);
-      aboutFacility.coverStartDateYear().should('have.value', now.getFullYear());
-      aboutFacility.coverEndDateDay().should('have.value', now.getDate());
```
AlexBramhill, line 223: I'd be super interested to know if there was an issue with exactly one year in the future (hopefully not!) or if this is done for simplicity of reading

AlexBramhill (dismissed):
approved, feel free to pm me the answer to that question -- I'd prefer exactly one day in the future if theres no other reason but happy to merge as is if there is good reason

AlexBramhill (dismissed):

AlexBramhill (approved):

PR #3342, opened 2024-07-31
feat(FN-1725): add mongo db path parameter validation tests

JoshBinns2000:
## Introduction :pencil2:
We want a commonised test helper for how to test mongo id path parameter validation. We also want a way of differentiating such an error from other errors so that we can be sure that the error we receive is the one we expect.

- Adds a test helper to libs/common
- Adds a new `API_ERROR_CODE` constant with associated `ApiErrorCode` type to libs/common
- Replaces the existing `mongoIdValidation` with a custom validator which sets the error response body to a type matching the `ApiErrorResponseBody` type added to libs/common
- Updates audit details tests to test for the correct error being sent by the api
- Updates the payload validator to set the error code


libs/common/src/test-helpers/test-cases/mongo-id-path-paramater-validation-tests.ts
```diff
@@ -0,0 +1,46 @@
...
+interface ErrorResponse extends Response {
+  body: ApiErrorResponseBody;
+}
+
+type MongoIdPathParamaterValidationTestsParams = {
+  baseUrl: string;
+  makeRequest: <TResponse extends ErrorResponse>(url: string) => Promise<TResponse>;
+};
+
+const extractParameters = (url: string): string[] =>
```
AlexBramhill, line 16: clever!

dtfs-central-api/src/v1/controllers/portal/deal/delete-deal.controller.ts
```diff
@@ -20,10 +23,11 @@ export const deleteDeal = async (req: CustomExpressRequest<{ params: { id: strin
     if (error instanceof InvalidAuditDetailsError) {
```
MarRobSoftwire, line 23: My code oops but you could consolidate both try/catches and check for `ApiError` rather than `invalidAuditDetailsError` then `DocumentNotDeletedError`

MarRobSoftwire: / once alexs pr is in can do `next(error)` too

dtfs-central-api/src/v1/controllers/tfm/amendments/tfm-get-amendments.controller.ts
```diff
@@ -72,7 +72,7 @@ export const getAmendmentsByFacilityId = async (req: Request, res: Response) =>
       default:
         if (amendmentIdOrStatus) {
           if (!ObjectId.isValid(amendmentIdOrStatus)) {
-            return res.status(400).send({ status: 400, message: 'Invalid amendment Id' });
+            return res.status(400).send({ status: 400, message: 'Invalid amendment Id', code: API_ERROR_CODE.INVALID_MONGO_ID_PATH_PARAMETER });
```
MarRobSoftwire, line 75: could we throw an error here instead? the ApiError check exists below anyway

libs/common/src/test-helpers/index.ts
```diff
@@ -1 +1,3 @@
 export * from './mock-data';
+// './test-cases' uses requires mongo related functionality, so should be in a separate export
+// export * from './test-cases';
```
MarRobSoftwire, line 3: Should this be commented out?

libs/common/package.json
```diff
@@ -24,7 +24,8 @@
     "./change-stream": "./src/change-stream/index.ts",
     "./change-stream/test-helpers": "./src/change-stream/test-helpers/index.ts",
     "./payload-verification": "./src/helpers/payload-verification/index.ts",
-    "./schemas": "./src/schemas/index.ts"
+    "./schemas": "./src/schemas/index.ts",
+    "./test-cases": "./src/test-helpers/test-cases/index.ts"
```
MarRobSoftwire, line 28: is test cases descriptive enough? This is isolated because it uses mongodb so that should be made clearer?

MarRobSoftwire (commented):
Happy with the way audit details are being validated :) Commented on a few things I saw along the way

libs/common/src/types/api-error-response-body.ts
```diff
@@ -0,0 +1,7 @@
+import { ApiErrorCode } from './api-error-code';
+
+export type ApiErrorResponseBody = {
```
AlexBramhill, line 3: do we want to commonise this or not part of scope?

dtfs-central-api/api-tests/v1/tfm/amendments/tfm-amendments-put.api-test.js
```diff
@@ -11,6 +12,8 @@ const { createDeal } = require('../../../helpers/create-deal');
 const { createFacility } = require('../../../helpers/create-facility');
 const { aPortalUser, aTfmUser } = require('../../../../test-helpers/test-data');
 
+console.error = jest.fn();
```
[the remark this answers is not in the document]
JoshBinns2000: I don't think so - it's nice to suppress the expected console.error output in tests so that it is easier to view test progress in the pipeline

dtfs-central-api/src/v1/controllers/tfm/amendments/tfm-get-amendments.controller.ts
JoshBinns2000: I don't think so. This particular case is actually an annoying one for me because the reason this check needs to exist here is that the path parameter can be either a mongo id or an amendment status, so I can't use the middleware like I have everywhere else. If I wanted to throw here, I would need to create a new error class for invalid path parameters which I don't want to do as I want to discourage validating path parameters in the controller and encourage using the middleware

libs/common/src/test-helpers/index.ts
JoshBinns2000: Yes - same idea as `libs/common/src/index.ts` where the `export * from './sql-db-connection'` is commented out

libs/common/package.json
JoshBinns2000: I think it's clear enough. You don't want a separate export for each test case we define (I'm planning on adding one for SQL id validation), and I can't think of a way of naming this which narrows the description without narrowing what we can include in the export. Did you have any other names in mind?

libs/common/src/types/api-error-response-body.ts
JoshBinns2000: I think this is as commonised as it can realistically get for now. Commonising further would involve looking at every controller which I think it a bit excessive for this PR, but it's definitely something we may want to try and encourage for new controllers we add going forward

AlexBramhill (commented):
lgtm -- once @MarRobSoftwire is happy, we can send this

libs/common/package.json
AlexBramhill: we could do `test-cases-backend` and `test-cases-frontend` and `test-cases` or even `test-cases-using-mongo-db`?... but I feel that might add confusion or is clunky and not very nice? Would this make using this for front end tests annoying as we'd have to update to be able to handle this?

To be honest this question has come up a few times now... I think every time we take the `best for now` solution, and here I wouldn't mind doing the same...

libs/common/src/test-helpers/index.ts
AlexBramhill: (sorry again, not looking at the file rn, but adding a comment to be annoying and document if anyone looks at this review in future) It's doing this to basically say "deliberate omission"? Happy with this -- I'd like the comment to quicklty touch on why mongo functionality is a no go with general exports

dtfs-central-api/src/v1/controllers/tfm/amendments/tfm-get-amendments.controller.ts
```diff
@@ -72,7 +72,7 @@ export const getAmendmentsByFacilityId = async (req: Request, res: Response) =>
       default:
         if (amendmentIdOrStatus) {
```
AlexBramhill, line 73: not your problem, but we could guard clause here and remove a layer of indentation

dtfs-central-api/src/v1/controllers/tfm/amendments/tfm-get-amendments.controller.ts
AlexBramhill: Interesting -- I agree with the discouraging, unsure how it'll work in practice as this is still a common pattern regardless of middleware handling in the legacy code.

`amendmentIdOrStatus` is certainly troublesome!

Sorry I just like being part of these discussions

libs/common/package.json
JoshBinns2000: I have decided that I don't need the `ObjectId` for this, so I should be able to include it in the main export

dtfs-central-api/src/v1/controllers/tfm/amendments/tfm-get-amendments.controller.ts
JoshBinns2000: This is a case where I'd end up refactoring the whole controller which I want to avoid 😅

libs/common/package.json
JoshBinns2000: Turns out `supertest` also is only used in backend so I have changed the export to `test-cases-backend`

AlexBramhill (dismissed):

AlexBramhill (approved):

PR #3358, opened 2024-08-01
fix(DTFS2-7305): code consistency

abhi-markan:
## Introduction :pencil2:
Ensure methods are called with a consistent pattern and where applicable follow same conventions.

* `axios.isAxiosError` changed to `isAxiosError`
* Used `HttpStatusCodes` where applicable.


AlexBramhill (approved):

AlexBramhill (approved):
lgtm, fwiw I think we are looking at replacing the status codes as and when we hit them -- I've relayed a message to do this to the whole file when people come across them.

regarding destructuring, wholey agree `isAxiosError` can be destructured as an import (it doesn't have any performance effect, and is just an aesthetic thing). I think we should be less hard and fast with this rule... for instance `axios.get` should not be `get`, and same with other ambigious imports (`create`/`options` area great examples).

PR #3362, opened 2024-08-02
feat(FN-1725): keying sheet fixed fee calculation

JoshBinns2000:
## Introduction :pencil2:
We want to implement the calculation of the fixed fee adjustment. The details of this calculation are quite complex and are detailed at the bottom of this PR description.

- Adds a function to calculate the fixed fee
- Adds a function to calculate the fixed fee adjustment
- Updates existing api tests to line up with what this calculation is expecting to see
- Adds a helper to convert timestamps to dates*

*this helper function is required as the current mongo setup can have timestamps stored either in seconds or in milliseconds using the data type of `Date | number | string`. We need a way to parse all of these possible cases which is what this helper does.

## Calculation description 🔢 
There are three important parameters when calculating the fixed fee. These are:
- the utilisation
- the interest percentage
- the days remaining in the cover period (divided by the day count basis)

### Utilisation
Based on which part of the fixed fee you are calculating, this value can come from two places. The utilisation for the previous report period comes from the `FacilityUtilisationData` table, where the report period will be one month before the current report period in the general case. To calculate the current fixed fee, the facility utilisation stored on the `FeeRecord` table should be used. The fixed fee adjustment is the difference between these two values (see example calculation below).

### Interest percentage
This values comes from the mongo DB `tfm-facilities` collection and cannot be amended.

### Days remaining in cover period (divided by day count basis)
The number of days remaining in the cover period can be calculated by comparing the start date of the current report period to the cover end date stored in the mongo DB `tfm-facilities` collection. The cover end date should either be the value stored in the tfm-facility facility snapshot or, if there have been *completed* tfm-facility amendments, the latest amended cover end date. The number of days remaining is then simply the number of days between these two dates*.

*There is one edge-case here where, if the cover start date is after the start of the report period, the cover start date should be used in place of the report period start. This value is also stored on the tfm-facility facility snapshot and cannot be amended.

The day count basis where this value is divided by is also stored on the tfm-facility facility snapshot and cannot be amended.

### Example calculation
```
--------------------------------------------------------
Reference values:
-                   currentDate: 02/08/2024
- currentReportPeriod (Monthly): 07/2024
-                coverStartDate: 15/06/2024
- coverEndDate (latest amended): 01/07/2024
--------------------------------------------------------

--------------------------------------------------------
Previous period (07/2024) fixed fee:
-        utilisation = 200000
-          startDate = 15/06/2024 (cover start date is after the report period month start)
-            endDate = 01/07/2025
-        daysBetween = 381
-      dayCountBasis = 730
- interestPercentage = 5

=> fixedFee = 200000 * 0.9 (bank margin) * 0.05 * 381 / 730
            = 4,697.26
--------------------------------------------------------

--------------------------------------------------------
Current period fixed fee:
-        utilisation = 100000
-          startDate = 01/07/2024 (cover start date is before the report period month start)
-            endDate = 01/07/2025
-        daysBetween = 365
-      dayCountBasis = 730
- interestPercentage = 5

=> fixedFee = 100000 * 0.9 (bank margin) * 0.05 * 365 / 730
            = 2,250.00
--------------------------------------------------------

--------------------------------------------------------
=> fixedFeeAdjustment = 4697.26 - 2250
                      = 2,447.26 (DECREASE)
--------------------------------------------------------
```


dtfs-central-api/src/services/state-machines/fee-record/event-handlers/helpers/calculate-fixed-fee-adjustment.test.ts
```diff
@@ -0,0 +1,56 @@
...
+    const feeRecord = FeeRecordEntityMockBuilder.forReport(aUtilisationReport()).withFacilityId('11111111').build();
+    const facilityUtilisationData = FacilityUtilisationDataEntityMockBuilder.forId('22222222').build();
+
+    // Act / Assert
+    await expect(calculateFixedFeeAdjustment(feeRecord, facilityUtilisationData, aReportPeriod())).rejects.toThrow(
+      new Error('Fee record facility id does not match the facility utilisation id'),
+    );
+  });
+
+  it('returns the difference between the current fixed fee and the facility utilisation data fixed fee rounded to 2 decimal places', async () => {
```
RLavender98, line 25: I think we either should go down the path of 'current' and 'previous' or down the path of 'fee record fixed fee' and 'facility utilisation fixed fee' and my preference would be 'current' and 'previous'. So e.g. rename 'facility utilisation data fixed fee' in this test name to 'previous fixed fee' and then also in the test could rename:
- `feeRecordFacilityUtilisation` -> `currentFacilityUtilisation`
- `facilityUtilisationDataUtilisation` -> `previousFacilityUtilisation`
- `feeRecordFixedFee` -> `currentFixedFee`
- ...

dtfs-central-api/src/services/state-machines/fee-record/event-handlers/helpers/calculate-fixed-fee-adjustment.test.ts
```diff
@@ -0,0 +1,56 @@
...
+    const facilityUtilisationDataFixedFee = 23.332;
+    when(calculateFixedFee)
+      .calledWith(facilityUtilisationDataUtilisation, facilityId, facilityUtilisationData.reportPeriod)
+      .mockResolvedValue(facilityUtilisationDataFixedFee);
+
+    // Act
+    const result = await calculateFixedFeeAdjustment(feeRecord, facilityUtilisationData, reportPeriod);
+
+    // Assert
+    expect(result).toBe(100.12);
```
RLavender98, line 50: Also, could we add a case where this the previous fixed fee is greater, and a case where they are equal just to cover all ~~edge~~ cases in unit tests?

dtfs-central-api/src/services/state-machines/fee-record/event-handlers/helpers/calculate-fixed-fee.ts
```diff
@@ -0,0 +1,76 @@
...
+ * @returns The fixed fee for the current report period
+ */
+export const calculateFixedFee = async (utilisation: number, facilityId: string, reportPeriod: ReportPeriod): Promise<number> => {
+  const { coverEndDate, coverStartDate, dayCountBasis, interestPercentage } = await getLatestTfmFacilityValues(facilityId);
+
+  const numberOfDaysRemainingInCoverPeriod = getNumberOfDaysRemainingInCoverPeriod(reportPeriod.start, coverStartDate, coverEndDate);
+
+  return new Big(utilisation)
+    .mul(interestPercentage)
+    .mul(0.9) // use 90% of bank margin
```
RLavender98, line 71: I think we should extract a named const for this and we should be explicit in our comment about the business reason we do this in case in the future someone looks at this and wonders why. Can probably just adapt from the ticket description:

> Bank fee adjustment - take 10% of the margin for their admin fee (fixed at 90%) multiplied by…

dtfs-central-api/src/services/state-machines/fee-record/event-handlers/helpers/calculate-fixed-fee.ts
```diff
@@ -0,0 +1,76 @@
...
+  const currentReportPeriodStartMonthStart = startOfMonth(new Date(`${reportPeriodStart.year}-${reportPeriodStart.month}`));
+
+  if (isBefore(currentReportPeriodStartMonthStart, coverStartDate)) {
+    return differenceInDays(coverEndDate, coverStartDate);
+  }
+  return differenceInDays(coverEndDate, currentReportPeriodStartMonthStart);
+};
+
+/**
+ * Calculates the fixed fee for the current report period
```
RLavender98, line 58: This should be `for the given report period`

dtfs-central-api/src/services/state-machines/fee-record/event-handlers/helpers/calculate-fixed-fee.ts
```diff
@@ -0,0 +1,76 @@
...
+  return {
+    coverEndDate: convertTimestampToDate(coverEndDate),
+    coverStartDate: convertTimestampToDate(coverStartDate),
+    dayCountBasis,
+    interestPercentage: new Big(interestPercentage).div(100).toNumber(),
+  };
+};
+
+const getNumberOfDaysRemainingInCoverPeriod = (reportPeriodStart: MonthAndYear, coverStartDate: Date, coverEndDate: Date): number => {
+  const currentReportPeriodStartMonthStart = startOfMonth(new Date(`${reportPeriodStart.year}-${reportPeriodStart.month}`));
```
RLavender98, line 49: Could we use the `getDateFromMonthAndYear` helper inside the startOfMonth call here?

dtfs-central-api/src/services/state-machines/fee-record/event-handlers/helpers/calculate-fixed-fee.ts
```diff
@@ -0,0 +1,76 @@
...
+import { NotFoundError } from '../../../../../errors';
+import { convertTimestampToDate } from '../../../../../helpers';
+
+const getLatestCompletedAmendmentCoverEndDate = (amendments: TfmFacilityAmendment[]): number | undefined => {
+  const completedAmendments = amendments.filter(({ status }) => status === 'Completed');
+  const latestAmendmentWithDefinedCoverEndDate = orderBy(completedAmendments, ['updatedAt'], ['desc']).find(({ coverEndDate }) => coverEndDate);
+  return latestAmendmentWithDefinedCoverEndDate?.coverEndDate ?? undefined;
+};
+
+const getLatestTfmFacilityValues = async (
```
RLavender98, line 15: I'm worried this is slightly hidden if e.g. what Andy highlighted about new types amendments coming in actually comes to fruition. Is having latest values for these things relevant in other parts of the codebase? If so, I wonder if we can write a more common helper that then might get updated for free if new amendments exist...

dtfs-central-api/src/services/state-machines/fee-record/event-handlers/helpers/calculate-fixed-fee.test.ts
```diff
@@ -0,0 +1,193 @@
...
+
+  const reportPeriodStartMonth = reportPeriodDate.getMonth() + 1; // one-indexed
+  const reportPeriodStartYear = reportPeriodDate.getFullYear();
+  const reportPeriod: ReportPeriod = {
+    start: { month: reportPeriodStartMonth, year: reportPeriodStartYear },
+    end: { month: reportPeriodStartMonth, year: reportPeriodStartYear },
+  };
+
+  const roundTo2Dp = (value: number): number => Number(value.toFixed(2));
+  const getExpectedFixedFee = (): number => roundTo2Dp((utilisation * 0.9 * interestPercentage) / 100);
```
RLavender98, line 47: We are repeating (at least some of) the calculation code here, which we should not do, because then we are just testing our implementation against a more or less copy of our implementation.

dtfs-central-api/src/services/state-machines/fee-record/event-handlers/helpers/calculate-fixed-fee.test.ts
```diff
@@ -0,0 +1,193 @@
...
+  const coverEndDate = addDays(reportPeriodDate, 365);
+  const coverStartDate = startOfMonth(subMonths(reportPeriodDate, 1));
+  const numberOfDaysRemainingInCoverPeriod = 365;
+
+  const reportPeriodStartMonth = reportPeriodDate.getMonth() + 1; // one-indexed
+  const reportPeriodStartYear = reportPeriodDate.getFullYear();
+  const reportPeriod: ReportPeriod = {
+    start: { month: reportPeriodStartMonth, year: reportPeriodStartYear },
+    end: { month: reportPeriodStartMonth, year: reportPeriodStartYear },
+  };
```
RLavender98, line 44: These details are highly relevant to the calculation under test so any test with an assertion that relies on their specific value should have them specified directly in the test itself please

e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/reconcile-fee-records.spec.js
```diff
@@ -16,10 +16,31 @@ context('PDC_RECONCILE users can reconcile fee records', () => {
...
   const FEE_RECORD_ID_TWO = '22';
+  const FACILITY_ID_ONE = '11111111';
+  const FACILITY_ID_TWO = '22222222';
+
+  const aTfmFacilityFacilitySnapshot = () => ({
+    coverStartDate: new Date().getTime(),
+    coverEndDate: new Date().getTime(),
+    interestPercentage: 5,
+    dayCountBasis: 365,
+  });
```
RLavender98, line 27: Looks like this doesn't have all the required fields of the type it should really be returning which is a `Facility` I think, could we make it do so please?

libs/common/src/sql-db-entities/fee-record/fee-record.entity.ts
```diff
@@ -188,6 +188,14 @@ export class FeeRecordEntity extends AuditableBaseEntity {
     return feesPaidToUkefForThePeriodAsBig.div(paymentExchangeRateAsBig).round(precision).toNumber();
   }
 
+  public getTotalFeesAccruedForThePeriodInTheBaseCurrency(): number {
```
RLavender98, line 191: Where is this used?

RLavender98 (changes requested):
Nice work! Unit tests for the calculations could do with a few changes but overall looking good!

e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/reconcile-fee-records.spec.js
JoshBinns2000: The reason I was hesitant to do that is that we don't have typescript enabled inside e2e-tests, so it would be really easy for a generic `aFacility` helper to fall out of sync with what we expect a facility to be

dtfs-central-api/src/services/state-machines/fee-record/event-handlers/helpers/calculate-fixed-fee.ts
JoshBinns2000: In matching the way we have helpers for fee records, I have added a `tfm-facilities.helper.ts` to the helpers directory which then gets the latest cover end date. In terms of where else this happens in the codebase, I believe this kind of logic is usually performed in `portal-api` and/or `trade-finance-manager-api` before it is sent to the respective UI server which is why this commonised logic doesn't really exist in `dtfs-central-api`

dtfs-central-api/src/helpers/convert-timestamp-to-date.ts
```diff
@@ -0,0 +1,19 @@
...
+export const convertTimestampToDate = (timestamp: string | number | Date): Date => {
+  if (!(timestamp instanceof Date) && !/^\d+$/.test(timestamp.toString())) {
+    throw new Error(`Expected timestamp to be either a date object, number or string of digits (received '${timestamp}')`);
+  }
+
+  const timestampAsNumber = Number(timestamp);
+  if (timestampAsNumber.toString().length === 10) {
+    return new Date(timestampAsNumber * 1000); // convert from seconds to milliseconds
+  }
+  return new Date(timestampAsNumber);
```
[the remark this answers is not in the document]
JoshBinns2000: Happy to make the one-line helper functions

Definitely should return early if the input is already a `Date`

I didn't think about the fact that the timestamp not being 10 digits doesn't mean it will be 13 - I'll make it so the function  throws if the timestamp does not match a format we expect

dtfs-central-api/src/services/state-machines/fee-record/event-handlers/helpers/calculate-fixed-fee.test.ts
JoshBinns2000: Sure. I personally don't see the difference between hard coding the expected result by using a calculator vs calculating it when the test runs, but I think I also wrote this at the point where I hadn't fixed the date and so I calculated it in the test suite to avoid flakiness related to dates. Now that the date logic is fixed though, there isn't any reason to avoid hard-coding it

e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/reconcile-fee-records.spec.js
RLavender98: Ah I had failed to realise we don't even have typescript for helper files in the e2e-tests, in that case I agree fine to leave this

JoshBinns2000: I have a few ideas on how we can reconcile that in a future fix for e2e tests, but it would also involve having to properly build `libs/common` which is why I'm holding off on opening any fix for it

dtfs-central-api/src/services/state-machines/fee-record/event-handlers/helpers/calculate-fixed-fee.ts
RLavender98: Hmm I see.. `findLatestCompletedAmendment` in tfm-api seems vaguely like something one would want in the common libs perhaps and then to use but tbh that does something a little weird that I'm not sure I follow, almost implying perhaps that a later amendment contains all the previous amendment fields not being amended again?

dtfs-central-api/src/services/state-machines/fee-record/event-handlers/helpers/calculate-fixed-fee.test.ts
```diff
@@ -81,99 +40,143 @@ describe('calculateFixedFeeAdjustment', () => {
...
-    await expect(calculateFixedFee(utilisation, facilityId, reportPeriod)).rejects.toThrow(
+    await expect(calculateFixedFee(0, facilityId, aReportPeriod())).rejects.toThrow(
       new NotFoundError(`Failed to find a cover start date for the tfm facility with ukef facility id '${facilityId}`),
     );
   });
 
   it('calculates the fixed fee using the facility snapshot cover start date if the report period start month starts before the cover start date', async () => {
     // Arrange
-    const coverStartDateAfterReportPeriod = addMonths(reportPeriodDate, 1);
+    const facilityId = '12345678';
```
RLavender98, line 92: These tests are looking better thank you! I think we can still make a few more improvements:
1. We are testing that the correct calculation happens but it feels almost inadvertent, would be good to detail the calculation, e.g. 'calculates the fixed fee by multiplying the utilisation, interest, ... and dividing by ....' 
2. Also are the tests covering cases with decimal places etc and that rounding happens correctly?
3. Finally the answers we assert against feel pulled slightly out of nowhere, I would recommend leaving a comment above the assert detailing/explaining the calculation of the correct value.

I'm actually tempted now I look at this in this format to suggest we actually have a pulled out method just for the last line of the `calculateFixedFee` method which has the actual calculation that we then unit test for the calculation logic so it is tested separately to all the 'fetching and choosing data' logic... I'm not sure what it would be called though... perhaps the existing method is `getFixedFee` and then the calculation line pulled out would be `calcualteFixedFee`?

JoshBinns2000: That's a nice idea - makes it much easier to separate the concerns of getting the correct data and calculating the value

dtfs-central-api/src/services/state-machines/fee-record/event-handlers/helpers/calculate-fixed-fee.ts
```diff
@@ -53,18 +17,35 @@ const getNumberOfDaysRemainingInCoverPeriod = (reportPeriodStart: MonthAndYear,
...
  */
-export const calculateFixedFee = async (utilisation: number, facilityId: string, reportPeriod: ReportPeriod): Promise<number> => {
-  const { coverEndDate, coverStartDate, dayCountBasis, interestPercentage } = await getLatestTfmFacilityValues(facilityId);
-
+export const calculateFixedFee = ({
+  utilisation,
+  reportPeriod,
+  coverStartDate,
+  coverEndDate,
+  interestPercentage,
```
RLavender98, line 45: Interest percentage here means the decimal representation but in the other files and on the facility it means out of a 100, I personally think percentage should be out of 100 one so I would probably move the dividing by 100 into here or name this slightly differently (and/or could be explicit in the docstring perhaps?)

RLavender98 (commented):
Just one final comment!

dtfs-central-api/src/services/state-machines/fee-record/event-handlers/helpers/calculate-fixed-fee.ts
JoshBinns2000: I've just performed the "divide by 100" in this function as keeping the format of `interestPercentage` consistent will probably be easier to understand

RLavender98 (approved):
Nice work!

PR #3363, opened 2024-08-02
fix(DTFS2-6889): ensure ACBS DOF payload is correctly constructed

abhi-markan:
## Introduction :pencil2:

* When invoking OS certain addresses do no have a `country` field if they are a valid UK postcodes, keeping this permutation in mind any deal with an empty country field will be set to default country code `GBR`.
* Refactored `createACBS` function:
  * Payload validation for both `deal` and `bank` properties
  * createACBS argument has been updated to `string` to accept a valid Mongo Object deal identifier, which is then used to fetch deal with relevant properties used by ACBS DOF.
  * Removed `auditRecord` from `deal` payload for ACBS DOF consumption.
  * 

* Improved unit and API test coverage for `createACBS` function.
* Updated `.eslintrc.js` to not allow prefixed `_` variable as this has been avoided in the DTFS code base.
* Updated `.eslintrc.js` `no-unused-vars` to allow object destructing.
* Improved logging.


trade-finance-manager-api/src/v1/controllers/acbs.controller.js
```diff
@@ -52,10 +57,17 @@ const addToACBSLog = async (payload) => {
 
 /**
  * Creates an ACBS task and adds it to the ACBS log.
- * @param {object} deal - The deal object containing the dealSnapshot property with the bank details.
+ * @param {String} deal - TFM deal object ID
```
JoshBinns2000, line 60: I think this type should be string and that the variable name should be dealId to match the function below

utils/mock-data-loader/insert-mocks-gef.js
```diff
@@ -24,7 +24,7 @@ const insertMocks = async (token) => {
 
   const makerToken = await portalApi.loginViaPortal(BANK1_MAKER1);
 
-  for (const [_index, item] of MOCKS.APPLICATION.entries()) {
+  for (const [item] of MOCKS.APPLICATION.entries()) {
```
JoshBinns2000, line 27: I think this may be a mistake as the first element of `.entries()` is the index of the array and the second element is the value of the element at that index. If `MOCKS.APPLICATION` is an array (which I think it must be if we are using `.entries()`) then this can just be `const item of MOCKS.APPLICATION` I think

trade-finance-manager-api/src/v1/controllers/acbs.controller.js
JoshBinns2000: I also think the return type should be `boolean`

trade-finance-manager-api/api-tests/v1/acbs/acbs.api-test.js
```diff
@@ -63,20 +58,25 @@ describe('acbs controller', () => {
...
-        dealSnapshot: {
-          ...MOCK_DEAL,
-          bank: null,
-        },
-      });
+    afterEach(() => {
+      jest.clearAllMocks();
+    });
+
+    it.each(invalidIds)('should return false if an invalid Mongo Object deal id has been supplied', async (dealId) => {
```
JoshBinns2000, line 65: This will appear in the test suite as the same test description for each case which would be hard to debug. Could we change this to be `should return false if the deal id is '%s' (not a valid mongo object id)`, async (dealId)`

trade-finance-manager-api/src/v1/controllers/acbs.controller.js
abhi-markan: Good spot, have updated all.

trade-finance-manager-api/api-tests/v1/acbs/acbs.api-test.js
abhi-markan: Good spot, have added string substitution.

utils/mock-data-loader/insert-mocks-gef.js
abhi-markan: Ah! yes it is apologies for that, have fixed. Thank you for pointing it out.

utils/mock-data-loader/insert-mocks-gef.js
```diff
@@ -24,12 +24,14 @@ const insertMocks = async (token) => {
 
   const makerToken = await portalApi.loginViaPortal(BANK1_MAKER1);
 
-  for (const [item] of MOCKS.APPLICATION.entries()) {
-    item.userId = makerUserId;
-    const application = await api.createApplication(item, makerToken);
+  for (const item of MOCKS.APPLICATION.entries()) {
+    const deal = item[1];
```
JoshBinns2000, line 28: Why can't this just be `for (const deal of MOCKS.APPLICATION) {`?

libs/common/src/constants/index.ts
```diff
@@ -9,10 +9,11 @@ export * from './portal';
 export * from './durable-functions-log.constant';
 export * from './fee-record-status';
 export * from './audit-user-types';
-export * as PAYLOAD_VERIFICATION from './payload-verification';
```
JoshBinns2000, line 12: What's the reason for moving this `export` to the bottom of the file?

utils/mock-data-loader/insert-mocks-gef.js
abhi-markan: Have removed `entries`, was trying to keep it consistent.

JoshBinns2000 (dismissed):

azure-functions/acbs-function/src/functions/acbs.js
```diff
@@ -34,6 +34,13 @@ df.app.orchestration('acbs', function* HDeal(context) {
         country = deal.dealSnapshot.exporter.registeredAddress.country;
       }
 
+      /**
+       * If no country is specified default it to `GBR`
+       */
+      if (!country) {
```
AlexBramhill, line 40: we could assign this initially as apposed to having it `undefined` initially

trade-finance-manager-api/src/v1/controllers/acbs.controller.test.js
```diff
@@ -22,9 +38,12 @@ const mockACBSTaskLink = {
   rewindPostUri: 'acbs',
   purgeHistoryDeleteUri: 'acbs',
 };
+
 const updateACBSfacilityMock = jest.spyOn(api, 'updateACBSfacility');
```
AlexBramhill, line 42: I know this isnt you, but could you update this to `updateACBSFacilityMock` (cap `F`) please

azure-functions/acbs-function/.eslintrc.js
```diff
@@ -34,12 +34,7 @@ const baseRules = {
     },
   ],
   'lines-between-class-members': ['error', 'always', { exceptAfterSingleLine: true }],
-  'no-unused-vars': [
-    'error',
-    {
-      varsIgnorePattern: '^_',
-    },
-  ],
+  'no-unused-vars': ['error', { ignoreRestSiblings: true }],
```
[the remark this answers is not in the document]
abhi-markan: As discussed before DTFS has stayed away from `_` prefixed variables.

dtfs-central-api/src/v1/controllers/portal/deal/update-deal.controller.js
```diff
@@ -156,6 +157,13 @@ exports.updateDeal = updateDeal;
 
 const addFacilityIdToDeal = async (dealId, newFacilityId, user, routePath, auditDetails) => {
   await findOneDeal(dealId, async (deal) => {
+    if (!deal) {
```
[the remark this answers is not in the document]
abhi-markan: Good shout have added.

AlexBramhill (dismissed):

AlexBramhill (approved):

PR #3366, opened 2024-08-05
feat(FN-2739): display report frequency

RLavender98:
## Introduction :pencil2:
We want users to see which reports are quarterly and which are monthly

- Update the table headers on the bank reports page to refer to the end of the reporting period
- Add frequency column to the bank reports table and display `Monthly` or `Quarterly` depending on the report period on the utilisation report

## Screenshots 📷 :
With feature flag on:
Screenshot 2024-08-05 at 16 47 28
With feature flag off:
Screenshot 2024-08-05 at 16 49 13


JoshBinns2000 (commented):
Nice job - mainly wondering if we can rename the `UtilisationReportingFrequency` to `UtilisationReportReportingFrequency` to match our other constants/types

trade-finance-manager-ui/server/controllers/utilisation-reports/index.test.ts
```diff
@@ -88,7 +88,6 @@ describe('controllers/utilisation-reports', () => {
       await getUtilisationReports(req, res);
 
       // Assert
-      expect(isTfmPaymentReconciliationFeatureFlagEnabledSpy).toHaveBeenCalledTimes(1);
```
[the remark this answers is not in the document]
RLavender98: It actually gets called twice now but that feels like more of an implementation detail rather than something we explicitly want to assert happens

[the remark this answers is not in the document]
JoshBinns2000: I thought that would be why - just wanted to confirm!

trade-finance-manager-ui/server/types/view-models/utilisation-reports-view-model.ts
```diff
@@ -2,7 +2,10 @@ import { IsoMonthStamp } from '@ukef/dtfs2-common';
 import { BaseViewModel } from './base-view-model';
 import { UtilisationReportReconciliationSummaryItem } from '../utilisation-reports';
 
+export type UtilisationReportingFrequency = 'Monthly' | 'Quarterly';
```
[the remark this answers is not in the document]
RLavender98: This is currently constrained to just the one page and just as a display value, but if we were ever to say want to return this from central api or use in more places then I would agree we should move into libs/common and use the constant/type pair linked with a function that maps to display value, but I'm not sure I see the value in doing that for something used just in the small way it currently is?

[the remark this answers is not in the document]
RLavender98: I have renamed it to `UtilisationReportDisplayFrequency`, I could name it `UtilisationReportDisplayReportingFrequency` but that felt slightly harder to read to me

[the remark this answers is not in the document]
JoshBinns2000: >but I'm not sure I see the value in doing that for something used just in the small way it currently is

I agree with this. Thanks for the name change as well and I agree that `UtilisationReportDisplayFrequency` is clear enough

JoshBinns2000 (approved):
LGTM

trade-finance-manager-ui/component-tests/utilisation-reports/utilisation-reports.component-test.js
```diff
@@ -24,7 +22,14 @@ describe(page, () => {
   });
 
   const getWrapper = async () => {
-    const reportPeriodSummaries = await getReportReconciliationSummariesViewModel(MOCK_UTILISATION_REPORT_RECONCILIATION_SUMMARY, 'user-token');
+    const reportPeriodSummaries = [
```
JoshBinns2000, line 25: Could we update this function to no longer be `async`? We should then be able to remove all the `await`s below as well :)

JoshBinns2000 (commented):
Really nice test updates! Just one comment around one `async` function which I think doesn't need to be `async`.

I'll leave this up to you, but we could also take this as a chance to update those files to typescript

trade-finance-manager-ui/server/controllers/utilisation-reports/index.test.ts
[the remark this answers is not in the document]
ttbarnes: Sorry, just stumbled across this - I don't understand why we would _not_ want to test this?

I think it should remain as is because with removing it - the flag could then be removed/disabled in the controller and the test would still pass.

If it's getting called twice, it sounds like the mock/spy needs to be reset or setup inbetween tests

trade-finance-manager-ui/server/types/view-models/utilisation-reports-view-model.ts
```diff
@@ -2,7 +2,10 @@ import { IsoMonthStamp } from '@ukef/dtfs2-common';
 import { BaseViewModel } from './base-view-model';
 import { UtilisationReportReconciliationSummaryItem } from '../utilisation-reports';
 
+export type UtilisationReportDisplayFrequency = 'Monthly' | 'Quarterly';
```
abhi-markan, line 5: Can this please be constants?

trade-finance-manager-ui/server/controllers/utilisation-reports/helpers/reconciliation-summary-helper.ts
```diff
@@ -42,19 +55,39 @@ export const getDueDateText = (reportDueDate: Date) => {
...
 
+const getReportPeriodHeadingWithPeriodEnd = (submissionMonth: IsoMonthStamp, reportPeriods: ReportPeriod[]) => {
+  const isCurrentSubmissionMonth = isSameMonth(new Date(submissionMonth), new Date());
+
+  const formattedReportPeriodEnd = format(getDateFromMonthAndYear(reportPeriods[0].end), 'MMMM yyyy');
+
+  return `${isCurrentSubmissionMonth ? 'Current' : 'Open reports'} reporting period end: ${formattedReportPeriodEnd}`;
+};
+
+export const getReportPeriodHeading = (submissionMonth: IsoMonthStamp, reportPeriods: ReportPeriod[]) => {
```
abhi-markan, line 74: Missing docstring.

trade-finance-manager-ui/server/controllers/utilisation-reports/helpers/reconciliation-summary-helper.ts
```diff
@@ -42,19 +55,39 @@ export const getDueDateText = (reportDueDate: Date) => {
...
 
   return `${isCurrentSubmissionMonth ? 'Current reporting period' : 'Open reports'}: ${formattedReportPeriods}`;
 };
 
+const getReportPeriodHeadingWithPeriodEnd = (submissionMonth: IsoMonthStamp, reportPeriods: ReportPeriod[]) => {
+  const isCurrentSubmissionMonth = isSameMonth(new Date(submissionMonth), new Date());
+
+  const formattedReportPeriodEnd = format(getDateFromMonthAndYear(reportPeriods[0].end), 'MMMM yyyy');
+
+  return `${isCurrentSubmissionMonth ? 'Current' : 'Open reports'} reporting period end: ${formattedReportPeriodEnd}`;
```
abhi-markan, line 71: Not sure if the constants already exists for these?

trade-finance-manager-ui/server/controllers/utilisation-reports/helpers/reconciliation-summary-helper.ts
```diff
@@ -12,11 +20,16 @@ export const reconciliationStatusCodeToDisplayStatus: Record<UtilisationReportRe
   RECONCILIATION_COMPLETED: 'Report completed',
 };
 
+const getUtilisationReportDisplayFrequency = (reportPeriod: ReportPeriod): UtilisationReportDisplayFrequency => {
```
abhi-markan, line 23: Functions missing docstring, all the functions should always have a docstring please.

abhi-markan (changes requested):

trade-finance-manager-ui/server/controllers/utilisation-reports/index.test.ts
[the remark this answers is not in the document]
RLavender98: The clean up is working correctly, the logic has been changed in the controller so we do genuinely call the helper twice (through two separate helper methods used by the page).

> the flag could then be removed/disabled in the controller and the test would still pass.

That is actually why we don't need this assertion, if the check gets removed and the assertion below still passes, then the behaviour is still the desired behaviour, since the assertion below this one is testing the outcome we want. This way the test is actually more resilient and if implementation details change but not desired behaviour, then the test doesn't need updating.

trade-finance-manager-ui/server/controllers/utilisation-reports/helpers/reconciliation-summary-helper.ts
RLavender98: No, I don't believe so

trade-finance-manager-ui/server/controllers/utilisation-reports/index.test.ts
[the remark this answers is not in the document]
ttbarnes: Thanks @RLavender98 I can see that we're testing the correct template is rendered which is great - can we add a test to make sure that `isTfmPaymentReconciliationFeatureFlagEnabled` is called please?

I appreciate the end result is tested, but something is being called and currently this is not tested. I think this should be tested so that we know the correct `dtfs2-common` feature flag is being called.

Otherwise, if the flag changes in this controller, it could return an incorrect result in the real world; Additionally the test will still pass regardless of which feature flag it's calling.

trade-finance-manager-ui/component-tests/utilisation-reports/utilisation-reports.component-test.js
RLavender98: Ah nice catch!

trade-finance-manager-ui/server/controllers/utilisation-reports/index.test.ts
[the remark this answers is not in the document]
RLavender98: Fair enough, I'll add it then.

JoshBinns2000 (approved):
LGTM

trade-finance-manager-ui/server/controllers/utilisation-reports/helpers/reconciliation-summary-helper.ts
abhi-markan: Can they please be created and used to keep it DRY. Thanks

abhi-markan (changes requested):

trade-finance-manager-ui/server/controllers/utilisation-reports/helpers/reconciliation-summary-helper.ts
RLavender98: I'm a little uncertain what a good name for this would be. It's only used and expected to be used in this one place, what do you think of this?
```
export const BANK_REPORTS_FOR_PERIOD_TABLE_HEADER_PREFIX = {
  CURRENT_REPORTING_PERIOD: 'Current',
  NOT_CURRENT_REPORTING_PERIOD: 'Open reports',
} as const;
```
So then it would become:
`${isCurrentSubmissionMonth ? BANK_REPORTS_FOR_PERIOD_TABLE_HEADER_PREFIX.CURRENT_REPORTING_PERIOD : BANK_REPORTS_FOR_PERIOD_TABLE_HEADER_PREFIX.NOT_CURRENT_REPORTING_PERIOD}`

ttbarnes (commented):
Can we add some E2E tests?

trade-finance-manager-ui/server/controllers/utilisation-reports/helpers/reconciliation-summary-helper.ts
abhi-markan: Hey @RLavender98 above snippet looks good thank you.

trade-finance-manager-ui/server/controllers/utilisation-reports/helpers/reconciliation-summary-helper.ts
```diff
@@ -36,25 +67,78 @@ const getDistinctReportPeriodsByStartMonth = (reportPeriods: ReportPeriod[]): Re
...
+ * @param submissionMonth - The submission month
+ * @param reportPeriods - List of report periods
+ * @returns Report period heading with the report periods listed
+ */
+const getReportPeriodHeadingWithAllPeriods = (submissionMonth: IsoMonthStamp, reportPeriods: ReportPeriod[]) => {
   const isCurrentSubmissionMonth = isSameMonth(new Date(submissionMonth), new Date());
 
   const formattedReportPeriods = reportPeriods.map((reportPeriod) => getFormattedReportPeriodWithShortMonth(reportPeriod, true)).join(' and ');
 
   return `${isCurrentSubmissionMonth ? 'Current reporting period' : 'Open reports'}: ${formattedReportPeriods}`;
```
abhi-markan, line 92: @RLavender98 could you also please update this, appreciate this is not in your PR. Thanks

abhi-markan (changes requested):

trade-finance-manager-ui/server/controllers/utilisation-reports/helpers/reconciliation-summary-helper.ts
RLavender98: Ahh I see, these are subtly different to other prefixes, I will see if I can find a commonisation.

RLavender98: I've made an attempt to commonise. It feels a little over-engineered to me but I'm not sure how else to do with constants whilst avoiding repetition of substrings. Let me know what you think!

abhi-markan (approved):
LGTM 👍🏼

trade-finance-manager-ui/server/constants/utilisation-report-display-frequency.ts
```diff
@@ -0,0 +1,4 @@
+export const UTILISATION_REPORT_DISPLAY_FREQUENCY = {
+  QUARTERLY: 'Quarterly',
+  MONTHLY: 'Monthly',
+} as const;
```
ttbarnes, line 4: Not a blocker - we should create DRY shared `FEE_FREQUENCY` constants, just searching the codebase I can see 5x different instances of `QUARTERLY: 'Quarterly'`.

`FEE_FREQUENCY` in ACBS functions seems to be the most comprehensive, I think this should be extracted into shared constants, not sure if we have such a shared package in place yet @abhi-markan

ttbarnes (approved):

trade-finance-manager-ui/server/constants/utilisation-report-display-frequency.ts
RLavender98: Interesting I had not seen fee frequency as a thing. I think here we do want a separate constant however because this is a report frequency not a fee frequency so only overlaps by chance rather than intent and they have separate business meanings I imagine as they allow different ranges of frequency. Generally not a good idea to link things that only relate coincidentally as it makes it harder to maintain down the line, even though it does look more DRY at a glance.

ttbarnes: @RLavender98 I was thinking we could have generic `FREQUENCY` constants, as regardless of where such frequencies are used, they are all simply frequencies of a year.

E.g from ACBS (just renamed the const here):

```ts
const FREQUENCY = {
  WEEKLY: 'Weekly',
  BI_WEEKLY: 'Bi-Weekly',
  MONTHLY: 'Monthly',
  BI_MONTHLY: 'Bi-Monthly',
  QUARTERLY: 'Quarterly',
  SEMI_ANNUALLY: 'Semi-annually',
  ANNUALLY: 'Annually',
  BI_ANNUALLY: 'Bi-annually',
};
```

Such frequencies should be aligned throughout all products, e.g it's not wise to have 1x service saying "Bi weekly", but another says "Bi-weekly", it should be standardised

PR #3369, opened 2024-08-06
fix(dtfs2-none): update integer in jsdocs when incorrect

AlexBramhill:
## Introduction :pencil2:
We incorrectly use `Integer` to refer to numbers which are not ints.

Update these to number

I know that `Integer` does not exist in JS -- To avoid this being a contentious PR, I've left `Integer` in where it makes sense (`months` etc), as JSDocs do not care this isn't a real type


JoshBinns2000 (approved):
LGTM - nice fix

abhi-markan (changes requested):
Nice update, ticket number as other PRs please.

AlexBramhill:
closed for https://github.com/UK-Export-Finance/dtfs2/pull/3374

PR #3370, opened 2024-08-06
fix(DTFS2-NONE): update-jsdocs-to-avoid-not-awaiting-promises

AlexBramhill:
## Introduction :pencil2:
If a jsdoc does not return a promise, we can call async functions without having to await them (no errors).

Updated jsdocs to reflect async functions

Remove unneeded types from ts file


azure-functions/acbs-function/api.js
```diff
@@ -42,7 +42,7 @@ const badRequest = {
 /**
  * Invokes TFS GET endpoint
  * @param {String} endpoint TFS endpoint
- * @returns {Object} API response object
+ * @returns {Promise<Object>} API response object
```
JoshBinns2000, line 45: I think this should be `Promise<object>`

azure-functions/acbs-function/mappings/facility/helpers/get-facility-currency-exchange-rate.js
```diff
@@ -2,7 +2,7 @@
  * Return facility's curency exchange rate when facility
  * currency is not in GBP
  * @param {Object} facility Facility object
- * @returns {Integer} Facility exchange rate
+ * @returns {number} Facility exchange rate
```
JoshBinns2000, line 5: Nice catch

e2e-tests/support/tasks.js
```diff
@@ -93,7 +93,7 @@ module.exports = {
     /**
      * Inserts utilisation report details to the SQL database
      * @param {Partial<import('@ukef/dtfs2-common').UtilisationReportEntity[]>} utilisationReports
-     * @returns {import('@ukef/dtfs2-common').UtilisationReportEntity[]} The inserted reports
+     * @returns {Promise<import('@ukef/dtfs2-common').UtilisationReportEntity[]>} The inserted reports
```
JoshBinns2000, line 96: I _think_ that because we actually import `UtilisationReportEntity`, we should be able to remove the `import('@ukef/dtfs2-common')`

gef-ui/server/controllers/unissued-facilities/index.js
```diff
@@ -13,7 +13,7 @@ const { validationErrorHandler } = require('../../utils/helpers');
  * if change true, changes 'cancel' + 'back' button href back to application preview
  * else renders back to unissued facilities list
  * @param {req} params, @param {req} query, @param {Boolean} change
- * @returns {Object} body
+ * @returns {Promise<object>} body
```
JoshBinns2000, line 16: I know the above isn't part of this PR, but `req` and `Boolean` do not appear to be the correct types (same for the below docstrings)

gef-ui/server/controllers/unissued-facilities/index.js
```diff
@@ -249,7 +249,7 @@ const postChangeUnissuedFacility = async (req, res) => {
  * post for changing unissued facilities from application preview
  * redirects to application preview once changed
  * @param {req}
- * @returns {res}
+ * @param {res}
```
JoshBinns2000, line 252: `res` and `req` should not be in the `{}` - my guess is that this should be `@param {import('express').Request} req` and `@param {import('express').Response} res`

portal/server/api.js
```diff
@@ -185,7 +185,7 @@ const updateDealName = async (id, newName, token) => {
  * @param {Object} statusUpdate - An object containing the `_id` property representing
  * the deal ID and the `status` property representing the new status of the deal.
  * @param {string} token - A token used for authorization in the API call.
- * @returns {Object|boolean} - An object containing the `status` code and the `data` from the API response, or `false` if the `_id` is not valid.
+ * @returns {Promise<Object|boolean>} - An object containing the `status` code and the `data` from the API response, or `false` if the `_id` is not valid.
```
JoshBinns2000, line 188: Nit but I think formatting rules suggest that this should be `Promise<object | boolean>`

trade-finance-manager-api/src/v1/helpers/activity.js
```diff
@@ -129,7 +129,7 @@ const getActivities = (deal) => {
  * Return array of activities object, comprising of
  * ACBS interaction records
  * @param {Object} acbs Durable function output
- * @returns {Array} An array of activities object
+ * @returns {Promise<object[]|{}>} An array of activities object
```
JoshBinns2000, line 132: This is an annoying return type, but I see why this is it. Either way, same formatting rules as above - `Promise<object[] | {}>`

trade-finance-manager-api/src/v1/helpers/create-tasks-amendment.helper.js
```diff
@@ -323,7 +323,7 @@ const updateAmendmentTasks = async (facilityId, amendmentId, taskUpdate) => {
  *                        a 'groupTasks' key with a list of tasks.
  * @param {string} groupName - The name of the team/group which tasks should be assigned.
  * @param {string} newUserId - The ID of the user to whom tasks should be assigned.
- * @returns {Array} A modified list of task groups.
+ * @returns {Promise<Array>} A modified list of task groups.
```
JoshBinns2000, line 326: I believe the generic argument for the `Array` interface is required, so this should either be `Promise<object[]>` to match the above, or `Promise<Array<object>>`

trade-finance-manager-api/src/v1/tasks/assign-group-tasks-to-one-user.js
```diff
@@ -7,7 +7,7 @@ const getAssigneeFullName = require('../helpers/get-assignee-full-name');
  * @param {Array} array of group titles that the should be assigned to the user
  * @param {String} user ID
  * @param {import("@ukef/dtfs2-common").AuditDetails} auditDetails - user making the request
- * @returns {Array} Updated tasks
+ * @returns {Promise<Array>} Updated tasks
```
JoshBinns2000, line 10: Same comment as above for `Array`

utils/reporting/tfm/rad-reporting.js
```diff
@@ -74,7 +74,7 @@ const constructRows = (deals) => {
 /**
  * Generates bespoke report as CSV
  * @param {Array} rows Array of processed deals
- * @returns {Null} Null is returned
+ * @returns {Promise<boolean>} Report generation status, true if successful
```
JoshBinns2000, line 77: I know it's not the purpose of this PR, but the `return Promise.resolve(true)` and `Promise.reject()` seems unnecessary - `return true` and `throw new Error()` should result in the same behaviour as far as I can see

JoshBinns2000 (commented):
Comments largely around formatting and semantics

gef-ui/server/controllers/unissued-facilities/index.js
AlexBramhill: am simply going to remove these for now -- as they are incorrect but working out types is beyond the scope

gef-ui/server/controllers/unissued-facilities/index.js
AlexBramhill: Lets keep this for now and ill address in another PR tomorrow

JoshBinns2000 (approved):
LGTM

abhi-markan (changes requested):
Nice work, need `DTFS2-ABC` ticket number for branch name and PR please.

AlexBramhill:
closed for https://github.com/UK-Export-Finance/dtfs2/pull/3375

PR #3372, opened 2024-08-07
fix(dtfs2-7309): update use of promises inside map and for each

AlexBramhill:
## Introduction :pencil2:
Some legacy code is not awaiting promises inside `forEach` and `map`

Updated these to work correctly

Refactored a few functions to reduce complexity


portal-api/src/v1/controllers/deal-status/send-status-update-emails.js
```diff
@@ -1,9 +1,11 @@
 const sendEmail = require('../../email');
 
 const sendEmailsToOwningBanks = async (templateId, emailVariables, owningBankEmails) => {
-  owningBankEmails.forEach(async (email) => {
-    await sendEmail(templateId, email, emailVariables);
-  });
+  await Promise.all(
+    owningBankEmails.map(async (email) => {
+      await sendEmail(templateId, email, emailVariables);
+    }),
```
JoshBinns2000, line 7: Maybe `owingBankEmails.map((email) => sendEmail(templateId, email, emailVariables))` for conciseness?

portal-api/src/v1/gef/controllers/application-submit.js
```diff
@@ -187,38 +187,41 @@ const submissionPortalActivity = async (application) => {
...
+    const isSubmittedToUkef = !!app.submissionCount;
+    const hasFacilities = facilities?.length > 0;
+    const isAIN = app.submissionType === CONSTANTS.DEAL.SUBMISSION_TYPE.AIN;
+
+    if (isSubmittedToUkef && !hasFacilities) {
+      return false;
     }
+
+    const updatedFacilities = await Promise.all(
+      facilities.reduce((promises, facility) => {
```
JoshBinns2000, line 205: I tend to use `.filter(...).map(...)` when dealing with `Promise` related logic as I think needing to figure out both what is going on with the `reduce` and with the `Promise` can be a bit tricky, but I don't mind leaving this one as it's quite small

portal-api/src/v1/gef/controllers/application.controller.js
```diff
@@ -219,19 +219,22 @@ const sendStatusUpdateEmail = async (user, existingApplication, status) => {
...
-      bankInternalRefName,
-      currentStatus: status,
-      previousStatus,
-      updatedByName: `${user.firstname} ${user.surname}`,
-      updatedByEmail: user.email,
-    });
-  });
+  const emailPromises = user.bank.emails.map(
+    async (email) =>
+      await sendEmail(EMAIL_TEMPLATE_IDS.UPDATE_STATUS, email, {
```
JoshBinns2000, line 224: Nit but if you are just returning a `Promise`, you don't need the `async () => await func()`

abhi-markan (changes requested):
Can we pause this PR please so I can investigate why this is required and if the current implementation is impacting any existing functionality. Thanks

portal-api/src/v1/gef/controllers/application-submit.js
AlexBramhill: thanks, lets leave this as is, but Im taking this into account for future stuff

AlexBramhill:
> Can we pause this PR please so I can investigate why this is required and if the current implementation is impacting any existing functionality. Thanks

Sure -- I'd be keen to get this in asap tho 

https://stackoverflow.com/questions/68026591/what-is-the-difference-between-async-await-foreach-and-promise-all-map

dtfs-central-api/src/v1/controllers/portal/gef-deal/add-min-activities.controller.js
```diff
@@ -84,25 +84,26 @@ const facilityChangePortalActivity = async (application, facilities) => {
...
-          maker,
-          checker,
-        };
-        // generates an activities object
-        const activityObj = portalActivityGenerator(activityParams);
-        // adds to beginning of portalActivities array so most recent displayed first
-        portalActivities.unshift(activityObj);
+    facilities.forEach((facility) => {
+      if (!facility.canResubmitIssuedFacilities) {
+        return;
```
ttbarnes, line 89: Is this required? Otherwise the one liner is simpler?:

```js
if (facility.canResubmitIssuedFacilities) { ...
```

portal-api/src/drivers/fileshare.js
```diff
@@ -165,9 +165,11 @@ const deleteMultipleFiles = async (fileshare, filePath, fileList) => {
   if (!fileList) return false;
 
   if (Array.isArray(fileList)) {
-    return fileList.map(async (filename) => {
-      await deleteFile(fileshare, `${filePath}/${filename}`);
-    });
+    return Promise.all(
```
ttbarnes, line 168:
```suggestion
    const promises = await Promise.all( ...

    return promises;
```

JoshBinns2000: Could I possibly ask what the reason for this change is? Also with the note that, semantically, the thing that gets returned when a `Promise.all` is `await`ed isn't a list of `promises`, but instead a list of whatever the `Promise` resolved to

portal-api/src/v1/controllers/deal-status/update-submitted-issued-facilities.js
```diff
@@ -48,11 +48,11 @@ const updateSubmittedIssuedFacilities = async (user, deal, auditDetails) => {
...
     }
 
-    const { data } = await facilitiesController.update(deal._id, facilityId, facility, user, auditDetails);
-
-    return data;
+    await facilitiesController.update(deal._id, facilityId, facility, user, auditDetails);
   });
 
+  await Promise.all(updateFacilitiesPromises);
+
```
ttbarnes, line 55: Will the deal returned in this function include the updated facilities?

dtfs-central-api/src/v1/controllers/portal/gef-deal/add-min-activities.controller.js
JoshBinns2000: I don't think this is necessarily simpler. If I remember correctly, this change was made because guard clauses are often easier to read than branching code - the early `return` here if the condition is met makes it clear to the reader that, if you can't resubmit issued facilities, then you do nothing. In contrast, a non-guard clause approach means that a reader would read this code as "if you can resubmit issued facilities, perform this action", which means they need to scroll to the bottom of the `if` block to figure out what happens if you can't resubmit facilities

portal-api/src/v1/gef/controllers/application-submit.js
```diff
@@ -187,38 +187,41 @@ const submissionPortalActivity = async (application) => {
...
-      }
-    } catch (error) {
-      console.error('Unable to set coverDateConfirmed for AIN facilities. %o', error);
+  if (!app) {
+    return false;
+  }
+
+  try {
+    const facilities = await getAllFacilitiesByDealId(app._id);
+    const isSubmittedToUkef = !!app.submissionCount;
```
ttbarnes, line 196: Maybe could use use `>= 1` check instead of `!!`?

portal-api/src/v1/gef/controllers/application-submit.js
```diff
@@ -187,38 +187,41 @@ const submissionPortalActivity = async (application) => {
...
+          promises.push(updateFacility(facility._id, { coverDateConfirmed }, auditDetails));
+        }
+
+        return promises;
+      }, []),
+    );
+
+    const hasUpdatedAtLeaseOneFacility = updatedFacilities.length > 0;
+
+    return hasUpdatedAtLeaseOneFacility;
```
ttbarnes, line 220: `AtLeast`?

portal-api/src/v1/gef/controllers/application-submit.js
```diff
@@ -187,38 +187,41 @@ const submissionPortalActivity = async (application) => {
...
+    const isAIN = app.submissionType === CONSTANTS.DEAL.SUBMISSION_TYPE.AIN;
+
+    if (isSubmittedToUkef && !hasFacilities) {
+      return false;
     }
+
+    const updatedFacilities = await Promise.all(
+      facilities.reduce((promises, facility) => {
+        const isIssuedFacility = facility.hasBeenIssued && !facility.coverDateConfirmed;
+        const isUnissuedFacility = !facility.hasBeenIssued && facility.coverDateConfirmed;
```
ttbarnes, line 207: maybe could destructure `hasBeenIssued` and `coverDateConfirmed`?

portal-api/src/v1/gef/controllers/application-submit.js
AlexBramhill: I'm not modifying the actual functions here -- it was originally negated, and I double negated it. We're not wanting this PR to scope creep -- it could be that we care about `>=1` but I'm avoiding heavy changes

portal-api/src/drivers/fileshare.js
ttbarnes: I've found this pattern to be helpful when debugging, including returning what is resolved in the promise(s)

portal-api/src/v1/gef/controllers/application-submit.js
ttbarnes: Understood RE scope creep, but the `submisssionCount` check and variable name has already been changed? IMO `>= 1`  would be clearer

ttbarnes (commented):
Can't see any unit tests updated, notably for the `return;` changes - do we not have unit tests for the relevant controllers etc?

dtfs-central-api/src/v1/controllers/portal/gef-deal/add-min-activities.controller.js
ttbarnes: Thanks @JoshBinns2000 that make sense RE readability. Happy with either approach.

portal-api/src/drivers/fileshare.js
JoshBinns2000: Because this function is marked as `async`, it will always return a list of resolved promises as long as the function is called with `await` (ie. `return Promise.all` is the same as `return await Promise.all` in terms of the result)

portal-api/src/v1/gef/controllers/application-submit.js
AlexBramhill: updated

portal-api/src/v1/gef/controllers/application-submit.js
AlexBramhill: nice

portal-api/src/v1/controllers/deal-status/update-submitted-issued-facilities.js
AlexBramhill: I'm not sure the previous implimentation did?

AlexBramhill: ie the following
```
const item = { subItems: [{ id: 1 }, { id: 2 }, { id: 3 }] };

item.subItems.forEach((subItem) => {
  return { id: subItem.id + 10 };
});

console.dir({ item }, { depth: null });
// {
//  item: { subItems: [ { id: 1 }, { id: 2 }, { id: 3 } ] }
// }
```

ttbarnes: Thanks @AlexBramhill I can see it didn't used to so all good

AlexBramhill:
> Can't see any unit tests updated, notably for the `return;` changes - do we not have unit tests for the relevant controllers etc?

Not sure why we'd need to update any here -- No testable behaviour has changed (the `return` you mentioned existed anyway, it already returned `void`. 

The only `behaviour` change is handling async properly, which is about as low level as people go in js :)  (and testing this would involve writing tests that target and explore JS event loop behaviour)

abhi-markan:
Closing this PR as this has been moved back to backlog. To be picked up when back in the sprint.

PR #3375, opened 2024-08-07
fix(dtfs2-7312): update jsdocs to avoid not awaiting promises

AlexBramhill:
## Introduction :pencil2:
If a jsdoc does not return a promise, we can call async functions without having to await them (no errors).

Updated jsdocs to reflect async functions

Remove unneeded types from ts file


gef-ui/server/controllers/unissued-facilities/index.js
```diff
@@ -154,7 +153,7 @@ const changeIssuedToUnissuedFacility = async (req, res) => {
  * validates first and gets parameters {Object} from validation function
  * displays success message and redirects to unissued facilities list
  * @param {req}
- * @returns {res}
+ * @param {res}
```
abhi-markan, line 156: Can we also please add `@returns`

external-api/src/v1/controllers/acbs.controller.ts
```diff
@@ -245,9 +245,9 @@ const amendAcbsFacility = async (amendment: Amendment) => {
 /**
  * ACBS facility amendment entry function.
  * Constructs acceptable payload by DOF.
- * @param {Object} req Request
- * @param {Object} res Response
- * @return {Object} Response object with HTTP code as `status` and response as `data`.
+ * @param req Request
+ * @param res Response
+ * @return Response object with HTTP code as `status` and response as `data`.
```
abhi-markan, line 250: Should be `@returns`

abhi-markan (changes requested):

gef-ui/server/controllers/unissued-facilities/index.js
JoshBinns2000: This docstring seems not to be correct as `req` is not a type but is being used as the type in this jsdoc. Could we update this to instead be `@param {object} req` or `@param {import('express').Request} req`? In terms of the return type, this function seems to return either a `res.redirect` or `res.render` which are both functions which return `void`, so not including the `@returns` tag is semantically correct as returning `void` is the equivalent of not returning anything. However, if we want to include the `returns` tag, then please could it be `@returns {void}`

external-api/src/v1/controllers/acbs.controller.ts
AlexBramhill: udpated

gef-ui/server/controllers/unissued-facilities/index.js
AlexBramhill: I'm not updating the doc strings here for these, as most of `res` `req` are also wrong in the same way, and it'll be it's own ticket to bring all of these into being correct (this ticket is addressing the issues regarding promises).

For the sake of completionism, I've added a return as promise void.

JoshBinns2000 (approved):

gef-ui/server/controllers/unissued-facilities/index.js
```diff
@@ -12,8 +12,7 @@ const { validationErrorHandler } = require('../../utils/helpers');
  * creates body/parameters for the template for unissued facilities
  * if change true, changes 'cancel' + 'back' button href back to application preview
  * else renders back to unissued facilities list
- * @param {req} params, @param {req} query, @param {Boolean} change
- * @returns {Object} body
+ * @returns {Promise<object>} body
  */
 const renderChangeFacilityPartial = async ({ params, query, change, userToken }) => {
```
abhi-markan, line 17: Missing @params.

gef-ui/server/controllers/unissued-facilities/index.js
abhi-markan: Agree with @JoshBinns2000 on `@param {object} req` , however it will be a separate PR and needs to be scoped in @AlexBramhill  could you please create a Jira ticket pertinent to the JSDoc please and then we can close this PR. Thanks

abhi-markan (changes requested):

gef-ui/server/controllers/unissued-facilities/index.js
AlexBramhill: updated

gef-ui/server/controllers/unissued-facilities/index.js
AlexBramhill: Will add a ticket

abhi-markan (dismissed):

abhi-markan (approved):

PR #3377, opened 2024-08-08
feat(FN-1725): util to create keying sheet tfm facilities

JoshBinns2000:
## Introduction :pencil2:
The SQL seeder inserts entries in the `FacilityUtilisationData` table, where the `id` column of this table should correspond to a Mongo DB facility `ukefFacilityId`. This PR adds a new util script which inserts these entries after the seeder is run so that the keying sheet functionality works without error.

- Adds a new util
- Makes the `npm run load` command run all the local data-loading related scripts
- Updates the test pipeline so that only the mock data loader gets run before each test step

- Fix a bug where mock data loader doesn't actually clear the `tfm facilities` collection
- Updates the utils `README.md`

### DIFF
--- a/utils/README.md
+++ b/utils/README.md
@@ -8,6 +8,60 @@ The `data-migration` directory contains scripts designed to facilitate the migra
 
 ## Mock Data Loader :page_with_curl:
 
-The `mock-data-loader` directory contains mock data that can be used for local development and in non-production environments. To learn more about how to use this mock data, please refer to the [mock-data-loader/README.md](mock-data-loader/README.md) file.
+The `mock-data-loader` directory contains mock data that can be used for local development and in non-production environments.
+
+You can run mock data loader directly using
+
+```shell
+npm run mock-data-loader
+```
+
+or as part of the all-purpose
+
+```shell
+npm run load
+```
+
+command.
+
+## SQL DB Seeder
+
+The `sql-db-seeder` directory contains a script which seeds random data into the SQL database for utilisation reports.
+
+You can run the SQL seeder directly using
+
+```shell
+npm run db:seed
+```
+
+or as part of the all-purpose
+
+```shell
+npm run load
+```
+
+command.
+
+## Create Keying Sheet TFM Facilities
+
+The `create-keying-sheet-tfm-facilities` directory contains a script which inserts data into the Mongo DB database to line up with the data inserted by the `sql-db-seeder`. This is needed as the "keying sheet" related functionality requires that the `FeeRecord` facility ids which are inserted correspond the UKEF facility ids in the Mongo DB `facilities` and `tfm-facilities` colletion.
+
+You can run this script directly using
+
+```shell
+npm run create-keying-sheet-tfm-facilities
+```
+
+or indirectly via both
+
+```shell
+npm run db:seed
+```
+
+or
+
+```shell
+npm run load
+```
 
 ---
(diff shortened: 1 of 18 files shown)


utils/create-keying-sheet-tfm-facilities/database-client/mongo-db-client.ts
```diff
@@ -0,0 +1,17 @@
...
+const mongoDbConfigSchema = z.object({
+  MONGODB_URI: z.string(),
+  MONGO_INITDB_DATABASE: z.string(),
+});
+
+const getMongoDbClient = (): MongoDbClient => {
+  const mongoDbConfig = mongoDbConfigSchema.parse(process.env);
+  return new MongoDbClient({
+    dbName: mongoDbConfig.MONGO_INITDB_DATABASE,
+    dbConnectionString: mongoDbConfig.MONGODB_URI,
```
RLavender98, line 13: I might be confused, but why does this not use `MONGODB_URI_QA`?

utils/create-keying-sheet-tfm-facilities/index.ts
```diff
@@ -0,0 +1,38 @@
...
+    const validGefDealIdGenerator = await getValidGefDealIdGenerator();
+
+    const facilities: Facility[] = facilityIdsToInsert.map((ukefFacilityId) => ({
+      ...aFacilityWithoutDealId(),
+      ukefFacilityId,
+      dealId: validGefDealIdGenerator(),
+    }));
+    await Promise.all(facilities.map((facility) => FacilityClient.insertIfNotExists(facility)));
+
+    const tfmFacilities: WithoutId<TfmFacility>[] = facilities.map((facilitySnapshot) => ({ facilitySnapshot }));
```
RLavender98, line 26: Do you think we should add some with amendments?

RLavender98 (commented):
Looks good, just a couple of questions. Please don't forget when merging to mention the change on UKEF slack, in particular I would be surprised if people not working on utilisation reports have actually been running sql db migrations so they might get caught out when running `npm run load`

utils/create-keying-sheet-tfm-facilities/database-client/mongo-db-client.ts
JoshBinns2000: The reason initially was that in order to run this against the feature DB I was updating the `MONGODB_URI` env variable and not realising that the db client was actually using the `MONGODB_URI_QA` variable. However, I think it makes more sense to keep things consistent and use `MONGODB_URI_QA` here so I'll update that

utils/create-keying-sheet-tfm-facilities/index.ts
JoshBinns2000: I considered it and eventually decided against as I assumed that, if people needed amendments, they could manually insert them. That seems like a bit of an unreasonable assumption now that I type it out, though, so I think I will randomly add in some amendments

utils/README.md
```diff
@@ -8,6 +8,60 @@ The `data-migration` directory contains scripts designed to facilitate the migra
...
+
+```shell
+npm run load
+```
+
+command.
+
+## Create Keying Sheet TFM Facilities
+
+The `create-keying-sheet-tfm-facilities` directory contains a script which inserts data into the Mongo DB database to line up with the data inserted by the `sql-db-seeder`. This is needed as the "keying sheet" related functionality requires that the `FeeRecord` facility ids which are inserted correspond the UKEF facility ids in the Mongo DB `facilities` and `tfm-facilities` colletion.
```
AlexBramhill, line 47: nice docs thanks

RLavender98 (approved):

PR #3378, opened 2024-08-08
feat(FN-1728): auto reconciliation of zero action keying sheet rows

RLavender98:
## Introduction :pencil2:
If a keying sheet row has no actions we should automatically move it to status reconciled

- If a fee record has no payments, no fixed fee adjustment and no principal balance adjustment then it should move to status 'RECONCILED' when we generate keying data


dtfs-central-api/src/services/state-machines/fee-record/event-handlers/generate-keying-data/generate-keying-data.event-handler.ts
RLavender98, line 1: There is a technical possibility that if the report only consists of fee records with zero fees paid to ukef for the period then the entire report could auto reconcile. I haven't added that check as I am waiting to hear back on if that is actually possible (it's not included in any ACs currently)

dtfs-central-api/src/services/state-machines/fee-record/event-handlers/generate-keying-data/generate-keying-data.event-handler.test.ts
```diff
@@ -32,121 +33,224 @@ describe('handleFeeRecordGenerateKeyingDataEvent', () => {
   const aMatchingFeeRecord = () => FeeRecordEntityMockBuilder.forReport(aReconciliationInProgressReport()).withStatus('MATCH').build();
 
   beforeEach(() => {
-    jest.resetAllMocks();
     jest.mocked(calculateFixedFeeAdjustment).mockResolvedValue(10);
```
sw-mattw, line 36: Is this line not redundant as the last mock implementation takes precedence?

sw-mattw (commented):
Looks good, one query comment.

Are we waiting to hear back about that report state before merging or would that be covered by a different PR?

dtfs-central-api/src/services/state-machines/fee-record/event-handlers/generate-keying-data/generate-keying-data.event-handler.test.ts
RLavender98: Ah whoops yes, the line below should be calculatePrincipleBalanceAdjustment! Will update now

RLavender98:
> Are we waiting to hear back about that report state before merging or would that be covered by a different PR?
Not heard back yet so I reckon a separate PR if it turns out it is needed

sw-mattw (approved):

PR #3379, opened 2024-08-08
feat(FN-3187): display many-to-many payments on keying sheet

JoshBinns2000:
## Introduction :pencil2:
We want to display many-to-many payments on the TFM keying sheet.

- Moves keying sheet related mapping to one place to isolate mapping testing
- Splits the payment amounts across fee records at the point where keying data is generated and persists this value on a new column on the fee record payment join table
- Uses the value stored on the join table when generating the keying data (this makes the creation of the keying sheet independent of whether or not the relation is one-to-many, many-to-one or many-to-many)


dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-report-reconciliation-details-by-id.controller/helpers/map-fee-record-payment-entity-groups-to-keying-sheet.ts
```diff
@@ -1,65 +1,88 @@
...
+      feePayments: [
+        {
+          ...mapFeeRecordEntityToReportedPayments(feeRecord),
+          dateReceived,
+        },
+      ],
+    }));
+  }
+
+  return getKeyingSheetRowsWithPaymentsAcrossFeeRecords({ feeRecords, payments });
```
RLavender98, line 85: I think we need to do the splitting once per group when we generate the keying data and then save the split values in the database against the fee records somehow and then read it from there, rather than split as we surface to the ui. We should do this because we want to be certain that the values displayed in the table are what PDC keyed when looking back at the table, so we don't want any uncertainty, the particular case being if we ever decide to change the way we split payments, previously keyed records should still display their split the same.

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-report-reconciliation-details-by-id.controller/helpers/map-fee-record-payment-entity-groups-to-keying-sheet.ts
```diff
@@ -1,65 +1,88 @@
...
-    amount: amountAbsoluteValue,
-    change: amount > 0 ? 'INCREASE' : 'DECREASE',
-  };
-};
+  return feeRecordsSortedDescending.map((feeRecord) => {
+    const feePaymentsForRow: KeyingSheetFeePayment[] = [];
+    let remainingFeeRecordAmount = new Big(feeRecord.getFeesPaidToUkefForThePeriodInThePaymentCurrency());
+    while (remainingFeeRecordAmount.gt(0)) {
+      const largestFeePayment = allFeePaymentsSortedAscending.pop()!;
+      feePaymentsForRow.push(largestFeePayment);
```
RLavender98, line 30: I'm not sure that I like that we add the full payment and then reassign the amount later on, would it be possible to figure out the amount here and then add it with the correct amount?

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-report-reconciliation-details-by-id.controller/helpers/map-fee-record-payment-entity-groups-to-keying-sheet.ts
```diff
@@ -1,65 +1,88 @@
...
 
-  return {
-    amount: amountAbsoluteValue,
-    change: amount > 0 ? 'INCREASE' : 'DECREASE',
-  };
-};
+  return feeRecordsSortedDescending.map((feeRecord) => {
+    const feePaymentsForRow: KeyingSheetFeePayment[] = [];
+    let remainingFeeRecordAmount = new Big(feeRecord.getFeesPaidToUkefForThePeriodInThePaymentCurrency());
+    while (remainingFeeRecordAmount.gt(0)) {
```
RLavender98, line 28: I have a feeling something will go wrong if there is a payment matching tolerance in effect and the fee record amount total is greater than the payments total. Either this will hang forever or if it doesn't it will be because something inside throws is my suspicion, I don't know off the top of my head what happens when you attempt to pop from an empty array

RLavender98 (changes requested):
Good first pass, I think the algorithm of choice should be good enough for the use cases we expect. A few things to address!

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-report-reconciliation-details-by-id.controller/helpers/map-fee-record-payment-entity-groups-to-keying-sheet.ts
JoshBinns2000: Popping from an empty array returns `undefined` I believe - technically, `pop` on an array of type `T[]` returns `T | undefined`, but I am overriding the with `pop()!`. Not sure how we want to integrate this with tolerance, but it is definitely going to require some more thinking on my end. I'll probably address this change from a test-driven POV and see what I can come up with

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-report-reconciliation-details-by-id.controller/helpers/map-fee-record-payment-entity-groups-to-keying-sheet.ts
JoshBinns2000: Given that I need to take tolerances into account, I will probably need to calculate the actual amount here anyway like you say. I originally didn't because it bloats the code quite a lot of perform that calculation in here

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-report-reconciliation-details-by-id.controller/helpers/get-keying-sheet-for-report-id.ts
```diff
@@ -0,0 +1,68 @@
...
+
+/**
+ * Gets the keying sheet for the report with id matching the supplied report id
+ * @param reportId - The report id
+ * @returns The keying sheet
+ */
+export const getKeyingSheetForReportId = async (reportId: number): Promise<KeyingSheet> => {
+  const joinTableEntities = await getAndValidateKeyingSheetFeeRecordPaymentJoinTableEntries(reportId);
+
+  const feeRecordIdToKeyingSheetRowMap = joinTableEntities.reduce(
```
RLavender98, line 38: What happens for zero payment fee records after this refactor?

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/helpers/update-fee-record-payment-join-table-payment-amounts-used-for-fee-record.ts
```diff
@@ -0,0 +1,150 @@
...
+    updatedFeeRecordPaymentJoinTableEntities.map(async ({ feeRecordId, paymentId, paymentAmountUsedForFeeRecord }) => {
+      try {
+        await entityManager.update(FeeRecordPaymentJoinTableEntity, { feeRecordId, paymentId }, { paymentAmountUsedForFeeRecord });
+      } catch (error) {
+        console.error(`Failed to update fee record payment join table entity with feeRecordId '${feeRecordId}' and paymentId '${paymentId}':`, error);
+        throw error;
+      }
+    }),
+  );
+};
```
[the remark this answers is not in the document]
JoshBinns2000: The reason I wanted to avoid using multiple files for this (even with the payment splitting) is because there isn't really a way to get around separating the payment splitting logic from mutating the list of payments without making code that is much harder to read than what I have currently written in my opinion (ie. passing around array indices or creating new versions of the payments array in memory which don't actually get cleared by the GC due to the objects in the array themselves not going out of scope). By keeping everything in the same file, I think it's much easier to figure out what the side effects of these functions are. If we want to highlight that splitting is taking place, I think changing the function name and return type would suffice, but I am hesistant to remove functions with side effects from the context where the side effects make sense.

dtfs-central-api/src/v1/controllers/utilisation-report-service/post-keying-data.controller/index.ts
```diff
@@ -16,7 +16,7 @@ export const postKeyingData = async (req: PostKeyingDataRequest, res: Response)
   const { user } = req.body;
 
   try {
-    const feeRecordsAtMatchStatus = await FeeRecordRepo.findByReportIdAndStatusesWithReport(Number(reportId), ['MATCH']);
+    const feeRecordsAtMatchStatus = await FeeRecordRepo.findByReportIdAndStatusesWithReportAndPayments(Number(reportId), ['MATCH']);
```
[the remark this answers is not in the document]
JoshBinns2000: This now loads payments on the fee records as, when updating the join table as part of the keying data generation, we first need to construct the fee record payment groups so that we can split the payments. If you do this again in the event handler, you would have to load every fee record again with all payments attached, which feels like a case where it makes more sense to do it as part of the initial repo call

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/helpers/update-fee-record-payment-join-table-payment-amounts-used-for-fee-record.ts
```diff
@@ -0,0 +1,150 @@
...
+ */
+const getFeeRecordPaymentJoinTableEntitiesForFeeRecordAndPayments = (
+  feeRecord: FeeRecordEntity,
+  paymentsSortedAscending: PaymentEntity[],
+): FeeRecordPaymentJoinTableEntity[] => {
+  const feeRecordId = feeRecord.id;
+  let remainingFeeRecordAmount = new Big(feeRecord.getFeesPaidToUkefForThePeriodInThePaymentCurrency());
+
+  const joinTableEntities: FeeRecordPaymentJoinTableEntity[] = [];
+  while (remainingFeeRecordAmount.gt(0)) {
```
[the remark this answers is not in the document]
JoshBinns2000: Fortunately we wouldn't end up in an infinite loop as `.pop` would give `undefined` if the list is empty and `.amount` would then throw a `TypeError`, so I left it in somewhat intentionally as something should really go wrong if the payments list ends up empty at some point during this algorithm - I'm going to spend some time thinking about whether or not we want that behaviour though

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-report-reconciliation-details-by-id.controller/helpers/get-keying-sheet-for-report-id.ts
```diff
@@ -0,0 +1,68 @@
...
+import { mapFeeRecordEntityToKeyingSheetRowWithoutFeePayments } from '../../../../../mapping/keying-sheet-mapping';
+import { KeyingSheet, KeyingSheetFeePayment, KeyingSheetRow } from '../../../../../types/fee-records';
+
+/**
+ * Gets all the keying sheet fee record payment join table entries
+ * linked to the report with id matching the supplied report id
+ * @param reportId - The report id
+ * @returns The join table entries linked to the report id
+ */
+const getAndValidateKeyingSheetFeeRecordPaymentJoinTableEntries = async (reportId: number): Promise<FeeRecordPaymentJoinTableEntity[]> => {
```
ChristianMcc, line 13: is the `andValidate` redundant here?

ChristianMcc (commented):
Left a nit pick, the fetching of data I wanted to look at via the join table LGTM. It's not my favourite piece of code but I agree it's the best we can do in time right now.

Have you tested this with like 200+ records being created with keying sheet data and then fetching that whole table on the keying sheet page?

I think it should be fine as its one large query that returns a bunch of data with a few joins but I'm aware with the relations we are fetching a lot of data at once :D

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/helpers/update-fee-record-payment-join-table-payment-amounts-used-for-fee-record.ts
```diff
@@ -0,0 +1,150 @@
...
+    const largestPayment = paymentsSortedAscending.pop()!;
+
+    remainingFeeRecordAmount = remainingFeeRecordAmount.sub(largestPayment.amount);
+
+    if (remainingFeeRecordAmount.lt(0)) {
+      const paymentAmountUsedForFeeRecord = new Big(largestPayment.amount).add(remainingFeeRecordAmount).toNumber();
```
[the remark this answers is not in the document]
JoshBinns2000: I have updated the way this loop works - hopefully it reads more easily now

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-report-reconciliation-details-by-id.controller/helpers/get-keying-sheet-for-report-id.ts
JoshBinns2000: Was removed by accident - I have re-included it and added more test cases to cover it

RLavender98 (commented):
Just one question, otherwise all looks good!

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-report-reconciliation-details-by-id.controller/helpers/get-keying-sheet-for-report-id.test.ts
```diff
@@ -272,6 +250,136 @@ describe('getKeyingSheetForReportId', () => {
...
+    // Assert
+    expect(keyingSheet).toHaveLength(1);
+    expect(keyingSheet[0].feePayments).toHaveLength(2);
+    expect(keyingSheet[0].feePayments[0].currency).toBe('GBP');
+    expect(keyingSheet[0].feePayments[0].amount).toBe(1000);
+    expect(keyingSheet[0].feePayments[0].dateReceived).toEqual(new Date('2021'));
+    expect(keyingSheet[0].feePayments[1].currency).toBe('USD');
+    expect(keyingSheet[0].feePayments[1].amount).toBe(0);
+    expect(keyingSheet[0].feePayments[1].dateReceived).toBeNull();
+  });
```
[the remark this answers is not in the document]
JoshBinns2000: Yes it would. I suppose we only want a zero fee payment to appear in the case when there are no other fee payments, so I'll update the logic and test to do that

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-report-reconciliation-details-by-id.controller/helpers/get-keying-sheet-for-report-id.ts
JoshBinns2000: There was originally some validation around this, but that was removed without updating the name 🙃

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-report-reconciliation-details-by-id.controller/helpers/get-keying-sheet-for-report-id.ts
```diff
@@ -24,28 +27,37 @@ const getAndValidateKeyingSheetFeeRecordPaymentJoinTableEntries = async (reportI
...
+  );
 
   const feeRecordIdToKeyingSheetRowMap = joinTableEntities.reduce(
     (map, { feeRecord, payment, paymentAmountUsedForFeeRecord }) => {
-      if (paymentAmountUsedForFeeRecord === null) {
-        throw new Error('Expected fee record at READY_TO_KEY or RECONCILED status to have a defined paymentAmountUsedForFeeRecord on the join table');
-      }
-
       const feeRecordId = feeRecord.id;
+      remove(keyingSheetFeeRecordsWithZeroFeePayment, ({ id }) => id === feeRecordId);
```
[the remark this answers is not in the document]
JoshBinns2000: Replied offline, but tl;dr we may as well load everything at once rather than loading all the fee records twice

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/helpers/get-keying-sheet-fee-payment-shares-for-fee-records.ts
```diff
@@ -3,7 +3,7 @@ import { FeeRecordEntity } from '@ukef/dtfs2-common';
 import { orderBy } from 'lodash';
 import { FeeRecordPaymentEntityGroup, getFeeRecordPaymentEntityGroupsFromFeeRecordEntities } from '../../../../../helpers';
 
-export type FeeRecordFeePayment = {
+export type KeyingSheetFeePaymentShare = {
```
RLavender98, line 6: Nice I like this name thanks!

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-report-reconciliation-details-by-id.controller/helpers/get-keying-sheet-for-report-id.ts
```diff
@@ -0,0 +1,89 @@
...
+      if (paymentAmountUsedForFeeRecord === null) {
+        return map;
+      }
+
+      const feeRecordId = feeRecord.id;
+      remove(keyingSheetFeeRecordsWithZeroFeePayment, ({ id }) => id === feeRecordId);
+
+      const feePayment: KeyingSheetFeePayment = {
+        currency: payment.currency,
+        amount: paymentAmountUsedForFeeRecord ?? 0,
```
RLavender98, line 63: I think this should now only be 0 or have a value so I don't think we need the `?? 0`?

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-report-reconciliation-details-by-id.controller/helpers/get-keying-sheet-for-report-id.test.ts
```diff
@@ -0,0 +1,405 @@
...
+    expect(keyingSheet[0].feePayments).toHaveLength(3);
+    expect(keyingSheet[0].feePayments[0].currency).toBe('GBP');
+    expect(keyingSheet[0].feePayments[0].dateReceived).toEqual(new Date('2021'));
+    expect(keyingSheet[0].feePayments[1].currency).toBe('USD');
+    expect(keyingSheet[0].feePayments[1].dateReceived).toEqual(new Date('2022'));
+    expect(keyingSheet[0].feePayments[2].currency).toBe('EUR');
+    expect(keyingSheet[0].feePayments[2].dateReceived).toEqual(new Date('2023'));
+  });
+
+  it('sets the keying sheet row fee payments only to the payment with a non-null paymentAmountUsedForFeeRecord column', async () => {
```
RLavender98, line 253: Could we add a test for a fee record, which has payments, but all the fee payments are null, that checks it gets set to a zero payment, no date, keying sheet row? The logic looks correct and should get handled similarly to the no payments case below, but would be goof to have test coverage.

RLavender98 (commented):
Just two small comments around the zero payments case!

RLavender98 (approved):
Nice job! Sorry for the the number of rounds of comments! 😅

PR #3381, opened 2024-08-08
feat(FN-3226): add util to check database facility ids

JoshBinns2000:
## Introduction :pencil2:
We want to add a script which can be used to check that, for each `facilityId` in the SQL `FeeRecord` table, there exists a document in the Mongo DB `tfm-facilities` collection with a matching `ukefFacilityId`.

- Adds the script
- Adds a utils npm script
- Updates the README


utils/validate-facility-ids/database-client/mongo-db-client.ts
```diff
@@ -0,0 +1,17 @@
+import z from 'zod';
+import { MongoDbClient } from '@ukef/dtfs2-common/mongo-db-client';
+
+const mongoDbConfigSchema = z.object({
+  MONGODB_URI_QA: z.string(),
+  MONGO_INITDB_DATABASE: z.string(),
```
RLavender98, line 6: Could we add in the README section that these need to be set to the values for the env you want to run against?

utils/validate-facility-ids/index.ts
```diff
@@ -0,0 +1,37 @@
...
+import { mongoDbClient } from './database-client';
+import { getUniqueFeeRecordFacilityIds, getUniqueTfmFacilitiesUkefFacilityIds } from './helpers';
+
+const run = async (): Promise<void> => {
+  const dataSource = await SqlDbDataSource.initialize();
+  try {
+    const tfmFacilitiesUkefFacilityIds = await getUniqueTfmFacilitiesUkefFacilityIds();
+    const feeRecordFacilityIds = await getUniqueFeeRecordFacilityIds(dataSource);
+
+    const tfmFacilitiesFacilityIdsNotInFeeRecord = difference(tfmFacilitiesUkefFacilityIds, feeRecordFacilityIds);
```
RLavender98, line 12: Nit pick but I think this should end with an 's'

utils/validate-facility-ids/index.ts
```diff
@@ -0,0 +1,37 @@
...
+    const feeRecordFacilityIdsNotInTfmFacilities = difference(feeRecordFacilityIds, tfmFacilitiesUkefFacilityIds);
+
+    if (tfmFacilitiesFacilityIdsNotInFeeRecord.length === 0 && feeRecordFacilityIdsNotInTfmFacilities.length === 0) {
+      console.info('All fee record facility ids exist in the tfm facilities collection');
+      return;
+    }
+
+    tfmFacilitiesFacilityIdsNotInFeeRecord.forEach((facilityId) => {
+      console.info(`Facility id '${facilityId}' exists in the "tfm-facilities" collection but not in the FeeRecord table`);
+    });
```
RLavender98, line 22: Is it not expected that there will be lots of facility ids without fee records and we don't care about that? e.g. banks which use the portal but not for utilisation reporting? I think we only care about fee records without a corresponding facility

utils/validate-facility-ids/index.ts
JoshBinns2000: This is now removed but I didn't include an 's' at the end because `FeeRecord` is a table

RLavender98 (approved):

ttbarnes (commented):
Could we look at resolving some of the `any` Codacy issues reported?

utils/validate-facility-ids/helpers/get-unique-fee-record-facility-ids.ts
```diff
@@ -0,0 +1,8 @@
+import { DataSource } from 'typeorm';
+import { FeeRecordEntity } from '@ukef/dtfs2-common';
+
+export const getUniqueFeeRecordFacilityIds = async (dataSource: DataSource): Promise<string[]> => {
```
ttbarnes, line 4: Should we add some documentation?

utils/validate-facility-ids/index.ts
```diff
@@ -0,0 +1,32 @@
+import { difference } from 'lodash';
+import { SqlDbDataSource } from '@ukef/dtfs2-common/sql-db-connection';
+import { mongoDbClient } from './database-client';
+import { getUniqueFeeRecordFacilityIds, getUniqueTfmFacilitiesUkefFacilityIds } from './helpers';
+
+const run = async (): Promise<void> => {
+  const dataSource = await SqlDbDataSource.initialize();
+  try {
```
ttbarnes, line 8: Should we add a `console.info` here? 

Can we add documentation?

utils/validate-facility-ids/index.ts
```diff
@@ -0,0 +1,32 @@
...
+    const feeRecordFacilityIdsNotInTfmFacilities = difference(feeRecordFacilityIds, tfmFacilitiesUkefFacilityIds);
+
+    if (feeRecordFacilityIdsNotInTfmFacilities.length === 0) {
+      console.info('All fee record facility ids exist in the tfm facilities collection');
+      return;
+    }
+
+    feeRecordFacilityIdsNotInTfmFacilities.forEach((facilityId) => {
+      console.info(`Facility id '${facilityId}' exists in the FeeRecord table but not in the "tfm-facilities" collection`);
+    });
```
ttbarnes, line 21: Should we return `feeRecordFacilityIdsNotInTfmFacilities`? Would also avoid having a `Promise<void>` return type

JoshBinns2000: This util is only here to display which facility ids do not match up - there is not benefit in returning the value (it would be immediately cleared up by the GC due to the script ending) and I don't see issues with returning `Promise<void>`

utils/validate-facility-ids/index.ts
JoshBinns2000: Done

utils/validate-facility-ids/helpers/get-unique-fee-record-facility-ids.ts
JoshBinns2000: Done

JoshBinns2000:
@ttbarnes The codacy issues are a common occurrence when working with typescript in this repo - fixing would require modifying the way codacy scans this repo which is beyond the scope of this work

RLavender98 (approved):

utils/validate-facility-ids/database-client/mongo-db-client.ts
```diff
@@ -0,0 +1,21 @@
...
+ */
+const getMongoDbClient = (): MongoDbClient => {
+  const mongoDbConfig = mongoDbConfigSchema.parse(process.env);
+  return new MongoDbClient({
+    dbName: mongoDbConfig.MONGO_INITDB_DATABASE,
+    dbConnectionString: mongoDbConfig.MONGODB_URI_QA,
+  });
+};
+
+export const mongoDbClient = getMongoDbClient();
```
ttbarnes, line 21: Noticed there is already a DB client in `utils/mock-data-loader/database` - maybe could extra this into the `utils` directory (or the other way around) to be DRY?

JoshBinns2000: Each of the mongo clients may technically need to use different connection strings depending on the task being performed, hence each util having its own db client. If we want to commonise this further down the line then we could, but the files are so small that there isn't much harm in keeping the multiple versions for now

PR #3382, opened 2024-08-09
feat(FN-3187): create entity for fee record payment join table

JoshBinns2000:
## Introduction :pencil2:
We want to create an entity for the existing join table so that we can add columns to it. This entity must be defined such that running the `npm run db:generate-miragtion` script in `libs/common` does not create a new migration (ie. all the indices and constraints should be given the same name as what they were in the migration where this table was initially added).

- Adds the new entity


libs/common/src/sql-db-entities/join-tables/fee-record-payments-payment/fee-record-payments-payment.entity.ts
```diff
@@ -0,0 +1,29 @@
...
+  @JoinColumn({ name: 'feeRecordId', foreignKeyConstraintName: 'FK_7a9b7aa849fc3bb09d80fa1f812' })
+  feeRecord!: FeeRecordEntity;
+
+  @PrimaryColumn({ nullable: false })
+  @Index('IDX_23bbb10be5f2136a5a5086654e')
+  paymentId!: number;
+
+  @ManyToOne(() => PaymentEntity, (payment) => payment.id, {
+    eager: false,
+    onDelete: 'CASCADE',
```
sw-mattw, line 25: Could I just double check that we're wanting to delete the payment as well as the fee records on delete? I noticed that [this constraint in the migration has `NO ACTION`](https://github.com/UK-Export-Finance/dtfs2/blob/82b874c6e64e4b7b1a4e1f48da38d267278e572a/libs/common/src/sql-db-connection/migrations/1716990167763-AddPaymentTable.ts#L40)

sw-mattw (commented):
Nice, one minor query just to double check something

--

IDs/constraint names matched those in https://github.com/UK-Export-Finance/dtfs2/blob/main/libs/common/src/sql-db-connection/migrations/1716990167763-AddPaymentTable.ts

Ran `npm run db:migrate && npm run db:generate-migration` in `libs/common` and got `No changes in database schema were found - cannot generate a migration.` as expected (no migration generated).

libs/common/src/sql-db-entities/join-tables/fee-record-payments-payment/fee-record-payments-payment.entity.ts
sw-mattw: Ah yes, thanks

sw-mattw (approved):

PR #3388, opened 2024-08-13
feat(FN-1749): add handling to add selected fees to an existing payment group

sw-mattw:
## Introduction :pencil2:
Adds functionality to add fee records to an existing payment group in a 'does not match' state.

- Implemented a new POST route in Central API and TFM API to handle adding fees to an existing payment.
- New controller to handle the 'add to an existing payment' flow.
- Additional state machine event types:
  - Utilisation Report: `ADD_FEES_TO_AN_EXISTING_PAYMENT_GROUP` - calls the two events below
  - Fee Record: `PAYMENT_ADDED` (existing) updates status of fee records being added to the group, `OTHER_FEE_ADDED_TO_PAYMENT_GROUP` updates status of fee records already part of the group.
- Added validation that the payment IDs being passed through form a complete payment group (rather than a partial)
- UI changes to link up a new continue button with this flow.
- UI changes to only show radio input when there is more than one payment group to select from
- Dynamic page title (fee vs fees) change from FN-1745

### Screenshots

One payment group with a single payment - no radio input:
fn-1749-query-add-fees-to-existing-payment-one-group-single-payment

One payment group with many payments (one without a reference) - no radio input:
fn-1749-query-add-fees-to-existing-payment-one-group-multiple-payments

Many payment groups with a varying number of payments (with and without references) - radio input for each payment group:
fn-1749-query-add-fees-to-existing-payment-multiple-groups

N/A


libs/common/src/sql-db-entities/payment/payment.entity.test.ts
```diff
@@ -0,0 +1,50 @@
...
+      payment.updateWithAdditionalFeeRecords({
+        feeRecords: [newFeeRecord1, newFeeRecord2],
+        requestSource: { platform: 'TFM', userId },
+      });
+
+      // Assert
+      expect(payment.feeRecords).toEqual(expect.arrayContaining([existingFeeRecord, newFeeRecord1, newFeeRecord2]));
+    });
+
+    it('does not allow duplicate fee records to be added to the payment', () => {
```
RLavender98, line 29: I think we should rename this test as the code does allow us to add two new fee records which are identical which technically is counter to what this test claims it is testing

libs/common/src/sql-db-entities/payment/payment.entity.test.ts
```diff
@@ -0,0 +1,50 @@
...
+
+      const feeRecord = FeeRecordEntityMockBuilder.forReport(utilisationReport).withId(1).build();
+      payment.feeRecords = [feeRecord];
+
+      const userId = 'abc123';
+
+      // Act & Assert
+      expect(() => {
+        payment.updateWithAdditionalFeeRecords({
+          feeRecords: [feeRecord],
```
RLavender98, line 41: Please can we use a separate in memory object for the existing one and the new one, our code doesn't and shouldn't rely on them being the same in memory so good to avoid in the test too I think.

e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/add-to-an-existing-payment.spec.js
```diff
@@ -12,22 +12,25 @@ import { NODE_TASKS } from '../../../../../e2e-fixtures';
 import relative from '../../relativeURL';
 
 context(`${PDC_TEAMS.PDC_RECONCILE} users can add fee records to existing payments`, () => {
+  const PAYMENT_ID_ONE = '1';
```
RLavender98, line 15: If there is no second payment this could just be `PAYMENT_ID`

e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/add-to-an-existing-payment.spec.js
```diff
@@ -85,7 +88,14 @@ context(`${PDC_TEAMS.PDC_RECONCILE} users can add fee records to existing paymen
...
+    pages.utilisationReportAddToAnExistingPaymentPage.availablePaymentGroups().should('contain', 'GBP 450.00');
     pages.utilisationReportAddToAnExistingPaymentPage.availablePaymentGroups().should('contain', 'Payment reference: REF01234');
   });
+
+  it('should add the selected fees to the payments and update the fee record status after clicking the continue button', () => {
+    pages.utilisationReportAddToAnExistingPaymentPage.continueButton().click();
+
+    pages.utilisationReportPage.premiumPaymentsTab.getPaymentLink(PAYMENT_ID_ONE).should('contain', 'GBP 450.00');
+    pages.utilisationReportPage.premiumPaymentsTab.getPremiumPaymentsTableRow(FEE_RECORD_ID_ONE).should('contain', 'MATCH');
+  });
```
RLavender98, line 100: Aware this is a bit counter to some of the existing tests, but we should be looking to combine e2e-tests into testing a journey as a whole where possible, given the before each does quite a lot and therefore takes a bit of time. Could we perhaps combine this with the above two tests to have one test that covers all the checking of things showing on the page and then at the end clicks continue?

Also how much effort would it be to add a second test which has multiple payment groups to test the case with radio button?

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/add-fees-to-an-existing-payment-group/add-fees-to-an-existing-payment-group.event-handler.ts
```diff
@@ -0,0 +1,90 @@
...
+  await addSelectedFeeRecordsToPaymentGroup(transactionEntityManager, feeRecordsToAdd, feeRecordsAndPaymentsMatchAfterFeeRecordsAdded, payments, requestSource);
+  await updateOtherFeeRecordsInPaymentGroup(
+    transactionEntityManager,
+    otherFeeRecordsInPaymentGroup,
+    feeRecordsAndPaymentsMatchAfterFeeRecordsAdded,
+    requestSource,
+  );
+
+  report.updateLastUpdatedBy(requestSource);
+  return await transactionEntityManager.save(UtilisationReportEntity, report);
```
RLavender98, line 89: I think the latest on the decision on what should count as an update to a report is that it should just be if a field on that entity itself has changed, so here the changes are not to the report entity itself so we should just return it without any updates.

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/add-fees-to-an-existing-payment-group/add-fees-to-an-existing-payment-group.event-handler.test.ts
```diff
@@ -0,0 +1,202 @@
...
+  };
+
+  const mockSave = jest.fn();
+  const mockEntityManager = {
+    save: mockSave,
+  } as unknown as EntityManager;
+
+  const aReconciliationInProgressReport = () => UtilisationReportEntityMockBuilder.forStatus('RECONCILIATION_IN_PROGRESS').build();
+
+  const aFeeRecordForReport = (report: UtilisationReportEntity, id: number) => FeeRecordEntityMockBuilder.forReport(report).withId(id).build();
```
RLavender98, line 31: I think this line is short enough that we should just inline calls to this, especially as this is what the entity mock builders are designed for.

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/add-fees-to-an-existing-payment-group/add-fees-to-an-existing-payment-group.event-handler.test.ts
```diff
@@ -0,0 +1,202 @@
...
+
+  const aPaymentWithIdAndFeeRecords = (id: number, feeRecords: FeeRecordEntity[]) =>
```
RLavender98, line 33: Ditto to the above

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/add-fees-to-an-existing-payment-group/add-fees-to-an-existing-payment-group.event-handler.test.ts
```diff
@@ -0,0 +1,202 @@
...
+      ...secondPayment,
+      feeRecords: [feeRecords[1], feeRecords[2]],
+    });
+  });
+
+  it.each([
+    { feeRecordsAndPaymentsMatch: true, testNameSuffix: 'the fee records match the payments' },
+    { feeRecordsAndPaymentsMatch: false, testNameSuffix: 'the fee records do not match the payments' },
+  ])(
+    "calls the fee record state machine event handler for the fee records to add with 'feeRecordsAndPaymentsMatch' set to '$feeRecordsAndPaymentsMatch' if $testNameSuffix after the new fee records have been added",
```
RLavender98, line 103: Could we add in the name of the event being called?

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/add-fees-to-an-existing-payment-group/add-fees-to-an-existing-payment-group.event-handler.test.ts
```diff
@@ -0,0 +1,202 @@
...
+        });
+      });
+    },
+  );
+
+  it.each([
+    { feeRecordsAndPaymentsMatch: true, testNameSuffix: 'the fee records match the payments' },
+    { feeRecordsAndPaymentsMatch: false, testNameSuffix: 'the fee records do not match the payments' },
+  ])(
+    "calls the fee record state machine event handlers for unselected fee records with 'feeRecordsAndPaymentsMatch' set to '$feeRecordsAndPaymentsMatch' if $testNameSuffix after the new fee records have been added",
```
RLavender98, line 145: Instead of unselected can we stick with the term 'existing fee records'?

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/add-fees-to-an-existing-payment-group/add-fees-to-an-existing-payment-group.event-handler.ts
```diff
@@ -0,0 +1,90 @@
+import { EntityManager } from 'typeorm';
+import { DbRequestSource, FeeRecordEntity, PaymentEntity, UtilisationReportEntity } from '@ukef/dtfs2-common';
+import { BaseUtilisationReportEvent } from '../../event/base-utilisation-report.event';
+import { FeeRecordStateMachine } from '../../../fee-record/fee-record.state-machine';
+import { feeRecordsMatchAttachedPayments } from '../helpers';
+
+type AddFeesToAnExistingPaymentGroupEventPayload = {
+  transactionEntityManager: EntityManager;
+  feeRecordsToAdd: FeeRecordEntity[];
+  otherFeeRecordsInPaymentGroup: FeeRecordEntity[];
```
RLavender98, line 10: Rather than `otherFeeRecordsInPaymentGroup` I think this should be `existingFeeRecordsInPaymentGroup`

RLavender98 (changes requested):
Nice job! Mostly just small comments/changes! Please could you also do another comb over of the changes to add some more docstrings?

trade-finance-manager-ui/server/controllers/utilisation-reports/helpers/payment-form-helpers.test.ts
```diff
@@ -295,4 +297,73 @@ describe('payment-form-helpers', () => {
...
+        addToAnExistingPaymentFormSubmission: 'true',
+        paymentGroup: 'paymentIds-1,2,3',
+      };
+
+      // Act
+      const result = extractAddToAnExistingPaymentRadioPaymentIdsAndValidateIfPresent(requestBody);
+
+      // Assert
+      expect(result.errors).not.toEqual(EMPTY_PAYMENT_ERRORS_VIEW_MODEL);
+    });
```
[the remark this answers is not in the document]
sw-mattw: Yep, this was the issue where the validation function didn't have a mock implementation. I've removed this test and moved validation checks to the other tests - under commit 1451da14b (not pushed yet)

trade-finance-manager-ui/server/types/view-models/add-to-an-existing-payment-view-model.ts
```diff
@@ -2,6 +2,12 @@ import { CurrencyAndAmountString } from '@ukef/dtfs2-common';
 import { SelectedReportedFeesDetailsViewModel } from './selected-reported-fees-details-view-model';
 import { BaseViewModel } from './base-view-model';
 import { AddToAnExistingPaymentRadioId } from '../add-to-an-existing-payment-radio-id';
+import { PremiumPaymentsTableCheckboxId } from '../premium-payments-table-checkbox-id';
+import { PaymentErrorsViewModel } from './payment-errors-view-model';
+
+export type AddToAnExistingPaymentErrorsViewModel = PaymentErrorsViewModel & {
```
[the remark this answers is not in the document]
sw-mattw: That's a good point

trade-finance-manager-ui/component-tests/utilisation-reports/_macros/payment-group-radio-input-group.component-test.ts
```diff
@@ -18,21 +18,22 @@ describe(component, () => {
...
-    wrapper.expectElement(radioInputSelector).toHaveCount(viewModel.paymentGroups.length);
-
-    wrapper.expectElement(`${radioInputSelector}:eq(0)`).toHaveAttribute('id', 'paymentIds-1,2');
-    wrapper.expectElement(`${radioInputSelector}:eq(0)`).toHaveAttribute('value', 'paymentIds-1,2');
-
-    wrapper.expectElement(`${radioInputSelector}:eq(1)`).toHaveAttribute('id', 'paymentIds-3');
-    wrapper.expectElement(`${radioInputSelector}:eq(1)`).toHaveAttribute('value', 'paymentIds-3');
+    wrapper.expectElement('div.govuk-radios__item').toHaveCount(0);
+    wrapper.expectElement(`input[type='radio']`).notToExist();
+    wrapper.expectElement(`label[data-cy='payment-group-label--paymentIds-1']`).notToExist();
```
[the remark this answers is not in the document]
sw-mattw: Good point, added a check for this input and that it has type 'hidden'

e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/add-to-an-existing-payment.spec.js
sw-mattw: Kept this suffix as we now have a second payment (to test that multiple payment groups can be rendered)

libs/common/src/sql-db-entities/payment/payment.entity.test.ts
sw-mattw: Yep thanks, removed this test

libs/common/src/sql-db-entities/payment/payment.entity.test.ts
sw-mattw: This test has been removed (other comment) but yeah I agree it'd be better like the other test case

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/add-fees-to-an-existing-payment-group/add-fees-to-an-existing-payment-group.event-handler.ts
sw-mattw: Ah that's nice, removed this now thanks

e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/add-to-an-existing-payment.spec.js
sw-mattw: Yep, combined these now thanks. Added two additional test cases too:
- `should automatically select the first payment group when there is only one group to choose from`
- `should display an error message when there are multiple payments to choose from and none have been selected`

trade-finance-manager-ui/templates/utilisation-reports/_macros/payment-group-radio-input-group.njk
```diff
@@ -1,33 +1,51 @@
+{% import "./payment-group-radio-input-group-details.njk" as paymentGroupRadioInputGroupDetails %}
+
 {% macro render(params) %}
   {% set legendText = params.legendText %}
   {% set paymentGroups = params.paymentGroups %}
+  {% set errorMessage = params.errorMessage %}
 
-  <div class="govuk-form-group" data-cy="payment-group-radio-input">
+  {% if errorMessage %}
+    <div class="govuk-form-group--error" data-cy="fee-record-details-error-wrapper">
```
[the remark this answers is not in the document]
sw-mattw: Updated, added component tests for this

trade-finance-manager-ui/templates/utilisation-reports/add-to-an-existing-payment.njk
```diff
@@ -25,11 +40,26 @@
...
-  {{ paymentGroupRadioInputGroup.render({
-    legendText: availablePaymentsHeading,
-    paymentGroups: availablePaymentGroups
-  }) }}
+  <form method="post">
+    <input type="hidden" name="_csrf" value="{{ csrfToken }}"/>
+    <input type="hidden" name="addToAnExistingPaymentFormSubmission" value="true"/>
+    {% for id in selectedFeeRecordCheckboxIds %}<input type="hidden" name={{ id }} value="on"/>{% endfor %}
+
+    {{ paymentGroupRadioInputGroup.render({
```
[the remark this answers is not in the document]
sw-mattw: Renamed this and the existing payment details component thanks

dtfs-central-api/src/v1/controllers/utilisation-report-service/post-add-fees-to-an-existing-payment-group.controller/index.ts
```diff
@@ -0,0 +1,68 @@
...
+    if (payments.length === 0) {
+      throw new NotFoundError('No payments found for the given payment IDs.');
+    }
+
+    validateThatSelectedPaymentsBelongToSamePaymentGroup(payments);
+    validateThatSelectedPaymentsFormACompletePaymentGroup(payments, paymentIds);
+
+    const existingPaymentGroupFeeRecords = payments.at(0)?.feeRecords;
+    validateThatPaymentGroupHasFeeRecords(existingPaymentGroupFeeRecords);
+
```
RLavender98, line 43: I think ideally we would also repeat the validation that let's them onto this page in the first place, the status checks are covered off by the state machines, so that just leaves a check for currency matching between all of the fees payment currencies and the payments' currency.

libs/common/src/sql-db-entities/payment/payment.entity.test.ts
RLavender98: Sorry, I think wires got crossed here a bit, I think the test is good just would be better named: `does not allow adding a fee record which is already attached to payment`

e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/add-to-an-existing-payment.spec.js
```diff
@@ -67,24 +94,49 @@ context(`${PDC_TEAMS.PDC_RECONCILE} users can add fee records to existing paymen
...
+
+  it('should automatically select the first payment group when there is only one group to choose from', () => {
+    cy.task(NODE_TASKS.REMOVE_ALL_UTILISATION_REPORTS_FROM_DB);
+    cy.task(NODE_TASKS.REMOVE_ALL_PAYMENTS_FROM_DB);
+    cy.task(NODE_TASKS.INSERT_UTILISATION_REPORTS_INTO_DB, [report]);
+    cy.task(NODE_TASKS.INSERT_FEE_RECORDS_INTO_DB, [feeRecordOne, feeRecordTwo]);
+
+    cy.reload();
```
RLavender98, line 125: I'm torn about this, do we do this elsewhere? I think it seems a bit odd to essentially remove all the set up and then restart it and also I guess this is why we are inserting the report separately which also seems less ideal than inserting a report with the fee records attached... I don't know if we nest many contexts, but I wonder if we can have two contexts and then they have slightly differing `beforeEachs`?

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/add-fees-to-an-existing-payment-group/add-fees-to-an-existing-payment-group.event-handler.test.ts
```diff
@@ -0,0 +1,176 @@
...
+        });
+      });
+    },
+  );
+
+  it.each([
+    { feeRecordsAndPaymentsMatch: true, testNameSuffix: 'the fee records match the payments' },
+    { feeRecordsAndPaymentsMatch: false, testNameSuffix: 'the fee records do not match the payments' },
+  ])(
+    "calls the fee record state machine event handlers for existing fee records with 'feeRecordsAndPaymentsMatch' set to '$feeRecordsAndPaymentsMatch' if $testNameSuffix after the new fee records have been added",
```
RLavender98, line 139: Please can we add the event name in the test name?

trade-finance-manager-ui/server/controllers/utilisation-reports/helpers/payment-form-helpers.test.ts
```diff
@@ -295,4 +297,69 @@ describe('payment-form-helpers', () => {
...
+        errorSummary: [{ text: 'Some text', href: '#some-href' }],
+      };
+      jest.mocked(validateAddToAnExistingPaymentRequestFormValues).mockReturnValue(paymentErrors);
+
+      // Act
+      const result = extractAddToAnExistingPaymentRadioPaymentIdsAndValidateIfPresent(requestBody);
+
+      // Assert
+      expect(result.paymentIds).toEqual([]);
+      expect(result.errors).toEqual(paymentErrors);
```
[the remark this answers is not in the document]
sw-mattw: Pulled this out to a separate test

trade-finance-manager-ui/server/controllers/utilisation-reports/helpers/payment-form-helpers.test.ts
```diff
@@ -295,4 +297,69 @@ describe('payment-form-helpers', () => {
...
+      const requestBody = {};
+
+      // Act
+      const result = extractAddToAnExistingPaymentRadioPaymentIdsAndValidateIfPresent(requestBody);
+
+      // Assert
+      expect(result.isAddingToAnExistingPayment).toBe(false);
+      expect(result.errors).toEqual(EMPTY_PAYMENT_ERRORS_VIEW_MODEL);
+      expect(result.paymentIds).toEqual([]);
+    });
```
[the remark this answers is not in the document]
sw-mattw: Pulled each of these out into separate tests, within a shared context

e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/add-to-an-existing-payment.spec.js
sw-mattw: Refactored this into the two contexts with a helper function to perform the actions that are shared.

libs/common/src/sql-db-entities/payment/payment.entity.test.ts
sw-mattw: Reintroduced test (other comment), made the mock a separate object with the same ID and improved naming of these mocks.

dtfs-central-api/src/v1/controllers/utilisation-report-service/post-add-fees-to-an-existing-payment-group.controller/index.ts
```diff
@@ -27,25 +27,27 @@ export type PostAddFeesToAnExistingPaymentGroupRequest = CustomExpressRequest<{
...
+    validateThatRequestedPaymentsMatchSavedPayments(payments, requestedPaymentIds);
 
-    const existingPaymentGroupFeeRecords = payments.at(0)?.feeRecords;
+    const existingPaymentGroupFeeRecords = payments[0].feeRecords;
     validateThatPaymentGroupHasFeeRecords(existingPaymentGroupFeeRecords);
 
     const selectedFeeRecords = await FeeRecordRepo.findByIdAndReportIdWithReport(feeRecordIds, Number(reportId));
+
+    const paymentGroupCurrency = payments[0].currency;
+    validateThatAllSelectedFeeRecordsAndPaymentGroupHaveSameCurrency(selectedFeeRecords, paymentGroupCurrency);
```
RLavender98, line 46: Do we have either a unit or api test covering this?

e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/add-to-an-existing-payment.spec.js
```diff
@@ -47,40 +39,23 @@ context(`${PDC_TEAMS.PDC_RECONCILE} users can add fee records to existing paymen
...
   });
 
-  beforeEach(() => {
-    cy.task(NODE_TASKS.REMOVE_ALL_UTILISATION_REPORTS_FROM_DB);
-    cy.task(NODE_TASKS.REMOVE_ALL_PAYMENTS_FROM_DB);
-    cy.task(NODE_TASKS.INSERT_UTILISATION_REPORTS_INTO_DB, [report]);
-
-    cy.task(NODE_TASKS.INSERT_FEE_RECORDS_INTO_DB, [feeRecordOne, feeRecordTwo, feeRecordThree]);
-
+  const navigateToAddToExistingPaymentScreenGivenSelectedFeeRecord = () => {
```
RLavender98, line 58: I'm not sure I follow the name of this since we select the fee record inside this helper don't we? Should it be: `navigateToAddToExistingPaymentScreenForFirstFeeRecord`?

RLavender98 (commented):
Just two final small comments!

dtfs-central-api/src/v1/controllers/utilisation-report-service/post-add-fees-to-an-existing-payment-group.controller/index.ts
sw-mattw: Added a unit test for this case, thanks

dtfs-central-api/src/v1/controllers/utilisation-report-service/post-add-fees-to-an-existing-payment-group.controller/index.test.ts
```diff
@@ -158,6 +158,36 @@ describe('post-fees-to-an-existing-payment-group.controller', () => {
...
 
+    it("responds with a '400' if the fee records payment currency does not equal the payment group currency", async () => {
+      // Arrange
+      const paymentId = 3;
+      const aPaymentInUSD = PaymentEntityMockBuilder.forCurrency('USD').withId(paymentId).withFeeRecords([]).build();
+      const aFeeRecordWithAPaymentInUSD = FeeRecordEntityMockBuilder.forReport(utilisationReport)
+        .withId(2)
+        .withPaymentCurrency('GBP')
+        .withPayments([aPaymentInUSD])
+        .build();
```
RLavender98, line 169: A few things:
- We need to set the status of this fee record in this test since there is validation in the endpoint of the fee record status to be in TO_DO status and we don't want this test to fail for that reason
- A fee record in TO_DO status won't have any payments attached, so please can we removed them and rename this to `aFeeRecordWithPaymentCurrencyGbp`
- The payment also needs a fee record to be allowed to have another fee record added so separately we need `aFeeRecordWithPaymentCurrencyUsd` which is attached to the payment and has a different id to the GBP one

RLavender98 (dismissed):

dtfs-central-api/src/repositories/payment-repo/payment.repo.ts
```diff
@@ -43,4 +44,34 @@ export const PaymentRepo = SqlDbDataSource.getRepository(PaymentEntity).extend({
...
+
+  /**
+   * Finds payment entities that are part of the same group as the payment
+   * with the supplied id.
+   * @param paymentId - The payment id
+   * @param reportId - The report id
+   * @returns An array of payment entities in the same group as the specified
+   * payment, including their fee records and associated reports.
+   */
+  async findPaymentsInGroupContainingPaymentByIdAndReportIdWithFeeRecords(paymentId: number, reportId: number): Promise<PaymentEntity[]> {
```
ttbarnes, line 56: Is there a simpler name we could use for this? It's very long and from reading it, it's not clear what it's doing, e.g it looks like:

- It might be getting "payments in a group that contains a payment *and* the group contains a report ID"
- Or it might be getting "payments in a group that contains a payment and the payment has a report ID"

Maybe it could be something like `findPaymentsByIds`?

Will there be a scenario where we want to get payments _without_ fee records?

sw-mattw: The two scenarios will always return the same thing as grouped payments will always be for the same report.

There are scenarios where we want to fetch payments without fee records, we have the naming convention on our SQL repos that when we include relations we include this as the 'with' part of the name. 

Given this, do you think `findPaymentsInGroupByPaymentIdAndReportIdWithFeeRecords` would be a better name that still fits within these naming conventions?

ttbarnes: Not sure what to suggest especially if this is the current convention. Long names like this are very hard to read/follow IMO. Appreciate this is not the scope of this PR, maybe object structured params could be nicer e.g

```ts
findPaymentsById({
  paymentId,
  withFeeRecords = false,
  reportId,
});
```

Just an idea off the top of my head, i'm not familiar with the product/service or usage of such functions.

sw-mattw: I've changed it to the shorter name above but kept the 'with' part of the method name as the relations we include with this particular repo method are a bit more complex than the other repo methods

RLavender98 (approved):

PR #3390, opened 2024-08-15
fix(DTFS2-6713): estore buyer, deal, facilities and documents upload

abhi-markan:
## Introduction :pencil2:
This PR mitigates issue with CRON execution pertinent `buyer`, `deal`, `facility` and `documents` upload, whilst also ensure strong TS static typing, improved documentation and enhanced logging.

* Deal CRON execution script.
* Buyer CRON execution script.
* Facility CRON execution script.
* Documents upload CRON execution script.
* Introduced `repositories` to `external-api` for `cron-job-logs`, this is a minimal implementation a typical implementation will require a complete `cron-job-log` interface definition with relevant types, which is outside this PR scope.

* Upgraded Husky to `9.1.4` and Husky GitHub hooks.
* Improved documentation.
* Consistent file naming conventions.
* Updated `publish.yml` to specify `release-type` argument.
* Updated `deployment.yml` to allow `MSSQL` data migration script execution @ChristianMcc 
* Have introduced `.markdownlint.json` to ensure consistent MD syntax linting and 80 characters limit to comply with `MD013` rule.
* Have enabled conditional replica set execution during `MongoDB` build process using `entrypoint.sh`, this approach allows for a singular `docker-compose.yml` file with build process configuration.
* Introduced `unit-test-ff` to execute FF specific unit test cases with FF environment variables set to truthy explicitly.
* Renamed screaming sentence `SESSION` and `TOKEN` to lowercase.


external-api/src/cron/eStoreBuyerDirectoryCreationJob.cron.ts
```diff
@@ -0,0 +1,91 @@
+import { ObjectId } from 'mongodb';
+import { HttpStatusCode } from 'axios';
+import { getCollection } from '../database';
+import { BuyerFolderResponse, EstoreErrorResponse, Estore } from '../interfaces';
+import { ESTORE_CRON_STATUS } from '../constants';
+import { createBuyerFolder } from '../v1/controllers/estore/eStoreApi';
+import { getNowAsEpoch } from '../helpers/date';
+import { eStoreDealDirectoryCreationJob } from './eStoreDealDirectoryCreationJob.cron';
+
+const acceptableStatuses = [HttpStatusCode.Ok, HttpStatusCode.Created];
```
ttbarnes, line 10: Maybe capitalize? `ACCEPTABLE_STATUSES`

external-api/src/cron/eStoreBuyerDirectoryCreationJob.cron.ts
```diff
@@ -0,0 +1,91 @@
...
+  if (eStoreData?.buyerName) {
+    const { dealId, siteId, exporterName, buyerName, dealIdentifier } = eStoreData;
+
+    console.info('Creating buyer directory %s for deal %s', buyerName, dealIdentifier);
+
+    // Create the buyer directory
+    const response: BuyerFolderResponse | EstoreErrorResponse = await createBuyerFolder(siteId, {
+      exporterName,
+      buyerName,
+    });
```
ttbarnes, line 52: Do we need to check for `siteId` and `exporterName`, as well as `buyerName` (line 43), before logging and calling `createBuyerFolder`?

external-api/src/cron/eStoreBuyerDirectoryCreationJob.cron.ts
```diff
@@ -0,0 +1,91 @@
...
+ *   .catch((error) => console.error('Job failed', error));
+ */
+export const eStoreBuyerDirectoryCreationJob = async (eStoreData: Estore): Promise<void> => {
+  const cronJobLogs = await getCollection('cron-job-logs');
+
+  // 1. Creating buyer directory
+  if (eStoreData?.buyerName) {
+    const { dealId, siteId, exporterName, buyerName, dealIdentifier } = eStoreData;
+
+    console.info('Creating buyer directory %s for deal %s', buyerName, dealIdentifier);
```
ttbarnes, line 46:
```suggestion
    console.info('Attempting to create buyer directory %s for deal %s', buyerName, dealIdentifier);
```

external-api/src/cron/eStoreBuyerDirectoryCreationJob.cron.ts
```diff
@@ -0,0 +1,91 @@
...
+        {
+          $set: {
+            'cron.buyer': {
+              response,
+              status: ESTORE_CRON_STATUS.FAILED,
+              timestamp: getNowAsEpoch(),
+            },
+          },
+        },
+      );
```
ttbarnes, line 88: One we could look into separately - I can see a lot of calls to `cronJobLogs.updateOne`, could potentially create a new helper to be DRY and reduce LOC

Not sure if it's worth looking into or doing, without knowing more about the differences. Also probably other things to do that have more value.

external-api/src/cron/eStoreDealDirectoryCreationJob.cron.ts
```diff
@@ -0,0 +1,97 @@
...
+ *   dealId: '507f1f77bcf86cd799439011',
+ *   destinationMarket: 'UK',
+ *   riskMarket: 'High'
+ * };
+ *
+ * eStoreDealDirectoryCreationJob(eStoreData)
+ *   .then(() => console.log('Job completed successfully'))
+ *   .catch((error) => console.error('Job failed', error));
+ */
+export const eStoreDealDirectoryCreationJob = async (eStoreData: Estore): Promise<void> => {
```
ttbarnes, line 42: For these CRON functions, should we wrap the functions in try/catch handlers and a `console.info` at the top?

external-api/src/cron/eStoreFacilityDirectoryCreationJob.cron.ts
```diff
@@ -0,0 +1,96 @@
...
+        createFacilityFolder(siteId, dealIdentifier, {
+          facilityIdentifier,
+          buyerName,
+          exporterName,
+        }),
+      ),
+    );
+
+    // Validate each and every response status code
+    if (response.every((facility) => acceptableStatuses.includes(facility?.status))) {
```
ttbarnes, line 61: Is any `status` property a sucessful response? Or do we need to be more specific e.g `facility?.status === A_STATUS`

external-api/src/cron/eStoreSiteCreationJob.cron.ts
```diff
@@ -79,17 +100,21 @@ export const eStoreSiteCreationCron = async (eStoreData: Estore): Promise<void>
     );
   } else {
     // Step 3: Site creation has failed
+
+    // Stop CRON job
+    cron(eStoreData, ENDPOINT.SITE, true);
```
ttbarnes, line 105: Probably not for this PR, but I wasn't sure what `true` meant here - this could be updated to have object structured params so it's easier to digest (IMO) e.g

```ts
cron({
  eStoreData,
  category: ENDPOINT.SITE,
  kill: true,
});
```

external-api/src/cron/eStoreBuyerDirectoryCreationJob.cron.ts
```diff
@@ -0,0 +1,91 @@
...
+ * This job performs the following tasks:
+ * 1. Checks if a buyer name is provided in the eStore data.
+ * 2. Creates a buyer directory using the provided buyer name and site ID.
+ * 3. Updates the cron job logs based on the success or failure of the directory creation.
+ *
+ * @param {Estore} eStoreData - The eStore data containing information about the deal, exporter, and buyer.
+ *
+ * @returns {Promise<void>} - A promise that resolves when the job is complete.
+ *
+ * @throws {Error} - Throws an error if there is an issue with database operations or API calls.
```
JoshBinns2000, line 24: We should avoid including the type in both the docstring and the function signature when in a typescript file

external-api/src/cron/eStoreBuyerDirectoryCreationJob.cron.ts
JoshBinns2000: The normal way of doing this would be to have a repo file for this collection which handles database-related logic - see `dtfs-central-api/src/repositories` for a few examples. However, creating a repo would probably need a refactor across `external-api` to ensure that every call made against the `cron-job-logs` collection uses this repo to introduce it as a consistent pattern

ttbarnes: Nice, didn't realise we have such a service (`repositories`). Makes sense

.github/workflows/test.yml
```diff
@@ -496,68 +487,6 @@ jobs:
           TZ: ${{ vars.TIMEZONE }}
         run: npx cypress run --config video=false,screenshotOnRunFailure=false --project ./gef --record false --spec "./gef/cypress/e2e/${{ matrix.spec }}"
 
-
-  # 5.1 E2E - GEF backwards compatibility
-  e2e-tests-gef-feature-flag-backwards-compatibility:
```
ttbarnes, line 501: Are these backwards compatible E2E tests now irrelevant?

.github/workflows/test.yml
```diff
@@ -765,9 +692,18 @@ jobs:
...
+          echo "::set-output name=directory::$directory"
+
       - name: Execute
         working-directory: ./e2e-tests
         env:
           SQL_DB_HOST: ${{ secrets.SQL_DB_HOST_LOCAL }}
           TZ: ${{ vars.TIMEZONE }}
-        run: npx cypress run --config video=false,screenshotOnRunFailure=false --project ./ukef --record false --spec "./ukef/cypress/e2e/journeys/portal/${{ matrix.spec }}"
+          # Various feature flags
+          FF_TFM_FACILITY_END_DATE_ENABLED: 'false'
```
ttbarnes, line 707: I'm not up to speed with the feature flags in DTFS - since this flow is called "E2E Feature flag", should this flag be enabled?

Is there more 1x feature flag? If so i'd expect them all to be listed and `true` here

external-api/src/cron/eStoreBuyerDirectoryCreationJob.cron.ts
```diff
@@ -0,0 +1,82 @@
...
+ *   .catch((error) => console.error('Job failed', error));
+ */
+export const eStoreBuyerDirectoryCreationJob = async (eStoreData: Estore): Promise<void> => {
+  // Argument validation
+  if (!eStoreData?.siteId || !eStoreData?.buyerName || !eStoreData?.exporterName || !eStoreData.dealIdentifier) {
+    console.error('Invalid arguments provided for eStore buyer directory creation');
+    return;
+  }
+
+  const { dealId, siteId, exporterName, buyerName, dealIdentifier } = eStoreData;
```
ttbarnes, line 46: Should we add a check for `dealId` in the above argument validation? It's currently the only param that's not checked

external-api/src/cron/eStoreBuyerDirectoryCreationJob.cron.ts
```diff
@@ -0,0 +1,82 @@
...
+
+  console.info('Attempting to create a buyer directory %s for deal %s', buyerName, dealIdentifier);
```
ttbarnes, line 48: I forget, what is `dealIdentifier`? Can't seem to find a decent mock of this.

I'm wondering, if it's different to `dealId` or makes it more troublesome to identify the deal ID (assuming this would be required when debugging/troubleshooting), maybe this should use `dealId` instead or in conjuction with?

external-api/src/cron/eStoreDealDirectoryCreationJob.cron.ts
```diff
@@ -0,0 +1,96 @@
...
+    !eStoreData?.siteId ||
+    !eStoreData?.exporterName ||
+    !eStoreData?.buyerName ||
+    !eStoreData?.dealIdentifier ||
+    !eStoreData?.destinationMarket ||
+    !eStoreData?.riskMarket
+  ) {
+    console.error('Invalid arguments provided for eStore deal directory creation');
+    return;
+  }
```
ttbarnes, line 55: Might be overkill - maybe could create a "Estore data check" helper that can handle generic and job specific field checks, thinking something like this (quick example), could save ~10 lines in each job and simplify the unit tests.


```ts
// with bespokeFieldNames being bespoke fields specific to each job
const validateEstoreData = (
  estoreData: Estore,
  cronJobName: string,
  bespokeFieldNames?: Array<string>,
) =>  {
  const allFieldNames = [
    'dealId'
    'siteId',
    'exporterName',
    'buyerName',
    'dealIdentifier',
    ...bespokeFieldNames,
  ];

  bespokeFieldNames.forEach((str) => {
    if (!estoreData[str]) {
      console.error(`Invalid generic arguments provided for eStore ${cronJobName}`); 
    }
  })
};
```

external-api/src/cron/eStoreFacilityDirectoryCreationJob.cron.ts
```diff
@@ -0,0 +1,94 @@
...
+      }),
+    ),
+  );
+
+  // Validate each and every response status code
+  if (response.every((facility) => ACCEPTABLE_STATUSES.includes(facility?.status))) {
+    console.info('Facility %s directory has been created for deal %s', facilityIdentifiers, dealIdentifier);
+
+    // Step 2: Update `cron-job-logs`
+    await update(new ObjectId(dealId), {
```
ttbarnes, line 74: Maybe one for the future, Is it worth looking into doing a bulk update? So we map/filter valid facilities/statuses and from that, generate an array and make 1x single bulk update DB call?

external-api/src/helpers/cron.ts
```diff
@@ -52,15 +76,15 @@ export const cron = (eStoreData: Estore, category: Category, kill: boolean = fal
...
 
     return false;
   }
 
   const cronJob = new CronJob(
     String(ESTORE_CRON_MANAGER_SCHEDULE), // Cron schedule
-    () => onTick(eStoreData), // On Tick
-    () => onComplete(eStoreData, category), // On complete
+    () => onTick(data), // On Tick
+    () => onComplete(data, category), // On complete
```
ttbarnes, line 87: I don't think we need these comments, function describes what it does 😄

external-api/src/interfaces/eStore.interface.ts
```diff
@@ -182,21 +195,25 @@ export interface DealFolderResponse {
...
  * @property {string} data.foldername - Name of the facility folder.
  * @property {string} data.error - Error message, if any.
+ * @property {unknown} data.message - Additional message, if any.
  */
 export interface FacilityFolderResponse {
   readonly status: number;
   readonly data: {
     foldername?: string;
     error?: string;
+    message?: unknown;
```
ttbarnes, line 210: Why is this unknown, could the response be anything or maybe 1 of a 2/3 possible responses?

external-api/src/repositories/estore/estore-repo.ts
```diff
@@ -0,0 +1,9 @@
+import { ObjectId, UpdateFilter, WithoutId } from 'mongodb';
+import { MONGO_DB_COLLECTIONS } from '@ukef/dtfs2-common';
+import { getCollection } from '../../database';
+
+export const update = async (dealId: ObjectId, document: UpdateFilter<WithoutId<object>>) => {
```
ttbarnes, line 5: Documentation?

external-api/src/cron/eStoreBuyerDirectoryCreationJob.cron.ts
ttbarnes: If I remember correctly, I now think `dealIdentifier` is the UKEF number generator ID, in which case this LGTM

.github/workflows/test.yml
abhi-markan: I amalgamated FF E2E under one GHA Job.

.github/workflows/test.yml
abhi-markan: Good point, @MarRobSoftwire I have refactored the `test.yml` to keep all FF E2E tests together I copied `FF_TFM_FACILITY_END_DATE_ENABLED` and `GEF_DEAL_VERSION` env variables and they both are falsey, should they be truthy for FF E2E to be tested effectively?

external-api/src/repositories/estore/estore-repo.ts
```diff
@@ -0,0 +1,86 @@
...
+ * @throws {Error} - Throws an error if the find operation fails.
+ *
+ * @example
+ * const dealId = new ObjectId('507f1f77bcf86cd799439011');
+ *
+ * find(dealId)
+ *   .then((document) => console.log('Document found', document))
+ *   .catch((error) => console.error('Find operation failed', error));
+ */
+export const find = async (dealId: ObjectId): Promise<WithId<ObjectId> | WithoutId<unknown> | null> => {
```
JoshBinns2000, line 21: Could this be `findByDealId` to make it specific? I'd expect a standard `find` to accept the same input parameters as the `mongodb` `Collection.find` method, so would be good to make it specific

external-api/src/repositories/estore/estore-repo.ts
```diff
@@ -0,0 +1,86 @@
...
+import { getCollection } from '../../database';
+
+/**
+ * Finds a document in the CRON_JOB_LOGS collection based on the provided deal ID.
+ *
+ * @param {ObjectId} dealId - The unique identifier for the deal.
+ *
+ * @returns {Promise<WithId<ObjectId> | WithoutId<unknown> | null>} - A promise that resolves to the found document, or `null` if no document is found.
+ *
+ * @throws {Error} - Throws an error if the find operation fails.
```
JoshBinns2000, line 12: Based on the below, the function does not throw an error if the `find` operation fails. Instead, it catches the error, logs and returns `null`

external-api/src/repositories/estore/estore-repo.ts
JoshBinns2000, line 21: `WithId<ObjectId>` is not the type you're expecting I think. `WithId` itself produces a type which contains the `_id: ObjectId` field. `ObjectId` itself has it's own type definition, however, so the type you've written is actually `ObjectId & { _id: ObjectId }` which I don't think you want.

Similarly, if `find` returns something, it will always contain the `_id: ObjectId` property by the nature of mongo, so the `WithoutId<unknown>` return type isn't quite right

external-api/src/repositories/estore/estore-repo.ts
```diff
@@ -0,0 +1,86 @@
...
+};
+
+/**
+ * Inserts a document into the CRON_JOB_LOGS collection.
+ *
+ * @param {object} document - The document to be inserted.
+ *
+ * @returns {Promise<InsertOneResult | boolean>} - A promise that resolves to the result of the insert operation, or `false` if an error occurs.
+ *
+ * @throws {Error} - Throws an error if the insert operation fails.
```
JoshBinns2000, line 39: Same comment as above with respect to throwing the error

external-api/src/repositories/estore/estore-repo.ts
```diff
@@ -0,0 +1,86 @@
...
+
+/**
+ * Updates a document in the CRON_JOB_LOGS collection based on the provided deal ID.
+ *
+ * @param {ObjectId} dealId - The unique identifier for the deal.
+ * @param {UpdateFilter<WithoutId<object>>} document - The update filter containing the fields to be updated.
+ *
+ * @returns {Promise<UpdateResult | boolean>} - A promise that resolves to the result of the update operation, or `false` if an error occurs.
+ *
+ * @throws {Error} - Throws an error if the update operation fails.
```
JoshBinns2000, line 67: Same comment as above

external-api/src/repositories/estore/estore-repo.ts
```diff
@@ -0,0 +1,86 @@
...
+ *
+ * @example
+ * const dealId = new ObjectId('507f1f77bcf86cd799439011');
+ * const document = { status: 'updated' };
+ *
+ * update(dealId, document)
+ *   .then((result) => console.log('Update successful', result))
+ *   .catch((error) => console.error('Update failed', error));
+ */
+export const update = async (dealId: ObjectId, document: UpdateFilter<WithoutId<object>>): Promise<UpdateResult | boolean> => {
```
JoshBinns2000, line 77: As above, could this be `updateByDealId`?

JoshBinns2000, line 77: This comment also applies to the `find` operation

If we are passing the supplied `dealId` to the `new ObjectId` constructor, then I think we should set the type of `dealId` to be `ObjectId | string`. However, if the decision to force the input to be `ObjectId` is intentional, then constructing a new `ObjectId` instance is unnecessary

JoshBinns2000 (changes requested):
Only looked at the repo file as that was the one I suggested. I would prefer that we use the `class` implementation like we have in `tfm-facilities-repo` so that you get the context of which repo you're using when you call it. For example, I find it easier to read

```typescript
import { EstoreRepo } from '../repositories/estore-repo.ts';

const document = await EstoreRepo.findByDealId(...);
```

than

```typescript
import { findByDealId } from '../repositories/estore-repo.ts';

const document = await findByDealId(...);
```

In the second version, especially in a larger file, it's difficult to figure out what we're finding by deal id, whereas the first version makes it clear that we are finding something from the `Estore` collection.

As a second note, I think this repo is based around the `CRON_JOB_LOGS` collection, so it should be called `cron-job-logs-repo` and not `estore-repo`

external-api/src/repositories/estore/estore-repo.ts
abhi-markan: Have updated to `Promise<WithId<object> | null>`

external-api/src/repositories/estore/estore-repo.ts
abhi-markan: Good spot, yes `ObjectId` type is intentional and is defined in `Estore` interface, have updated.

external-api/src/repositories/estore/estore-repo.ts
```diff
@@ -0,0 +1,80 @@
+import { InsertOneResult, ObjectId, UpdateFilter, UpdateResult, WithId, WithoutId } from 'mongodb';
+import { MONGO_DB_COLLECTIONS } from '@ukef/dtfs2-common';
+import { getCollection } from '../../database';
+
+/**
+ * Finds a document in the CRON_JOB_LOGS collection based on the provided deal ID.
+ *
+ * @param {ObjectId} dealId - The unique identifier for the deal.
```
JoshBinns2000, line 8: Sorry I didn't pick up on this last time, but we shouldn't have types in both the TSDoc and the function signature

external-api/src/repositories/estore/estore-repo.ts
```diff
@@ -0,0 +1,80 @@
...
+ *
+ * @returns {Promise<WithId<ObjectId> | WithoutId<unknown> | null>} - A promise that resolves to the found document, or `null` if no document is found.
+ *
+ * @example
+ * const dealId = new ObjectId('507f1f77bcf86cd799439011');
+ *
+ * find(dealId)
```
JoshBinns2000, line 15:
```suggestion
 * findByDealId(dealId)
```

external-api/src/repositories/estore/estore-repo.ts
```diff
@@ -0,0 +1,80 @@
...
+ * @returns {Promise<InsertOneResult | boolean>} - A promise that resolves to the result of the insert operation, or `false` if an error occurs.
+ *
+ * @example
+ * const document = { dealId: '507f1f77bcf86cd799439011', status: 'new' };
+ *
+ * insert(document)
+ *   .then((result) => console.log('Insert successful', result))
+ *   .catch((error) => console.error('Insert failed', error));
+ */
+export const insertOne = async (document: object): Promise<InsertOneResult | boolean> => {
```
JoshBinns2000, line 44: I think the type of `document` here can be the `Document` type (imported from `mongodb`). The result type can also be narrowed to be `Promise<InsertOneResult | false>`

external-api/src/repositories/estore/estore-repo.ts
```diff
@@ -0,0 +1,80 @@
...
+ * Inserts a document into the CRON_JOB_LOGS collection.
+ *
+ * @param {object} document - The document to be inserted.
+ *
+ * @returns {Promise<InsertOneResult | boolean>} - A promise that resolves to the result of the insert operation, or `false` if an error occurs.
+ *
+ * @example
+ * const document = { dealId: '507f1f77bcf86cd799439011', status: 'new' };
+ *
+ * insert(document)
```
JoshBinns2000, line 40:
```suggestion
 * insertOne(document)
```

external-api/src/repositories/estore/estore-repo.ts
```diff
@@ -0,0 +1,80 @@
...
+ * @param {ObjectId} dealId - The unique identifier for the deal.
+ * @param {UpdateFilter<WithoutId<object>>} document - The update filter containing the fields to be updated.
+ *
+ * @returns {Promise<UpdateResult | boolean>} - A promise that resolves to the result of the update operation, or `false` if an error occurs.
+ *
+ * @example
+ * const dealId = new ObjectId('507f1f77bcf86cd799439011');
+ * const document = { status: 'updated' };
+ *
+ * update(dealId, document)
```
JoshBinns2000, line 67:
```suggestion
 * updateByDealId(dealId, document)
```

external-api/src/repositories/estore/estore-repo.ts
```diff
@@ -0,0 +1,80 @@
...
+ *   .then((result) => console.log('Update successful', result))
+ *   .catch((error) => console.error('Update failed', error));
+ */
+export const updateByDealId = async (dealId: ObjectId, document: UpdateFilter<WithoutId<object>>): Promise<UpdateResult | boolean> => {
```
JoshBinns2000, line 71: Should the type of `document` be `UpdateFiler` or `Document`?

Return type can be narrowed again to `Promise<UpdateResult | boolean>`

JoshBinns2000 (changes requested):

external-api/src/repositories/estore/estore-repo.ts
abhi-markan: I replied back to your previous comment regarding the same, we are still using types in arguments (JSDocs - https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#param-and-returns) across the repo I would unify this in an epic.

Signature is always a good to specify explicitly, albeit not mandatory.

external-api/src/repositories/estore/estore-repo.ts
abhi-markan: Updated.

external-api/src/repositories/estore/estore-repo.ts
abhi-markan: Updated.

external-api/src/repositories/estore/estore-repo.ts
abhi-markan: Updated.

external-api/src/repositories/estore/estore-repo.ts
abhi-markan: Have updated to `Document`.

external-api/src/repositories/estore/estore-repo.ts
JoshBinns2000: I think I'm going to push back on this a bit simply citing something that has already happened below in this PR - you updated the return type of this function but didn't have to update the return type in the jsdoc and they're now out of sync.

The link you provided doesn't actually say anything about providing both jsdoc types and typescript types - it is discussing the fact that you _can_ use the typescript type checker with jsdoc inside of js files, which is useful if you set the tsconfig `checkJs` option to `true`. However, minimising confusion by having one source of truth with respect to types is quite important in my opinion, and avoiding any risk of confusing the LSP by having two sources for type definitions is also important for the developer experience

external-api/src/cron/eStoreDealDirectoryCreationJob.cron.ts
ttbarnes: Sounds great and makes sense, don't want to make this PR even larger.

I would suggest some kind of object structured function export for such a `validateEstoreData` helper, because then in the individual unit tests, you can simply just call `toHaveBeenCalledWith()` 👍

external-api/src/repositories/estore/estore-repo.ts
JoshBinns2000: I didn't see the reply to the previous comment. We are likely to have a code base which forever uses a mix of JSDoc and TSDoc, so using TSDoc in typescript files and JSDoc in javascript files seems like the standard we should stick to for future code

ttbarnes (dismissed):
LGTM great work

external-api/src/repositories/estore/estore-repo.ts
abhi-markan: Previous response - https://github.com/UK-Export-Finance/dtfs2/pull/3390/files/2f0284dbe0e71d65cdf67c321a8d47d23f42cd3f#r1722864347

* I have updated the return signature which I should have updated, however rightly so this has been picked up in a review.
* Majority of the TS files still uses JSDocs and until a convention is decided to migrate all TS files to use TSDocs.

JoshBinns2000: Form a quick search it appears that this is not the norm in most typescript files and is instead simply a common occurrence in `external-api` typescript files. I would prioritise fixing this issue in `external-api` sooner rather than later as continuing to add JSDoc types in typescript files will just add to the pile of tech debt and discourage performing this refactor in future. If there is a ticket number related to updating this, can we please add it as a TODO and then I'll be happy to resolve this comment thread

ttbarnes (approved):

external-api/src/repositories/estore/estore-repo.ts
abhi-markan: I am happy add an existing ticket number to the `README.md` rather than doc string as this affects all the micro-services including existing PRs.

JoshBinns2000: If you can please add this as a TODO in the `README` then I'm happy to approve

abhi-markan: Thanks @JoshBinns2000  have added to a new `docs/docstring.md` file.

ttbarnes (approved):

JoshBinns2000 (dismissed):

ttbarnes (dismissed):

mongodb/Dockerfile
```diff
@@ -1,10 +1,15 @@
 # Base image
 FROM mongo:5.0.5
 
-# Create keyfile
-RUN openssl rand -base64 756 > /mongo-keyfile
+# Enviornment variable
```
abhi-markan, line 4: @ChristianMcc @ttbarnes I have updated `MongoDB` dockerfile to ensure Docker build process can be customised as per `ENABLE_REPLICA_SET` environment variable more information on the PR description. A quick review from you both will be great.

mongodb/entrypoint.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+# Start MongoDB with replica set if `ENABLE_REPLICA_SET` is true
+if [ "$ENABLE_REPLICA_SET" == "true" ]; then
+  mongod --bind_ip_all --keyFile /mongo-keyfile --replSet rs0
+else
+  mongod --bind_ip_all
```
ChristianMcc, line 7: Is this needed if there is no replica set?

ChristianMcc:
Note I can't see the changes mentioned under `Updated deployment.yml to allow MSSQL data migration script execution @ChristianMcc` in the PR description, was that done as part of a different PR now?

external-api/src/cron/eStoreBuyerDirectoryCreationJob.cron.ts
```diff
@@ -0,0 +1,80 @@
...
+
+  // Step 1: Create the buyer directory
+  const response: BuyerFolderResponse | EstoreErrorResponse = await createBuyerFolder(siteId, {
```
Zainzzkk, line 49: just wondering, for readability, is it better for response to be `let response:  BuyerFolderResponse | EstoreErrorResponse` and then response = await ...

Zainzzkk, line 49: try catch?

external-api/src/cron/eStoreDealDirectoryCreationJob.cron.ts
```diff
@@ -0,0 +1,94 @@
...
+ * };
+ *
+ * eStoreDealDirectoryCreationJob(eStoreData)
+ *   .then(() => console.log('Job completed successfully'))
+ *   .catch((error) => console.error('Job failed', error));
+ */
+export const eStoreDealDirectoryCreationJob = async (eStoreData: Estore): Promise<void> => {
+  // Argument validation
+  if (
+    !eStoreData?.dealId ||
```
Zainzzkk, line 43: could this be a const? then `if (const)`

external-api/src/cron/eStoreDealDirectoryCreationJob.cron.ts
```diff
@@ -0,0 +1,94 @@
...
+    console.error('Invalid arguments provided for eStore deal directory creation');
+    return;
+  }
+
+  const { dealId, siteId, exporterName, buyerName, dealIdentifier, destinationMarket, riskMarket } = eStoreData;
+
+  console.info('Attempting to create a deal directory for deal %s', dealIdentifier);
+
+  // Step 1: Create the deal directory
+  const response: DealFolderResponse | EstoreErrorResponse = await createDealFolder(siteId, {
```
Zainzzkk, line 60: try catch?

external-api/src/cron/eStoreFacilityDirectoryCreationJob.cron.ts
```diff
@@ -0,0 +1,92 @@
...
+ * };
+ *
+ * eStoreFacilityDirectoryCreationJob(eStoreData)
+ *   .then(() => console.log('Job completed successfully'))
+ *   .catch((error) => console.error('Job failed', error));
+ */
+export const eStoreFacilityDirectoryCreationJob = async (eStoreData: Estore): Promise<void> => {
+  // Argument validation
+  if (
+    !eStoreData?.dealId ||
```
Zainzzkk, line 41: same as above

external-api/src/cron/eStoreFacilityDirectoryCreationJob.cron.ts
```diff
@@ -0,0 +1,92 @@
...
+    console.error('Invalid arguments provided for eStore facility directory creation');
+    return;
+  }
+
+  const { dealId, siteId, facilityIdentifiers, exporterName, buyerName, dealIdentifier } = eStoreData;
+
+  console.info('Attempting to create a facility directory %s for deal %s', facilityIdentifiers, dealIdentifier);
+
+  // Step 1: Create the facility directory
+  const response: FacilityFolderResponse[] | EstoreErrorResponse[] = await Promise.all(
```
Zainzzkk, line 57: same as above

external-api/src/cron/eStoreFacilityDirectoryCreationJob.cron.ts
```diff
@@ -0,0 +1,92 @@
...
+      createFacilityFolder(siteId, dealIdentifier, {
+        facilityIdentifier,
+        buyerName,
+        exporterName,
+      }),
+    ),
+  );
+
+  // Validate each and every response status code
+  if (response.every((facility) => ACCEPTABLE_STATUSES.includes(facility?.status))) {
```
Zainzzkk, line 68: const for this?

external-api/src/cron/eStoreSiteCreationJob.cron.ts
```diff
@@ -9,97 +10,121 @@ import { siteExists } from '../v1/controllers/estore/eStoreApi';
...
   // Step 1: Initiate the CRON job
-  cron(eStoreData, ENDPOINT.SITE);
+  cron({
+    data: eStoreData,
+    category: ENDPOINT.SITE,
+  });
 
   // Step 2: Site existence check
-  const siteExistsResponse: SiteExistsResponse | EstoreErrorResponse = await siteExists(eStoreData.exporterName);
+  const siteExistsResponse: SiteExistsResponse | EstoreErrorResponse = await siteExists(exporterName);
```
Zainzzkk, line 57: or instead of let, could this be `const siteExistsResponse = await siteExists(exporterName) as SiteExistsResponse | EstoreErrorResponse ;`

external-api/src/cron/eStoreSiteCreationJob.cron.ts
Zainzzkk, line 1: do we need a try catch here?

azure-functions/acbs-function/unit.ff.jest.config.js
```diff
@@ -0,0 +1,8 @@
+const commonSettings = require('./jest.common.config');
```
Zainzzkk, line 1: can we move towards using import and export? ES6

dtfs-central-api/unit.ff.jest.config.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
```
Zainzzkk, line 1: same as above

external-api/src/cron/eStoreDealDirectoryCreationJob.cron.ts
abhi-markan: Have added.

external-api/src/cron/eStoreBuyerDirectoryCreationJob.cron.ts
abhi-markan: Have added.

azure-functions/acbs-function/unit.ff.jest.config.js
abhi-markan: Yes good shout, can I do this in a new PR so I can refactor all these files otherwise I fear this PR might get cumbersome to review?

external-api/src/cron/eStoreSiteCreationJob.cron.ts
abhi-markan: Have added.

dtfs-central-api/unit.ff.jest.config.js
abhi-markan: Yes good shout, can I do this in a new PR so I can refactor all these files otherwise I fear this PR might get cumbersome to review?

Zainzzkk: sure, ticket for it?

external-api/src/cron/eStoreTermStoreCreationJob.cron.ts
```diff
@@ -1,97 +1,106 @@
...
-      await cronJobLogs.updateOne(
-        { 'payload.dealId': { $eq: new ObjectId(eStoreData.dealId) } },
-        {
-          $set: {
-            'cron.term': {
-              status: ESTORE_CRON_STATUS.FAILED,
-              timestamp: getNowAsEpoch(),
-            },
+      // Validate each and every response status code
+      if (response.every((term) => ACCEPTABLE_STATUSES.includes(term?.status))) {
```
Zainzzkk, line 63: const for this?

external-api/src/cron/eStoreFacilityDirectoryCreationJob.cron.ts
```diff
@@ -0,0 +1,127 @@
...
+ *
+ * eStoreFacilityDirectoryCreationJob(eStoreData)
+ *   .then(() => console.log('Job completed successfully'))
+ *   .catch((error) => console.error('Job failed', error));
+ */
+export const eStoreFacilityDirectoryCreationJob = async (eStoreData: Estore): Promise<void> => {
+  try {
+    // Argument validation
+    if (
+      !eStoreData?.dealId ||
```
Zainzzkk, line 42: const for these?

external-api/src/interfaces/eStore.interface.ts
```diff
@@ -36,49 +40,54 @@ export interface EstoreSite {
...
   readonly destinationMarket: string;
   readonly riskMarket: string;
 }
 
 /**
  * Represents a facility folder within the eStore, identified by a facility identifier.
+ *
  * @interface EstoreFacilityFolder
  * @extends EstoreBuyer
  * @property {string} facilityIdentifier - Unique identifier for the facility.
```
Zainzzkk, line 82: is this the correct type?

external-api/src/v1/controllers/estore/eStore.controller.ts
```diff
@@ -130,34 +159,35 @@ export const create = async (req: EstoreRequest, res: Response) => {
...
-            $set: {
-              'cron.site.create': {
-                status: ESTORE_CRON_STATUS.COMPLETED,
-                response: siteExistsResponse.data.status,
-                timestamp: getNowAsEpoch(),
-                id: siteExistsResponse.data.siteId,
-              },
-              'cron.site.status': ESTORE_CRON_STATUS.COMPLETED,
-            },
+        await EstoreRepo.updateByDealId(eStoreData?.dealId, {
```
Zainzzkk, line 182: `eStoreData?.dealId` is repeated a few times and on line 139 it does not have optional chaining.  Is optional chaining required and could dealId be dereferenced with an if(!dealId) { do something }?

external-api/src/cron/eStoreTermStoreCreationJob.cron.ts
abhi-markan: Have created.

external-api/src/cron/eStoreFacilityDirectoryCreationJob.cron.ts
abhi-markan: Good spot, apologies I missed this have added.

external-api/src/interfaces/eStore.interface.ts
abhi-markan: Good spot have updated with an additional `@extends`

external-api/src/v1/controllers/estore/eStore.controller.ts
abhi-markan: Good spot, have updated and referred to the variables from `body`

```typescript
    const { dealId, siteId, facilityIdentifiers, supportingInformation, exporterName, buyerName, dealIdentifier, destinationMarket, riskMarket } = body;
```

external-api/src/v1/controllers/estore/eStore.controller.ts
```diff
@@ -187,46 +217,36 @@ export const create = async (req: EstoreRequest, res: Response) => {
...
-            {
-              $set: {
-                'cron.site.create': {
-                  response: siteCreationResponse?.data,
-                  status: ESTORE_CRON_STATUS.FAILED,
-                  timestamp: getNowAsEpoch(),
-                },
-                'cron.site.status': ESTORE_CRON_STATUS.FAILED,
-              },
+          await EstoreRepo.updateByDealId(eStoreData?.dealId, {
```
Zainzzkk, line 225: Think this one has been missed, can be done in the next PR

external-api/src/v1/controllers/estore/eStore.controller.ts
Zainzzkk: Think this one has been missed, can be done in the next PR

Zainzzkk (dismissed):

azure-functions/acbs-function/package.json
```diff
@@ -25,7 +25,8 @@
     "prettier": "prettier --no-error-on-unmatched-pattern --check **/*.{ts,js} --ignore-path ../../.prettierignore",
     "prettier:fix": "prettier --write **/*.{ts,js} --ignore-path ../../.prettierignore",
     "unit-test": "jest --coverage --verbose --config=unit.jest.config.js",
-    "unit-test-dev": "jest --verbose --config=unit.jest.config.js"
+    "unit-test-dev": "jest --verbose --config=unit.jest.config.js",
+    "unit-test-ff": "jest --coverage --verbose --config=unit.ff.jest.config.js --passWithNoTests"
```
ttbarnes, line 29: What's the reason for adding this please, especially with the `passWithNoTests` flag?

Looks like this isn't consumed anywhere

external-api/src/constants/emailTemplates.constant.ts
```diff
@@ -1,3 +1,3 @@
 export const EMAIL_TEMPLATES = {
-  ESTORE_FAILED: '50f4ba6c-9eef-4e90-bf86-5945cd08d0cb',
+  ESTORE_FAILED: '345ae136-7442-47b4-b26f-1a3201fead41',
```
ttbarnes, line 2: Just to confirm - do we need a ticket to remove the `50f4ba6c-9eef-4e90-bf86-5945cd08d0cb` template once deployed?

abhi-markan: I have already removed that template and updated with the latest one as the old template was not fit for purpose.

external-api/src/v1/controllers/estore/eStore.controller.ts
abhi-markan: Ah! apologies have updated.

Zainzzkk (dismissed):

dtfs-central-api/unit.ff.jest.config.js
abhi-markan: A ticket already exist.

azure-functions/acbs-function/package.json
abhi-markan: This is to ensure feature flag unit test cases can be executed and there can be occasions when no unit tests for feature flags exists. I have added this across all the micro-services.

ttbarnes: Thanks @abhi-markan makes sense - should we add this to the CI/CD pipeline? If so I would suggest in a new PR

ChristianMcc: Most definitely not for this PR but I'm quite strongly against having config to run unit tests in particular with feature flags. Unit tests test the actual implementation on the lowest level and should be aware of feature flags, so our tests should be in charge of mocking what the value of certain env variables/feature flags are rather than the CI server. This comment definitely doesn't apply to other tests like E2E tests which are abstracted away where I do think the CI/CD pipeline should be in charge of setting the variables.

Ignore this comment for this PR as not for now but let's definitely organise that chat re feature flag testing especially as we're looking to bring more in soon

Zainzzkk (dismissed):

trade-finance-manager-api/src/v1/rest-mappings/mappings/facilities/getLatestFacilityEndDateValues.test.ts
```diff
@@ -45,24 +45,6 @@ describe('getLatestFacilityEndDateValues', () => {
       // Assert
       expect(result).toEqual({ isUsingFacilityEndDate: true, facilityEndDate: mockFacilityEndDate });
     });
-
-    it('should return the the latest amendment value when the FED has been amended', () => {
```
ttbarnes, line 49: Why has this been removed please? AFAIK, facility end date is unrelated to Estore?

trade-finance-manager-ui/server/controllers/case/index.ff-test.js
```diff
@@ -0,0 +1,121 @@
...
+jest.mock('@ukef/dtfs2-common', () => ({
+  ...jest.requireActual('@ukef/dtfs2-common'),
+  isTfmFacilityEndDateFeatureFlagEnabled: jest.fn(),
+}));
+
+const res = mockRes();
+
+const TOKEN = 'test-token';
+
+const SESSION = {
```
ttbarnes, line 15: Can we use lowercase consts for `TOKEN` and `SESSION` please? Otherwise this will be the only instance of uppercase token/session consts

trade-finance-manager-api/src/v1/rest-mappings/mappings/facilities/getLatestFacilityEndDateValues.test.ts
abhi-markan: I have moved this to `trade-finance-manager-api/src/v1/rest-mappings/mappings/facilities/getLatestFacilityEndDateValuest.ff-test.ts` a FF specific test.

external-api/src/cron/eStoreDocumentsCreationJob.cron.ts
```diff
@@ -0,0 +1,119 @@
...
+import { EstoreRepo } from '../repositories/estore/estore-repo';
+import { AzureDocument, EstoreDocument, DocumentCreationResponse, EstoreErrorResponse, Estore } from '../interfaces';
+import { cron } from '../helpers/cron';
+import { ENDPOINT, ESTORE_CRON_STATUS } from '../constants';
+import { uploadSupportingDocuments } from '../v1/controllers/estore/eStoreApi';
+import { getNowAsEpoch } from '../helpers/date';
+
+const ACCEPTABLE_STATUSES = [HttpStatusCode.Ok, HttpStatusCode.Created];
+
+export const eStoreDocumentsCreationJob = async (eStoreData: Estore): Promise<void> => {
```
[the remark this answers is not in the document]
abhi-markan: Have added.

trade-finance-manager-ui/server/controllers/case/index.ff-test.js
abhi-markan: Good spot, have updated.

Zainzzkk (approved):

ttbarnes (dismissed):

ttbarnes (dismissed):

ChristianMcc (dismissed):

ttbarnes (approved):

PR #3392, opened 2024-08-19
feat(FN-2594): add dynamic dropdown for find previous reports year input

JoshBinns2000:
## Introduction :pencil2:
We want to add a dynamic drop down to the find utilisation reports by year which provides a list of years where a report exists for the specified bank.

- Adds a new query parameter to the `/v1/banks` `dtfs-central-api` endpoint
- Adds a series of `datalist` tags to the nunjucks template*
- Add a new script
- Add the `ts-loader` webpack loader so that scripts can be written in typescript
- Updates unit and api tests

*The method for generating the dropdown is to render a data list for each of the banks with reporting years with element id `datalist--bankId-<bankId>`. You can attach this data list as a dropdown to an input by including the following tag:

```html
<datalist id="datalist--bankId-123"> ... </datalist>
<input list="datalist--bankId-123" ... />
```

The script then dynamically updates the data list id which populates the `list` attribute depending on which bank the user has selected.


dtfs-central-api/src/repositories/utilisation-reports-repo/utilisation-report.repo.ts
```diff
@@ -236,4 +236,24 @@ export const UtilisationReportRepo = SqlDbDataSource.getRepository(UtilisationRe
...
     });
   },
+
+  /**
+   * Finds the years where the bank with the supplied id submitted
+   * a utilisation report
+   * @param bankId - The bank id
+   * @returns A set of years where the bank submitted a report
+   */
+  async findReportingYearsByBankId(bankId: string): Promise<Set<number>> {
```
JoshBinns2000, line 246: Considering that this only gets tested in the api tests, do we think it is worth adding a unit test?

trade-finance-manager-ui/scripts/find-utilisation-reports-by-year-year-input-dropdown.ts
```diff
@@ -0,0 +1,29 @@
+const setupFindUtilisationReportsByYearYearInput = (): void => {
```
ttbarnes, line 1: `YearYear`?

Maybe could be called something like `addListAttributesToYearInputs`, `setupFindUtilisationReports` feels strange to me

trade-finance-manager-ui/scripts/find-utilisation-reports-by-year-year-input-dropdown.ts
```diff
@@ -0,0 +1,29 @@
...
+      yearInput.setAttribute('list', dataListId);
+    }
+
+    bankRadio.addEventListener('click', () => {
+      yearInput.setAttribute('list', dataListId);
+    });
+  });
+};
+
+setupFindUtilisationReportsByYearYearInput();
```
ttbarnes, line 29: Why do we need a client side script to set some attributes, can this not be done in nunjucks?

If not, will the core functionality continue to work when JavaScript is disabled?

JoshBinns2000: This cannot be done in nunjucks - it is something that updates attributes based on user input. The page is already designed to work without JS being enabled - this is an add-on which was requested to get some nicer behaviour when JS is enabled

trade-finance-manager-ui/scripts/find-utilisation-reports-by-year-year-input-dropdown.ts
JoshBinns2000: This script is specific only to the `find-utilisation-reports-by-year` page, and I think it's important to have the function and file name to reflect this. The function itself also does not actually add any attributes but instead sets up the event listeners, hence `setup`. The choice to include the fact that the input specifically being targeted by this script is the year input is that there are radio inputs for each of the banks on the page, so saying `input` by itself would not be clear

dtfs-central-api/api-tests/v1/bank/banks-get.api-test.ts
```diff
@@ -0,0 +1,107 @@
...
+    });
+  });
+
+  function* idGenerator(): Generator<number, number, unknown> {
+    let id = 1;
+    while (true) {
+      yield id;
+      id += 1;
+    }
+  }
```
ttbarnes, line 106: I can see this is a duplicate in at least 4x places now, could we move to a DRY test helper?

dtfs-central-api/src/repositories/utilisation-reports-repo/utilisation-report.repo.test.ts
```diff
@@ -0,0 +1,55 @@
...
+      const bankId = '123';
+      const reports = [
+        UtilisationReportEntityMockBuilder.forStatus('RECONCILIATION_IN_PROGRESS')
+          .withBankId(bankId)
+          .withReportPeriod({
+            start: { month: 1, year: 2020 },
+            end: { month: 1, year: 2021 },
+          })
+          .build(),
+        UtilisationReportEntityMockBuilder.forStatus('PENDING_RECONCILIATION')
```
ttbarnes, line 29: Constant for statuses?

dtfs-central-api/src/repositories/utilisation-reports-repo/utilisation-report.repo.ts
```diff
@@ -236,4 +236,24 @@ export const UtilisationReportRepo = SqlDbDataSource.getRepository(UtilisationRe
...
+        bankId,
+        status: Not('REPORT_NOT_RECEIVED'),
+      },
+      select: ['reportPeriod'],
+    });
+    return utilisationReports.reduce(
+      (uniqueReportPeriodYears, { reportPeriod }) => uniqueReportPeriodYears.add(reportPeriod.start.year).add(reportPeriod.end.year),
+      new Set<number>(),
+    );
+  },
```
ttbarnes, line 258: This is becoming large, nearly 300 lines, anyway we can split up / extract and import? Probably not for this PR.

dtfs-central-api/src/repositories/utilisation-reports-repo/utilisation-report.repo.test.ts
JoshBinns2000: Because this is a typescript file, there is no difference between using the constant and the string literal because of how we have set everything up

dtfs-central-api/src/repositories/utilisation-reports-repo/utilisation-report.repo.ts
JoshBinns2000: I don't think there is anything wrong with the size of this file (see the various `api.js` files for instance). The reason we should keep these methods in this file aside from anything else is that they are appending to the base repository api through `.extend`, which keeps our binding of `this` consistent. If we try to move the functions outside of this file and them import them, we start having to worry about binding `this` to the correct value, which we don't want to do

dtfs-central-api/src/v1/controllers/bank/get-banks.controller.ts
```diff
@@ -0,0 +1,51 @@
+import { Response } from 'express';
+import { HttpStatusCode } from 'axios';
+import { CustomExpressRequest, Bank, ApiError, ApiErrorResponseBody } from '@ukef/dtfs2-common';
+import { getAllBanks } from '../../../repositories/banks-repo';
```
ttbarnes, line 4: I'm finding having this new TS controller and the existing JS controller with the same name confusing, which controller is used for what?

dtfs-central-api/api-tests/v1/bank/banks-get.api-test.ts
JoshBinns2000: The specifics of an `idGenerator` could vary depending on what each file wants to test. For instance, here we are specifically testing ids which are positive integers, but someone may mistakenly import this to their test file and expect that it generates some other form of an "id" (`ObjectId`, for instance) which they would then need to check the actual implementation of `idGenerator` in order to see what was going wrong with their test

dtfs-central-api/src/v1/routes/bank-routes.js
```diff
@@ -17,6 +17,24 @@ const handleExpressValidatorResult = require('../validation/route-validators/exp
 /**
  * @openapi
  * /bank:
+ *   get:
+ *     summary: Get an array of all banks in the banks collection
```
ttbarnes, line 21: Not for this PR - curious why the route is `/bank`, without an "S", since it is returning multiple banks.

dtfs-central-api/src/v1/controllers/bank/get-banks.controller.ts
JoshBinns2000: They don't have the same name as far as I can see - one is called `get-banks.controller` and the other is called `get-bank.controller`

dtfs-central-api/src/v1/routes/bank-routes.js
JoshBinns2000: I thought the same thing - the route I was expecting to see was `/banks` and `/banks/:id`, but for some reason the `/bank` route was chosen instead

trade-finance-manager-ui/component-tests/utilisation-reports/find-utilisation-reports-by-year.component-test.ts
```diff
@@ -0,0 +1,145 @@
...
+    const bankReportingYearsDataLists: BankReportingYearsDataListViewModel[] = [{ bankId: '123', reportingYears: [2020, 2021, 2022, 2023] }];
+    const wrapper = getWrapper({ bankReportingYearsDataLists });
+
+    // Assert
+    wrapper.expectElement('datalist#datalist--bankId-123').toExist();
+    wrapper.expectElement('datalist#datalist--bankId-123 option').toHaveCount(4);
+    wrapper.expectElement('datalist#datalist--bankId-123 option[value="2020"]').toExist();
+    wrapper.expectElement('datalist#datalist--bankId-123 option[value="2021"]').toExist();
+    wrapper.expectElement('datalist#datalist--bankId-123 option[value="2022"]').toExist();
+    wrapper.expectElement('datalist#datalist--bankId-123 option[value="2023"]').toExist();
```
ttbarnes, line 82: DRY variable for `datalist#datalist--bankId-123`? Maybe `selector` const?

trade-finance-manager-ui/scripts/find-utilisation-reports-by-year-year-input-dropdown.ts
ttbarnes: Designed to work without JS being enabled? 😞

JoshBinns2000: Yes - you can always input a year without JS being enabled. This script simply means that you will have a helpful dropdown that appears displaying the years where we know reports exist when you are typing a value into the year input

trade-finance-manager-ui/scripts/find-utilisation-reports-by-year-year-input-dropdown.ts
ttbarnes: Should we remove the dupicate `year-year` in the file and function name? It's a bit strange IMO.

I don't feel that naming reflects what it's doing.

The event listeners are called to add HTML attributes, to me the file name reads like it's just finding for something and nothing else.

trade-finance-manager-ui/templates/utilisation-reports/find-utilisation-reports-by-year.njk
```diff
@@ -1,39 +1,35 @@
...
 {% block content %}
-  <div class="govuk-grid-row govuk-!-padding-top-7">
-    <div
-      class="govuk-grid-column-full">
+  <div class='govuk-grid-row govuk-!-padding-top-7'>
+    <div class='govuk-grid-column-full'>
       {% if errorSummary | length %}
         {{ govukErrorSummary({
-          titleText: "There is a problem",
+          titleText: 'There is a problem',
```
ttbarnes, line 15: Do we need to change from double quote to single quote? Looks like some digression has been happening - when searching the nunjucks files:

- `text: "` = 706 results
- `text: '` = 67 results

dtfs-central-api/src/repositories/utilisation-reports-repo/utilisation-report.repo.test.ts
ttbarnes: There is a difference, it's not DRY and If the `PENDING_RECONCILIATION` string needs to change, we'll need to update (currently) 166 instances in 75 files.

JoshBinns2000: Using the constant is just as DRY as using a string literal - you either repeat the string everywhere or repeat the constant everywhere. In terms of updating the constant, we discussed this internally before, but the constant updating would result in the same number of updates whether we use the constant or the string literal. This is because these constants are functionally the same as an enum, so the key will always be the same as the value and updating the `PENDING_RECONCILIATION` status would require updates on both the constant and the string

trade-finance-manager-ui/templates/index.njk
```diff
@@ -56,5 +56,6 @@
     <script src="/assets/js/mojFrontend.js" type="text/javascript" integrity="sha512-VJYfsojYZY4stpSFP22EQ/CXqqrW9BWYWKaiEnYGmyw22rggV4DCe+Lzn8L7Gt+bRq3AyWuNYFadfrCmao0o+A==" crossorigin="anonymous"></script>
     <script src="/assets/js/disableFormSubmitOnSubmission.js" type="text/javascript" integrity="sha512-tCINlYTcYERel2bdysTMl5KaemDdEyog5JSS3hYQqHS6ZIyT4ATLu4oTPerbZ+vAOn4AAzDEsRBSQVacGhoieQ==" crossorigin="anonymous"></script>
     <script src="/assets/js/enableSelectAllTableCheckbox.js" type="text/javascript" integrity="sha512-ER/5raNAJX5VTcyrRBjVExC/f39WaVY4r8k1CyGqJUad95QZ0A1H6zHtsFpdf0JDBLV0acXzSA1B2mkKC8g5Tg==" crossorigin="anonymous"></script>
+    <script src="/assets/js/findUtilisationReportsByYearYearInputDropdown.js" type="text/javascript" crossorigin="anonymous"></script>
```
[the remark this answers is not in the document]
JoshBinns2000: For some reason this file didn't need one

dtfs-central-api/src/repositories/utilisation-reports-repo/utilisation-report.repo.test.ts
ttbarnes: It depends on the change of the constant really.

I think all statuses should be constants as this seems to be the standard in DTFS, if you search for e.g `STATUS = {`

Appreciate this is not for this PR.

trade-finance-manager-ui/templates/index.njk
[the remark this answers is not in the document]
ttbarnes: All client scripts should have an intergrity hash? Would be strange for this to be the only script without an intergrity hash

dtfs-central-api/src/repositories/utilisation-reports-repo/utilisation-report.repo.test.ts
JoshBinns2000: This status is a constant - we have the `UTILISATION_REPORT_RECONCILIATION_STATUS` constant and the `UtilisationReportReconciliationStatus` type which is generated from that constant - it is the type alias which allows us to safely use the string literal and reduce code bloat

>It depends on the change of the constant really.

There should never be a change to this constant where the key becomes different to the value - if there ever was, many more things would go wrong than just our references to that constant

trade-finance-manager-ui/templates/index.njk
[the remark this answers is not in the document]
JoshBinns2000: I know they should, but running locally did not generate an integrity hash for this file and so I didn't have one that I could add

trade-finance-manager-ui/templates/utilisation-reports/find-utilisation-reports-by-year.njk
RLavender98: Definitely not for this PR but wondering if we have ever considered linting for nunjucks files? It's hard to infer the intended patterns for going forwards given how inconsistent the existing files are on things like quotes and spacing and depending on individual IDE setup it definitely suggests different things on spacing etc to others

trade-finance-manager-ui/templates/index.njk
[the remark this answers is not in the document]
RLavender98: Interesting, I'm not familiar with how the integrity hashes get generated, I'm guessing this is to do with the script being in typescript? Do we need to do some additional config perhaps?

RLavender98 (commented):
Neat solution to the problem, I had been imagining it being a lot more gnarly so nice job!

dtfs-central-api/src/v1/routes/bank-routes.js
```diff
@@ -17,6 +17,24 @@ const handleExpressValidatorResult = require('../validation/route-validators/exp
 /**
  * @openapi
  * /bank:
+ *   get:
```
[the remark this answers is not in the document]
JoshBinns2000: I was meaning to do that but forgot - thanks for catching!

dtfs-central-api/src/repositories/utilisation-reports-repo/utilisation-report.repo.test.ts
```diff
@@ -0,0 +1,55 @@
...
+
+      when(findSpy)
+        .calledWith({
+          where: {
+            bankId,
+            status: Not<UtilisationReportReconciliationStatus>('REPORT_NOT_RECEIVED'),
+          },
+          select: ['reportPeriod'],
+        })
+        .mockResolvedValue(reports);
```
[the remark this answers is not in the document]
JoshBinns2000: That is true, but typescript would be annoyed at me doing that. However, I think because we are purposefully doing that here it does make more sense to only include that property and do some explicit generics (type-hacking, if you will) to make typescript happy

trade-finance-manager-api/src/v1/api.js
```diff
@@ -1208,6 +1208,22 @@ const getAllBanks = async () => {
...
+  const url = `${DTFS_CENTRAL_API_URL}/v1/bank`;
+  const response = await axios.get(url, {
+    headers: headers.central,
+    params: {
+      includeReportingYears: true,
+    },
+  });
+
+  return response.data;
+};
```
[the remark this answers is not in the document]
JoshBinns2000: We did, but without a generic we wouldn't get nice typing on these functions (ie. if we set the query to `"true"`, we know that the type will include the `reportingYears` array) - I'll have a think about how to contain this in one method

dtfs-central-api/src/v1/controllers/bank/get-banks.controller.test.ts
```diff
@@ -0,0 +1,87 @@
...
+        { ...aBank(), id: '456' },
+      ];
+      jest.mocked(getAllBanks).mockResolvedValue(banks);
+
+      when(findReportingYearsByBankIdSpy)
+        .calledWith('123')
+        .mockResolvedValue(new Set([2021, 2022]));
+      when(findReportingYearsByBankIdSpy)
+        .calledWith('456')
+        .mockResolvedValue(new Set([2022, 2023, 2024]));
```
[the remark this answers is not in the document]
JoshBinns2000: Done as part of resolving the check for `isVisibleInTfmUtilisationReports`

trade-finance-manager-ui/scripts/find-utilisation-reports-by-year-year-input-dropdown.ts
```diff
@@ -0,0 +1,29 @@
...
+  bankRadios.forEach((bankRadio) => {
+    const bankId = bankRadio.value;
+    const dataListId = `datalist--bankId-${bankId}`;
+
+    if (bankRadio.checked) {
+      yearInput.setAttribute('list', dataListId);
+    }
+
+    bankRadio.addEventListener('click', () => {
+      yearInput.setAttribute('list', dataListId);
```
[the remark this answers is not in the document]
JoshBinns2000: I hadn't heard of data lists either, but they were added as part of HTML 5 and seem to offer a nice, light-weight solution to drop-downs

trade-finance-manager-ui/templates/index.njk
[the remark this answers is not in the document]
JoshBinns2000: From some testing, it seems like none of the scripts require an integrity hash anymore - I imagine there has been some update to TFM that has gone unnoticed which turned off this requirement

trade-finance-manager-api/src/v1/api.js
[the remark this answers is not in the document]
JoshBinns2000: Decided we don't need any generic behaviour here as TFM-API doesn't need to know the specifics of the response body depending on the query

trade-finance-manager-ui/scripts/find-utilisation-reports-by-year-year-input-dropdown.ts
JoshBinns2000: To match the other script which performs a similar action with respect to event listeners, I have renamed the file to `enable-find-utilisation-reports-by-year-dropdown-autocomplete` and function to `enableFindUtilisationReportsByYearDropdownAutocomplete`

dtfs-central-api/api-tests/v1/bank/banks-get.api-test.ts
JoshBinns2000: I have added a test helper called `getIncrementingPositiveIntegerIdGenerator`

trade-finance-manager-ui/templates/index.njk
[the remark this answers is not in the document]
JoshBinns2000: I have posted a message in devs because it seems these hashes aren't even required on `main` anymore. Going to resolve this thread

trade-finance-manager-ui/templates/utilisation-reports/find-utilisation-reports-by-year.njk
JoshBinns2000: No idea why this change happened, but we definitely should look into adding better and more consistent prettier for nunjucks files

dtfs-central-api/test-helpers/test-data/payment-matching-tolerances.ts
```diff
@@ -1,16 +1,9 @@
 import { CURRENCY, PaymentMatchingToleranceEntityMockBuilder } from '@ukef/dtfs2-common';
+import { getIncrementingPositiveIntegerIdGenerator } from '../get-integer-id-generator';
 
 export const aListOfZeroThresholdActivePaymentMatchingTolerances = () => {
-  const toleranceIdGenerator = idGenerator();
+  const toleranceIdGenerator = getIncrementingPositiveIntegerIdGenerator(1);
```
ttbarnes, line 5: What's the reason for introducing `getIncrementingPositiveIntegerIdGenerator` instead of using `idGenerator`?

trade-finance-manager-api/src/v1/api.js
```diff
@@ -1198,14 +1198,21 @@ const findBankById = async (bankId) => {
   }
 };
 
+/**
+ * @typedef {object} GetBanksQuery
+ * @property {boolean | undefined} includeReportingYears - Whether or not to include the bank reporting years
+ */
+
```
ttbarnes, line 1205: Documentation seems funny here, can't see `includeReportingYears` anywhere?

dtfs-central-api/src/v1/swagger-definitions/bank.js
```diff
@@ -4,27 +4,53 @@
...
  *         items:
  *           type: string
- *           example: maker1@ukexportfinance.gov.uk
- *       mga:
+ *       emails:
  *         type: array
  *         items:
  *           type: string
- *           example: 'document.pdf'
+ *           format: email
```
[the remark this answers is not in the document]
JoshBinns2000: `format` is a special swagger docs field which tells you that the string is an email. For instance, a `Date` does not actually exist in JSON, so we document that using `type: string ... format: date` as `type: date` is invalid.
I will re-introduce an example email though

dtfs-central-api/test-helpers/test-data/payment-matching-tolerances.ts
JoshBinns2000: Because `idGenerator` is not very specific. When defined inside the test file, it is easy enough to figure out what it is doing, but having an export called `idGenerator` could mean anything. The use of `get` is to communicate that calling this function returns a generator to you

trade-finance-manager-api/src/v1/api.js
JoshBinns2000: That's because we just pass in the query supplied to the `getBanks` function. This type becomes necessary inside the `banks.controller.ts` file where it now requires that the supplied query parameter matches this type definition

dtfs-central-api/src/repositories/utilisation-reports-repo/utilisation-report.repo.test.ts
```diff
@@ -0,0 +1,52 @@
...
+
+describe('UtilisationReportRepo', () => {
+  afterEach(() => {
+    jest.resetAllMocks();
+  });
+
+  describe('findReportingYearsByBankId', () => {
+    const findSpy = jest.spyOn(UtilisationReportRepo, 'find');
+
+    const aUtilisationReportEntityWithOnlyReportPeriodSelected = (reportPeriod: ReportPeriod): UtilisationReportEntity =>
```
ttbarnes, line 14: Can we shorten some of the variable names, they are really long and hard to follow e.g

```ts
- aUtilisationReportEntityWithOnlyReportPeriodSelected
- aUtilisationReportForStatusBankIdAndMonthlyReportPeriod (what does `ForStatusBankId` mean?)
- getIncrementingPositiveIntegerIdGenerator
```
Maybe could be something more like this:

```ts
- utilisationReport({ selectedReportPeriods: 1 });
- utilisationReport({ monthlyReportPeriod: true });
- idGenerator
```

JoshBinns2000: Shorter variable names often lack any context and can therefore usually be extremely hard to follow. Especially considering that this is a test file, I don't see the downside to having these longer variable/function names

dtfs-central-api/test-helpers/test-data/payment-matching-tolerances.ts
ttbarnes: Do we now have multiple ID generators used for testing purposes? Maybe could call it `getIdGenerator`, I don't think that's too generic.

IMO, `getIncrementingPositiveIntegerIdGenerator` is too long and descriptive.

dtfs-central-api/src/repositories/utilisation-reports-repo/utilisation-report.repo.test.ts
ttbarnes: I am finding the long variables names extremely hard to follow.

I worry that someone new (like me), could look at such functions/names in the future and find it hard to follow.

How else could this be improved without shortening the names?

dtfs-central-api/test-helpers/test-data/payment-matching-tolerances.ts
JoshBinns2000: Before SQL, id in this project would usually refer to an `ObjectId`. It is because of SQL that id can now mean integer or `ObjectId`, hence the need to distinguish between them so that a future developer does not use this generator expecting it to produce `ObjectId`s

ttbarnes: The 2x function names do not highlight this at all, neither `idGenerator` or `getIncrementingPositiveIntegerIdGenerator` say to me "could be an integer or `ObjectId`", it just reads "i'm generating an incremental ID" really.

Should we maybe have 1x ID generator that conditionally creates a different ID?

dtfs-central-api/src/repositories/utilisation-reports-repo/utilisation-report.repo.test.ts
JoshBinns2000: I don't believe the long variable names are hard to follow, and actually think that shorter variable names make things much harder. For example, as someone who joined this code base nearly a year ago, existing tests often were and still are very hard to follow due to the lack of descriptive variable names.

If the choice is between longer and shorter variable names with readability being cited as the threshold breaker, then I do believe longer variable names are the way to go

dtfs-central-api/test-helpers/test-data/payment-matching-tolerances.ts
JoshBinns2000: `getIncrementingPositiveIntegerIdGenerator` communicates to me that you create a generator which yields positive integers which increment after each `.next()` is called. 

`idGenerator` just communicates that it is an id generator (which it technically isn't because you have to call it to get the `Generator` object) which is lacking context around what it means by id

dtfs-central-api/src/repositories/utilisation-reports-repo/utilisation-report.repo.test.ts
ttbarnes: What would the limit be for such long variable names, what happens if we want to add more params/conditions? It becomes unreadable e.g

```ts
aUtilisationReportEntityWithOnlyReportPeriodSelected
aUtilisationReportEntityWithOnlyReportPeriodSelectedWithXDate
aUtilisationReportEntityWithTwoReportPeriodsSelectedWithOnePastADeadline
aUtilisationReportEntityWithOneInvalidAndOneValidReportPeriodsSelected
```

JoshBinns2000: I don't think those names are necessarily unreadable, but then you can start citing context for a specific state if the variable name becomes unreasonably long. For an actual example, we have cases where we want to define a report to have fields which reflect that the report has been uploaded. Rather than calling the function `aUtilisationReportEntityWithUploadedByUserIdAndDateUploadedAnd...`, we call the function `anUploadedUtilisationReportEntity` which can accept custom parameters. It's all context specific, but in the context of this file the thing we care about is that the utilisation report only has a report period defined on it because we are selecting that field in the typeorm query, which is something that does not normally happen. The extra documentation we get for free that the entity yielded by this function is a report _only_ with a report period selected would be hard to document without context of how this function is implemented

dtfs-central-api/test-helpers/test-data/payment-matching-tolerances.ts
RLavender98: Perhaps a compromise would be to go with `getSqlIdGenerator` and then that makes it clear and disambiguates it from other ids we have in use in the codebase? And if we ever wanted another incrementing positive integer id generator for a different kind of id, we could make a second helper and have both the sql one and the other one point to the same function which does that behind the scenes?

JoshBinns2000: I have tried to stay away from `SQL id` like names because an SQL id does not need to be an integer (for example, our `FacilityUtilisationData` table has a primary key which is `NVARCHAR(255)`). However, for most cases we will be using integers for SQL ids so I am happy with that name

dtfs-central-api/src/repositories/utilisation-reports-repo/utilisation-report.repo.test.ts
RLavender98: In specific case I would actually be tempted to just inline this as it is just one line and only used twice in quite a specific niche case and then avoids the need to try and come up with a good name? I think should be accompanied by a comment as well since it is a little odd and worth outlining it is intentional that we are using the `as`.

RLavender98 (approved):
LGTM - nice work!

dtfs-central-api/api-tests/sql-db-helper.ts
```diff
@@ -7,7 +7,7 @@ import {
   UtilisationReportEntity,
 } from '@ukef/dtfs2-common';
 import { SqlDbDataSource } from '@ukef/dtfs2-common/sql-db-connection';
-import { aListOfZeroThresholdActivePaymentMatchingTolerances } from '../test-helpers/test-data/payment-matching-tolerances';
+import { aListOfZeroThresholdActivePaymentMatchingTolerances } from '../test-helpers';
```
abhi-markan, line 10: Can we please make this function names and variables more readable, other one looses the context behind length names.

dtfs-central-api/api-tests/v1/bank/banks-get.api-test.ts
```diff
@@ -0,0 +1,102 @@
...
+
+    it('responds with a 200 (Ok) and all banks with reporting years', async () => {
+      // Arrange
+      const requestUrl = axios.getUri({
+        url: BASE_URL,
+        params: { includeReportingYears: true },
+      });
+
+      const utilisationReports = [
+        aUtilisationReportForStatusBankIdAndMonthlyReportPeriod('PENDING_RECONCILIATION', BANK_ID_ONE, { month: 1, year: 2021 }),
```
abhi-markan, line 79: * Variable names are too long.
* CAn these values please be picked from constants?

dtfs-central-api/src/repositories/utilisation-reports-repo/utilisation-report.repo.ts
```diff
@@ -236,4 +236,24 @@ export const UtilisationReportRepo = SqlDbDataSource.getRepository(UtilisationRe
...
+   * Finds the years where the bank with the supplied id submitted
+   * a utilisation report
+   * @param bankId - The bank id
+   * @returns A set of years where the bank submitted a report
+   */
+  async findReportingYearsByBankId(bankId: string): Promise<Set<number>> {
+    const utilisationReports = await this.find({
+      where: {
+        bankId,
+        status: Not('REPORT_NOT_RECEIVED'),
```
abhi-markan, line 250: Constant please.

dtfs-central-api/test-helpers/get-sql-id-generator.ts
```diff
@@ -0,0 +1,11 @@
+/**
```
abhi-markan, line 1:
```suggestion
/**
 * Generates a sequence of unique SQL IDs starting from 1.
 *
 * This function returns a generator that yields an incrementing number
 * each time it is called. It can be used to simulate the behavior of
 * an auto-incrementing SQL ID, where each call to the generator provides
 * the next ID in the sequence.
 *
 * @generator
 * @yields {number} The next unique SQL ID in the sequence.
 * @returns {number} The final value of the ID, although in typical use cases,
 * this generator function is intended to run indefinitely.
 *
 * @example
 * const idGenerator = getSqlIdGenerator();
 * console.log(idGenerator.next().value); // 1
 * console.log(idGenerator.next().value); // 2
 * console.log(idGenerator.next().value); // 3
 */
```

trade-finance-manager-ui/server/controllers/utilisation-reports/find-reports-by-year/index.ts
```diff
@@ -25,28 +30,35 @@ const getBankIdQueryAndYearQueryAsString = (
   return { bankIdQuery, yearQuery };
 };
 
+const mapBankWithReportingYearsToDataListViewModel = (bank: BankWithReportingYearsResponseBody): BankReportingYearsDataListViewModel => ({
```
abhi-markan, line 33: Missing doc string, amongst other functions in this controller.

trade-finance-manager-ui/templates/index.njk
```diff
@@ -56,5 +56,6 @@
     <script src="/assets/js/mojFrontend.js" type="text/javascript" integrity="sha512-VJYfsojYZY4stpSFP22EQ/CXqqrW9BWYWKaiEnYGmyw22rggV4DCe+Lzn8L7Gt+bRq3AyWuNYFadfrCmao0o+A==" crossorigin="anonymous"></script>
     <script src="/assets/js/disableFormSubmitOnSubmission.js" type="text/javascript" integrity="sha512-tCINlYTcYERel2bdysTMl5KaemDdEyog5JSS3hYQqHS6ZIyT4ATLu4oTPerbZ+vAOn4AAzDEsRBSQVacGhoieQ==" crossorigin="anonymous"></script>
     <script src="/assets/js/enableSelectAllTableCheckbox.js" type="text/javascript" integrity="sha512-ER/5raNAJX5VTcyrRBjVExC/f39WaVY4r8k1CyGqJUad95QZ0A1H6zHtsFpdf0JDBLV0acXzSA1B2mkKC8g5Tg==" crossorigin="anonymous"></script>
+  <script src="/assets/js/enableFindUtilisationReportsByYearDropdownAutocomplete.js" integrity="sha512-b1kS23cn0PI6KoBLTw3xeNJ2qi11c5nGrnDP3zk9DCuayz0/fu4PbzbUiqPaD5062eI9KJRp1e1ZvApQLGjpug==" type="text/javascript" crossorigin="anonymous"></script>
```
abhi-markan, line 59: `findYearlyUtilisationReportsDropdown` might be a suitable file name.

trade-finance-manager-ui/templates/utilisation-reports/find-utilisation-reports-by-year.njk
```diff
@@ -2,31 +2,27 @@
...
-            'data-cy': 'error-summary'
+            "data-cy": "error-summary"
           }
         }) }}
       {% endif %}
       <h1 class="govuk-heading-xl govuk-!-margin-bottom-6" data-cy="find-reports-by-year-heading">
         Find reports by year
       </h1>
-      <form method="GET" data-cy="form">
+      <form method="GET" data-cy="form" id="find-utilisation-reports-by-year--form">
```
abhi-markan, line 25: We haven't used `id` attribute in form tag, can `data-cy` attribute rather please be used?

abhi-markan (changes requested):

trade-finance-manager-ui/templates/utilisation-reports/find-utilisation-reports-by-year.njk
JoshBinns2000: The `id` attribute is used in the public script to select this specific form, and using `data-cy` as a selector for your public facing code is not best practice

trade-finance-manager-ui/templates/index.njk
JoshBinns2000: I think that name is confusing because it suggests that the script is "finding the yearly utilisation reports dropdown", which doesn't make sense. It also removes the intentional linking to the `find utilisation reports by year` page by including `FindUtilisationReportsByYear` in the name

dtfs-central-api/api-tests/sql-db-helper.ts
JoshBinns2000: I'm not sure what you mean. Could you provide an example shorter name for this which maintains the same level of context for reference?

trade-finance-manager-ui/templates/index.njk
abhi-markan: In that case can be just `findUtilisationReportsByYear` please and what the script actually does can be explained in-depth as comments please.

JoshBinns2000: Again, that name implies that the script performs an action ("_finds_ utilisation reports by year") which can lead to confusion even with a comment in the file as the comment will only be in the source file; not the actually built script (ie. `find-utilisation-reports-by-year.ts` would have the comment, but `public/js/findUtilisationReportsByYear.js` would not)

dtfs-central-api/api-tests/sql-db-helper.ts
abhi-markan: Can this please be `activeZerotThresholdCurrencies` it defeats the purpose to actively name every single imperative operation a function is performing this is where doc string will help which also seems to be missing from the helper function `aListOfZeroThresholdActivePaymentMatchingTolerances` can this also please be added (appreciate this is not in your PR scope).

trade-finance-manager-ui/templates/utilisation-reports/find-utilisation-reports-by-year.njk
abhi-markan: Understood thanks for explaning.

dtfs-central-api/api-tests/sql-db-helper.ts
JoshBinns2000: As this test helper was originally added by @RLavender98, could I leave this with you?

dtfs-central-api/api-tests/v1/bank/banks-get.api-test.ts
JoshBinns2000: Could you please make a suggestion for another name which retains the same context? If we were to inline this using our builders, it would become

```typescript
const utilisationReports = [
  UtilisationReportEntityMockBuilder.forStatus('PENDING_RECONCILIATION')
    .withBankId(BANK_ID_ONE)
    .withReportPeriod({
      start: { month: 1, year: 2021 },
      end: { month: 1, year: 2021 },
    })
    .build(),
    // ... repeated for each case
];
```

which I'm sure you can agree is much more bloated

ttbarnes (commented):
Can we add some E2E tests?

JoshBinns2000:
@ttbarnes We add e2e tests in separate PRs as they tend to be large and can often draw the focus away from the actual implementation. Similarly, this particular set of logic is very small and would be very hard to capture in an e2e test, and the benefit of having an e2e test for it would likely not make up for the added run time for all future PRs in the QA pipeline

dtfs-central-api/api-tests/v1/bank/banks-get.api-test.ts
abhi-markan: Can it be rather `monthlyUtilisationBankReport` or `monthlyUtlisationBankOneReport`?

JoshBinns2000: I have replaced it with the less clear but shorter `aUtilisationReportFor` helper - hopefully this is now more readable

trade-finance-manager-ui/scripts/enable-find-utilisation-reports-by-year-dropdown-autocomplete.ts
```diff
@@ -0,0 +1,29 @@
+const enableFindUtilisationReportsByYearDropdownAutocomplete = (): void => {
```
abhi-markan, line 1: Missing docstring.

trade-finance-manager-ui/scripts/enable-find-utilisation-reports-by-year-dropdown-autocomplete.ts
```diff
@@ -0,0 +1,29 @@
...
+  const findUtilisationReportsByYearForm: HTMLFormElement | null = document.querySelector('form#find-utilisation-reports-by-year--form');
+
+  if (!findUtilisationReportsByYearForm) {
+    return;
+  }
+
+  const yearInput: HTMLInputElement | null = document.querySelector('input#find-utilisation-reports-by-year--year-input');
+  if (!yearInput) {
```
abhi-markan, line 9: Can you please add line feed before LOC:9.

trade-finance-manager-ui/scripts/enable-find-utilisation-reports-by-year-dropdown-autocomplete.ts
```diff
@@ -0,0 +1,29 @@
+const enableFindUtilisationReportsByYearDropdownAutocomplete = (): void => {
+  const findUtilisationReportsByYearForm: HTMLFormElement | null = document.querySelector('form#find-utilisation-reports-by-year--form');
```
abhi-markan, line 2: Just had a look on other public scripts they seems to be referring to a DOM element using `data-cy` can the existing pattern please be used?

ttbarnes:
> @ttbarnes We add e2e tests in separate PRs as they tend to be large and can often draw the focus away from the actual implementation. Similarly, this particular set of logic is very small and would be very hard to capture in an e2e test, and the benefit of having an e2e test for it would likely not make up for the added run time for all future PRs in the QA pipeline

Appreciate this particular functionality might be tricky to capture in an E2E test, but I think we should try. IMO, everything as much as possible should have an E2E test so that we know if something has broken or not working as expected.

I don't think additional time in the CI/CD should be a concern, it can be optimised and improved by e.g splitting up tests so there are more runners. Having E2E test coverage at the cost of an additional few minutes in the pipeline (without optimisation) is definitely worth it.

There is a huge benefit for having a test for this. For example, in the future, if the new client side script's intergrity hash changes and the hash is not updated, the script will not run and the functionality will not work. This could then be picked up by QA in another environment and cause additional time/effort to raise, debug and fix. If there is an E2E test for it, this would be picked up in PR.

trade-finance-manager-ui/scripts/enable-find-utilisation-reports-by-year-dropdown-autocomplete.ts
```diff
@@ -0,0 +1,29 @@
...
+  }
+
+  const bankRadios: NodeListOf<HTMLInputElement> = findUtilisationReportsByYearForm.querySelectorAll('input.govuk-radios__input');
+
+  bankRadios.forEach((bankRadio) => {
+    const bankId = bankRadio.value;
+    const dataListId = `datalist--bankId-${bankId}`;
+
+    if (bankRadio.checked) {
+      yearInput.setAttribute('list', dataListId);
```
abhi-markan, line 20: Is `list ` custom data attribute, if yes then it should be `data-list` please https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes.

trade-finance-manager-ui/scripts/enable-find-utilisation-reports-by-year-dropdown-autocomplete.ts
```diff
@@ -0,0 +1,29 @@
+const enableFindUtilisationReportsByYearDropdownAutocomplete = (): void => {
+  const findUtilisationReportsByYearForm: HTMLFormElement | null = document.querySelector('form#find-utilisation-reports-by-year--form');
+
+  if (!findUtilisationReportsByYearForm) {
+    return;
+  }
+
+  const yearInput: HTMLInputElement | null = document.querySelector('input#find-utilisation-reports-by-year--year-input');
```
abhi-markan, line 8: Are these inputs part of utilisation report year form, wondering if yes then can we do `findUtilisationReportsByYearForm.querySelectorAll...`

trade-finance-manager-ui/templates/index.njk
abhi-markan: * The source file still needs to have comments.
* Having reviewed the file it seems to be adding bank id to year input unless I have misunderstood it can it be rather called either `addBankIdToYearInput` or something along those lines?

abhi-markan (changes requested):

trade-finance-manager-ui/scripts/enable-find-utilisation-reports-by-year-dropdown-autocomplete.ts
JoshBinns2000: There is only one input for the year, hence `querySelector` (not to mention that we want to differentiate between the radio inputs and this specific input)

trade-finance-manager-ui/scripts/enable-find-utilisation-reports-by-year-dropdown-autocomplete.ts
JoshBinns2000: It is not a custom data attribute. It is an attribute on `input` elements which allow it to link up with a `datalist` tag (this is builtin HTML 5 functionality)

trade-finance-manager-ui/scripts/enable-find-utilisation-reports-by-year-dropdown-autocomplete.ts
JoshBinns2000: Using `data-cy` in production is an antipattern as `data-cy` should only be used for testing

trade-finance-manager-ui/scripts/enable-find-utilisation-reports-by-year-dropdown-autocomplete.ts
JoshBinns2000: Could I ask why? I think it reads more clearly if you get the element and have a guard clause immediately after

trade-finance-manager-ui/templates/index.njk
JoshBinns2000: It does not add a bank id to the year input. This may be worth taking offline as it seems the missing context around this ticket is causing confusion

trade-finance-manager-ui/scripts/enable-find-utilisation-reports-by-year-dropdown-autocomplete.ts
JoshBinns2000: I also can't find any TFM-UI scripts which use the `data-cy` tag to access any elements

ChristianMcc: Just weighing in here as I asking for this to be an ID instead of `data-cy` initially. Cypress explicitly wants `data-cy` to only be for tests and not be referenced in either css or JS as per https://docs.cypress.io/guides/references/best-practices. I realise we've got an existing pattern of using it in portal at points but is keeping that pattern our preference over the cypress advice of how to use it?

JoshBinns2000: Further to this, [jquery best practices](https://www.freecodecamp.org/news/jquery-selectors) suggest using IDs as the primary way of selecting elements due to their high specificity (in contrast to other attributes such as `class` or `data-cy`)

>ID selectors have a high specificity making them difficult to override. Classes have a much lower specificity and are generally the preferred way to style elements in order to avoid specificity issues.

JoshBinns2000:
@ttbarnes I don't believe there exists e2e tests specifically for the other public scripts, so this is something that wouldn't be caught for existing scripts either way. Nevertheless, I am happy to try and add an e2e test, but as part of a separate PR (the scope of this PR has already grown beyond where it should be, and adding more code to review will slow development down futher)

trade-finance-manager-ui/scripts/enable-find-utilisation-reports-by-year-dropdown-autocomplete.ts
JoshBinns2000: Added and changed name

dtfs-central-api/api-tests/sql-db-helper.ts
JoshBinns2000: Going to resolve as beyond PR scope

JoshBinns2000:
@ttbarnes I found a way to add a minimal e2e test for this functionality - if you could have a look that would be great

JoshBinns2000:
Going to squash and rebase this PR to re-trigger the mend security check - sorry for any issues this may cause with respect to previous comments

dtfs-central-api/api-tests/sql-db-helper.ts
abhi-markan: Can we please add a TODO, then I am happy mark it as resolved. Thanks

trade-finance-manager-ui/templates/index.njk
abhi-markan: Renamed to `enableFindReportsByYearDropdown`.

trade-finance-manager-ui/scripts/enable-find-utilisation-reports-by-year-dropdown-autocomplete.ts
JoshBinns2000: Added

trade-finance-manager-ui/scripts/enable-find-utilisation-reports-by-year-dropdown-autocomplete.ts
JoshBinns2000: Going to leave this as is and update other form selection under [FN-3390](https://ukef-dtfs.atlassian.net/browse/FN-3390)

dtfs-central-api/api-tests/sql-db-helper.ts
JoshBinns2000: Added TODO to the actual test helper file referencing [FN-3389](https://ukef-dtfs.atlassian.net/browse/FN-3389)

abhi-markan (approved):
LGTM 👍🏼

PR #3400, opened 2024-08-20
fix(dtfs2-7313): add logging to identify remote only issue

AlexBramhill:
## Introduction :pencil2:
we have a bug creating facility amendments that only occurs on remote

add logging to middleware to identify the issue


abhi-markan (approved):
Approving with caveat these lines to be removed post troubleshooting.

PR #3405, opened 2024-08-21
feat(FN-3274): add script to migrate ACBS utilisation data to TFM

sw-mattw:
## Introduction :pencil2:
Added a utils data migration script to aid with migrating ACBS facility utilisation data to TFM. Reads a JSON file containing this data and inserts it into the SQL database.

- Added script which reads the provided JSON file and inserts this data in to the DB.
- Added README with usage instructions

Tested locally using the following JSON file, the resulting database change can be seen in the screenshot below that (includes all columns even though it's quite squashed):

```
[
  {
    "facilityId": "12",
    "utilisation": 876543.21,
    "fixedFeePremium": 1234.56
  },
  {
    "facilityId": "34",
    "utilisation": 76543.21,
    "fixedFeePremium": 234.56
  },
  {
    "facilityId": "56",
    "utilisation": 6543.21,
    "fixedFeePremium": 34.56
  },
  {
    "facilityId": "78",
    "utilisation": 543.21,
    "fixedFeePremium": 4.56
  }
]
```

and the report period constant:
```
{
  start: { month: 1, year: 2020 },
  end: { month: 3, year: 2020 },
}
```

Screenshot 2024-08-21 at 09 13 33


utils/data-migration/migrate-acbs-utilisation-data-to-tfm/index.ts
```diff
@@ -0,0 +1,61 @@
...
+  end: { month: 3, year: 2020 },
+};
+
+type UtilisationDataEntry = {
+  facilityId: string;
+  utilisation: number;
+  fixedFeePremium: number;
+};
+
+const validateReportPeriod = (reportPeriod: ReportPeriodPartialEntity): void => {
```
sw-mattw, line 20: Omitted basic month/year range validation from this function and only validated the slightly more complex case of end date < start date

utils/data-migration/migrate-acbs-utilisation-data-to-tfm/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+To modify the report period, update the `REPORT_PERIOD` constant at the top of the `index.ts` file with appropriate values.
+
+### JSON File Path
+
+To change the location of the input JSON file, modify the `JSON_FILE_PATH` constant at the top of the `index.ts` file.
+
+## Running the Script
+
+Follow these steps to run the migration script:
```
ChristianMcc, line 32: Nice readme, could we add one section to say add the .env file to wherever we need to with the sql DB connection parameters in case the user is running this in isolation of having run any of the other scripts in here please

sw-mattw: Ah yes, thanks for spotting, added a section on this and added a step linking to this

utils/data-migration/migrate-acbs-utilisation-data-to-tfm/index.ts
abhi-markan, line 20: Can we please add docstring for all the new functions.

utils/data-migration/migrate-acbs-utilisation-data-to-tfm/index.ts
```diff
@@ -0,0 +1,61 @@
...
+  const { start, end } = reportPeriod;
+  if (start.year > end.year || (start.year === end.year && start.month > end.month)) {
+    throw new Error('Report period end date cannot be before start date');
+  }
+};
+
+const insertFacilityUtilisationEntry = async (dataSource: DataSource, { facilityId, utilisation, fixedFeePremium }: UtilisationDataEntry) => {
+  const facilityUtilisationData = new FacilityUtilisationDataEntity();
```
abhi-markan, line 28: * Singleton pattern is not widely used in DTFS code base, is there a reason this pattern has been used.
* Could you also please refer to Codacy SCA alerts on unsafe member access please.

utils/data-migration/migrate-acbs-utilisation-data-to-tfm/README.md
abhi-markan: Good documentation indeed.

utils/data-migration/migrate-acbs-utilisation-data-to-tfm/README.md
```diff
@@ -0,0 +1,55 @@
+# ACBS Utilisation Data Migration to TFM
+
+This script migrates ACBS facility utilisation data to TFM by reading from a JSON file and inserting the data into the SQL database.
+
+## JSON Input File Structure
+
+The script expects a JSON file with the following structure:
+
+```
+[
```
abhi-markan, line 10: @sw-mattw could you please review Codacy alerts.

utils/data-migration/migrate-acbs-utilisation-data-to-tfm/index.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  try {
+    if (!existsSync(JSON_FILE_PATH)) {
+      throw new Error(`Migration JSON file not found at path: ${JSON_FILE_PATH}`);
+    }
+    const jsonData = readFileSync(JSON_FILE_PATH, 'utf8');
+
+    const utilisationData = JSON.parse(jsonData) as UtilisationDataEntry[];
+
+    await Promise.all(utilisationData.map((entry) => insertFacilityUtilisationEntry(dataSource, entry)));
```
ChristianMcc, line 51: I imagine the numbers of entries will only be in the low 200s - could you test this script against your local DB with a file with 250 entries please?
if it works then im more than happy with this and no changes needed - in future I'd probably map to the entities to get the full list of entities then call a single save to the database with a batch size of 100 (so 100 get saved at a time)

Sorry missed this on first review!

utils/data-migration/migrate-acbs-utilisation-data-to-tfm/index.ts
```diff
@@ -0,0 +1,61 @@
...
+
+type UtilisationDataEntry = {
+  facilityId: string;
+  utilisation: number;
+  fixedFeePremium: number;
+};
+
+const validateReportPeriod = (reportPeriod: ReportPeriodPartialEntity): void => {
+  const { start, end } = reportPeriod;
+  if (start.year > end.year || (start.year === end.year && start.month > end.month)) {
```
abhi-markan, line 22: Can this please be refactored to two conditional variables.

```javascript
const cond1 = ..
const cond2 = ..

if (cond1 || cond2) { ...
```

utils/data-migration/migrate-acbs-utilisation-data-to-tfm/index.ts
```diff
@@ -0,0 +1,61 @@
...
+  facilityUtilisationData.id = facilityId;
+  facilityUtilisationData.reportPeriod = REPORT_PERIOD;
+  facilityUtilisationData.utilisation = utilisation;
+  facilityUtilisationData.fixedFee = fixedFeePremium;
+  facilityUtilisationData.updateLastUpdatedBy({ platform: 'SYSTEM' });
+
+  await dataSource.manager.save(FacilityUtilisationDataEntity, facilityUtilisationData);
+};
+
+const run = async () => {
```
abhi-markan, line 38: Can the function name please be a bit more descriptive.

utils/data-migration/migrate-acbs-utilisation-data-to-tfm/index.ts
```diff
@@ -0,0 +1,61 @@
...
+  validateReportPeriod(REPORT_PERIOD);
+
+  const dataSource = await SqlDbDataSource.initialize();
+
+  try {
+    if (!existsSync(JSON_FILE_PATH)) {
+      throw new Error(`Migration JSON file not found at path: ${JSON_FILE_PATH}`);
```
abhi-markan, line 45: String substitution please.

utils/data-migration/migrate-acbs-utilisation-data-to-tfm/index.ts
```diff
@@ -0,0 +1,61 @@
...
+    }
+    const jsonData = readFileSync(JSON_FILE_PATH, 'utf8');
+
+    const utilisationData = JSON.parse(jsonData) as UtilisationDataEntry[];
+
+    await Promise.all(utilisationData.map((entry) => insertFacilityUtilisationEntry(dataSource, entry)));
+  } catch (error) {
+    console.error('Failed to migrate ACBS utilisation data to TFM:', error);
+  } finally {
+    await dataSource.destroy();
```
abhi-markan, line 55: Can this be in `catch()` block, happy for it to be in `finally()` just trying to understand if moving to `catch()` will have any implications.

utils/data-migration/migrate-acbs-utilisation-data-to-tfm/index.ts
```diff
@@ -0,0 +1,61 @@
...
+  }
+};
+
+(async () => {
```
abhi-markan, line 59: Can the execution of `run()` be controlled via a promise pattern please?

abhi-markan (changes requested):

utils/data-migration/migrate-acbs-utilisation-data-to-tfm/index.ts
ChristianMcc: I don't think this is a singleton pattern - we're instantiating a new FacilityUtilisationDataEntity every time we run this function rather than re-using a single entity.

Codacy errors on the readme look normal - on the typescript codacy has been broken for a long time - this FacilityUtilisationDataEntity has a type and is well typed and isn't any - but Codacy seems unable to pick this up properly and so complains. Think this is a wider problem with the Codacy setup unfortunately :(

utils/data-migration/migrate-acbs-utilisation-data-to-tfm/index.ts
sw-mattw: We need to close the DB connection regardless of whether we encounter an error or not

utils/data-migration/migrate-acbs-utilisation-data-to-tfm/index.ts
sw-mattw: Ran with 250 entries in the JSON, didn't get any errors. Happy to implement the batching if you'd prefer

Screenshot 2024-08-22 at 16 20 04

utils/data-migration/migrate-acbs-utilisation-data-to-tfm/index.ts
sw-mattw: Added to all the functions and the constants thanks

utils/data-migration/migrate-acbs-utilisation-data-to-tfm/index.ts
sw-mattw: Made this change, thanks for spotting forgot to pull these out

utils/data-migration/migrate-acbs-utilisation-data-to-tfm/index.ts
sw-mattw: I've changed the `console.error` statements below this to be substitution, I don't think we can do that on this line as we're throwing an exception

utils/data-migration/migrate-acbs-utilisation-data-to-tfm/README.md
sw-mattw: Seem to be resolved now thanks

abhi-markan: HEy @sw-mattw still I can see few on this README.md file 
image

utils/data-migration/migrate-acbs-utilisation-data-to-tfm/index.ts
abhi-markan: Thank you :-)

utils/data-migration/migrate-acbs-utilisation-data-to-tfm/index.ts
abhi-markan: Understood thanks.

utils/data-migration/migrate-acbs-utilisation-data-to-tfm/index.ts
```diff
@@ -0,0 +1,91 @@
...
+    if (!existsSync(JSON_FILE_PATH)) {
+      throw new Error(`Migration JSON file not found at path: ${JSON_FILE_PATH}`);
+    }
+    const jsonData = readFileSync(JSON_FILE_PATH, 'utf8');
+
+    const utilisationData = JSON.parse(jsonData) as UtilisationDataEntry[];
+
+    await Promise.all(utilisationData.map((entry) => insertFacilityUtilisationEntry(dataSource, entry)));
+  } catch (error) {
+    console.error('Failed to migrate ACBS utilisation data to TFM: %s', error);
```
abhi-markan, line 79: Can you please change this to `%o` object for verbose logging.

utils/data-migration/migrate-acbs-utilisation-data-to-tfm/index.ts
abhi-markan: Perfect, can you please reflect this in the docstring.

utils/data-migration/migrate-acbs-utilisation-data-to-tfm/README.md
```diff
@@ -0,0 +1,65 @@
...
+## JSON Input File Structure
+
+The script expects a JSON file with the following structure:
+
+```json
+[
+  {
+    "facilityId": "string",
+    "utilisation": number,
+    "fixedFeePremium": number
```
JoshBinns2000, line 15: Nit but because this is `json`, `number` is not valid - could we replace with example number literals?
```suggestion
    "utilisation": 100000,
    "fixedFeePremium": 1000
```

utils/data-migration/migrate-acbs-utilisation-data-to-tfm/index.ts
abhi-markan: * @ChristianMcc just noticed this pattern is already in use was implemented last month, if we are instantiating a class object I would have expected to use `setter` methods to update the individual properties perhaps we can have a chat about this approach so I understand the context.

* Some of these codacy errors are easily resolavable link to codacy suggestions https://app.codacy.com/gh/UK-Export-Finance/dtfs2/pull-requests/3405, of!course please ignore `import` related errors as they are genuine issue with codacy unable to refer to the deps. Happy to jump on a call regarding this as well if it helps.

abhi-markan (changes requested):

utils/data-migration/migrate-acbs-utilisation-data-to-tfm/index.ts
JoshBinns2000: Batching shouldn't be required here as we are running one insert per document :)

utils/data-migration/migrate-acbs-utilisation-data-to-tfm/index.ts
JoshBinns2000: This is not so much a pattern choice with our entity classes as much as it is something specific to typeorm - we can't start defining `set` and `get` methods as that will start to conflict with the way typeorm works (for example, you can't have `private` properties as the repository API needs to be able to access them). As a compromise, I think we should add a static `create` method to the `FacilityUtilisationData` class which constructs this entity with all the supplied parameters (see `FeeRecordEntity.create` for an example). This also gives us better future proofing if we ever add a property to the table and forget to update this script

utils/data-migration/migrate-acbs-utilisation-data-to-tfm/README.md
sw-mattw: Hmm weird, they weren't showing up inline. I've just reviewed the [latest report on Codacy itself](https://app.codacy.com/gh/UK-Export-Finance/dtfs2/pull-requests/3405/issues#issue-1c26324f42316ce74690e95e09d86af1) and all but one (excluding the `unsafe` issues) seem to be resolved. The one remaining issue is:

`Expected: [None]; Actual: # ACBS Utilisation Data Migration to TFM`

I've not seen this pattern of setting required headings before and not sure why it's not defaulting to an empty array. Is this warning okay to skip? if not, do you have any suggestions on how to resolve?

Thank you

utils/data-migration/migrate-acbs-utilisation-data-to-tfm/index.ts
sw-mattw: Added the create method Josh referenced as I quite like that pattern and it's on the other entities already, thanks!

libs/common/src/sql-db-entities/facility-utilisation-data/facility-utilisation-data.entity.ts
```diff
@@ -37,6 +41,16 @@ export class FacilityUtilisationDataEntity extends AuditableBaseEntity {
   @MonetaryColumn({ defaultValue: 0 })
   fixedFee!: number;
 
+  public static create({ id, reportPeriod, utilisation, fixedFee, requestSource }: CreateFacilityUtilisationDataParams): FacilityUtilisationDataEntity {
```
abhi-markan, line 44: Can you please add doc string on all the functions.

utils/data-migration/migrate-acbs-utilisation-data-to-tfm/index.ts
```diff
@@ -0,0 +1,93 @@
...
+
+    await Promise.all(utilisationData.map((entry) => insertFacilityUtilisationEntry(dataSource, entry)));
+  } catch (error) {
+    console.error('Failed to migrate ACBS utilisation data to TFM: %o', error);
+  } finally {
+    await dataSource.destroy();
+  }
+};
+
+runMigrationOfAcbsUtilisationDataToTfm()
```
abhi-markan, line 87: This is much better adhering to promise pattern. Thank you.

utils/data-migration/migrate-acbs-utilisation-data-to-tfm/index.ts
```diff
@@ -0,0 +1,93 @@
...
+const runMigrationOfAcbsUtilisationDataToTfm = async () => {
+  validateReportPeriod(REPORT_PERIOD);
+
+  const dataSource = await SqlDbDataSource.initialize();
+
+  try {
+    if (!existsSync(JSON_FILE_PATH)) {
+      throw new Error(`Migration JSON file not found at path: ${JSON_FILE_PATH}`);
+    }
+    const jsonData = readFileSync(JSON_FILE_PATH, 'utf8');
```
abhi-markan, line 75: One minor comment, a leed feed before LOC:75 (after ending bracket), I'll raise a PR to update the ESLint configuration. Thank you.

utils/data-migration/migrate-acbs-utilisation-data-to-tfm/index.ts
```diff
@@ -0,0 +1,93 @@
...
+ * @param utilisationDataEntry.utilisation - The utilisation value.
+ * @param utilisationDataEntry.fixedFeePremium - The fixed fee premium value.
+ */
+const insertFacilityUtilisationEntry = async (dataSource: DataSource, { facilityId, utilisation, fixedFeePremium }: UtilisationDataEntry) => {
+  const facilityUtilisationData = FacilityUtilisationDataEntity.create({
+    id: facilityId,
+    reportPeriod: REPORT_PERIOD,
+    utilisation,
+    fixedFee: fixedFeePremium,
+    requestSource: { platform: 'SYSTEM' },
```
abhi-markan, line 56: Can `SYSTEM` please be picked up from a constant, if not already exist can one please be created.

utils/data-migration/migrate-acbs-utilisation-data-to-tfm/index.ts
```diff
@@ -0,0 +1,93 @@
...
+  });
+
+  await dataSource.manager.save(FacilityUtilisationDataEntity, facilityUtilisationData);
```
abhi-markan, line 59: I would suggestion following improvements to this operation please:

1. Wrap it in a `try-catch` block.
2. `facilityUtilisationData` void check before execution a call (please ignore if static typing is in place already)

```typescript
  // Ensure `facilityUtilisationData` is not null or undefined
  if (!facilityUtilisationData) {
    throw new Error("facilityUtilisationData is null or undefined");
  }
  ```

utils/data-migration/migrate-acbs-utilisation-data-to-tfm/README.md
abhi-markan: Thanks for this @sw-mattw I do not see any issue with this as it is clearly not violating `MD043` MD syntax, however if you wish you can try following suggestions see if they work:

1. Reduce heading text
2. Add LF before heading (not ideal, just to test)
3. If neither works then I am happy for you to skip it.

Thank you!

abhi-markan (changes requested):

utils/data-migration/migrate-acbs-utilisation-data-to-tfm/index.ts
JoshBinns2000: I would actually argue that having a try-catch inside this function is redundant as it is expected that the caller of the function performs any error handling, and the redundancy of having an extra try-catch which re-throws the error provides no benefit.

Similarly, the guard clause for `facilityUtilisationData` is redundant for a few reasons:
1. `dataSource.manager.save` would itself throw if the supplied entity is `null` or `undefined` (which is then caught and handled by the caller)
2. Introducing redundant `null` or `undefined` check defeats the point of using typescript, especially when we're in a situation where we are not relying on things like the shape of an api response

utils/data-migration/migrate-acbs-utilisation-data-to-tfm/index.ts
JoshBinns2000: Just out of curiosity, what `Promise` pattern is this adhering to? As far as I know, it is not considered best practice to mix and match `.then` and `.catch` with `async`/`await` syntax due to the unexpected ways in which they can interact with each other, but maybe I am missing something?

utils/data-migration/migrate-acbs-utilisation-data-to-tfm/index.ts
sw-mattw: Just to check, is this a missing line break/blank line after the closing curly of the if statement? If so this should be resolved now thanks

utils/data-migration/migrate-acbs-utilisation-data-to-tfm/README.md
sw-mattw: Changed the heading text and tested adding a blank line before the heading but these didn't rectify the issue (see commits with the Codacy warning below)

Adding blank line - https://github.com/UK-Export-Finance/dtfs2/pull/3405/commits/afa4b58dd3b5f1cb647bdd2ad4f81e4376771542
Removing blank line - https://github.com/UK-Export-Finance/dtfs2/commit/ea79b107626d958d899f3dc5b85df97f6650859f

abhi-markan: Thank you @sw-mattw  for trying happy to mark this as resolved.

utils/data-migration/migrate-acbs-utilisation-data-to-tfm/index.ts
abhi-markan: DTFS code base widely uses two patterns generally:
* Middleware
* Promise

Promise pattern is a widely used for async calls as they are design to handle returned promise resolves or rejects. as it keeps the code simple (KISS), DRY, adheres to SOLID principles and most importantly helps de-couple crucial logics and exceptional handling.

There are other patterns which provide similar advantages but provided wide use of this pattern in DTFS code base, it makes sense to use it.

utils/data-migration/migrate-acbs-utilisation-data-to-tfm/index.ts
abhi-markan: 1. If we are propagating the exception to a higher level function which has a better context of the error message then  I am happy for `try-catch` block not to be introduced as long as this function scope remains the same.

2. `facilityUtilisationData` - I mentioned if static typing exists then it can be ignored but a good practice to limit operations by checking variables are not void as `dataSource.manager.save` throwing an exception because `facilityUtilisationData` is null still counts as an await operation which will trigger low-level event-driven architecture and if a call can be saved with argument validation it is a recommended approach.

abhi-markan (changes requested):

utils/data-migration/migrate-acbs-utilisation-data-to-tfm/index.ts
ChristianMcc: This is a script so I don't mind how this is written in this particular case but just wanted to check the above - in our codebase we almost entirely use async/await. E.g. in api.js files it's all async/await and then in our controllers (both old and new ones) we use async/await with async code not promises. 
Are there different cases where we are trying to use promises over the normal async/await we use in the core of our controller and API logic?

utils/data-migration/migrate-acbs-utilisation-data-to-tfm/index.ts
JoshBinns2000: The use of typescript typing means the null/undefined check isn't required, but I'm not sure I understand the latter part of what you're saying. As far as I know, the node JS runtime always runs using an event loop independent of whether or not `async` tasks are being executed. All `async` tasks do is place the function call on the microtask queue rather than immediately on the call stack, but the use of `await Promise.all` in the parent scope means that all of these `async` tasks will be immediately put on the call stack whenever the thread has down time. So, whether or not you introduce a guard clause, you will always be dealing with a series of tasks executed concurrently which utilise the event loop

sw-mattw: @abhi-markan static typing is already in place so I think (2) should be resolved, and happy with your explanation for (1), thanks

Screenshot 2024-08-28 at 08 42 12

utils/data-migration/migrate-acbs-utilisation-data-to-tfm/index.ts
sw-mattw: Pulled this out into a new constant and replaced references in this PR, other platform string literals to be updated under a new ticket (FN-3398)

utils/data-migration/migrate-acbs-utilisation-data-to-tfm/index.ts
abhi-markan: Thanks @sw-mattw, as discussed you will explicitly throw an exception in `create` method to check whether it will be caught and `save` method will not be invoked.

abhi-markan (approved):
Good work, LGTM 👍🏼

PR #3409, opened 2024-08-21
feat(DTFS2-7219): add bank review date page

MarRobSoftwire:
## Introduction :pencil2:
Users should be able to input a bank review date when issuing a facility

- add `bank review date` page to unissued facilities journey


gef-ui/server/controllers/bank-review-date/post-bank-review-date.ts
```diff
@@ -0,0 +1,146 @@
...
+    req,
+    res,
+    uris: {
+      nextPage: `/gef/application-details/${req.params.dealId}`,
+      saveAndReturn: `/gef/application-details/${req.params.dealId}`,
+      previousPage: `/gef/application-details/${req.params.dealId}/unissued-facilities/${req.params.facilityId}/change`,
+    },
+  });
+
+export const postBankReviewDateFromUnissuedFacilitiesPage = async (req: PostBankReviewDateRequest, res: Response) =>
```
AlexBramhill, line 126: look at this lovely pattern

e2e-tests/gef/cypress/e2e/facility/version-0-unissued-facility.spec.js
```diff
@@ -64,5 +64,17 @@ context('About unissued facility page', () => {
 
       cy.url().should('eq', relative(`/gef/application-details/${version0DealId}/unissued-facilities/${version0FacilityId}/change`));
     });
+
+    it('redirects to application details page when visiting bank review date page from initially issuing', () => {
```
AlexBramhill, line 68: is there a better phrase than `initially issuing` to describe this

e2e-tests/gef/cypress/e2e/facility/version-0-facility.spec.js
```diff
@@ -49,7 +49,7 @@ context('About Facility Page', () => {
     it('redirects to application details page when visiting bank review date page', () => {
```
AlexBramhill, line 49: test title out of sync

e2e-tests/gef/cypress/e2e/facility/bank-review-date.spec.js
```diff
@@ -238,13 +238,13 @@ context('Bank Review Date Page', () => {
 
       cy.visit(relative(`/gef/application-details/${application.id}/facilities/${facilityId}/bank-review-date`));
 
-      cy.url().should('eq', relative(`/gef/application-details/${application.id}`));
+      cy.url().should('eq', relative(`/gef/application-details/${application.id}/facilities/${facilityId}/about-facility`));
```
AlexBramhill, line 241: test title out of sync

gef-ui/server/types/view-models/view-model-errors.ts
```diff
@@ -0,0 +1,4 @@
+export type ViewModelErrors = {
+  errorSummary: { text: string; href: string }[];
```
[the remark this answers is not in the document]
MarRobSoftwire: ya this is the existing pattern in gef-ui but agree. I guess in the context of the `errors` object it makes sense but :/

e2e-tests/gef/cypress/e2e/facility/unissued-facilities/unissued-facilities-bank-review-date.spec.js
```diff
@@ -0,0 +1,159 @@
...
+      cy.saveSession();
+      cy.login(BANK1_MAKER1);
+    });
+
+    describe('Visiting from unissued facilities page', () => {
+      beforeEach(() => {
+        cy.visit(relative(`/gef/application-details/${dealId}/unissued-facilities`));
+        unissuedFacilityTable.updateIndividualFacilityButton(0).click();
+
+        aboutFacilityUnissued.issueDateDay().type(dateConstants.threeDaysDay);
```
Zainzzkk, line 62: can these be de-referenced please?

e2e-tests/gef/cypress/e2e/facility/unissued-facilities/unissued-facilities-bank-review-date.spec.js
```diff
@@ -0,0 +1,159 @@
...
+        aboutFacilityUnissued.coverEndDateMonth().type(dateConstants.threeMonthsOneDayMonth);
+        aboutFacilityUnissued.coverEndDateYear().type(dateConstants.threeMonthsOneDayYear);
+
+        aboutFacilityUnissued.isUsingFacilityEndDateNo().click();
+
+        aboutFacilityUnissued.continueButton().click();
+      });
+
+      it('should redirect user to the about unissued facility page when clicking back link', () => {
+        cy.url().should('eq', relative(`/gef/application-details/${dealId}/unissued-facilities/${facilityOneId}/bank-review-date`));
```
Zainzzkk, line 81: can all the urls be lets and populated in the beforeEach please? Easier to read and some are repeated more than once

e2e-tests/gef/cypress/e2e/facility/unissued-facilities/unissued-facilities-bank-review-date.spec.js
```diff
@@ -0,0 +1,159 @@
...
+        aboutFacilityUnissued.coverEndDateDay().type(dateConstants.threeMonthsOneDayDay);
+        aboutFacilityUnissued.coverEndDateMonth().type(dateConstants.threeMonthsOneDayMonth);
+        aboutFacilityUnissued.coverEndDateYear().type(dateConstants.threeMonthsOneDayYear);
+
+        aboutFacilityUnissued.isUsingFacilityEndDateNo().click();
+
+        aboutFacilityUnissued.continueButton().click();
+      });
+
+      it('should redirect user to the about unissued facility page when clicking back link', () => {
```
Zainzzkk, line 80: this could be `should redirect user to ${aboutUnissuedFacilityUrl} when clicking the back link`

e2e-tests/gef/cypress/e2e/facility/unissued-facilities/unissued-facilities-bank-review-date.spec.js
```diff
@@ -0,0 +1,159 @@
...
+        cy.url().should('eq', relative(`/gef/application-details/${dealId}/unissued-facilities/${facilityOneId}/bank-review-date`));
+
+        bankReviewDate.backLink().click();
+
+        cy.url().should('eq', relative(`/gef/application-details/${dealId}/unissued-facilities/${facilityOneId}/about`));
+      });
+
+      it('should display error messages', () => {
```
Zainzzkk, line 88: this could go in a describe explaining why there are error messages, eg `describe('when the date is entered in the incorrect format', () => {...`

e2e-tests/gef/cypress/e2e/facility/unissued-facilities/unissued-facilities-bank-review-date.spec.js
```diff
@@ -0,0 +1,159 @@
...
+
+        bankReviewDate.continueButton().click();
+
+        cy.url().should('eq', relative(`/gef/application-details/${dealId}/unissued-facilities`));
+      });
+
+      it('should redirect user to the unissued facility page when clicking save and return', () => {
+        bankReviewDate.bankReviewDateDay().clear().type(dateConstants.todayDay);
+        bankReviewDate.bankReviewDateMonth().clear().type(dateConstants.todayMonth);
+        bankReviewDate.bankReviewDateYear().clear().type(dateConstants.todayYear);
```
Zainzzkk, line 109: these could go in a form-submission describe and have the 3 repeated lines in a beforeEach

e2e-tests/gef/cypress/e2e/facility/unissued-facilities/unissued-facilities-bank-review-date.spec.js
```diff
@@ -0,0 +1,159 @@
...
+
+      it('should redirect user to the about unissued facility page when clicking back link', () => {
+        cy.url().should('eq', relative(`/gef/application-details/${dealId}/unissued-facilities/${facilityOneId}/bank-review-date/change`));
+
+        bankReviewDate.backLink().click();
+
+        cy.url().should('eq', relative(`/gef/application-details/${dealId}/unissued-facilities/${facilityOneId}/change`));
+      });
+
+      it('should display error messages', () => {
```
Zainzzkk, line 131: same as above

e2e-tests/gef/cypress/e2e/facility/unissued-facilities/unissued-facilities-bank-review-date.spec.js
```diff
@@ -0,0 +1,159 @@
...
+    });
+
+    describe('Visiting from application preview page', () => {
+      beforeEach(() => {
+        cy.visit(relative(`/gef/application-details/${dealId}`));
+        applicationPreview.facilitySummaryListTable(4).bankReviewDateAction().click();
+      });
+
+      it('should redirect user to the about unissued facility page when clicking back link', () => {
+        cy.url().should('eq', relative(`/gef/application-details/${dealId}/unissued-facilities/${facilityOneId}/bank-review-date/change`));
```
Zainzzkk, line 124: this can go in the beforeEach

e2e-tests/gef/cypress/e2e/facility/unissued-facilities/unissued-facilities-change-all-MIN.spec.js
```diff
@@ -402,6 +410,13 @@ context('Unissued Facilities MIN - change all to issued from unissued table', ()
 
       aboutFacilityUnissued.continueButton().click();
 
+      if (facilityEndDateEnabled) {
+        bankReviewDate.bankReviewDateDay().clear().type(dateConstants.threeMonthsDay);
```
Zainzzkk, line 414: this could be a cypress command, it is repeated

e2e-tests/gef/cypress/e2e/facility/version-0-facility.spec.js
```diff
@@ -46,10 +46,10 @@ context('About Facility Page', () => {
       cy.url().should('eq', relative(`/gef/application-details/${version0DealId}/facilities/${version0FacilityId}/about-facility`));
     });
 
-    it('redirects to application details page when visiting bank review date page', () => {
+    it('redirects to about facility page when visiting bank review date page', () => {
```
Zainzzkk, line 49: to the

e2e-tests/gef/cypress/e2e/facility/version-0-unissued-facility.spec.js
```diff
@@ -53,16 +53,28 @@ context('About unissued facility page', () => {
       aboutFacilityUnissued.isUsingFacilityEndDateNo().should('not.exist');
     });
 
-    it('redirects to application details page when visiting facility end date page from initially issuing', () => {
+    it('redirects to about unissued facility page when visiting facility end date page from unissued facility page', () => {
```
Zainzzkk, line 56: this it is quite long, could have a describe before

gef-ui/api-tests/bank-review-date.get.api-test.js
```diff
@@ -52,56 +70,55 @@ describe('bank review date routes', () => {
...
+        const response = await getWithSessionCookie(url, sessionCookie);
 
         // Assert
         expect(response.status).toBe(302);
-        expect(response.headers.location).toBe(`/gef/application-details/${dealId}`);
+        expect(response.headers.location).toBe(previousPageUrl);
       });
 
-      it('redirects the user to the application details page if version is 1 & isUsingFacilityEndDate is null ', async () => {
+      it('redirects the user to the previous page if version is 1 & isUsingFacilityEndDate is null ', async () => {
```
Zainzzkk, line 98: could use more describes so shorter it statements

libs/common/src/constants/facility-type.ts
```diff
@@ -1,6 +1,11 @@
-export const FACILITY_TYPE = {
-  BOND: 'Bond',
-  LOAN: 'Loan',
+export const GEF_FACILITY_TYPE = {
   CASH: 'Cash',
   CONTINGENT: 'Contingent',
 } as const;
+
+export const BSS_EWCS_FACILITY_TYPE = {
```
Zainzzkk, line 6: could we use types instead please?

libs/common/src/constants/deal-type.ts
```diff
@@ -0,0 +1,4 @@
+export const DEAL_TYPE = {
```
Zainzzkk, line 1: could we use types instead please?

gef-ui/server/utils/helpers.js
```diff
@@ -66,7 +66,7 @@ const ErrorMessagesMap = {
  * Maps validation errors so both Summary Error component and field component display the error messages correctly.
  * @param {import('../types/validation-error').ValidationError[] | import('../types/validation-error').ValidationError} errs - errors to be mapped
  * @param {string} href - the current page URL
- * @returns {import('../types/view-models/facility-end-date-view-model').ViewModelErrors | null}
+ * @returns {import('../types/view-models/view-model-errors').ViewModelErrors | null}
```
Zainzzkk, line 69: could this be ViewModelErrors |  null (whatever the type is in the file being referenced)

gef-ui/server/services/api.js
```diff
@@ -137,6 +143,12 @@ const createFacility = async ({ payload, userToken }) => {
   }
 };
 
+/**
+ * @param {object} param
+ * @param {string} param.facilityId
+ * @param {string} param.userToken
+ * @returns {Promise<{ details: import('../types/facility').Facility }>}
```
Zainzzkk, line 150: this can be Promise<Facility> .  Same for the one above

gef-ui/server/controllers/unissued-facilities/index.js
```diff
@@ -328,13 +329,14 @@ const postChangeUnissuedFacilityPreview = async (req, res) => {
...
     const aboutFacilityRedirectUri = `/gef/application-details/${dealId}`;
 
-    if (isUsingFacilityEndDate === true) {
-      return res.redirect(facilityEndDateRedirectUri);
+    if (isUsingFacilityEndDate === undefined) {
+      return res.redirect(aboutFacilityRedirectUri);
     }
 
-    return res.redirect(aboutFacilityRedirectUri);
+    return res.redirect(isUsingFacilityEndDate ? facilityEndDateRedirectUri : bankReviewDateRedirectUri);
```
Zainzzkk, line 339: could this be if statements? this is quite difficult to read

gef-ui/server/controllers/unissued-facilities/index.js
```diff
@@ -230,13 +230,14 @@ const postChangeUnissuedFacility = async (req, res) => {
...
     const unissuedFacilitiesRedirectUri = `/gef/application-details/${dealId}/unissued-facilities`;
 
-    if (isUsingFacilityEndDate === true) {
-      return res.redirect(facilityEndDateRedirectUri);
+    if (isUsingFacilityEndDate === undefined) {
+      return res.redirect(unissuedFacilitiesRedirectUri);
     }
 
-    return res.redirect(unissuedFacilitiesRedirectUri);
+    return res.redirect(isUsingFacilityEndDate ? facilityEndDateRedirectUri : bankReviewDateRedirectUri);
```
Zainzzkk, line 240: same here, this is difficult to read

gef-ui/server/controllers/unissued-facilities/index.js
```diff
@@ -230,13 +230,14 @@ const postChangeUnissuedFacility = async (req, res) => {
     });
 
     const facilityEndDateRedirectUri = `/gef/application-details/${dealId}/unissued-facilities/${facilityId}/facility-end-date`;
+    const bankReviewDateRedirectUri = `/gef/application-details/${dealId}/unissued-facilities/${facilityId}/bank-review-date`;
     const unissuedFacilitiesRedirectUri = `/gef/application-details/${dealId}/unissued-facilities`;
 
-    if (isUsingFacilityEndDate === true) {
-      return res.redirect(facilityEndDateRedirectUri);
+    if (isUsingFacilityEndDate === undefined) {
```
Zainzzkk, line 236: could this not be if (!isUsingFacilityEndDate)

gef-ui/server/controllers/unissued-facilities/index.js
Zainzzkk, line 339: have these changes been tested?

gef-ui/server/controllers/facility-end-date/post-facility-end-date.ts
```diff
@@ -36,74 +30,14 @@ type HandlePostFacilityEndDateParams = {
...
-
-export const getFacilityEndDateFromApplicationDetailsPage = async (req: GetFacilityEndDateRequest, res: Response) =>
-  handleGetFacilityEndDate({ req, res, previousPage: `/gef/application-details/${req.params.dealId}/facilities/${req.params.facilityId}/about-facility` });
-
-const updateFacilityEndDateIfChanged = async (
-  existingFacility: Record<string, unknown>,
-  facilityEndDate: Date,
-  { userToken, user }: LoggedInUserSession,
-): Promise<void> => {
+const updateFacilityEndDateIfChanged = async (existingFacility: Facility, facilityEndDate: Date, { userToken, user }: LoggedInUserSession): Promise<void> => {
```
Zainzzkk, line 33: could this just be updateFacilityEndDate? And documentation?

gef-ui/server/controllers/facility-end-date/get-facility-end-date.ts
```diff
@@ -0,0 +1,74 @@
...
+      day: String(getDate(facilityEndDate)),
+      month: String(getMonth(facilityEndDate) + 1),
+      year: String(getYear(facilityEndDate)),
+    };
+  }
+
+  return facilityEndDateViewModel;
+};
+
+const handleGetFacilityEndDate = async ({ req, res, previousPage }: HandleGetFacilityEndDateParams) => {
```
Zainzzkk, line 43: could this be getFacilityEndDate?

gef-ui/server/controllers/facility-end-date/get-facility-end-date.ts
```diff
@@ -0,0 +1,74 @@
...
+    params: { dealId, facilityId },
+    query: { status },
+  } = req;
+  const { userToken } = asLoggedInUserSession(req.session);
+
+  try {
+    const { details: facility } = await api.getFacility({ facilityId, userToken });
+    const deal = await api.getApplication({ dealId, userToken });
+
+    const shouldRedirectFromPage = !isFacilityEndDateEnabledOnGefVersion(parseDealVersion(deal.version)) || !facility.isUsingFacilityEndDate;
```
Zainzzkk, line 54: could parseDealVersion(deal.version) be a const? or could it be added to isFacilityEndDateEnabledOnGefVersion

gef-ui/server/controllers/facility-end-date/get-facility-end-date.ts
```diff
@@ -0,0 +1,74 @@
...
+
+    if (shouldRedirectFromPage) {
+      return res.redirect(previousPage);
+    }
+
+    return res.render('partials/facility-end-date.njk', getFacilityEndDateViewModel(facility, previousPage, status));
+  } catch (error) {
+    console.error(error);
```
Zainzzkk, line 62: could a desription be added to this error? including the function name

gef-ui/server/controllers/bank-review-date/post-bank-review-date.ts
```diff
@@ -0,0 +1,143 @@
...
+    userToken,
+  });
+
+  const applicationUpdate = {
+    editorId: user._id,
+  };
+  await api.updateApplication({ dealId: existingFacility.dealId, application: applicationUpdate, userToken });
+};
+
+const handlePostBankReviewDate = async ({ req, res, uris }: HandlePostBankReviewDateParams) => {
```
Zainzzkk, line 55: could this just be postBankReviewDate

gef-ui/server/controllers/facility-end-date/get-facility-end-date.ts
Zainzzkk, line 1: These files are too long and complicated.  Can this be split up
Params could go in a types file
The main work could be done by a helper
Which would make this much less complicated, where only the post controllers are in this file

gef-ui/api-tests/facility-end-date.post.api-test.js
```diff
@@ -192,6 +192,8 @@ describe('facility end date routes', () => {
           // Assert
           expect(response.status).toBe(HttpStatusCode.Found);
           expect(response.headers.location).toBe(saveAndReturnRedirectUrl);
+
+          expect(api.updateFacility).toHaveBeenCalledTimes(0);
```
Zainzzkk, line 196: this should have its own it statement

e2e-tests/gef/cypress/e2e/pages/bank-review-date.js
```diff
@@ -9,6 +11,16 @@ const bankReviewDate = {
...
+
+  /**
+   * Fill in the bank review date form
+   * @param {Date} date
+   */
+  fillInBankReviewDate: (date) => {
+    this.bankReviewDateDay().clear().type(format(date, 'd'));
+    this.bankReviewDateMonth().clear().type(format(date, 'M'));
+    this.bankReviewDateYear().clear().type(format(date, 'yyyy'));
+  },
```
ttbarnes, line 23: Is this pattern (putting functions into `page` selectors), something we have in other places?

I'm not a fan of this because:

1) Page selectors are no longer pure selectors
2) Individual tests have to import the page in order to call the function
3) Makes it difficult to move to another place in the future

Unless this is a pattern in other places, could we move this to a non-page specific location, i.e a standalone cypress command? So it can be called like this for example:

```js
cy.completeBankReviewDate();
```

"fillIn" or "complete" or something else is up for debate I think.

ttbarnes, line 23: Is this pattern (putting functions into `page` selectors), something we have in other places?

I'm not a fan of this because:

1) Page selectors are no longer pure selectors
2) Individual tests have to import the page in order to call the function
3) Makes it difficult to move to another place in the future

Unless this is a pattern in other places, could we move this to a non-page specific location, i.e a standalone cypress command? So it can be called like this for example:

```js
cy.completeBankReviewDate();
```

"fillIn" or "complete" or something else is up for debate I think.

ttbarnes (changes requested):

e2e-tests/gef/cypress/e2e/pages/bank-review-date.js
MarRobSoftwire: I'd seen reusable commands requested on one of Beths PRs & this was how she resolved it (see navigateThroughFacilityEndDateAmendmentPages in e2e-tests\tfm\cypress\e2e\pages\amendments\amendmentsPage.js). I agree that these pages files should be pure & only contain selectors, but I'm not much more of a fan of commands on the root level cy object. Will implement for now, but think a nicer pattern we could adopt would be importing these from a page-specific helpers file

ttbarnes: It's an interesting topic @MarRobSoftwire - the problem with having such functions in a page file is that, it's no longer "a page", if that makes sense. Also, it means anytime you want to run such a function, you have to import it. There could be 5-10+ more instances in an E2E suite where we want to e.g complete and submit X form, having this as a root command would be cleaner and DRYer

It's open for discussion, there's a few ways to do such things but I think top level cy commands are the cleanest way to go, as long as they're named and structured well

e2e-tests/gef/cypress/e2e/facility/unissued-facilities/unissued-facilities-bank-review-date.spec.js
Zainzzkk: sorry missed this.  it is better to dereference if it is repeated a few times

e2e-tests/gef/cypress/e2e/facility/unissued-facilities/unissued-facilities-bank-review-date.spec.js
Zainzzkk: true, can you put about unissued facility page in " " please

gef-ui/server/controllers/bank-review-date/post-bank-review-date.ts
```diff
@@ -0,0 +1,161 @@
...
+  nextPage: string;
+  previousPage: string;
+  saveAndReturn: string;
+};
+
+type PostBankReviewDateParams = {
+  req: PostBankReviewDateRequest;
+  res: Response;
+  uris: PostRequestUris;
+};
```
ttbarnes, line 31: I can see there are duplicates of the above 5x types, can we commonise these / be DRY?

gef-ui/server/controllers/bank-review-date/post-bank-review-date.ts
```diff
@@ -0,0 +1,161 @@
...
+    facilityId: existingFacility._id,
+    payload: {
+      bankReviewDate,
+    },
+    userToken,
+  });
+
+  const applicationUpdate = {
+    editorId: user._id,
+  };
```
ttbarnes, line 57: Can we add an empty line after this please

gef-ui/server/controllers/bank-review-date/post-bank-review-date.ts
```diff
@@ -0,0 +1,161 @@
...
+    const bankReviewDateIsBlank = !bankReviewDateYear && !bankReviewDateMonth && !bankReviewDateDay;
+
+    // If the user clicks save and return with no values filled in, we do not update the database
+    if (isTrueSet(saveAndReturn) && bankReviewDateIsBlank) {
+      return res.redirect(uris.saveAndReturn);
+    }
+
+    const { details: facility } = await api.getFacility({ facilityId, userToken });
+
+    const bankReviewDateErrorsAndDate = validateAndParseBankReviewDate(
```
ttbarnes, line 83: Could/should we destructure the results here, thinking maybe:

```ts
const { validationErrors, submittedDate } = validateAndParseBankReviewDate()
```

Would simplify some instances below

gef-ui/server/controllers/facility-end-date/get-facility-end-date.ts
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  const facilityEndDateViewModel: FacilityEndDateViewModel = {
+    dealId: facility.dealId,
+    facilityId: facility._id,
+    previousPage,
+    status,
+  };
+
+  if (typeof facility.facilityEndDate === 'string') {
```
ttbarnes, line 37: Same comments in this file as above for the `get-bank-review-date` controller

gef-ui/server/controllers/unissued-facilities/index.js
ttbarnes: I agree this is a little odd, could we do this instead @MarRobSoftwire? Reversing the condition, also simplifies the conditional URLs. Something like this


```js
if (isUsingFacilityEndDate) {
  return res.redirect(facilityEndDateRedirectUri);
}

// is undefined
return res.redirect(bankReviewDateRedirectUri);

File: gef-ui/server/controllers/unissued-facilities/index.js (hunk 22 above)
Comment (ttbarnes) [reply gef-ui/server/controllers/unissued-facilities/index.js:240]: My suggestion above would simplify this 😎

File: gef-ui/server/controllers/unissued-facilities/index.js (hunk 21 above)
Comment (ttbarnes) [reply gef-ui/server/controllers/unissued-facilities/index.js:339]: Same comment/suggestion as above

If we don't want to use if statements can we move the ternary into it's own `redirectUrl` const, so that what's called with `res.redirect` is simpler

File: gef-ui/server/types/deal.ts [hunk 35]
```diff
@@ -0,0 +1,23 @@
...
+  bank: AnyObject;
+  exporter: AnyObject;
+  bankInternalRefName: string | null;
+  createdAt: IsoDateTimeStamp;
+  updatedAt: IsoDateTimeStamp;
+  submissionCount: number;
+  supportingInformation: AnyObject;
+  editedBy: string[];
+  additionalRefName: string | null;
+};
```
ttbarnes, line 23: Wondering if this should be in common instead of specific to GEF UI?

Does TFM have a similar GEF deal types?

gef-ui/server/types/facility.ts
```diff
@@ -0,0 +1,36 @@
...
+  ukefFacilityId?: string;
+  dayCountBasis?: number | null;
+  specialIssuePermission: AnyObject | null;
+  canResubmitIssuedFacilities?: boolean | null;
+  unissuedToIssuedByMaker?: AnyObject | null;
+  // The following are only found on facilities from version 1 or higher deals
+  isUsingFacilityEndDate?: boolean | null;
+  facilityEndDate?: IsoDateTimeStamp | null;
+  bankReviewDate?: IsoDateTimeStamp | null;
+};
```
ttbarnes, line 36: Same as above RE common (potentially?)

gef-ui/server/types/view-models/bank-review-date-view-model.ts
```diff
@@ -0,0 +1,10 @@
+import { ViewModelErrors } from './view-model-errors';
+
+export type BankReviewDateViewModel = {
+  dealId: string;
+  facilityId: string;
+  bankReviewDate?: { day: string; month: string; year: string };
+  previousPage: string;
+  status?: string;
+  errors?: ViewModelErrors | null;
```
ttbarnes, line 9: I don't think we need `null` if errors is optional?

gef-ui/server/types/view-models/view-model-errors.ts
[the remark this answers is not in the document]
ttbarnes: I think error summaries should keep `text` to this as it makes for simpler consumption with the GOV component - a lot of the GOV components have `text` as a standard

gef-ui/server/utils/get-cover-start-date-or-start-of-today.ts
```diff
@@ -1,7 +1,7 @@
 import { parseISO, startOfDay } from 'date-fns';
-import { AnyObject } from '@ukef/dtfs2-common';
+import { Facility } from '../types/facility';
 
-export const getCoverStartDateOrStartOfToday = (facility: AnyObject): Date => {
+export const getCoverStartDateOrStartOfToday = (facility: Facility): Date => {
```
ttbarnes, line 4: Looks like this file doesn't have a unit test 😞 

Can we create a tech debt ticket to add a test?

ttbarnes (changes requested):
Nice work

e2e-tests/gef/cypress/e2e/facility/unissued-facilities/unissued-facilities-change-all-MIN.spec.js
```diff
@@ -295,6 +296,11 @@ context('Unissued Facilities MIN - change all to issued from unissued table', ()
 
       aboutFacilityUnissued.continueButton().click();
 
+      if (facilityEndDateEnabled) {
```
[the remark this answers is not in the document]
MarRobSoftwire: I've got a PR to update this separately but need this merged in first :) (https://github.com/UK-Export-Finance/dtfs2/pull/3433/)

gef-ui/server/controllers/bank-review-date/get-bank-review-date.ts
```diff
@@ -0,0 +1,92 @@
...
+/**
+ * @param facility - the facility
+ * @param previousPage - the previous page url
+ * @param status - the query parameter status
+ * @returns view model for the bank review date template
+ */
+const getBankReviewDateViewModel = (facility: Facility, previousPage: string, status: string | undefined): BankReviewDateViewModel => {
+  if (typeof facility.dealId !== 'string' || typeof facility._id !== 'string') {
+    throw new Error('Invalid facility or deal id provided');
+  }
```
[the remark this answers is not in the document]
MarRobSoftwire: Oops, this was added before I added the `Facility` type. It all boils down to whether we trust the api response!

gef-ui/server/controllers/bank-review-date/get-bank-review-date.ts
```diff
@@ -0,0 +1,92 @@
...
+
+  const bankReviewDateViewModel: BankReviewDateViewModel = {
+    dealId: facility.dealId,
+    facilityId: facility._id,
+    previousPage,
+    status,
+  };
+
+  if (typeof facility.bankReviewDate === 'string') {
```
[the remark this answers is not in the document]
MarRobSoftwire: Yep this is needed for TS because the `BankReviewDateViewModel` needs this to be a string and `getDate` returns a number

gef-ui/server/controllers/bank-review-date/post-bank-review-date.ts
MarRobSoftwire: These are different depending on being bank review date/ facility end date and post/get requests

gef-ui/server/controllers/bank-review-date/post-bank-review-date.ts
MarRobSoftwire: `validateAndParseBankReviewDate` returns a `ErrorsOrDate` object - we have to check if the `errors` object exists in the return value before we can use it

gef-ui/server/types/deal.ts
MarRobSoftwire: No this is based on what portal-api sends and is not a commonisable type

gef-ui/server/utils/get-cover-start-date-or-start-of-today.ts
MarRobSoftwire: Have added since it was added by my last PR

gef-ui/server/controllers/bank-review-date/post-bank-review-date.ts
ttbarnes: Sorry didn't notice - maybe could create a new type and extend it? Thinking something like this:

```ts
interface aNewType {
  res: Response;
  uris: PostRequestUris;  
}

interface HandlePostFacilityEndDateParams extends aNewType {
  req: PostFacilityEndDateRequest;  
}
```

Or would this be overkill / future PR ?

gef-ui/server/controllers/bank-review-date/post-bank-review-date.ts
ttbarnes: Interesting approach - not a blocker since it's an established pattern, I think it could be simpler 😊

e2e-tests/e2e-fixtures/dateConstants.js
Zainzzkk, line 1: great work in this file

e2e-tests/gef/cypress/e2e/facility/bank-review-date.spec.js
```diff
@@ -215,19 +193,17 @@ context('Bank Review Date Page', () => {
...
-      bankReviewDate.bankReviewDateYear().clear().type(nextYear);
+      cy.fillInBankReviewDate(twoYears);
 
       cy.clickContinueButton();
 
       cy.visit(relative(`/gef/application-details/${application.id}/facilities/${facilityId}/bank-review-date`));
-      bankReviewDate.bankReviewDateDay().should('have.value', today.getDate());
-      bankReviewDate.bankReviewDateMonth().should('have.value', today.getMonth() + 1);
-      bankReviewDate.bankReviewDateYear().should('have.value', today.getFullYear() + 1);
+      bankReviewDate.bankReviewDateDay().should('have.value', twoYears.getDate());
```
Zainzzkk, line 201: for the future, lets look to have a shared cypress command to check the dates

e2e-tests/gef/cypress/e2e/facility/unissued-facilities/unissued-facilities-facility-end-date.spec.js
```diff
@@ -0,0 +1,160 @@
...
+      it('should redirect user to the application details page when clicking continue', () => {
+        facilityEndDate.facilityEndDateDay().clear().type(dateConstants.todayDay);
+        facilityEndDate.facilityEndDateMonth().clear().type(dateConstants.todayMonth);
+        facilityEndDate.facilityEndDateYear().clear().type(dateConstants.todayYear);
+        cy.clickContinueButton();
+
+        cy.url().should('eq', relative(`/gef/application-details/${dealId}`));
+      });
+
+      it('should redirect user to the application details page when clicking save and return', () => {
```
Zainzzkk, line 149: just to check, do we have any tests for testing data has been saved when coming back to the page? And also, what happens if data is entered which has validation errors and save and return is clicked?

MarRobSoftwire: Theres a more detailed api test to cover saveAndReturn behaviour -  `gef-ui\api-tests\facility-end-date.post.api-test.js` & unit tests to cover validation in more detail `gef-ui\server\controllers\facility-end-date\validation.test.ts`.
I'll add an e2e test that the data is saved on returning

gef-ui/server/controllers/bank-review-date/get-bank-review-date.ts
```diff
@@ -0,0 +1,87 @@
...
+    };
+  }
+
+  return bankReviewDateViewModel;
+};
+
+/**
+ * Handle get bank review date requests
+ */
+const getBankReviewDate = async ({ req, res, previousPage }: HandleGetBankReviewDateParams) => {
```
Zainzzkk, line 46: has this function and the one above been tested?

e2e-tests/gef/cypress/e2e/facility/unissued-facilities/unissued-facilities-facility-end-date.spec.js
Zainzzkk: thanks and it would be good to also have an e2e test to ensure that behaviour on save and return with validation errors occurs

gef-ui/server/controllers/bank-review-date/post-bank-review-date.ts
ttbarnes: bump

gef-ui/server/types/deal.ts
ttbarnes, line 23: Could `bankInternalRefName` and `additionalRefName be the following?:

```ts
bankInternalRefName?: string;
additionalRefName?: string;
```

gef-ui/server/types/facility.ts
```diff
@@ -0,0 +1,36 @@
...
+  updatedAt?: IsoDateTimeStamp;
+  ukefFacilityId?: string;
+  dayCountBasis?: number | null;
+  specialIssuePermission: AnyObject | null;
+  canResubmitIssuedFacilities?: boolean | null;
+  unissuedToIssuedByMaker?: AnyObject | null;
+  // The following are only found on facilities from version 1 or higher deals
+  isUsingFacilityEndDate?: boolean | null;
+  facilityEndDate?: IsoDateTimeStamp | null;
+  bankReviewDate?: IsoDateTimeStamp | null;
```
ttbarnes, line 35: Same as above for all `... | null` instances here?

ttbarnes (changes requested):

gef-ui/server/types/deal.ts
MarRobSoftwire: `?` means it can be `undefined`. But these are always `string | null` (based on the `Facility` class in portal-api)

ttbarnes: Got it, thanks. Not for this PR, but ideally could look into updating the `Facility` class to return undefined instead of null - could simplify some things, but maybe unintended consequences from this, one for the future.

gef-ui/server/controllers/bank-review-date/post-bank-review-date.ts
MarRobSoftwire: My instinct is overkill since they're only used in 2 places & aren't really reusable for anything else

gef-ui/server/types/view-models/bank-review-date-view-model.ts
ttbarnes: bump

gef-ui/server/controllers/bank-review-date/get-bank-review-date.ts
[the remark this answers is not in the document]
ttbarnes: It seems simpler to remove this I think @MarRobSoftwire but happy to keep in if you think this might be a scenario that happens. 

I'm not sure how or if the API response would ever return a deal or facility ID that is _not_ a string?

gef-ui/server/controllers/bank-review-date/get-bank-review-date.ts
```diff
@@ -0,0 +1,87 @@
...
+    query: { status },
+  } = req;
+  const { userToken } = asLoggedInUserSession(req.session);
+
+  try {
+    const { details: facility } = await api.getFacility({ facilityId, userToken });
+    const deal = await api.getApplication({ dealId, userToken });
+
+    const dealVersion = parseDealVersion(deal.version);
+    const shouldRedirectFromPage = !isFacilityEndDateEnabledOnGefVersion(parseDealVersion(dealVersion)) || facility.isUsingFacilityEndDate !== false;
```
ttbarnes, line 58: There's a duplicate `parseDealVersion` here. Also looks like the "not false" check could just check if its true?

```suggestion
  const shouldRedirectFromPage = !isFacilityEndDateEnabledOnGefVersion(dealVersion) || facility.isUsingFacilityEndDate;
```

gef-ui/server/controllers/bank-review-date/get-bank-review-date.ts
```diff
@@ -0,0 +1,92 @@
...
+  previousPage: string;
+};
+
+/**
+ * @param facility - the facility
+ * @param previousPage - the previous page url
+ * @param status - the query parameter status
+ * @returns view model for the bank review date template
+ */
+const getBankReviewDateViewModel = (facility: Facility, previousPage: string, status: string | undefined): BankReviewDateViewModel => {
```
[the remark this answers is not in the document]
ttbarnes: Resolving this since this is in place in a few places and want to be consistent.

gef-ui/server/controllers/facility-end-date/get-facility-end-date.ts
```diff
@@ -0,0 +1,91 @@
...
+    const dealVersion = parseDealVersion(deal.version);
+    const shouldRedirectFromPage = !isFacilityEndDateEnabledOnGefVersion(dealVersion) || !facility.isUsingFacilityEndDate;
+
+    if (shouldRedirectFromPage) {
+      return res.redirect(previousPage);
+    }
+
+    return res.render('partials/facility-end-date.njk', getFacilityEndDateViewModel(facility, previousPage, status));
+  } catch (error) {
+    console.error('Error in getFacilityEndDate', error);
```
Zainzzkk, line 70: sorry need a %o at the end of the string

gef-ui/server/controllers/unissued-facilities/index.js
Zainzzkk: I missed this comment, do we also have unit tests for these?

gef-ui/server/types/view-models/bank-review-date-view-model.ts
MarRobSoftwire: `null` is different to `undefined`, `validationErrorHandler` returns `null`

gef-ui/server/controllers/bank-review-date/get-bank-review-date.ts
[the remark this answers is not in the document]
MarRobSoftwire: Removed

gef-ui/server/controllers/unissued-facilities/index.js
MarRobSoftwire: Yes in `gef-ui\server\controllers\unissued-facilities\index.test.js`

gef-ui/server/types/view-models/bank-review-date-view-model.ts
ttbarnes: @MarRobSoftwire Noticed we have a duplicate type here, this is the same as `FacilityEndDateViewModel`, could they both be generalised, maybe something like `DateViewModel`? I'm not familiar with other view model type namings

gef-ui/server/controllers/bank-review-date/get-bank-review-date.ts
```diff
@@ -0,0 +1,87 @@
...
+    query: { status },
+  } = req;
+  const { userToken } = asLoggedInUserSession(req.session);
+
+  try {
+    const { details: facility } = await api.getFacility({ facilityId, userToken });
+    const deal = await api.getApplication({ dealId, userToken });
+
+    const dealVersion = parseDealVersion(deal.version);
+    const shouldRedirectFromPage = !isFacilityEndDateEnabledOnGefVersion(dealVersion) || facility.isUsingFacilityEndDate !== false;
```
ttbarnes, line 58: Not sure if we discussed before

```suggestion
    const shouldRedirectFromPage = !isFacilityEndDateEnabledOnGefVersion(dealVersion) || facility.isUsingFacilityEndDate;
```

ttbarnes (commented):
2x minor comments, otherwise LGTM, nice work

gef-ui/server/controllers/bank-review-date/get-bank-review-date.ts
MarRobSoftwire: We also need to redirect if `isUsingFacilityEndDate` is `undefined` or `null`

gef-ui/server/types/view-models/bank-review-date-view-model.ts
MarRobSoftwire: `FacilityEndDateViewModel` has `facilityEndDate` whilst this one has `bankReviewDate` - i could commonise with some sort of generic but I think that would just obsfucate the code & tie the two together

gef-ui/server/controllers/bank-review-date/get-bank-review-date.ts
MarRobSoftwire: have added more unit tests

gef-ui/server/controllers/bank-review-date/get-bank-review-date.ts
ttbarnes: Thanks @MarRobSoftwire Any chance we could add a comment about this?

Just thinking, a future developer might spot this and try to simplify like I suggested, and then break things 😄

ttbarnes (approved):
Great work 🚀 🎉

Zainzzkk (approved):

PR #3413, opened 2024-08-22
feat(FN-2406): move utilisation report data validation to central api

RLavender98:
## Introduction :pencil2:
We need to add some validation for the report data which requires making database calls in a future ticket ([FN-3162](https://ukef-dtfs.atlassian.net/browse/FN-3162)). Since we need to find and display all errors at the same time and that will require a call to dtfs-central-api, we have decided to move all the report data validation to dtfs-central-api since it provides the following benefits:
- We only need to define the validation and run it once if it happens only in dtfs-central-api
- We can use the same validation for the intitial file selection and the subsequent file validation when we actually save the report, which will remove any possibility for discrepancy between the two

- Move utilisation report data validation and error message generation to a new endpoint in dtfs-central-api `POST /utilsiation-reports/validate`

- Update the validation file to typescript
- Add docstrings to the validation functions

## Out of Scope
- A second PR will follow which will contain the changes to move the saving of the report data endpoint to use the same validation


portal-api/src/generateApp.js
```diff
@@ -30,7 +30,7 @@ const generateApp = () => {
   app.use(createRateLimit());
   app.use(healthcheck);
   app.use(passport.initialize());
-  app.use(express.json());
+  app.use(express.json({ limit: '500kb' }));
```
RLavender98, line 33: Sending the report data with the locations as structured JSON is quite big for large files. I spoke to Christian about this and he agreed for now we can up this limit to match the dtfs-central-api limit, but we might consider the priority or supporting even larger files and change how we send the data (perhaps zipping it) after discussing with product in a follow up PR

dtfs-central-api/src/v1/controllers/utilisation-report-service/post-validate-utilisation-report-data.controller/index.test.ts
```diff
@@ -0,0 +1,117 @@
...
+
+      jest.mocked(validateUtilisationReportCsvData).mockReturnValue([{ errorMessage: 'Data invalid!' }]);
+
+      // Act
+      postValidateUtilisationReportData(req, res);
+
+      // Assert
+      expect(validateUtilisationReportCsvData).toHaveBeenCalledWith(reportData);
+      expect(res._getStatusCode()).toBe(HttpStatusCode.Ok);
+      expect(res._getData()).toEqual({ csvValidationErrors: [{ errorMessage: 'Data invalid!' }] });
```
JoshBinns2000, line 40: It feels so strange to respond with a 200 with validation errors, but I like this solution

dtfs-central-api/src/v1/controllers/utilisation-report-service/post-validate-utilisation-report-data.controller/index.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export type PostValidateUtilisationReportDataRequest = CustomExpressRequest<{
+  reqBody: PostValidateUtilisationReportDataPayload;
+}>;
+
+/**
+ * Controller to handle post request to validate utilisation report data route
+ * @param req - The request
+ * @param res - The response
+ * @returns - The response with the body set to contain any error messages to display
```
JoshBinns2000, line 15: So this is something that may become relevant at some point - the return type of `RequestHandler` in `express` should be `void` or `Promise<void>` (as of express 5 I think). This is not a problem in this file, but can become a problem in routes files ie. `router.post(postValidateUtilisationReportData)` would display a typescript error. I know this is something that we have done in most of our controller files which would need to be fixed, but could we update this controller so that the return type is `void`

dtfs-central-api/src/v1/routes/utilisation-reports-routes.js
```diff
@@ -69,12 +71,50 @@ const utilisationReportsRouter = express.Router();
...
+ */
+utilisationReportsRouter.route('/validate').post(validatePostValidateUtilisationReportDataPayload, postValidateUtilisationReportData);
+
 /**
  * @openapi
  * /utilisation-reports/:id:
  *   get:
  *     summary: Get utilisation report with the specified id ('id')
- *     tags: [Utilisation Report]
+ *     tags: c
```
JoshBinns2000, line 117: Was this unintentional?

dtfs-central-api/src/v1/routes/utilisation-reports-routes.js
```diff
@@ -355,8 +395,6 @@ utilisationReportsRouter
  *               $ref: '#/definitions/FeeRecordsToKeyResponseBody'
  *       400:
  *         description: Bad request
- *       404:
- *         description: Not Found
```
JoshBinns2000, line 359: Why has this been removed?

dtfs-central-api/src/v1/routes/utilisation-reports-routes.js
```diff
@@ -69,12 +71,50 @@ const utilisationReportsRouter = express.Router();
...
+ *                     csvValidationErrors:
+ *                       type: array
+ *                       items:
+ *                         $ref: '#/definitions/CsvValidationError'
+ *       500:
+ *         description: Internal server error
+ *       400:
+ *         description: Invalid payload
+ *       409:
+ *         description: Server conflict
```
JoshBinns2000, line 108: Out of curiosity, when can this return a `409`?

dtfs-central-api/src/v1/routes/utilisation-reports-routes.js
```diff
@@ -69,12 +71,50 @@ const utilisationReportsRouter = express.Router();
  */
 utilisationReportsRouter.route('/').post(postUploadUtilisationReportPayloadValidator, postUploadUtilisationReport);
 
+/**
+ * @openapi
+ * /utilisation-reports/validate:
```
JoshBinns2000, line 76: Don't feel too strongly about this, but `validate` feels like it may be too broad and it might be better to use something more specific like `validate-report-csv`. What do you think?

dtfs-central-api/src/v1/swagger-definitions/utilisation-reports/utilisation-report-data-validation.js
```diff
@@ -0,0 +1,39 @@
...
+ *         properties:
+ *           value:
+ *             type: string
+ *             nullable: true
+ *           column:
+ *             type: string
+ *           row:
+ *             oneOf:
+ *               - type: string
+ *               - type: number
```
JoshBinns2000, line 19: Haven't seen this syntax before. Have we checked that it works?

portal-api/src/v1/controllers/utilisation-report-service/validate-utilisation-report-data.controller.ts
```diff
@@ -0,0 +1,31 @@
...
+
+type ValidateUtilisationReportDataRequest = CustomExpressRequest<{
+  reqBody: ValidateUtilisationReportDataRequestBody;
+}>;
+
+/**
+ * Calls the DTFS Central API to get a validation errors for utilisation report data
+ * @param req - The request object containing information about the HTTP request.
+ * @param res - The response object used to send the HTTP response.
+ * @returns An object containing the validations errors for the utilisation report data
```
JoshBinns2000, line 18: Same comment as above for express `RequestHandler` return type

portal-api/src/v1/controllers/utilisation-report-service/validate-utilisation-report-data.controller.ts
```diff
@@ -0,0 +1,31 @@
...
+ */
+export const validateUtilisationReportData = async (req: ValidateUtilisationReportDataRequest, res: Response) => {
+  try {
+    const { reportData } = req.body;
+    const validationErrors = await api.validateUtilisationReportData(reportData);
+    return res.status(200).send(validationErrors);
```
JoshBinns2000, line 24: Could we use `HttpStatusCode.Ok` here please

JoshBinns2000 (changes requested):
Very nice solution to this problem. Many of the comments are just repeats of the same thing in multiple files

dtfs-central-api/src/v1/controllers/utilisation-report-service/post-validate-utilisation-report-data.controller/index.ts
RLavender98: I think I have resolved this by just removing the returns, let me know if that was not the correct solution!

dtfs-central-api/src/v1/routes/utilisation-reports-routes.js
RLavender98: Yes 😅 Will revert!

dtfs-central-api/src/v1/routes/utilisation-reports-routes.js
RLavender98: Ah whoops I think I got a bit confused editing this file at some point, reverting

dtfs-central-api/src/v1/routes/utilisation-reports-routes.js
RLavender98: Good question, I don't know I copied it from the above route without thinking, have removed

dtfs-central-api/src/v1/routes/utilisation-reports-routes.js
RLavender98: I've gone for `report-data-validation`, let me know if that is ok!

dtfs-central-api/src/services/utilisation-report-data-validator/utilisation-report-cell-validators/types/validation-error-generator.ts
```diff
@@ -0,0 +1,8 @@
+import { UtilisationReportDataValidationError, UtilisationReportCsvCellData, UtilisationReportCsvRowData } from '@ukef/dtfs2-common';
+
+export type UtilisationReportCellValidationErrorGenerator = (
+  cellData: UtilisationReportCsvCellData | undefined,
+  exporterName?: string | null,
```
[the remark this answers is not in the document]
RLavender98: Yes, so cell object `value` can either be string or null, but then if the exporter heading was missing there would be no object to pull the value from and then in that case we would pass in undefined.

dtfs-central-api/src/v1/swagger-definitions/utilisation-reports/utilisation-report-data-validation.js
RLavender98: Yes, it's here in the docs: https://swagger.io/docs/specification/data-models/oneof-anyof-allof-not/
Screenshot 2024-08-30 at 08 46 01

dtfs-central-api/src/v1/controllers/utilisation-report-service/post-validate-utilisation-report-data.controller/index.ts
RLavender98: Ah ok that did not work because it sends a response but continues the current function. So the option is to move to having an empty return immediately following the call, but I'm not sure the internet agrees that is the nicest pattern and I do think perhaps this is standard e.g. https://dev.to/adamkatora/why-you-should-always-use-return-before-ressend-in-express-apis-and-applications-k9k#:~:text=In%20Express%2C%20calling%20the%20res,not%20exit%20the%20current%20function.

dtfs-central-api/src/v1/routes/utilisation-reports-routes.js
JoshBinns2000: Sounds good to me

dtfs-central-api/src/services/utilisation-report-data-validator/index.ts
```diff
@@ -0,0 +1,140 @@
...
+
+/**
+ * Validate the utilisation report csv data
+ * @param csvData - The data from the utilisation report csv
+ * @returns An array of errors pertaining to the report if there are any
+ */
+export const validateUtilisationReportCsvData = (csvData: UtilisationReportCsvRowData[]): UtilisationReportDataValidationError[] => {
+  const { missingHeaderErrors, availableHeaders } = validateUtilisationReportCsvHeaders(csvData[0]);
+
+  const dataValidationErrors = validateUtilisationReportCsvCellData(csvData, availableHeaders);
```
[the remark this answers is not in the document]
JoshBinns2000: I thought as much and am happy to leave as it - I just have a preference for column-by-column with respect to readability, but transforming the spreadsheet to that format would itself be hard to read

dtfs-central-api/src/v1/controllers/utilisation-report-service/post-validate-utilisation-report-data.controller/index.ts
JoshBinns2000: That's interesting - it seems this is a semantic thing with `RequestHandler` as depending on where you use the type it has different behaviour (which means that returning the `res.send` call is actually fine so long as you don't say that the function itself is a `RequestHandler` type and use it directly in the `router.use` params). Agree that leaving this as is would be better for consistency either way

JoshBinns2000 (dismissed):
Small nitpick, but going to approve anyway. Nice job

dtfs-central-api/api-tests/v1/utilisation-reports/post-validate-utilisation-report-data.api-test.ts
```diff
@@ -0,0 +1,45 @@
+import { HttpStatusCode } from 'axios';
+import { testApi } from '../../test-api';
+
+console.error = jest.fn();
+
+const URL = '/v1/utilisation-reports/report-data-validation';
+
+describe(`POST ${URL}`, () => {
+  it('returns a 200 when request body is valid', async () => {
```
Zainzzkk, line 9: could we format this so it is `returns a ${HttpStatusCode.Ok} when request body is valid` and the same for the rest in this file please

dtfs-central-api/src/services/utilisation-report-data-validator/index.test.ts
```diff
@@ -0,0 +1,140 @@
...
+
+      // Assert
+      expect(error).toEqual(`${header} header is missing or spelt incorrectly`);
+    });
+  });
+
+  describe('validateUtilisationReportCsvHeaders', () => {
+    it('returns an error if a header is missing', () => {
+      const csvDataRowWithMissingHeader = {
+        'bank facility reference': { value: 'Britannia Energy GEF', column: 'A', row: 1 },
```
Zainzzkk, line 38: are any of these stored in constants?

portal-api/src/generateApp.js
ttbarnes: Should this be in a constant?

dtfs-central-api/src/services/utilisation-report-data-validator/index.test.ts
```diff
@@ -0,0 +1,140 @@
...
+      };
+
+      const { missingHeaderErrors } = validateUtilisationReportCsvHeaders(csvDataRowWithMissingHeader);
+
+      expect(missingHeaderErrors.length).toBe(1);
+      expect(missingHeaderErrors[0].errorMessage).toBe('UKEF facility ID header is missing or spelt incorrectly');
+    });
+
+    it('returns no errors when no headers are missing', () => {
+      const csvDataRowWithCorrectHeaders = {
```
Zainzzkk, line 56: could we have a test-mock for this.  And then we can modify each component of mock which needs to be modified

dtfs-central-api/src/services/utilisation-report-data-validator/index.test.ts
```diff
@@ -0,0 +1,140 @@
...
+
+      const availableHeaders = [UTILISATION_REPORT_HEADERS.UKEF_FACILITY_ID, UTILISATION_REPORT_HEADERS.BASE_CURRENCY];
+
+      validateUtilisationReportCsvCellData(csvData, availableHeaders);
+      expect(generateUkefFacilityIdError).toHaveBeenCalledWith({ value: '20001371', column: 'B', row: 1 }, 'test exporter');
+      expect(generateBaseCurrencyError).toHaveBeenCalledWith({ value: 'GBP', column: 'D', row: 1 }, 'test exporter');
+      expect(generateFacilityUtilisationError).not.toHaveBeenCalled();
+    });
+
+    it('calls the generate payment currency and exchange rate error functions even if no headers are present', () => {
```
Zainzzkk, line 120: can all these its start with 'should'?

portal/server/api.js
```diff
@@ -884,6 +884,27 @@ const uploadUtilisationReportData = async (uploadingUser, reportPeriod, csvData,
   }
 };
 
+/**
+ * Generates validation errors for all errors in the utilisation report data
+ * @param {import('@ukef/dtfs2-common').UtilisationReportCsvRowData[]} reportData - The csv data with location information
+ * @param {string} userToken - The user token
+ * @returns {Promise<{csvValidationErrors: import('@ukef/dtfs2-common').UtilisationReportDataValidationError[]}>} - The validation errors for the utilisation report data
```
ttbarnes, line 891: Can we have a simple return type instaed of `import('...`?

dtfs-central-api/src/services/utilisation-report-data-validator/index.test.ts
```diff
@@ -0,0 +1,140 @@
...
+          'total fees accrued for the period': { value: '367.23', column: 'G', row: 1 },
+          'fees paid to ukef for the period': { value: '367.23', column: 'H', row: 1 },
+          'payment reference': { value: 'Britannia Energy / 3001175147', column: 'I', row: 1 },
+        },
+      ];
+
+      const availableHeaders = [UTILISATION_REPORT_HEADERS.UKEF_FACILITY_ID, UTILISATION_REPORT_HEADERS.BASE_CURRENCY];
+
+      validateUtilisationReportCsvCellData(csvData, availableHeaders);
+      expect(generateUkefFacilityIdError).toHaveBeenCalledWith({ value: '20001371', column: 'B', row: 1 }, 'test exporter');
```
Zainzzkk, line 115: can these be referenced from csvData (csvData[field]

portal/server/controllers/utilisation-report-service/utilisation-report-upload/index.js
```diff
@@ -117,7 +116,7 @@ const postUtilisationReportUpload = async (req, res) => {
     }
 
     // File is valid so we can start processing and validating its data
-    const { csvJson, fileBuffer, error } = await extractCsvData(req.file); // do we here catch some errors and return generic something went wrong here?
+    const { csvJson, fileBuffer, error } = await extractCsvData(req.file);
```
ttbarnes, line 119: Do we need to have `extractCsvData` here? I think such file buffers should be handled by the API and the UI just consumes it.

Would like to avoid the UI having to do XYZ because the API doesn't return a desired/ideal response

dtfs-central-api/src/services/utilisation-report-data-validator/utilisation-report-cell-validators/generate-base-currency-error.test.ts
```diff
@@ -40,10 +40,10 @@ describe('generateBaseCurrencyError', () => {
     expect(baseCurrencyError).toEqual(expectedError);
   });
 
-  it('returns null if the value is a valid currency', async () => {
+  it('returns null if the value is a valid currency', () => {
```
Zainzzkk, line 43: can use toBeNull for the expect statement for this test

dtfs-central-api/src/services/utilisation-report-data-validator/utilisation-report-cell-validators/generate-base-currency-error.ts
```diff
@@ -0,0 +1,30 @@
...
+  if (!currencyCellData?.value) {
+    return {
+      errorMessage: 'Base currency must have an entry',
+      column: currencyCellData?.column,
+      row: currencyCellData?.row,
+      value: currencyCellData?.value,
+      exporter: exporterName,
+    };
+  }
+  if (!validator.isISO4217(currencyCellData.value)) {
```
Zainzzkk, line 20: empty line here and line 59 please

dtfs-central-api/src/services/utilisation-report-data-validator/utilisation-report-cell-validators/generate-facility-utilisation-error.ts
Zainzzkk, line 1: empty lines between the ifs and the return please

dtfs-central-api/src/v1/controllers/utilisation-report-service/post-report-data-validation.controller/index.ts
```diff
@@ -0,0 +1,30 @@
...
+ * Controller to handle post request to report data validation route
+ * @param req - The request
+ * @param res - The response
+ */
+export const postReportDataValidation = (req: PostReportDataValidationRequest, res: Response) => {
+  const { reportData } = req.body;
+
+  try {
+    const csvValidationErrors = validateUtilisationReportCsvData(reportData);
+    return res.status(HttpStatusCode.Ok).send({ csvValidationErrors });
```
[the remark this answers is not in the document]
RLavender98: Yes, but can I ask why? Just in case there is a way we can perhaps capture the requirement in our formatting config!

dtfs-central-api/src/v1/controllers/utilisation-report-service/post-report-data-validation.controller/index.ts
```diff
@@ -0,0 +1,30 @@
...
+  } catch (error) {
+    const errorMessage = 'Failed to validate report data';
+    console.error(errorMessage, error);
+    if (error instanceof ApiError) {
+      return res.status(error.status).send(`${errorMessage}: ${error.message}`);
+    }
+    return res.status(HttpStatusCode.InternalServerError).send(errorMessage);
```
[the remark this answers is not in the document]
RLavender98: I think this is the standard pattern we have in our controllers for handling `ApiError`s, so keen to keep it for consistency. Note that the status used and message is slightly different between lines 26 and 28 so it isn't the same.

dtfs-central-api/src/services/utilisation-report-data-validator/index.ts
```diff
@@ -0,0 +1,149 @@
...
+ * @returns The header is missing error message for the provided header
+ */
+export const getHeaderIsMissingErrorMessage = (header: string) => `${header} ${HEADER_IS_MISSING_BASE_ERROR_MESSAGE}`;
+
+/**
+ * Validate utilisation report csv headers
+ * @param csvDataRow - A row of data
+ * @returns Errors for any missing required headers and a list of the present headers
+ */
+export const validateUtilisationReportCsvHeaders = (
```
Zainzzkk, line 34: should present headers also be returned in this function? as the function is called validate

dtfs-central-api/src/services/utilisation-report-data-validator/utilisation-report-cell-validators/generate-fees-paid-for-the-period-currency-error.ts
Zainzzkk, line 1: empty lines between the ifs and return please

dtfs-central-api/src/services/utilisation-report-data-validator/utilisation-report-cell-validators/generate-fees-paid-for-the-period-currency-error.ts
```diff
@@ -0,0 +1,33 @@
...
+  if (!feesPaidForThePeriodCurrencyCellData?.value) {
+    return {
+      errorMessage: 'Fees paid to UKEF currency must have an entry',
+      column: feesPaidForThePeriodCurrencyCellData?.column,
+      row: feesPaidForThePeriodCurrencyCellData?.row,
+      value: feesPaidForThePeriodCurrencyCellData?.value,
+      exporter: exporterName,
+    };
+  }
+  if (!validator.isISO4217(feesPaidForThePeriodCurrencyCellData.value)) {
```
Zainzzkk, line 23: should this also be ?.value ?

dtfs-central-api/src/services/utilisation-report-data-validator/utilisation-report-cell-validators/generate-payment-currency-error.ts
```diff
@@ -1,11 +1,18 @@
...
-const generatePaymentCurrencyError = (csvDataRow) => {
-  if (!csvDataRow[UTILISATION_REPORT_HEADERS.PAYMENT_EXCHANGE_RATE]?.value && !csvDataRow[UTILISATION_REPORT_HEADERS.PAYMENT_CURRENCY]?.value) {
+/**
+ * Generate and return error for payment currency entry in the csv row if value is invalid
+ * @param csvDataRow - The row data for which to validate the payment currency
+ * @returns The error if the payment currency entry is invalid, null if the payment currency entry is valid
+ */
+export const generatePaymentCurrencyError: UtilisationReportRowValidationErrorGenerator = (csvDataRow) => {
+  const paymentCurrencyValue = csvDataRow[UTILISATION_REPORT_HEADERS.PAYMENT_CURRENCY]?.value;
+  if (!csvDataRow[UTILISATION_REPORT_HEADERS.PAYMENT_EXCHANGE_RATE]?.value && !paymentCurrencyValue) {
```
Zainzzkk, line 12: could this not also have a const for `csvDataRow[UTILISATION_REPORT_HEADERS.PAYMENT_EXCHANGE_RATE]?.value` like there is for paymentCurrencyValue

dtfs-central-api/src/services/utilisation-report-data-validator/utilisation-report-cell-validators/generate-total-fees-accrued-currency-error.ts
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * Generate and return error for total fees accrued currency entry in csv row if value is invalid
+ * @param csvDataRow - The row data for which to validate the total fees accrued currency
+ * @returns The error if the total fees accrued currency entry is invalid, null if the total fees accrued currency entry is valid
+ */
+export const generateTotalFeesAccruedCurrencyError: UtilisationReportRowValidationErrorGenerator = (csvDataRow) => {
+  const totalFeesAccruedCurrencyValue = csvDataRow[UTILISATION_REPORT_HEADERS.TOTAL_FEES_ACCRUED_CURRENCY]?.value;
+
+  if (!totalFeesAccruedCurrencyValue && !csvDataRow[UTILISATION_REPORT_HEADERS.TOTAL_FEES_ACCRUED_EXCHANGE_RATE]?.value) {
```
Zainzzkk, line 13: could there be a const for EXCHANGERATE too please

dtfs-central-api/src/services/utilisation-report-data-validator/utilisation-report-cell-validators/generate-total-fees-accrued-exchange-rate-error.ts
```diff
@@ -1,21 +1,24 @@
...
+ * @returns - object comprising of error message and error location or null if valid.
  */
-const generateTotalFeesAccruedExchangeRateError = (csvDataRow) => {
+export const generateTotalFeesAccruedExchangeRateError: UtilisationReportRowValidationErrorGenerator = (csvDataRow) => {
+  const totalFeesAccruedExchangeRateValue = csvDataRow[UTILISATION_REPORT_HEADERS.TOTAL_FEES_ACCRUED_EXCHANGE_RATE]?.value;
+
   if (
-    !csvDataRow[UTILISATION_REPORT_HEADERS.TOTAL_FEES_ACCRUED_EXCHANGE_RATE]?.value &&
+    !totalFeesAccruedExchangeRateValue &&
     (!csvDataRow[UTILISATION_REPORT_HEADERS.TOTAL_FEES_ACCRUED_CURRENCY]?.value ||
```
Zainzzkk, line 16: could these each be const and the or statement be another const.  Will be easier to read

dtfs-central-api/src/services/utilisation-report-data-validator/utilisation-report-cell-validators/generate-ukef-facility-id-error.ts
```diff
@@ -0,0 +1,30 @@
+import { UKEF_FACILITY_ID_REGEX } from '../../../constants/regex';
+import { UtilisationReportCellValidationErrorGenerator } from './types/validation-error-generator';
+
+/**
+ * Generate and return error for the facility id cell if value is invalid
```
Zainzzkk, line 5: can the descriptions all be checked, and add `an error` if they do not already

dtfs-central-api/src/services/utilisation-report-data-validator/utilisation-report-cell-validators/generate-ukef-facility-id-error.ts
```diff
@@ -0,0 +1,30 @@
...
+  if (!facilityIdCellData?.value) {
+    return {
+      errorMessage: 'UKEF facility ID must have an entry',
+      column: facilityIdCellData?.column,
+      row: facilityIdCellData?.row,
+      value: facilityIdCellData?.value,
+      exporter: exporterName,
+    };
+  }
+  if (!UKEF_FACILITY_ID_REGEX.test(facilityIdCellData.value)) {
```
Zainzzkk, line 20: I think this could be a helper.  Can pass it the REGEX, facilityIdCellData and exporterName and it returns the error for us

dtfs-central-api/src/v1/controllers/utilisation-report-service/post-report-data-validation.controller/index.test.ts
```diff
@@ -0,0 +1,121 @@
...
+      });
+
+      // Act
+      postReportDataValidation(req, res);
+
+      // Assert
+      expect(res._getData()).toBe(`Failed to validate report data: ${errorMessage}`);
+    });
+
+    it("responds with a '500' if an unknown error occurs", () => {
```
Zainzzkk, line 85: same as above. could the it have `HttpStatusCode.InternalServerError` instead of the hardcoded number

dtfs-central-api/src/services/utilisation-report-data-validator/utilisation-report-cell-validators/generate-total-fees-accrued-error.ts
```diff
@@ -0,0 +1,33 @@
...
+  if (!value) {
+    return generateError('Total fees accrued for the period must have an entry');
+  }
+  if (!CURRENCY_NUMBER_REGEX.test(value)) {
+    return generateError('Total fees accrued for the period must be a number with a maximum of two decimal places');
+  }
+  if (value.length > FILE_UPLOAD.MAX_CELL_CHARACTER_COUNT) {
+    return generateError(`Total fees accrued for the period must be ${FILE_UPLOAD.MAX_CELL_CHARACTER_COUNT} characters or less`);
+  }
+  return null;
```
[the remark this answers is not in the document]
RLavender98: I think a switch case might be a little harder to read and also quite a large refactor so I will add empty lines. 

Is that a general rule having blank lines after if statements or just a readability issue for these specific files? If it is a general rule do you think you could raise a ticket to cover adding it to the formatting (either prettier/eslint) config?

dtfs-central-api/src/v1/controllers/utilisation-report-service/post-report-data-validation.controller/index.ts
```diff
@@ -0,0 +1,30 @@
...
+ */
+export const postReportDataValidation = (req: PostReportDataValidationRequest, res: Response) => {
+  const { reportData } = req.body;
+
+  try {
+    const csvValidationErrors = validateUtilisationReportCsvData(reportData);
+    return res.status(HttpStatusCode.Ok).send({ csvValidationErrors });
+  } catch (error) {
+    const errorMessage = 'Failed to validate report data';
+    console.error(errorMessage, error);
```
Zainzzkk, line 24: @abhi-markan should this be %s %s?

dtfs-central-api/src/v1/routes/middleware/payload-validation/schemas/utilisation-report-csv-row-data.schema.ts
```diff
@@ -0,0 +1,9 @@
+import z from 'zod';
+
+const UtilisationReportCsvCellDataSchema = z.object({
+  value: z.string().nullable(),
+  column: z.string(),
+  row: z.union([z.string(), z.number()]),
+});
+
+export const UtilisationReportCsvRowDataSchema = z.record(z.string(), UtilisationReportCsvCellDataSchema);
```
Zainzzkk, line 9: have some documentation here please.  Useful to see what z.record etc does on first glance if new

portal-api/src/v1/api.js
```diff
@@ -269,6 +269,29 @@ const findLatestGefMandatoryCriteria = async () => {
   }
 };
 
+/**
+ * Validates utilisation report data returning any errors to display to the user
+ * @param {import('@ukef/dtfs2-common').UtilisationReportCsvRowData[]} reportData
+ * @returns {Promise<import('./api-response-types').ValidateUtilisationReportDataResponseBody>} Object containing the validation errors to display to the user
```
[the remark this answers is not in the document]
RLavender98: I'm not sure I follow what you are suggesting? I have created a type for this response that we refer to here to try and make this as simple as possible.

portal-api/src/v1/controllers/utilisation-report-service/validate-utilisation-report-data.controller.ts
```diff
@@ -0,0 +1,30 @@
...
+  try {
+    const { reportData } = req.body;
+    const validationErrors = await api.validateUtilisationReportData(reportData);
+    return res.status(HttpStatusCode.Ok).send(validationErrors);
+  } catch (error) {
+    const errorMessage = 'Failed to validate utilisation report data';
+    console.error(errorMessage, error);
+    const errorStatus = (isAxiosError(error) && error.response?.status) || HttpStatusCode.InternalServerError;
+    return res.status(errorStatus).send(errorMessage);
+  }
```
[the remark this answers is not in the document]
RLavender98: Have done this, but again curious what the exact rule here is? I personally find it a little harder to read with a blank line between each line of code 😅

portal/server/api.js
RLavender98: Sorry I'm not sure I understand what alternative way there is to type this?

portal-api/src/v1/controllers/utilisation-report-service/validate-utilisation-report-data.controller.ts
Zainzzkk, line 1: some empty lines please

portal/server/controllers/utilisation-report-service/utilisation-report-upload/index.js
RLavender98: Totally agree with you on that, here we do multiple things with the extracted data, including attach it to the users session I believe so that we can use it on the next page since they have to confirm before uploading, so it is required to live here.

RLavender98: I did briefly think about this when I started but discarded the idea but thinking about it more now I agree with where you are coming from. It is however a lot of additional work and having spoken to Christian there would be a few other things to consider:
- Increasing the number of file scans we do through clamscan
- Introducing multer and a bunch of context around file uploading to dtfs-central-api when the actual file gets uploaded to Azure in portal-api

If we were going to make this change it would need to be a follow on ticket, but would need to consider a bit more if it is something we should do.

dtfs-central-api/src/v1/controllers/utilisation-report-service/post-report-data-validation.controller/index.ts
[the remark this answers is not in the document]
ttbarnes: Actually sorry @RLavender98 I had a look, seems that the approach in place here (without empty lines) is the pattern in a lot of central API controllers, should be aligned like you've already done 👍 It's easier to read with the empty lines

dtfs-central-api/src/v1/controllers/utilisation-report-service/post-report-data-validation.controller/index.ts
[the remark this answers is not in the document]
ttbarnes: Thanks @RLavender98 - looks like this is only in place for some utilisation report and some amendment controllers though so perhaps it isn't a standard pattern.

I did notice the statuses are different - what's the reasoning for this, why do we need to do this? Do any consumers of this endpoint react differently depending on the status?

dtfs-central-api/src/services/utilisation-report-data-validator/utilisation-report-cell-validators/generate-total-fees-accrued-error.ts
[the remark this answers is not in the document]
ttbarnes: I think it's subjective 😄 but happy with if statements and new lines.

Yes it's a readability thing, it's a lot easier to read if the if statements are broken up more IMO.

@abhi-markan @Zainzzkk What do you think please, should we look into an eslint/prettier rule to enforce a new line before/after if statements?

dtfs-central-api/src/v1/controllers/utilisation-report-service/post-report-data-validation.controller/index.ts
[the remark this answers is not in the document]
ttbarnes: Sorry, for this controller, same comment as here RE alignment that I didn't realise https://github.com/UK-Export-Finance/dtfs2/pull/3413/files#r1741022249

portal-api/src/v1/api.js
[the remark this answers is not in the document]
ttbarnes: The import documentation seems overkill to me, I think something like this would be a lot simpler and easier to read/maintain:

```ts
* @param {UtilisationReportCsvRowData} reportData
 * @returns {Promise<ValidateUtilisationReportDataResponseBody>}
```

portal/server/api.js
ttbarnes: Same reply as here => https://github.com/UK-Export-Finance/dtfs2/pull/3413#discussion_r1741064160

dtfs-central-api/src/services/utilisation-report-data-validator/index.test.ts
```diff
@@ -0,0 +1,140 @@
...
+      const csvData = [
+        {
+          'bank facility reference': { value: 'Britannia Energy GEF', column: 'A', row: 1 },
+          'ukef facility id': { value: '20001371', column: 'B', row: 1 },
+          exporter: { value: 'test exporter', column: 'C', row: 1 },
+          'base currency': { value: 'GBP', column: 'D', row: 1 },
+          'facility limit': { value: '600000', column: 'E', row: 1 },
+          'facility utilisation': { value: '34538e.54', column: 'F', row: 1 },
+          'total fees accrued for the period': { value: '367.23', column: 'G', row: 1 },
+          'fees paid to ukef for the period': { value: '367.23', column: 'H', row: 1 },
```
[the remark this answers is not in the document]
RLavender98: I've refactored these a bit and noticed a slight gap in coverage so have rectified that also 😄

dtfs-central-api/src/services/utilisation-report-data-validator/utilisation-report-cell-validators/generate-payment-exchange-rate-error.test.ts
```diff
@@ -32,33 +32,33 @@ describe('generatePaymentExchangeRateError', () => {
...
         value: null,
-        column: 2,
+        column: 'C',
         row: 1,
       },
       'payment exchange rate': {
         value: 'abc',
         row: 1,
-        column: 3,
+        column: 'D',
```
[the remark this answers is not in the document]
RLavender98: I think in the majority of cases the fees paid to ukef currency is highly relevant to the test case, so I'd actually be uncomfortable pulling it out. Appreciate the importance of being DRY, but I think with the data row arrays refactors are likely to make the tests much more brittle for not much benefit.

I'm not sure I understand the value of pulling out the exporter name?

If we were wanting to truly avoid any repetition at all we would define `MOCK_NON_EMPTY_STRING = 's'` and then each test becomes:

```
const expectedError = {
      errorMessage: 'Base currency must have an entry',
      column: 'A',
      row: 1,
      value: null,
      exporter: MOCK_NON_EMPTY_STRING,
    };

    const baseCurrencyError = generateBaseCurrencyError(nullBaseCurrency, MOCK_NON_EMPTY_STRING);
```
But I personally think that is harder to read and follow?

portal-api/src/v1/api.js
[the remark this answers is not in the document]
RLavender98: So the issue is if I just type it like that then you can't click through to the type definition, and also we don't get type safety in the following file as it has no idea what that type is.

If I do your change and add `const { UtilisationReportResponseBody } = require('./api-response-types');` at the top of the api.js file, then I can click through to the definition, and we do get the type safety in the files calling the function but I get a linting error violating the `no-unused-vars` rule since this is only used in the docstring.

Let me know if you have a solution!

dtfs-central-api/src/v1/controllers/utilisation-report-service/post-report-data-validation.controller/index.ts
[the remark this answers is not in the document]
RLavender98: My understanding is ApiError is a new pattern that was brought in shortly before I joined the project which is the reason it isn't used in more places. 

The idea is we can have standard errors that extend this type and that store the status on themselves so that the controller error handling can be simpler.

Say you have an endpoint that could throw 10 different kinds of expected errors. We could check for each of these errors here one by one to allow us to return the correct status code and message depending on the error type (some might be errors that need to be BadRequest, or Forbidden for example). Instead, using the ApiError class, we have each of those 10 errors extend the ApiError type and then we can handle them all in one nice block since we get the error to own it's own status. e.g. `InvalidPayloadError` extends `ApiError` and it's error code is a `BadRequest`.

The line on 28 is for any unexpected errors, in which case that would fall under InternalServerError and we want to send a more generic message since we don't know what is in the error message since we didn't write it ourselves and we don't want to accidentally reveal sensitive information which could be in the error message.

As for consumers responding differently, I don't think they do, but I don't think that is a good reason to not use correct error codes. Distinct error codes can also be useful for logging and debugging purposes. Also if we ever wanted to have alerts set up on different status codes for example, this would be really useful, because then we could track 500s and 400s separately which would be pretty nice.

dtfs-central-api/src/services/utilisation-report-data-validator/index.test.ts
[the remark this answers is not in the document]
ttbarnes: Nice one thanks @RLavender98 😄

dtfs-central-api/src/services/utilisation-report-data-validator/utilisation-report-cell-validators/generate-payment-exchange-rate-error.test.ts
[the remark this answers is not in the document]
ttbarnes: @RLavender98 I mean the overall CSV row mocks, not just the exporter name. E.g if you search for this: `column: 'D', looks like there's a lot of repitition and we could create a common fixture for mock CSV rows?

E.g in here `generate-payment-currency-error.test`, the same mock row is duplicated 3x places with only 1/2 very small differences

Could we create e.g `mockCsvDataRow`? Looks like it could remove a fair amount of LOC 🚀

dtfs-central-api/src/services/utilisation-report-data-validator/index.test.ts
RLavender98: I've reworked the tests after a similar comment from Tony to try and make them better.

dtfs-central-api/src/services/utilisation-report-data-validator/index.test.ts
RLavender98: Have reworked this a bit and introduced a helper

dtfs-central-api/src/services/utilisation-report-data-validator/index.test.ts
RLavender98: This doesn't seem to be a consistent pattern from a search through the codebase, so I can change it but keen to understand if this is something we should be doing going forwards/something all devs are aware of?

dtfs-central-api/src/services/utilisation-report-data-validator/index.ts
RLavender98: I am worried the name would be too long if we make it say `validateUtilisationReportCsvHeadersAndReturnAvailableHeaders` but if we move that logic out then we would be duplicating some of the logic and it could fall out of sync so here I'd been keen to keep it as it is. I have updated the docstring to add more detail.

dtfs-central-api/src/services/utilisation-report-data-validator/utilisation-report-cell-validators/generate-payment-currency-error.ts
RLavender98: I didn't pull it out because unlike the other one it is only used in one place and the other one needing pulling out to get the type checking follow on from the if statement checking it was undefined. But I will pull it out now.

dtfs-central-api/src/services/utilisation-report-data-validator/utilisation-report-cell-validators/generate-ukef-facility-id-error.ts
RLavender98: We'd have to pass it quite a lot more information to make the error messaging accurate so I'm not sure the end outcome would be super valuable. Additionally, I think the amount of refactoring required would increase the scope of the ticket a fair amount which is to move the logic, not rework it extensively.

dtfs-central-api/src/v1/controllers/utilisation-report-service/post-report-data-validation.controller/index.test.ts
```diff
@@ -0,0 +1,121 @@
...
+      // Arrange
+      const req = httpMocks.createRequest<PostReportDataValidationRequest>({
+        body: aValidRequestBody(),
+      });
+      const res = httpMocks.createResponse();
+
+      jest.mocked(validateUtilisationReportCsvData).mockImplementation(() => {
+        throw new Error('Some error');
```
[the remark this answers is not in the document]
RLavender98: This is only repeated once the type of error/message are crucial to the assertions in the tests in these files so I think we lose a fair amount of readability and maintainability for not much gain in value.

dtfs-central-api/src/v1/routes/middleware/payload-validation/schemas/utilisation-report-csv-row-data.schema.ts
RLavender98: I've added a docstring, cautious I don't want to describe what z.record() does however because it is a zod method and as such has it's own documentation: https://zod.dev/?id=records

dtfs-central-api/src/v1/controllers/utilisation-report-service/post-report-data-validation.controller/index.test.ts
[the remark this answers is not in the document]
ttbarnes: I agree a DRY fixture for all tests for such errors would be nice, e.g `mockErrorMessage` imported from a mock/fixtures file, to avoid having the same "Some error" string repeated everywhere

But this is a tech improvement that's not part of this ticket. If we're going to create a DRY error message, we should do it it one swoop.

dtfs-central-api/src/services/utilisation-report-data-validator/utilisation-report-cell-validators/generate-payment-exchange-rate-error.test.ts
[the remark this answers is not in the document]
ttbarnes: Thanks @RLavender98 , I think it would be worth doing, but not a blocker for this PR since they're existing mocks and this PR only updates one value (column) in these mocks.

Even with test specifics, it looks like we could reduce a lot of LOC for example a test could e.g:

```
csvDataRow['payment exchange rate'].value = null;
```

And it'd save ~22 lines of mocks, for 1x test. Multiplying this for all the instances, we could be saving ~500 lines of mocks in total

dtfs-central-api/src/services/utilisation-report-data-validator/utilisation-report-cell-validators/generate-ukef-facility-id-error.ts
ttbarnes: I said similar @Zainzzkk https://github.com/UK-Export-Finance/dtfs2/pull/3413/files#r1740882018 , but I agree with @RLavender98 as this PR is about moving the validation to another place rather than code improvements. Would be nice to have a jira ticket to do in the future.

portal/server/controllers/utilisation-report-service/utilisation-report-upload/utilisation-report-filename-validator.ts
```diff
@@ -0,0 +1,112 @@
...
+    return { filenameError };
+  }
+
+  if (!isReportPeriodInFilename(dueReportPeriod, filename)) {
+    const exampleFilenameReportPeriod = getExampleFilenameReportPeriod(dueReportPeriod);
+    const filenameError = `The selected file must contain the reporting period as part of its name, for example '${exampleFilenameReportPeriod}'`;
+    return { filenameError };
+  }
+  return {};
+};
```
ttbarnes, line 112: 4x functions in here are untested 😢  Not a blocker for this PR as it's purely moving things around.

dtfs-central-api/src/v1/controllers/utilisation-report-service/post-report-data-validation.controller/index.ts
ttbarnes: There are currently 34 instances of this throughout the codebase. If it needs to change, I suggest we do it all in one go.

ttbarnes (dismissed):

dtfs-central-api/src/services/utilisation-report-data-validator/index.test.ts
Zainzzkk: I believe that this is what we will be doing going forward but can discuss first

dtfs-central-api/src/services/utilisation-report-data-validator/utilisation-report-cell-validators/generate-ukef-facility-id-error.ts
Zainzzkk: Would be good to have a ticket for this

dtfs-central-api/src/v1/routes/middleware/payload-validation/schemas/utilisation-report-csv-row-data.schema.ts
ttbarnes: Would be great if we could add at least a little information about what it's checking exactly.

I agree we don't want to go into hugh depth about `z.record()` and `z.string()`, but equally, if you're not familiar with it, you have to go hunting around in external documentation. I think we should have a high level description in our own documentation, something simple like "checks that the object contains a string via zod methods" would be perfect

portal-api/src/v1/api.js
```diff
@@ -269,6 +269,29 @@ const findLatestGefMandatoryCriteria = async () => {
...
+      url: `${DTFS_CENTRAL_API_URL}/v1/utilisation-reports/report-data-validation`,
+      headers: headers.central,
+      data: {
+        reportData,
+      },
+    });
+
+    return response.data;
+  } catch (error) {
+    console.error('Unable to validate utilisation report data', error);
```
Zainzzkk, line 290: needs %o after data

dtfs-central-api/src/v1/routes/middleware/payload-validation/schemas/utilisation-report-csv-row-data.schema.ts
RLavender98: Have added more detail

Zainzzkk (approved):
lgtm

ttbarnes (approved):

JoshBinns2000 (approved):

PR #3418, opened 2024-08-23
feat(FN-1725): add e2e test for keying sheets

JoshBinns2000:
## Introduction :pencil2:
We want to add some e2e tests for generating keying data.

- Adds an e2e test

### DIFF
--- a/e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/generate-keying-sheet.spec.js
+++ b/e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/generate-keying-sheet.spec.js
@@ -1,0 +1,146 @@
+import { startOfMonth, subMonths, subYears, addYears } from 'date-fns';
+import {
+  FeeRecordEntityMockBuilder,
+  PaymentEntityMockBuilder,
+  UTILISATION_REPORT_RECONCILIATION_STATUS,
+  UtilisationReportEntityMockBuilder,
+  FacilityUtilisationDataEntityMockBuilder,
+  CURRENCY,
+  FEE_RECORD_STATUS,
+} from '@ukef/dtfs2-common';
+import pages from '../../pages';
+import USERS from '../../../fixtures/users';
+import { NODE_TASKS } from '../../../../../e2e-fixtures';
+import relative from '../../relativeURL';
+
+context('PDC_RECONCILE users can generate keying data', () => {
+  const BANK_ID = '961';
+  const REPORT_ID = 1;
+  const TODAY = new Date();
+
+  const START_OF_CURRENT_REPORT_PERIOD = startOfMonth(subMonths(TODAY, 1));
+  const CURRENT_REPORT_PERIOD = {
+    start: { month: START_OF_CURRENT_REPORT_PERIOD.getMonth() + 1, year: START_OF_CURRENT_REPORT_PERIOD.getFullYear() },
+    end: { month: START_OF_CURRENT_REPORT_PERIOD.getMonth() + 1, year: START_OF_CURRENT_REPORT_PERIOD.getFullYear() },
+  };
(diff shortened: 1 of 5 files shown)


e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/generate-keying-sheet.spec.js
```diff
@@ -0,0 +1,146 @@
...
+    cy.task(NODE_TASKS.INSERT_UTILISATION_REPORTS_INTO_DB, [CURRENT_PERIOD_UTILISATION_REPORT]);
+
+    pages.landingPage.visit();
+    cy.login(USERS.PDC_RECONCILE);
+  });
+
+  it('can generate keying data when only one fee record for a facility is at MATCH status but does not generate the fixed fee adjustment or principal balance adjustment', () => {
+    const matchingPayment = PaymentEntityMockBuilder.forCurrency(CURRENCY.GBP).withAmount(100).build();
+    const firstFeeRecord = firstFeeRecordBuilder.withStatus(FEE_RECORD_STATUS.MATCH).withPayments([matchingPayment]).build();
+    const secondFeeRecord = secondFeeRecordBuilder.withStatus(FEE_RECORD_STATUS.TO_DO).build();
```
RLavender98, line 99: I'm not clear on how reusing the same builder across multiple tests is actually working since we are calling build multiple times now? 
Also, is this not a risk of inter test dependency, in that if you were to call `.withBaseCurrency(CURRENCY.EUR)` here would the builder have the value `CURRENCY.GBP` or `CURRENCY.EUR` at the start of the next test?

e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/generate-keying-sheet.spec.js
```diff
@@ -0,0 +1,146 @@
...
+    pages.utilisationReportPage.premiumPaymentsTab.generateKeyingDataButton().click();
+    pages.checkKeyingDataPage.generateKeyingSheetDataButton().click();
+
+    cy.url().should('eq', relative(`/utilisation-reports/${REPORT_ID}#keying-sheet`));
+
+    pages.utilisationReportPage.keyingSheetTab.keyingSheetTableRow(FIRST_FEE_RECORD_ID).should('contain', 'TO DO');
+    pages.utilisationReportPage.keyingSheetTab.keyingSheetTableRow(SECOND_FEE_RECORD_ID).should('contain', 'TO DO');
+
+    const EXPECTED_PRINCIPAL_BALANCE_ADJUSTMENT = 20000; // 200000 - 180000, DECREASE
+    const EXPECTED_FIXED_FEE_ADJUSTMENT = 7100; // 8100 - 1000, INCREASE
```
RLavender98, line 134: Please can these be lower case since they are just used within the one test?

RLavender98, line 134: Please could we include more detail to where the 8100 comes from in case this ever fails and someone less knowledgable about keying sheet generation needs to try and figure out the failure?

RLavender98 (commented):
Looks good, thanks for adding and nice job making the set up pretty easy to read!

e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/generate-keying-sheet.spec.js
JoshBinns2000: I didn't actually think of that - my brain was in a functional world when I was writing this so I forgot about the side effect issues. Will re-work this

RLavender98 (approved):

RLavender98 (approved):

RLavender98 (approved):

PR #3424, opened 2024-08-28
feat(FN-2310): display payment link on payment details page

JoshBinns2000:
## Introduction :pencil2:
We want to render payments not at the `RECONCILED` or `READY_TO_KEY` status as links to the `edit-payment` page.

- Updates the template
- Updates the mapping function to retain the payment id (so that the correct link can be generated)
- Updates tests


trade-finance-manager-ui/templates/utilisation-reports/_macros/payment-details-table-row.njk
```diff
@@ -1,13 +1,28 @@
...
   {% set payment = paymentDetails.payment %}
   {% set feeRecords = paymentDetails.feeRecords %}
   {% set reconciledBy = paymentDetails.reconciledBy or "-" %}
   {% set dateReconciled = paymentDetails.dateReconciled or "-" %}
+  {% set paymentAmountHtml %}
+  {% if userCanEdit and groupStatus !== 'READY_TO_KEY' and groupStatus !== 'RECONCILED' %}
+    <a href="/utilisation-reports/{{ reportId }}/edit-payment/{{ payment.id }}">
+      {{ payment.amount.formattedCurrencyAndAmount }}
+    </a>
+  {% else %}
```
RLavender98, line 15: I think I would prefer it if the logic of which to display as a link lived in the javascript, and then the view model had a field on paymentDetails that is the `editPaymentLink` which is either a string or undefined and we make it a link if it is defined and the userCanEdit here? That way we can use the status constant for the logic rather than hardcoding it here. What do you think, is there a reason for preferring this?

JoshBinns2000: Both approaches are valid and I don't have a preference one way or the other. The reason for having the status check at this point was purely to match the way our existing templates perform this logic (we do a similar thing on the premium payments table for rendering links), so I think we should move both sets of logic to the view model if we prefer the `editPaymentLink?: string | undefined` approach to keep this consistent

RLavender98 (approved):
LGTM - Happy for you to merge once you also have approval from someone from UKEF

trade-finance-manager-ui/component-tests/utilisation-reports/_macros/payment-details-table-row.component-test.ts
```diff
@@ -75,11 +80,71 @@ describe(component, () => {
...
+
+        wrapper.expectElement('tr td:contains(a)').toExist();
+        wrapper.expectLink('td a').toLinkTo('/utilisation-reports/12/edit-payment/24', 'GBP 123.45');
+      },
+    );
+  });
+
+  describe('when userCanEdit is set to false', () => {
+    const userCanEdit = false;
+
```
abhi-markan, line 131: Should we add assertions for `it.each([FEE_RECORD_STATUS.READY_TO_KEY, FEE_RECORD_STATUS.RECONCILED])(...` when `userCanEdit` is false?

trade-finance-manager-ui/server/controllers/utilisation-reports/helpers/reconciliation-for-report-helper.ts
```diff
@@ -236,7 +237,7 @@ export const mapFeeRecordPaymentGroupsToPaymentDetailsViewModel = (feeRecordPaym
     {},
   );
 
-  return feeRecordPaymentGroups.reduce((paymentDetails, { feeRecords, paymentsReceived }) => {
+  return feeRecordPaymentGroups.reduce((paymentDetails, { feeRecords, paymentsReceived, status }) => {
```
abhi-markan, line 240: Can we rename `status` to `feeRecordPaymentGroupStatus`, so the return object is clean.

abhi-markan (changes requested):
Only few minor comments to address please.

trade-finance-manager-ui/component-tests/utilisation-reports/_macros/payment-details-table-row.component-test.ts
JoshBinns2000: We do - we have `it.each(Object.values(FEE_RECORD_STATUS))` which includes all the fee record statuses

trade-finance-manager-ui/server/controllers/utilisation-reports/helpers/reconciliation-for-report-helper.ts
JoshBinns2000: I have performed the rename in the destructuring of the type, but I would prefer not to rename the `FeeRecordPaymentGroup` `status` property itself. The reason to call it `feeRecordPaymentGroupStatus` in this context is that the individual payment does not have a "status", but the group it is in does

abhi-markan (approved):
LGTM 👍🏼

PR #3425, opened 2024-08-28
feat(FN-3400): add tfm field to tfm facility seeding

RLavender98:
## Introduction :pencil2:
An error was being thrown on feature because of inserted facilities with no tfm object

- Add tfm field to TfmFacility type
- Set tfm field to empty object for test data and facility seeding (appears to be the default as far as I cantell)


libs/common/src/types/mongo-db-models/tfm-facilities.ts
```diff
@@ -121,4 +121,5 @@ export type TfmFacility = {
   _id: ObjectId;
   facilitySnapshot: Facility;
   amendments?: TfmFacilityAmendment[];
+  tfm: object;
```
MarRobSoftwire, line 124: Do we want to add any more details to the type or is the plan to incrementally update this?

MarRobSoftwire (approved):
LGTM

trade-finance-manager-api/src/v1/rest-mappings/mappings/facilities/getLatestFacilityEndDateValues.test.ts
```diff
@@ -37,6 +38,7 @@ describe('getLatestFacilityEndDateValues', () => {
           { ...MOCK_AMENDMENT, updatedAt: 1723653111, version: 1 },
           { ...MOCK_AMENDMENT, updatedAt: 1723653222, version: 2 },
         ],
+        tfm: {},
```
ttbarnes, line 41: I think this can be removed since `facility` (line 8/35) has a TFM object?

libs/common/src/types/mongo-db-models/tfm-facilities.ts
ttbarnes: Would be great if we could add some types to this as the other existing types seem to be fully populated 🙏

RLavender98: The code regarding this tfm object is quite hard to follow and I am not close to this at all, so felt best to leave it vague and then incrementally improve when someone with more context has time to do so. If either can explain the purpose of this field, the possible fields it can have and their types then that would be really helpful!

trade-finance-manager-api/src/v1/rest-mappings/mappings/facilities/getLatestFacilityEndDateValues.test.ts
RLavender98: Ah whoops, thanks!

libs/common/src/types/mongo-db-models/tfm-facilities.ts
JoshBinns2000: For reference, adding the full `TfmFacilityAmendment` type took about a week of running e2e tests and tracking and documenting every field which was sent to the `dtfs-central-api` `PUT /v1/tfm/.../amendments` endpoint. That was technically easier than this `Facility.tfm` type though because it was very straightforward to track down that this was the only endpoint where amendments were modified

JoshBinns2000 (approved):

libs/common/src/types/mongo-db-models/tfm-facilities.ts
MarRobSoftwire: Happy to leave for now :)

ttbarnes: I'm a bit rusty on this, but if I remember correctly, `facility.tfm` is modified/created during deal submission to TFM, before any amendments.

Can see where the initial data is created starting from here => https://github.com/UK-Export-Finance/dtfs2/blob/df67bba1f917bdd9f04717f990b92786e327ca97/trade-finance-manager-api/src/v1/controllers/update-facilities.js#L23

Does look fiddly to obtain the amendments data structure along with this, not a blocker

trade-finance-manager-api/src/v1/rest-mappings/mappings/facilities/getLatestFacilityEndDateValues.test.ts
ttbarnes: Can't see this being removed @RLavender98 - not sure if it's a Github issue. Happy for this to be done in the next PR since it's very minor.

ttbarnes (approved):

ttbarnes (approved):

PR #3429, opened 2024-08-29
feat(FN-3389): add docstrings to event handlers and entities

JoshBinns2000:
## Note
This PR is primarily for documentation with minimal modifications to actual implementation. Any comments around implementation are therefore outside the scope of this PR.

## Introduction :pencil2:
We want to add docstrings to event handlers and typeORM entities.

- Adds docstrings to the event handlers
- Adds docstrings to the entities

### DIFF
--- a/libs/common/src/sql-db-entities/partial-entities/month-and-year.partial-entity.ts
+++ b/libs/common/src/sql-db-entities/partial-entities/month-and-year.partial-entity.ts
@@ -1,15 +1,16 @@
 import { Column } from 'typeorm';
+import { OneIndexedMonth } from '../../types';
 
 export class MonthAndYearPartialEntity {
   /**
-   * The month
+   * The month (one-indexed)
    */
-  @Column()
-  month!: number;
+  @Column({ type: 'int' })
+  month!: OneIndexedMonth;
 
   /**
    * The year
    */
-  @Column()
+  @Column({ type: 'int' })
   year!: number;
 }
--- a/libs/common/src/sql-db-entities/partial-entities/month-and-year.partial-entity.ts
+++ b/libs/common/src/sql-db-entities/partial-entities/month-and-year.partial-entity.ts
@@ -1,9 +1,15 @@
 import { Column } from 'typeorm';
 
 export class MonthAndYearPartialEntity {
+  /**
+   * The month
+   */
   @Column()
   month!: number;
 
+  /**
+   * The year
+   */
   @Column()
   year!: number;
 }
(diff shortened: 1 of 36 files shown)


libs/common/src/sql-db-entities/utilisation-report/utilisation-report.entity.ts
```diff
@@ -82,10 +91,18 @@ export class UtilisationReportEntity extends AuditableBaseEntity {
...
+   * Updates the report with fee records
+   * @param param - The parameters to update the report with
+   */
   public updateWithFeeRecords({ feeRecords }: UpdateWithFeeRecordsParams): void {
     this.feeRecords = feeRecords;
   }
 
+  /**
+   * Updates the report with a supplied status
+   * @param param - The parameters to update the report with
```
abhi-markan, line 104: Can the parameters names please be reflected, I would encourage adding `@returns` as well if below snippet can be used as a reference please that will be great. Specifying types in TSDocs is optional.

```typescript
/**
 * Updates the report with a supplied status.
 *
 * @param {UpdateWithStatusParams} params - An object containing the status and request source.
 * @param {string} params.status - The new status to update.
 * @param {string} params.requestSource - The source of the update request.
 * @returns {void} - This method does not return a value.
 */
```

libs/common/src/sql-db-entities/partial-entities/month-and-year.partial-entity.ts
```diff
@@ -1,9 +1,15 @@
 import { Column } from 'typeorm';
 
 export class MonthAndYearPartialEntity {
+  /**
```
abhi-markan, line 4: Minor touches to the type.

```typescript
/**
   * The year
   * @required true
   * @example 2024
   */
```

Is the month stored in JS month number (0 - Jan, 11 - Dec) or standard (1-12), an explanation on that will be great. Thanks

abhi-markan (changes requested):
Nice work, minor comments to address please.

libs/common/src/sql-db-entities/partial-entities/month-and-year.partial-entity.ts
JoshBinns2000: As far as I can see, `@required` is not a JSDoc tag. I will make the change documenting it as a one-indexed month though

abhi-markan: Much cleaner, thanks.

libs/common/src/sql-db-entities/utilisation-report/utilisation-report.entity.ts
abhi-markan: Thanks for adding properties to the docstring.
TSDocs reference on `@returns` https://tsdoc.org/pages/tags/returns/, happy to approve on this instance as DocStrings needs to be unified under an established tech-debt ticket.

abhi-markan (approved):

libs/common/src/sql-db-entities/utilisation-report/utilisation-report.entity.ts
JoshBinns2000: That reference highlights that `@returns` documents a return _value_, where `void` is designed to document the absence or a return value, hence the distinction in this case

PR #3434, opened 2024-08-30
feat(FN-2594): disable input before bank is selected

JoshBinns2000:
## Introduction :pencil2:
We want to disable the year input before a bank is selected.

- Updates the script
- Updates the e2e test


e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/find-utilisation-reports-by-bank-and-year.spec.js
```diff
@@ -85,14 +85,16 @@ context('PDC_RECONCILE users can search for reports by bank and year', () => {
     pages.searchUtilisationReportsResultsPage.noReportsText().should('exist');
   });
 
-  it('should display the dropdown for banks which have reports and should not display anything for banks without reports', () => {
+  it('should disable the input until a bank is selected and display the dropdown for banks which have reports and should not display anything for banks without reports', () => {
```
ttbarnes, line 88: Could we split this test up? Lots of "ands" in 1x test, thinking something like this?:

```js
- when a bank is not selected
  - it should disable the input

- when a bank is selected
  - it should display the dropdown for banks which have reports
  - it should not display anything for banks without reports
```

ttbarnes (changes requested):
Nice work and great small PR 😎 

e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/find-utilisation-reports-by-bank-and-year.spec.js
JoshBinns2000: Decided to just add a new test for the disable/enable functionality as I'd like to leave the datalist selection all inside of one test

ttbarnes (approved):

PR #3447, opened 2024-09-02
feat(FN-2406): use common report data validation for upload

RLavender98:
## Introduction :pencil2:
In a previous PR we moved the validation of report data to dtfs-central-api for the 'check the report page'.
We should use the same validation for the upload journey since:
- The new validation is slightly stricter than the previous validation in dtfs-central-api so this will ensure consistency across the validation
- We will only have to update one set of validation rules moving forwards as rules change

- Replace the report data validation in the upload utilisation report payload validator with the newly moved stricter report data validation logic from portal-api.

- Moved the payload validation to the `routes/middleware` folder where the other payload validators are defined


sw-mattw (dismissed):
Nice, looks good, needs UKEF a review though

dtfs-central-api/src/v1/routes/middleware/payload-validation/validate-post-upload-utilisation-report-payload.test.ts
```diff
@@ -0,0 +1,128 @@
...
+      });
+      jest.mocked(validateUtilisationReportCsvData).mockReturnValue([{ errorMessage: 'That is not valid report data' }]);
+
+      // Act
+      validatePostUploadUtilisationReportPayload(req, res, mockNext);
+
+      // Assert
+      expect(mockNext).not.toHaveBeenCalled();
+      expect(res._getStatusCode()).toBe(HttpStatusCode.BadRequest);
+    });
```
ttbarnes, line 126: Can we add a test to check the validation errors are returned in the response? e.g

```ts
expect(res.send).toHaveBeenCalledWith(expected);
```

dtfs-central-api/src/v1/routes/middleware/payload-validation/validate-post-upload-utilisation-report-payload.ts
```diff
@@ -0,0 +1,49 @@
...
+      return { ...rowData, [key]: { value: reportDataRow[key], row: rowIndex + 2 } };
+    }, {} as UtilisationReportCsvRowData);
+  });
+
+  const validationErrors = [
+    validateReportId(reportId),
+    ...validateFileInfo(fileInfo),
+    ...validateUtilisationReportCsvData(reportDataMapped).map(({ errorMessage, value, row }) => ({ errorMessage, value, row })),
+    ...validateReportUser(user),
+  ].filter(Boolean);
```
ttbarnes, line 41: Can we extract `reportDataMapped` and the `validationErrors` mapping and filtering logic into their own standalone functions? As there's a fair amount going on here. Would also then be able to write unit tests for these and make the middleware and it's unit test simpler

ttbarnes (changes requested):
Great work 🎉 Just a couple of comments

ttbarnes (approved):

Zainzzkk (approved):

PR #3433, opened 2024-08-30
chore(DTFS2-7334): refactor e2e test feature flagging

MarRobSoftwire:
## Introduction :pencil2:
After a discussion we're changing our approach to testing feature flags going forward.

## Resolution :heavy_check_mark:
Remove all FF references in GEF e2e tests

## Note ⚠️ 
- This is branched off #3409 and probably shouldn't be reviewed until that has been merged in

PR #3448, opened 2024-09-02
chore(DTFS2-7335): remove GEF deal version regression tests

MarRobSoftwire:
## Introduction :pencil2:
When implementing deal versioning, we added lots of tests to check the existing behaviour was maintained when  GEF_DEAL_VERSION is set to 0 (much like a feature flag).

Remove the deal version 0 regression tests

## Note ⚠️ 
- This **should not be merged in** until the feature has gone live & we're happy we dont want to rollback any changes
- The files `version-0-facility.spec` and `version-0-unissued-facility.spec` capture behaviour we want to keep, even with `GEF_DEAL_VERSION=1` and should not be removed
- This is branched off #3433 and probably shouldn't be reviewed until that has been merged in


MarRobSoftwire:
Closing due to merge conflicts on #3433

PR #3449, opened 2024-09-02
feat(FN-3417): add batch saving for facility utilisation data

JoshBinns2000:
## Introduction :pencil2:
There was an issue with uploading reports with a large number of new facilities, where new refers to a facility which does not have an entry in the `FacilityUtilisationData` table.

- Adds a new api test to test for a large number of facilities
  - This api test was run before and after the new changes were included to replicate the existing issue and ensure it remains working in future
- Updates the event handler

- Adds a missing docstring

### DIFF
--- a/dtfs-central-api/api-tests/v1/utilisation-reports/post-upload-utilisation-report.api-test.ts
+++ b/dtfs-central-api/api-tests/v1/utilisation-reports/post-upload-utilisation-report.api-test.ts
@@ -60,7 +60,7 @@ describe(`POST ${getUrl()}`, () => {
     const response = await testApi.post(payload).to(getUrl());
 
     // Assert
-    expect(response.status).toBe(HttpStatusCode.NotFound);
+    expect(response.status).toEqual(HttpStatusCode.NotFound);
   });
 
   it("responds with a 201 (Created) with a valid payload and sets the report status to 'PENDING_RECONCILIATION'", async () => {
@@ -68,7 +68,7 @@ describe(`POST ${getUrl()}`, () => {
     const response = await testApi.post(aValidPayload()).to(getUrl());
 
     // Assert
-    expect(response.status).toBe(HttpStatusCode.Created);
+    expect(response.status).toEqual(HttpStatusCode.Created);
 
     const report = await SqlDbHelper.manager.findOneByOrFail(UtilisationReportEntity, { id: reportId });
     expect(report.status).toBe<UtilisationReportReconciliationStatus>('PENDING_RECONCILIATION');
@@ -88,10 +88,10 @@ describe(`POST ${getUrl()}`, () => {
     const response = await testApi.post(payload).to(getUrl());
 
     // Assert
-    expect(response.status).toBe(HttpStatusCode.Created);
+    expect(response.status).toEqual(HttpStatusCode.Created);
 
     const feeRecordCount = await SqlDbHelper.manager.count(FeeRecordEntity, {});
-    expect(feeRecordCount).toBe(3);
+    expect(feeRecordCount).toEqual(3);
   });
 
   it('creates an entry in the FacilityUtilisationData table for each facility id in the report csv data', async () => {
@@ -108,7 +108,7 @@ describe(`POST ${getUrl()}`, () => {
     const response = await testApi.post(payload).to(getUrl());
 
     // Assert
-    expect(response.status).toBe(HttpStatusCode.Created);
+    expect(response.status).toEqual(HttpStatusCode.Created);
 
     const facilityIdExists = await Promise.all(
       facilityIds.map((facilityId) => SqlDbHelper.manager.existsBy(FacilityUtilisationDataEntity, { id: facilityId })),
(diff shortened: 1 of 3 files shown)


RLavender98 (dismissed):

dtfs-central-api/api-tests/v1/utilisation-reports/post-upload-utilisation-report.api-test.ts
```diff
@@ -127,13 +127,37 @@ describe(`POST ${getUrl()}`, () => {
...
   });
 
-  it('saves the report when there is a large number of fee records', async () => {
+  it('saves the report when there is a large number of fee records all for the same facility', async () => {
     // Arrange
     const numberOfReportDataEntriesToCreate = 500;
 
+    const facilityId = '12345678';
+    const reportData: UtilisationReportRawCsvData[] = [];
+    while (reportData.length < numberOfReportDataEntriesToCreate) {
```
Zainzzkk, line 136: could this be in a test helper

dtfs-central-api/api-tests/v1/utilisation-reports/post-upload-utilisation-report.api-test.ts
```diff
@@ -127,13 +127,37 @@ describe(`POST ${getUrl()}`, () => {
...
+
+    const payload: PostUploadUtilisationReportRequestBody = { ...aValidPayload(), reportData };
+
+    // Act
+    const response = await testApi.post(payload).to(getUrl());
+
+    // Assert
+    expect(response.status).toBe(HttpStatusCode.Created);
+
+    const numberOfInsertedFeeRecords = await SqlDbHelper.manager.count(FeeRecordEntity, {});
```
Zainzzkk, line 148: can this be in a different it statement please

dtfs-central-api/api-tests/v1/utilisation-reports/post-upload-utilisation-report.api-test.ts
JoshBinns2000: I don't think there is a reason to have a test helper for a single array population (YAGNI)

dtfs-central-api/api-tests/v1/utilisation-reports/post-upload-utilisation-report.api-test.ts
JoshBinns2000: I'd rather keep this all in the same test, but this assertion doesn't actually need to be in this test so I have removed it

JoshBinns2000: EDIT: I have decided to keep these assertions in as they come under the statement of "saves the report". Please let me know if you disagree

Zainzzkk: I think it should be different it statements.  If one fails, it is easier to see exactly which is failing from the it statement description

JoshBinns2000: Jest allows you to see exactly which part of the test failed. In the case where you don't get a `201`, that will be very clear from the test output. Similarly, in the case where you do get a `201`, it would be clear that it was the number of fee records inserted which caused this to fail.

I actually think having different `it` blocks would make things less clear. For instance, the it block which says "saves the report" may pass if a `201` is returned, but the below test block which asserts that the correct number of fee records is saved may fail. This is not semantically accurate as the return of a `201` accompanied by saving an incorrect number of fee records implies that "saving the report" did not actually work successfully, but the separate `it` blocks may cause the debugger to believe that "saving the report" was successful and so the error must be elsewhere

dtfs-central-api/api-tests/v1/utilisation-reports/post-upload-utilisation-report.api-test.ts
Zainzzkk: I think that this still should be in a test helper but can be done later, resolving

dtfs-central-api/api-tests/v1/utilisation-reports/post-upload-utilisation-report.api-test.ts
Zainzzkk: This still should be split.  It allows for future proofing as well incase we wanted to add more tests for this bit.  we can change the it statement to make more sense - eg it('returns a status of ${HttpStatusCode.Created}' ... ) and it('should have correct number of records...).  Can have a describe around it too.

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/report-uploaded/report-uploaded.event-handler.ts
```diff
@@ -33,6 +42,8 @@ const createFacilityUtilisationDataEntityIfNotExists = async (
   return FacilityUtilisationDataEntity.createWithoutUtilisationAndFixedFee({ id: facilityId, reportPeriod, requestSource });
 };
 
+const CHUNK_SIZE_FOR_BATCH_SAVING = 100;
```
Zainzzkk, line 45: maybe this can be done in the next PR, but can this be moved to constants?

JoshBinns2000: This constant only makes sense in the context of this file and exists purely for readability. Outside the context of this file it would have no meaning, and it is likely that another place which wants to use this would want a different batch size (it depends on many features which are implementation specific like the number of columns, relations etc.), so I don't think we should be adding this to the `constants` directory

Zainzzkk: even if this is only being used once, it should be stored in the constants directory as it is a constant.  It makes it easier to manage/change.  In the future, if we use a shared package for all UKEF constants, it would have to be in there.
PLease can this be moved in the next PR.

PR #3450, opened 2024-09-02
feat(FN-2069): align columns across all report reconciliation tables on bank reports page

sw-mattw:
## Introduction :pencil2:
The columns in all report reconciliation tables on the bank reports page should be properly aligned with each other for consistency and readability.

**UI prior to this PR** (note that this is just to illustrate the alignment issue, the data shown has changed since this screenshot):

FN-2069_PR_prior-design

- Styling changes:
  - Fixed width columns (percentage)
  - Table placed within a scrollable pane

Note that changes to the files within `public` are a result of running `npm run build` in the TFM UI directory to recompile the CSS/JS.

Designs with horizontally scrollable pane approved.

**Screenshots:**

**Full width browser, 16pt font:**

FN-2069_scrollable-pane-approach_full-width

**Full width browser, 40pt font:**

FN-2069_scrollable-pane-approach_full-width_40pt-font

**Narrow browser, 16pt font, (note that the scrollable pane is horizontally scrollable):**

FN-2069_scrollable-pane-approach_narrow

N/A


trade-finance-manager-ui/styles/table.scss
```diff
@@ -156,4 +156,26 @@ table.premium-payments-table {
...
+    width: 25% !important;
+  }
+}
+@media (max-width:65em) {
+  table.utilisation-report-reconciliation-table {
+    width: 55em;
+
+    .govuk-table__header,
+    .govuk-table__cell {
+      white-space: normal;
```
sw-mattw, line 178: We require this to override the `nowrap` value the scrollable pane applies (which causes elements to overlap)

trade-finance-manager-ui/styles/table.scss
```diff
@@ -156,4 +156,26 @@ table.premium-payments-table {
...
   }
-}
\ No newline at end of file
+}
+
+table.utilisation-report-reconciliation-table {
+  table-layout: fixed;
+
+  .ukef-width-one-tenth {
+    width: 10% !important;
```
sw-mattw, line 165: These live inside the table (vs in `layout.css`) as they shouldn't be used for general grid styling

trade-finance-manager-ui/styles/table.scss
```diff
@@ -156,4 +156,26 @@ table.premium-payments-table {
...
+  }
+
+  .ukef-width-one-quarter {
+    width: 25% !important;
```
ttbarnes, line 169: Do we really need `!important`, I suspect so so that it overrides a design library class?

trade-finance-manager-ui/styles/table.scss
```diff
@@ -156,4 +156,26 @@ table.premium-payments-table {
...
+@media (max-width:65em) {
+  table.utilisation-report-reconciliation-table {
+    width: 55em;
+
+    .govuk-table__header,
+    .govuk-table__cell {
+      white-space: normal;
+    }
+  }
+}
```
ttbarnes, line 181: Since we're using SCSS we can save some lines here by nesting the media query in the selector like this:

```css
table.utilisation-report-reconciliation-table {
  table-layout: fixed;

  /* ... */

  @media (max-width:65em) {
    width: 55em;

    .govuk-table__header,
    .govuk-table__cell {
      white-space: normal;
    }
  }
}
```

ttbarnes (changes requested):
Great PR, 2x minor comments otherwise LGTM

RLavender98 (commented):
Nothing to add to Tony's comments - consider this E2ETF team approval for once those comments are resolved!

trade-finance-manager-ui/styles/table.scss
sw-mattw: Removed and retested - works as it did previously. Thanks for mentioning

(I think this was only required for a previous approach)

trade-finance-manager-ui/styles/table.scss
sw-mattw: Cool, I wasn't sure which to adopt due to `layout.scss` using the other way. Changed this now in line with your suggestion thanks

ttbarnes (approved):

PR #3453, opened 2024-09-03
chore(DTFS2-7333): dry cypress partials, click event commands - tfm

ttbarnes:
## Introduction :pencil2:
This PR introduces some common partials and click events used throughout the TFM E2E suite.

- Update the partials file for the following selectors: 
  - `backLink`
  - `continueButton`
  - `errorSummary`
  - `errorSummaryItems`
  - `submitButton`
- Remove the above selectors from individual page selectors.
- Create new click event cypress commands for common buttons and links: 
  - `clickBackLink`
  - `clickContinueButton`
  - `clickSubmitButton`
- Update all TFM E2E tests.

- Align some `data-cy` values.
- Various partials destructuring improvements.


e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/remove-fees-from-payment.spec.js
```diff
@@ -87,7 +88,7 @@ context(`${PDC_TEAMS.PDC_RECONCILE} users can remove fees from payments`, () =>
 
   it('should display errors when form submitted with invalid fee selections and persist the selected fees and inputted values', () => {
     pages.utilisationReportEditPaymentPage.clickRemoveSelectedPaymentsButton();
-    pages.utilisationReportEditPaymentPage.errorSummary().contains('Select fee or fees to remove from the payment');
+    errorSummary().contains('Select fee or fees to remove from the payment');
```
JoshBinns2000, line 91: It appears that this is being used as an assertion, but `.contains` just yields a DOM element. Could we replace this with `.should('contain', <text>)`

ttbarnes: Completely agree - there are currently 1944 instances of this. Would like to replace all instances of "contains" (whether it's checking the DOM or text), to use "to equal" in a separate ticket/PR, as this PR is only replacing selectors and introducing common commands for click events

JoshBinns2000: I don't think we can just replace all of them - some of them are likely being used to yield the element which contains the text which is the intended usage, whilst others will be incorrectly used as an assertion like this. That is to say that replacing them will require some care and is likely non-trivial, but agree that it is not for this PR

ttbarnes: `.contains` is inherently risky and should not be used, they should all be using equal, otherwise the test is not well written or reliable.

It is a big task though and probably not a high priority

JoshBinns2000: I'm not sure what you mean by "risky" - `.contains` just works in the same was as jQuery, with `$('h1:contains("My title")')` yielding the first `h1` in the DOM which contains the text `My title`. This is fairly standard practice when querying the DOM, so I'm not sure I agree that the use of `contains` as a DOM selector has any link to the test reliability.

To reiterate, in the cases where `.contains` may be used as a DOM selector, replacing it with "equals" would not be correct as this would no longer yield the same DOM element

ttbarnes: It is risky because it is only partially asserting something. For example, given this element:

```html
<p data-cy="a-selector">Deal 1234</p>
```

And this cypress assertion:

```js
aSelector().should('contain', 'Deal 1234');
```

The test will still pass if *anything* but the text changes in the element - for example the test would still pass if the element is changed to:


```html
<p data-cy="a-selector">Deal 1234!</p>
<p data-cy="a-selector"> Deal 1234</p>
<p data-cy="a-selector">This is the Deal 1234</p>
```

Using strict equality checking for the element will cause the test to fail.

JoshBinns2000: Ah, it seems we're talking past each other - I thought you were suggesting that using `.contains` to query the DOM is risky. I agree that we should use more strict checks for the actual assertion than `.should('contain', ...)` in most cases

abhi-markan (dismissed):
Really good work on DRY.

Question: When we merge this, are you okay to update the existing PRs with the latest Cypress commands unless respective devs have capacity to do so.

ttbarnes:
> Really good work on DRY.
> 
> Question: When we merge this, are you okay to update the existing PRs with the latest Cypress commands unless respective devs have capacity to do so.

Thanks @abhi-markan and yes I will ensure the other PRs are updated, have done this for 2x other similar DRY PRs 👍

abhi-markan (approved):

PR #3458, opened 2024-09-04
feat(DTFS2-7294): add cancel deal button and Feature flag

BethThomas141:
## Introduction :pencil2:
- Add the cancel deal button to the deal page on the front end
- Hide it behind a new feature flag `FF_TFM_DEAL_CANCELLATION_ENABLED`
- In addition only display if the deal type is AIN or MIN 
- Currently clicking 'cancel deal' has no functionality - this will be added in later tickets

- E2E tests will be added as a separate PR

image


.github/workflows/test.yml
```diff
@@ -143,6 +143,7 @@ env:
   CONTACT_US_EMAIL_ADDRESS: ${{ vars.CONTACT_US_EMAIL_ADDRESS }}
   FF_TFM_PAYMENT_RECONCILIATION_ENABLED: "true"
   FF_TFM_FACILITY_END_DATE_ENABLED: "true"
+  FF_TFM_DEAL_CANCELLATION: "false"
```
BethThomas141, line 146: I've set it to false here as the default values on the github workflows as I think that's what Abhi wants to move to. It shouldn't impact the unit tests as these mock the value anyway

trade-finance-manager-ui/server/controllers/case/index.test.js
```diff
@@ -7,6 +7,11 @@ import mapAssignToSelectOptions from '../../helpers/map-assign-to-select-options
 
 const res = mockRes();
 
+jest.mock('@ukef/dtfs2-common', () => ({
```
BethThomas141, line 10: I'm not super sure what to do in this test file - the existing tests just check very broad high level things such as 'when the deal exists the deal template is rendered' and doesn't check specific sub cases of values. 

I could break that pattern and add in conditional testing for the `showCancelDealButton` e.g. tests of the form: 

When the TFM Deal cancellation FF is enabled -> 
- If the deal type is AIN then showCancelDealButton should be true
-  If the deal type is MIN then showCancelDealButton should be true
-  If the deal type is MIA then showCancelDealButton should be false

When the TFM Deal cancellation FF is disabled -> 
- If the deal type is AIN then showCancelDealButton should be false
-  If the deal type is MIN then showCancelDealButton should be false
-  If the deal type is MIA then showCancelDealButton should be false

But I feel this logic is tested in unit tests, and ultimately will be tested in E2E tests too. 
It still needs some mock of `isTfmDealCancellationFeatureFlagEnabled` hence this at the top of the file. 

What do you think is best here?

trade-finance-manager-ui/server/helpers/deal-cancellation-enabled.test.ts
```diff
@@ -0,0 +1,55 @@
...
+
+    it('should return false if the deal type is MIA', () => {
+      const result = dealCancellationEnabled(DEAL_SUBMISSION_TYPE.MIA);
+
+      expect(result).toEqual(false);
+    });
+
+    it('should return true if the deal type is MIN', () => {
+      const result = dealCancellationEnabled(DEAL_SUBMISSION_TYPE.MIN);
+
```
BethThomas141, line 21: Do you know how to resolve these TS errors?

docker-compose.yml
```diff
@@ -124,6 +124,7 @@ services:
       - CONTACT_US_EMAIL_ADDRESS
       - FF_TFM_PAYMENT_RECONCILIATION_ENABLED
       - FF_TFM_FACILITY_END_DATE_ENABLED
+      - FF_TFM_DEAL_CANCELLATION
```
BethThomas141, line 127: I can also pre-emptively add this to tfm-api (& central-api?) if we expect we'll be using them in either place?

.github/workflows/test.yml
AlexBramhill: We've got a new pattern for working with feature flags -- see https://github.com/UK-Export-Finance/dtfs2/pull/3390/commits. Might be worth having a chat with Abhi about how to get this up to the new pattern (I think that should be done in this PR -- and I do think that we should look at maybe creating the e2e-test here for the button to build up over the next few tickets so we can establish the new FF e2e pattern)

Of course -- I'm not dead set on this and let me know if you disagree or want to chat about ti

docker-compose.yml
AlexBramhill: That's a really good point, I think either way is fine -- it might be easier to do this as soon as we create a new route in tfm-ui (or anywhere else), as we want to get common middleware in place for this...

.github/workflows/test.yml
AlexBramhill: Youll need to add this to the docker files too :)

libs/common/src/constants/deal-submission-type.ts
```diff
@@ -0,0 +1,5 @@
+export const DEAL_SUBMISSION_TYPE = {
```
AlexBramhill, line 1: Very glad to see this has now made it here! Thank you!

trade-finance-manager-ui/templates/case/deal/deal.njk
```diff
@@ -113,4 +114,16 @@
     </div>
   </div>
 
+  {% if showDealCancelButton %}
+    <div class="govuk-!-padding-top-7">
+      {{ govukButton({
+        text: "Cancel deal",
+        classes: "govuk-button--warning",
+        attributes: {
+          'data-cy': 'cancel-deal-button'
```
AlexBramhill, line 123: I think I'd be interested for you to chat to Tony about how he'd approach naming this button in terms of data-cy. We're looking to name these more genericly -- but I wonder if this can be a specific button.

trade-finance-manager-ui/templates/case/deal/deal.njk
```diff
@@ -113,4 +114,16 @@
     </div>
   </div>
 
+  {% if showDealCancelButton %}
+    <div class="govuk-!-padding-top-7">
```
AlexBramhill, line 118: Thank you for including a screenshot on this -- I was a little worried how large this padding was but it allowed me to cross reference the designs and see everythings fine!

trade-finance-manager-ui/server/helpers/deal-cancellation-enabled.test.ts
```diff
@@ -0,0 +1,55 @@
+import * as ukefCommon from '@ukef/dtfs2-common';
+import { DEAL_SUBMISSION_TYPE } from '@ukef/dtfs2-common';
+import { dealCancellationEnabled } from './deal-cancellation-enabled';
+
+describe('dealCancellationEnabled', () => {
```
AlexBramhill, line 5: I'd be interested for you to have a conversation with Abhi about how we want to set up these files as I know this is how we have traditionally mocked it but I think we have a new approach, and it'd be good to get notes on the approach that I can add to a living document on testing around feature flags :)

trade-finance-manager-ui/server/helpers/deal-cancellation-enabled.test.ts
AlexBramhill: I am not seeing a ts error?

trade-finance-manager-ui/server/helpers/deal-cancellation-enabled.test.ts
```diff
@@ -0,0 +1,55 @@
...
+  const isTfmDealCancellationFeatureFlagEnabledSpy = jest.spyOn(ukefCommon, 'isTfmDealCancellationFeatureFlagEnabled');
```
AlexBramhill, line 6: I've actually not often seen spy be used for this -- Is there a reason to favour it? Personally, I think a `mock` is probably the right choice here as we're `mocking` it rather than `spying` on it, but keen to understand!

trade-finance-manager-ui/server/helpers/deal-cancellation-enabled.test.ts
AlexBramhill: Just a heads up -- I'll review this more once this conversation is had, as I feel that this might change a few things!

trade-finance-manager-ui/server/helpers/deal-cancellation-enabled.test.ts
```diff
@@ -0,0 +1,55 @@
...
+      const result = dealCancellationEnabled(DEAL_SUBMISSION_TYPE.MIN);
+
+      expect(result).toEqual(false);
+    });
+
+    it('should return false if the deal type is AIN', () => {
+      const result = dealCancellationEnabled(DEAL_SUBMISSION_TYPE.AIN);
+
+      expect(result).toEqual(false);
+    });
```
AlexBramhill, line 53: We could combine these all in an `it.each([MIA,MIN,AIN])` :)

trade-finance-manager-ui/server/controllers/case/index.test.js
AlexBramhill: Is this an api test or a unit test? It seems like it wants to be a unit test (as per its name), so I think the only thing we care for is what this file sees, and no further -- ie it does not care about implimentation of ` dealCancellationEnabled(deal.dealSnapshot.submissionType)`, and therefore I dont think it cares about the feature flag at all.

If I were to approach this, I'd look at mocking `dealCancellationEnabled(deal.dealSnapshot.submissionType)`, asserting that `dealCancellationEnabled` is called with the expected value, and any value it returns (which we mock in the test) is passed on... How does that sound?

So we could have `it('should check whether deal cancellation is enabled'` and a `it('should render the template with the deal cancellation value')` or something similar -- the namings slightly off there!

trade-finance-manager-ui/component-tests/case/deal/deal.component-test.js
```diff
@@ -188,3 +188,29 @@ describe(`${page} when deal is NDB`, () => {
     wrapper.expectElement('[data-cy="bank-security-facility-text"]').toExist();
   });
 });
+
+describe(`${page} cancel deal button`, () => {
```
AlexBramhill, line 192: Nice job

AlexBramhill (commented):
Some initial comments

trade-finance-manager-ui/server/helpers/deal-cancellation-enabled.test.ts
MarRobSoftwire: It's Codacy being broken in TS :(

trade-finance-manager-ui/server/helpers/deal-cancellation-enabled.ts
```diff
@@ -0,0 +1,5 @@
+import { DEAL_SUBMISSION_TYPE, dealSubmissionType, isTfmDealCancellationFeatureFlagEnabled } from '@ukef/dtfs2-common';
+
+export const dealCancellationEnabled = (submissionType: dealSubmissionType): boolean => {
+  return isTfmDealCancellationFeatureFlagEnabled() && (submissionType === DEAL_SUBMISSION_TYPE.AIN || submissionType === DEAL_SUBMISSION_TYPE.MIN);
```
MarRobSoftwire, line 4: Think its something to cover in a future ticket (once we've implemented cancelling!) but we'll want to add `deal.tfm.stage !== 'Cancelled` (or however we do this)

libs/common/src/types/tfm/deal-submission-type.ts
```diff
@@ -0,0 +1,4 @@
+import { DEAL_SUBMISSION_TYPE } from '../../constants';
+import { ValuesOf } from '..';
+
+export type dealSubmissionType = ValuesOf<typeof DEAL_SUBMISSION_TYPE>;
```
MarRobSoftwire, line 4:
```suggestion
export type DealSubmissionType = ValuesOf<typeof DEAL_SUBMISSION_TYPE>;
```

trade-finance-manager-ui/templates/case/deal/deal.njk
MarRobSoftwire: I think this shouldnt be confused with a `cancel` link & we dont want to commonise it into that.
I think this is its own unique component & can be accessed from the deal page file

trade-finance-manager-ui/templates/case/deal/deal.njk
BethThomas141: Ah awesome, I found this Gov spacing guide https://design-system.service.gov.uk/styles/spacing/ . Our designs have the spacing on Figma as 40px, which is equivalent to using `padding-top-7` on large screens, but it will be less padded on smaller screens. The alternative was `padding-top-9` so that the designs align with small screens instead, but then it does look a big huge on large screens so I think this is the better option.

MarRobSoftwire (commented):
Looks really good :)

trade-finance-manager-ui/server/helpers/deal-cancellation-enabled.ts
BethThomas141: Ooh yes good point. I can add a TODO with ticket number DTFS2-7298?

libs/common/src/constants/deal-submission-type.ts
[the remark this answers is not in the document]
BethThomas141: It's a good question - there are a lot of uses of deal types across the services - for instance in `portal-api/src/constants/deal.js` we have `SUBMISSION_TYPE` with the exact same values already. If we replace all uses across the files where that's used it would be a lot of files changed. I guess a quick way around it so that it is still using the new common values but doesn't impact lots of files would be to simply e.g. put: 

`const SUBMISSION_TYPE = DEAL_SUBMISSION_TYPE;` in `portal-api/src/constants/deal.js`

so at least the values aren't being repeated. But not sure what we think about that. @AlexBramhill do you have thoughts on this?

trade-finance-manager-ui/server/helpers/deal-cancellation-enabled.test.ts
BethThomas141: No that makes sense - so I originally did my usual mocking, but it was causing all sorts of TS errors on my IDE, so I did a hunt around and found other test files mocking feature flags in TS files and found a lot of instances with `isTfmPaymentReconciliationFeatureFlagEnabledSpy` - so thought (in potentially in a moment of madness) that maybe we do things differently in TS tests... but I wasn't super convinced by what I was doing so glad you've flagged.

I think I'm concluding that my IDE needs some reconfig to stop it showing non-existent TS errors! :)

trade-finance-manager-ui/server/helpers/deal-cancellation-enabled.test.ts
MarRobSoftwire: ooh I'll take a look if its a genuine thing, just so used to codacy i assumed that was it :(

MarRobSoftwire: Yeah looks fine locally for me - probably just ts/eslint server not recognising a new file has been made

.github/workflows/test.yml
BethThomas141: Oh yep, didn't realise Abhi already had a PR open for it. I'll look into it!

docker-compose.yml
BethThomas141: Sure, I'll leave it for now

trade-finance-manager-ui/server/helpers/deal-cancellation-enabled.test.ts
BethThomas141: Have switched it back to the usual mocking

trade-finance-manager-ui/server/helpers/deal-cancellation-enabled.test.ts
BethThomas141: We absolutely could, have now done that thanks! :)

trade-finance-manager-ui/templates/case/deal/deal.njk
AlexBramhill: I agree

libs/common/src/constants/deal-submission-type.ts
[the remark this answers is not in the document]
BethThomas141: For the sake of keeping this PR small I'm going to do another small PR afterwards that refactors some of the submission type declarations in various services to use the new libs/common values.

.github/workflows/test.yml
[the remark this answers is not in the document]
BethThomas141: Discussed elsewhere, leaving updating them as these are out of date

trade-finance-manager-ui/server/controllers/case/index.test.js
BethThomas141: Great point, I've now added unit tests to this effect 👍

trade-finance-manager-ui/templates/case/deal/deal.njk
BethThomas141: @ttbarnes Are you happy with this `data-cy` attribute naming? It's not named generically, but it is important to distinguish this 'cancel deal' button from general 'cancel' links

trade-finance-manager-ui/server/controllers/case/index.ff-test.js
```diff
@@ -1,13 +1,8 @@
-import { AMENDMENT_STATUS, isTfmFacilityEndDateFeatureFlagEnabled } from '@ukef/dtfs2-common';
+import { AMENDMENT_STATUS } from '@ukef/dtfs2-common';
 import caseController from '.';
 import api from '../../api';
 import { mockRes } from '../../test-mocks';
 
-jest.mock('@ukef/dtfs2-common', () => ({
-  ...jest.requireActual('@ukef/dtfs2-common'),
-  isTfmFacilityEndDateFeatureFlagEnabled: jest.fn(),
-}));
```
BethThomas141, line 9: Removed this mock as I don't think it's needed now we have the new `ff` file name way of adding feature flag tests

.github/workflows/test.yml
BethThomas141: Have refactored to the new pattern. I'm going to leave e2e tests out of this PR to help get this PR in quickly as it's a bit of a blocker on other work. I'll add as a subtask to the ticket to add them in separately.

trade-finance-manager-ui/server/helpers/deal-cancellation-enabled.test.ts
BethThomas141: Have refactored using the new `ff` test file namings as Abhi implemented. I think Abhi will do a call with us all in at some point for the full feature flag low down - agreed would be good do have documentation on the approach, happy to create / contribute to that.

trade-finance-manager-ui/server/controllers/case/index.test.js
```diff
@@ -79,11 +87,44 @@ describe('controllers - case', () => {
...
+
+        expect(res.render).toHaveBeenCalledWith(
+          'case/deal/deal.njk',
+          expect.objectContaining({
+            showDealCancelButton: true,
+          }),
+        );
+      });
+
+      it('should render the template with the deal cancellation value when it is false', async () => {
```
ttbarnes, line 116: For both of these tests I think some describe blocks would be nice:

```js
describe('when deal cancellation is enabled'
  it('should render the template with showDealCancelButton=true'

describe('when deal cancellation is disabled'
  it('should render the template with showDealCancelButton=false'
```

trade-finance-manager-ui/server/controllers/helpers/deal-cancellation-enabled.helper.test.ts
```diff
@@ -0,0 +1,12 @@
...
+
+describe('dealCancellationEnabled', () => {
+  describe('when `FF_TFM_FACILITY_END_DATE_ENABLED` is set to false', () => {
+    it.each([DEAL_SUBMISSION_TYPE.MIA, DEAL_SUBMISSION_TYPE.MIN, DEAL_SUBMISSION_TYPE.AIN])('should return false if the deal type is %s', (type) => {
+      const result = dealCancellationEnabled(type);
+
+      expect(result).toEqual(false);
+    });
+  });
+});
```
ttbarnes, line 12: Bit confused about the 2x tests:

```ts
deal-cancellation-enabled.helper.test.ts
deal-cancellation-enabled.helper.ff-test.ts
```

Should we combine them into one?

trade-finance-manager-ui/templates/case/deal/deal.njk
ttbarnes: Thank you both, can we call this `cancel-button` please?

Recently changed all "cancel" buttons (amongst other things) to have the same `data-cy`.

This way they all use the same E2E partial and click event command 😎

trade-finance-manager-ui/server/controllers/case/index.ff-test.js
abhi-markan: NICE!!

trade-finance-manager-ui/server/controllers/helpers/deal-cancellation-enabled.helper.test.ts
abhi-markan: Good question @ttbarnes 
`deal-cancellation-enabled.helper.ff-test.ts` runs on pipeline with FF set to `true` (future state of the system)
`deal-cancellation-enabled.helper.test.ts` runs on pipeline with FF set to `false` or not set all (current state of the system)

due to above they can't be combined, as we are running two different states concurrently.

trade-finance-manager-ui/server/controllers/helpers/deal-cancellation-enabled.helper.ts
```diff
@@ -0,0 +1,6 @@
+import { DEAL_SUBMISSION_TYPE, DealSubmissionType, isTfmDealCancellationFeatureFlagEnabled } from '@ukef/dtfs2-common';
+
+export const dealCancellationEnabled = (submissionType: DealSubmissionType): boolean => {
+  // TODO: DTFS2-7298: also check that the deal hasn't already been cancelled.
+  return isTfmDealCancellationFeatureFlagEnabled() && (submissionType === DEAL_SUBMISSION_TYPE.AIN || submissionType === DEAL_SUBMISSION_TYPE.MIN);
```
abhi-markan, line 5: Minor point for clarity, may be deconstruct `DEAL_SUBMISSION_TYPE` so the above can be 

```typescript
const acceptableSubmissionType = submissionType === AIN || submissionType === MIN;

return isTfmDealCancellationFeatureFlagEnabled() && acceptableSubmissionType;
```

trade-finance-manager-ui/server/controllers/case/index.test.js
```diff
@@ -2,14 +2,19 @@ import { AMENDMENT_STATUS, isTfmFacilityEndDateFeatureFlagEnabled } from '@ukef/
...
-import { getTask } from '../helpers';
+import { dealCancellationEnabled, getTask } from '../helpers';
 import mapAssignToSelectOptions from '../../helpers/map-assign-to-select-options';
 
 jest.mock('@ukef/dtfs2-common', () => ({
   ...jest.requireActual('@ukef/dtfs2-common'),
   isTfmFacilityEndDateFeatureFlagEnabled: jest.fn(),
 }));
 
+jest.mock('../helpers', () => ({
```
abhi-markan, line 13: Is this `env` variable mock, not sure if this is required anymore?

trade-finance-manager-ui/server/controllers/case/index.test.js
```diff
@@ -60,16 +68,16 @@ describe('controllers - case', () => {
...
           });
-      });
 
-      it('should render deal template with data', async () => {
-        const req = {
+        req = {
           params: {
             _id: mockDeal._id,
           },
           session,
```
abhi-markan, line 76: Do we need a user session to ensure only `PIM` user can action this command. Thanks

abhi-markan (changes requested):
Nice work!

trade-finance-manager-ui/server/controllers/helpers/deal-cancellation-enabled.helper.test.ts
ttbarnes: Makes sense, thanks @abhi-markan however I think it could be made clearer by renaming both tests to the following:

```
deal-cancellation-enabled.helper.ff-enabeld.test.ts
deal-cancellation-enabled.helper.ff-disabled.test.ts
```

Or maybe

```
deal-cancellation-enabled.helper/ff-enabled.test.ts
deal-cancellation-enabled.helper/ff-disabled.test.ts
```

Perhaps this should be in a different PR so we can progress this PR and update all FF unit tests to have the same approach

trade-finance-manager-ui/server/controllers/case/index.test.js
BethThomas141: Good idea, have changed

trade-finance-manager-ui/server/controllers/helpers/deal-cancellation-enabled.helper.ff-test.ts
```diff
@@ -0,0 +1,24 @@
+import { DEAL_SUBMISSION_TYPE } from '@ukef/dtfs2-common';
+import { dealCancellationEnabled } from './deal-cancellation-enabled.helper';
+
+describe('dealCancellationEnabled', () => {
+  describe('when `FF_TFM_FACILITY_END_DATE_ENABLED` is set to false', () => {
```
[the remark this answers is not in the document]
BethThomas141: Good spot, have changed

trade-finance-manager-ui/server/controllers/helpers/deal-cancellation-enabled.helper.test.ts
BethThomas141: Agree with leaving this to a separate PR as that would require changing some of the test set-up I think

trade-finance-manager-ui/server/controllers/helpers/deal-cancellation-enabled.helper.ts
BethThomas141: Good idea, I've changed it to this, and I've also added a check here that the user is a PIM user

trade-finance-manager-ui/server/controllers/case/index.test.js
BethThomas141: This is mocking the return result of `dealCancellationEnabled` - given this is a unit test of the `index` file which doesn't handle the `env` values, I'm just mocking the return result of the function it calls, and the different feature flag values are instead unit tested within the `dealCancellationEnabled` test files.

trade-finance-manager-ui/server/controllers/helpers/deal-cancellation-enabled.helper.test.ts
ttbarnes: Do we have a ticket for this or could this be done as part of @MarRobSoftwire 's feature flag test PR?

BethThomas141: Can I leave this with you @abhi-markan to decide if you want to implement this change in another ticket or not? I think I'm happy with the existing naming as long as we have some documentation about it somewhere so it's clear.

trade-finance-manager-ui/server/controllers/case/index.js
```diff
@@ -49,6 +52,7 @@ const getCaseDeal = async (req, res) => {
     amendments,
     amendmentsInProgress,
     hasAmendmentInProgress,
+    showDealCancelButton: dealCancellationEnabled(deal.dealSnapshot.submissionType, user),
```
Zainzzkk, line 55: can submissionType be dereferenced from deal please

trade-finance-manager-ui/server/controllers/helpers/deal-cancellation-enabled.helper.ff-test.ts
```diff
@@ -0,0 +1,38 @@
...
+
+      it('should return true if the deal type is AIN', () => {
+        const result = dealCancellationEnabled(DEAL_SUBMISSION_TYPE.AIN, pimUser);
+
+        expect(result).toEqual(true);
+      });
+    });
+
+    describe('when user is a non PIM user', () => {
+      it.each([DEAL_SUBMISSION_TYPE.MIA, DEAL_SUBMISSION_TYPE.MIN, DEAL_SUBMISSION_TYPE.AIN])('should return false if the deal type is %s', (type) => {
```
Zainzzkk, line 31: if these can be dereferenced, will be a look better I think (same with other instances in this file)

trade-finance-manager-ui/server/controllers/helpers/deal-cancellation-enabled.helper.ts
```diff
@@ -0,0 +1,13 @@
...
+import { TfmSessionUser } from '../../types/tfm-session-user';
+
+export const dealCancellationEnabled = (submissionType: DealSubmissionType, user: TfmSessionUser): boolean => {
+  // TODO: DTFS2-7298: also check that the deal hasn't already been cancelled.
+
+  const userAllowedToCancelDeal = userIsInTeam(user, [TEAM_IDS.PIM]);
+  const dealCancellationFeatureFlagEnabled = isTfmDealCancellationFeatureFlagEnabled();
+  const acceptableSubmissionType = submissionType === DEAL_SUBMISSION_TYPE.AIN || submissionType === DEAL_SUBMISSION_TYPE.MIN;
+
+  return userAllowedToCancelDeal && dealCancellationFeatureFlagEnabled && acceptableSubmissionType;
```
Zainzzkk, line 12: can this be a const please?

trade-finance-manager-ui/server/controllers/helpers/deal-cancellation-enabled.helper.ff-test.ts
BethThomas141: I agree that destructuring is generally useful as it allows us to reduce the amount of unneeded noise, but in this case I think it might be worth keeping as `MIA` is not descriptive in and of itself to work out what it is referring to

trade-finance-manager-ui/server/controllers/helpers/deal-cancellation-enabled.helper.ts
BethThomas141: I'm not sure there's any need to allocate the memory to reference this variable as it's the return value, and so it's clear it's returning a boolean which is referring to `isDealCancellationEnabled`. I think it might just be an extra line of code which is not particularly needed? I also think it's good to have how the boolean is calculated in easy access right at the end of the function. 

What do you think? Is there a reason why you'd want it to be a const?

trade-finance-manager-ui/server/controllers/helpers/deal-cancellation-enabled.helper.test.ts
AlexBramhill: > Do we have a ticket for this or could this be done as part of @MarRobSoftwire 's feature flag test PR?

As this is adding new features not present in Marks PR, I'd like to avoid it please. Also as this is a change in how we deal with feature flags in their whole, I think this is a bit of scope creep and we should capture this in a properly pointed and planned ticket if possible.

AlexBramhill: Just a note that we should start to write down these feature flag concepts and schedule a meeting to talk about it -- I'll try and get some time cleared with Tom to write this up as as discussed in the office it would be incredibly beneficial to have a technical decision register to refer to

trade-finance-manager-ui/templates/case/deal/deal.njk
BethThomas141: I've renamed to `cancel-button`. Marking as resolved 😃

trade-finance-manager-ui/server/controllers/case/index.js
BethThomas141: I've pulled this out 👍

trade-finance-manager-ui/server/controllers/helpers/deal-cancellation-enabled.helper.test.ts
ttbarnes: Sounds good @AlexBramhill . Do you want to have a session before creating a ticket?

I just think the naming of the FF unit tests are confusing so would like to rename them https://github.com/UK-Export-Finance/dtfs2/pull/3458#discussion_r1756771561

AlexBramhill: Great idea, I think this is a good example of the need for said meeting. Will need to run by Tom, which we can do in Monday standup :)

ttbarnes: I've added DTFS2-7371 to the backlog.

Resolving

trade-finance-manager-ui/server/controllers/helpers/deal-cancellation-enabled.helper.ts
Zainzzkk: it looks better rather than many &&s.  However to keep it consistent, these can all be changed later

trade-finance-manager-ui/server/controllers/helpers/deal-cancellation-enabled.helper.ff-test.ts
```diff
@@ -0,0 +1,38 @@
...
+      });
+
+      it('should return true if the deal type is AIN', () => {
+        const result = dealCancellationEnabled(DEAL_SUBMISSION_TYPE.AIN, pimUser);
+
+        expect(result).toEqual(true);
+      });
+    });
+
+    describe('when user is a non PIM user', () => {
```
abhi-markan, line 30:
```suggestion
    describe('when user is not a PIM user', () => {
```

trade-finance-manager-ui/server/controllers/helpers/deal-cancellation-enabled.helper.ff-test.ts
```diff
@@ -0,0 +1,38 @@
...
+import { dealCancellationEnabled } from './deal-cancellation-enabled.helper';
+import { TfmSessionUser } from '../../types/tfm-session-user';
+
+const pimUser = { teams: [TEAM_IDS.PIM] } as TfmSessionUser;
+const nonPimUser = { teams: [TEAM_IDS.UNDERWRITERS] } as TfmSessionUser;
+
+describe('dealCancellationEnabled', () => {
+  describe('when `FF_TFM_FACILITY_END_DATE_ENABLED` is set to true', () => {
+    describe('when user is a PIM user', () => {
+      it('should return false if the deal type is MIA', () => {
```
abhi-markan, line 11: Nice assertions could you please add one for an edge case:

* `dealCancellationEnabled({}, {});`
* `dealCancellationEnabled(undefined, undefined);`

trade-finance-manager-ui/server/controllers/helpers/deal-cancellation-enabled.helper.test.ts
```diff
@@ -0,0 +1,15 @@
+import { DEAL_SUBMISSION_TYPE, TEAM_IDS } from '@ukef/dtfs2-common';
+import { dealCancellationEnabled } from './deal-cancellation-enabled.helper';
+import { TfmSessionUser } from '../../types/tfm-session-user';
+
+const pimUser = { teams: [TEAM_IDS.PIM] } as TfmSessionUser;
+
+describe('dealCancellationEnabled', () => {
+  describe('when `FF_TFM_FACILITY_END_DATE_ENABLED` is set to false', () => {
```
abhi-markan, line 8: Even though the flag is `false` could you please add above assertions where you assert whether the user session is `PIM` or not please. :-)

abhi-markan (changes requested):
Nice work, few minor repeating comments. Thanks

trade-finance-manager-ui/server/controllers/case/index.test.js
BethThomas141: Resolving this as implemented👍

trade-finance-manager-ui/server/controllers/helpers/deal-cancellation-enabled.helper.ff-test.ts
BethThomas141: As we're working in Typescript, this won't work unless we tell the function that we may genuinely expect an undefined or empty object as parameters. If this is genuinely a concern we can change the function to e.g. take `submissionType?: DealSubmissionType | null, user?: TfmSessionUser | null)`: boolean => {`
 but it would be nicer if we don't have to as that would make the function a bit messy. I think elsewhere we generally assume that if the deal exists then the deal submission type exists and so on. What do you think?

trade-finance-manager-ui/server/controllers/helpers/deal-cancellation-enabled.helper.ff-test.ts
BethThomas141: Changed 👍

trade-finance-manager-ui/server/controllers/case/index.js
```diff
@@ -49,6 +54,7 @@ const getCaseDeal = async (req, res) => {
     amendments,
     amendmentsInProgress,
     hasAmendmentInProgress,
+    showDealCancelButton: dealCancellationEnabled(dealSubmissionType, user),
```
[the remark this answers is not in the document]
BethThomas141: Changed 👍

trade-finance-manager-ui/server/controllers/case/index.js
```diff
@@ -39,6 +42,8 @@ const getCaseDeal = async (req, res) => {
     deal.tfm.stage = DEAL.DEAL_STAGE.AMENDMENT_IN_PROGRESS;
   }
 
+  const dealSubmissionType = deal.dealSnapshot.submissionType;
```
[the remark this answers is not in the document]
BethThomas141: Changed 👍

trade-finance-manager-ui/server/controllers/helpers/deal-cancellation-enabled.helper.ts
```diff
@@ -0,0 +1,13 @@
+import { DEAL_SUBMISSION_TYPE, DealSubmissionType, isTfmDealCancellationFeatureFlagEnabled, TEAM_IDS } from '@ukef/dtfs2-common';
+import { userIsInTeam } from '../../helpers/user';
+import { TfmSessionUser } from '../../types/tfm-session-user';
+
+export const dealCancellationEnabled = (submissionType: DealSubmissionType, user: TfmSessionUser): boolean => {
```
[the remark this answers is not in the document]
BethThomas141: Added 👍

trade-finance-manager-ui/server/controllers/helpers/deal-cancellation-enabled.helper.test.ts
BethThomas141: Added 👍

trade-finance-manager-ui/server/controllers/helpers/deal-cancellation-enabled.helper.ff-test.ts
abhi-markan: Ah apologies I thought this was in JS.

abhi-markan (approved):
LGTM 👍🏼

trade-finance-manager-ui/server/controllers/helpers/deal-cancellation-enabled.helper.ff-test.ts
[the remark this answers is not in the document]
BethThomas141: Resolving as this is fixed

trade-finance-manager-ui/server/controllers/helpers/deal-cancellation-enabled.helper.ts
BethThomas141: Resolving as this has been addressed

trade-finance-manager-ui/server/controllers/helpers/deal-cancellation-enabled.helper.ts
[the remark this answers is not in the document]
BethThomas141: Resolving as now added

PR #3463, opened 2024-09-06
feat(FN-2069): hide horizontal scrollbar on report reconciliation table when at full width

sw-mattw:
## Introduction :pencil2:
There is a small design issue with the scrollbar for the report reconciliation table. The table element is wrapped in a [scrollable pane](https://design-patterns.service.justice.gov.uk/components/scrollable-pane/) and the [example for this](https://design-patterns.service.justice.gov.uk/examples/scrollable-pane/) shows no bar when at full width.

The changes in #3450, however, show the scrollbar at full width which isn't preferable from a design perspective.

The column widths total to 100%, but the pixel values for these add up to slightly greater than the panes width. The pane is 1440px at its maximum, and the columns sum to 1440.02px (`(350.76 * 2) + (151.38 * 4) + (132.98)`) which appears to be causing this issue.

Current version in main (resulting from #3450):
Screenshot 2024-09-06 102401

A simple solution is to assume this will be an off by one error and reduce the tables width by 1px. Alternatively we could reduce the width further (for example to 99% which equates to 14.4px) if we want more safety.

Happy to discuss a solution further with anyone.

Screenshot 2024-09-06 094341

Table width now reduced by 1px to 1439px, scrollbar no longer showing when at full width:
Screenshot 2024-09-06 100059

N/A


trade-finance-manager-ui/styles/table.scss
```diff
@@ -160,6 +160,7 @@ table.premium-payments-table {
 
 table.utilisation-report-reconciliation-table {
   table-layout: fixed;
+  width: calc(100% - 1px);
```
RLavender98, line 163: Please could we add a comment in the code here with a brief explanation as to why?

ttbarnes, line 163: Is it possible to reduce one of the columns by 1 pixel instead of creating a "negative" table width?

sw-mattw: Apologies, added a block comment to provide reasoning for this

sw-mattw: I'm not sure this creates the negative table width. This just sets the table width to slightly less than 100% of its outer container and as the columns themselves have percentage-based widths they'll distribute their widths according to these within the table's slightly reduced width.

I also feel like reducing the width of the nth column would also be harder to understand vs applying this on the table element itself

ttbarnes: Sorry I meant like "less than 100% width" rather than "negative" table width, my fault for poor description 😄 

I know this is all minor, but feels strange to _not_ have the table width as 100% - it sounds like the root cause is one of the table's child elements/columns are making the overall table width more than 100%? If you temporary delete one of the columns and remove the new `calc(100% - 1px);` style, is the issue still present?

Am concerned that this issue could arise in the future with the "-1px" rule, if something is pushing the table out to be more than 1px.

trade-finance-manager-ui/styles/table.scss
```diff
@@ -160,6 +160,13 @@ table.premium-payments-table {
 
 table.utilisation-report-reconciliation-table {
   table-layout: fixed;
+  /* 
+   * Column widths are percentage-based, totalling 100%, but sum to 1440.02px,
+   * slightly exceeding the 1440px maximum scrollable pane width. To fix,
+   * reduce table width by 1px. The column widths will adapt automatically as
+   * they are percentage-based.
```
RLavender98, line 167: I'd suggest we add slightly more detail to this, perhaps along the lines of:
```suggestion
   * Column widths are percentage-based, totalling 100%, but when translated into pixels, 
   * they sum to 1440.02px, slightly exceeding the 1440px maximum scrollable pane width which wraps the table. 
   * This causes a scroll bar to appear when it shouldn't.
   * To fix, we reduce table width by 1px. The column widths will adapt automatically as
   * they are percentage-based.
```

trade-finance-manager-ui/styles/table.scss
sw-mattw: In the situation you describe, the issue of the scrollbar being shown is unfortunately still present (see screenshot at the bottom of this comment).

The issue is due to the browsers rounding when converting the percentage-based widths on each of the columns to their pixel values. The widths are converted to pixels with two decimal places meaning small rounding errors are present. Such errors should always be less than 1px in total (0.02px error observed in this case) and if not for some strange reason the scrollbar will be shown anyway (like in the screenshot below) so the page won't break.


Screenshot 2024-09-09 110137

trade-finance-manager-ui/styles/table.scss
sw-mattw: Cool, updated thanks

trade-finance-manager-ui/styles/table.scss
ttbarnes: Thanks @sw-mattw I think in the future we might need to refactor the table styles and/or potentially (with design) input maybe shortern or remove a column or two to help avoid such issues.

Happy for this to be merged in the meantime since it fixes an immediate issue (FN-2069) 👍

ttbarnes (approved):

RLavender98 (approved):

PR #3464, opened 2024-09-06
feat(FN-2310): add custom behaviour for edit payment redirect

JoshBinns2000:
## Introduction :pencil2:
We want to redirect to different reconciliation for report tabs from the edit payment page depending on where we access it from. There are four main ways to redirect from the "edit payment" journey which are modified in this PR:

1. Clicking the back link
2. Clicking the cancel button
3. Saving the changes to the payment
4. Navigating to delete payment

Cases 1 and 2 have the same behaviour, but cases 3 and 4 can be a little more complicated depending on whether or not there are any errors with the form validation and which actions the user takes.

- Adds new parameters to the edit payment related `URL`s to allow for smart redirects
- Adds a helper to generate the reconciliation for report href link
- Updates e2e tests to cover the required behaviour
- Adds a new `RECONCILIATION_FOR_REPORT_TABS` and `ReconciliationForReportTab` constant and type

- Makes an update to the `pageRenderer` and `componentRenderer` so that the generic can be included at the call of these functions, allow for better typing when the `render` method is called, ie.

```typescript
import { pageRenderer } from '<path>/pageRenderer';
import { EditPaymentViewModel } from '<path>/server/types/view-models';

const render = pageRenderer<EditPaymentViewModel>(page);

render({
  // the type of this object must now satisfy the `EditPaymentViewModel` type
});
```


e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/remove-fees-from-payment.spec.js
```diff
@@ -161,12 +161,12 @@ context(`${PDC_TEAMS.PDC_RECONCILE} users can remove fees from payments`, () =>
...
     pages.utilisationReportPage.premiumPaymentsTab.clickPaymentLink(paymentId);
 
-    cy.url().should('eq', relative(`/utilisation-reports/${reportId}/edit-payment/${paymentId}`));
+    cy.url().should('eq', relative(`/utilisation-reports/${reportId}/edit-payment/${paymentId}?redirectTab=premium-payments`));
 
     getFeeRecordCheckbox(1).check();
     pages.utilisationReportEditPaymentPage.clickRemoveSelectedPaymentsButton();
 
-    cy.url().should('eq', relative(`/utilisation-reports/${reportId}/edit-payment/${paymentId}`));
+    cy.url().should('eq', relative(`/utilisation-reports/${reportId}/edit-payment/${paymentId}?redirectTab=premium-payments`));
```
JoshBinns2000, line 169: @sw-mattw this assertion is failing because I haven't had time to add the logic to retain the redirect tab in the "remove selected fees" journey - once that is added, this test should pass

e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/edit-payment.spec.js
```diff
@@ -77,24 +77,54 @@ context(`${PDC_TEAMS.PDC_RECONCILE} users can edit payments`, () => {
...
+  it('should return to the payment details page when the back link is clicked after accessing the edit payment page from the payment details table', () => {
+    cy.visit(`/utilisation-reports/${reportId}#payment-details`);
+
+    pages.utilisationReportPage.paymentDetailsTab.clickPaymentLink(paymentId);
+
+    cy.url().should('eq', relative(`/utilisation-reports/${reportId}/edit-payment/${paymentId}?redirectTab=payment-details`));
+
+    pages.utilisationReportEditPaymentPage.clickBackLink();
+
+    cy.url().should('eq', relative(`/utilisation-reports/${reportId}#payment-details`));
```
ttbarnes, line 113: Rather than having a long single "it" block, should we split up like this?

```js
describe('when ...', () => {
  it('should ...', () => {
    
  });
});

If the current approach (single its) is used throughout then happy to keep as is so that it's aligned 👍 
```

e2e-tests/tfm/cypress/e2e/pages/utilisation-reports/utilisationReportEditPaymentPage.js
```diff
@@ -3,6 +3,7 @@ const utilisationReportEditPaymentPage = {
   clickSaveChangesButton: () => cy.get('input[data-cy="save-changes-button"]').click(),
   clickDeletePaymentButton: () => cy.get('a[data-cy="delete-payment-button"]').click(),
   clickRemoveSelectedPaymentsButton: () => cy.get('input[data-cy="remove-selected-fees-button"]').click(),
+  clickBackLink: () => cy.get('[data-cy="back-link"]').click(),
```
ttbarnes, line 6: Can we move this into a different structure please - there is a PR open to have back link and other common buttons/links non-specific to the page:

https://github.com/UK-Export-Finance/dtfs2/pull/3453/files#diff-209e86a9ed4a341306fb6f673efa64efbf30abf821716d084696d7f69798084fR1

Happy to pair on this if you'd like, will only take 5mins 😎

trade-finance-manager-ui/server/controllers/utilisation-reports/confirm-delete-payment/index.ts
```diff
@@ -28,16 +30,23 @@ const getPaymentSummaryListRows = (paymentResponse: GetPaymentDetailsWithoutFeeR
   ];
 };
 
-export const getConfirmDeletePayment = async (req: Request, res: Response) => {
+type GetConfirmDeletePaymentRequest = CustomExpressRequest<{
+  query: {
+    redirectTab?: ReconciliationForReportTab;
```
ttbarnes, line 35: Is there only 1x tab in the page?

If not, i'm wondering if this should have a more specific name e.g `redirectXTab` - just thinking, if a new tab redirect is added we'd have to rename this

trade-finance-manager-ui/server/controllers/utilisation-reports/helpers/edit-payment-helper.test.ts
```diff
@@ -367,6 +379,31 @@ describe('edit-payment-helper', () => {
...
+    });
+
+    it('sets the back link to the reconciliation for report href using the supplied redirectTab', () => {
+      // Arrange
+      const redirectTab = RECONCILIATION_FOR_REPORT_TABS.KEYING_SHEET;
+
+      jest.mocked(getReconciliationForReportHref).mockReturnValue('/utilisation-reports/123#keying-sheet');
+
+      // Act
+      const viewModel = getEditPaymentViewModel(aPaymentDetailsWithFeeRecordsResponseBody(), reportId, paymentId, isCheckboxChecked, undefined, redirectTab);
```
ttbarnes, line 401: Not a blocker as this is an established pattern with the view model functions - these could have object structured params - then avoids having to pass in `undefined` params

Can improve these in the future

trade-finance-manager-ui/server/controllers/utilisation-reports/helpers/edit-payment-helper.ts
```diff
@@ -67,6 +69,7 @@ export const getEditPaymentViewModel = (
   paymentId: string,
   isCheckboxChecked: (checkboxId: string) => boolean,
   errors: EditPaymentErrorsViewModel = EMPTY_PAYMENT_ERRORS_VIEW_MODEL,
+  redirectTab: ReconciliationForReportTab | undefined = undefined,
```
ttbarnes, line 72: Do we need to do `undefined = undefined`, could it just be the following?:

```suggestion
  redirectTab?: ReconciliationForReportTab
```

trade-finance-manager-ui/test-helpers/fee-record-payment-group-view-model-item.ts
```diff
@@ -0,0 +1,28 @@
...
+    dataSortValue: 0,
+  },
+  paymentsReceived: [],
+  totalPaymentsReceived: {
+    formattedCurrencyAndAmount: undefined,
+    dataSortValue: 0,
+  },
+  status: FEE_RECORD_STATUS.TO_DO,
+  displayStatus: 'TO DO',
+  checkboxId: 'feeRecordIds-1-reportedPaymentsCurrency-GBP-status-TO_DO',
```
ttbarnes, line 25: Should/could line 24 and 25 use a constant like in line 23?

ttbarnes (changes requested):
Nice work, few minor comments

e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/delete-payment.spec.js
```diff
@@ -64,16 +64,16 @@ context(`${PDC_TEAMS.PDC_RECONCILE} users can delete payments`, () => {
 
     pages.utilisationReportPage.premiumPaymentsTab.clickPaymentLink(payment.id);
 
-    cy.url().should('eq', relative(`/utilisation-reports/${report.id}/edit-payment/${payment.id}`));
+    cy.url().should('eq', relative(`/utilisation-reports/${report.id}/edit-payment/${payment.id}?redirectTab=premium-payments`));
```
Zainzzkk, line 67: Could a const be made for `/utilisation-reports/${report.id}/edit-payment/${payment.id}`  is repeated

e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/edit-payment.spec.js
```diff
@@ -77,24 +77,54 @@ context(`${PDC_TEAMS.PDC_RECONCILE} users can edit payments`, () => {
     cy.visit(`/utilisation-reports/${reportId}`);
   });
 
-  it('should allow the user to navigate to the edit payment page', () => {
+  it('should allow the user to navigate to the edit payment page from the premium payments table', () => {
     pages.utilisationReportPage.premiumPaymentsTab.clickPaymentLink(paymentId);
 
-    cy.url().should('eq', relative(`/utilisation-reports/${reportId}/edit-payment/${paymentId}`));
+    cy.url().should('eq', relative(`/utilisation-reports/${reportId}/edit-payment/${paymentId}?redirectTab=premium-payments`));
```
Zainzzkk, line 83: same as above regard const for first part of url

e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/remove-fees-from-payment.spec.js
```diff
@@ -82,7 +82,7 @@ context(`${PDC_TEAMS.PDC_RECONCILE} users can remove fees from payments`, () =>
     pages.landingPage.visit();
     cy.login(USERS.PDC_RECONCILE);
 
-    cy.visit(`/utilisation-reports/${reportId}/edit-payment/${paymentId}`);
+    cy.visit(`/utilisation-reports/${reportId}/edit-payment/${paymentId}?redirectTab=premium-payments`);
```
Zainzzkk, line 85: same as above for url

trade-finance-manager-ui/test-helpers/fee-record-payment-group-view-model-item.ts
Zainzzkk: also should TODO be in the code or should it be a comment instead with the ticket number?

ttbarnes: @Zainzzkk `TO_DO` is a valid status. I believe there is a ticket to use more status constants. In the meantime would be great to _not_ introduce more hard coded statuses

e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/edit-payment.spec.js
```diff
@@ -77,176 +79,250 @@ context(`${PDC_TEAMS.PDC_RECONCILE} users can edit payments`, () => {
...
+      cy.getInputByLabelText('Month').should('have.value', 'nonsense');
+      cy.getInputByLabelText('Year').should('have.value', 'nonsense');
+      cy.getInputByLabelText('Payment reference (optional)').should('have.value', 'This is valid');
+    });
   });
 
-  it('should return to the premium payments table after the user clicks the save changes button', () => {
-    pages.utilisationReportPage.premiumPaymentsTab.clickPaymentLink(paymentId);
+  describe('when the user clicks the save changes button', () => {
+    it('should return to the premium payments table', () => {
```
RLavender98, line 192: This test duplicates one below with a better name so I think this should be removed

trade-finance-manager-ui/test-helpers/fee-record-payment-group-view-model-item.ts
sw-mattw: Yep, apologies, I've replaced the checkboxId string with constants for the currency and status.

`displayStatus` is a bit more complex - the `'TO DO'` string isn't exported as its made available through a helper function to get the display text from the actual status, therefore I think the current approach of using the string literal is preferable (this is also used across other files) so I've left this as is.

trade-finance-manager-ui/server/controllers/utilisation-reports/helpers/edit-payment-helper.ts
sw-mattw: Sorry, missed this from Josh's initial work - resolved this now thanks

e2e-tests/tfm/cypress/e2e/partials/index.js
```diff
@@ -3,11 +3,13 @@ const caseSummary = require('./caseSummary');
...
 const primaryNavigation = require('./primaryNavigation');
+const clickEvents = require('./clickEvents');
 
 module.exports = {
   caseSubNavigation,
   caseSummary,
   header,
   underwritingSubNav,
   primaryNavigation,
+  clickEvents,
```
ttbarnes, line 14: Can we change this to simply be this please:

```js
backLink: () => cy.get('[data-cy="back-link"]'),
```

So that TFM E2E will be aligned with the same structure as GEF/Portal (and 2x other incoming PRs)

Then we can remove `e2e-tests/tfm/cypress/e2e/partials/clickEvents.js`

e2e-tests/tfm/cypress/support/ui/click-events/click-back-link.js
```diff
@@ -0,0 +1,11 @@
+import { clickEvents } from '../../../e2e/partials';
+
+/**
+ * clickBackLink
+ * Click the back link.
+ */
+const clickBackLink = () => {
+  clickEvents.backLink().click();
```
ttbarnes, line 8: After the above comment this can just import `backLink` and then be `backLink().click();`

trade-finance-manager-ui/server/controllers/utilisation-reports/helpers/edit-payment-helper.ts
```diff
@@ -92,6 +97,7 @@ export const getEditPaymentViewModelWithFormValues = (
   isCheckboxChecked: (checkboxId: EditPaymentsTableCheckboxId) => boolean,
   formValues: EditPaymentFormValues,
   errors: EditPaymentErrorsViewModel = EMPTY_PAYMENT_ERRORS_VIEW_MODEL,
+  redirectTab: ReconciliationForReportTab | undefined = undefined,
```
ttbarnes, line 100: Do we need `| undefined = undefined`? Could it just be this?:

```ts
redirectTab?: ReconciliationForReportTab
```

ttbarnes (commented):
Just a few minor comments otherwise LGTM 👍

trade-finance-manager-ui/server/controllers/utilisation-reports/confirm-delete-payment/index.ts
sw-mattw: There are multiple tabs on the reconciliation for report page and this query param specifies which of these we want to be shown when we perform a redirect. This page is the only template within the utilisation reports group that uses tabs at present. We also need to be aware when naming that this is passed around as a query parameter.

We could rename it something like `reconciliationForReportRedirectTab` if you'd prefer something like that? Or leave as is? Open to either, thanks

e2e-tests/tfm/cypress/e2e/pages/utilisation-reports/utilisationReportEditPaymentPage.js
sw-mattw: Changed to adopt this, thanks for the link

e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/edit-payment.spec.js
sw-mattw: Refactored to use the describe blocks thanks

trade-finance-manager-ui/server/controllers/utilisation-reports/confirm-delete-payment/index.ts
ttbarnes: Thanks for explaining @sw-mattw, I think what we have already makes sense 👍

trade-finance-manager-ui/server/controllers/utilisation-reports/helpers/edit-payment-helper.ts
sw-mattw: Updated to match that pattern thanks, I've checked over the other touched files and there don't seem to be any other occurrences

trade-finance-manager-ui/server/controllers/utilisation-reports/helpers/edit-payment-helper.test.ts
```diff
@@ -3,11 +3,23 @@ import { Currency, CurrencyAndAmount } from '@ukef/dtfs2-common';
...
 import { PaymentErrorsViewModel, SortedAndFormattedCurrencyAndAmount } from '../../../types/view-models';
 import { EditPaymentFormValues } from '../../../types/edit-payment-form-values';
 import { EMPTY_PAYMENT_ERRORS_VIEW_MODEL } from './payment-form-helpers';
+import { getReconciliationForReportHref } from './get-reconciliation-for-report-href';
+
+jest.mock('./get-reconciliation-for-report-href');
 
 describe('edit-payment-helper', () => {
+  beforeEach(() => {
+    jest.mocked(getReconciliationForReportHref).mockReturnValue('/utilisation-reports/1');
```
ttbarnes, line 16: Sorry for not spotting before, do we have to mock this? As it's just returning a URL, I think should be consumed in the unit test(s) rather than mocking it? e.g

```ts
const expected = getReconciliationForReportHref(reportId, redirectTab);

expect(viewModel.backLinkHref).toEqual(expected);
```

sw-mattw: Updated this as it is returning something fairly simple as you say, thanks

RLavender98 (approved):

sw-mattw:
Some jobs seem to be hanging, retriggered all jobs by merging main and pushing:

Screenshot 2024-09-11 121923

ttbarnes:
> Some jobs seem to be hanging, retriggered all jobs by merging main and pushing:
> 

I've been getting the same issue randomly in other PRs, seems to be since [this PR](https://github.com/UK-Export-Finance/dtfs2/pull/3390) was merged

PR #3465, opened 2024-09-09
feat(FN-2309): introduce a DRY aUtilisationReport method

RLavender98:
## Introduction :pencil2:
It was noted on another PR that there were 14 duplicate definitions for a test helper method.

Move the `aUtilisationReport` test helper to common place and remove all duplicates.

### DIFF
--- a/dtfs-central-api/test-helpers/test-data/utilisation-report.ts
+++ b/dtfs-central-api/test-helpers/test-data/utilisation-report.ts
@@ -1,0 +1,12 @@
+import { UTILISATION_REPORT_RECONCILIATION_STATUS, UtilisationReportEntity, UtilisationReportEntityMockBuilder } from '@ukef/dtfs2-common';
+
+/**
+ * Test helper which returns a utilisation report entity.
+ *
+ * This is to be used when the test just requires a utilisation report but does not care what the values of any fields are.
+ * To build a report for a test where it does matter what the fields are then you should use the entity mock builder directly.
+ *
+ * @returns a utilisation report entity
+ */
+export const aUtilisationReport = (): UtilisationReportEntity =>
+  UtilisationReportEntityMockBuilder.forStatus(UTILISATION_REPORT_RECONCILIATION_STATUS.RECONCILIATION_IN_PROGRESS).build();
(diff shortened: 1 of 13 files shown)


ttbarnes (approved):

dtfs-central-api/test-helpers/test-data/utilisation-report.ts
```diff
@@ -0,0 +1,12 @@
...
+/**
+ * Test helper which returns a utilisation report entity.
+ *
+ * This is to be used when the test just requires a utilisation report but does not care what the values of any fields are.
+ * To build a report for a test where it does matter what the fields are then you should use the entity mock builder directly.
+ *
+ * @returns a utilisation report entity
+ */
+export const aUtilisationReport = (): UtilisationReportEntity =>
+  UtilisationReportEntityMockBuilder.forStatus(UTILISATION_REPORT_RECONCILIATION_STATUS.RECONCILIATION_IN_PROGRESS).build();
```
toddym42, line 12: I feel like this is just creating a test helper for a test helper. Test mock builders are made exactly for this purpose (i.e., getting an instance of a valid entity when we don't care what the values are), so we could have just built this into the the mock builder without a further separate level of test helper abstraction.

Most mock builders will give a valid default instance when no values are explicitly set. For this one we chose specifically not to allow this by making the constructor private, forcing us to instead use the static factory method `forStatus`. If we've decided now that we don't want to be forced to do that we can just change the constructor from:
```
private constructor(report: UtilisationReportEntity) {
  this.report = report;
}
```
to
```
public constructor(report?: UtilisationReportEntity) {
  this.report = report ?? UtilisationReportEntityMockBuilder.forStatus('RECONCILIATION_IN_PROGRESS').report;
}
```
then call it with
```
new UtilisationReportEntityMockBuilder().build()
```

ttbarnes: Sounds good @toddym42 RE having this in the mock builder. I assume this will simplify some tests? 

Want to create a ticket to do this?

toddym42: I'll leave it to you and the dtfs team to decide how to want to go forward with this (if at all). Was just sticking my nose in after seeing something I'd previously worked on 😅

PR #3469, opened 2024-09-09
feat(DTFS2-7294): add deal cancellation update backend

BethThomas141:
## Introduction :pencil2:
This PR implements the planned backend infrastructure for the PUT endpoint for updating deal cancellation: (proposed plan by @MarRobSoftwire below:)

image

The cancellation object lives on the `tfm` object and has the structure: 
```
cancellation: {
  reason: "lorem ipsum",
  bankRequestDate: 1234567890, // Unix time
  effectiveFrom: 1234567890, // Unix time
}
```


trade-finance-manager-ui/server/api-ssrf.test.js
```diff
@@ -3,6 +3,9 @@ const MockAdapter = require('axios-mock-adapter');
 const api = require('./api');
 
 const mockAxios = new MockAdapter(axios);
+const localIp = '127.0.0.1';
+const urlTraversal = '../../../etc/stealpassword';
+const validId = '5ce819935e539c343f141ece';
```
BethThomas141, line 8: I've added a test to the bottom of this file, but also done a bit of a refactor to reduce some DRY content (moved these variables out up here, and commonised some more of the amendments tests)

dtfs-central-api/src/repositories/tfm-deals-repo/tfm-deals-cancellation-repo.test.ts
```diff
@@ -0,0 +1,83 @@
...
+  const getCollectionMock = jest.fn();
+
+  afterEach(() => {
+    jest.resetAllMocks();
+  });
+
+  describe('findDealCancellationByDealId', () => {
+    beforeEach(() => {
+      findOneMock.mockResolvedValue(mockDealObject);
+      getCollectionMock.mockResolvedValue({});
```
MarRobSoftwire, line 24: What does this do?

dtfs-central-api/src/repositories/tfm-deals-repo/tfm-deals-cancellation-repo.test.ts
```diff
@@ -0,0 +1,83 @@
...
+describe('tfm-deals-cancellation-repo', () => {
+  const findOneMock = jest.fn();
+  const updateOneMock = jest.fn();
+  const getCollectionMock = jest.fn();
+
+  afterEach(() => {
+    jest.resetAllMocks();
+  });
+
+  describe('findDealCancellationByDealId', () => {
```
MarRobSoftwire, line 21: Could you test the correct error is thrown if its not found?

dtfs-central-api/src/repositories/tfm-deals-repo/tfm-deal-cancellation.repo.ts
```diff
@@ -0,0 +1,48 @@
...
+    return matchingDeal.tfm.cancellation;
+  }
+
+  /**
+   * Updates the deal tfm object with the supplied cancellation
+   * @param dealId - The deal id
+   * @param update - The deal cancellation update to apply
+   * @returns The update result
+   */
+  public static async updateOneDealWithCancellation(dealId: string | ObjectId, update: UpdateFilter<TfmDealCancellation>): Promise<UpdateResult> {
```
MarRobSoftwire, line 32: The two names have a slightly different pattern - should this be `updateOneDealCancellation`

dtfs-central-api/src/repositories/tfm-deals-repo/tfm-deals-cancellation-repo.test.ts
```diff
@@ -0,0 +1,83 @@
...
+    it('returns the found deal cancellation', async () => {
+      // Act
+      const result = await TfmDealCancellationRepo.findDealCancellationByDealId(dealId);
+
+      // Assert
+      expect(result).toEqual(mockDealCancellationObject);
+    });
+  });
+
+  describe('updateOneDealWithCancellation', () => {
```
MarRobSoftwire, line 50: Could you add test for error thrown & correct update result returned?

dtfs-central-api/src/v1/controllers/tfm/deal-cancellation/tfm-put-update-deal-cancellation.controller.ts
```diff
@@ -0,0 +1,43 @@
...
+
+    await TfmDealCancellationRepo.updateOneDealWithCancellation(
+      dealId,
+      flatten({
+        'tfm.cancellation': dealCancellationUpdate,
+        auditRecord: generateAuditDatabaseRecordFromAuditDetails(auditDetails),
+      }),
+    );
+
+    const dealCancellation = await TfmDealCancellationRepo.findDealCancellationByDealId(dealId);
```
MarRobSoftwire, line 30: Rather than calling the repo twice, do we want to use `findOneAndUpdate`?

MarRobSoftwire: Or should we just return the update result if we're not using the response body

dtfs-central-api/src/v1/routes/tfm-routes.js
```diff
@@ -530,6 +535,22 @@ tfmRouter
...
+ *         description: Not found
+ */
+tfmRouter
+  .route('/deals/:dealId/cancellation')
+  .put(
+    validation.mongoIdValidation('dealId'),
+    handleExpressValidatorResult,
+    validatePutDealCancellationPayload,
+    tfmPutUpdateDealCancellationController.updateTfmDealCancellation,
+  );
```
MarRobSoftwire, line 552: This is so nice.

Do we want more swagger docs?

libs/common/src/types/mongo-db-models/deal.ts
```diff
@@ -0,0 +1,9 @@
+import { AnyObject } from '../any-object';
+
+/**
+ * Type of the mongo db "deal" collection
+ *
+ * This type is likely incomplete and should be added
+ * to as and when new properties are discovered
+ */
+export type Deal = AnyObject;
```
MarRobSoftwire, line 9: 👨‍🍳

libs/common/src/types/mongo-db-models/index.ts
```diff
@@ -7,3 +7,4 @@ export * from './users';
 export * from './deletion-audit-logs';
 export * from './tfm-facilities';
 export * from './facility';
+export * from './tfm-deals';
```
MarRobSoftwire, line 10: Do we want to export the deals type too, even though its just a stub?

libs/common/src/types/mongo-db-models/tfm-deals.ts
```diff
@@ -0,0 +1,22 @@
+import { ObjectId } from 'mongodb';
+import { AuditDatabaseRecord } from '../audit-database-record';
+import { Deal } from './deal';
+
+export type TfmDealCancellation = {
+  reason: string;
+  bankRequestDate: number;
+  effectiveFrom: number;
+};
```
MarRobSoftwire, line 9: Should this be stored in its own file a level up? We'll use the deal cancellation a lot not as a mongo-db-model

trade-finance-manager-api/src/v1/deals/routes.js
```diff
@@ -76,6 +77,10 @@ dealsAuthRouter
   .get(validation.dealIdValidation, handleExpressValidatorResult, dealController.getDeal)
   .put(validation.dealIdValidation, handleExpressValidatorResult, dealController.updateDeal);
 
+dealsAuthRouter
+  .route('/deals/:dealId/cancellation')
+  .put(validation.dealIdValidation, handleExpressValidatorResult, dealCancellationController.updateDealCancellation);
```
MarRobSoftwire, line 82: Do we want zod schema validation here as well? It'd be a new pattern but we could possibly move some stuff from dtfs-central to libs common (perhaps in a different subtask)

trade-finance-manager-ui/server/api.js
```diff
@@ -1230,6 +1230,35 @@ const addFeesToAnExistingPayment = async (reportId, feeRecordIds, paymentIds, us
   return response.data;
 };
 
+/**
+ * @param {string} dealId - The deal ID
+ * @param {Partial<import('@ukef/dtfs2-common').TfmDealCancellation>} cancellationUpdate - The deal cancellation update object
+ * @param {string} userToken - The user token
+ * @returns {Promise<void>}
```
MarRobSoftwire, line 1237: This doesnt return `void`

trade-finance-manager-ui/server/api.js
```diff
@@ -1230,6 +1230,35 @@ const addFeesToAnExistingPayment = async (reportId, feeRecordIds, paymentIds, us
...
+ */
+const updateDealCancellation = async (dealId, cancellationUpdate, userToken) => {
+  try {
+    const isValidDealId = isValidMongoId(dealId);
+
+    if (!isValidDealId) {
+      console.error('updateDealCancellation: Invalid deal id provided %s', dealId);
+      return { status: 400, data: 'Invalid deal id' };
```
MarRobSoftwire, line 1245: I assume this is an existing pattern but its weird to return `{ status: 400 }` when we've not even made an api request?

MarRobSoftwire: Looking at the payment record functions above they dont try/catch or even validate - would that be a better plan - let the api's deal with this stuff?

trade-finance-manager-ui/server/api-ssrf.test.js
```diff
@@ -229,206 +211,145 @@ describe('API is protected against SSRF attacks', () => {
...
-
-    it('Makes an axios request when the deal id is valid', async () => {
-      const validDealId = '5ce819935e539c343f141ece';
-
-      const response = await api.getAmendmentsByDealId(validDealId, 'mock token');
-
-      expect(response.data).toEqual(mockResponse);
-    });
-  });
+  const getAmendmentDetailsByDealIdFunctionsToTest = [
```
MarRobSoftwire, line 238: Nice work :)

MarRobSoftwire: Can we add every test to this describe.each?

MarRobSoftwire: Happy to leave of course

trade-finance-manager-api/src/v1/controllers/deal-cancellation.controller.test.ts
```diff
@@ -0,0 +1,42 @@
...
+
+const mockDealId = new ObjectId();
+const mockUserId = new ObjectId();
+
+describe('controllers - deal cancellation', () => {
+  beforeEach(() => {
+    jest.resetAllMocks();
+  });
+
+  describe('PUT - updateDealCancellation', () => {
```
MarRobSoftwire, line 22: Could you add tests for error handling & returning the correct thing when successful

MarRobSoftwire (commented):
Looks really good!

Assume you'll add api tests under a future ticket?

dtfs-central-api/src/repositories/tfm-deals-repo/tfm-deal-cancellation.repo.ts
```diff
@@ -0,0 +1,48 @@
...
+    return await mongoDbClient.getCollection('tfm-deals');
+  }
+
+  /**
+   * Finds the cancellation by dealId
+   * @param dealId - The deal id
+   * @returns the found deal cancellation
+   */
+  public static async findDealCancellationByDealId(dealId: string | ObjectId): Promise<Document> {
+    const dealCollection = await this.getCollection();
```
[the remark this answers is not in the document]
BethThomas141: So if I change it away from `Document` I get type errors in the controller file when trying to return the json output with the cancellation object - but then this is possibly an odd pattern I was trying to copy (a lot of the existing controllers do this where they e.g. update something in the database, call it again with a different endpoint where it's returned as a `Document` type, and then send that in json format with the status code - not sure what the best implementation is here
image

libs/common/src/types/mongo-db-models/index.ts
BethThomas141: Ooh I don't follow - it's being exported here, is there something else you think should be exported too?

BethThomas141: Ah I see your comment below about it being in the mongo-db directory

trade-finance-manager-api/src/v1/controllers/deal-cancellation.controller.test.ts
BethThomas141: Yep added!

dtfs-central-api/src/repositories/tfm-deals-repo/tfm-deal-cancellation.repo.ts
```diff
@@ -0,0 +1,48 @@
...
+   * @param dealId - The deal id
+   * @param update - The deal cancellation update to apply
+   * @returns The update result
+   */
+  public static async updateOneDealWithCancellation(dealId: string | ObjectId, update: UpdateFilter<TfmDealCancellation>): Promise<UpdateResult> {
+    const dealCollection = await this.getCollection();
+
+    const matchingDeal = await dealCollection.findOne({ _id: { $eq: new ObjectId(dealId) } });
+
+    if (!matchingDeal) {
```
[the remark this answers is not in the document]
BethThomas141: Great idea, have updated.

dtfs-central-api/src/repositories/tfm-deals-repo/tfm-deal-cancellation.repo.ts
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  /**
+   * Finds the cancellation by dealId
+   * @param dealId - The deal id
+   * @returns the found deal cancellation
+   */
+  public static async findDealCancellationByDealId(dealId: string | ObjectId): Promise<Document> {
+    const dealCollection = await this.getCollection();
+    const matchingDeal = await dealCollection.findOne({ _id: { $eq: new ObjectId(dealId) } });
```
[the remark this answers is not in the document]
BethThomas141: Good idea - I've added that full check to the `update` one below, and for this one have added the check but without the check that it's not 'Cancelled' - as I imagine we may want to `get` the deal cancellation information using this when the deal is cancelled.

dtfs-central-api/src/repositories/tfm-deals-repo/tfm-deals-cancellation-repo.test.ts
BethThomas141: Absolutely nothing 😃 have removed!

dtfs-central-api/src/repositories/tfm-deals-repo/tfm-deals-cancellation-repo.test.ts
BethThomas141: Have added

libs/common/src/types/mongo-db-models/tfm-deals.ts
BethThomas141: Good idea, have moved

trade-finance-manager-ui/server/api.js
BethThomas141: That's a much better idea - I've removed it and let the api handle it. It feels like all the calls in this file should not have the valid id check given they're all checked anyway in the API layer! So a lot of repetition around.

libs/common/src/types/mongo-db-models/index.ts
BethThomas141: Hmm sorry for the rollercoaster of messages, I'm currently using the `TfmDeal` type as a type in `dtfs-central-api/src/repositories/tfm-deals-repo/tfm-deal-cancellation.repo.ts` so using the export in the one place. I think fine to export it for use there?

trade-finance-manager-api/src/v1/deals/routes.js
BethThomas141: I think that would be nice - but might cause this PR to take longer to get through. Happy for this to go as another slightly lower priority subtask?

trade-finance-manager-api/src/v1/controllers/deal-cancellation.controller.test.ts
```diff
@@ -0,0 +1,100 @@
...
+
+      // Act
+      await updateDealCancellation(req, res);
+
+      // Assert
+      expect(res._getStatusCode()).toBe(HttpStatusCode.Ok);
+      expect(res._getData()).toEqual(expect.objectContaining(dealCancellationUpdate));
+    });
+
+    it('should throw an error when there is an API error', async () => {
```
MarRobSoftwire, line 62: should this be `return an error` and similarly below

trade-finance-manager-api/src/v1/deals/routes.js
MarRobSoftwire: Sounds good - feel free to resolve this comment

libs/common/src/types/mongo-db-models/index.ts
```diff
@@ -7,4 +7,4 @@ export * from './users';
 export * from './deletion-audit-logs';
 export * from './tfm-facilities';
 export * from './facility';
-export * from './tfm-deals';
+export * from './tfm-deal';
```
MarRobSoftwire, line 10: It looks like the other files here use the plural (as in the same as the collection name)

trade-finance-manager-api/src/v1/api.js
```diff
@@ -269,6 +269,33 @@ const submitDeal = async (dealType, dealId, auditDetails) => {
   }
 };
 
+/**
+ * @param {object} params
+ * @param {string} params.dealId - deal cancellation to update
+ * @param {Partial<import('@ukef/dtfs2-common').TfmDealCancellation>} params.dealCancellationUpdate - deal cancellation update to make
+ * @param {import('@ukef/dtfs2-common').AuditDetails} params.auditDetails - user making the request
+ * @returns {Promise<Partial<import('@ukef/dtfs2-common').TfmDealCancellation>>} updated deal cancellation on success
```
MarRobSoftwire, line 277: This should be an update result now :)

trade-finance-manager-ui/server/api.js
```diff
@@ -1230,6 +1230,23 @@ const addFeesToAnExistingPayment = async (reportId, feeRecordIds, paymentIds, us
   return response.data;
 };
 
+/**
+ * @param {string} dealId - The deal ID
+ * @param {Partial<import('@ukef/dtfs2-common').TfmDealCancellation>} cancellationUpdate - The deal cancellation update object
+ * @param {string} userToken - The user token
+ * @returns {Partial<import('@ukef/dtfs2-common').TfmDealCancellation>} - The deal cancellation update object
```
MarRobSoftwire, line 1237: Also should be an update result (though thats a mongodb type which we cant import in tfm-ui so maybe want something else - honestly could just be void right?)

BethThomas141: Yep good point, we don't need the returned object so have changed it to not return anything / updated docstrings to `@returns {Promise<void>}`

trade-finance-manager-api/src/v1/api.js
BethThomas141: Good spot, have changed!

libs/common/src/types/mongo-db-models/index.ts
BethThomas141: Changed to the plural👍

libs/common/src/types/mongo-db-models/index.ts
BethThomas141: Gotcha, not sure how I didn't read it as such!! Now exporting `Deal` too

trade-finance-manager-api/src/v1/deals/routes.js
BethThomas141: Added it as a subtask to the ticket👍

trade-finance-manager-api/api-tests/api-ssrf.api-test.js
```diff
@@ -9,6 +11,9 @@ const api = jest.requireActual('../src/v1/api');
 
 const mockAxios = new MockAdapter(axios);
 
+const localIp = '127.0.0.1';
+const urlTraversal = '../../../etc/stealpassword';
+
```
BethThomas141, line 16: Pulled some constants out at the top to make things more DRY in this file :)

trade-finance-manager-api/src/v1/deals/routes.js
ttbarnes, line 82: In a future PR are we going to add middleware here to check that it's a user that can cancel a deal (PIM group/user?)

ttbarnes (changes requested):
Great work @BethThomas141 just some minor comments otherwise LGTM 🚀

trade-finance-manager-api/src/v1/api.js
```diff
@@ -269,6 +269,33 @@ const submitDeal = async (dealType, dealId, auditDetails) => {
   }
 };
 
+/**
+ * @param {object} params
```
abhi-markan, line 273: Missing function summary.

trade-finance-manager-api/src/v1/api.js
```diff
@@ -269,6 +269,33 @@ const submitDeal = async (dealType, dealId, auditDetails) => {
...
+ * @param {string} params.dealId - deal cancellation to update
+ * @param {Partial<import('@ukef/dtfs2-common').TfmDealCancellation>} params.dealCancellationUpdate - deal cancellation update to make
+ * @param {import('@ukef/dtfs2-common').AuditDetails} params.auditDetails - user making the request
+ * @returns {Promise<import('mongodb').UpdateResult>} update result object
+ */
+const updateDealCancellation = async ({ dealId, dealCancellationUpdate, auditDetails }) => {
+  const isValidDealId = isValidMongoId(dealId);
```
abhi-markan, line 280: Can we wrap this in `try-catch` block please.

trade-finance-manager-api/src/v1/controllers/deal-cancellation.controller.ts
```diff
@@ -0,0 +1,40 @@
...
+import api from '../api';
+
+export type UpdateDealCancellationRequest = CustomExpressRequest<{
+  params: {
+    dealId: string;
+  };
+  reqBody: TfmDealCancellation;
+}>;
+
+export const updateDealCancellation = async (req: UpdateDealCancellationRequest, res: Response) => {
```
abhi-markan, line 14: DocString please.

trade-finance-manager-api/src/v1/deals/routes.js
```diff
@@ -76,6 +77,10 @@ dealsAuthRouter
   .get(validation.dealIdValidation, handleExpressValidatorResult, dealController.getDeal)
   .put(validation.dealIdValidation, handleExpressValidatorResult, dealController.updateDeal);
 
+dealsAuthRouter
```
abhi-markan, line 80: Is this restricted to `PIM` only?

trade-finance-manager-ui/server/api.js
```diff
@@ -1230,6 +1230,21 @@ const addFeesToAnExistingPayment = async (reportId, feeRecordIds, paymentIds, us
   return response.data;
 };
 
+/**
+ * @param {string} dealId - The deal ID
```
abhi-markan, line 1234: Summary please.

trade-finance-manager-ui/server/api.js
```diff
@@ -1230,6 +1230,21 @@ const addFeesToAnExistingPayment = async (reportId, feeRecordIds, paymentIds, us
...
+ * @param {Partial<import('@ukef/dtfs2-common').TfmDealCancellation>} cancellationUpdate - The deal cancellation update object
+ * @param {string} userToken - The user token
+ * @returns {Promise<void>}
+ */
+const updateDealCancellation = async (dealId, cancellationUpdate, userToken) => {
+  await axios({
```
abhi-markan, line 1240: try-catch block please.

abhi-markan (commented):
Nice work, few repeating comments.
We should add a restriction where only a PIM user can call this endpoint.

dtfs-central-api/src/v1/controllers/tfm/deal-cancellation/tfm-put-update-deal-cancellation.controller.ts
```diff
@@ -0,0 +1,43 @@
+import { Response } from 'express';
```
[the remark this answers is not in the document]
BethThomas141: A good question, I guess we were thinking that cancelling a deal feels like the same type of things as amending a facility, and `amendments` is it's own folder (not in `facility`). Hence making it a distinct folder for this following the `amendments` pattern.

dtfs-central-api/src/repositories/tfm-deals-repo/tfm-deal-cancellation.repo.ts
```diff
@@ -0,0 +1,53 @@
...
+  private static async getCollection(): Promise<Collection<WithoutId<TfmDeal>>> {
+    return await mongoDbClient.getCollection('tfm-deals');
+  }
+
+  /**
+   * Finds the cancellation by dealId
+   * @param dealId - The deal id
+   * @returns the found deal cancellation
+   */
+  public static async findDealCancellationByDealId(dealId: string | ObjectId): Promise<TfmDealCancellation> {
```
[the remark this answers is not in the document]
BethThomas141: In all the cases I can see in the tfm-facility repos we always use `dealId: string | ObjectId` so I'm a little nervous about committing to one. Do you know what it should be here? 

image

dtfs-central-api/src/v1/controllers/tfm/deal-cancellation/tfm-put-update-deal-cancellation.controller.ts
```diff
@@ -0,0 +1,43 @@
...
+
+type UpdateTfmDealCancellationResponse = Response<ApiErrorResponseBody | UpdateResult>;
+
+export const updateTfmDealCancellation = async (req: UpdateTfmDealCancellationRequest, res: UpdateTfmDealCancellationResponse) => {
+  const { dealCancellationUpdate, auditDetails } = req.body;
+  const { dealId } = req.params;
+
+  try {
+    validateAuditDetailsAndUserType(auditDetails, AUDIT_USER_TYPES.TFM);
+
```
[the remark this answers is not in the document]
BethThomas141: This happens already when calling `updateOneDealCancellation` - it only matches on deals that don't have `tfm.stage` equal to Cancelled

trade-finance-manager-ui/server/api.js
BethThomas141: My expectation is that this will be called by the controller that will wrap this in a `try/catch` block itself (see implementation of the e.g. `addFeesToAnExistingPayment`, `removeFeesFromPayment` `markKeyingDataAsDone` functions in this file. Adding a try/catch block here as well would be unnecessary and reduce DRY of the code

libs/common/src/types/mongo-db-models/tfm-deals.ts
```diff
@@ -0,0 +1,17 @@
...
+ * Type of the mongo db "tfm-deals" collection
+ *
+ * This type is likely incomplete and should be added
+ * to as and when new properties are discovered
+ */
+export type TfmDeal = {
+  _id: ObjectId;
+  dealSnapshot: Deal;
+  tfm: { cancellation: TfmDealCancellation };
+  auditRecord: AuditDatabaseRecord;
```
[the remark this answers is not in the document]
BethThomas141: @MarRobSoftwire can I get your input here? 👀

trade-finance-manager-api/src/v1/api.js
BethThomas141: The only place where this is called and will ever be called is the controller in tfm-api and that wraps this in a `try/catch` block, so it would be unnecessary repetition to also have it here

dtfs-central-api/src/v1/controllers/tfm/deal-cancellation/tfm-put-update-deal-cancellation.controller.ts
```diff
@@ -0,0 +1,43 @@
...
+import { generateAuditDatabaseRecordFromAuditDetails, validateAuditDetailsAndUserType } from '@ukef/dtfs2-common/change-stream';
+import { UpdateResult } from 'mongodb';
+import { TfmDealCancellationRepo } from '../../../../repositories/tfm-deals-repo';
+import { PutDealCancellationPayload } from '../../../routes/middleware/payload-validation';
+
+type UpdateTfmDealCancellationRequest = CustomExpressRequest<{
+  reqBody: PutDealCancellationPayload;
+}>;
+
+type UpdateTfmDealCancellationResponse = Response<ApiErrorResponseBody | UpdateResult>;
```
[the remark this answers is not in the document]
BethThomas141: It's fairly generic, however it's only `PUT` requests that return an `UpdateResult` and this is the first time there's a type that returns `Response<ApiErrorResponseBody | UpdateResult>` so this would be the only use. I'd prefer to keep it as is, and if someone else find they also use the same return type they can pull it out at that point to follow YAGNI principles.

libs/common/src/types/mongo-db-models/tfm-deals.ts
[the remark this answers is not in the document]
MarRobSoftwire: This is a good point, I suppose it is optional (but will exist on any deal made after audit logs is released). It's nice that this will cause a type error if you try to create a deal without `auditRecord`. The code doesn't actually use the auditRecord on any `findOne` methods or similar - its just sent to the data lake.

If we do want to make it optional it would be good to have a comment explaining that its only optional on in flight deals, pre audit log release

```suggestion
  // Audit records may not exist on a deal if it has not been modified after Audit Logs is released
  auditRecord?: AuditDatabaseRecord;
```

dtfs-central-api/api-tests/v1/tfm/deal-cancellation/tfm-deal-cancellation-put.api-test.js
```diff
@@ -0,0 +1,100 @@
...
+      expect(status).toEqual(404);
+    });
+
+    it('should return 400 if invalid dealId', async () => {
+      const invalidDealId = '1234';
+
+      const { status } = await testApi.put({ dealCancellationUpdate, auditDetails: tfmAuditDetails }).to(`/v1/tfm/deals/${invalidDealId}/cancellation`);
+      expect(status).toEqual(400);
+    });
+  });
```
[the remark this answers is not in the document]
BethThomas141: There's a test for this on line 64 in this file

dtfs-central-api/src/repositories/tfm-deals-repo/tfm-deal-cancellation.repo.ts
```diff
@@ -0,0 +1,53 @@
...
+   * @param update - The deal cancellation update to apply
+   * @returns The update result
+   */
+  public static async updateOneDealCancellation(dealId: string | ObjectId, update: UpdateFilter<TfmDealCancellation>): Promise<UpdateResult> {
+    const dealCollection = await this.getCollection();
+
+    const updateResult = await dealCollection.updateOne(
+      {
+        _id: { $eq: new ObjectId(dealId) },
+        'tfm.stage': { $ne: 'Cancelled' },
```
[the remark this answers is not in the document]
ttbarnes: Resolving this as the comment below says the same thing

trade-finance-manager-api/src/v1/api.js
```diff
@@ -269,6 +269,33 @@ const submitDeal = async (dealType, dealId, auditDetails) => {
...
+    url: `${DTFS_CENTRAL_API_URL}/v1/tfm/deals/${dealId}/cancellation`,
+    headers: headers.central,
+    data: {
+      dealCancellationUpdate,
+      auditDetails,
+    },
+  });
+
+  return response.data;
+};
```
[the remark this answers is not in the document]
BethThomas141: The only place it is called (and will ever be called) is in the tfm-api controller, and this wraps it in a try-catch block there. So adding it here too would be redundant and reduce DRYness of the code. Some of the recently added items to this file such as `markKeyingDataAsDone` also follow this cleaner approach of just using try/catch a level up 🙂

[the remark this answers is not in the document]
ttbarnes: Resolving in favour of the duplicate comment above

trade-finance-manager-api/src/v1/api.js
ttbarnes: I left the same comment - would also like a try/catch here please like the majority of other functions in this file; Helpful for debugging even if it's only called in one place

trade-finance-manager-api/src/v1/controllers/deal-cancellation.controller.test.ts
```diff
@@ -0,0 +1,108 @@
...
+  effectiveFrom: 1794418808,
+};
+
+const mockUpdateResult = {
+  acknowledged: true,
+  modifiedCount: 1,
+  upsertedId: new ObjectId(),
+  upsertedCount: 0,
+  matchedCount: 1,
+};
```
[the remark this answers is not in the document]
BethThomas141: Hmm due to this being in TypeScript and us using the `UpdateResult` type I can't easily change this here without changing the types we expect elsewhere. Are MongoDB known for changing their response types?

trade-finance-manager-api/src/v1/deals/routes.js
BethThomas141: It will be, we currently don't have any nice middleware validation to add in the tfm-api layer to check this. We've got a subtask on the ticket to create middleware on the tfm-api layer to do this

trade-finance-manager-ui/server/api.js
BethThomas141: Added 👍

trade-finance-manager-api/src/v1/controllers/deal-cancellation.controller.ts
BethThomas141: Added 👍

trade-finance-manager-api/src/v1/api.js
BethThomas141: Added 👍

trade-finance-manager-api/src/types/express.d.ts
```diff
@@ -0,0 +1,9 @@
+import { TfmSessionUser } from './tfm-session-user.ts';
+
+declare global {
```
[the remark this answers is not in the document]
BethThomas141: Added 👍

dtfs-central-api/src/v1/routes/tfm-routes.js
```diff
@@ -530,6 +535,50 @@ tfmRouter
...
+ *               bankRequestDate:
+ *                 type: number
+ *                 example: 1725977352
+ *               effectiveFrom:
+ *                 type: number
+ *                 example: 1725977352
+ *     responses:
+ *       200:
+ *         description: OK
+ *       404:
```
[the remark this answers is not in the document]
BethThomas141: Have added that to the responses

dtfs-central-api/src/v1/routes/tfm-routes.js
```diff
@@ -530,6 +535,50 @@ tfmRouter
     tfmPutAmendmentController.updateTfmAmendment,
   );
 
+/**
+ * @openapi
+ * /tfm/deals/:id/cancellation:
+ *   put:
+ *     summary: Updates tfm deal cancellation object
+ *     tags: [TFM, deals, cancellation]
```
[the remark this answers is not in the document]
BethThomas141: Added 👍

dtfs-central-api/src/v1/routes/tfm-routes.js
```diff
@@ -530,6 +535,50 @@ tfmRouter
     tfmPutAmendmentController.updateTfmAmendment,
   );
 
+/**
+ * @openapi
+ * /tfm/deals/:id/cancellation:
+ *   put:
+ *     summary: Updates tfm deal cancellation object
```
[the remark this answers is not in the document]
BethThomas141: Added 👍

dtfs-central-api/src/v1/controllers/tfm/deal-cancellation/tfm-put-update-deal-cancellation.controller.ts
```diff
@@ -0,0 +1,43 @@
...
+    return res.status(HttpStatusCode.Ok).json(updateResult);
+  } catch (error) {
+    console.error('Error updating deal cancellation:', error);
+    if (error instanceof ApiError) {
+      const { status, message, code } = error;
+      return res.status(status).send({ status, message, code });
+    }
+    return res.status(HttpStatusCode.InternalServerError).send({
+      status: HttpStatusCode.InternalServerError,
+      message: 'Unknown error',
```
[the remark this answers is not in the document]
BethThomas141: We have to send the `message` as a string and if you search the code base for If you search the code base for `return res.status(HttpStatusCode.InternalServerError).send({` we do generally send 500 errors with a custom message like this. I have updated the message to be a bit clearer.

dtfs-central-api/src/v1/routes/middleware/payload-validation/validate-put-deal-cancellation-payload.ts
```diff
@@ -0,0 +1,19 @@
+import z from 'zod';
+import { AuditDetails, TfmDealCancellation } from '@ukef/dtfs2-common';
+import { AuditDetailsSchema } from './schemas';
+import { createValidationMiddlewareForSchema } from './create-validation-middleware-for-schema.ts';
+
+const PutDealCancellationSchema: z.ZodType<{ dealCancellationUpdate: Partial<TfmDealCancellation>; auditDetails: AuditDetails }> = z.object({
```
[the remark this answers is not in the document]
BethThomas141: This is custom middleware to check the types of the fields we are sending which I think is good to have, so would like to keep as is.

dtfs-central-api/src/v1/controllers/tfm/deal-cancellation/tfm-put-update-deal-cancellation.controller.ts
```diff
@@ -0,0 +1,43 @@
...
+type UpdateTfmDealCancellationResponse = Response<ApiErrorResponseBody | UpdateResult>;
+
+export const updateTfmDealCancellation = async (req: UpdateTfmDealCancellationRequest, res: UpdateTfmDealCancellationResponse) => {
+  const { dealCancellationUpdate, auditDetails } = req.body;
+  const { dealId } = req.params;
+
+  try {
+    validateAuditDetailsAndUserType(auditDetails, AUDIT_USER_TYPES.TFM);
+
+    const updateResult = await TfmDealCancellationRepo.updateOneDealCancellation(
```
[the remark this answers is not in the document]
BethThomas141: I'm actually not sure how to / if that's possible - `TfmDealCancellationRepo` is a class and `TfmDealCancellationRepo` is a member of it (Is there a way to just import a class member? )

dtfs-central-api/src/repositories/tfm-deals-repo/tfm-deal-cancellation.repo.ts
```diff
@@ -0,0 +1,53 @@
...
+        'dealSnapshot.submissionType': { $in: [DEAL_SUBMISSION_TYPE.AIN, DEAL_SUBMISSION_TYPE.MIN] },
+      },
+      update,
+    );
+
+    if (updateResult.matchedCount === 0) {
```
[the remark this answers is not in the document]
BethThomas141: Changed 👍

dtfs-central-api/src/repositories/tfm-deals-repo/tfm-deal-cancellation.repo.ts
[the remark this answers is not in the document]
BethThomas141: Changed 👍

dtfs-central-api/src/repositories/tfm-deals-repo/tfm-deal-cancellation.repo.ts
```diff
@@ -0,0 +1,53 @@
...
+    return matchingDeal.tfm.cancellation;
+  }
+
+  /**
+   * Updates the deal tfm object with the supplied cancellation
+   * @param dealId - The deal id
+   * @param update - The deal cancellation update to apply
+   * @returns The update result
+   */
+  public static async updateOneDealCancellation(dealId: string | ObjectId, update: UpdateFilter<TfmDealCancellation>): Promise<UpdateResult> {
```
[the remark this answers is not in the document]
BethThomas141: have replied above, will resolve here as I'll make the changes in both places if we reach a conclusion on the other thread

dtfs-central-api/src/repositories/tfm-deals-repo/tfm-deal-cancellation.repo.ts
```diff
@@ -0,0 +1,53 @@
+import { Collection, ObjectId, UpdateFilter, UpdateResult, WithoutId } from 'mongodb';
+import { DEAL_SUBMISSION_TYPE, DealNotFoundError, TfmDeal, TfmDealCancellation } from '@ukef/dtfs2-common';
+import { mongoDbClient } from '../../drivers/db-client';
+
+export class TfmDealCancellationRepo {
+  private static async getCollection(): Promise<Collection<WithoutId<TfmDeal>>> {
+    return await mongoDbClient.getCollection('tfm-deals');
```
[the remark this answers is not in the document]
BethThomas141: Ah nice didn't know that existed. Changed 👍

trade-finance-manager-api/api-tests/v1/deal-cancellation/deal-cancellation.api-test.ts
```diff
@@ -0,0 +1,57 @@
...
+      // Arrange
+      const tokenUser = await testUserCache.initialise(app);
+      const url = getTfmDealCancellationUpdateUrl({ id: validId });
+
+      // Act
+      const response = await as(tokenUser).put(payload).to(url);
+
+      // Assert
+      expect(response.status).toEqual(200);
+      expect(response.body).not.toBeNull();
```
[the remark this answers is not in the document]
BethThomas141: Actually the body it returns isn't important as it's a `PUT` request. So I've just removed that line to check that the response is 200

libs/common/src/types/mongo-db-models/deals.ts
```diff
@@ -0,0 +1,9 @@
+import { AnyObject } from '../any-object';
+
+/**
+ * Type of the mongo db "deal" collection
+ *
+ * This type is likely incomplete and should be added
+ * to as and when new properties are discovered
+ */
+export type Deal = AnyObject;
```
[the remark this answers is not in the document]
BethThomas141: Yep can create a ticket in the backlog!

dtfs-central-api/src/v1/controllers/tfm/deal-cancellation/tfm-put-update-deal-cancellation.controller.ts
```diff
@@ -0,0 +1,43 @@
...
+import { TfmDealCancellationRepo } from '../../../../repositories/tfm-deals-repo';
+import { PutDealCancellationPayload } from '../../../routes/middleware/payload-validation';
+
+type UpdateTfmDealCancellationRequest = CustomExpressRequest<{
+  reqBody: PutDealCancellationPayload;
+}>;
+
+type UpdateTfmDealCancellationResponse = Response<ApiErrorResponseBody | UpdateResult>;
+
+export const updateTfmDealCancellation = async (req: UpdateTfmDealCancellationRequest, res: UpdateTfmDealCancellationResponse) => {
```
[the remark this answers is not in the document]
BethThomas141: Added 👍

dtfs-central-api/src/repositories/tfm-deals-repo/tfm-deal-cancellation.repo.ts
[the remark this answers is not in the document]
BethThomas141: Good idea, now added 👍

dtfs-central-api/api-tests/v1/tfm/deal-cancellation/tfm-deal-cancellation-put.api-test.js
```diff
@@ -0,0 +1,100 @@
...
+const { testApi } = require('../../../test-api');
+const { DEALS } = require('../../../../src/constants');
+const aDeal = require('../../deal-builder');
+const { createDeal } = require('../../../helpers/create-deal');
+const { aPortalUser, aTfmUser } = require('../../../../test-helpers');
+const { MOCK_PORTAL_USER } = require('../../../mocks/test-users/mock-portal-user');
+
+console.error = jest.fn();
+
+describe('PUT TFM deal cancellation', () => {
```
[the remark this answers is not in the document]
BethThomas141: Good spot have changed!

trade-finance-manager-api/api-tests/v1/deal-cancellation/deal-cancellation.api-test.ts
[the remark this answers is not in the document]
ttbarnes: Can we please try to add a test - appreciate it's a PUT request, but it's still returning something via `.send(dealCancellationResponse)`

dtfs-central-api/src/repositories/tfm-deals-repo/tfm-deal-cancellation.repo.ts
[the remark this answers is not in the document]
ttbarnes: I suggest for this we stay aligned with existing approach

If we want to do something different we should make the change in a separate PR.

dtfs-central-api/src/v1/controllers/tfm/deal-cancellation/tfm-put-update-deal-cancellation.controller.ts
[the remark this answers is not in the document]
ttbarnes: Makes sense @BethThomas141 , but what if this function is called without previously calling `updateOneDealCancellation`

Maybe the check for this (in `updateOneDealCancellation`), should be in this function?

dtfs-central-api/src/v1/controllers/tfm/deal-cancellation/tfm-put-update-deal-cancellation.controller.ts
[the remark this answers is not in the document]
ttbarnes: Seems that all the instances of e.g `xRepo.doSomething` are called like this, I suggest we stick to this.

Not sure if it's possible to destructure either - if it is and we want to do this, we should do in a separate PR.

trade-finance-manager-api/src/v1/deals/routes.js
ttbarnes: Resolving this, I left the same comment - this will be added in another ticket/PR to keep PRs small 👍 @abhi-markan

trade-finance-manager-ui/server/api.js
ttbarnes: I think this would only be beneficial, regardless of how it's called and if the caller is wrapped in a try/catch. Seems that half of the functions in this file have a try/catch, the other half don't. I think having a try/catch cannot hurt. Helpful for debugging

trade-finance-manager-api/src/v1/api.js
ChristianMcc: This came up a fair amount almost a year ago about the error handling in these files and how it's not very good - we're returning a hand rolled status object instead of letting the axios error propagate and losing what axios gives us for free in the error object. The benefit of that as well is with the axios error it's very obvious where the error came from.
If the answer here is we want a consistent (but genuinely worse imo) pattern in these files that's fine but feel the need to point out the above in these files. (This applies to the other api.js file in tfm-ui too but won't copy paste a comment :D )

ttbarnes: Makes sense to me @ChristianMcc 👍

dtfs-central-api/src/v1/controllers/tfm/deal-cancellation/tfm-put-update-deal-cancellation.controller.ts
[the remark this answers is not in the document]
BethThomas141: So this function will go on to call `updateOneDealCancellation` in the central layer where it will be checked. I think the check there is enough and would prevent us otherwise calling the `getDeals` endpoint from another Repo, finding the matching deal and checking the state here which would add a bit of complexity. 

On the `tfm-ui` layer we won't be showing the deal cancellation button or any of the pages if the deal state is `Cancelled`, so this isn't a scenario we expect unless people are calling endpoints directly, and in either case the check on the central layer will prevent them cancelling the deal.

dtfs-central-api/src/v1/routes/middleware/payload-validation/validate-put-deal-cancellation-payload.ts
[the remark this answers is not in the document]
abhi-markan: The current implementation is good and I can see it is being used extensively my only concern is we seems have three different methods to validate a payload which becomes a tech-debt in my opinion.

libs/common/src/types/tfm-deal-cancellation.ts
```diff
@@ -0,0 +1,5 @@
+export type TfmDealCancellation = {
```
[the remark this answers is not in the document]
abhi-markan: Agree needs a re-think one for a conversation.

abhi-markan (changes requested):

trade-finance-manager-api/api-tests/v1/deal-cancellation/deal-cancellation.api-test.ts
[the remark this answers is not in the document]
BethThomas141: So it returns an empty object on success, which we can test? But if you look at all other similar test files in the tfm-api api-tests, they all either just check `expect(response.status).toEqual(200);` or that and the `expect(response.body).not.toBeNull();` check. (e.g. `get-all-banks.api-test.ts`, `feedback.api-test.js`)

I've added the not null check back in, and I believe that to be consistent with the other tests here. Let me know what you think!

trade-finance-manager-api/src/v1/api.js
ChristianMcc: Yep part of my points were about the custom error handling rather than the rethrowing, but even on the the catch and re-throw we do lose access to the stack trace (without some custom code being written). The stack trace can often be very useful for debugging if you are a few functions deep so where possible I'd still avoid catching and re-throwing unless we really really need some extra context in that particular function/extra behaviour

dtfs-central-api/src/v1/controllers/tfm/deal-cancellation/tfm-put-update-deal-cancellation.controller.ts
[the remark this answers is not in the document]
BethThomas141: Just had a chat with @AlexBramhill about this to improve my own understanding - there's a reason we don't want to throw the error message using `error` as there's a risk we might pass on sensitive data from the error to tfm-ui. Tfm-ui doesn't care for the reason of the failure, just that it's failed with a 500.

trade-finance-manager-api/src/v1/api.js
BethThomas141: @abhi-markan @ttbarnes I've added a try/catch block here as discussed on Friday 👍

trade-finance-manager-ui/server/api.js
BethThomas141: @abhi-markan @ttbarnes I've added a try/catch block here as discussed on Friday 👍

trade-finance-manager-api/api-tests/v1/deal-cancellation/deal-cancellation.api-test.ts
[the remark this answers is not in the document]
BethThomas141: @ttbarnes I've now mocked the API response at the top of the file so it now tests the expected received object correctly. Good idea to do that, thanks!👍

trade-finance-manager-ui/server/api.js
abhi-markan: Thank you @BethThomas141

trade-finance-manager-api/src/v1/api.js
```diff
@@ -269,6 +269,39 @@ const submitDeal = async (dealType, dealId, auditDetails) => {
...
+      data: {
+        dealCancellationUpdate,
+        auditDetails,
+      },
+    });
+
+    return response.data;
+  } catch (error) {
+    console.error(error);
+    throw error;
```
ttbarnes, line 301: Thanks @BethThomas141 , but this seems to be a new approach compared to the other functions, there are so many different approaches here 😞 => 

```js
// new approach in this PR
} catch (error) {
  console.error();
  throw error;
}

// existing
} catch ({ response }) {
  console.error();

  return false;
}

// existing
} catch (error) {
  console.error();
  return onError( ... );
}

// existing
} catch (error) {
  console.error();
  return { ... };
}
```

Can we use one of the existing approaches please so that we don't introduce another new approach

ttbarnes (dismissed):

abhi-markan (dismissed):

ttbarnes (dismissed):

ttbarnes (approved):

ttbarnes (dismissed):

trade-finance-manager-api/api-tests/v1/deal-cancellation/deal-cancellation.api-test.ts
```diff
@@ -0,0 +1,123 @@
...
+import { generateTfmAuditDetails } from '@ukef/dtfs2-common/change-stream';
+import createApi from '../../api';
+import app from '../../../src/createApp';
+import testUserCache from '../../api-test-users';
+import { MOCK_TFM_SESSION_USER } from '../../../src/v1/__mocks__/mock-tfm-session-user.ts';
+
+const updateDealCancellationMock = jest.fn();
+const isTfmDealCancellationFeatureFlagEnabledMock = jest.fn();
+
+// eslint-disable-next-line @typescript-eslint/no-unsafe-return
```
abhi-markan, line 11: Just noticed these, can they please be removed.

abhi-markan (changes requested):

ttbarnes (approved):

dtfs-central-api/src/v1/routes/tfm-routes.js
```diff
@@ -530,6 +536,53 @@ tfmRouter
...
+ *         description: OK
+ *       404:
+ *         description: Not found
+ *       500:
+ *         description: Internal server error
+ */
+tfmRouter
+  .route('/deals/:dealId/cancellation')
+  .put(
+    validateDealCancellationEnabled,
```
abhi-markan, line 579: Should we add PIM role in `dtfs-central-api` as an another safety net?

abhi-markan (changes requested):
Two minor comments for clarification please.

trade-finance-manager-api/src/v1/deals/routes.js
```diff
@@ -76,6 +78,10 @@ dealsAuthRouter
...
+  .route('/deals/:dealId/cancellation')
+  .put(validateDealCancellationEnabled, validation.dealIdValidation, handleExpressValidatorResult, dealCancellationController.updateDealCancellation);
```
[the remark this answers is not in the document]
MarRobSoftwire: https://github.com/UK-Export-Finance/dtfs2/pull/3498/files will address this

abhi-markan (approved):

PR #3472, opened 2024-09-10
chore(tests): simplify some unit test descriptions, align file names

ttbarnes:
## Introduction :pencil2:
- Some unit tests had an "it" in the description e.g `it('it should ...`
- Some util mocks had a different file name convention (using _ instead of -)

- Simplify unit test "it" blocks.
- Rename util mocks.


dtfs-central-api/src/services/utilisation-report-data-validator/utilisation-report-cell-validators/generate-ukef-facility-id-error.test.ts
```diff
@@ -52,7 +52,7 @@ describe('generateUkefFacilityIdError', () => {
     expect(ukefFacilityIdError).toBeNull();
   });
 
-  it('it returns the correct column and row when an error is found', () => {
+  it('returns the correct column and row when an error is found', () => {
```
Zainzzkk, line 55: can we start making this 'should return' so its ready for when that standard is introduced?

ttbarnes: Would like to do this in a separate PR and when we don't have a lot of PRs waiting to merge. There will be a lot of things to check and change, for example there are 779 instances of  `it('returns` alone

Zainzzkk (approved):

PR #3482, opened 2024-09-11
chore(tests): split up e2e tests - unissued facilities

ttbarnes:
## Introduction :pencil2:
The "unissued facilities" GEF E2E tests seem to be the longest running E2E tests, taking ~18-23minutes to run.

This often results in having to wait for these tests to pass in order to merge a PR or re-run other tests.

These changes now take the following times to execute, instead of 18-23 mins:

- Suite 1 = 5mins
- Suite 2 = 3mins
- Suite 3 = 1min
- Suite 4 = 5mins
- Suite 5 = 4.5mins

This is part of a series of E2E test splitting to reduce E2E GHA time.

- Split up the "unissued facilities" into AIN and MIN directories.

- Ensure that all of the tests hav the deal type (AIN or MIN) in the test description.


.github/workflows/test.yml
```diff
@@ -732,7 +732,9 @@ jobs:
             './gef/cypress/e2e/eligibility/**/*.spec.js',
             './gef/cypress/e2e/exporter/**/*.spec.js',
             './gef/cypress/e2e/external/**/*.spec.js',
-            './gef/cypress/e2e/facility/**/*.spec.js',
+            './gef/cypress/e2e/facility/*.spec.js',
```
MarRobSoftwire, line 735: Could you also split apart the non-ff tests too?

MarRobSoftwire: Also would it be more clear if there was a subdirectory for the remaining tests e.g 
```suggestion
            './gef/cypress/e2e/facility/pre-submission/**/*.spec.js',
```

Though some of the tests wouldn't match this description (e.g.review-ukef-approval-with-issued-and-unissued-facilities-MIA-and-resubmit-to-ukef.spec.js) & are not pre-submission - they could be moved to `unissued-facilities).

MarRobSoftwire (commented):
Thanks for doing this :) These tests had got huge with the latest facility end date changes

.github/workflows/test.yml
ttbarnes: Thanks @MarRobSoftwire great shout, i've updated both E2E suites - agree some others could be split up, i've created the following directories to keep inline with the existing test names. It's structured like this now:

```
facility/*.spec.js
facility/end-date/*.spec.js
facility/review-ukef-approval/*.spec.js
facility/version-0/*.spec.js
facility/unissued-facilities/AIN/*.spec.js
facility/unissued-facilities/MIN/*.spec.js
```

What do you think?



Also, before merging this, want to monitor the time differences to see if we can make further improvements before merging.

abhi-markan: Any `ff` specific E2E will need to be added to `e2e-tests-ff` job please, so we run all explicit truthy FF E2E coverage there and elsewhere it is standard system state.

abhi-markan (changes requested):

.github/workflows/test.yml
MarRobSoftwire: This sounds great, what's left in `facility/*.spec.js`? Could they be moved into something like `'./gef/cypress/e2e/facility/pre-submission/**/*.spec.js'`?

Could `facility/end-date/*.spec.js` be renamed to `facility/facility-end-date/*.spec.js` & include the `bank-review-date` test? `facility/facility-end-date/*.spec.js` will not need to be tested when the ff is off

ttbarnes: @abhi-markan It's already been added to `e2e-tests-ff`, it's hard to see with the git diff, need to expand previous lines 

https://github.com/UK-Export-Finance/dtfs2/pull/3482/commits/1056fb64084b8640fd24bdbdb5ba785692961ce6#diff-faff1af3d8ff408964a57b2e475f69a6b7c7b71c9978cccc8f471798caac2c88R714

ttbarnes: Thanks @MarRobSoftwire 

I'm not keen on adding a `pre-submission` directory, I think it could be misleading/confusing because some existing tests/directories are "pre submission".

Have moved the `bank-review-date` test into the `end-date` directory, good shout

I think adding a `facility-` prefix to the `end-date` directory would be strange - in this case i'd expect all the directories to have a prefix and it'd become this which seems overkill because we have a `facility` parent directory

```js
facility/facility-end-date
facility/facility-review-ukef-approval
facility/facility-unissued-facilities
facility/facility-version-0
``` 

The existing structure/naming feels cleaner/simpler:

```js
facility/end-date
facility/review-ukef-approval
facility/unissued-facilities
facility/version-0
```

MarRobSoftwire: `facility-end-date` is a field the user submits (e.g. see https://ukef-dtfs.atlassian.net/browse/DTFS2-7220 ), like `bank-review-date` or `cover-start-date`. By calling it `end-date` this could be confused with `cover-end-date` (which is sometimes referred to as `endDate` in existing code). Moreover, if I'm searching for facility end date tests I wouldn't expect them to be under `end-date`

As to which tests are pre-submission, I think its only the facility end date ones so we could just put them in the same directory. Its more personal preference but I don't really like having some tests existing at a different level to others. I think it could encourage laziness to put a new facility test suite in the top directory, rather than checking if it could exist a level down.

Now I think about, the facility-end-date directory is maybe unnecessary and could be a point of confusion - `unissued-facilities-facility-end-date-AIN.spec.js` could live in the `unissued-facilities` or the `facility-end-date` directory. The tests currently in `facility-end-date` could just be moved into a pre-submission directory.

ttbarnes: Thanks @MarRobSoftwire agreed, I have renamed `facility-end-date` to `pre-submission`.

I know what you mean about having tests in the parent directory, I had a think and I cannot think of a sensible way to group these tests (any ideas?), it seems they have nothing in common. In which case I think it makes sense to have them in a parent `facility` directory.

Sorry for any confusion, I am unfamiliar with theses tests, the intention of this PR is to split them up so that they run faster in GHA.

.github/workflows/test.yml
```diff
@@ -503,7 +503,11 @@ jobs:
             'eligibility/**/*.spec.js',
             'exporter/**/*.spec.js',
             'external/**/*.spec.js',
-            'facility/**/*.spec.js',
+            'facility/pre-submission/*.spec.js',
+            'facility/review-ukef-approval/*.spec.js',
+            'facility/version-0/*.spec.js',
```
abhi-markan, line 508: What is `version-0` please?

ttbarnes: it's a version 0 deal

MarRobSoftwire (approved):

abhi-markan (approved):

PR #3493, opened 2024-09-12
feat(FN-3400): seed utilisation report test facilities against one deal

RLavender98:
## Introduction :pencil2:
The `create-keying-sheet-tfm-facilities` seeding script was creating not quite valid facilities and adding them to random existing deals. This was not ideal as was causing errors locally when loading deals and facilities and also making it hard to separate from other mock data.

- Insert a deal specifically for the facilities needed for the utilisation report seeding data
- Add all facilities to this specific deal
- Use realistic data by copying from manually created deals/facilities from mongo db and only making minor modifications

- Set utilisation data seeded by the sql-db-seeding script to use the previous report period not current so it is ready for keying sheet generation


utils/create-keying-sheet-tfm-facilities/helpers/get-portal-user-id.ts
```diff
@@ -0,0 +1,12 @@
+import { PortalUser } from '@ukef/dtfs2-common';
+import { ObjectId } from 'mongodb';
+import { mongoDbClient } from '../database-client';
+
+export const getPortalUserIdOrFail = async (): Promise<ObjectId> => {
```
ChristianMcc, line 5: Where is this used? Wonder if we need to be careful with what type of user we are attributing to the deal (and if someone with a wrong role attached to an action might cause errors later?) let me know if I'm overthinking this

utils/sql-db-seeder/src/fee-record-payment-group/fee-record-payment-group.seed.ts
```diff
@@ -1,76 +1,186 @@
...
+  UtilisationReportEntity,
+  getDateFromMonthAndYear,
+  getPreviousReportPeriodForBankScheduleByMonth,
+  toIsoMonthStamp,
+} from '@ukef/dtfs2-common';
+import { subMonths } from 'date-fns';
 import { FeeRecordPaymentGroupSeeder } from './fee-record-payment-group.seeder';
+import { MongoDbDataLoader } from '../mongo-db-client';
+
+const getPreviousReportPeriod = (reportingSchedule: BankReportPeriodSchedule, reportPeriod: ReportPeriod): ReportPeriod => {
```
ChristianMcc, line 14: Can you add a docstring please

utils/sql-db-seeder/src/fee-record-payment-group/fee-record-payment-group.seed.ts
```diff
@@ -1,76 +1,186 @@
...
+  const startOfReportPeriod = getDateFromMonthAndYear(reportPeriod.start);
+  const dateInPreviousReportingPeriod = subMonths(startOfReportPeriod, 1);
+  return getPreviousReportPeriodForBankScheduleByMonth(reportingSchedule, toIsoMonthStamp(dateInPreviousReportingPeriod));
+};
 
 export const seedFeeRecordPaymentGroups = async (dataSource: DataSource) => {
```
ChristianMcc, line 20: ditto even though it's existing

ChristianMcc (commented):
This approach looks good to me, separates us out from all the other deals added and means we actually have one in TFM which is nice! Don't think we can have any other approach which won't take a huge amount of time to implement

Nice work, left a minor questions and nit picks

utils/create-keying-sheet-tfm-facilities/helpers/get-portal-user-id.ts
RLavender98: It's actually only used in the audit records, so is perhaps actually overkill anyway since I'm not sure we parse them? But basically I replaced any ids that were pulled as object ids with real object ids. There are other ids stored as just string which I have not made dynamically accurate e.g. deal has a nested field inside `portalActivities` for the checker on one of the objects in there. Just opening the deal and looking didn't throw any errors. I am hoping if someone wanted to test something about a deal they would not continue a journey from this deal however, and instead would ideally use a fresh one.

utils/create-keying-sheet-tfm-facilities/index.ts
```diff
@@ -1,30 +1,41 @@
...
 import { SqlDbDataSource } from '@ukef/dtfs2-common/sql-db-connection';
 import { Facility, TfmFacility } from '@ukef/dtfs2-common';
-import { FacilityClient, mongoDbClient, TfmFacilityClient } from './database-client';
-import { generateRandomTfmFacilityForFacility, getSqlFacilityIds, getValidGefDealIdGenerator } from './helpers';
-import { aFacilityWithoutDealId } from './mock-data';
+import { FacilityClient, mongoDbClient, TfmFacilityClient, DealClient, TfmDealClient } from './database-client';
+import { generateRandomTfmFacilityForFacility, getSqlFacilityIds, getPortalUserIdOrFail } from './helpers';
+import { aDeal, aFacility, aTfmDeal } from './mock-data';
 
 const run = async () => {
```
Zainzzkk, line 8: documentation for this function please?

utils/create-keying-sheet-tfm-facilities/index.ts
```diff
@@ -1,30 +1,41 @@
...
+    const tfmFacilities: WithoutId<TfmFacility>[] = facilities.map((facility) => generateRandomTfmFacilityForFacility(facility, portalUserId));
     await Promise.all(tfmFacilities.map((tfmFacility) => TfmFacilityClient.insertIfNotExists(tfmFacility)));
+
+    const deal = aDeal(ukefDealId, dealId, portalUserId);
+    await DealClient.insertIfNotExists(deal);
+
+    const tfmDeal = { ...aTfmDeal(dealId, portalUserId, deal) };
+    await TfmDealClient.insertIfNotExists(tfmDeal);
   } catch (error) {
     console.error('Failed to create tfm facilities:', error);
```
Zainzzkk, line 40: %o after `facilities:` please

utils/create-keying-sheet-tfm-facilities/database-client/tfm-deal-client.ts
```diff
@@ -0,0 +1,31 @@
...
+  public static async insertIfNotExists(tfmDeal: TfmDeal): Promise<void> {
+    if (!this.collection) {
+      throw new Error('Tfm deal client has not yet been initialised');
+    }
+
+    const { ukefDealId } = tfmDeal.dealSnapshot;
+
+    const numberOfDeals = await this.collection.count({ 'dealSnapshot.ukefDealId': { $eq: ukefDealId } });
+    if (numberOfDeals !== 0) {
+      console.info(`Tfm deal with ukef deal id '${ukefDealId}' already exists`);
```
Zainzzkk, line 24: Can this be: `console.info('Deal %s already exists in TFM', ukefDealId);`

utils/create-keying-sheet-tfm-facilities/database-client/tfm-deal-client.ts
```diff
@@ -0,0 +1,31 @@
...
+      return;
+    }
+
+    console.info(`Inserting tfm deal with ukef deal id '${ukefDealId}'`);
```
Zainzzkk, line 28: Can this be ` console.info('Inserting deal %s into TFM', ukefDealId);`

utils/create-keying-sheet-tfm-facilities/database-client/deal-client.ts
```diff
@@ -0,0 +1,31 @@
...
+  public static async insertIfNotExists(deal: Deal): Promise<void> {
+    if (!this.collection) {
+      throw new Error('Deal client has not yet been initialised');
+    }
+
+    const { ukefDealId } = deal;
+
+    const numberOfDeals = await this.collection.count({ ukefDealId: { $eq: ukefDealId } });
+    if (numberOfDeals !== 0) {
+      console.info(`Facility with UKEF facility id '${ukefDealId}' already exists`);
```
Zainzzkk, line 24: Should this be ukefDealId? and can it use the same %s format as below please

utils/create-keying-sheet-tfm-facilities/database-client/deal-client.ts
```diff
@@ -0,0 +1,31 @@
...
+      return;
+    }
+
+    console.info(`Inserting deal with ukef deal id '${ukefDealId}'`);
```
Zainzzkk, line 28: same as above

CHANGELOG.md
```diff
@@ -2,26 +2,24 @@
...
+- **FN-3447:** add utilisation tab data to response ([#3651](https://github.com/UK-Export-Finance/dtfs2/issues/3651)) ([5f1a9e7](https://github.com/UK-Export-Finance/dtfs2/commit/5f1a9e729b853619d490309f9ea8f493f022a02d))
+- **FN-3447:** utilisation tab e2e tests ([#3666](https://github.com/UK-Export-Finance/dtfs2/issues/3666)) ([d717e6e](https://github.com/UK-Export-Finance/dtfs2/commit/d717e6e46d36997705aec73a89ef32838c20b85e))
+- **FN-3447:** utilisation tab table ([#3652](https://github.com/UK-Export-Finance/dtfs2/issues/3652)) ([005f64a](https://github.com/UK-Export-Finance/dtfs2/commit/005f64a70a3a276f3693037fc8d793ed45085d5e))
 
 ### Bug Fixes
 
-* **FN-3422:** calculate fixed fee from report period end date ([#3650](https://github.com/UK-Export-Finance/dtfs2/issues/3650)) ([92b22f2](https://github.com/UK-Export-Finance/dtfs2/commit/92b22f204d52019e67f2e814e77e7befa97a165c))
+- **FN-3422:** calculate fixed fee from report period end date ([#3650](https://github.com/UK-Export-Finance/dtfs2/issues/3650)) ([92b22f2](https://github.com/UK-Export-Finance/dtfs2/commit/92b22f204d52019e67f2e814e77e7befa97a165c))
```
RLavender98, line 22: Not sure where this change has come from, will investigate

utils/create-keying-sheet-tfm-facilities/mock-data/deal.ts
```diff
@@ -0,0 +1,262 @@
...
+/**
+ * Creates an active deal that can be used for utilisation reporting testing.
+ * It can be replaced by creating a deal manually and then pulling from the mongo db collection.
+ * This way we ensure we are inserting a valid deal.
+ * @param ukefDealId - ukef deal id to use
+ * @param dealId - id of deal
+ * @param portalUserId - id of a portal user
+ * @returns a deal
+ */
+export const aDeal = (ukefDealId: string, dealId: ObjectId, portalUserId: ObjectId): Deal => ({
```
ChristianMcc, line 13: Love this function, think it's definitely the best way to do this seeding even if it might go out of date. Just wanted to comment to say writing the shape of deals into this format so you can see what it is clearly makes me realise it's grim aha

ChristianMcc (approved):
LGTM :) thanks for improving this seeding

utils/create-keying-sheet-tfm-facilities/mock-data/deal.ts
```diff
@@ -0,0 +1,262 @@
...
+  _id: dealId,
+  dealType: 'GEF',
+  version: 1,
+  maker: {
+    username: 'maker1@ukexportfinance.gov.uk',
```
Zainzzkk, line 18: could any of these come out of other files.  For example 

`const USERS = require('../portal-users');

const A_MAKER = USERS.BANK1_MAKER1;`

utils/create-keying-sheet-tfm-facilities/mock-data/tfm-facility.ts
```diff
@@ -0,0 +1,36 @@
...
+ * @param facilitySnapshot - a facility
+ * @param portalUserId - id of a portal user
+ * @returns a tfm facility
+ */
+export const aTfmFacility = (facilityId: ObjectId, facilitySnapshot: Facility, portalUserId: ObjectId): TfmFacility => ({
+  _id: facilityId,
+  facilitySnapshot,
+  tfm: {
+    hasBeenIssuedAndAcknowledged: true,
+    feeRecord: 5622.222222222223,
```
Zainzzkk, line 18: should this number be not rounded?

utils/create-keying-sheet-tfm-facilities/mock-data/deal.ts
RLavender98: I intentionally didn't do that because it would be a lot of effort and then if we ever wanted to replace with a fresh one would have to do the whole replacement job again given to ensure this is valid it is pulled from the database on dev.

utils/create-keying-sheet-tfm-facilities/mock-data/tfm-facility.ts
RLavender98: This facility was created by creating a facility on deployed env and then copying from mongo. If this should be being rounded then that would be a bug completely unrelated to this work.

Zainzzkk (approved):

PR #3471, opened 2024-09-10
feat(DTFS2-7348): add reason for cancelling template

MarRobSoftwire:
## Introduction :pencil2:
Include a summary of the changes and related feature(s) or issue(s).

## Resolution :heavy_check_mark:
List all changes made to the codebase.

## Miscellaneous :heavy_plus_sign:
List any additional fixes or improvements.

PR #3494, opened 2024-09-12
feat(DTFS2-7349): add validation to reason for cancelling

MarRobSoftwire:
## Introduction :pencil2:
The input should be validated when the user submits a reason for cancelling over 1200 characters

- Add post endpoint
- Add validation helper, called by post endpoint

## Note ⚠️ 
This is merged off #3471


trade-finance-manager-ui/server/controllers/case/cancellation/validation/validate-reason-for-cancelling.ts
```diff
@@ -0,0 +1,19 @@
+import { ErrorSummaryViewModel, ReasonForCancellingErrorsViewModel } from '../../../../types/view-models';
+
+const MAXIMUM_LENGTH = 1200;
```
abhi-markan, line 3: Is there `libs/common` constant for this, if not might be worth importing there as it will be quite common for current and future textareas.

trade-finance-manager-ui/server/controllers/case/cancellation/validation/validate-reason-for-cancelling.ts
```diff
@@ -0,0 +1,19 @@
...
+const REASON_TOO_LONG_MESSAGE = `Reason for cancelling must be ${MAXIMUM_LENGTH} characters or less`;
+
+/**
+ * @param reason The reason for cancelling
+ * @returns a reason for cancelling errors view model
+ */
+export const validateReasonForCancelling = (reason: string | undefined): ReasonForCancellingErrorsViewModel => {
```
abhi-markan, line 10: Why is the reason `undefined`?

trade-finance-manager-ui/server/controllers/case/cancellation/validation/validate-reason-for-cancelling.ts
```diff
@@ -0,0 +1,19 @@
...
+  const errorSummary: ErrorSummaryViewModel[] = [];
+
+  const reasonForCancellingErrorMessage = reason && reason.length > MAXIMUM_LENGTH ? REASON_TOO_LONG_MESSAGE : undefined;
```
abhi-markan, line 13:
```suggestion
  const reasonForCancellingErrorMessage = reason?.length > MAXIMUM_LENGTH ? REASON_TOO_LONG_MESSAGE : undefined;
```

trade-finance-manager-ui/server/controllers/case/cancellation/reason-for-cancelling.controller.ts
```diff
@@ -3,19 +3,51 @@ import { CustomExpressRequest } from '@ukef/dtfs2-common';
 import { PRIMARY_NAVIGATION_KEYS } from '../../../constants';
 import { asUserSession } from '../../../helpers/express-session';
 import { ReasonForCancellingViewModel } from '../../../types/view-models';
+import { validateReasonForCancelling } from './validation/validate-reason-for-cancelling';
+
+export type GetReasonForCancellingRequest = CustomExpressRequest<{ params: { _id: string } }>;
+export type PostReasonForCancellingRequest = CustomExpressRequest<{ params: { _id: string }; reqBody: { 'reason-for-cancelling': string | undefined } }>;
```
abhi-markan, line 9:
```suggestion
export type PostReasonForCancellingRequest = CustomExpressRequest<{ params: { _id: string }; reqBody: { 'reason': string} }>;
```

I like the custom Express request approach for an effective payload validation, but I think the body parameters should not allow `undefined` and can we keep them to pascal case please.

trade-finance-manager-ui/server/controllers/case/cancellation/reason-for-cancelling.controller.ts
```diff
@@ -3,19 +3,51 @@ import { CustomExpressRequest } from '@ukef/dtfs2-common';
...
-export const getReasonForCancelling = (req: CustomExpressRequest<{ params: { dealId: string } }>, res: Response) => {
-  const { dealId } = req.params;
+export const getReasonForCancelling = (req: GetReasonForCancellingRequest, res: Response) => {
+  const { _id } = req.params;
   const { user } = asUserSession(req.session);
 
   const reasonForCancellingViewModel: ReasonForCancellingViewModel = {
     activePrimaryNavigation: PRIMARY_NAVIGATION_KEYS.ALL_DEALS,
     user,
     ukefDealId: '0040613574', // TODO: DTFS2-7350 get values from database
```
abhi-markan, line 21: Appreciate not in your PR, buw where did this number come from and why is it set to this deal number please.

trade-finance-manager-ui/server/controllers/case/cancellation/reason-for-cancelling.controller.ts
```diff
@@ -3,19 +3,51 @@ import { CustomExpressRequest } from '@ukef/dtfs2-common';
...
-    dealId, // TODO: DTFS2-7350 get values from database
+    dealId: _id,
   };
   return res.render('case/cancellation/reason-for-cancelling.njk', reasonForCancellingViewModel);
 };
+
+/**
+ * controller to update the reason for cancelling
```
abhi-markan, line 28: DocString missing @params

trade-finance-manager-ui/server/controllers/case/cancellation/reason-for-cancelling.controller.ts
```diff
@@ -3,19 +3,51 @@ import { CustomExpressRequest } from '@ukef/dtfs2-common';
...
+ */
+export const postReasonForCancelling = (req: PostReasonForCancellingRequest, res: Response) => {
+  const { _id } = req.params;
+  const { 'reason-for-cancelling': reason } = req.body;
```
abhi-markan, line 32:
```suggestion
  const { 'reason': reason } = req.body;
```

trade-finance-manager-ui/server/controllers/case/cancellation/reason-for-cancelling.controller.ts
```diff
@@ -3,19 +3,51 @@ import { CustomExpressRequest } from '@ukef/dtfs2-common';
...
+  const { user } = asUserSession(req.session);
+
+  const validationErrors = validateReasonForCancelling(reason);
+
+  const formHasErrors = validationErrors.errorSummary.length !== 0;
+
+  if (formHasErrors) {
+    const reasonForCancellingViewModel: ReasonForCancellingViewModel = {
+      activePrimaryNavigation: PRIMARY_NAVIGATION_KEYS.ALL_DEALS,
+      user,
```
abhi-markan, line 42: Same as above.

trade-finance-manager-ui/server/controllers/case/cancellation/reason-for-cancelling.controller.ts
```diff
@@ -3,19 +3,51 @@ import { CustomExpressRequest } from '@ukef/dtfs2-common';
...
+
+/**
+ * controller to update the reason for cancelling
+ */
+export const postReasonForCancelling = (req: PostReasonForCancellingRequest, res: Response) => {
+  const { _id } = req.params;
+  const { 'reason-for-cancelling': reason } = req.body;
+  const { user } = asUserSession(req.session);
+
+  const validationErrors = validateReasonForCancelling(reason);
```
abhi-markan, line 35: Does this function also ensure user session is from `PIM` only?

abhi-markan (changes requested):

trade-finance-manager-ui/server/controllers/case/cancellation/reason-for-cancelling.controller.ts
ttbarnes: Good shout @abhi-markan 

I think in all instances it will always be sent a string - if an empty form is submitted it will be an empty string

trade-finance-manager-ui/server/controllers/case/cancellation/validation/validate-reason-for-cancelling.test.ts
```diff
@@ -0,0 +1,46 @@
...
+    // Arrange
+    const reason = new Array(1202).join('x');
+
+    // Act
+    const result = validateReasonForCancelling(reason);
+
+    // Assert
+    const expected: ReasonForCancellingErrorsViewModel = {
+      errorSummary: [{ text: 'Reason for cancelling must be 1200 characters or less', href: 'reason-for-cancelling' }],
+      reasonForCancellingErrorMessage: 'Reason for cancelling must be 1200 characters or less',
```
ttbarnes, line 42: Can we use the `MAXIMUM_LENGTH` constant for this instead of hard coding?

trade-finance-manager-ui/server/controllers/case/cancellation/reason-for-cancelling.controller.ts
MarRobSoftwire: This is stub data with the TODO to get this from the database in a future ticket.

trade-finance-manager-ui/server/controllers/case/cancellation/validation/validate-reason-for-cancelling.ts
```diff
@@ -0,0 +1,19 @@
...
+const REASON_TOO_LONG_MESSAGE = `Reason for cancelling must be ${MAX_CHARACTER_COUNT} characters or less`;
+
+/**
+ * @param reason The reason for cancelling
+ * @returns a reason for cancelling errors view model
+ */
+export const validateReasonForCancelling = (reason: string): ReasonForCancellingErrorsViewModel => {
+  const errorSummary: ErrorSummaryViewModel[] = [];
+
+  const reasonForCancellingErrorMessage = reason.length > MAX_CHARACTER_COUNT ? REASON_TOO_LONG_MESSAGE : undefined;
```
ttbarnes, line 13:
```suggestion
  const reasonForCancellingErrorMessage = reason.length > MAX_CHARACTER_COUNT && REASON_TOO_LONG_MESSAGE;
```

MarRobSoftwire: This would pass `false` into the `ReasonForCancellingErrorsViewModel`, I'm not sure how it would be handled by the nunjucks template but thats not the current pattern

ttbarnes: Feels very strange that we need to explicitly return an undefined value. Is this approach used in other places?

MarRobSoftwire: This just allows me to have one return at the end
```
  return { errorSummary, reasonForCancellingErrorMessage };
```

rather than 

```
export const validateReasonForCancelling = (reason: string): ReasonForCancellingErrorsViewModel => {
  if (reason.length > MAX_CHARACTER_COUNT) {
    return {
      errorSummary: [{{ text: REASON_TOO_LONG_MESSAGE, href: 'reason-for-cancelling' }}],
      reasonForCancellingErrorMessage: REASON_TOO_LONG_MESSAGE,
    }
  }

  return { errorSummary: [] };
};
```

MarRobSoftwire: It's based on the other typescript validation files (e.g. `trade-finance-manager-ui\server\controllers\utilisation-reports\find-reports-by-year\search-input-validator.ts`) which allows this validation to be extending on if there was more than one input to be validated.

ttbarnes: Thanks @MarRobSoftwire I think there are some other ways we could do it, but happy with this existing approach.

trade-finance-manager-ui/server/controllers/case/cancellation/validation/validate-reason-for-cancelling.test.ts
```diff
@@ -0,0 +1,43 @@
+import { MAX_CHARACTER_COUNT } from '@ukef/dtfs2-common';
+import { ReasonForCancellingErrorsViewModel } from '../../../../types/view-models';
+import { validateReasonForCancelling } from './validate-reason-for-cancelling';
+
+describe('validateReasonForCancelling', () => {
+  const errorFreeReasons = [
```
ttbarnes, line 6: `validAnswers`?

trade-finance-manager-ui/server/controllers/case/cancellation/validation/validate-reason-for-cancelling.test.ts
```diff
@@ -0,0 +1,43 @@
...
+    {
+      description: 'an empty string',
+      reason: '',
+    },
+    {
+      description: `${MAX_CHARACTER_COUNT} characters`,
+      reason: new Array(Number(MAX_CHARACTER_COUNT) + 1).join('x'),
```
ttbarnes, line 13:
```suggestion
      reason: 'x'.repeat(MAX_CHARACTER_COUNT),
```

ttbarnes (changes requested):
2x minor comments/suggestions otherwise LGTM 👍

trade-finance-manager-ui/server/controllers/case/cancellation/validation/validate-reason-for-cancelling.test.ts
MarRobSoftwire: Ooh thats much cleaner!

ttbarnes (approved):

abhi-markan (approved):
LGTM 👍🏼

PR #3497, opened 2024-09-13
feat(FN-2309): handle reconciled records with no payments

RLavender98:
## Introduction :pencil2:
If a fee record has no payments it can be reconciled but have no reconciled by user:
- If the fee record belongs to a manually reconciled report from before the payment reconciliation feature is enabled
- If the fee record is auto reconciled
In both cases we won't actually display any reconciled by data since that is only displayed on the payments tab, so we just need to handle this gracefully.

- Do not throw an error if a reconciled fee record with no payments has no reconciled by date or user

- Update the sql db seeder to not seed reconciliation data for the manually reconciled reports fee records


sw-mattw (dismissed):

dtfs-central-api/src/helpers/get-fee-record-payment-entity-group-reconciliation-data.ts
```diff
@@ -23,14 +23,17 @@ export const getFeeRecordPaymentEntityGroupReconciliationData = async (
 
   const reconciledFeeRecords = group.feeRecords.filter(({ status }) => status === FEE_RECORD_STATUS.RECONCILED);
 
-  if (reconciledFeeRecords.length === 0) {
+  if (reconciledFeeRecords.length !== group.feeRecords.length) {
```
ttbarnes, line 26: Can we add a comment/documentation here explaining why we need to do this?

sw-mattw: Just picked up this PR, I believe this is to ensure we only return the reconciliation data if all fee records in the group are reconciled. I've pulled this condition out into a self-describing constant to try and avoid the comment - is this ok? Thanks

sw-mattw: This also means we can simplify the filtering to be an `every` check (see diff), simplifying the function

ttbarnes (approved):

dtfs-central-api/src/helpers/get-fee-record-payment-entity-group-reconciliation-data.ts
ttbarnes: Nice!

PR #3498, opened 2024-09-13
feat(DTFS2-7370): add RBAC middleware to tfm-api

MarRobSoftwire:
## Introduction :pencil2:
TFM-API doesn't validate the users teams on any endpoints. The new cancellation endpoints should only be accessible to PIM users

- Add new `validateUserHasAtLeastOneAllowedRole` middleware to tfm-api
- Add test helper
- Add test user fluent builder to allow middleware to be tested

- Remove types from tfm-api that are duplicated in libs/common 
- convert test api set up to typescript


trade-finance-manager-api/src/types/express.d.ts
MarRobSoftwire, line 1: Taken from #3469

trade-finance-manager-api/api-tests/api-test-users.ts
```diff
@@ -0,0 +1,133 @@
...
+    withTeam: (team: TeamId) => {
+      users = users.filter((user) => user.teams.includes(team));
+      return fluentBuilder;
+    },
+    withoutTeam: (team: TeamId) => {
+      users = users.filter((user) => !user.teams.includes(team));
+      return fluentBuilder;
+    },
+  };
+};
```
josh-b-2000, line 35: Might be nice to make this more functional:
```typescript
const fluentBuilder = (users: TestUser[]) => ({
  all: () => users,
  one: () => users[0],
  withTeam: (team: TeamId) => {
    const filteredUsers = users.filter((user) => user.teams.includes(team));
    return fluentBuilder(filteredUsers);
  },
  withoutTeam: (team: TeamId) => {
    const filteredUsers = users.filter((user) => !user.teams.includes(team));
    return fluentBuilder(filteredUsers);
  },
});
```

trade-finance-manager-api/src/v1/deals/routes.js
```diff
@@ -78,7 +80,12 @@ dealsAuthRouter
 
 dealsAuthRouter
   .route('/deals/:dealId/amendments/:status?/:type?')
-  .get(validation.dealIdValidation, handleExpressValidatorResult, amendmentController.getAmendmentsByDealId);
+  .get(
+    validateUserHasAtLeastOneAllowedTeam([TEAM_IDS.PIM]),
+    validation.dealIdValidation,
```
MarRobSoftwire, line 85: TODO: move this to new endpoint & test it

trade-finance-manager-api/api-tests/api.ts
```diff
@@ -0,0 +1,158 @@
...
+export type TestAs = (user: TestUser) => {
+  post: TestRequestWithoutHeaders;
+  postEach: TestPostEach;
+  put: TestRequestWithoutHeaders;
+  putMultipartForm: TestPutMultipartForm;
+  get: TestGetWithoutHeaders;
+  remove: TestRequestWithoutHeaders;
+};
+
+type TestApi = {
```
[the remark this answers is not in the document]
MarRobSoftwire: Good call

trade-finance-manager-api/.eslintrc.js
```diff
@@ -35,12 +35,7 @@ const baseRules = {
   ],
   'lines-between-class-members': ['error', 'always', { exceptAfterSingleLine: true }],
   'no-unused-vars': ['error', { ignoreRestSiblings: true }],
-  'object-curly-newline': [
-    'error',
-    {
-      consistent: true,
-    },
-  ],
+  'object-curly-newline': 'off',
```
MarRobSoftwire, line 38: This was contradicting prettier (see https://stackoverflow.com/questions/59318013/eslint-object-curly-newline-conflicts-prettier-multi-line-rationale)

trade-finance-manager-api/api-tests/v1/utils/submitDeal.js
MarRobSoftwire, line 1: Note to self: why is this in a `utils` folder?

trade-finance-manager-api/api-tests/api-test-users.ts
```diff
@@ -0,0 +1,130 @@
...
+  await collection.insertOne(userToInsert);
+};
+
+/**
+ * Create tests user with an initial user token
+ * @param params
+ * @param params.user - user to create
+ * @param params.as - supertest as user
+ * @param params.initialuser - already created initial user with token
+ */
```
ttbarnes, line 89: `* Create a test user* with ...`

`* @param params.initialUser - Initially created user with token`

?

trade-finance-manager-api/api-tests/api.ts
```diff
@@ -0,0 +1,101 @@
...
+    'x-api-key': TFM_API_KEY,
+  };
+  if (token) {
+    headers.Authorization = token;
+  }
+
+  return headers;
+};
+
+export const createApi = (app: unknown): TestApi => ({
```
ttbarnes, line 25: Documentation?

trade-finance-manager-api/api-tests/api.ts
```diff
@@ -0,0 +1,101 @@
...
+          const apiRequest = request(app).put(url).set(getHeaders(token));
+
+          if (files.length) {
+            await Promise.all(files.map((file) => apiRequest.attach(file.fieldname, file.filepath)));
+          }
+
+          await Promise.all(
+            Object.entries(data).map(([fieldname, value]) => {
+              // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
+              return apiRequest.field(fieldname, value as any);
```
ttbarnes, line 63: Are we able to remove the eslint disable and `any`?

trade-finance-manager-api/api-tests/api.ts
```diff
@@ -0,0 +1,101 @@
...
+            }),
+          );
+
+          return apiRequest;
+        },
+      }),
+
+      get: async (url: string) => request(app).get(url).set(getHeaders(token)),
```
ttbarnes, line 71: Might be able to remove the `async` from such functions in this file(?)

trade-finance-manager-api/api-tests/common-tests/with-team-authorisation.api-tests.ts
```diff
@@ -0,0 +1,40 @@
...
+  expect(status).toBe(HttpStatusCode.Forbidden);
+  expect(body).toStrictEqual({
+    success: false,
+    msg: "You don't have access to this page",
+  });
+};
+
+const allTeams = Object.values(TEAM_IDS);
+
+export const withTeamAuthorisationTests = ({ allowedTeams, getUserWithTeam, makeRequestAsUser, successStatusCode }: WithTeamAuthorisationTests) => {
```
ttbarnes, line 24: Why are we exporting this?

trade-finance-manager-api/api-tests/database-helper.ts
```diff
@@ -0,0 +1,11 @@
+import { MongoDbCollectionName } from '@ukef/dtfs2-common';
+import { mongoDbClient } from '../src/drivers/db-client';
+
+export const wipeCollection = async (collectionNames: MongoDbCollectionName[]) => {
```
ttbarnes, line 4: documentation

trade-finance-manager-api/api-tests/types/test-api.ts
```diff
@@ -0,0 +1,59 @@
...
+};
+
+type TestGetWithHeaders = (
+  url: string,
+  {
+    headers,
+    query,
+  }?: {
+    headers: IncomingHttpHeaders | undefined;
+    query: AnyObject | undefined;
```
ttbarnes, line 18:
```suggestion
    headers?: IncomingHttpHeaders;
    query?: AnyObject
```

trade-finance-manager-api/api-tests/types/test-user.ts
```diff
@@ -0,0 +1,16 @@
...
+  _id?: string;
+  username: string;
+  password: string;
+  email: string;
+  teams: TeamId[];
+  timezone: string;
+  firstName: string;
+  lastName: string;
+  token: string;
+};
```
ttbarnes, line 16: Do we already have a `User` type that could be used here instead of a new `TestUser` type?

If there's an existing `User` type, but the tests require additional fields, perhaps we could extend the test type instead

trade-finance-manager-api/api-tests/v1/deal-cancellation/deal-cancellation.api-test.ts
```diff
@@ -76,36 +85,44 @@ describe('/v1/deals/:id/cancellation', () => {
         jest.resetAllMocks();
       });
 
+      withTeamAuthorisationTests({
```
ttbarnes, line 88: Do we have other team specific routes? I believe there are some for underwriters at least.

If so, seems `withTeamAuthorisationTests` is a new pattern. If we want to use this new pattern we should either update the other tests or use the original pattern, rather than introducing a different way of testing the same thing.

Ignore me if i'm wrong or missing something 😊

trade-finance-manager-api/src/v1/__mocks__/mock-users.js
```diff
@@ -1,10 +1,12 @@
+const { TEAM_IDS } = require('@ukef/dtfs2-common');
+
 const MOCK_USERS = [
   {
     _id: '6051d94564494924d38ce67c',
     username: 'BUSINESS_SUPPORT_USER_1',
     password: 'AbC!2345',
     email: 'test@testing.com',
-    teams: ['BUSINESS_SUPPORT'],
+    teams: [TEAM_IDS.BUSINESS_SUPPORT],
```
ttbarnes, line 9: Can we destructure `TEAM_IDS`?

trade-finance-manager-api/src/v1/controllers/user/passport.js
```diff
@@ -14,15 +14,19 @@ const options = {
...
+  teams: user.teams,
   lastLogin: user.lastLogin,
-  firstname: user.firstname,
-  surname: user.surname,
+  firstName: user.firstName,
+  lastName: user.lastName,
   email: user.email,
   timezone: user.timezone,
-  'user-status': user['user-status'],
+  status: user.status,
```
ttbarnes, line 29: Why are these values changing? `roles => teams`, `firstname => firstName` etc

trade-finance-manager-api/src/v1/deals/routes.js
ttbarnes: Do we have a ticket for this

trade-finance-manager-api/src/v1/middleware/validate-user-is-in-at-least-one-allowed-team.test.ts
```diff
@@ -0,0 +1,83 @@
+import { createMocks } from 'node-mocks-http';
+import { HttpStatusCode } from 'axios';
+import { TEAM_IDS, TeamId } from '@ukef/dtfs2-common';
+import { validateUserHasAtLeastOneAllowedTeam } from './validate-user-is-in-at-least-one-allowed-team';
+import { aTfmSessionUser } from '../../../test-helpers';
+
+describe('validateUserHasAtLeastOneAllowedTeam returns middleware that', () => {
```
ttbarnes, line 7: Can we simplify this please

```ts
describe('validateUserHasAtLeastOneAllowedTeam', () => {
```

trade-finance-manager-api/src/v1/middleware/validate-user-is-in-at-least-one-allowed-team.test.ts
```diff
@@ -0,0 +1,83 @@
...
+  beforeEach(() => {
+    jest.resetAllMocks();
+  });
+
+  describe('when the user is not in one of the teams', () => {
+    const userTeams: TeamId[] = [TEAM_IDS.PDC_READ];
+    const endpointTeams: TeamId[] = [TEAM_IDS.PIM, TEAM_IDS.RISK_MANAGERS];
```
ttbarnes, line 14: `allowedTeams`? Same with the test below

trade-finance-manager-api/src/v1/middleware/validate-user-is-in-at-least-one-allowed-team.ts
```diff
@@ -0,0 +1,23 @@
...
+ */
+export const validateUserHasAtLeastOneAllowedTeam =
+  (allowedTeams: TeamId[]): RequestHandler =>
+  (req, res, next) => {
+    const { user } = req;
+
+    const userHasAtLeastOneAllowedTeam = allowedTeams.some((allowedTeam) => user.teams.includes(allowedTeam));
+
+    if (!userHasAtLeastOneAllowedTeam) {
+      console.error('Unauthorised access for %s.', user);
```
ttbarnes, line 17: Can we have a more descriptive log here that mentions it's a team check?

ttbarnes (changes requested):
Nice work

trade-finance-manager-api/src/v1/deals/routes.js
MarRobSoftwire: Have done this as part of the PR now :)

trade-finance-manager-api/src/v1/middleware/validate-user-is-in-at-least-one-allowed-team.ts
```diff
@@ -0,0 +1,23 @@
...
+ */
+export const validateUserHasAtLeastOneAllowedTeam =
+  (allowedTeams: TeamId[]): RequestHandler =>
+  (req, res, next) => {
+    const { user } = req;
+
+    const userHasAtLeastOneAllowedTeam = allowedTeams.some((allowedTeam) => user.teams.includes(allowedTeam));
+
+    if (!userHasAtLeastOneAllowedTeam) {
+      console.error('Unauthorised access, expected user team to be one of: %s, but user was: %s.', allowedTeams, user);
```
ttbarnes, line 17: Nice 😍

trade-finance-manager-api/src/v1/controllers/user/passport.js
ttbarnes: Wowza. Nice spot

trade-finance-manager-api/api-tests/types/test-api.ts
ttbarnes: Thanks @MarRobSoftwire 

Not for this PR but it would be nice to some day look into improving these, seeing a lot of instances of `| undefined`, I don't understand why we should explicitly need to define something as "or undefined" 😢

trade-finance-manager-api/api-tests/api.ts
ttbarnes: Resolving this as it seems these functions are purely being moved/migrated to TS.

trade-finance-manager-api/api-tests/types/test-user.ts
ttbarnes: Not sure what you mean by hacky - do we have an existing `User` type? I can't seem to find 

Just want to avoid having duplicate or almost identical types when they could have a common interface

trade-finance-manager-api/api-tests/common-tests/with-team-authorisation.api-tests.ts
ttbarnes: Makes sense, nice one - could we just add some documentation please

trade-finance-manager-api/api-tests/api-test-users.ts
ttbarnes: Thanks @MarRobSoftwire makes sense to me just 2x small typos/niggles please 😊 

- `Create tests user` should be `Create test user` or `Creates a test user`
- `params.initialuser` should be `params.initialUser`

trade-finance-manager-api/api-tests/api.ts
ttbarnes: Thanks @MarRobSoftwire good finds - I would vote for no.2 and definitely not no.1 🙈 

I know it bloats the code, but it's better than disabling eslint rules, what do you think?

Conscious that we already have a lot of eslint disabling (e.g search for `@typescript-eslint/no-unsafe-argument` and click through each file), really keen to _not_ add anymore such disables

trade-finance-manager-api/api-tests/types/test-user.ts
MarRobSoftwire: Can you see the latest version of the file? I think there is a case for separate types for those stored in the database (database models) and when they've been serialised (Data Transfer Objects - i think).

Happy to somewhat commonise them here if you think thats best

AlexBramhill: I've hit this type issue on SSO (I'll bring it up tomorrow in standup as we need to decide an agreed pattern).

We should have the following types for any given thing ideally:
- Database model
- Internal service use type (can be the same as DTO)
- Data transfer object (as it leaves our service)

Then we'll need to have an incoming DTO for the service recieving the object (basically a serialised DTO object), a mapper that maps the serialised DTO either back to the DTO object or return a new, leaner object (the leaner object is slightly pointless imo).

As we pass through multiple services, we might need more than one DTO for an object (for instance, UI layer should only recieve things that the UI cares about, but the API layer might care about more things depending on responsibility of services, a good example of this would be `facilities` or `deals` etc -- we wouldn't want to return things to the UI layer, but we would want to be able to pass them around our 
API services, so the need for multiple DTOs could be present).

Commonisation is interesting, as the commonisation appears to have occured prior to libs common, and may already be out of sync (they are on SSO ticket).

I've seen a libs common and a TFM API `TfmUser` model for instance.

Regardless, my understanding so far is that we have a `TfmUser` db model -- and we should, on all repo calls, be transforming that type into another type, removing information like salt and hash etc (ideally we would do this in the mongo queries themselves). At the minute, my understanding is this is pretty much `TfmSessionUser` object for all intents and purposes.

ttbarnes: Good reply @AlexBramhill 

With my current knowledge/understanding, I don't see why a model should be different throughout various areas. IMO they should ideally be aligned, avoiding mapping, multiple types, potential confusion etc.

Appreciate there are scenarios where they _should_ be different - e.g for a user, probably only the API/DB should care about e.g salt/hash fields and the UI shouldn't consume or know about such fields. So in this scenario there could be a common user type and the API/DB extends the type.

This feels like a separate thing to tackle outside of this PR and sounds like it doesn't affect just users.

Shall we setup a call to talk about this and agree on a pattern like @AlexBramhill said?

trade-finance-manager-api/api-tests/api.ts
```diff
@@ -0,0 +1,110 @@
...
+        to: async (url: string) => {
+          const apiRequest = request(app).put(url).set(getHeaders(token));
+
+          if (files.length) {
+            await Promise.all(files.map((file) => apiRequest.attach(file.fieldname, file.filepath)));
+          }
+
+          await Promise.all(
+            Object.entries(data).map(([fieldname, value]) => {
+              // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
```
ttbarnes, line 71: Same as below RE removing eslint disable

MarRobSoftwire: Good spot - wasn't doing anything anyway :)

trade-finance-manager-api/api-tests/types/test-user.ts
MarRobSoftwire: Would be great to have a call about this sometime :) 

I definitely agree that we should be avoiding mapping these objects around too much, but unfortunately theres a lot of that already.
The issue is particularly visible when you look at the objects being consumed by the ui (e.g. I created a front end facility type `gef-ui\server\types\facility.ts` because the final object is so different to `libs\common\src\types\mongo-db-models\facility.ts`. The UnixTimestampStrings are mapped to IsoDateTimeStamp somewhere, ObjectIds are serialised to strings)

MarRobSoftwire: If we did have the ideal world where things weren't mapped, we could use a [Jsonify](https://effectivetypescript.com/2020/04/09/jsonify/) type helper to get around the serialisation issue

ttbarnes (approved):

trade-finance-manager-api/api-tests/types/test-user.ts
MarRobSoftwire: Resolving but will chat about the plan for this going forward at some point :)

PR #3526, opened 2024-09-17
feat(FN-2311): add initial payment details filter UI components

sw-mattw:
## Introduction :pencil2:
Adds the initial UI components for the payment details tab filter. This includes a collapsible panel when JS is enabled, which is always shown when JS is disabled.

Note that the error message UI and the view model will be updated in a subsequent PR.

- New UI template macros for the payment details tab filters, including component tests for these. Makes use of the [MOJ filter component](https://design-patterns.service.justice.gov.uk/components/filter/).
- New JS script to toggle the collapsible filter panel.
- JQuery types added to TFM UI
- New docs on adding and updating scripts and integrity hashes

Note that the screenshots are missing the payment currency radio inputs, this is as these are yet to be provided in the view model (this will be added in a subsequent PR for this ticket).

FN-2311_pr-1_panel-hidden

FN-2311_pr-1_panel-shown

Once the currency options are passed in through the view model, the filter component will look as follows:
FN-2311_pr-1_panel-with-currency

- Note that running the `build` script within TFM UI has recompiled all files, leading so some diffs on other files


trade-finance-manager-ui/scripts/moj-filter-hide.ts
```diff
@@ -0,0 +1,52 @@
...
+
+if (cachedShownStatus) {
+  setFilterVisibility(cachedShownStatus);
+}
+
+// Initially display filters, but hide them if they were previously hidden
+const startHidden = cachedShownStatus === 'HIDDEN';
+
+// eslint-disable-next-line @typescript-eslint/no-unsafe-call, no-new, @typescript-eslint/no-unsafe-member-access
+new window.MOJFrontend.FilterToggleButton({
```
sw-mattw, line 32: The new keyword is necessary here because it creates a new instance of the FilterToggleButton class, which sets up event listeners and modifies the DOM to enable the filter toggle behavior. Disabled in line with `portal/scripts/moj-filters.js`.

doc/adding-and-updating-scripts.md
```diff
@@ -0,0 +1,60 @@
...
+2. Copy this output and update the integrity attribute value in the
+corresponding script tag in the relevant `njk` template.
+    - For example:
+      - If the script tag is currently: `<script
+      src="/assets/js/your-script.js" type="text/javascript"
+      integrity="sha512-3UQ..." crossorigin="anonymous"></script>`
+      - And the output from the above command is `sha512-QLv...`
+      - Our updated script tag would be: `<script
+      src="/assets/js/your-script.js" type="text/javascript"
+      integrity="sha512-QLv..." crossorigin="anonymous"></script>`
```
ttbarnes, line 60: Do we need this? We have documentation in the README - maybe could be updated instead? I am not sure what exists or is the plan with this `doc` directory

trade-finance-manager-ui/templates/index.njk
```diff
@@ -51,11 +51,12 @@
...
     <script src="/assets/js/disableFormSubmitOnSubmission.js" type="text/javascript" integrity="sha512-tCINlYTcYERel2bdysTMl5KaemDdEyog5JSS3hYQqHS6ZIyT4ATLu4oTPerbZ+vAOn4AAzDEsRBSQVacGhoieQ==" crossorigin="anonymous"></script>
     <script src="/assets/js/enableSelectAllTableCheckbox.js" type="text/javascript" integrity="sha512-ER/5raNAJX5VTcyrRBjVExC/f39WaVY4r8k1CyGqJUad95QZ0A1H6zHtsFpdf0JDBLV0acXzSA1B2mkKC8g5Tg==" crossorigin="anonymous"></script>
-  <script src="/assets/js/enableFindReportsByYearDropdown.js" integrity="sha512-QLvf2zIp3WAaUOyTIf4SlZ9JLBWVw1CCwwY8Ozzc3+mTsFv4WClwZ9ZYNdNwf9nSFNd2Y4WNr1GL/Bc6ABo1Fw==" type="text/javascript" crossorigin="anonymous"></script>
+    <script src="/assets/js/enableFindReportsByYearDropdown.js" integrity="sha512-QLvf2zIp3WAaUOyTIf4SlZ9JLBWVw1CCwwY8Ozzc3+mTsFv4WClwZ9ZYNdNwf9nSFNd2Y4WNr1GL/Bc6ABo1Fw==" type="text/javascript" crossorigin="anonymous"></script>
+    <script src="/assets/js/mojFilterHide.js" type="text/javascript" integrity="sha512-SopYqjSHGoALKGfzoUMnm3ovAiTAQP0zz9UQPgo8N+IGD9XPscbhKGWg0fQ5txgzrfm2kURyK9D5iKalsMLwMA==" crossorigin="anonymous"></script>
```
ttbarnes, line 60: Are we able to include this only in the page that requires this script? Otherwise every page will be loading this resource.

Appreciate the other scripts aren't part of this PR, but we should be doing the same with e.g the following:

```
enableSelectAllTableCheckbox
enableFindReportsByYearDropdown
```

FYI @Zainzzkk

trade-finance-manager-ui/scripts/moj-filter-hide.ts
```diff
@@ -0,0 +1,52 @@
...
+const toggleButtonContainer = document.querySelector(toggleButtonContainerSelector)!;
+
+const params = (toggleButtonContainer as HTMLElement).dataset;
+const FILTER_VISIBILITY_SESSION_KEY = `${params.filteredTableId}-filters-visibility-state`;
+
+const setFilterVisibility = (visibility: FilterVisibility) => {
+  sessionStorage.setItem(FILTER_VISIBILITY_SESSION_KEY, visibility);
+};
+
+const cachedShownStatus = sessionStorage.getItem(FILTER_VISIBILITY_SESSION_KEY) as FilterVisibility | null;
```
ttbarnes, line 22: Are we able to make something optional rather than `| null`?

trade-finance-manager-ui/scripts/moj-filter-hide.ts
```diff
@@ -0,0 +1,52 @@
...
+
+if (cachedShownStatus) {
+  setFilterVisibility(cachedShownStatus);
+}
+
+// Initially display filters, but hide them if they were previously hidden
+const startHidden = cachedShownStatus === 'HIDDEN';
+
+// eslint-disable-next-line @typescript-eslint/no-unsafe-call, no-new, @typescript-eslint/no-unsafe-member-access
```
ttbarnes, line 31: Can we remove such eslint rules please

Maybe it'd be best to have this as a JS file instead, like the other client side scripts

trade-finance-manager-ui/scripts/moj-filter-hide.ts
```diff
@@ -0,0 +1,52 @@
...
+  },
+});
+
+const toggleButton = document.querySelector(`${toggleButtonContainerSelector} > button`) as HTMLButtonElement;
+
+toggleButton.addEventListener('click', () => {
+  const currentStatus = sessionStorage.getItem(FILTER_VISIBILITY_SESSION_KEY) as FilterVisibility;
+
+  setFilterVisibility(currentStatus === 'HIDDEN' ? 'SHOWN' : 'HIDDEN');
+});
```
ttbarnes, line 52: Can we add some documentation for this entire file please

trade-finance-manager-ui/scripts/moj-filter-hide.ts
ChristianMcc: @abhi-markan I'd quite like to keep new files to typescript if possible for maintainability, is this one of the places where we might want to look at disabling some of the eslint rules on files within the script folders you mentioned before rather than having the eslint-disables in line?

trade-finance-manager-ui/templates/utilisation-reports/utilisation-report-reconciliation-for-report.njk
```diff
@@ -158,18 +160,39 @@
...
-    {{ paymentDetailsTable.render({
-      reportId: reportId,
-      paymentDetails: paymentDetails,
-      userCanEdit: userCanEdit
-    }) }}
+      {{ paymentDetailsFiltersPanel.render({
+        filters: filters,
+        selectedFilters: selectedFilters,
+        keyword: keyword,
+        csrfToken: csrfToken
```
ChristianMcc, line 169: where does this get rendered to be passed back? Is it part of the MoJ implementation or should I be expecting a csrfToken to be passed back in a form somewhere? (Sorry if I'm missing something obvious!)

ttbarnes (commented):
E2E tests?

trade-finance-manager-ui/scripts/moj-filter-hide.ts
sw-mattw: I'd also prefer TS

Please see [the comment below](https://github.com/UK-Export-Finance/dtfs2/pull/3526/files#r1763142121) - disabling these eslint rules for this line only would also be required in JS and is what the current JS implementation in portal (`portal/scripts/moj-filters.js`) also has to do

trade-finance-manager-ui/scripts/moj-filter-hide.ts
sw-mattw: `getItem` returns `string | null` - I think it's good to apply the type casting straight away and also keep the consistency with the underlying API. Not sure how I feel about changing the type casting to `FilterVisibility | undefined`.

trade-finance-manager-ui/templates/utilisation-reports/utilisation-report-reconciliation-for-report.njk
sw-mattw: oops I forgot to update these, sorted it now thanks - removed the CSRF token as it isn't needed/used (at least for now)

trade-finance-manager-ui/scripts/moj-filter-hide.ts
sw-mattw: Refactored this whole file now thanks - should be a lot more readable and has docs added

trade-finance-manager-ui/templates/index.njk
sw-mattw: Moved these 'single-use' scripts into their respective templates (tried to work this out from the PRs they were merged in with) - had to pull the mojFrontend.js depdency up to the top of the body though as some of these scripts are dependent on this being included first

doc/adding-and-updating-scripts.md
sw-mattw: Good point, missed that... removed

sw-mattw:
Added an initial E2E tests, planning to address testing the actual filtering in a subsequent PR once both the central API and UI server implementations are completed.

e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/filter-payment-details.spec.js
```diff
@@ -0,0 +1,76 @@
...
+
+        if (isFirstFeeRecordRow) {
+          pages.utilisationReportPage.paymentDetailsTab.paymentDetailsTable
+            .paymentCurrencyAndAmount(paymentId, feeRecordId)
+            .should('contain', `${currency} ${amount}`);
+
+          pages.utilisationReportPage.paymentDetailsTab.paymentDetailsTable.paymentReference(paymentId, feeRecordId).should('contain', reference);
+        }
+
+        pages.utilisationReportPage.paymentDetailsTab.paymentDetailsTable.facilityId(paymentId, feeRecordId).should('contain', facilityId);
```
ttbarnes, line 72: For the above assertions can we please use `cy.assertText` (strict equals check) instead of `.contains`?

e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/filter-payment-details.spec.js
```diff
@@ -0,0 +1,76 @@
...
+
+  beforeEach(() => {
+    cy.task(NODE_TASKS.REMOVE_ALL_UTILISATION_REPORTS_FROM_DB);
+
+    cy.task(NODE_TASKS.INSERT_UTILISATION_REPORTS_INTO_DB, [utilisationReport]);
+
+    pages.landingPage.visit();
+    cy.login(USERS.PDC_RECONCILE);
+
+    cy.visit(`/utilisation-reports/${reportId}`);
```
ttbarnes, line 33: Not a blocker as can see a few instances of this - we should mimic an actual user journey rather than navigating to a URL directly

ttbarnes (changes requested):

e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/filter-payment-details.spec.js
sw-mattw: I see your point about wanting to truly test the end-to-end journey, I'm not sure how practical this is though (see below):

Instead of `cy.visit` we could try to do something like:

```
pages.utilisationReportsSummaryPage.visit();
pages.utilisationReportsSummaryPage.reportLink(BANK_ID, SUBMISSION_MONTH).click();
```

We have `BANK_ID`, but `SUBMISSION_MONTH` is a bit trickier to construct/deal with - [this is done in `e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/reconcile-fee-records.spec.js`](https://github.com/UK-Export-Finance/dtfs2/blob/9159cd13c17e74be99d3ea03b159eb9677e137b4/e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/reconcile-fee-records.spec.js#L55) - and requires a few calls to different helpers.

Is this the sort of approach you were thinking? I can add a similar implementation here, it just seems a bit more overhead/adding more complexity to the tests

ttbarnes: Thanks @sw-mattw I know it adds some complexity but `.visit()` doesn't replicate "typical / happy path" kind of user experience, we could have something like `cy.goToReport(bankId)` for example and use throughout various tests, so it's all handled in one place and reduces such lines in individual tests.

I think this should be done in one PR and for many other things; Happy to resolve since there are other instances of this unrelated to this PR and don't want them to have different approaches.

trade-finance-manager-ui/scripts/moj-filter-hide.ts
```diff
@@ -0,0 +1,91 @@
...
+};
+
+/**
+ * Initialises the MOJFrontend FilterToggleButton with specified configuration.
+ * This function sets up the filter panel toggle functionality for the UI.
+ */
+const initialiseFilterToggleButton = (): void => {
+  const startHidden = initialiseFilterVisibility();
+
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-call, no-new, @typescript-eslint/no-unsafe-member-access
```
ttbarnes, line 52: Are we able to remove this eslint disabling?

trade-finance-manager-ui/scripts/moj-filter-hide.ts
```diff
@@ -0,0 +1,91 @@
+import $ from 'jquery';
```
ttbarnes, line 1: Last comment - is the jquery import required purely because of the MOJ package?

sw-mattw: Yep, the [MOJ component expects elements with jQuery functions](https://github.com/ministryofjustice/moj-frontend/blob/1a02b936d8ff76279200fcc6185df2dce87bf5da/src/moj/components/filter-toggle-button/filter-toggle-button.js#L66) (e.g. `addClass` and `removeClass`).

I tried to achieve this with `document.querySelector` but this returns a standard DOM Element without these methods, so I've used jQuery to provide them rather than adding a wrapper around Element.

I've just added a comment to the script to summarise this

ttbarnes (approved):

e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/filter-payment-details.spec.js
sw-mattw: @ttbarnes I've added a ticket to address this thanks - FN-3483

PR #3540, opened 2024-09-18
feat(DTFS2-7354): add get cancellation endpoint in central api

MarRobSoftwire:
## Introduction :pencil2:
dtfs-central-api should get the cancellation object from the tfm-deals collection

Add cancellation endpoint, using cancellation repo.
Add api and unit tests


libs/common/src/types/mongo-db-models/tfm-deals.ts
```diff
@@ -12,7 +12,7 @@ import { AuditDatabaseRecord } from '../audit-database-record';
 export type TfmDeal = {
   _id: ObjectId;
   dealSnapshot: Deal;
-  tfm: { cancellation: TfmDealCancellation };
+  tfm?: { cancellation: TfmDealCancellation };
```
ttbarnes, line 15: What's the reason for this please, a TFM deal should always have a `tfm` object?

libs/common/src/errors/index.ts
```diff
@@ -12,3 +12,4 @@ export * from './invalid-facility-id.error';
 export * from './deal-version.error';
 export * from './invalid-parameter-error';
 export * from './invalid-payload-error';
+export * from './cancellation-not-found.error';
```
ttbarnes, line 15: `deal-cancellation-not-found.error`?

Incase we add "cancellation" features to something else

ttbarnes (changes requested):

libs/common/src/types/mongo-db-models/tfm-deals.ts
MarRobSoftwire: I thought so too but when running an api-test:
In `dtfs-central-api\api-tests\v1\tfm\deal-cancellation\tfm-deal-cancellation-get.api-test.ts`, the test with title `'should return 404 if the deal does not have an existing cancellation'` fails if the `tfm-deal-cancellation.repo.ts` doesn't use optional chaining on `!matchingDeal.tfm?.cancellation` (line 34).

It looks like the `deal.tfm` object isnt initialised when the submit request is made to `'/v1/tfm/deals/submit'`, but will be initialised whenever anything is first done on the tfm side.

I'll double check this is the case, but i think even if it will never happen in practice, the way the endpoints are configured mean its possible in theory

ttbarnes: Thanks @MarRobSoftwire maybe that unit test is missing something compared to other existing tests?

Concern is that it's only since desired with the changes (new test) in this PR

MarRobSoftwire: Yeah I wonder if this is the case - when done through portal tfm is initalised with 
``` "tfm": {
    "activities": [],
    "dateReceived": "23-09-2024",
    "dateReceivedTimestamp": 1727085149,
    "exporterCreditRating": "Acceptable (B+)",
    "lastUpdated": 1727085149571,
    "lossGivenDefault": 50,
    "parties": {},
    "probabilityOfDefault": 12,
    "product": "GEF",
    "stage": "Confirmed"
  }
 ```
I'll chase up where thats done

MarRobSoftwire: Yeah it looks like this is done in `trade-finance-manager-api\src\v1\controllers\deal.submit.controller.js` in a separate call `addTfmDealData`

ttbarnes: Nice one @MarRobSoftwire 🎉

ttbarnes (approved):

PR #3542, opened 2024-09-18
feat(DTFS2-7353): add get cancellation endpoint to tfm api

MarRobSoftwire:
## Introduction :pencil2:
tfm-api should get the cancellation object from dtfs-central-api. The dtfs-central-api is added in #3540 

Add cancellation endpoint, getting data from dtfs-central-api
Add api and unit tests


trade-finance-manager-api/src/v1/deals/routes.js
```diff
@@ -80,7 +81,9 @@ dealsAuthRouter
 
 dealsAuthRouter
   .route('/deals/:dealId/cancellation')
-  .put(validateDealCancellationEnabled, validation.dealIdValidation, handleExpressValidatorResult, dealCancellationController.updateDealCancellation);
+  .all(validateDealCancellationEnabled, validation.dealIdValidation, handleExpressValidatorResult)
```
MarRobSoftwire, line 84: Note to self: add RBAC to the new endpoint once 7370 is merged in

libs/common/src/middleware/feature-flags/deal-cancellation.ts
```diff
@@ -5,8 +5,9 @@ import { isTfmDealCancellationFeatureFlagEnabled } from '../../helpers';
 /**
  * Middleware to check if the deal cancellation feature flag is enabled
  */
-export const validateDealCancellationEnabled: RequestHandler = (_req, res, next) => {
+export const validateDealCancellationEnabled: RequestHandler = (req, res, next) => {
   if (!isTfmDealCancellationFeatureFlagEnabled()) {
+    console.error(`Deal Cancellation feature flag disabled, accessing ${req.originalUrl}`);
```
ttbarnes, line 10: Should this be `.info` since technically it's not an error

trade-finance-manager-api/api-tests/v1/deal-cancellation/deal-cancellation.get.api-test.ts
```diff
@@ -0,0 +1,121 @@
...
+      testUsers = await initialiseTestUsers(app);
+      aPimUser = testUsers().withTeam(TEAM_IDS.PIM).one();
+    });
+
+    afterAll(() => {
+      jest.restoreAllMocks();
+      process.env = originalProcessEnv;
+    });
+
+    const getTfmDealCancellationUrl = ({ id }: { id: string }) => `/v1/deals/${id}/cancellation`;
```
ttbarnes, line 40: Could we create a DRY test helper for this, noticed it's repeated and with different names:

```ts
const getTfmDealCancellationUrl = ({ id }: { id: string }) => `/v1/deals/${id}/cancellation`;
const getTfmDealCancellationUpdateUrl = ({ id }: { id: string }) => `/v1/deals/${id}/cancellation`;
```

trade-finance-manager-api/src/v1/api.js
```diff
@@ -302,6 +302,32 @@ const updateDealCancellation = async ({ dealId, dealCancellationUpdate, auditDet
   }
 };
 
+/**
+ * Updates the deal cancellation object on a TFM AIN or MIN deal
```
ttbarnes, line 306: Do we need to mention AIN and MIN here, this function doesn't specifically care about this?

ttbarnes (changes requested):
Nice work, few minor comments

ttbarnes (approved):

PR #3546, opened 2024-09-19
fix(FN-3460): fix numerical rounding errors during report upload

sw-mattw:
## Introduction :pencil2:
During bank report upload to portal, we occasionally observe floating-point rounding errors for numerical cells. These are presented as validation errors (see screenshot below), meaning the report cannot be submitted.

Facility utilisation values of `3938753.80` and `761579.37` are two examples of this.

A spreadsheet with this issue has been attached to the ticket itself.

**This affects production.**

**Validation errors for the above figures:**
Note that the value displayed in Excel (3938753.80) does not match the value in the entry column of the first row (due to this rounding issue).

369045787-3683f4e4-c47e-4884-a6e4-0b8918516b42

- Added a handler function to handle any floating-point rounding errors and correct these
    - Checks if the rounded 2 decimal place value is within `1e-6` of the original figure. If it is, we take the rounded value, otherwise we use the original figure. This approach means we can still handle genuine errors such as `12.345` (which isn't 2 DP, but is not a floating-point rouding error either).
- Updated valid report spreadsheets in the e2e tests to include the two rows with the above problematic facility utilisation values.

N/A


e2e-tests/portal/cypress/fixtures/valid-utilisation-report-February_2023_monthly.xlsx
sw-mattw, line 1: Each of these spreadsheets have had two additional rows added to include the problematic facility utilisation values in the ticket description.

portal/server/utils/csv-utils.js
```diff
@@ -41,6 +41,32 @@ const excelColumnToColumnIndex = (column) => {
...
+  }
+
+  const TOLERANCE = 1e-6;
+  const DECIMAL_PLACES = 2;
+
+  const roundedNumber = Number(number.toFixed(DECIMAL_PLACES));
+
+  const isRoundedNumberWithinErrorTolerance = Math.abs(number - roundedNumber) < TOLERANCE;
+
+  if (isRoundedNumberWithinErrorTolerance) {
```
abhi-markan, line 62:
```suggestion
return isRoundedNumberWithinErrorTolerance ? roundedNumber : number;
```

portal/server/utils/csv-utils.js
```diff
@@ -41,6 +41,32 @@ const excelColumnToColumnIndex = (column) => {
...
+ */
+const handleFloatingPointRoundingErrors = (number) => {
+  if (typeof number !== 'number') {
+    throw new TypeError('Input must be a number');
+  }
+
+  const TOLERANCE = 1e-6;
+  const DECIMAL_PLACES = 2;
+
+  const roundedNumber = Number(number.toFixed(DECIMAL_PLACES));
```
AlexBramhill, line 58: Out of interest -- have you considered calling `Number(number.toPrecision(2))`?

Not looked into it much however...

 https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toPrecision

portal/server/utils/csv-utils.js
```diff
@@ -41,6 +41,32 @@ const excelColumnToColumnIndex = (column) => {
...
+
+  const isRoundedNumberWithinErrorTolerance = Math.abs(number - roundedNumber) < TOLERANCE;
+
+  if (isRoundedNumberWithinErrorTolerance) {
+    return roundedNumber;
+  }
+
+  return number;
```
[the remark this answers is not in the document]
sw-mattw: Fair point, pushed a change thanks

portal/server/utils/csv-utils.js
AlexBramhill: Ah my bad, thanks :)

portal/server/utils/csv-utils.test.js
```diff
@@ -245,5 +276,49 @@ describe('csv-utils', () => {
...
+    it.each([
+      { input: -761579.3699999999, expected: -761579.37 },
+      { input: -3938753.8000000007, expected: -3938753.8 },
+      { input: -123.456, expected: -123.456 },
+      { input: Number.MIN_SAFE_INTEGER, expected: Number.MIN_SAFE_INTEGER },
+    ])('should return $expected for negative number $input', ({ input, expected }) => {
+      expect(handleFloatingPointRoundingErrors(input)).toBe(expected);
+    });
+
+    it.each(['123.456', null, undefined])('should throw TypeError for non-number input %o', (input) => {
```
abhi-markan, line 320: Nice assertions for type checking, can you please add one with `1e-6` and `1e-5`.

portal/server/utils/csv-utils.js
```diff
@@ -41,6 +41,28 @@ const excelColumnToColumnIndex = (column) => {
...
+ * @param {number} number - The number to check and potentially round.
+ * @returns {number} - The rounded number to two decimal places if within tolerance, otherwise the original number.
+ * @throws {TypeError} - If the input is not a number.
+ */
+const handleFloatingPointRoundingErrors = (number) => {
+  if (typeof number !== 'number') {
+    throw new TypeError('Input must be a number');
+  }
+
+  const TOLERANCE = 1e-6;
```
abhi-markan, line 55: Can these be added to `libs/common` to future proof these very specific constants.

portal/server/utils/csv-utils.js
abhi-markan: Good suggestion.

abhi-markan (changes requested):

portal/server/utils/csv-utils.js
```diff
@@ -41,6 +41,28 @@ const excelColumnToColumnIndex = (column) => {
...
+  const DECIMAL_PLACES = 2;
+
+  const roundedNumber = Number(number.toFixed(DECIMAL_PLACES));
+
+  const isRoundedNumberWithinErrorTolerance = Math.abs(number - roundedNumber) < TOLERANCE;
+
+  return isRoundedNumberWithinErrorTolerance ? roundedNumber : number;
```
AlexBramhill, line 62: My final little review comment in passing is why would we want to ever not return the rounded number? I feel this has some odd behaviour as a result?

ie if we had a number that was 1.234 parsed in, this could be the `actual` value in the database, so we'd want to display it as this. Which is great -- this would work fine.

However, if we had performed an operation on this in the database and it had a rounding error (1.2340000001), this would still return 1.2340000001 -- which still has a floating point error.

As we're always wanting 2dp (as its hard coded in) we could just do something like the below maybe?

```
  const handleFloatingPointRoundingErrorsTwo = (number) => {
    if (typeof number !== 'number') {
      throw new TypeError('Input must be a number');
    }
    const DECIMAL_PLACES = 2;

    return Number(number.toFixed(DECIMAL_PLACES));;
  };
```

AlexBramhill: in your PR you mention 'This approach means we can still handle genuine errors such as 12.345 (which isn't 2 DP, but is not a floating-point rounding error either).', but arent these also potentially exhibiting floating point errors due to them being subject to the same calcs that are being done to give the others floating point errors? :)

AlexBramhill: (I think the approach should be that we round to our tolerance and compare that vs the actual number? So `1.1230001` would be rounded to `1.1230000` and the diff would be `0.0000001`), so it would return `1.1230000`.

As we are dealing with such low values however, are we not getting into the territory of the answer having a floating error (so it'd just make sense to just return the rounded and be done with it -- no checks)?

portal/server/utils/csv-utils.test.js
sw-mattw: Sorry, bit confused - these are valid numbers (and the string case is covered by `123.456`).

If we want to check the boundary, I think these values would be more suitable in the respective tests above:
- `{ input: 9e-7, expected: 0 }` added to the 'inside tolerance' case
- `1e-6` added to the 'outside tolerance' case

Happy to make this change if you're happy with it

portal/server/utils/csv-utils.js
sw-mattw: Pulled these out into `libs/common` thanks

portal/server/utils/csv-utils.js
ChristianMcc: Would be keen to catch up with you this morning on this Alex.
I think I kind of understand your point, where maybe we just round to 7 decimal places for example before doing anything else so that if we need to return an error on 1.123 we can say the error is 1.123 not 1.123000001 in the error message. toFixed wouldn't work here but some math.round function might, will have a play in node

portal/server/utils/csv-utils.js
```diff
@@ -41,6 +42,25 @@ const excelColumnToColumnIndex = (column) => {
   return result - 1;
 };
 
+/**
+ * Checks for floating-point rounding errors and rounds to two decimal places if within error tolerance.
+ * @param {number} number - The number to check and potentially round.
+ * @returns {number} - The rounded number to two decimal places if within tolerance, otherwise the original number.
+ * @throws {TypeError} - If the input is not a number.
+ */
+const handleFloatingPointRoundingErrors = (number) => {
```
ChristianMcc, line 51: also could we add (number): number to give the type of it properly please

portal/server/utils/csv-utils.js
ChristianMcc: Spoke to Matt, we're going to do:

Round to x decimal places, truncate 0s, that's now our number in string format that we run our validations on etc.

If it's javascript float precision error then even with extra decimal places we won't round away from them, but if it's a valid entry we will truncate our 0s down to 2 decimal places and we should be happy

portal/server/utils/csv-utils.js
sw-mattw: JS file so can't add type annotations - the param and return types are specified in the function doc above though so the IDE should pick this up

portal/server/utils/csv-utils.js
sw-mattw: Omitted the truncating as this is done for us automatically in `parseXlsxToCsvArrays` when we concatenate our cell values together when writing the CSV. Implemented the rounding approach though and updated the tests and pushed this work.

Agree with the end of the previous comment.

This approach still flags genuine errors such as 12.345 during validation (our BA expressed that we want to enforce 2 DP, so we need to keep support for those cases).

Tested locally with problematic report, works as expected.

portal/server/utils/csv-utils.test.js
sw-mattw: Added boundary cases to the tests (in line with the latest changes) `987654321.1200005` and `987654321.1200006`.

portal/server/utils/csv-utils.js
sw-mattw: And just for reference, the updated test cases cover a wider range of currency values (hundreds of millions, down to £1.XX), e.g.: `987654321.1200005`, `9999.9999999999`, `1.230001`

abhi-markan (approved):

portal/server/utils/csv-utils.js
ChristianMcc: he he I'm blind, sorry

portal/server/utils/csv-utils.js
sw-mattw: Spoke to Alex - happy to resolve

abhi-markan:
Ignoring failing company house tests due to head branch ahead with recent company house API response change and possibly due to APIM migration.

PR #3547, opened 2024-09-19
chore(DTFS2): allow release branch to release to staging for release-2.0.1

ChristianMcc:
## Introduction :pencil2:
Update release-2.0.1 branch to allow deployment to staging

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -66,7 +66,7 @@ on:
       - feature
       - staging
       - prod
-      - release-2.0
+      - "release-*"
 
     paths:
       - ".github/workflows/deployment.yml"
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -66,7 +66,7 @@ on:
       - feature
       - staging
       - prod
-      - release-2.0
+      - "release-*"
```
abhi-markan, line 69: No double quotes please.

abhi-markan (changes requested):

.github/workflows/test.yml
ChristianMcc: Ah I did this purposefully as this file still has double quotes across it everywhere int he prod version, but will update :D

PR #3548, opened 2024-09-20
feat(DTFS2-7361): add middleware validation to deal cancellation endpoint in tfm api

MarRobSoftwire:
## Introduction :pencil2:
dtfs-central-api uses zod to validate payloads in middleware. We should also validate the payload in tfm-api

- Commonise the middleware validation generator
- Commonise the deal cancellation schema
- Add middleware validation to tfm api

Update api tests & add unit test coverage to dtfs-central-api's deal cancellation middleware validation


libs/common/src/schemas/index.ts
```diff
@@ -1,4 +1,4 @@
 export * as PORTAL_USER from './portal-user';
-export * as AUDIT_DATABASE_RECORD from './audit-database-record';
```
MarRobSoftwire, line 2: This is never used

dtfs-central-api/src/v1/routes/middleware/payload-validation/validate-put-deal-cancellation-payload.test.ts
```diff
@@ -0,0 +1,100 @@
+import { AuditDetails, MAX_CHARACTER_COUNT, TfmDealCancellation } from '@ukef/dtfs2-common';
+import { generateTfmAuditDetails } from '@ukef/dtfs2-common/change-stream';
+import httpMocks from 'node-mocks-http';
+import { HttpStatusCode } from 'axios';
+import { aTfmSessionUser } from '../../../../../test-helpers';
+import { validatePutDealCancellationPayload } from './validate-put-deal-cancellation-payload';
+
+describe('validatePostPaymentPayload', () => {
```
ttbarnes, line 8: Seems this should be PUT instead of POST?

dtfs-central-api/src/v1/routes/middleware/payload-validation/validate-put-deal-cancellation-payload.test.ts
```diff
@@ -0,0 +1,100 @@
...
+    const next = jest.fn();
+
+    req.body = aValidPayload();
+
+    // Act
+    validatePutDealCancellationPayload(req, res, next);
+
+    // Assert
+    expect(next).toHaveBeenCalled();
+    expect(res._isEndCalled()).toBe(false);
```
ttbarnes, line 98: Can we test what is returned?

libs/common/src/middleware/create-validation-middleware-for-schema.ts
```diff
@@ -1,7 +1,8 @@
 import { ZodError, ZodType } from 'zod';
 import { NextFunction, Request, Response } from 'express';
 import { HttpStatusCode } from 'axios';
-import { AnyObject, API_ERROR_CODE, ApiErrorCode, ApiErrorResponseBody } from '@ukef/dtfs2-common';
+import { AnyObject, ApiErrorCode, ApiErrorResponseBody } from '../types';
+import { API_ERROR_CODE } from '../constants';
```
ttbarnes, line 5: Whats the reason for importing from local constants instead of the common library please?

libs/common/src/schemas/deal-cancellation.ts
```diff
@@ -0,0 +1,11 @@
+import z from 'zod';
+import { TfmDealCancellation, ZodTypeFromObject } from '../types';
+import { MAX_CHARACTER_COUNT } from '../constants';
+
+export const DEAL_CANCELLATION: z.ZodObject<ZodTypeFromObject<TfmDealCancellation>> = z
```
ttbarnes, line 5: Documentation?

Can we also use `export default` as there's only 1x function exported

libs/common/src/types/zod-type-from-object.ts
```diff
@@ -0,0 +1,5 @@
+import z from 'zod';
+
+export type ZodTypeFromObject<T extends object> = {
+  [K in keyof T]: z.ZodType<T[K]>;
+};
```
ttbarnes, line 5: Can we add some documentation please, not clear what this is doing

trade-finance-manager-api/src/v1/middleware/validate-put-deal-cancellation-payload.test.ts
```diff
@@ -0,0 +1,80 @@
+import { MAX_CHARACTER_COUNT, TfmDealCancellation } from '@ukef/dtfs2-common';
+import httpMocks from 'node-mocks-http';
+import { HttpStatusCode } from 'axios';
+import { validatePutDealCancellationPayload } from './validate-put-deal-cancellation-payload';
+
+describe('validatePostPaymentPayload', () => {
```
ttbarnes, line 6: Same as above, seems this should be PUT instead of POST?

trade-finance-manager-api/src/v1/middleware/validate-put-deal-cancellation-payload.test.ts
```diff
@@ -0,0 +1,80 @@
...
+      },
+    },
+    {
+      description: "the 'bankRequestDate' is a string",
+      payload: {
+        ...aValidPayload(),
+        bankRequestDate: new Date().toString(),
+      },
+    },
+  ];
```
ttbarnes, line 48: Seems we have both central API and TFM API testing the same validation/response?

trade-finance-manager-api/src/v1/middleware/validate-put-deal-cancellation-payload.test.ts
```diff
@@ -0,0 +1,80 @@
...
+    req.body = aValidPayload();
+
+    // Act
+    validatePutDealCancellationPayload(req, res, next);
+
+    // Assert
+    expect(next).toHaveBeenCalled();
+    expect(res._isEndCalled()).toBe(false);
+  });
+});
```
ttbarnes, line 80: Same as above, can we test what is returned please

ttbarnes (changes requested):

dtfs-central-api/src/v1/routes/middleware/payload-validation/validate-put-deal-cancellation-payload.test.ts
MarRobSoftwire: This is an express handler & doesnt return anything

libs/common/src/middleware/create-validation-middleware-for-schema.ts
MarRobSoftwire: This is a file in libs/common

libs/common/src/schemas/deal-cancellation.ts
MarRobSoftwire: I wasn't aware export default is the prefered.

I've got a slight preference against it - its not extensible & it does slightly weird things if you import into a common js file & you have to write 
```
const { default: theDefaultExport } = require('../path/to/file');
```
Happy to do if this is the pattern we want to run with

libs/common/src/middleware/create-validation-middleware-for-schema.ts
ttbarnes: Sorry I did not realise 😂 🤦

libs/common/src/schemas/deal-cancellation.ts
ttbarnes: Not a blocker thanks @MarRobSoftwire I think its subjective, pros and cons.

Personally I think should always a default export if there's only one thing, I believe it's the default linting rule and we have it disabled

But we already have a mix of both approaches so happy to resolve

ttbarnes: Just some documentation on this function though please 🙏

MarRobSoftwire: Have added :)

trade-finance-manager-api/src/v1/middleware/validate-put-deal-cancellation-payload.test.ts
MarRobSoftwire: this is an express request handler so I dont think it really makes sense to test the returns - I think its void anyway

trade-finance-manager-api/src/v1/middleware/validate-put-deal-cancellation-payload.test.ts
MarRobSoftwire: Yep, this is intentional. My understanding is each microservice should validate anything it recieves & and not rely on this being done elsewhere

trade-finance-manager-api/src/v1/middleware/validate-put-deal-cancellation-payload.ts
```diff
@@ -0,0 +1,9 @@
+import z from 'zod';
+import { TfmDealCancellation, createValidationMiddlewareForSchema } from '@ukef/dtfs2-common';
+import { DEAL_CANCELLATION } from '@ukef/dtfs2-common/schemas';
+
+const PutDealCancellationSchema: z.ZodType<Partial<TfmDealCancellation>> = DEAL_CANCELLATION.partial();
```
ttbarnes, line 5: Why does this have `partial`? I'm not familiar with Zod but I imagine we want to validate the whole schema and not part of it?

MarRobSoftwire: This is just updating the cancellation request so its possible to fill in the `reason` on its own & the other fields be blank

ttbarnes (dismissed):

ttbarnes (approved):

ttbarnes (approved):

PR #3549, opened 2024-09-20
feat(dtfs2-6892): add FF_TFM_SSO_ENABLED feature flag

AlexBramhill:
## Introduction :pencil2:
Establish a pattern for the use of feature flag in tfm for sso

This looks at creating config getters at a router level. This is done as using standard shared interface patterns are not possible due to the increased complexity of 2fa sign in/lack of classes in current implementation. As this feature develops, I'll be sure to review this as we go -- potentially looking at strategies.


.github/workflows/test.yml
```diff
@@ -753,6 +755,7 @@ jobs:
       FF_TFM_PAYMENT_RECONCILIATION_ENABLED: 'true'
       FF_TFM_FACILITY_END_DATE_ENABLED: 'true'
       FF_TFM_DEAL_CANCELLATION_ENABLED: 'true'
+      # TODO DTFS2-6892: add FF_TFM_SSO_ENABLED feature flag here when e2e tests implimented
```
ttbarnes, line 758: Could/should we instead add this in now with a false value?

AlexBramhill: No I dont think so -- this will be set to true eventually. Our FF helper defaults to false anyway.

ttbarnes (approved):

.github/workflows/staging_infrastructure.yml
```diff
@@ -1292,17 +1292,7 @@ jobs:
...
-        set-environment,
-        vnet,
-        private-dns,
-        cosmos-db-private,
-        external-api,
-        dtfs-central-api,
-        file-storage,
-        trade-finance-manager-api,
-      ]
+    needs: [set-environment, vnet, private-dns, cosmos-db-private, external-api, dtfs-central-api, file-storage, trade-finance-manager-api]
```
abhi-markan, line 1295: Formatting issue, can this please be reverted.

docker-compose.yml
```diff
@@ -124,6 +124,7 @@ services:
       - FF_TFM_PAYMENT_RECONCILIATION_ENABLED
       - FF_TFM_FACILITY_END_DATE_ENABLED
       - FF_TFM_DEAL_CANCELLATION_ENABLED
+      - FF_TFM_SSO_ENABLED
```
abhi-markan, line 127: I have added this to GitHub environment values as `false`.

AlexBramhill: thx!

abhi-markan (approved):

PR #3551, opened 2024-09-20
feat(dtfs2-6892): add feature flag around tfm UI router

AlexBramhill:
## Introduction :pencil2:
We need to introduce a feature flag around SSO

This adds in a feature flag to the tfm ui router level.

Due to the additional complexities with flow of SSO and not using classes/typescript, an abstract interface is not possible to handle this logic.

I think this way is probably the best way of doing it for now, and will review as we progress with work


trade-finance-manager-ui/server/controllers/login/loginNonSso/index.js
```diff
@@ -37,12 +37,12 @@ const postLogin = async (req, res) => {
...
 
   if (loginErrors.length) {
-    return res.render('login.njk', {
+    res.render('login.njk', {
       errors: validationErrorHandler(loginErrors),
     });
   }
 
-  return res.redirect('/home');
+  res.redirect('/home');
```
ttbarnes, line 45: Why are the returns removed?

trade-finance-manager-ui/server/routes/login/configs/loginNonSso.ts
```diff
@@ -0,0 +1,16 @@
+/* eslint-disable @typescript-eslint/no-misused-promises */
```
ttbarnes, line 1: Can we remove this?

trade-finance-manager-ui/server/routes/login/configs/loginNonSso.ts
```diff
@@ -0,0 +1,16 @@
...
+import express from 'express';
+import * as loginController from '../../../controllers/login/loginNonSso';
+import { GetConfiguredRouter } from '../../../types/get-configured-router';
+
+export const getConfiguredLoginNonSsoRouter: GetConfiguredRouter = () => {
```
ttbarnes, line 6: Could be simplified here and below to

```ts
getLoginNonSsoRouter
getLoginSsoRouter
```

ttbarnes (changes requested):

trade-finance-manager-ui/server/routes/login/configs/loginNonSso.ts
AlexBramhill: No -- is present in all TS files for routes

trade-finance-manager-ui/server/controllers/login/loginNonSso/index.js
AlexBramhill: Good spot - I dont think we technically need them (although it might be that we dont technically need them only on the last call of `res`, in which case thats inconsistent. reverted

trade-finance-manager-ui/server/routes/user/configs/userNonSso.ts
```diff
@@ -0,0 +1,14 @@
+/* eslint-disable @typescript-eslint/no-misused-promises */
+import express from 'express';
+import { GetConfiguredRouter } from '../../../types/get-configured-router';
+import * as userController from '../../../controllers/user/userNonSso';
+
+export const getConfiguredUserNonSsoRouter: GetConfiguredRouter = () => {
```
ttbarnes, line 6: `getLoginNonSsoRouter`

ttbarnes (approved):

ttbarnes (dismissed):

ttbarnes (approved):

ttbarnes (approved):

PR #3561, opened 2024-09-24
feat(DTFS2-7296): add validation to bank request date

BethThomas141:
## Introduction :pencil2:
The input should be validated when the user submits a bank request date, both the usual date field validation checks, and also that it shouldn't be more than 12 months in the future or more than 12 months in the past. 

- Add post endpoint
- Add validation helper, called by post endpoint

## Screenshots 📸
image
image
image
image


trade-finance-manager-ui/server/controllers/case/cancellation/bank-request-date.controller.test.ts
```diff
@@ -1,8 +1,18 @@
...
 import { PRIMARY_NAVIGATION_KEYS } from '../../../constants';
-import { BankRequestDateViewModel } from '../../../types/view-models';
-import { getBankRequestDate, GetBankRequestDateRequest } from './bank-request-date.controller';
+import { BankRequestDateValidationViewModel, BankRequestDateViewModel } from '../../../types/view-models';
+import { getBankRequestDate, GetBankRequestDateRequest, postBankRequestDate, PostBankRequestDateRequest } from './bank-request-date.controller';
+
+const validateBankRequestDateMock: jest.Mock<BankRequestDateValidationViewModel> = jest.fn(() => ({
+  errors: null,
+  bankRequestDate: undefined,
+}));
```
ttbarnes, line 11: Do we need to define `null` and `undefined` values here?

trade-finance-manager-ui/server/controllers/case/cancellation/validation/validate-bank-request-date.test.ts
```diff
@@ -0,0 +1,100 @@
...
+    const expectedError = {
+      summary: [{ text: expectedMessage, href: '#bank-request-date-day' }],
+      bankRequestDateError: { message: expectedMessage, fields: ['bank-request-date-day', 'bank-request-date-month', 'bank-request-date-year'] },
+    };
+
+    expect(result.errors).toEqual(expectedError);
+  });
+
+  it('returns an error if a field is left blank', () => {
+    const result = validateBankRequestDate({ day: '', month: '1', year: '2024' });
```
ttbarnes, line 76: Can we add some tests for if `month` or `year` are left empty?

trade-finance-manager-ui/server/controllers/case/cancellation/validation/validate-bank-request-date.ts
```diff
@@ -0,0 +1,51 @@
...
+      bankRequestDateError: { message, fields: refs },
+    },
+  };
+};
+
+/**
+ * @param date The entered bank request date
+ * @returns a reason for cancelling errors view model
+ */
+export const validateBankRequestDate = (date: DayMonthYearInput): BankRequestDateValidationViewModel => {
```
ttbarnes, line 26: Do we have any other date validations we could use? I imagine they all have the same rules

ttbarnes (changes requested):
Great PR - can we add some E2E tests please?

trade-finance-manager-ui/server/controllers/case/cancellation/validation/validate-bank-request-date.ts
BethThomas141: On line 27 we call `applyStandardValidationAndParseDateInput` which is the commonised date validation checker I made a few weeks back. So this handles all the standard date checks (e.g. missing fields, unexpected values etc.) and this new validator is just to add the specific handling on top of that that is unique to bank request date - which is that it's not more than 12 months in the future or more than 12 months in the past.

trade-finance-manager-ui/server/controllers/case/cancellation/bank-request-date.controller.test.ts
BethThomas141: I've changed it to: 

```
const validateBankRequestDateMock: jest.Mock<BankRequestDateValidationViewModel> = jest.fn(() => ({
  errors: null,
  bankRequestDate: new Date(),
}));
```

which is more realistic, as we expect either a date (and errors to be null) or errors and the bank request date to be undefined - thanks! Any errors we test are mocked specifically further down the file so fine for them to be null here.

trade-finance-manager-ui/server/controllers/case/cancellation/validation/validate-bank-request-date.test.ts
ttbarnes: Thanks @BethThomas141 makes sense, I don't think we should have duplicate tests like you say.

I do think we need to ensure that in here, or any other function that calls/consumes `applyStandardValidationAndParseDateInput`, need to make sure that it is called - not sure what's in place already but maybe something like this would suffice?

```ts
describe('X field validation', () => {
  describe('custom date rules', () => {
    it('should ...'
    it('should ...'
  });

  describe('when all custom date rules are valid', () => {
    it('should return the result of applyStandardValidationAndParseDateInput', () => {
      const expected = applyStandardValidationAndParseDateInput();

    });
  });
});
```

BethThomas141: Good idea, thanks! I've now done something to this effect - have a look and let me know what you think.

trade-finance-manager-ui/server/controllers/case/cancellation/validation/validate-bank-request-date.test.ts
```diff
@@ -0,0 +1,90 @@
...
+  describe('base date validation rules', () => {
+    it('should return the result of applyStandardValidationAndParseDateInput', () => {
+      const dateObjectToTest = { day: '', month: '1', year: '2024' };
+
+      const { errors: receivedErrors } = validateBankRequestDate(dateObjectToTest);
+
+      const { error: expectedError } = applyStandardValidationAndParseDateInput(dateObjectToTest, 'bank request date', 'bank-request-date');
+
+      expect(receivedErrors?.bankRequestDateError.message).toEqual(expectedError?.message);
+      expect(receivedErrors?.bankRequestDateError.fields).toEqual(expectedError?.fieldRefs);
```
ttbarnes, line 87: Could we just test the whole thing? e.g

```ts
const result = validateBankRequestDate(dateObjectToTest);

// ...

const expected = getErrorObjectFromMessageAndRefs();

```

trade-finance-manager-ui/server/controllers/case/cancellation/validation/validate-bank-request-date.test.ts
ttbarnes: Thanks @BethThomas141 looks good to me 

Appreciate this isn't part of DTFS2-7296 - are we able to check and apply the same treatment to other date unit tests that use `applyStandardValidationAndParseDateInput`, so they're all aligned? Not sure if this would be a quick thing to do

trade-finance-manager-ui/server/controllers/case/cancellation/validation/validate-bank-request-date.test.ts
BethThomas141: Ah the format we get from `applyStandardValidationAndParseDateInput` and `validateBankRequestDate` is different, so as you've mentioned there we'd have to call `getErrorObjectFromMessageAndRefs` in the test. I think it's best not to do that, as that function isn't exported, so it would be being exported just for a test here which I don't think is worth it.

trade-finance-manager-ui/server/controllers/case/cancellation/validation/validate-bank-request-date.test.ts
BethThomas141: I think we leave it for now as it's out of scope of this bit of work, but certainly going forwards we can

trade-finance-manager-ui/server/controllers/case/cancellation/validation/validate-bank-request-date.test.ts
ttbarnes: Ah, I think we should though @BethThomas141 - didn't notice that `getErrorObjectFromMessageAndRefs` isn't tested, can we add a test for this please

Then we can add a "full" test as mentioned above 😎

At the moment I could remove `getErrorObjectFromMessageAndRefs` and create something else and the test  would still pass

trade-finance-manager-ui/server/controllers/case/cancellation/validation/validate-bank-request-date.test.ts
ttbarnes: Can we create a ticket for this please? So we don't forget to make it aligned with the nice new approach we have here 😊

BethThomas141: It looks like we have tests that would be impacted in only 2 files - so once we've nailed down the approach in this PR, I'll do a PR straight away afterwards this afternoon to amend those other two (but will keep it out of this PR).

trade-finance-manager-ui/server/controllers/case/cancellation/validation/validate-bank-request-date.test.ts
BethThomas141: I've exported `getErrorObjectFromMessageAndRefs` and added a test for it. 

I've just had a chat with @AlexBramhill and realised the best way to test this function as a unit test is to mock the response of `applyStandardValidationAndParseDateInput`. So I've now instead mocked this and check that the mocked response is returned. Have a look and let me know what you think! :)

ttbarnes: Thanks @BethThomas141 

Why do we need to mock `applyStandardValidationAndParseDateInput please?

ttbarnes: Can we update the `validateBankRequestDate` tests to use `getErrorObjectFromMessageAndRefs` and test the whole returned object please? So e.g instead of this:

```ts
const expectedError = {
  summary: [{ text: expectedMessage, href: '#bank-request-date-day' }],
  bankRequestDateError: { message: expectedMessage, fields: ['bank-request-date-day', 'bank-request-date-month', 'bank-request-date-year'] },
};

expect(result.errors).toEqual(expectedError);
```

We do this for example:

```ts
const expected = getErrorObjectFromMessageAndRefs('err message', fieldRefs);

```

BethThomas141: @ttbarnes `applyStandardValidationAndParseDateInput` should be mocked as that is a function call from outside the file - as this is a unit test we should be mocking returns like this. This unit test doesn't care what `applyStandardValidationAndParseDateInput` returns, just that it is called and we use it's result - what this returns is tested in the unit test for that function.

ttbarnes (approved):

PR #3563, opened 2024-09-24
feat(DTFS2-7300): add effective from date template

BethThomas141:
## Introduction :pencil2:
The PIM user needs to be able to provide the effective from date for the deal cancellation. This ticket does not include validating the values or making updates to the database

Add bank request date template
Add a new GET endpoint to render this template

## Screenshots 📸
image

### DIFF
--- a/trade-finance-manager-ui/server/types/view-models/deal-cancellation/effective-from-date-view-model.ts
+++ b/trade-finance-manager-ui/server/types/view-models/deal-cancellation/effective-from-date-view-model.ts
@@ -12,5 +12,5 @@ export type EffectiveFromDateViewModel = BaseViewModel & {
   day?: string;
   month?: string;
   year?: string;
-  errors?: EffectiveFromDateErrorsViewModel | null;
+  errors?: EffectiveFromDateErrorsViewModel;
 };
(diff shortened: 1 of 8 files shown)


trade-finance-manager-ui/server/types/view-models/deal-cancellation/effective-from-date-view-model.ts
```diff
@@ -0,0 +1,16 @@
...
+  effectiveFromDateError: { message: string; fields: string[] };
+};
+
+export type EffectiveFromDateViewModel = BaseViewModel & {
+  ukefDealId: string;
+  dealId: string;
+  day?: string;
+  month?: string;
+  year?: string;
+  errors?: EffectiveFromDateErrorsViewModel | null;
```
ttbarnes, line 15: Would this work?


`errors?: EffectiveFromDateErrorsViewModel`

ttbarnes (commented):

trade-finance-manager-ui/server/types/view-models/deal-cancellation/effective-from-date-view-model.ts
BethThomas141: Ah yes, I think so thank you! Have changed

ttbarnes (approved):

PR #3541, opened 2024-09-18
chore(tests): split up e2e tests - tfm case

ttbarnes:
## Introduction :pencil2:
This PR splits up the "TFM - case" E2E tests into child runners/directories.

Prior to this change, these tests took ~15 minutes to run.

Each directory now takes ~1 minute - 1.5 minutes to execute, with 1x test taking the most amount of time at 2.5 minutes.

This is part of a series of E2E test splitting to reduce E2E GHA time.

## Resolution :heavy_check_mark:
- Split up some root "TFM - case" E2E tests.
- Split up root "TFM - case - parties" E2E tests.
- Split up root "TFM - case - tasks" E2E tests.
- Update `test.yml`.

## Miscellaneous :heavy_plus_sign:
- Fix a typo (`bankResponse` => `bankResponses`).
- Remove an unnecessary `.clear()`.

PR #3571, opened 2024-09-25
feat(DTFS2-7350): update db on reason for cancelling page

MarRobSoftwire:
## Introduction :pencil2:
The reason for cancelling page should update the database

- get reason for cancelling from the database
- update reason for cancelling
- validate deal exists & submission type is valid on both endpoints

- Add GetDealResponse data transfer object
- Make get deal cancellation response partial
- Dont throw 404 error when cancellation doesnt already exist


dtfs-central-api/src/repositories/tfm-deals-repo/tfm-deal-cancellation.repo.ts
```diff
@@ -32,7 +24,7 @@ export class TfmDealCancellationRepo {
     }
 
     if (!matchingDeal.tfm.cancellation) {
-      throw new DealCancellationNotFoundError(dealId.toString());
+      return {};
```
ttbarnes, line 27: Why do we want to return an object here - is the cancellation object not created when a deal reaches TFM (and therefore all deals should have a cancellation object)?

ttbarnes, line 27: Why do we want to return an object here - is the cancellation object not created when a deal reaches TFM (and therefore all deals should have a cancellation object)?

trade-finance-manager-ui/server/controllers/case/cancellation/reason-for-cancelling.controller.ts
```diff
@@ -14,17 +16,34 @@ export type PostReasonForCancellingRequest = CustomExpressRequest<{ params: { _i
...
-    user,
-    ukefDealId: '0040613574', // TODO: DTFS2-7350 get values from database
-    dealId: _id,
-  };
-  return res.render('case/cancellation/reason-for-cancelling.njk', reasonForCancellingViewModel);
+  const { user, userToken } = asUserSession(req.session);
+  try {
+    const deal = await api.getDeal(_id, userToken);
+
+    if (!deal || 'status' in deal) {
```
ttbarnes, line 25:
```suggestion
    if (!deal) {
```

trade-finance-manager-ui/server/controllers/case/cancellation/reason-for-cancelling.controller.post.test.ts
```diff
@@ -0,0 +1,244 @@
...
+  validateReasonForCancelling: (reason: string | undefined) => validateReasonForCancellingMock(reason),
+}));
+
+jest.mock('../../../api', () => ({
+  getDeal: jest.fn(),
+  updateDealCancellation: jest.fn(),
+}));
+
+jest.mock('../../helpers', () => ({
+  canSubmissionTypeBeCancelled: jest.fn(() => true),
```
ttbarnes, line 25: Can we test without mocking this helper? I.e pass different mock data into the helper rather than mocking the helper's response

trade-finance-manager-ui/server/controllers/helpers/deal-cancellation-enabled.helper.ts
```diff
@@ -2,6 +2,15 @@ import { DEAL_SUBMISSION_TYPE, DealSubmissionType, isTfmDealCancellationFeatureF
...
 import { TfmSessionUser } from '../../types/tfm-session-user';
 
+/**
+ * Checks if the deal submission type can be cancelled
+ * @param submissionType - the deal submission type
+ * @returns true if the deal is AIN or MIN
+ */
+export const canSubmissionTypeBeCancelled = (submissionType: DealSubmissionType): boolean => {
+  return submissionType === DEAL_SUBMISSION_TYPE.AIN || submissionType === DEAL_SUBMISSION_TYPE.MIN;
+};
```
ttbarnes, line 12: Could we destructure `AIN` and `MIN` then maybe could do this:
```ts
export const canSubmissionTypeBeCancelled = (submissionType: DealSubmissionType): boolean =>
  submissionType === AIN || submissionType === MIN;
```

trade-finance-manager-ui/server/routes/case/cancellation.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-misused-promises */
```
ttbarnes, line 1: Is this intentional?

ttbarnes (changes requested):

dtfs-central-api/src/repositories/tfm-deals-repo/tfm-deal-cancellation.repo.ts
MarRobSoftwire: Nope, the trend in tfm deals seems to be create objects when they're needed (e.g. ammendments) so I've stuck to that.
I think this is the easiest thing to handle from a front end perspective - (I could return a 404 but that raises some difficult questions given the slightly odd error handling in tfm-api/ui. Alternatively `null` or `undefined` would work but I think I prefer an empty object). 

I expect we'll add a `stage` to the cancellation object at some point & then it might make more sense for this to return something like `stage: 'Not started'`

trade-finance-manager-ui/server/controllers/case/cancellation/reason-for-cancelling.controller.post.test.ts
MarRobSoftwire: I think as a unit test, it makes sense to mock other dependencies

MarRobSoftwire: The `canSubmissionTypeBeCancelled` helper is tested in its own unit test

trade-finance-manager-ui/server/controllers/case/cancellation/reason-for-cancelling.controller.ts
MarRobSoftwire: `getDeal` returns a really weird object if the deal id isnt valid so I'm accounting for that here.

trade-finance-manager-ui/server/routes/case/cancellation.ts
MarRobSoftwire: Yep, this is an issue in express (fixed in v5 I believe) that the typing incorrectly expects a `(req, res ) => void` rather than `(req, res) => Promise<void>`.

As far as I know (short of updating express) there's no fix to this :(

(e.g. see https://stackoverflow.com/questions/65670264/typescript-node-express-router-and-second-argument-typescript-eslint-no-misused)

dtfs-central-api/src/repositories/tfm-deals-repo/tfm-deals-cancellation-repo.test.ts
```diff
@@ -47,6 +47,19 @@ describe('tfm-deals-cancellation-repo', () => {
       await expect(TfmDealCancellationRepo.findDealCancellationByDealId(dealId)).rejects.toThrow(new DealNotFoundError(dealId.toString()));
     });
 
+    it('returns an empty object if a deal is found without a cancellation', async () => {
```
abhi-markan, line 50: Can we add an assertion with an invalid `dealId` please.

dtfs-central-api/src/repositories/tfm-deals-repo/tfm-deal-cancellation.repo.ts
abhi-markan: Can we please add a line on why we are returning an empty object please (pretty much what you have explained above), for a better DevX.

trade-finance-manager-api/src/v1/api.js
```diff
@@ -305,7 +305,7 @@ const updateDealCancellation = async ({ dealId, dealCancellationUpdate, auditDet
 /**
  * Gets the deal cancellation object on a TFM deal
  * @param {string} dealId - deal cancellation to update
- * @returns {Promise<import('@ukef/dtfs2-common').TfmDealCancellation>} - Deal cancellation object
+ * @returns {Promise<Partial<import('@ukef/dtfs2-common').TfmDealCancellation>>} - Deal cancellation object
```
abhi-markan, line 308: Nice use of utility type.

trade-finance-manager-ui/server/api.js
```diff
@@ -1251,6 +1258,27 @@ const updateDealCancellation = async (dealId, cancellationUpdate, userToken) =>
...
+ */
+const getDealCancellation = async (dealId, userToken) => {
+  try {
+    const response = await axios({
+      method: 'get',
+      url: `${TFM_API_URL}/v1/deals/${dealId}/cancellation`,
+      headers: generateHeaders(userToken),
+    });
+
+    return response.data;
```
abhi-markan, line 1275:
```suggestion
    return response?.data;
```

trade-finance-manager-ui/server/controllers/case/cancellation/reason-for-cancelling.controller.get.test.ts
```diff
@@ -0,0 +1,114 @@
...
+const ukefDealId = 'ukefDealId';
+const mockUser = aTfmSessionUser();
+
+describe('getReasonForCancelling', () => {
+  beforeEach(() => {
+    jest.resetAllMocks();
+    jest.mocked(canSubmissionTypeBeCancelled).mockReturnValue(true);
+  });
+
+  it('redirects to not found if the deal does not exist', async () => {
```
abhi-markan, line 29: An assertion for a non-PIM user please.

trade-finance-manager-ui/server/controllers/case/cancellation/reason-for-cancelling.controller.get.test.ts
```diff
@@ -0,0 +1,114 @@
...
+import api from '../../../api';
+import { canSubmissionTypeBeCancelled } from '../../helpers';
+
+jest.mock('../../../api', () => ({
+  getDeal: jest.fn(),
+  getDealCancellation: jest.fn(),
+}));
+
+jest.mock('../../helpers', () => ({
+  canSubmissionTypeBeCancelled: jest.fn(() => true),
```
abhi-markan, line 16: Do we need to mock this, can the actual function be used rather?

trade-finance-manager-ui/server/controllers/case/cancellation/reason-for-cancelling.controller.post.test.ts
abhi-markan: I would recommend testing the function without mocking for an effecting and a better coverage of the helper function.

trade-finance-manager-ui/server/controllers/case/cancellation/reason-for-cancelling.controller.ts
abhi-markan: Can it be accomplished using optional chaining please?

trade-finance-manager-ui/server/controllers/case/cancellation/reason-for-cancelling.controller.ts
```diff
@@ -14,17 +16,34 @@ export type PostReasonForCancellingRequest = CustomExpressRequest<{ params: { _i
...
+      return res.redirect('/not-found');
+    }
+
+    if (!canSubmissionTypeBeCancelled(deal.dealSnapshot.submissionType)) {
+      return res.redirect(`/case/${_id}/deal`);
+    }
+
+    const cancellation = await api.getDealCancellation(_id, userToken);
+
```
abhi-markan, line 34: Should we verify `cancellation` object for essential properties?

trade-finance-manager-ui/server/controllers/case/cancellation/reason-for-cancelling.controller.ts
```diff
@@ -33,27 +52,44 @@ export const getReasonForCancelling = (req: GetReasonForCancellingRequest, res:
...
   const { reason } = req.body;
-  const { user } = asUserSession(req.session);
+  const { user, userToken } = asUserSession(req.session);
 
-  const validationErrors = validateReasonForCancelling(reason);
+  try {
+    const deal = await api.getDeal(_id, userToken);
 
-  const formHasErrors = validationErrors.errorSummary.length !== 0;
+    if (!deal || 'status' in deal) {
```
abhi-markan, line 63: Same above.

trade-finance-manager-ui/server/controllers/helpers/deal-cancellation-enabled.helper.test.ts
```diff
@@ -26,3 +26,23 @@ describe('dealCancellationEnabled', () => {
     });
   });
 });
+
+describe('canSubmissionTypeBeCancelled', () => {
+  it('returns true when deal submission type is AIN', () => {
```
abhi-markan, line 31: One minor not a blocker, might be effective to use `it.each(...`

dtfs-central-api/src/repositories/tfm-deals-repo/tfm-deal-cancellation.repo.ts
ttbarnes: Interesting - not for this PR, but maybe we should create it during TFM deal creation if it's giong to cause some odd behaviours/logic/returns et

trade-finance-manager-ui/server/controllers/case/cancellation/reason-for-cancelling.controller.ts
ttbarnes: This is checking for the status though; I'm not seeing this in other areas?

trade-finance-manager-ui/server/types/data-transfer-objects/get-deal-response.ts
```diff
@@ -0,0 +1,9 @@
+import { DealSubmissionType } from '@ukef/dtfs2-common';
+
+/**
+ * Data Transfer Object sent from the trade-finance-manager-api endpoint /v1/deals/:id
+ *
+ * This type is likely incomplete and should be added
+ * to as and when new properties are discovered
+ */
+export type GetDealResponse = { dealSnapshot: { details: { ukefDealId: string }; submissionType: DealSubmissionType } } | undefined;
```
abhi-markan, line 9: Can it be multiline please to improve readability.
Is this union type because this DTO can also be used when deal is not found?

abhi-markan (changes requested):

trade-finance-manager-ui/server/controllers/case/cancellation/reason-for-cancelling.controller.post.test.ts
ttbarnes: I can't see any reason to mock this helper in the unit test (or any helper in a unit test really that is not async) - can we please update it so it's not mocked? E.g

```ts

// in test 1 - deal can be cancelled
const mockXDeal = {};
jest.mocked(api.getDeal).mockResolvedValue(mockXDeal);


// in test 2 - deal cannot be cancelled
const mockYDeal = {};
jest.mocked(api.getDeal).mockResolvedValue(mockYDeal);
```

Then we don't need things like this:
```ts
jest.mock('../../helpers', () => ({
  canSubmissionTypeBeCancelled: jest.fn(() => true),
}));

jest.mocked(canSubmissionTypeBeCancelled).mockReturnValueOnce(false);
```

trade-finance-manager-ui/server/routes/case/cancellation.ts
ttbarnes: Thanks @MarRobSoftwire , unfortunate 😞 I can also see we already have this disabling rule in many places.

trade-finance-manager-ui/server/controllers/case/cancellation/reason-for-cancelling.controller.post.test.ts
ttbarnes: Same with `validateReasonForCancelling` as well @MarRobSoftwire , I can't see any reason to mock these.

The controllers consume/call these functions, the tests can just call these with different parms/arguments, instead of mocking the other functions responses. I would say that by _not_ doing this, it's not representative of what the controller is actually doing

trade-finance-manager-ui/server/api.js
MarRobSoftwire: I dont think `response` is ever falsey?

Pasting it into a ts file and adding this
```
if (!response) {
  console.log(response);
}
```
`response` has type `never` inside the block

trade-finance-manager-ui/server/controllers/case/cancellation/reason-for-cancelling.controller.get.test.ts
MarRobSoftwire: The PIM user's access is prevented in middleware so cant be tested in this unit test.

TFM-UI doesn't have any api-tests currently, which makes this a bit harder to test - I'd like to add some but I think we'll need to get confirmation from Tom that hes happy with that, and to check this wont conflict too painfully with SSO for @AlexBramhill

trade-finance-manager-ui/server/controllers/case/cancellation/reason-for-cancelling.controller.get.test.ts
MarRobSoftwire: Since this is a unit test I think it makes sense to mock this dependency & rely on the unit tests for `canSubmissionTypeBeCancelled` to check AIN/MIA.

trade-finance-manager-ui/server/controllers/case/cancellation/reason-for-cancelling.controller.ts
MarRobSoftwire: I'm not sure what you mean? Do you mean check that the object returned by `tfm-api` is valid? I can add something like zod validation to this response type but its not done anywhere else in tfm-ui

trade-finance-manager-ui/server/types/data-transfer-objects/get-deal-response.ts
MarRobSoftwire: Yep, this is what `getDeal` returns from the api.
I'm not sure if `response?.data` will ever be undefined, but it seems safest to assume that it could be

trade-finance-manager-ui/server/controllers/case/cancellation/reason-for-cancelling.controller.get.test.ts
ttbarnes: Thing is @MarRobSoftwire it's not representative of what it's actually doing. For example, with the current approach, if I change the mock for `api.getDeal` to return a deal with an MIA `submissionType`, the `it('renders the reason for cancelling page'` test will still pass, because `canSubmissionTypeBeCancelled` is marked to return true regardless of the `submissionType`. Outside of a unit test, the page would not render and it'd redirect instead

AlexBramhill: Just weighing in some thoughts here -- not sure if any of this has already been said.

My understanding is that a unit test should test the smallest isolatable testable unit incredibly thoroughly. The `unit` in this case is a function.

The question over mocking external areas is always interesting. I always lean to mock everything that we're importing (exceptions being external modules). This is because:
1. We test these imported `units` elsewhere with 100% coverage (if we don't, as part of this PR we should).
2. The tests should not care about the implementation of other units, otherwise these stop being unit tests and start being half unit, half api tests. It should only care that `given I provide a function with x, it sends back y`
3. Testing these units here does not add any additional coverage to our testing as long as point 1 above is true (and it should be)

We should look to have a test double folder for these mocks to avoid having to repeat them.

My thoughts are in the screenshot below -- with the red showing what we should unit test.

{D17CBDD5-D13F-4299-A975-54A6D9701800}

This would achieve 100% code coverage as this ethos of smallest possible testing unit is continued throughout the codebase

{F71900A9-6505-4FF0-9A21-BC9989D184BC}

I think not mocking helpers is a bit odd -- as we have unit tests which look like this, and covers more than a minimal tested unit: in purple 
{22FA3D1E-1DAF-426A-9D7D-BD57272ADD52}

AlexBramhill: alternatively, if we are wanting to test this, which I think is a bit odd as effectively we have unit tests relying on the implementation of imported functions (and all the reasons above), we should 100% look to, as part of creating helpers, have a `withBlaHelper({makeRequest, addBlaParameterToRequest: (param)=> {baseReq, field: param}})` and then have all the tests in a helper to keep tests dry

trade-finance-manager-ui/server/controllers/case/cancellation/reason-for-cancelling.controller.ts
BethThomas141: I think this is why we were originally keen to `throw` errors to be handled at this layer - without `throw`ing the error and returning objects like this, we have to check if the returned object has a `status` field to know it's errored, instead of just catching an error (and I think as we spotted live on call with Alex, quite a few places currently don't handle this correctly when they're returning this arbitrary `status` object).

trade-finance-manager-ui/server/controllers/case/cancellation/reason-for-cancelling.controller.ts
abhi-markan:
```javascript
if (!cancellation?.reason) {...}
```

I meant something like above, this is quite common across DTFS.

trade-finance-manager-ui/server/controllers/case/cancellation/reason-for-cancelling.controller.ts
abhi-markan: Is `!deal?.status` not an option?

trade-finance-manager-ui/server/controllers/case/cancellation/reason-for-cancelling.controller.get.test.ts
abhi-markan: Can we not mock a PIM user in the user session and assert with the expected behaviour?

```typescript
  session: {
        user: mockPimUser,
        userToken: 'a pim user token',
      },
```

trade-finance-manager-ui/server/controllers/case/cancellation/reason-for-cancelling.controller.get.test.ts
abhi-markan: Thanks @AlexBramhill  for the above detailed explanation.
I can see the helper function is quite a simple implementation

```typescript
export const canSubmissionTypeBeCancelled = (submissionType: DealSubmissionType): boolean => {
  return submissionType === AIN || submissionType === MIN;
}
```

Ideally I would have suggested to mock a database call or an expensive external API call but given this function merely ascertain whether the deal submission type is either notice or not I think mocking is not required.

Though agree with unit test coverage but mocking a simple implementation will not harm based on mock deal it should return `true` anyways. This also works well for future refactorisation of this controller if the scope is expanded and we will also need to cater of deal type and its associated facility types.

abhi-markan (changes requested):
Thanks for the update, few remaining changes to address please.

trade-finance-manager-ui/server/controllers/case/cancellation/reason-for-cancelling.controller.post.test.ts
ttbarnes: Can we please replace the mocks for the following @BethThomas141 

```ts
jest.mock('./validation/validate-reason-for-cancelling'
jest.mock('../../helpers'
```

Same principle as comment from @abhi-markan  here https://github.com/UK-Export-Finance/dtfs2/pull/3571#discussion_r1780772262

trade-finance-manager-ui/server/controllers/case/cancellation/reason-for-cancelling.controller.get.test.ts
AlexBramhill: Sounds great to me

trade-finance-manager-ui/server/controllers/case/cancellation/reason-for-cancelling.controller.ts
BethThomas141: `!deal?.status` would be different: 

Suppose we have an error object returned that is `{status: 400, data: 'abc'}`

**Then with the current implementation: `if (!deal || 'status' in deal) {`**
`status` is in `{status: 400, data: 'abc'}` so this is `true` and this `if` block runs. 

**With the suggestion: `if (!deal?.status) {`**
`deal?.status` is `true` so `!deal?.status` is `false` so this block would not run. 

So these are two different behaviours. I don't think we can make Mark's implementation more succinct.

trade-finance-manager-ui/server/controllers/helpers/deal-cancellation-enabled.helper.test.ts
BethThomas141: This is now implemented, thanks!👍

trade-finance-manager-ui/server/controllers/case/cancellation/reason-for-cancelling.controller.get.test.ts
BethThomas141: @abhi-markan Mocking this in this test won't do anything as the PIM validation happens in middleware outside of this test file. 
We have two options: 
1) test this in an api-test (preferred) 
OR 2) test this in an E2E test (breaks consistency)

With option 1, this is what we'd prefer to do, but we currently don't have any api-tests at all in `tfm-ui`. Mark has a ticket up for this and I've noted that we should test the PIM/non-PIM users in deal cancellation there when that ticket is scoped and run by Tom. https://ukef-dtfs.atlassian.net/browse/DTFS2-7363

With option 2, we don't currently do this for e2e tests (as in e.g. for amendments we don't check non-PIM users can't amend deals in an e2e test so this would introduce a new pattern). 

So our preference is 1, and that will be deferred to a separate PR as part of ticket 7363 linked to above when that gets approved. How does that sound?

trade-finance-manager-ui/server/controllers/case/cancellation/reason-for-cancelling.controller.get.test.ts
BethThomas141: @abhi-markan @ttbarnes I've refactored the tests to no longer mock `canSubmissionTypeBeCancelled` 👍

abhi-markan: Thank you both, will mark it as resolved.

trade-finance-manager-ui/server/controllers/case/cancellation/reason-for-cancelling.controller.post.test.ts
```diff
@@ -0,0 +1,240 @@
...
+import api from '../../../api';
+
+const validateReasonForCancellingMock: jest.Mock<ReasonForCancellingErrorsViewModel> = jest.fn(() => ({
+  errorSummary: [],
+  reasonForCancellingErrorMessage: undefined,
+}));
+
+jest.mock('./validation/validate-reason-for-cancelling', () => ({
+  validateReasonForCancelling: (reason: string | undefined) => validateReasonForCancellingMock(reason),
+}));
```
ttbarnes, line 16: Can we replace the 2x mocks above please @BethThomas141  ?

trade-finance-manager-ui/server/controllers/case/cancellation/reason-for-cancelling.controller.post.test.ts
BethThomas141: @ttbarnes I've now replaced the mock for `canSubmissionTypeBeCancelled` 👍 

I'd be keen to keep mocking `validateReasonForCancelling` - the reason for this one, is it's a more complicated function call and the unit tests for our controller here should just check that the validation response we get back is displayed and formatted correctly on the Nunjucks page. The actual message that we get is tested in the unit test for `validate-reason-for-cancelling.test.ts` so doesn't need re-testing here. 

If we don't mock it here, every time we change any of the logic in `validate-reason-for-cancelling.ts` we will have to change the unit tests in _both_ `validate-reason-for-cancelling.test.ts` and the tests here, touching two files instead of one. 

Would you be ok with a compromise of removing the mock for  `canSubmissionTypeBeCancelled`, but keeping the mock in for `validateReasonForCancelling`? 🙂

ttbarnes: Thanks @BethThomas141 I see the point - the issue I find with the mocking of the validation (`validateReasonForCancelling`) (or any other type functions called in a unit test, excluding async/await calls), is that essentially:

What happens outside of unit tests:
1. Controller receives a body.
2. Controller calls the validation function with that body.
3. Controller does XYZ if there are or aren't validation errors.

But the test is doing this:

1. Controller receives a body.
2. Validation function returns whatever we mock, instead of it's true logic, based on what is in the payload.
  a. This can also lead to misleading mocks/tests.
  b. Currently you can pass in an invalid payload and it'll still pass - it'd expect it to only pass with a valid payload.
4. Controller does XYZ if there are or aren't validation errors.

We could have an e.g `validBody` object and 1x simple mock object with errors or even just an empty object and then remove the mock,. Then it's mimicking what the controller is actually doing - calling the validation function instead of mocking/stubbing it.

There are lots of mock instances I think we should remove/replace, but one for another day. Therefore happy to go with current approach, resolving

e2e-tests/tfm/cypress/e2e/journeys/deal-cancellation/reason-for-cancelling.spec.js
```diff
@@ -75,6 +74,16 @@ context('Deal cancellation - reason for cancelling', () => {
 
       cy.url().should('eq', relative(`/case/${dealId}/cancellation/cancel`));
     });
+
+    it('returning to the page should display saved data', () => {
+      const reason = 'A Reason';
+      cy.keyboardInput(reasonForCancellingPage.reasonForCancellingTextBox().clear(), reason);
```
ttbarnes, line 80: Need to remove `.clear()` here, test is probably failing as is @BethThomas141 🙏

trade-finance-manager-ui/server/controllers/case/cancellation/reason-for-cancelling.controller.get.test.ts
ttbarnes: I don't think we need to test a PIM user here at all @abhi-markan as it is handled in middleware which will be called before the controller is hit

abhi-markan (dismissed):
Good work, thank you both. 👍🏼

trade-finance-manager-ui/server/controllers/case/cancellation/reason-for-cancelling.controller.ts
BethThomas141: Discussed and resolved in other thread👍

e2e-tests/tfm/cypress/e2e/journeys/deal-cancellation/reason-for-cancelling.spec.js
ttbarnes: Lets get this merged, I will pick this up in [my PR](https://github.com/UK-Export-Finance/dtfs2/pull/3541) 

I don't think the deal cancellation tests are running in GHA - otherwise this test should be failing because of the `.clear()`

```
'./tfm/cypress/e2e/journeys/deal-cancellation/**/*.spec.js'

to 

'./tfm/cypress/e2e/journeys/deal-cancellation/*.spec.js'
```

Or we could do this in the next PR?

ttbarnes (approved):

dtfs-central-api/src/repositories/tfm-deals-repo/tfm-deals-cancellation-repo.test.ts
BethThomas141: Looks like Mark has added a test for this in👍

trade-finance-manager-ui/server/controllers/case/cancellation/reason-for-cancelling.controller.post.test.ts
BethThomas141: Thanks Tony - appreciate you writing that out. I can see pros and cons of both sides here, it's a tricky one. Thanks for resolving, happy to revisit the discussion in future.

e2e-tests/tfm/cypress/e2e/journeys/deal-cancellation/reason-for-cancelling.spec.js
BethThomas141: Ooh interesting, so I've removed it, but I don't think the `.clear()` does impact whether or not the test passes/fails. 

It has the right structure in the `test.yml` file, it's consistent with the other similar subfolders, and the tests are indeed running:

https://github.com/UK-Export-Finance/dtfs2/actions/runs/11109438117/job/30864938680?pr=3571

image

So I think it's just a case that `.clear()` isn't needed but won't stop a test passing. Either way, have removed it here 👍

abhi-markan (approved):

e2e-tests/tfm/cypress/e2e/journeys/deal-cancellation/reason-for-cancelling.spec.js
ttbarnes: Thank you @BethThomas141 , this seems very strange as in another PR, `.clear()` stopped the test from passing, 

`cy.keyboardInput` (correctly) fails if a selector is not passed, and `.clear()` is not a selector, rather it's a method on a selector

I would like to run the test locally with `.clear()` to see if it passes but i'm currently unable to do so.

Anyway, i've removed this in another PR

PR #3582, opened 2024-09-26
feat(FN-2311): refactor existing premium payments UI logic

sw-mattw:
## Introduction :pencil2:
At present, the utilisation report reconciliation controller only has filtering logic for the premium payments tab. The naming of filters (such as `facility id`) does not include reference to the associated tab, neither do the input fields.

This ticket introduces filtering for the payment details tab, and the `facility id` filter also applies to this tab. We require unique naming for these filters to associate them with a particular tab. 

- Naming:
  - Added `premiumPayments` prefix to `facilityId` filter naming.
    - Including changing the `facilityIdQuery` query parameter naming to `premiumPaymentsFacilityId` (as we'll also have a `facility id` filter on the payment details tab).
  - Changed the naming of the existing `premium payments` tab errors to include reference to this tab (which is also the container they're rendered within).
- Made `validateFacilityIdQuery` take in the anchor href to the `facility id` input to allow reuse of this function when validating the same input on the payment details tab.
- Encapsulated the `premium payments` filters in an object for readability (in line with the approach taken by other PRs for this ticket for the multiple `payment details` filters).
- Pulled out parsing of the `premium payments` filters into a separate function (`parsePremiumPaymentsFilters`) to aid readability.

No visual changes have been made, only name and small logic improvements (above) have been made.


RLavender98 (dismissed):
LGTM!

trade-finance-manager-ui/server/controllers/utilisation-reports/utilisation-report-reconciliation-for-report/extract-query-and-session-data.test.ts
```diff
@@ -7,59 +7,61 @@ jest.mock('./validate-facility-id-query');
...
   const ORIGINAL_URL = '/original-url';
-  const FACILITY_ID_QUERY = '1234';
+  const PREMIUM_PAYMENTS_FACILITY_ID_QUERY = '1234';
 
   beforeEach(() => {
     jest.resetAllMocks();
 
     jest.mocked(handleRedirectSessionData).mockReturnValue({
-      tableDataError: undefined,
+      premiumPaymentsTableDataError: undefined,
```
ttbarnes, line 16: Instead of setting an undefined value, can we declare `premiumPaymentsTableDataError` as an optional type?

trade-finance-manager-ui/server/controllers/utilisation-reports/utilisation-report-reconciliation-for-report/extract-query-and-session-data.test.ts
```diff
@@ -7,59 +7,61 @@ jest.mock('./validate-facility-id-query');
...
-    // Arrange
-    const queryParams = {
-      facilityIdQuery: FACILITY_ID_QUERY,
-    };
-    const sessionData = {};
+  describe('premium payments filters', () => {
+    it('uses provided premiumPaymentsFacilityId', () => {
+      // Arrange
+      const queryParams = {
+        premiumPaymentsFacilityId: PREMIUM_PAYMENTS_FACILITY_ID_QUERY,
```
ttbarnes, line 25: What's the reason for renaming the query param please? I think `facilityId` is simpler and cleaner than `premiumPaymentsFacilityId`.

I'm don't know the URL off the top of my head, but i'm imagining with the current changes it will be something like this:

`dtfs.com/utlisation-report-reconciliation?premiumPaymentsFacilityId=1234,5678`

Where IMO this would be cleaner:

`dtfs.com/utlisation-report-reconciliation?facilityIds=1234,5678`

trade-finance-manager-ui/server/controllers/utilisation-reports/utilisation-report-reconciliation-for-report/extract-query-and-session-data.test.ts
```diff
@@ -91,12 +93,12 @@ describe('extractQueryAndSessionData', () => {
...
     // Arrange
     const queryParams = {
-      facilityIdQuery: FACILITY_ID_QUERY,
+      premiumPaymentsFacilityId: PREMIUM_PAYMENTS_FACILITY_ID_QUERY,
     };
     const sessionData = {};
 
     jest.mocked(handleRedirectSessionData).mockReturnValue({
-      tableDataError: undefined,
+      premiumPaymentsTableDataError: undefined,
```
ttbarnes, line 101: Same as above, can we have an optional type instead of having to pass `undefined` values around?

trade-finance-manager-ui/server/controllers/utilisation-reports/utilisation-report-reconciliation-for-report/extract-query-and-session-data.ts
```diff
@@ -46,9 +69,9 @@ export const extractQueryAndSessionData = (
   const isCheckboxChecked = getIsCheckboxChecked(selectedFeeRecordIds);
 
   return {
-    facilityIdQueryString,
-    filterError,
-    tableDataError,
+    premiumPaymentsFilters,
+    premiumPaymentsFilterError,
+    premiumPaymentsTableDataError,
```
ttbarnes, line 74: Similar to above, why do we want `premiumPayments` prefixes? We're already in the context of "reports", it seems like the original naming is simpler

In other words (and not for this PR), but I think e.g this:

```
controllers/utilisation-reports/utilisation-report-reconciliation-for-report/*
```

Could just be something like this:

```
controllers/utilisation-reports/reconciliation/*
```

I am not familiar with the functionality/URL etc, but I don't think we need to be as specific when in a child directory. What do you think?

trade-finance-manager-ui/server/controllers/utilisation-reports/utilisation-report-reconciliation-for-report/validate-facility-id-query.ts
```diff
@@ -2,26 +2,28 @@ import { isNonEmptyString } from '@ukef/dtfs2-common';
 import { ErrorSummaryViewModel } from '../../../types/view-models';
 import { REGEX } from '../../../constants';
 
-const FACILITY_ID_INPUT_ID = '#facility-id-filter';
-
-export const validateFacilityIdQuery = (facilityIdQuery: string | undefined, originalUrl: string): ErrorSummaryViewModel | undefined => {
-  if (originalUrl.includes('?') && originalUrl.includes('facilityIdQuery')) {
+export const validateFacilityIdQuery = (
+  facilityIdQuery: string | undefined,
```
ttbarnes, line 6:
```suggestion
  facilityIdQuery?: string,
```

trade-finance-manager-ui/server/controllers/utilisation-reports/utilisation-report-reconciliation-for-report/validate-facility-id-query.ts
```diff
@@ -2,26 +2,28 @@ import { isNonEmptyString } from '@ukef/dtfs2-common';
...
+  originalUrl: string,
+  facilityIdInputId: string,
+): ErrorSummaryViewModel | undefined => {
```
ttbarnes, line 9:
```suggestion
): ErrorSummaryViewModel? => {
```

ttbarnes (changes requested):

trade-finance-manager-ui/component-tests/utilisation-reports/utilisation-report-reconciliation-for-report.component-test.ts
```diff
@@ -61,192 +63,198 @@ describe(page, () => {
...
-    wrapper.expectElement(`${premiumPaymentsTabSelector} table[data-cy="premium-payments-table"]`).toExist();
-  });
-
-  it('should render the facility ID filter input', () => {
-    const wrapper = getWrapper();
-    wrapper.expectElement('[data-cy="facility-filter-input"]').toExist();
-    wrapper.expectText('[data-cy="facility-filter-form"]').toContain('Filter by facility ID');
-  });
+  describe('premium payments tab', () => {
+    it('should render the premium payments tab with headings, text, table and buttons', () => {
```
Zainzzkk, line 67: it would be nice to have a different it for each element being tested, eg it for header, it for text etc, but to keep it consistent is fine to leave it as is for now

trade-finance-manager-ui/component-tests/utilisation-reports/utilisation-report-reconciliation-for-report.component-test.ts
Zainzzkk, line 1: this will clash with my changes made

trade-finance-manager-ui/server/controllers/utilisation-reports/utilisation-report-reconciliation-for-report/extract-query-and-session-data.ts
```diff
@@ -1,42 +1,65 @@
...
 };
 
+/**
+ * Parses premium payments filters from the query parameters.
+ * @param premiumPaymentsFacilityId - The premium payments facility ID query.
+ * @param originalUrl - The original URL of the request.
+ * @returns An object containing the parsed premium payments filters and any
+ * error resulting from validating the facility ID query.
+ */
+const parsePremiumPaymentsFilters = (facilityIdQuery: string | undefined, originalUrl: string) => {
```
Zainzzkk, line 20: can this be optional (will need to change the order) but then will not need to be | undefined

trade-finance-manager-ui/server/middleware/validatePostAddPaymentRequestBody/index.ts
```diff
@@ -12,19 +12,19 @@ import { PremiumPaymentsTableCheckboxId } from '../../types/premium-payments-tab
 
 const isRequestBodyAnObject = (body: unknown): body is object => !body || typeof body === 'object';
 
-const FACILITY_ID_QUERY_REGEX = /facilityIdQuery=(?<facilityIdQuery>\d{4,10})/;
+const PREMIUM_PAYMENTS_FACILITY_ID_QUERY_REGEX = /premiumPaymentsFacilityId=(?<premiumPaymentsFacilityId>\d{4,10})/;
 
-const getFacilityIdQueryFromReferer = (req: Request): string | undefined => {
+const getPremiumPaymentsFacilityIdQueryFromReferer = (req: Request): string | undefined => {
```
Zainzzkk, line 17: documentation in this file please so it is easier to understand what is happening.  For the REGEX and functions

trade-finance-manager-ui/server/controllers/utilisation-reports/utilisation-report-reconciliation-for-report/extract-query-and-session-data.test.ts
sw-mattw: We have different tabs on this page, each with their own filters (some of which filter the same attributes). Previously we only had filters on the premium payments tab, but now we've also added them to to the payment details tab. Both sets of these filters are passed through as query parameters so the prefix has been added to distinguish which tab the filter applies to - hope that makes sense

Screenshot 2024-10-01 at 11 32 39

Screenshot 2024-10-01 at 11 32 32

trade-finance-manager-ui/server/controllers/utilisation-reports/utilisation-report-reconciliation-for-report/validate-facility-id-query.ts
sw-mattw: Thanks, updated this on another WIP branch and forgot to update here - just made this change

trade-finance-manager-ui/server/controllers/utilisation-reports/utilisation-report-reconciliation-for-report/validate-facility-id-query.ts
sw-mattw: I don't believe we can do this as it's on the return type, kept as is

trade-finance-manager-ui/server/controllers/utilisation-reports/utilisation-report-reconciliation-for-report/extract-query-and-session-data.ts
sw-mattw: I agree that it could be good for this directory to be renamed.

Regarding the `premiumPayments` prefixes, this is due to us having multiple tabs on this page, each with their own error view models - please see this comment https://github.com/UK-Export-Finance/dtfs2/pull/3582/files#r1782538503

trade-finance-manager-ui/server/controllers/utilisation-reports/utilisation-report-reconciliation-for-report/extract-query-and-session-data.test.ts
sw-mattw: Made this change changes, removed this from the stub as it's no longer required (as its now optional)

trade-finance-manager-ui/server/controllers/utilisation-reports/utilisation-report-reconciliation-for-report/extract-query-and-session-data.test.ts
ttbarnes: Great thanks for the explanation @sw-mattw makes sense 👍

trade-finance-manager-ui/server/middleware/validatePostAddPaymentRequestBody/index.ts
```diff
@@ -10,23 +10,47 @@ import {
...
-const FACILITY_ID_QUERY_REGEX = /facilityIdQuery=(?<facilityIdQuery>\d{4,10})/;
-
-const getFacilityIdQueryFromReferer = (req: Request): string | undefined => {
+/**
+ * Regular expression to extract the premium payments facility ID from a URL query string.
+ * The pattern uses {4,10} to match between 4 and 10 digits, ensuring the facility ID
+ * is at least 4 digits long but no more than 10 digits.
+ * The matched digits are captured in a named group 'premiumPaymentsFacilityId'.
+ */
+const PREMIUM_PAYMENTS_FACILITY_ID_QUERY_REGEX = /premiumPaymentsFacilityId=(?<premiumPaymentsFacilityId>\d{4,10})/;
```
ttbarnes, line 26: Are we able to partially use the existing `PARTIAL_FACILITY_ID` regex? Not sure if possible without having a go myself

ttbarnes (commented):

ttbarnes (approved):

trade-finance-manager-ui/component-tests/utilisation-reports/utilisation-report-reconciliation-for-report.component-test.ts
sw-mattw: Agree we should test these in smaller groups - leaving as is for now as this change is just to nest these in a describe block for readability

trade-finance-manager-ui/server/middleware/validatePostAddPaymentRequestBody/index.ts
```diff
@@ -10,23 +10,47 @@ import {
 } from '../../helpers/premium-payments-table-checkbox-id-helper';
 import { PremiumPaymentsTableCheckboxId } from '../../types/premium-payments-table-checkbox-id';
 
+/**
+ * Checks if the given value is an object.
+ * @param body - The value to check.
+ * @returns True if the value is an object, false otherwise.
+ */
 const isRequestBodyAnObject = (body: unknown): body is object => !body || typeof body === 'object';
```
Zainzzkk, line 18: this could be moved to a helper and used elsewhere as well and needs testing

ttbarnes: Wait - just spotted this. @Zainzzkk Didn't you already create a helper or use a lodash method for object checking? We should use the same thing here

Why do we need to check the request body is an object?

Zainzzkk: Mine was to check if 2 objects are the same so this is different

ttbarnes: OK thanks.

Seems this PR is just adding documentation to this function so not a blocker

Zainzzkk: Ah in that case can be left as is

Zainzzkk (dismissed):

trade-finance-manager-ui/component-tests/utilisation-reports/utilisation-report-reconciliation-for-report.component-test.ts
```diff
@@ -61,210 +63,200 @@ describe(page, () => {
...
-
-    wrapper.expectElement(`${premiumPaymentsTabSelector} table[data-cy="premium-payments-table"]`).toExist();
-  });
+      wrapper.expectElement(`${premiumPaymentsTabSelector} input[data-cy="add-a-payment-button"]`).toExist();
+      wrapper.expectElement(`${premiumPaymentsTabSelector} input[data-cy="add-a-payment-button"]`).toHaveAttribute('value', 'Add a payment');
+      wrapper
+        .expectElement(`${premiumPaymentsTabSelector} input[data-cy="add-a-payment-button"]`)
+        .toHaveAttribute('formaction', `/utilisation-reports/${reportId}/add-payment`);
 
-  it('should render the premium payments with the relevant text for non-PDC_RECONCILE users', () => {
```
RLavender98, line 102: Has this test been deleted by accident?

trade-finance-manager-ui/component-tests/utilisation-reports/utilisation-report-reconciliation-for-report.component-test.ts
```diff
@@ -61,210 +63,200 @@ describe(page, () => {
...
 
-    wrapper.expectText(`${keyingSheetTabSelector} h2[data-cy="keying-sheet-heading"]`).toRead('Keying sheet');
-    wrapper.expectText('[data-cy="select-payments-text"]').notToExist();
-    wrapper
-      .expectText('[data-cy="payments-on-premium-payments-tab-text"]')
-      .toMatch(/Payments on the premium payments tab will show as reconciled when they have been marked as done here./);
+      wrapper.expectText(`${keyingSheetTabSelector} h2[data-cy="keying-sheet-heading"]`).toRead('Keying sheet');
+      wrapper.expectText('[data-cy="select-payments-text"]').notToExist();
+      wrapper
+        .expectText(`${keyingSheetTabSelector} p`)
```
RLavender98: I think this selector has changed incorrectly in the merge

sw-mattw: Yep, spotted this too - pushed a change at the same time

trade-finance-manager-ui/component-tests/utilisation-reports/utilisation-report-reconciliation-for-report.component-test.ts
sw-mattw: Thanks, readded

ttbarnes (approved):

RLavender98 (approved):

PR #3583, opened 2024-09-26
chore(DTFS2-7394): refactor facility end date feature flag in e2e-tests/tfm

MarRobSoftwire:
## Introduction :pencil2:
We want to refactor how ffs are tested to remove the if statements from tests.

update test.yml and remove if statements from e2e-tests/tfm tests


.github/workflows/test.yml
```diff
@@ -821,7 +820,6 @@ jobs:
             './tfm/cypress/e2e/journeys/case-amendments/amendment-underwriting-page-multiple-amendments/*.spec.js',
             './tfm/cypress/e2e/journeys/case-amendments/amendments-version-order/*.spec.js',
             './tfm/cypress/e2e/journeys/case-amendments/facility-end-date/bss-deal-expect-no-facility-end-date-features/*.spec.js',
-            './tfm/cypress/e2e/journeys/case-amendments/facility-end-date/feature-flag-disabled/*.spec.js',
```
ttbarnes, line 824: Looks like this should remain, to cover this test?

```ts
e2e-tests/tfm/cypress/e2e/journeys/case-amendments/facility-end-date/feature-flag-disabled/gef-deal-expect-no-facility-end-date-features.spec.js
```

ttbarnes (changes requested):

.github/workflows/test.yml
MarRobSoftwire: That is a feature-flag-disabled test & will fail if the feature flag is on (now I've removed the if statement checking this)

ttbarnes: Ah sorry, didn't realise this is in the "FF enabled" suite

PR #3584, opened 2024-09-26
fix(FN-3460): fix invalid facility utilisation values in e2e report fixtures

sw-mattw:
## Introduction :pencil2:
Facility utilisation values for the same facility ID must match. The E2E report fixtures were updated and this wasn't the case - these need updating to the same value.

- Updated each of the 5 `valid_` spreadsheets to include the same facility utilisation value for the same facility ID. All files now have the same content (as previously, apart from the invalid utilisation values which differed) - see screenshot

Screenshot 2024-09-26 152400

Note that the two large values are to test the floating-point rounding issue from FN-3460.

N/A


Zainzzkk (approved):
please merge main in first

PR #3589, opened 2024-09-27
feat(FN-3272): store initial utilisation and fee based on supplied and facility values

Zainzzkk:
## Introduction :pencil2:
This PR saves the utilisation and fixed fee based on the provided utilisation report and stored facility values

* Updated report-uploaded.event-handler to call a helper to obtain utilisation and fixed fee values
* Added helpers to calculate fixed fee and utilisation
* Refactored files to use shared helpers
* Added documentation

* Updated optional variable for validate-rows
* Moved some types to libs/common
* Moved calculateDrawnAmount to libs/common


dtfs-central-api/src/helpers/calculate-fixed-fee-from-days-remaining.ts
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * calculateFixedFeeFromDaysRemaining
+ * Calculates the fixed fee for the given parameters and number of days provided
+ * @param param - The parameters to calculate the fixed fee with
+ * @param {Number} param.utilisation - The facility utilisation
+ * @param {Number} param.numberOfDaysRemainingInCoverPeriod - The number of days remaining in the cover period
+ * @param {Number} param.interestPercentage - The facility interest percentage
+ * @param {Number} param.dayCountBasis - The facility day count basis
+ * @returns {Number} The fixed fee for the current report period
```
ttbarnes, line 18: `@param {number}`?

Can't see any other instances of this in uppercase

Zainzzkk: changed to number

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/helpers/calculate-fixed-fee.test.ts
```diff
@@ -0,0 +1,42 @@
...
+      const today = new Date();
+      const future = addDays(today, difference);
+
+      const result = getNumberOfDaysInCoverPeriod(today, future);
+
+      expect(result).toEqual(difference);
+    });
+  });
+
+  describe('calculateFixedFee', () => {
```
ChristianMcc, line 19: Do we still have the tests that test the actual values of the calculateFixedFee with the raw numbers?

trade-finance-manager-api/src/v1/helpers/calculate-gef-facility-fee-record.js
```diff
@@ -1,3 +1,5 @@
+const { calculateDrawnAmount } = require('@ukef/dtfs2-common');
```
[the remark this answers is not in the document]
ChristianMcc: ah yes you do, ignore this!

dtfs-central-api/src/helpers/calculate-fixed-fee-from-days-remaining.ts
```diff
@@ -0,0 +1,34 @@
+import Big from 'big.js';
+import { CalculateFixedFeeFromDaysRemainingParams } from '@ukef/dtfs2-common';
+
+/**
+ * An admin fee (fixed at 10%) is applied to the fixed fee, meaning
+ * we subtract this amount when calculating it
+ */
+const BANK_ADMIN_FEE_ADJUSTMENT = 0.9;
```
ttbarnes, line 8: Why does the documentation say 10% when the const is `0.9`?

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/helpers/calculate-fixed-fee.test.ts
```diff
@@ -0,0 +1,42 @@
...
+    it('should return the value of "calculateFixedFeeFromDaysRemaining"', () => {
+      const utilisation = 100000;
+      const interestPercentage = 5;
+      const dayCountBasis = 365;
+      const coverStartDate = new Date();
+      const coverEndDate = addDays(coverStartDate, 368);
```
ttbarnes, line 25: I'm curious why this is 368 and not e.g 1?

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/helpers/calculate-initial-utilisation-and-fixed-fee.ts
```diff
@@ -0,0 +1,58 @@
...
+ * @param coverStartDate
+ * @param coverEndDate
+ * @returns true if all values are present and in the correct format
+ */
+export const hasRequiredValues = (
+  value: number,
+  interestPercentage: number,
+  dayCountBasis: number,
+  coverStartDate: string | Date | null,
+  coverEndDate: string | Date | null,
```
ttbarnes, line 20: Would this work? Ignore if not

```ts
coverStartDate?: Date,
coverEndDate?: Date,
```

ttbarnes, line 20: Would this work? Ignore if not

```ts
coverStartDate?: Date,
coverEndDate?: Date,
```

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/helpers/calculate-initial-utilisation-and-fixed-fee.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  if (hasRequiredValues(value, interestPercentage, dayCountBasis, coverStartDate, coverEndDate)) {
+    throw new Error(`TFM facility values for ${facilityId} are missing`);
+  }
+
+  const utilisation = calculateInitialUtilisation(value);
+  const fixedFee = calculateFixedFee({
+    utilisation,
+    coverStartDate: coverStartDate as Date,
+    coverEndDate: coverEndDate as Date,
```
ttbarnes, line 49: Do we have to do this, shouldn't the dates already have a `Date` type?

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/report-uploaded/report-uploaded.event-handler.test.ts
```diff
@@ -99,15 +105,20 @@ describe('handleUtilisationReportReportUploadedEvent', () => {
...
         id: facilityId,
         reportPeriod: reportReportPeriod,
         requestSource,
+        utilisation,
+        fixedFee,
       });
     });
-    expect(mockSave).toHaveBeenCalledWith(FacilityUtilisationDataEntity, createdFacilityUtilisationDataEntities, { chunk: 100 });
+
+    expect(mockSave).toHaveBeenCalledWith(FacilityUtilisationDataEntity, await Promise.all(createdFacilityUtilisationDataEntities), { chunk: 100 });
```
ttbarnes, line 122: Could we use an `expected` here?

libs/common/src/helpers/date-validation.test.ts
```diff
@@ -239,4 +239,27 @@ describe('date-validation helpers', () => {
...
+    });
+
+    describe('when date is invalid', () => {
+      it('should return false', () => {
+        const result = isDate('invalid date');
+        expect(result).toEqual(false);
+      });
+    });
+
+    describe('when date is null', () => {
```
ttbarnes, line 258: Can we add a test for undefined/empty param since the type is `any`?

libs/common/src/helpers/facility-calculations/index.test.ts
```diff
@@ -0,0 +1,26 @@
+import { calculateDrawnAmount, calculateInitialUtilisation } from '.';
+
+describe('helpers/facility-calculation', () => {
```
ttbarnes, line 3: calculations*

libs/common/src/helpers/facility-calculations/index.test.ts
```diff
@@ -0,0 +1,26 @@
...
+    it('should return correct calculation', () => {
+      const result = calculateDrawnAmount(mockFacilityValue, mockCoverPercentage);
+
+      const expected = calculateInitialUtilisation(mockFacilityValue) * (mockCoverPercentage / 100);
+
+      expect(result).toEqual(expected);
+    });
+  });
+
+  describe('calculateInitialUtilisation', () => {
```
ttbarnes, line 17: Can we move this to the top so it's the same as the index?

libs/common/src/sql-db-entities/facility-utilisation-data/facility-utilisation-data.entity.ts
```diff
@@ -75,15 +71,19 @@ export class FacilityUtilisationDataEntity extends AuditableBaseEntity {
...
    * @returns The created entity
    */
-  public static createWithoutUtilisationAndFixedFee({
+  public static createWithUtilisationAndFixedFee({
     id,
     reportPeriod,
     requestSource,
-  }: CreateFacilityUtilisationDataWithoutUtilisationParams): FacilityUtilisationDataEntity {
+    utilisation,
+    fixedFee,
```
ttbarnes, line 79: Can we update the documentation please

ttbarnes (changes requested):

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/helpers/calculate-initial-utilisation-and-fixed-fee.ts
Zainzzkk: it does not work as ? is undefined and it is expecting null

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/helpers/calculate-initial-utilisation-and-fixed-fee.ts
Zainzzkk: Yep it's needed or linting errors

dtfs-central-api/src/helpers/calculate-fixed-fee-from-days-remaining.ts
Zainzzkk: I am not sure, I moved this from an existing file

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/helpers/calculate-fixed-fee.test.ts
Zainzzkk: yes we do in dtfs-central-api\src\services\state-machines\fee-record\event-handlers\helpers\calculate-fixed-fee.test.ts and the new test I have added

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/helpers/calculate-fixed-fee.test.ts
Zainzzkk: I chose a random number

ttbarnes: I read this like 368 or more than 365 was required for some particular logic or something. Can we use e.g 1 instead? So the `coverEndDate` is just `coverStartDate` + 1

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/report-uploaded/report-uploaded.event-handler.test.ts
```diff
@@ -99,15 +105,22 @@ describe('handleUtilisationReportReportUploadedEvent', () => {
...
         id: facilityId,
         reportPeriod: reportReportPeriod,
         requestSource,
+        utilisation,
+        fixedFee,
       });
     });
-    expect(mockSave).toHaveBeenCalledWith(FacilityUtilisationDataEntity, createdFacilityUtilisationDataEntities, { chunk: 100 });
+
+    const expected = await Promise.all(createdFacilityUtilisationDataEntities);
```
ttbarnes, line 122: I think `expectedEntities` would be better here

dtfs-central-api/src/helpers/calculate-fixed-fee-from-days-remaining.test.ts
```diff
@@ -0,0 +1,30 @@
...
+    // Act
+    const result = calculateFixedFeeFromDaysRemaining({
+      utilisation,
+      numberOfDaysRemainingInCoverPeriod,
+      interestPercentage,
+      dayCountBasis,
+    });
+
+    // Assert
+    expect(result).toBe(5671.23); // 100000 * (5 / 100) * 0.9 * 4603 / 365 = 5,671.2328767... = 5671.23 (ROUNDED)
```
ttbarnes, line 28: - `.toEqual`
- `expected`

Can we move the comment into it's own line - also wondering if we should use some variables instead of the numbers, to understand the calculation you have to go back and forth between the variables and the numbers

Instead of having a hardcoded number, can we/should we use methods like in the actual function?

RLavender98: We shouldn't use methods like in the actual function because that is duplicating the implementation and means the test would be verifying we can write the same code twice, rather than giving us confidence that the calculation is correct.

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/helpers/calculate-fixed-fee.ts
```diff
@@ -0,0 +1,36 @@
+import { differenceInDays } from 'date-fns';
+import { CalculateFixedFeeUtilisationReportParams } from '@ukef/dtfs2-common';
+import { calculateFixedFeeFromDaysRemaining } from '../../../../../helpers/calculate-fixed-fee-from-days-remaining';
+
+/**
+ * getNumberOfDaysInCoverPeriod
+ * calculates number of days in the cover period between the cover start and end dates
+ * @param coverStartDate
+ * @param coverEndDate
+ * @returns {Number} number of days in the cover period
```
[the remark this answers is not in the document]
RLavender98: Can we actually remove this type from the docstring instead please, we shouldn't use types in docstrings in ts files

dtfs-central-api/src/helpers/calculate-fixed-fee-from-days-remaining.test.ts
RLavender98, line 28: Where does 4603 come from?

dtfs-central-api/src/helpers/calculate-fixed-fee-from-days-remaining.test.ts
```diff
@@ -0,0 +1,30 @@
...
+    const utilisation = 100000;
+    const interestPercentage = 5;
+    const dayCountBasis = 365;
+
+    const coverStartDateBeforeReportPeriod = subMonths(new Date(), 1);
+
+    // Number of days remaining = 432
+    const coverEndDate = addDays(startOfMonth(new Date()), 432);
+
+    const numberOfDaysRemainingInCoverPeriod = getNumberOfDaysInCoverPeriod(coverStartDateBeforeReportPeriod, coverEndDate);
```
RLavender98, line 17:
```suggestion
    const numberOfDaysRemainingInCoverPeriod = 432;
```

dtfs-central-api/src/helpers/calculate-fixed-fee-from-days-remaining.test.ts
Zainzzkk: typo

Zainzzkk: changed to toEqual and moved to its own line

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/helpers/calculate-fixed-fee.ts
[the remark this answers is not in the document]
Zainzzkk: thought I had removed this one

dtfs-central-api/src/helpers/calculate-fixed-fee-from-days-remaining.test.ts
ttbarnes: Pros and cons - currently the test assertion is effectively manually / hard coding the result of the implemention which uses `Big`. In which case why do we need to use `Big`?

I am happy for the current approach to be merged since we probably have this in all other Big related unit tests.

Zainzzkk: This is consistent with already existing tests and this was moved across.  We should update this later though

ChristianMcc: It's stay like this for now so all good etc., but I'd really like to not update this. Big is very important for catching the complete edge cases of how javascript sucks at doing calculations, and having the tiny chance of being 1p off is worth using Big for. If we have a use case of numbers that we know won't cause this edge effect all we are doing here is making sure that no changes happen impact the result of our calculations without us knowing. If we start using methods then we might run the risk of something changing in the calculation but the test not knowing.

I think all I'm getting to is why would we not want to have (at least one) test with this behaviour?

ttbarnes: Makes sense @ChristianMcc just I think that such tests read very messily. Should be able to read a line and understand what it's doing, this is not the case here:

```ts
expect(result).toBe(5671.23); // 100000 * (5 / 100) * 0.9 * 4603 / 365 = 5,671.2328767... = 5671.23 (ROUNDED)
```

I would expect to have e.g `expect(result).toEqual(expected)`. Or even better something like this:

```ts
const someSubCalculation = varA + varB;
const someOtherNumber = varC * varD;

const expected = someSubCalculation / someOtherNumber + etc
```

This part in particular is messy, hard to read and maintain.

```ts
100000 * (5 / 100) * 0.9 * 4603 / 365 = 5,671.2328767... = 5671.23 (ROUNDED)
```

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/helpers/calculate-initial-utilisation-and-fixed-fee.test.ts
```diff
@@ -0,0 +1,89 @@
...
+      expect(result).toEqual(true);
+    });
+
+    it('should return false if a value is missing', () => {
+      const result = hasRequiredValues(1, 2, 3, new Date(), null);
+      expect(result).toEqual(false);
+    });
+
+    it('should return false if a date is the wrong format', () => {
+      const result = hasRequiredValues(1, 2, 3, 'a', new Date());
```
ttbarnes, line 20: Should we have `hasRequiredValues` with object structured params so it's less ambiguous when calling?

```ts
hasRequiredValues({
  value: 1,
  interestPercentage: 2,
  dayCountBasis: 3,
  coverStartDate: 'a',
  coverEndDate: new Date(),
});
```

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/helpers/calculate-initial-utilisation-and-fixed-fee.test.ts
```diff
@@ -0,0 +1,89 @@
...
+import { calculateFixedFee } from './calculate-fixed-fee';
+
+describe('helpers/calculate-initial-utilisation-and-fixed-fee', () => {
+  describe('hasRequiredValues', () => {
+    it('should return true if all values are present', () => {
+      const result = hasRequiredValues(1, 2, 3, new Date(), new Date());
+      expect(result).toEqual(true);
+    });
+
+    it('should return false if a value is missing', () => {
```
ttbarnes, line 14: Can we add a test for each individual param missing? i.e

```ts
when value is not provided
when interestPercentage is not provided
when dayCountBasis is not provided
when coverStartDate is not provided
when coverEndDate is not provided
```

libs/common/src/helpers/date-validation.ts
```diff
@@ -214,3 +214,17 @@ export const applyStandardValidationAndParseDateInput = (inputtedDate: DayMonthY
...
+ * @returns if the provided date is an instance of Date
+ */
+export const isDate = (date: any): boolean => date instanceof Date && !Number.isNaN(date.getTime());
+
+/**
+ * checks if provided value is a date and is not invalid
+ * @param date - provided date
+ * @returns if the provided date is an instance of Date and is not invalid
+ */
+export const isValidDate = (date: any): boolean => date instanceof Date && !Number.isNaN(date.getTime());
```
ttbarnes, line 230: `isDate` and `isValidDate` are doing the same thing, can we just have `isValidDate`?

Zainzzkk: I am not sure if there will be any issues with adding `Number.isNaN(date.getTime())` to the existing functionality so have added a second function where I know it is required

ttbarnes: I don't understand, they're both doing exactly the same thing, we don't need 2x instances of a function doing the same thing

Zainzzkk: I need to recommit, forgot to remove something 

`export const isDate = (date: any): boolean => date instanceof Date;` isDate should only be this without the !Number.isNaN(date.getTime())

dtfs-central-api/src/helpers/calculate-fixed-fee-from-days-remaining.test.ts
```diff
@@ -0,0 +1,29 @@
...
+    const result = calculateFixedFeeFromDaysRemaining({
+      utilisation,
+      numberOfDaysRemainingInCoverPeriod,
+      interestPercentage,
+      dayCountBasis,
+    });
+
+    // Assert
+    // utilisation * (interestPercentage / 100) * BANK_ADMIN_FEE_ADJUSTMENT * numberOfDaysRemainingInCoverPeriod / dayCountBasis = 5,326.02739... = 5,326.03 (ROUNDED)
+    const calculation = (utilisation * (interestPercentage / 100) * 0.9 * numberOfDaysRemainingInCoverPeriod) / dayCountBasis;
```
ChristianMcc, line 24: Can we just assert that it's equal to 5326.03 please as we know the answer? In this case I want the test to check it's giving the right answer, not re-write the implementation and then check the implementation in the controller is the same as the implementation written in the test, I just care about the end result

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/report-uploaded/report-uploaded.event-handler.ts
```diff
@@ -39,7 +42,16 @@ const createFacilityUtilisationDataEntityIfNotExists = async (
   if (entityExists) {
     return null;
   }
-  return FacilityUtilisationDataEntity.createWithoutUtilisationAndFixedFee({ id: facilityId, reportPeriod, requestSource });
+
+  const { utilisation, fixedFee } = await calculateInitialUtilisationAndFixedFee(facilityId);
+
+  return FacilityUtilisationDataEntity.createWithUtilisationAndFixedFee({
+    id: facilityId,
+    reportPeriod,
```
RLavender98, line 50: When we generate keying data we check that keying data for that report has not already been uploaded by checking that the report period on the utilisation data has not already been calculated (see [here](https://github.com/UK-Export-Finance/dtfs2/blob/b4ef04d7a9956a284bea44ae23f9468ec438122e/dtfs-central-api/src/services/state-machines/fee-record/event-handlers/helpers/calculate-fixed-fee-adjustment.ts#L25)).
So imagine if we upload a report with entries for new facilities, it would then throw errors when we generate the keying data later in the flow?
@ChristianMcc do you think we should just use the previous report period here instead? Also just noticed that check we do doesn't explicitly check for the previous report period, it just checks for 'not the current' report period which seems wrong since we want to have protection against reconciling reports out of order ideally.

ChristianMcc: Agreed on using the previous report period instead of current report period, that's what we should do for this.

Re previous report period work, I checked this with both Ian/Soma and Ruth and they have said they will never process them out of order so we don't need to cover that edge case (thankfully) :)

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/helpers/calculate-fixed-fee.ts
```diff
@@ -0,0 +1,36 @@
...
+ * @returns The fixed fee for the current report period
+ */
+export const calculateFixedFee = ({
+  utilisation,
+  coverStartDate,
+  coverEndDate,
+  interestPercentage,
+  dayCountBasis,
+}: CalculateFixedFeeUtilisationReportParams): number => {
+  const numberOfDaysRemainingInCoverPeriod = getNumberOfDaysInCoverPeriod(new Date(coverStartDate), new Date(coverEndDate));
```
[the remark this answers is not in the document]
Zainzzkk: Because for some reason, some dates which look like dates in the database are not dates and cause issues here

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/helpers/calculate-initial-utilisation-and-fixed-fee.ts
```diff
@@ -0,0 +1,69 @@
...
+ * @returns {Object} fixedFee and utilisation values
+ */
+export const calculateInitialUtilisationAndFixedFee = async (facilityId: string) => {
+  const tfmFacility = await TfmFacilitiesRepo.findOneByUkefFacilityId(facilityId);
+
+  if (!tfmFacility?.facilitySnapshot) {
+    throw new Error(`TFM facility ${facilityId} could not be found`);
+  }
+
+  const { value, coverStartDate, coverEndDate, interestPercentage, dayCountBasis } = tfmFacility.facilitySnapshot;
```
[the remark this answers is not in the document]
Zainzzkk: does this require a seperate ticket?

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/helpers/calculate-initial-utilisation-and-fixed-fee.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  if (!hasRequiredValues({ value, interestPercentage, dayCountBasis, coverStartDate, coverEndDate })) {
+    throw new Error(`TFM facility values for ${facilityId} are missing`);
+  }
+
+  const utilisation = calculateInitialUtilisation(value);
+  const fixedFee = calculateFixedFee({
+    utilisation,
+    coverStartDate: coverStartDate as Date,
+    coverEndDate: coverEndDate as Date,
```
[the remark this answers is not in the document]
Zainzzkk: it's because timestamp from the database could be `string | Date | null`.  Even if we use `convertTimestampToDate`, there is a linting error as it expects `string | number | Date`

libs/common/src/sql-db-entities/facility-utilisation-data/facility-utilisation-data.entity.ts
```diff
@@ -68,22 +64,28 @@ export class FacilityUtilisationDataEntity extends AuditableBaseEntity {
...
    * @param facilityUtilisationData - Parameters for creating the entity
    * @param facilityUtilisationData.id - The ID of the entity
    * @param facilityUtilisationData.reportPeriod - The report period details
    * @param facilityUtilisationData.requestSource - The source of the request
+   * @param facilityUtilisationData.utilisation - The utilisation value
+   * @param facilityUtilisationData.fixedFee - The fixed fee value
    * @returns The created entity
    */
-  public static createWithoutUtilisationAndFixedFee({
+  public static createWithUtilisationAndFixedFee({
```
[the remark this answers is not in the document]
Zainzzkk: removed

libs/common/src/helpers/facility-calculations/index.test.ts
```diff
@@ -0,0 +1,26 @@
...
+    });
+  });
+
+  describe('calculateDrawnAmount', () => {
+    it('should return correct calculation', () => {
+      const result = calculateDrawnAmount(mockFacilityValue, mockCoverPercentage);
+
+      const expected = calculateInitialUtilisation(mockFacilityValue) * (mockCoverPercentage / 100);
+
+      expect(result).toEqual(expected);
```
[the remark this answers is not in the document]
Zainzzkk: I do not agree with this.  We have the confidence from the `calculateInitialUtilisation` tests and this helper uses the returned value from that function.

dtfs-central-api/src/helpers/calculate-fixed-fee-from-days-remaining.test.ts
Zainzzkk: I have changed this.  I do not fully agree with this as if we change the mocks, then this test will fail and it will be hard to see why the test is failing for someone new.  Whereas with the old implementation, it will be resilient to mock changes but a change in function will cause it to fail (as the calculation will change)

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/helpers/calculate-initial-utilisation-and-fixed-fee.ts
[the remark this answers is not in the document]
ChristianMcc: I *think* we don't need to because the update that will be generated in the keying sheet report will be accurate for what we should update the value to. While we are generating this late all we want this to be is the 'initial' values that the facility would have at it's inception when it gets into ACBS for the first time

dtfs-central-api/src/helpers/calculate-fixed-fee-from-days-remaining.test.ts
ChristianMcc: Thanks, can completely see your point of view but for a calculation test like this I don't mind it being a bit brittle if it increases my confidence that the test is doing the exact right calculation and output . We have some other tests closer to what you've got so at least we have a mix to cover everything now :D

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/helpers/calculate-initial-utilisation-and-fixed-fee.ts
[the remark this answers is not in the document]
RLavender98: Ahh I see, that makes sense. I guess on the ticket about facility value changing and manual updates we have comments already about finding out if PDC would manually update fixed fee in that case and this is actually similar to that.

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/helpers/calculate-fixed-fee.ts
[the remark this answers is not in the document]
RLavender98: If we can have things which aren't dates here then we shouldn't be claiming they are dates in the function param typing

libs/common/src/helpers/facility-calculations/index.test.ts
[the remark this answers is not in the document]
RLavender98: The `* (mockCoverPercentage / 100)` is duplicating implementation rather than actually testing the correct value is returned. It would be better to mock the 'calculateInitialUtilisation` helper in this test so that we can make it return a set known value in this test and then we can make the test better as suggested.

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/helpers/calculate-initial-utilisation-and-fixed-fee.ts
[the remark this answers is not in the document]
RLavender98: If it can be `string` or `Date` or `null` you shouldn't do `as Date` since it will likely cause an error to be thrown if it isn't a `Date` down the line since all subsequent code will assume it is a `Date`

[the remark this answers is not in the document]
RLavender98: You could write the hasRequiredFields as a user defined type guard instead perhaps to resolve this in a safer way. I think that would look something like this (haven't tested this so might need some changes):
```
export type ValidatedRequiredParams = {
  value: number;
  interestPercentage: number;
  dayCountBasis: number;
  coverStartDate: string | Date;
  coverEndDate: string | Date;
};

export const hasRequiredValues = ({ value, interestPercentage, dayCountBasis, coverStartDate, coverEndDate }: RequiredParams) is ValidatedRequiredParams =>
  Boolean(
    value &&
      coverStartDate &&
      coverEndDate &&
      interestPercentage &&
      dayCountBasis &&
      isValidDate(new Date(coverStartDate)) &&
      isValidDate(new Date(coverEndDate)),
  );
```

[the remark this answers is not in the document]
Zainzzkk: It will always be Date given `hasRequiredValues` checks it is a valid date.  However, the linting error will keep occurring if I remove as Date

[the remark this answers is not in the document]
RLavender98: It won't necessarily be a `Date` because `isValidDate(new Date(params.coverStartDate))` just checks that it can be converted into a `Date`. It could very well be a string that _can_ be converted into a `Date`.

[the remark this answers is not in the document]
RLavender98: How about simply:
```suggestion
    coverStartDate: new Date(coverStartDate),
    coverEndDate: new Date(coverEndDate),
```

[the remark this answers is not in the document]
RLavender98: It isn't using `convertTimestampToDate` so we are being slightly inconsistent in how we read these values but it is at least safer as a compromise.

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/helpers/calculate-fixed-fee.ts
```diff
@@ -0,0 +1,36 @@
...
+ * gets the number of days in the cover period
+ * calculates and returns the fixed free from calculateFixedFeeFromDaysRemaining
+ * @param utilisation - The facility utilisation
+ * @param coverStartDate - The facility cover start date
+ * @param coverEndDate - The facility cover end date
+ * @param interestPercentage - The facility interest percentage
+ * @param dayCountBasis - The facility day count basis
+ * @returns The fixed fee for the current report period
+ */
+export const calculateFixedFee = ({
```
ChristianMcc, line 26: We already have this function in dtfs-central-api/src/services/state-machines/fee-record/event-handlers/helpers/calculate-fixed-fee.ts. 
If this is going to be the function that calculates initialFixedFee for a new facility can we call it something like that e.g. calculateInitialFacilityFixedFee?
We should probably update the documentation at the top as well to be 
`Calculates the fixed fee for a facility` or something as it's facility based not utilisation report based

Zainzzkk: changed the name

libs/common/src/helpers/facility-calculations/index.test.ts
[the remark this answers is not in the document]
ttbarnes: I agree with @Zainzzkk 

1. Hard coding is not a good practice.
2. `calculateInitialUtilisation` is tested.
3. `calculateDrawnAmount` is simply calling `calculateInitialUtilisation` with some certain values (2 of which are "dynamic" / from params).
4. We only need to test that `calculateInitialUtilisation` is calling `calculateDrawnAmount` with the expected params. That is exactly what the function/implementation does, I do not see why it should be tested any differently.
5. If the implementation changes, this test will correctly fail.
6. If we mock the helper and the implementation changes - depending on the mock, the test might still pass.
7. I don't see any reason to mock helpers in any unit test. I have never seen such a pattern in any project; With the expection for API/DB calls etc.

The current test is perfect IMO.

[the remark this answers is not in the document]
RLavender98: I disagree a few counter points:

1.  "We only need to test that  `calculateInitialUtilisation` is calling `calculateDrawnAmount `with the expected params." - This test isn't actually testing this, it is testing that the outcome of the function under test is _equal_ to if you had called this function then done some modifications to the result. If you want to test this function is _actually_ you could spy on it and check it is actually called.

2. `(mockCoverPercentage / 100)` is used in the test and in the function itself. If this part fo the calculation didn't do what was intended for some reason then the test would still pass. The point of a test is to fail if the function doesn't work as intended by the developer. 
3. "I don't see any reason to mock helpers in any unit test. I have never seen such a pattern in any project; With the expection of API/DB calls etc." - I've got the exact opposite experience, all the projects I have worked on have always in cases like this mocked the helper method instead of calling it to set up the expected values.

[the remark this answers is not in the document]
RLavender98: If that still doesn't convince you I will resolve the thread in the interest of this getting merged but I really disagree with this and would like us to come to a place of agreement for future work.

dtfs-central-api/src/helpers/get-previous-report-period.test.ts
```diff
@@ -0,0 +1,57 @@
...
+      const response = await getPreviousReportPeriod(bankId, utilisationReport);
+
+      const { start } = utilisationReport.reportPeriod;
+      const paddedMonth = String(start.month).padStart(2, '0');
+      const monthFormatted = `${start.year}-${paddedMonth}`;
+
+      const expected = getPreviousReportPeriodForBankScheduleByMonth(mockGetBankByIdResponse.utilisationReportPeriodSchedule, monthFormatted);
+
+      expect(response).toEqual(expected);
+    });
```
RLavender98, line 55: In this function you are duplicating logic from the function under test and therefore not validating it. Instead we should do something like this:
```suggestion
    it('should return the result of getPreviousReportPeriodForBankScheduleByMonth', async () => {
    // Arrange
    const reportPeriod = {start: { month: 2, year: 2024 }, end: { month: 4, year 2024 }};
    const utilisationReport = UtilisationReportEntityMockBuilder.forStatus('PENDING_RECONCILIATION').withReportPeriod(reportPeriod).build();
    
    	// Act
      const result = await getPreviousReportPeriod(bankId, utilisationReport);

			// Assert
      const expected = getPreviousReportPeriodForBankScheduleByMonth(mockGetBankByIdResponse.utilisationReportPeriodSchedule, '2024-02');

      expect(response).toEqual(expected);
    });
```

dtfs-central-api/src/helpers/get-previous-report-period.test.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  describe('when a bank is found', () => {
+    beforeEach(() => {
+      jest.mocked(getBankById).mockImplementation(jest.fn().mockResolvedValueOnce(mockGetBankByIdResponse));
+    });
+
+    it('should return the result of getPreviousReportPeriodForBankScheduleByMonth', async () => {
+      const response = await getPreviousReportPeriod(bankId, utilisationReport);
+
+      const { start } = utilisationReport.reportPeriod;
```
RLavender98, line 48: Please can we do tests for report periods with single digit start months and two digit start months please?

RLavender98 (changes requested):
Just a couple of comments on the new tests

dtfs-central-api/src/helpers/get-previous-report-period.test.ts
```diff
@@ -11,7 +11,7 @@ describe('getPreviousReportPeriod', () => {
     jest.resetAllMocks();
   });
 
-  const utilisationReport = UtilisationReportEntityMockBuilder.forStatus('PENDING_RECONCILIATION').build();
+  let utilisationReport = UtilisationReportEntityMockBuilder.forStatus('PENDING_RECONCILIATION').build();
```
RLavender98, line 14: We shouldn't use let here and then redeclare in the tests as then the value of the report will be different within each test depending on the order the tests are run.

dtfs-central-api/src/helpers/get-previous-report-period.test.ts
```diff
@@ -37,19 +37,49 @@ describe('getPreviousReportPeriod', () => {
...
-      const paddedMonth = String(start.month).padStart(2, '0');
-      const monthFormatted = `${start.year}-${paddedMonth}`;
+      const expected = getPreviousReportPeriodForBankScheduleByMonth(mockGetBankByIdResponse.utilisationReportPeriodSchedule, '2024-02');
 
-      const expected = getPreviousReportPeriodForBankScheduleByMonth(mockGetBankByIdResponse.utilisationReportPeriodSchedule, monthFormatted);
+      expect(response).toEqual(expected);
+    });
+  });
+
+  describe('when a bank is found and the month has 2 digits', () => {
```
RLavender98, line 56: Nitpick but for clarity given there are multiple months defined in the test:
```suggestion
  describe('when a bank is found and the start month of the report period has 2 digits', () => {
```

dtfs-central-api/src/helpers/get-previous-report-period.test.ts
```diff
@@ -37,19 +37,49 @@ describe('getPreviousReportPeriod', () => {
...
+  });
+
+  describe('when a bank is found and the month has 2 digits and starts with 0', () => {
+    beforeEach(() => {
+      jest.mocked(getBankById).mockImplementation(jest.fn().mockResolvedValueOnce(mockGetBankByIdResponse));
+    });
+
+    it('should return the result of getPreviousReportPeriodForBankScheduleByMonth', async () => {
+      const reportPeriod = { start: { month: 0o3, year: 2024 }, end: { month: 4, year: 2024 } };
+      utilisationReport = UtilisationReportEntityMockBuilder.forStatus('PENDING_RECONCILIATION').withReportPeriod(reportPeriod).build();
```
RLavender98, line 79:
```suggestion
      const reportPeriod = { start: { month: 3, year: 2024 }, end: { month: 4, year: 2024 } };
      utilisationReport = UtilisationReportEntityMockBuilder.forStatus('PENDING_RECONCILIATION').withReportPeriod(reportPeriod).build();
```

dtfs-central-api/src/helpers/get-previous-report-period.test.ts
```diff
@@ -37,19 +37,49 @@ describe('getPreviousReportPeriod', () => {
...
+      const response = await getPreviousReportPeriod(bankId, utilisationReport);
+
+      const expected = getPreviousReportPeriodForBankScheduleByMonth(mockGetBankByIdResponse.utilisationReportPeriodSchedule, '2024-10');
+
+      expect(response).toEqual(expected);
+    });
+  });
+
+  describe('when a bank is found and the month has 2 digits and starts with 0', () => {
```
RLavender98, line 72:
```suggestion
  describe('when a bank is found and the start month of the report period is a single digit', () => {
```

Zainzzkk: this is testing it works when the date is 03 format

Zainzzkk: it is doing 03 basically (as it does not accept 03 written like that).  It probably would never but its a nice test just to make sure it works

libs/common/src/helpers/date-validation.ts
ttbarnes: Resolving this with a fresh comment

libs/common/src/helpers/date-validation.ts
```diff
@@ -214,3 +214,17 @@ export const applyStandardValidationAndParseDateInput = (inputtedDate: DayMonthY
...
+ * @returns if the provided date is an instance of Date
+ */
+export const isDate = (date: any): boolean => date instanceof Date;
+
+/**
+ * checks if provided value is a date and is not invalid
+ * @param date - provided date
+ * @returns if the provided date is an instance of Date and is not invalid
+ */
+export const isValidDate = (date: any): boolean => date instanceof Date && !Number.isNaN(date.getTime());
```
ttbarnes, line 230: Can we remove `isDate` and use `isValidDate` throughout? Feels strange to have both of these functions.

Currently, `isDate` could return an invalid date

dtfs-central-api/src/helpers/get-previous-report-period.test.ts
RLavender98: I'll approve since this is low stakes but I disagree, I believe this is an octal number and I don't see the point in using an octal number when we are never using octal numbers for report periods

RLavender98 (approved):

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/helpers/calculate-initial-utilisation-and-fixed-fee.test.ts
```diff
@@ -0,0 +1,218 @@
...
+    });
+
+    it('should return false if the dayCountBasis is not provided', () => {
+      const params = {
+        value: 1,
+        interestPercentage: 2,
+        dayCountBasis: null,
+        coverStartDate: new Date(),
+        coverEndDate: new Date(),
+      };
```
ttbarnes, line 95: Could we create a `baseParams` const or something so that each test is shorter/one line params? E.g for this test:

```ts
 const params = { ...validParams, dayCountBasis: null };
```

ttbarnes: Could even just do this:

```ts
const result = hasRequiredValues({ ...params, dayCountBasis: null });
```

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/helpers/calculate-initial-utilisation-and-fixed-fee.ts
```diff
@@ -0,0 +1,82 @@
+import { calculateInitialUtilisation, isValidDate } from '@ukef/dtfs2-common';
+import { TfmFacilitiesRepo } from '../../../../../repositories/tfm-facilities-repo';
+import { calculateInitialFixedFee } from './calculate-initial-fixed-fee';
+
+export type RequiredParams = {
+  value?: number | null;
+  interestPercentage?: number | null;
+  dayCountBasis?: number | null;
+  coverStartDate?: string | Date | null;
+  coverEndDate?: string | Date | null;
```
ttbarnes, line 10: Do we need to do `number | null;` and `string | Date | null;`, since they're optional - would the following work? E.g

```
value?: number;
coverStartDate?: string | Date;
```

Zainzzkk: doesn't work as it gets `string | Date | null` from the facility and ? is undefined so causes a linting error

libs/common/src/helpers/date-validation.ts
Zainzzkk: removed

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/helpers/calculate-initial-utilisation-and-fixed-fee.ts
ttbarnes: Would be nice to refactor so that it returns something more consistent rather than 3x different optional types; One for another day...

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/helpers/calculate-initial-utilisation-and-fixed-fee.test.ts
Zainzzkk: done

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/helpers/calculate-initial-utilisation-and-fixed-fee.test.ts
```diff
@@ -0,0 +1,176 @@
...
+
+    it('should return false if the value is not provided', () => {
+      baseParams.value = null;
+
+      const result = hasRequiredValues(baseParams);
+      expect(result).toEqual(false);
+    });
+
+    it('should return false if the interestPercentage is not provided', () => {
+      baseParams.interestPercentage = null;
```
ttbarnes, line 70: Doing this in each test isn't working as expected, e.g in this test, `value` is also null, so this test for "if the interestPercentage is not provided" is actually testing "if value _and_ interestPercentage is not provided"

Need to do this so it only nullifies X field in Y test specifically

```ts
const result = hasRequiredValues({ ...params, dayCountBasis: null });
```

Or this, but I prefer the one liner.

```ts
const params = { ...validParams, dayCountBasis: null };
const result = hasRequiredValues(params);
```

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/helpers/calculate-initial-utilisation-and-fixed-fee.test.ts
ttbarnes: Resolving in place of a new comment

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/helpers/calculate-initial-utilisation-and-fixed-fee.ts
```diff
@@ -0,0 +1,82 @@
...
+};
+
+/**
+ * parses date to the correct format
+ * date could be passed as string or different type of timestamp
+ * @param date
+ * @returns parsed date in correct format
+ */
+export const parseDate = (date: string | Date | number | null): Date => {
```
ttbarnes, line 19: Not a blocker for this PR. With this, we now have 3x `parseDate` functions and they are all different 🙃

One for another day.

ttbarnes (approved):
LGTM

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/helpers/calculate-initial-utilisation-and-fixed-fee.ts
```diff
@@ -0,0 +1,40 @@
...
+export const calculateInitialUtilisationAndFixedFee = async (facilityId: string) => {
+  const tfmFacility = await TfmFacilitiesRepo.findOneByUkefFacilityId(facilityId);
+
+  if (!tfmFacility?.facilitySnapshot) {
+    throw new Error(`TFM facility ${facilityId} could not be found`);
+  }
+
+  const { value, coverStartDate, coverEndDate, interestPercentage, dayCountBasis } = tfmFacility.facilitySnapshot;
+
+  if (!value || !coverStartDate || !coverEndDate || !interestPercentage || !dayCountBasis) {
```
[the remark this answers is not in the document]
Zainzzkk: this has been done

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/helpers/calculate-initial-utilisation-and-fixed-fee.ts
Zainzzkk: yes, have the same issue with isDate, need to all be moved to common another day

PR #3590, opened 2024-09-27
fix(DTFS2-6616): amend facility fixed fee record orchestrator function

abhi-markan:
## Introduction :pencil2:
* TODO

* TODO

* Documentation improvement.


azure-functions/acbs-function/helpers/mandatoryFields.js
```diff
@@ -1,5 +1,12 @@
-const findMissingMandatory = (data, mandatoryFields) =>
-  mandatoryFields.map((mf) => (data[mf] !== null || data[mf] !== undefined || data[mf] !== '' ? '' : mf)).filter((mf) => mf !== '');
+/**
+ * Finds missing mandatory fields in the provided data object.
+ *
+ * @param {Object} data - The data object to check for missing fields.
```
AlexBramhill, line 4:
```suggestion
 * @param {object} data - The data object to check for missing fields.
```

azure-functions/acbs-function/helpers/mandatoryFields.js
```diff
@@ -1,5 +1,12 @@
...
+ * @param {Array<string>} mandatoryFields - An array of field names that are mandatory.
+ * @returns {Array<string>} - An array of missing mandatory field names. If all mandatory fields are present, returns an empty array.
+ */
+
+const findMissingMandatory = (data, mandatoryFields) => mandatoryFields.map((field) => (data[field] ? false : field)).filter((field) => field);
```
Zainzzkk, line 9: not sure what others think, worth splitting this into a 2 functions so easier to read/understand?

AlexBramhill, line 9: This fails on falsy values -- ie
```
aTemplate = ['totalToPayThisMonth', 'shouldDoSomethingImportant']
```

if we gave it the following:
```
{
    totalToPayThisMonth: 0,
    shouldDoSomethingImportant: false
}
```

it will return that the fields are missing.


Working example:
```
const findMissingMandatory = (data, mandatoryFields) => mandatoryFields.map((field) => (data[field] ? false : field)).filter((field) => field);

const data = {
    tester: 0,
    tester2: false
}

console.log(findMissingMandatory(data, ['tester', 'tester2', 'tester3']));
```

azure-functions/acbs-function/helpers/retry.js
```diff
@@ -0,0 +1,6 @@
+const df = require('durable-functions');
```
AlexBramhill, line 1: Any reason for the rename here? It feels like `retryOptions` is more specific to me ie if I read the below, I would assume that `retry` was the method to retry, not the `retryOptions`?
```
      const facilityBuyerTask = context.df.callActivityWithRetry('create-facility-guarantee', retry, {
        facilityIdentifier,
        acbsFacilityGuaranteeInput: acbsFacilityBuyerGuaranteeInput,
      });
```

libs/common/src/constants/retry.ts
```diff
@@ -0,0 +1,13 @@
+// Durable orchestrated functions
```
AlexBramhill, line 1: If this is only used for durable orchestrated functions, we could just keep them in there maybe? or are we going to need these elsewhere?

I'd also say if these are specific durable orchestrated functions configuration, that we follow the same pattern as other files like `deal-submission-type` for instance to provide some context to them -- Maybe something like:

```
const DURABLE_ORCHESTRATED_FUNCTIONS = {
    RETRY_OPTIONS: {
        INTERVAL_MS: 5000;
        MAX_RETRY: 5
    }
} as const
```

Then we can remove the comment at the top of the file and have it self documenting in the constant's name? I'm a little cautious of adding a word/concept so widely used such as `retry` without an explicit name as to where it's being used if it's not used everywhere (and even then I'd probably want it wrapped in `API` or `APP` or a config file to denote where it's being used, as we can configure retry on both our app and the api calls we make).

utils/mock-data-loader/re-insert-mocks.js
```diff
@@ -53,7 +53,7 @@ const init = async () => {
   }
   await db.close();
 
-  console.info('✅ Insertion successfull');
+  console.info('✅ Insertion successful');
```
AlexBramhill, line 56: Thanks!

azure-functions/acbs-function/package.json
```diff
@@ -29,6 +29,7 @@
     "unit-test-ff": "jest --coverage --verbose --config=unit.ff.jest.config.js --passWithNoTests"
   },
   "dependencies": {
+    "@ukef/dtfs2-common": "1.0.0",
```
AlexBramhill, line 32: Will this add a lot of bloat to the azure functions node_modules?

azure-functions/acbs-function/mappings/facility/facility-fixed-fee-amend.js
```diff
@@ -0,0 +1,42 @@
...
+const { FACILITY } = CONSTANTS;
+
+const facilityFixedFeeAmend = (amendments) => {
+  try {
+    const records = [];
+    const { amendment } = amendments;
+    const { amount } = amendment;
+
+    // 1. UKEF Exposure
+    if (amount) {
```
AlexBramhill, line 13: I know this code is generally non-changable in terms of style due to the importance of keeping it all in sync, so I'll avoid commenting things for readability (for instance here, a guard clause would remove a level of complexity)

azure-functions/acbs-function/mappings/facility/facility-loan-amend.js
```diff
@@ -4,16 +4,28 @@ const helpers = require('./helpers');
 
 const { FACILITY } = CONSTANTS;
 
+/**
+ * Processes amendments to a facility loan record.
+ *
+ * @param {Object} amendments - The amendments to be applied.
```
AlexBramhill, line 10: These should be lower case `object` I believe (all of these)

azure-functions/acbs-function/src/functions/acbs-amend-facility-covenant-record.js
```diff
@@ -24,9 +24,26 @@
...
 
+/**
+ * Durable Orchestration Function (DOF) for amending a facility covenant record in the ACBS system.
+ *
+ * This function is triggered by an HTTP trigger function and performs the following operations:
+ * 1. Validates the input payload.
+ * 2. Maps the amendments to the facility covenant record.
+ * 3. Updates the facility covenant record in the ACBS system.
+ *
+ * @param {Object} context - The context object provided by Durable Functions.
```
AlexBramhill, line 38: Might be worth doing a find replace for `{Object}` to `{object}` :)

azure-functions/acbs-function/src/functions/acbs-amend-facility.js
```diff
@@ -77,21 +89,27 @@ df.app.orchestration('acbs-amend-facility', function* amendFacility(context) {
...
         facilityCovenantRecord: {
           error: incorrectFacilityStageErrorMessage,
         },
+        facilityGuaranteeRecord: {
+          error: incorrectFacilityStageErrorMessage,
+        },
+        facilityLoanRecord: {
+          error: incorrectFacilityStageErrorMessage,
+        },
+        facilityFixedFeeRecord: {
```
AlexBramhill, line 110: good catch!

azure-functions/acbs-function/src/functions/activity-update-facility-covenant.js
```diff
@@ -15,33 +15,23 @@
...
- * If the API request is successful, it returns an object containing the status, timestamps of when the request was sent and received, the data sent, and the data received from the API.
- * If the API request fails, it throws an error with details about the request and the error.
- * If the payload is not valid or does not contain all mandatory fields, it returns an object with the missing mandatory fields.
- * If any other error occurs, it logs the error and throws a new error.
- *
- * @param {object} payload - The payload containing the facilityIdentifier and acbsFacilityCovenantInput.
- * @param {string} payload.facilityIdentifier - The identifier of the facility.
- * @param {import('../../mappings/facility/facility-covenant-amend').MappedFacilityCovenantAmendment} payload.acbsFacilityCovenantInput - The acbsFacilityCovenantInput object containing the covenant details.
- * @returns {object} - An object containing the status, timestamps of when the request was sent and received, the data sent, and the data received from the API.
- * @throws {Error} - Throws an error if the payload is invalid, if the API request to generate the covenant ID fails, or if any other error occurs.
```
AlexBramhill, line 34: are we worriedabout the loss of typing?

azure-functions/acbs-function/src/functions/activity-update-facility-fixed-fee.js
```diff
@@ -0,0 +1,88 @@
...
+
+/**
+ * Handles the amendment of a facility fixed fee record in the ACBS system.
+ *
+ * This function performs the following operations:
+ * 1. Validates the input payload.
+ * 2. Checks for missing mandatory fields in the ACBS facility fixed fee input.
+ * 3. Submits the amendment to the ACBS system.
+ * 4. Handles the response from the ACBS system and returns the result.
+ *
```
AlexBramhill, line 33: We've removed the other typings/docs here and kept this, is there a reason for this?

azure-functions/acbs-function/api.js
```diff
@@ -154,7 +205,7 @@ const put = async (endpoint, payload, etag) => {
 /**
  * Invokes TFS PATCH endpoint
  * @param {string} endpoint TFS endpoint
- * @param {object} payload Payload object
+ * @param {Object} payload Payload object
```
ttbarnes, line 208: I think these should be lowercase - outside of this PR:

@param {Object} - 4 instances
@param {object} - 385 instances

azure-functions/acbs-function/helpers/mandatoryFields.js
ttbarnes: We have a mix/match with Array/array as well, take your pick for this PR I guess 😄 

```ts
@param {Array} - 46 instances
@param {array} - 14 instances
```

azure-functions/acbs-function/helpers/mandatoryFields.js
ttbarnes: IMO something like this would be easier to read

```ts
const mapMissingField = (field) => data[field] ? false : field;

const findMissingMandatoryFields = (data, mandatoryFields) => {
  const mapped = mandatoryFields.map((field) => mapMissingField(field))

  // I am not sure why we need this (?)
  const filtered = mapped.filter((field) => field);

  return filtered;
};
```

azure-functions/acbs-function/helpers/retry.js
ttbarnes: Agree I think `retryOptions` is clearer especially as it returns `new df.RetryOptions(...)`

azure-functions/acbs-function/mappings/facility/facility-fixed-fee-amend.js
```diff
@@ -0,0 +1,42 @@
+const CONSTANTS = require('../../constants');
+const { now } = require('../../helpers/date');
+
+const { FACILITY } = CONSTANTS;
+
+const facilityFixedFeeAmend = (amendments) => {
```
ttbarnes, line 6: Can we add some ACBS style documentation please

azure-functions/acbs-function/src/functions/acbs-amend-facility.js
```diff
@@ -77,21 +89,27 @@ df.app.orchestration('acbs-amend-facility', function* amendFacility(context) {
 
     if (!acceptableFacilityStage.includes(facilityStageCode)) {
       // Error upon unacceptable facility stage
-      const incorrectFacilityStageErrorMessage = `Facility ${facilityId} stage is ${facilityStageCode}, amendment will not be processed`;
+      const incorrectFacilityStageErrorMessage = `Facility ${facilityId} stage is ${facilityStageCode}, amendments are accepted for 07 stage.`;
```
ttbarnes, line 92: What is the meanint of "07" Seems to be an ACBS specific facility stage - wondering if it's worth adding to line 85 what this actually means rather than an ambiguous number

azure-functions/acbs-function/src/functions/activity-update-facility-master.js
```diff
@@ -68,15 +44,37 @@ const handler = async (payload) => {
     }
 
     const { facilityId, acbsFacilityMasterInput, updateType, etag } = payload;
-
+    const mandatoryFields = [
```
ttbarnes, line 47: What's the reason for moving this inside the function?

Seems that all other similar ACBS functions have this defined at the top, I think this is easier to read/consume/maintain and should be aligned

gef-ui/Dockerfile
```diff
@@ -8,7 +8,6 @@ ENV GITHUB_SHA=$GITHUB_SHA
 # Node setup
 WORKDIR /app
 COPY package.json .
-COPY package-lock.json .
```
ttbarnes, line 11: Nice

ttbarnes (changes requested):
Few comments, nothing major. Nice work especially around improved ACBS documentation

azure-functions/acbs-function/helpers/mandatoryFields.js
AlexBramhill: Should be `Array` I believe

azure-functions/acbs-function/api.js
AlexBramhill: I actually just looked this up due to the array question -- and I think we've made a mistake here, where it should be `Object` technically as objects arent primatives. But I think that's another PR thing

azure-functions/acbs-function/api.js
```diff
@@ -144,7 +148,11 @@ const post = async (endpoint, payload) => {
  * Invokes MDM POST endpoint
  * @param {string} endpoint MDM endpoint
  * @param {Object} payload Payload object
+<<<<<<< HEAD
  * @returns {Promise<object>} API response object
+=======
+ * @returns {Promise<Object>} API response object
+>>>>>>> main
```
ttbarnes, line 155: Merge conflict 😛

azure-functions/acbs-function/helpers/mandatoryFields.js
abhi-markan: Have updated all and merged to `main` under PR #3600

azure-functions/acbs-function/mappings/facility/facility-loan-amend.js
abhi-markan: Merged under PR #3600

azure-functions/acbs-function/package.json
abhi-markan: Yes it will in a final docker build image.

azure-functions/acbs-function/src/functions/acbs-amend-facility-covenant-record.js
abhi-markan: Merged under PR #3600

azure-functions/acbs-function/src/functions/activity-update-facility-fixed-fee.js
abhi-markan: Good point, have refactored and made docstring consistent across all DAF.

azure-functions/acbs-function/src/functions/activity-update-facility-master.js
abhi-markan: Good spot, have moved all `mandatoryFields` all inside function.
Reason for moving inside function is to ensure variables are declared in memory after passing payload validation.

libs/common/src/constants/retry.ts
abhi-markan: Good suggestion, have implemented in similar pattern as `deal-submission-type`.

azure-functions/acbs-function/helpers/mandatoryFields.js
ttbarnes: Bump @abhi-markan

azure-functions/acbs-function/src/functions/activity-create-deal-investor.js
```diff
@@ -27,7 +26,7 @@ const handler = async (payload) => {
     }
 
     const { dealIdentifier, investor } = payload;
-
+    const mandatoryFields = ['effectiveDate', 'currency'];
```
ttbarnes, line 29: I think this remain at the top like other ACBS functions

ttbarnes: Actually just noticed these are being moved in other functions so all good I think - just want to make sure they're aligned.

azure-functions/acbs-function/src/functions/activity-update-facility-master.js
ttbarnes: Not a blocker - maybe one for another PR.

Some instances of `mandatoryFields` take up a lot of space in the functions, maybe we could extract into some ACBS constants or something e.g

```ts
REQUIRED_FIELDS.FUNCTION_X = [...];
REQUIRED_FIELDS.FUNCTION_Y = [...];
REQUIRED_FIELDS.FUNCTION_Y = null;
```

azure-functions/acbs-function/src/functions/activity-create-deal-investor.js
abhi-markan: Yep good point, they all are aligned with local scope.

azure-functions/acbs-function/src/functions/activity-update-facility-master.js
abhi-markan: Sounds good.

azure-functions/acbs-function/mappings/facility/code-value-transaction.js
```diff
@@ -5,6 +5,17 @@
...
 
+/**
+ * Generates a facility code value transaction record with default values.
+ *
+ * This function performs the following operations:
+ * 1. Sets default values for lender type code, initial bundle status code, and facility transaction code value code.
+ * 2. Returns the generated facility code value transaction record.
+ * 3. Handles any errors that occur during the process and logs them.
+ *
+ * @returns {Object|boolean} - An object containing the default values for the facility code value transaction record, or false if an error occurs.
```
AlexBramhill, line 16: This can me more specific with `* @returns { Object | false }` I think :)

azure-functions/acbs-function/mappings/facility/code-value-transaction.js
```diff
@@ -5,6 +5,17 @@
...
+ * @throws {Error} - Logs the error if any error occurs during the process.
```
AlexBramhill, line 17: This never throws due to the wrapping in a `try`/`catch` block that returns false :) I think if we're wanting to say we log errors as part of this function's error handling it might be better to add it to the description (fwiw I'm not sure there's much point)

azure-functions/acbs-function/mappings/facility/facility-covenant.js
```diff
@@ -10,6 +10,21 @@
...
+ * 1. Extracts the guarantee expiry date and effective date from the facility's guarantee dates.
+ * 2. Retrieves the currency of the facility, defaulting to a constant if not provided.
+ * 3. Calculates the maximum liability using a helper function.
+ * 4. Constructs and returns the facility covenant record object.
+ *
+ * @param {Object} deal - The deal object containing deal details.
+ * @param {Object} facility - The facility object containing facility details.
+ * @param {string} covenantType - The type of the covenant.
+ * @returns {Object} - The mapped facility covenant record, including covenant type, maximum liability, currency, guarantee expiry date, and effective date.
+ * @throws {Error} - Logs the error and returns an empty object if any error occurs during the mapping process.
```
AlexBramhill, line 26: I wont comment everywhere regarding the `@throws` -- but I do think we shouldn't say things throw an error when we're using other error handling (the return of an empty object) to handle errors as currently this would read as an Error object being thrown

azure-functions/acbs-function/mappings/facility/facility-loan.js
```diff
@@ -22,6 +22,35 @@ const date = require('../../helpers/date');
...
+ * This function performs the following operations:
+ * 1. Extracts the issue date and guarantee expiry date from the facility's guarantee dates.
+ * 2. Retrieves the currency of the facility, defaulting to a constant if not provided.
+ * 3. Calculates the maximum liability and other financial details using helper functions.
+ * 4. Constructs and returns the facility loan record object.
+ * 5. Adds additional fields if the facility is not in GBP.
+ * 6. Adds specific fields for BSS/EWCS deals.
+ *
+ * @param {Object} deal - The deal object containing deal details.
+ * @param {Object} deal.dealSnapshot - The snapshot of the deal details.
```
AlexBramhill, line 37: do we want to commonise these object definitions to make the code more maintainable and easier to read?

azure-functions/acbs-function/package.json
AlexBramhill: Are we still using libs common as it appears we're now using constants from the acbs folder?

it'd be quite nice to remove this, for micro benefits/spped gains when deploying. 

There's also the potential future issue regarding how the default libs common imports run at startup. As we are making use of `index` files, and a lot of constants that are evaluated at run time, we have to separate some imports, such as `mongo-db-client`, which would otherwise cause a failed startup. I'd like to avoid that issue being tied to the DF as this microservice should be really lean.

azure-functions/acbs-function/helpers/mandatoryFields.js
AlexBramhill, line 9: This still has the bug noted here I think: https://github.com/UK-Export-Finance/dtfs2/pull/3590/files#r1781405185

azure-functions/acbs-function/src/functions/activity-update-facility-covenant.js
AlexBramhill: I mean that the use of `imports` allow for reusable types -- so `acbsFacilityCovenantInput` is fully typed with intellisense etc and allows us to reuse this throughout (meaning we don't have to repeat definitions everywhere they're used)

azure-functions/acbs-function/src/functions/activity-create-facility-master.js
```diff
@@ -1,43 +1,41 @@
...
+ * Handles the creation of a facility master record in the ACBS system.
+ *
+ * This function performs the following operations:
+ * 1. Validates the input payload.
+ * 2. Checks for missing mandatory fields in the payload.
+ * 3. Submits the creation request to the ACBS system.
+ * 4. Handles the response from the ACBS system and returns the result.
  *
  * @param {Object} payload - The payload containing the mandatory fields for creating a facility master record.
+ * @param {string} payload.dealIdentifier - The identifier of the deal.
```
AlexBramhill, line 17: I think this is another example of why might want to use imported typed -- so we can avoid large walls of jsdocs like this

azure-functions/acbs-function/src/functions/activity-update-facility-covenant.js
```diff
@@ -1,56 +1,43 @@
...
     const submittedToACBS = getNowAsIsoString();
-    const { status, data } = await api.updateFacilityCovenant(facilityIdentifier, acbsFacilityCovenantInput);
+    const { status, data } = await api.updateFacilityCovenant(facilityId, acbsFacilityCovenantInput);
 
     if (isHttpErrorStatus(status)) {
       throw new Error(
         JSON.stringify(
           {
             name: 'ACBS Facility Covenant amend error',
-            facilityIdentifier,
```
AlexBramhill, line 53: is there a reason we dont want to have an identifier for the error? Would this not make it harder to track any side effects (if, for instance, there is only a partial write of information)

libs/common/src/constants/index.ts
```diff
@@ -24,5 +24,6 @@ export * from './url-character-allowances';
 export * from './max-character-count';
 export * from './deal-submission-type';
 export * from './floating-point-rounding-decimal-places';
+export * from './retry-options';
```
AlexBramhill, line 27: dont think we need any of this now (hopefully!)

AlexBramhill (changes requested):

PR #3597, opened 2024-10-01
feat(DTFS2-7296): update db on bank request date page

BethThomas141:
## Introduction :pencil2:
The bank request date page should update the database when a user presses 'continue'. It should also display previously entered values (e.g. if they press back)

I've followed the implementation precedent set on the deal cancellation page for consistency. 

- get bank request date from the database
- update bank request date
- validate deal exists & submission type is valid on both post and get endpoints


trade-finance-manager-ui/server/controllers/case/cancellation/validation/validate-bank-request-date.ts
```diff
@@ -46,7 +46,7 @@ export const validateBankRequestDate = (date: DayMonthYearInput): BankRequestDat
   }
 
   return {
-    bankRequestDate: parsedDate,
```
BethThomas141, line 49: If a user e.g. enters:
Day: 2
Month: 3
Year: 2024

I think it's sensible to store this as the start of the 2/3/2024 in the database, to make sure e.g. if they enter today's date, then the deal cancellation will have taken place immediately on submit. But happy to hear other thoughts here!

trade-finance-manager-ui/server/controllers/case/cancellation/bank-request-date.controller.get.test.ts
```diff
@@ -0,0 +1,140 @@
...
+jest.mock('../../../api', () => ({
+  getDeal: jest.fn(),
+  getDealCancellation: jest.fn(),
+}));
+
+const dealId = 'dealId';
+const ukefDealId = 'ukefDealId';
+const mockUser = aTfmSessionUser();
+
+describe('getBankRequestDate', () => {
```
BethThomas141, line 18: I've split the tests out into a `.get` and `.post` test file, which is why it looks like lots of new tests

trade-finance-manager-ui/server/controllers/case/cancellation/bank-request-date.controller.ts
```diff
@@ -17,19 +20,43 @@ export type PostBankRequestDateRequest = CustomExpressRequest<{
...
+      return res.redirect('/not-found');
+    }
+
+    if (!canSubmissionTypeBeCancelled(deal.dealSnapshot.submissionType)) {
+      return res.redirect(`/case/${_id}/deal`);
+    }
+
+    const cancellation = await api.getDealCancellation(_id, userToken);
+
+    const previouslyEnteredBankRequestDate = cancellation?.bankRequestDate ? new Date(cancellation.bankRequestDate) : undefined;
```
ttbarnes, line 40: Would this work?

```ts
const previouslyEnteredBankRequestDate = cancellation?.bankRequestDate && new Date(cancellation.bankRequestDate);
```

trade-finance-manager-ui/server/controllers/case/cancellation/bank-request-date.controller.ts
```diff
@@ -17,19 +20,43 @@ export type PostBankRequestDateRequest = CustomExpressRequest<{
...
+
+    const day = previouslyEnteredBankRequestDate ? format(previouslyEnteredBankRequestDate, 'd') : '';
+    const month = previouslyEnteredBankRequestDate ? format(previouslyEnteredBankRequestDate, 'M') : '';
+    const year = previouslyEnteredBankRequestDate ? format(previouslyEnteredBankRequestDate, 'yyyy') : '';
```
ttbarnes, line 44: Same comment as above

trade-finance-manager-ui/server/controllers/case/cancellation/validation/validate-bank-request-date.ts
ttbarnes: @BethThomas141 i'm not familiar with what happens with deal cancellation,

Feels strange to change the time of cancellation to be _before_ a user actually cancels the deal.

What happens if we leave the timestamp as is, instead of changing to start of the day?

ttbarnes (changes requested):
Nice work, few minor comments

trade-finance-manager-ui/server/controllers/case/cancellation/validation/validate-bank-request-date.ts
BethThomas141: That's a good point, I've changed it back to keep the timestamp to preserve the current time (so removed the `startOfDay` here)

trade-finance-manager-ui/server/controllers/case/cancellation/bank-request-date.controller.get.test.ts
```diff
@@ -0,0 +1,140 @@
...
+      // Assert
+      expect(res._getRenderView()).toEqual('case/cancellation/bank-request-date.njk');
+      expect(res._getRenderData() as BankRequestDateViewModel).toEqual({
+        activePrimaryNavigation: PRIMARY_NAVIGATION_KEYS.ALL_DEALS,
+        user: mockUser,
+        ukefDealId,
+        dealId,
+        day: '21',
+        month: '3',
+        year: '2024',
```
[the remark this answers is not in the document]
BethThomas141: Good idea, have changed :)

trade-finance-manager-ui/server/controllers/case/cancellation/bank-request-date.controller.ts
BethThomas141: Yes, have changed thanks!

trade-finance-manager-ui/server/controllers/case/cancellation/bank-request-date.controller.ts
BethThomas141: Have implemented above, but wouldn't work here - we explicitly want day/month/year to be an empty string to be passed in the nunjucks template in the case that there isn't a previously entered bank request date

ttbarnes: Makes sense, thanks 👍

ttbarnes (dismissed):

ttbarnes (approved):

PR #3600, opened 2024-10-01
fix(DTFS2-7441): consistent docstring - object param decorator

abhi-markan:
## Introduction :pencil2:
Ensure docstring is using consistent type.

* Updated all to lowercase `{object}` to `{Object}` for [`@param` decorator](https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#param-and-returns) when referring to an instance of type `Object` or if the parameter should be a generic object.
* Corrected primitive types.
* Correct `Array` type.


azure-functions/acbs-function/mappings/facility/helpers/get-loan-maximum-liability.js
```diff
@@ -6,8 +6,8 @@ const { to2Decimals } = require('../../../helpers/currency');
  * `Loan` (EWCS) = Disbursement amount * (UKEF cover percentage / 100)
  * `Bond` (BSS) = Amount
  * @param {Float} amount Facility UKEF exposure
- * @param {object} facility Facility
- * @param {object} dealType Deal type
+ * @param {Object} facility Facility
+ * @param {Object} dealType Deal type
```
ttbarnes, line 10: Appreciate this isn't your part of your changes but `dealType` is a string

ttbarnes (dismissed):

azure-functions/acbs-function/mappings/facility/helpers/get-loan-maximum-liability.js
abhi-markan: Good point @ttbarnes have updated alongside with additional types.

ttbarnes (approved):

AlexBramhill (commented):
Looks good, but we need to capture all changes of `object`

`\*.*\{.*object.*\}` is a regex that should help

ttbarnes (dismissed):

AlexBramhill (approved):

ttbarnes (approved):

PR #3604, opened 2024-10-01
feat(DTFS2-7300): update db on effective from date page

BethThomas141:
## Introduction :pencil2:
The effective from date page should update the database when a user presses 'continue'. It should also display previously entered values (e.g. if they press back)

I've followed the implementation precedent set on the deal cancellation reason page and [bank request date](https://github.com/UK-Export-Finance/dtfs2/pull/3597) page for consistency.

- get effective from date from the database
- update effective from date
- validate deal exists & submission type is valid on both post and get endpoints


trade-finance-manager-ui/server/controllers/case/cancellation/effective-from-date.controller.ts
```diff
@@ -17,19 +20,43 @@ export type PostEffectiveFromDateRequest = CustomExpressRequest<{
...
+      return res.redirect(`/case/${_id}/deal`);
+    }
+
+    const cancellation = await api.getDealCancellation(_id, userToken);
+
+    const previouslyEnteredEffectiveFromDate = cancellation?.effectiveFrom && new Date(cancellation.effectiveFrom);
+
+    const day = previouslyEnteredEffectiveFromDate ? format(previouslyEnteredEffectiveFromDate, 'd') : '';
+    const month = previouslyEnteredEffectiveFromDate ? format(previouslyEnteredEffectiveFromDate, 'M') : '';
+    const year = previouslyEnteredEffectiveFromDate ? format(previouslyEnteredEffectiveFromDate, 'yyyy') : '';
```
ttbarnes, line 44: Not a blocker for this PR - we should probably create a helper for this, I can see such logic is already in place in 2/3 other files, it could become something like this in individual controllers and unit tests 😎 

```ts
const effectiveFromDateViewModel: EffectiveFromDateViewModel = {
  activePrimaryNavigation: PRIMARY_NAVIGATION_KEYS.ALL_DEALS,
  user,
  ukefDealId: deal.dealSnapshot.details.ukefDealId,
  dealId: _id,
  ...formattedDate(previouslyEnteredEffectiveFromDate),   // <<<< { day, month, year }
};
```

ttbarnes (approved):

trade-finance-manager-ui/server/controllers/case/cancellation/effective-from-date.controller.ts
BethThomas141: Good point! I'll merge this one as is to keep things moving, but will bear in mind going forwards.

PR #3631, opened 2024-10-02
feat(DTFS2-7300): fix effective from date page test

BethThomas141:
## Introduction :pencil2:
Fix a failing test - currently the page following the effective from date page is not implemented, and so clicking 'continue' on the effective from date page goes to the 'Not found' page which doesn't have a 'back' button.

This PR fixes this by instead returning to the effective from date page by using the URL instead of the back button

* Set the GitHub actions feature flags to boolean instead of string.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -349,10 +349,10 @@ jobs:
     needs: setup
     env:
       # Various feature flags
-      FF_TFM_PAYMENT_RECONCILIATION_ENABLED: true
-      FF_TFM_FACILITY_END_DATE_ENABLED: true
-      FF_TFM_DEAL_CANCELLATION_ENABLED: true
-      FF_TFM_SSO_ENABLED: true
+      FF_TFM_PAYMENT_RECONCILIATION_ENABLED: 'true'
+      FF_TFM_FACILITY_END_DATE_ENABLED: 'true'
+      FF_TFM_DEAL_CANCELLATION_ENABLED: 'true'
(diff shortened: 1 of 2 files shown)


ttbarnes (dismissed):

Zainzzkk (approved):

.github/workflows/test.yml
```diff
@@ -349,10 +349,10 @@ jobs:
     needs: setup
     env:
       # Various feature flags
-      FF_TFM_PAYMENT_RECONCILIATION_ENABLED: 'true'
-      FF_TFM_FACILITY_END_DATE_ENABLED: 'true'
-      FF_TFM_DEAL_CANCELLATION_ENABLED: 'true'
-      FF_TFM_SSO_ENABLED: 'true'
+      FF_TFM_PAYMENT_RECONCILIATION_ENABLED: true
```
AlexBramhill, line 352: Env vars are parsed as strings, so we should keep these as strings to maintain reliability in our tests (I'm not sure if github will pass these as strings or not magically, but we can be explicit to avoid having to research this please)

AlexBramhill: @abhi-markan Do you mind if we remove this commit from this hotfix please as I feel it needs a bit more thought

abhi-markan (approved):

.github/workflows/test.yml
abhi-markan: As discussed I will raise a separate PR. Thanks

ttbarnes (approved):

PR #3634, opened 2024-10-02
Feat/dtfs2 7447/add cancel cancellation endpoint central api

MarRobSoftwire:
## Introduction :pencil2:
Include a summary of the changes and related feature(s) or issue(s).

## Resolution :heavy_check_mark:
List all changes made to the codebase.

## Miscellaneous :heavy_plus_sign:
List any additional fixes or improvements.

PR #3636, opened 2024-10-03
feat(DTFS2-7446): add delete cancellation endpoint in tfm-api

MarRobSoftwire:
## Introduction :pencil2:
The user can click cancel during the cancel deal journey. This should delete all cancellation data in the database

- Add deleteDealCancellation to api service, calling new endpoint in dtfs-central-api (#3634)
- Add delete cancellation endpoint and controller
- Add api and unit test coverage

### DIFF
--- a/trade-finance-manager-api/api-tests/api.ts
+++ b/trade-finance-manager-api/api-tests/api.ts
@@ -106,4 +106,11 @@ export const createApi = (app: unknown): TestApi => ({
     }
     return requestInProgress.send(data);
   },
+  remove: async (url: string, data: AnyObject, { headers }: { headers: IncomingHttpHeaders | undefined } = { headers: undefined }) => {
+    const requestInProgress = request(app).delete(url);
+    if (headers) {
+      await requestInProgress.set(headers);
+    }
+    return requestInProgress.send(data);
+  },
 });
--- a/trade-finance-manager-api/src/v1/api.js
+++ b/trade-finance-manager-api/src/v1/api.js
@@ -328,6 +328,35 @@ const getDealCancellation = async (dealId) => {
   }
 };
 
+/**
+ * Deletes the deal cancellation object on a TFM AIN or MIN deal
+ * @param {Object} params
+ * @param {string} params.dealId - deal cancellation to update
+ * @param {import('@ukef/dtfs2-common').AuditDetails} params.auditDetails - user making the request
+ * @returns {Promise<void>} update result object
+ */
+const deleteDealCancellation = async ({ dealId, auditDetails }) => {
+  try {
+    const isValidDealId = isValidMongoId(dealId);
+
+    if (!isValidDealId) {
+      throw new InvalidDealIdError(dealId);
+    }
+
+    await axios({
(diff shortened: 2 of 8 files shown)


trade-finance-manager-api/api-tests/api.ts
```diff
@@ -106,4 +106,11 @@ export const createApi = (app: unknown): TestApi => ({
     }
     return requestInProgress.send(data);
   },
+  remove: async (url: string, data: AnyObject, { headers }: { headers: IncomingHttpHeaders | undefined } = { headers: undefined }) => {
```
ttbarnes, line 109: Would this work? To avoid defining `undefined`
```
headers?: IncomingHttpHeaders
```

trade-finance-manager-api/src/v1/api.js
```diff
@@ -328,6 +328,35 @@ const getDealCancellation = async (dealId) => {
   }
 };
 
+/**
+ * Deletes the deal cancellation object on a TFM AIN or MIN deal
```
ttbarnes, line 332: Do we need to mentino AIN/MIN deal here? If MIA is added for example this could become out of sync/misleading

trade-finance-manager-api/src/v1/controllers/deal-cancellation/delete-deal-cancellation.controller.test.ts
```diff
@@ -0,0 +1,88 @@
...
+
+      // Act
+      await deleteDealCancellation(req, res);
+
+      // Assert
+      expect(res._getStatusCode()).toBe(HttpStatusCode.NoContent);
+    });
+
+    it('should return an error when there is an API error', async () => {
+      const testErrorStatus = 418;
```
ttbarnes, line 53: Should we use maybe 400/500 instead of the tea pot status code

ttbarnes (changes requested):
Nice work, few minor comments

trade-finance-manager-api/api-tests/api.ts
MarRobSoftwire: I think it makes sense to keep it consistent with the `get` and `put` methods above

trade-finance-manager-api/src/v1/api.js
MarRobSoftwire: MIA deals will be abandoned on a separate endpoint - I think this is a business logic thing & I dont expect it to change

trade-finance-manager-api/src/v1/controllers/deal-cancellation/delete-deal-cancellation.controller.test.ts
MarRobSoftwire:
```suggestion
      const testErrorStatus = 400;
```

trade-finance-manager-api/api-tests/api.ts
ttbarnes: Makes sense, thanks

ttbarnes (approved):

PR #3226, opened 2024-06-10
chore(DTFS2-NONE): add sleep to utilisation recon summary test

MarRobSoftwire:
## Introduction :pencil2:
`get-utilisation-reports-reconciliation-summary.api-test.ts` is failing because it runs code before the SQL db connection has been initialised.

## Resolution :heavy_check_mark:
Add a timeout to prevent the code from running before the DB connection is initialised. This should only be a temporary fix but a large refactor of the `createApi` & `createApp` files would be needed otherwise

PR #3638, opened 2024-10-03
feat(FN-2311): fix broken premium payments column sorting

sw-mattw:
## Introduction :pencil2:
A previous PR for this ticket (#3226) moved a few of the `script` tags from the `index.njk` template into the `njk` templates which made use of them.

This broke sorting of the premium payments table as the MOJ frontend script was moved above the content for the filter button to work. This sorting functionality relied on the MOJ frontend script being after the content so that the DOM could be manipulated at script load (to add the sorting functionality). There were no tests in place to test this JS sorting worked as expected.

- Added a new nunjuck block for page scripts, allowing the majority of script tags to be placed after the content (ideal scenario) and scripts only used by a few templates to be included in only those templates through use of this new code block.
- Added E2E tests for each sortable column in the premium payments table to ensure that the script is imported correctly.

- The `cy.visit` functionality will be updated by FN-3483 (Refactor Cypress tests to use reportLink for navigating to utilisation reports) and is out of scope for this ticket


e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/sort-premium-payments.spec.js
```diff
@@ -0,0 +1,92 @@
...
+
+    cy.task(NODE_TASKS.INSERT_UTILISATION_REPORTS_INTO_DB, [utilisationReport]);
+
+    cy.task(NODE_TASKS.INSERT_FEE_RECORDS_INTO_DB, [firstFeeRecord, secondFeeRecord, thirdFeeRecord]);
+    cy.task(NODE_TASKS.INSERT_PAYMENTS_INTO_DB, [firstPayment, secondPayment]);
+
+    pages.landingPage.visit();
+    cy.login(USERS.PDC_RECONCILE);
+
+    cy.visit(`/utilisation-reports/${reportId}`);
```
sw-mattw, line 60: Updating this is out of scope, see PR description

e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/sort-premium-payments.spec.js
```diff
@@ -0,0 +1,92 @@
...
+  });
+
+  describe('when status column heading is clicked', () => {
+    it('should sort the rows by status', () => {
+      pages.utilisationReportPage.premiumPaymentsTab.premiumPaymentsTable.header.status().click();
+
+      pages.utilisationReportPage.premiumPaymentsTab.premiumPaymentsTable.rows().eq(0).first('th').contains('33333333');
+      pages.utilisationReportPage.premiumPaymentsTab.premiumPaymentsTable.rows().eq(1).first('th').contains('22222222');
+      pages.utilisationReportPage.premiumPaymentsTab.premiumPaymentsTable.rows().eq(2).first('th').contains('11111111');
```
RLavender98, line 69: Since the assertions don't actually explicitly mention the status please could we add comments explaining why we expect these numbers in these orders?

RLavender98 (commented):
Thanks for adding some tests so we don't see the same regression in the future!

trade-finance-manager-ui/templates/utilisation-reports/find-utilisation-reports-by-year.njk
```diff
@@ -83,7 +83,9 @@
       </form>
     </div>
   </div>
+{% endblock %}
+{% block sub_content %}{% endblock %}
```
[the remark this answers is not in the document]
sw-mattw: This seems to be a convention across the other templates, I've moved this up from line 89.

It's defining an empty block to be used by the parent template, however, I don't think this affects the functionality of the page - more convention. Happy to remove if we want

e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/sort-premium-payments.spec.js
sw-mattw: Added docs to these thanks

e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/sort-premium-payments.spec.js
```diff
@@ -0,0 +1,117 @@
...
+
+      // Facility 22222222 has status 'DOES_NOT_MATCH' (alphabetically first) and has the lowest total reported payments
+      // (secondary sorting criteria, ascending), so it should appear first.
+      pages.utilisationReportPage.premiumPaymentsTab.premiumPaymentsTable.rows().eq(0).first('th').contains('22222222');
+
+      // Facility 11111111 also has status 'DOES_NOT_MATCH' but has a slightly higher total reported payments, so it
+      // should appear second.
+      pages.utilisationReportPage.premiumPaymentsTab.premiumPaymentsTable.rows().eq(1).first('th').contains('11111111');
+
+      // Facility 33333333 has status 'TO_DO' (alphabetically last), so it should appear last..
```
RLavender98, line 77:
```suggestion
      // Facility 33333333 has status 'TO_DO' (alphabetically last), so it should appear last.
```

e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/sort-premium-payments.spec.js
```diff
@@ -0,0 +1,117 @@
...
+
+      // NOTE: This column is the default sorting column and defaults to ascending, therefore the descending case is being tested.
+
+      // Facility 11111111 has the highest total reported payments (30 EUR), so it should appear first.
+      pages.utilisationReportPage.premiumPaymentsTab.premiumPaymentsTable.rows().eq(0).first('th').contains('11111111');
+
+      // Facility 22222222 has the second highest total reported payments (20 GBP), so it should appear second.
+      pages.utilisationReportPage.premiumPaymentsTab.premiumPaymentsTable.rows().eq(1).first('th').contains('22222222');
+
+      // Facility 33333333 has the lowest total reported payments (10 USD), so it should appear last.
```
RLavender98, line 113: I actually think these are in this order because first we order by the currency alphabetically? (e.g. 1000 USD would still come last even though actually highest)

RLavender98 (dismissed):

e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/sort-premium-payments.spec.js
```diff
@@ -0,0 +1,117 @@
...
+      // NOTE: This column is the default sorting column and defaults to ascending, therefore the descending case is being tested.
+
+      // Facility 11111111 has reported payment currency 'EUR' (alphabetically first), so it should appear first.
+      pages.utilisationReportPage.premiumPaymentsTab.premiumPaymentsTable.rows().eq(0).first('th').contains('11111111');
+
+      // Facility 22222222 has reported payment currency 'GBP' (alphabetically second), so it should appear second.
+      pages.utilisationReportPage.premiumPaymentsTab.premiumPaymentsTable.rows().eq(1).first('th').contains('22222222');
+
+      // Facility 33333333 has reported payment currency 'USD' (alphabetically last), so it should appear last.
+      pages.utilisationReportPage.premiumPaymentsTab.premiumPaymentsTable.rows().eq(2).first('th').contains('33333333');
```
ttbarnes, line 114: Can we use `cy.assertText` instead of `contains`?

ttbarnes, line 114: Can we use `cy.assertText` instead of `contains`?

e2e-tests/tfm/cypress/e2e/pages/utilisation-reports/utilisationReportPage.js
```diff
@@ -12,6 +12,12 @@ const utilisationReportPage = {
     generateKeyingDataButton: () => cy.get('[data-cy="generate-keying-data-button"]'),
     addAPaymentButton: () => cy.get('[data-cy="add-a-payment-button"]'),
     premiumPaymentsTable: {
+      header: {
+        totalReportedPayments: () => cy.get('th[data-cy="premium-payments-table--total-reported-payments"] button'),
+        totalPaymentsReceived: () => cy.get('th[data-cy="premium-payments-table--total-payments-received"] button'),
+        status: () => cy.get('th[data-cy="premium-payments-table--status"] button'),
+      },
+      rows: () => cy.get('tr[data-cy^="premium-payments-table-row--feeRecordId-"]'),
```
ttbarnes, line 20: Looks like a typo here? `feeRecordId-`

e2e-tests/tfm/cypress/e2e/pages/utilisation-reports/utilisationReportPage.js
```diff
@@ -12,6 +12,12 @@ const utilisationReportPage = {
     generateKeyingDataButton: () => cy.get('[data-cy="generate-keying-data-button"]'),
     addAPaymentButton: () => cy.get('[data-cy="add-a-payment-button"]'),
     premiumPaymentsTable: {
+      header: {
+        totalReportedPayments: () => cy.get('th[data-cy="premium-payments-table--total-reported-payments"] button'),
+        totalPaymentsReceived: () => cy.get('th[data-cy="premium-payments-table--total-payments-received"] button'),
```
ttbarnes, line 17: Is it possible to add a `data-cy`to the button, or is this not controlled by us?

ttbarnes (changes requested):
Few minor comments please

e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/sort-premium-payments.spec.js
sw-mattw: Yes thanks, had my currencies mixed up missed that, pushed updates for this

e2e-tests/tfm/cypress/e2e/pages/utilisation-reports/utilisationReportPage.js
sw-mattw: The button is inserted by the MOJ frontend script I believe so I don't think we can pass any attributes down.

I tried this without the `button` and the tests don't wait for the script to manipulate the DOM (to add the button) - so the sorting wouldn't be tested correctly without selecting the button element itself

e2e-tests/tfm/cypress/e2e/pages/utilisation-reports/utilisationReportPage.js
sw-mattw: This is to get all rows with that prefix (which is then followed by the specific fee record ID) - added a comment just above this line

e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/sort-premium-payments.spec.js
sw-mattw: _In case you see the pushed commit for this that's causing the test failures, I'm working on this still_

sw-mattw: Yep apologies, keep forgetting that's been added. Pushed a change to use `cy.assertText` instead thanks

e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/sort-premium-payments.spec.js
```diff
@@ -0,0 +1,117 @@
...
+  describe('when total reported payments column heading is clicked', () => {
+    it('should sort the rows by total reported payments', () => {
+      pages.utilisationReportPage.premiumPaymentsTab.premiumPaymentsTable.header.totalReportedPayments().click();
+
+      // Check that the rows are sorted by total reported payments by checking the facility IDs.
+
+      // NOTE: This column is the default sorting column and defaults to ascending, therefore the descending case is being tested.
+
+      // Facility 11111111 has reported payment currency 'EUR' (alphabetically first), so it should appear first.
+      cy.assertText(pages.utilisationReportPage.premiumPaymentsTab.premiumPaymentsTable.rows().eq(0).find('th').first(), '11111111');
```
ttbarnes, line 108: Could we move the selectors into `premiumPaymentsTable` please, keep the actual test simpler and DRY? Not sure about the naming/structure but would like it to be e.g:

```js
cy.assertText(aSelector, '11111111');
cy.assertText(anotherSelector, '22222222');
```

e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/sort-premium-payments.spec.js
ttbarnes: Now worries thanks @sw-mattw 😊

e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/sort-premium-payments.spec.js
sw-mattw: Thanks, pulled out some of this into: ...`premiumPaymentsTable.facilityIdByRowIndex(0)`

Is this similar to what you had in mind/okay? Thanks

ttbarnes (approved):

PR #3639, opened 2024-10-03
feat(DTFS2-7297): add deal cancellation check details page

BethThomas141:
## Introduction :pencil2:
Add the check details page to the deal cancellation flow

- Add nunjucks template
- Add controller (note the proper implementation of the POST endpoint will be handled in a separate ticket)
- Add component and unit tests
- Add e2e tests

## Screenshots
image
image

### DIFF
--- a/e2e-tests/tfm/cypress/e2e/journeys/deal-cancellation/check-details.spec.js
+++ b/e2e-tests/tfm/cypress/e2e/journeys/deal-cancellation/check-details.spec.js
@@ -93,10 +93,11 @@ context('Deal cancellation - check details', () => {
       cy.url().should('eq', relative(`/case/${dealId}/deal`));
     });
 
-    it('return to deal summary link should take you to the deal summary page', () => {
+    // TODO: DTFS2-7359 - add this test once cancel link is implemented
+    it.skip('return to deal summary link should take you to confirm cancellation page', () => {
       checkDetailsPage.returnLink().click();
 
-      cy.url().should('eq', relative(`/case/${dealId}/deal`));
+      cy.url().should('eq', relative(`/case/${dealId}/cancellation/cancel`));
     });
 
     it('reason change link should take you to the reason for cancelling page and correctly update the reason', () => {
--- a/e2e-tests/tfm/cypress/e2e/journeys/deal-cancellation/check-details.spec.js
+++ b/e2e-tests/tfm/cypress/e2e/journeys/deal-cancellation/check-details.spec.js
@@ -1,0 +1,168 @@
+import { format } from 'date-fns';
+import relative from '../../relativeURL';
+import MOCK_DEAL_AIN from '../../../fixtures/deal-AIN';
+import { ADMIN, BANK1_MAKER1, PIM_USER_1, T1_USER_1 } from '../../../../../e2e-fixtures';
+import caseDealPage from '../../pages/caseDealPage';
+import { backLink } from '../../partials';
+import effectiveFromDatePage from '../../pages/deal-cancellation/effective-from-date';
+import dateConstants from '../../../../../e2e-fixtures/dateConstants';
+import bankRequestDatePage from '../../pages/deal-cancellation/bank-request-date';
+import checkDetailsPage from '../../pages/deal-cancellation/check-details';
+import reasonForCancellingPage from '../../pages/deal-cancellation/reason-for-cancelling';
+
+context('Deal cancellation - check details', () => {
+  let dealId;
+  const dealFacilities = [];
+
+  before(() => {
+    cy.insertOneDeal(MOCK_DEAL_AIN, BANK1_MAKER1).then((insertedDeal) => {
+      dealId = insertedDeal._id;
+
+      const { dealType, mockFacilities } = MOCK_DEAL_AIN;
+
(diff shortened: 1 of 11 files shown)


e2e-tests/tfm/cypress/e2e/journeys/deal-cancellation/check-details.spec.js
```diff
@@ -0,0 +1,169 @@
...
+
+      checkDetailsPage.reasonLink();
+
+      checkDetailsPage.bankRequestDateResponse();
+      checkDetailsPage.bankRequestDateResponse().should('have.text', format(dateConstants.today, 'd MMMM yyyy'));
+
+      checkDetailsPage.bankRequestDateLink();
+
+      checkDetailsPage.effectiveFromResponse();
+      checkDetailsPage.effectiveFromResponse().should('have.text', format(dateConstants.tomorrow, 'd MMMM yyyy'));
```
ttbarnes, line 76: Can we use `cy.assertText` instead of `should('have.text'` please 🙏

trade-finance-manager-ui/server/controllers/case/cancellation/check-details.controller.ts
```diff
@@ -0,0 +1,82 @@
...
+      return res.redirect('/not-found');
+    }
+
+    if (!canSubmissionTypeBeCancelled(deal.dealSnapshot.submissionType)) {
+      return res.redirect(`/case/${_id}/deal`);
+    }
+
+    const { reason, bankRequestDate, effectiveFrom } = await api.getDealCancellation(_id, userToken);
+
+    const bankRequestDateFormatted = bankRequestDate ? format(new Date(bankRequestDate), 'd MMMM yyyy') : undefined;
```
ttbarnes, line 36:
```suggestion
    const bankRequestDateFormatted = bankRequestDate && format(new Date(bankRequestDate), 'd MMMM yyyy');
```

trade-finance-manager-ui/templates/case/cancellation/check-details.njk
```diff
@@ -0,0 +1,115 @@
...
+        <label class="govuk-label govuk-label--l" for="check-details">
+          Check details before cancelling deal
+        </label>
+      </h1>
+
+      {{ govukTable({
+        rows: optionsWithHTML,
+       firstCellIsHeader: true,
+       classes: "govuk-!-margin-top-4"
+      }) }}
```
ttbarnes, line 90: indentation is slightly off here

ttbarnes (changes requested):
Nice work, few minor comments

trade-finance-manager-ui/server/controllers/case/cancellation/check-details.controller.ts
BethThomas141: Due to using Typescript these two implementations are subtly different: 

**Current implementation:**
image
`bankRequestDateFormatted` returns either a `string` or `undefined`



**Suggested implementation:**
image
`bankRequestDateFormatted` returns either a `string`, `undefined` OR `0`
This is due to possibly surprising behaviour / took me a while to think about! But bankRequestDate is received as either a number or undefined, and obviously when it's a number we expect it to be a Unix time stamp, but Typescript doesn't know this so it think the number _could be 0_. So if we use the `&&` operator, if it was hypothetically `0` then it would return `0` as the LHS of the operand which is why it considers this a possible outcome of the statement. 

Due to it being a typescript file, this causes problems elsewhere in the file if it thinks it could be 0 (even though we know it won't ever be). So I'd rather keep it as is to make it clear that we're passing in `undefined` or a string. 

What do you reckon?

e2e-tests/tfm/cypress/e2e/journeys/deal-cancellation/check-details.spec.js
BethThomas141: Ah yes of course!! Have changed thank you!

trade-finance-manager-ui/templates/case/cancellation/check-details.njk
BethThomas141: Good spot, have fixed thanks!

e2e-tests/tfm/cypress/e2e/journeys/deal-cancellation/check-details.spec.js
```diff
@@ -0,0 +1,169 @@
...
+    });
+
+    it('delete deal button should take you to the deal summary page', () => {
+      checkDetailsPage.dealDeletionButton().click();
+
+      cy.url().should('eq', relative(`/case/${dealId}/deal`));
+    });
+
+    // TODO: DTFS2-7359 - add this test once cancel link is implemented
+    it.skip('return to deal summary link should take you to confirm cancellation page', () => {
```
ttbarnes, line 97: `.skip`?

ttbarnes (approved):

e2e-tests/tfm/cypress/e2e/journeys/deal-cancellation/check-details.spec.js
BethThomas141: Have removed, thanks!

ttbarnes (dismissed):

ttbarnes (approved):

PR #3640, opened 2024-10-03
feat(DTFS2-7445): add cancel cancellation template

MarRobSoftwire:
## Introduction :pencil2:
The user can click cancel during the cancel deal journey. This should redirect to a confirmation page. 

- Add `cancel.njk` nunjucks template
- Add GET endpoint to render the template

## Screenshots 📸 
image

- Move view model test helpers into subdirectories

## Note ⚠️ 
- This PR does not connect the existing `cancel` links to the new page. 
- This PR does not add a POST endpoint to delete the deal cancellation nor redirect back to the previous page.


trade-finance-manager-ui/server/controllers/case/cancellation/cancel-cancellation.controller.ts
```diff
@@ -0,0 +1,42 @@
...
+  const { _id } = req.params;
+  const { user, userToken } = asUserSession(req.session);
+  try {
+    const deal = await api.getDeal(_id, userToken);
+
+    if (!deal || 'status' in deal) {
+      return res.redirect('/not-found');
+    }
+
+    if (!canSubmissionTypeBeCancelled(deal.dealSnapshot.submissionType)) {
```
ttbarnes, line 27: I'm wondering if the `canSubmissionTypeBeCancelled` check should be in middleware, we have this repeated in 4x places. What do you think?

Not a blocker for this PR.

ttbarnes (approved):

trade-finance-manager-ui/server/controllers/case/cancellation/cancel-cancellation.controller.ts
MarRobSoftwire: Its worth thinking about, but it needs the request to be made to get the deal so probably has to happen in the controller. Something I've been thinking about though - we could do something like the middleware calls the api, checks its valid, & passes this in as `req.body.deal` or similar. 
Would be a large change of pattern though!

ttbarnes (dismissed):

ttbarnes (approved):

PR #3641, opened 2024-10-03
feat(DTFS2-7448): update cancel link hrefs

MarRobSoftwire:
## Introduction :pencil2:
The existing cancellation pages have cancel links redirecting to the deal summary page. These should take the user to the new cancel cancellation page

Update the cancel link hrefs
Update the e2e tests
Update the component tests

## Note ⚠️ 
This page doesnt exist yet (Implemented by #3640)

### DIFF
--- a/trade-finance-manager-ui/templates/case/cancellation/bank-request-date.njk
+++ b/trade-finance-manager-ui/templates/case/cancellation/bank-request-date.njk
@@ -100,7 +100,7 @@
             text: "Continue",
             attributes: { "data-cy": "continue-button" }
           }) }}
-          <a class="govuk-link" href="/case/{{ dealId }}/deal" data-cy="cancel-link">Cancel</a>
+          <a class="govuk-link" href="cancel" data-cy="cancel-link">Cancel</a>
         </div>
 
       </form>
--- a/trade-finance-manager-ui/templates/case/cancellation/bank-request-date.njk
+++ b/trade-finance-manager-ui/templates/case/cancellation/bank-request-date.njk
@@ -100,7 +100,7 @@
             text: "Continue",
             attributes: { "data-cy": "continue-button" }
           }) }}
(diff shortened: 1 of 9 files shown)


trade-finance-manager-ui/templates/case/cancellation/bank-request-date.njk
```diff
@@ -100,7 +100,7 @@
             text: "Continue",
             attributes: { "data-cy": "continue-button" }
           }) }}
-          <a class="govuk-link" href="/case/{{ dealId }}/deal" data-cy="cancel-link">Cancel</a>
+          <a class="govuk-link" href="cancel" data-cy="cancel-link">Cancel</a>
```
ttbarnes, line 103: Shouldn't this be `/cancel` or similar?

MarRobSoftwire: This is working as shorthand for `/case/${dealId}/cancellation/cancel`. I'm never quite sure how links work in HTML but this seems to be replacing the `bank-request-date` at the end of the URL with `cancel`

ttbarnes: I think a failsafe is to have a backslash - can we add this in please? Should still pass

This might be some very oldschool thinking from me 🙈 but regardless it feels odd to _not_ have a backslash

MarRobSoftwire: Just played around with this & the `cancel` will break if the URL has a `/` on the end! Great spot, will replace it with the full `/case/{{ dealId }}/cancellation/cancel`

ttbarnes: Ah ha, it's been a while for me, but now I remember this is why 😄 Great spot/testing 👍

ttbarnes (approved):

PR #3643, opened 2024-10-04
feat(DTFS2-7449): add post cancel endpoint

MarRobSoftwire:
## Introduction :pencil2:

### DIFF
--- a/trade-finance-manager-ui/server/api.js
+++ b/trade-finance-manager-ui/server/api.js
@@ -1259,7 +1259,7 @@ const updateDealCancellation = async (dealId, cancellationUpdate, userToken) => 
 };
 
 /**
- * Updates the deal cancellation object on a TFM MIN or AIN deal
+ * Gets the deal cancellation object on a TFM MIN or AIN deal
  * @param {string} dealId - The deal ID
  * @param {string} userToken - The user token
  * @returns {Promise<Partial<import('@ukef/dtfs2-common').TfmDealCancellation>>}
@@ -1273,6 +1273,25 @@ const getDealCancellation = async (dealId, userToken) => {
     });
 
     return response.data;
+  } catch (error) {
+    console.error('Failed to get deal cancellation', error);
+    throw error;
+  }
+};
+
+/**
+ * Deletes the deal cancellation object on a TFM MIN or AIN deal
+ * @param {string} dealId - The deal ID
+ * @param {string} userToken - The user token
+ * @returns {Promise<void>}
+ */
+const deleteDealCancellation = async (dealId, userToken) => {
+  try {
+    await axios({
+      method: 'delete',
+      url: `${TFM_API_URL}/v1/deals/${dealId}/cancellation`,
+      headers: generateHeaders(userToken),
+    });
   } catch (error) {
     console.error('Failed to get deal cancellation', error);
     throw error;
@@ -1337,4 +1356,5 @@ module.exports = {
   addFeesToAnExistingPayment,
   updateDealCancellation,
   getDealCancellation,
+  deleteDealCancellation,
 };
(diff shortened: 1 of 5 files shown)


trade-finance-manager-ui/server/controllers/case/cancellation/cancel-cancellation.controller.post.test.ts
```diff
@@ -0,0 +1,145 @@
...
+
+      // Act
+      await postCancelCancellation(req, res);
+
+      // Assert
+      expect(api.deleteDealCancellation).toHaveBeenCalledTimes(1);
+      expect(api.deleteDealCancellation).toHaveBeenCalledWith(dealId, userToken);
+    });
+
+    it('redirects to the deal summary page', async () => {
```
ttbarnes, line 126: Can we add a test to make sure it does not call `api.deleteDealCancellation`?

trade-finance-manager-ui/server/controllers/case/cancellation/cancel-cancellation.controller.post.test.ts
```diff
@@ -0,0 +1,145 @@
...
+    });
+
+    // Act
+    await postCancelCancellation(req, res);
+
+    // Assert
+    expect(res._getRedirectUrl()).toBe(`/not-found`);
+  });
+
+  it('redirects to deal summary page if the submission type is invalid (MIA)', async () => {
```
ttbarnes, line 84: Can we add a test to make sure it does not call `api.deleteDealCancellation`?

ttbarnes (changes requested):
E2E test?

trade-finance-manager-ui/server/controllers/case/cancellation/cancel-cancellation.controller.post.test.ts
ttbarnes: Thanks @MarRobSoftwire looks good, can we just wrap into one describe block though please? Little more DRY e.g

```ts
describe('when the type is invalid (MIA)', () => {
  it('should redirect to deal summary page ...

  it('should not call ...
});
```

MarRobSoftwire: Have updated :)

trade-finance-manager-ui/server/controllers/case/cancellation/validation/validate-previous-page.ts
```diff
@@ -0,0 +1,26 @@
...
+ *  - effective from date page
+ *  - check details page
+ *
+ * and returns a new relative URL for that page.
+ * Otherwise, returns the deal summary page URL.
+ * @param url - The URL of the previous page, provided in the post request
+ * @param dealId - The deal id
+ * @returns the previous page URL
+ */
+export const validatePreviousPage = (url: string, dealId: string): string => {
```
ttbarnes, line 14: Do we really need to do this? The design has:

_"Are you sure you want to cancel the request?
You'll lose all the data you have entered **and return to deal summary**"_

Instead of this and the following in the controller:

```ts
return res.redirect(validatePreviousPage(req.body.previousPage, _id));
```

Could we not just do the following?

```ts
return res.redirect('/case-summary-url');
```

MarRobSoftwire: The secondary button `No go back` needs to redirect the user to the previous page.

when the user clicks this button, a POST request is made from (and to) `cancellation/cancel` meaning the `req.headers.referrer` is not the previous page.

To get around this, I'm passing the `previousPage` as a hidden input on the POST request. The downside is that the previous page is now _technically_ a user supplied value and needs to be validated before using it to redirect the user. 

I could get around this by adding the previous page to flash/session but I think this is the easiest solution to implement.

ttbarnes: Thanks @MarRobSoftwire it's not my call, but this feels like a strange design - visual and therefore technical solution to achieve it.

The `req.body` or DOM could also be manipulated and then there'd be no `previousPage` and the URL would be undefined. This is edge case though.

Can we at least rename `validatePreviousPage` to `getPreviousPageUrl` and remove the `validation` directory please because at the moment It could be read that "previous page" is a validation rule in terms of the form POST.

I get that it's consuming a value from the form body but i'd typically expect anything in a controller's validation directory to be validation rules that are presented to a user.

MarRobSoftwire: Sure will do,
as to the `previousPage` being undefined - this function will return `/case/${dealId}/deal` if thats the case

ttbarnes (approved):

PR #3644, opened 2024-10-07
fix(DTFS2-7444): eStore void job validation

abhi-markan:
## Introduction :pencil2:
Ensure documents CRON job validated

* Void job validation, useful due to unexpected behaviour and micro-service restart.
* Added document upload retries for high volume files (this is currently a workaround as re-calling document upload API once will suffice however due to an-going issue of upload with 5+ files a re-upload is being tested).

### DIFF
--- a/external-api/src/helpers/cron.ts
+++ b/external-api/src/helpers/cron.ts
@@ -9,6 +9,7 @@ import {
   eStoreBuyerDirectoryCreationJob,
   eStoreDealDirectoryCreationJob,
   eStoreFacilityDirectoryCreationJob,
+  eStoreDocumentsCreationJob,
 } from '../cron';
 
 dotenv.config();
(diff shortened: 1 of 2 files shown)


external-api/src/helpers/cron.ts
```diff
@@ -33,6 +34,8 @@ const onTick = async (eStoreData: Estore, category: Category) => {
       return eStoreDealDirectoryCreationJob(eStoreData);
     case ENDPOINT.FACILITY:
       return eStoreFacilityDirectoryCreationJob(eStoreData);
+    case ENDPOINT.DOCUMENT:
+      return eStoreDocumentsCreationJob(eStoreData);
```
ttbarnes, line 38: Not a blocker as it's already existing, the constant is `DOCUMENT`, but the function is `eStoreDocuments...`, should we rename `DOCUMENT` to `DOCUMENTS`?

ttbarnes (approved):

Zainzzkk (approved):

external-api/src/helpers/cron.ts
abhi-markan: Good point, I have kept all of them singular as document can be either none, one or more than one.

```typescript
export const ENDPOINT = {
  SITE: 'sites',
  TERM: 'terms/facilities',
  BUYER: 'buyers',
  DEAL: 'deals',
  FACILITY: 'facilities',
  DOCUMENT: 'documents',
} as const;
```

ttbarnes: OK makes sense for now - not a blocker but I would vote for adding S to all of them to match the actual value

PR #3647, opened 2024-10-07
feat(DTFS2-7303): change links on deal cancellation summary page

BethThomas141:
## Introduction :pencil2:
When a user presses 'Change' next to one of the three fields on the deal cancellation Check details page it should take the user to that respective page (already implemented). 

Additionally, **when they press 'Back' having come from pressing 'Change' on the  Check details page then they should be redirected back to the Check details page.** This is what this PR implements. 

image

- when a user presses 'change' next to one of those fields, they are redirected to the respective page, and a query parameter `{status: change}` is passed in as well. This is consistent with a pattern also used in gef-ui (`gef-ui\templates\partials\facility-value.njk`)

- If this query parameter is detected, then the back link will go to the Check details page. Otherwise it will go to the default back url.


trade-finance-manager-ui/component-tests/case/cancellation/effective-from-date.component-test.ts
```diff
@@ -31,17 +31,18 @@ describe(page, () => {
     wrapper.expectLink('[data-cy="cancel-link"]').toLinkTo(`/case/${dealId}/cancellation/cancel`, 'Cancel');
   });
 
-  it('should render back link button linking to the deal bank request date page', () => {
+  it('should render back link button with the passed in URL', () => {
     // Arrange
     const dealId = 'dealId';
-    const effectiveFromDateViewModel: EffectiveFromDateViewModel = { ...aEffectiveFromDateViewModel(), dealId };
+    const backUrl = 'testUrl';
+    const effectiveFromDateViewModel: EffectiveFromDateViewModel = { ...aEffectiveFromDateViewModel(), dealId, backUrl };
```
MarRobSoftwire, line 38: FWIW I've been using `previousPage` - happy to stick with this if both are used but want to try keep consistent :)

trade-finance-manager-ui/server/controllers/case/cancellation/effective-from-date.controller.ts
```diff
@@ -51,6 +52,7 @@ export const getEffectiveFromDate = async (req: GetEffectiveFromDateRequest, res
       day,
       month,
       year,
+      backUrl: status === 'change' ? `/case/${_id}/cancellation/check-details` : `/case/${_id}/cancellation/bank-request-date`,
```
MarRobSoftwire, line 55: FWIW could add a function to abstract this logic
```suggestion
      backUrl: getPreviousPage(_id, status),
```

trade-finance-manager-ui/component-tests/case/cancellation/effective-from-date.component-test.ts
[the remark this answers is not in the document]
BethThomas141: Ah nice, I've now changed to `previousPage` to keep consistent, thanks!

trade-finance-manager-ui/server/controllers/case/cancellation/effective-from-date.controller.ts
BethThomas141: Extracted, thanks!

e2e-tests/tfm/cypress/e2e/journeys/deal-cancellation/check-details.spec.js
```diff
@@ -99,60 +107,136 @@ context('Deal cancellation - check details', () => {
...
+        cy.url().should('eq', relative(`/case/${dealId}/cancellation/effective-from-date?status=change`));
+
+        cy.keyboardInput(effectiveFromDatePage.effectiveFromDateYear(), '1');
+        cy.clickContinueButton();
+        errorSummary();
+
+        cy.clickBackLink();
+
+        cy.url().should('eq', relative(`/case/${dealId}/cancellation/check-details`));
+      });
```
ttbarnes, line 239: I'm finding these tests a bit hard to read, especially as there are multiple pages and scenarios involved and I don't have the context of the full journey; I think we should split these up. Something like this, what do you think?

```js
describe("'x row' - change link", () => {
  beforeEach(() => {
    checkDetailsPage.bankRequestDateLink().click();
  });

  it('should redirect to the `bank request date` page', () => {
    cy.url().should('eq', relative(`/case/${dealId}/cancellation/bank-request-date?status=change`));
  });

  describe('when submitting X form', () => {
    it('should render updates', () => {
      // cy.keyboardInput(...)
      // cy.assertText(...)
    });
  });

  describe('when clicking the back link', ( ) => {
    it('should redirect to X page', () => {
    });
  });

  describe('when clicking the back link with validation errors', () => {
    it('should redirect to X page', () => {
    });

    it('should NOT render updates', () => {
    });
  });
});
```

trade-finance-manager-ui/server/controllers/case/cancellation/reason-for-cancelling.controller.ts
```diff
@@ -7,8 +7,18 @@ import { validateReasonForCancelling } from './validation/validate-reason-for-ca
...
+
+/**
+ * Gets the previous page URL to pass into the back link
+ *
+ * @param dealId - The deal ID
+ * @param status - 'change' if user comes from the check answers page, otherwise undefined
+ */
+const getPreviousPage = (dealId: string, status?: string): string => {
+  return status === 'change' ? `/case/${dealId}/cancellation/check-details` : `/case/${dealId}/deal`;
+};
```
ttbarnes, line 21: I don't agree with this design pattern - Issue with this approach is there are some UX / potential accessibility issues:

- The browser back link will always go to the previous page.
- With custom logic like this for a back link in the page, the back link acts differently to the browser back button.
- Using "back" link text could be confusing for some users.

I am not sure what's already in place in the codebase for back links, but they should all generally use `req.headers.referer`, mimicking a browser's back button and aligning all instances of "back".

[GOV design](https://design-system.service.gov.uk/components/back-link/) mentions alternative text, which seems appropriate for DTFS2-7303 - should/could we ask design for their input?

- "_Make sure the link takes users to the previous page they were on, in the state they last saw it._"

- "_For more complex user journeys, consider using different link text, like ‘Go back to [page]’. For example, in an admin system with many different areas. In this case, if you used ‘Back’, it might not be clear to users what they are going back to._"

ttbarnes (changes requested):

trade-finance-manager-ui/server/controllers/case/cancellation/reason-for-cancelling.controller.ts
BethThomas141: @ttbarnes interesting - I've kept to the ACs, but whether or not there should be more ACs I'm not sure (but if so this might have to be a new ticket). 

The reason we can't just use `req.headers.referer` is as follows: 

- Say you're on the `reason for cancelling` page
- Then you enter a reason and click 'Continue' so you're now on the `bank request date` page
- Then you click `back` - the previous page as per `req.headers.referer` was the reason for cancelling page so you get taken there as expected. 
- Then you click `back` again - here a user expects to go to the previous page in the cancellation flow (in this case the deal summary page). However the previous page as per `req.headers.referer` is the `bank request date` page, so you get taken there. 
- Then you get stuck in an infinite loop - on the bank request date page,  `req.headers.referer` has the previous page as the reason for cancelling page, and the reason for cancelling page has `req.headers.referer` as the bank request date page. 

Does that make sense? It's tricky to know what the desired implementation is (especially without using Javascript) but we can't just rely on `req.headers.referer` due to this odd infinite loop behaviour.

e2e-tests/tfm/cypress/e2e/journeys/deal-cancellation/check-details.spec.js
BethThomas141: Good idea, have now split them out as suggested - have a look and see what you think!

e2e-tests/tfm/cypress/e2e/journeys/deal-cancellation/check-details.spec.js
```diff
@@ -99,60 +107,136 @@ context('Deal cancellation - check details', () => {
       cy.url().should('eq', relative(`/case/${dealId}/cancellation/cancel`));
     });
 
-    it('reason change link should take you to the reason for cancelling page and correctly update the reason', () => {
-      const testReason = 'test reason';
+    describe('reason change link', () => {
```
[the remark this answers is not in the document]
BethThomas141: Good idea, have changed!

trade-finance-manager-ui/server/controllers/case/cancellation/reason-for-cancelling.controller.ts
BethThomas141: As per discussion in the teams channel, leaving as the current implementation for now and this can be revisited codebase wide as a separate, more involved, piece of work.

trade-finance-manager-ui/server/controllers/case/cancellation/reason-for-cancelling.controller.ts
```diff
@@ -7,8 +7,18 @@ import { validateReasonForCancelling } from './validation/validate-reason-for-ca
...
+export type GetReasonForCancellingRequest = CustomExpressRequest<{ params: { _id: string }; query: { status?: string } }>;
+export type PostReasonForCancellingRequest = CustomExpressRequest<{ params: { _id: string }; query: { status?: string }; reqBody: { reason: string } }>;
+
+/**
+ * Gets the previous page URL to pass into the back link
+ *
+ * @param dealId - The deal ID
+ * @param status - 'change' if user comes from the check answers page, otherwise undefined
+ */
+const getPreviousPage = (dealId: string, status?: string): string => {
```
ttbarnes, line 19: Can we rename this to `getPreviousPageUrl` and move into a helpers directory please, so we have the same approach as @MarRobSoftwire 's recent PR

Looks like we might need to move the existing `getPreviousPageUrl` helper into a child directory so it's clear, maybe something like

`/controllers/case/cancellation/helpers/X-page/get-previous-page-url.ts`
`/controllers/case/cancellation/helpers/Y-page/get-previous-page-url.ts`

What do you think @BethThomas141 @MarRobSoftwire ?

BethThomas141: I've renamed it to `getPreviousPageUrl` thanks! I'm not sure if it's worth moving each of these out into a child directory - each of these functions is different in the 3 respective files, and only used in one place. 
The logic in these functions is also a lot simpler than in the case that Mark made a separate helper function for. 

I'd prefer to reduce file numbers and keep these at the top of these files (instead of making an additional 3 directories and 3 files).  

Would that be ok? Let me know what you think!

trade-finance-manager-ui/server/controllers/case/cancellation/bank-request-date.controller.post.test.ts
```diff
@@ -151,6 +152,37 @@ describe('postBankRequestDate', () => {
...
+      });
+
+      it('renders the page with the back URL as the check details page when "change" is passed in as a query parameter', async () => {
+        // Arrange
+        const { req, res } = createMocks<PostBankRequestDateRequest>({
+          params: { _id: dealId },
+          query: { status: 'change' },
+          session: {
+            user: mockUser,
+            userToken: 'a user token',
```
ttbarnes, line 166: If Mark's PR [here](https://github.com/UK-Export-Finance/dtfs2/pull/3661) gets merged before this, can we update such instances to use `aRequestSession` please 🙏

trade-finance-manager-ui/server/controllers/case/cancellation/effective-from-date.controller.post.test.ts
```diff
@@ -23,6 +23,7 @@ jest.mock('../../../api', () => ({
 const dealId = 'dealId';
 const ukefDealId = 'ukefDealId';
 const mockUser = aTfmSessionUser();
+const defaultBackUrl = `/case/${dealId}/cancellation/bank-request-date`;
```
ttbarnes, line 26: I'm not sure if this should be here, but rather on line 156 since it's only used once, unless i'm missing something

trade-finance-manager-ui/server/controllers/case/cancellation/effective-from-date.controller.ts
```diff
@@ -8,12 +8,23 @@ import { validateEffectiveFromDate } from './validation/validate-effective-from-
...
   params: { _id: string };
+  query: { status?: string };
   reqBody: { 'effective-from-date-day': string; 'effective-from-date-month': string; 'effective-from-date-year': string };
 }>;
 
+/**
+ * Gets the previous page URL to pass into the back link
+ *
+ * @param dealId - The deal ID
+ * @param status - 'change' if user comes from the check answers page, otherwise undefined
```
ttbarnes, line 22:
```suggestion
 * @param status - 'change' if user comes from the check answers page, otherwise 'cancellation - bank request date' 
```

trade-finance-manager-ui/server/controllers/case/cancellation/effective-from-date.controller.post.test.ts
ttbarnes: Same in other controllers

trade-finance-manager-ui/server/types/view-models/deal-cancellation/bank-request-date-view-model.ts
```diff
@@ -14,6 +14,7 @@ export type BankRequestDateValidationViewModel = {
 export type BankRequestDateViewModel = BaseViewModel & {
   ukefDealId: string;
   dealId: string;
+  previousPage: string;
```
ttbarnes, line 17: Not a blocker for this PR - seems like a lot of repetition in these view models - maybe `ukefDealId`, `dealId` and `previousPage` could go into `BaseViewModel` (i'm not sure what it's used for exactly - or maybe a new `DealViewModel`/`CaseViewModel`

trade-finance-manager-ui/server/controllers/case/cancellation/reason-for-cancelling.controller.ts
ttbarnes: Thank you both, agree on the points

Thing is, if they're moved to controller specific directories with a helpers directory - there's a lot of repeated code. All the new `getPreviousPageUrl` instances (in this PR at least), do exactly the same thing, except they have a different sub string at the end.

We could pass e.g a `route` param (for lack of a beter name) to a generic, DRY `getPreviousPageUrl`.

How about having this in a parent `/controllers/case/cancellation/helpers/get-previous-page-url` structure?

Then, the controllers become simpler and the controller tests just need to call `getPreviousPageUrl`, since we'd have specific unit tests for `getPreviousPageUrl`

Would that work / any other ideas? What do you think?

trade-finance-manager-ui/server/controllers/case/cancellation/effective-from-date.controller.ts
BethThomas141: This is the query param in the URL - so it's correct as is! (`?change` is at the end of the URL if the user has arrived at the page by clicking 'change' on the review page, otherwise there's nothing extra at the end of the URL)

ttbarnes: Sorry, was reading this as the return type 🤦

trade-finance-manager-ui/server/types/view-models/deal-cancellation/bank-request-date-view-model.ts
BethThomas141: Have made a `BaseCancellationViewModel` for the 3 view models to build up from 👍

trade-finance-manager-ui/server/controllers/case/cancellation/effective-from-date.controller.post.test.ts
BethThomas141: I'm indifferent either way - I think there's some value in having a fixed constant at the top of the file here. But I've move it so it's inline on controllers where it's only used once (and kept it at the top on two controllers where it's used twice, but renamed it to `previousPage` so it looks a bit neater passed in).

trade-finance-manager-ui/server/controllers/case/cancellation/reason-for-cancelling.controller.ts
BethThomas141: @ttbarnes Sure! Have pulled it out into the same `/controllers/case/cancellation/helpers/get-previous-page-url` helpers file that Mark's one lives in. Let me know what you think.

trade-finance-manager-ui/server/controllers/case/cancellation/helpers/get-previous-page-url-for-cancel-cancellation.ts
```diff
@@ -0,0 +1,40 @@
+/**
```
ttbarnes, line 1: Shall we change the file name to `get-previous-page-url` because we're already in the "case cancellation" directory", keep it simple?

trade-finance-manager-ui/server/controllers/case/cancellation/reason-for-cancelling.controller.ts
ttbarnes: Nice one @BethThomas141 👏

trade-finance-manager-ui/server/controllers/case/cancellation/helpers/get-previous-page-url.test.ts
```diff
@@ -45,9 +45,34 @@ describe('getPreviousPageUrl', () => {
...
+describe('getPreviousPageUrlForCancellationFlow', () => {
+  it('returns the correct relative url when status is not passed in', () => {
+    // Arrange
+    const defaultPreviousPage = `/defaultPreviousPage`;
+
+    // Act
+    const response = getPreviousPageUrlForCancellationFlow(dealId, defaultPreviousPage);
+
+    // Assert
+    expect(response).toBe(`/case/${dealId}/defaultPreviousPage`);
```
ttbarnes, line 64: Can we use `toEqual` throughout please

trade-finance-manager-ui/server/controllers/case/cancellation/reason-for-cancelling.controller.get.test.ts
```diff
@@ -92,6 +92,36 @@ describe('getReasonForCancelling', () => {
         ukefDealId,
         dealId,
         reasonForCancelling: reason,
+        previousPage: `/case/${dealId}/deal`,
+      });
+    });
+
+    it('renders the page with the back URL as the check details page when "change" is passed in as a query parameter', async () => {
```
ttbarnes, line 99: Seems that we don't need this test, can just update the test above to consume `getPreviousPageUrlForCancellationFlow` in `previousPage` 😎

trade-finance-manager-ui/server/controllers/case/cancellation/reason-for-cancelling.controller.post.test.ts
```diff
@@ -142,6 +142,38 @@ describe('postReasonForCancelling', () => {
           dealId,
           errors: validationErrors,
           reasonForCancelling,
+          previousPage: `/case/${dealId}/deal`,
+        });
+      });
+
+      it('renders the page with the back URL as the check details page when "change" is passed in as a query parameter', async () => {
```
ttbarnes, line 149: Seems that we don't need this test, can just update the test above to consume `getPreviousPageUrlForCancellationFlow` in `previousPage` 😎

trade-finance-manager-ui/server/controllers/case/cancellation/helpers/get-previous-page-url.test.ts
BethThomas141: Yes, have changed!

trade-finance-manager-ui/server/controllers/case/cancellation/helpers/get-previous-page-url-for-cancel-cancellation.ts
BethThomas141: Great point, have changed

trade-finance-manager-ui/server/controllers/case/cancellation/helpers/get-previous-page-url-for-cancel-cancellation.ts
```diff
@@ -0,0 +1,40 @@
...
+ * Validates the if the provided URL is one of
```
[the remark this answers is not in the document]
BethThomas141: Have fixed!

trade-finance-manager-ui/server/controllers/case/cancellation/reason-for-cancelling.controller.get.test.ts
BethThomas141: I'm keen to keep this as is - I like the reassurance of checking on the page level that different results of `getPreviousPageUrlForCancellationFlow` mean different parameters get passed in as part of the view model which requires two tests. 

The alternative to me would be mocking the result of `getPreviousPageUrlForCancellationFlow` and having two test for the two different results of the mocks. But given the preference it seems of minimising mocking, what I've got already here is my next best compromise!

So keen to leave as is if that's ok 👍

trade-finance-manager-ui/server/controllers/case/cancellation/reason-for-cancelling.controller.post.test.ts
BethThomas141: See above, hoping to leave this as is🤞

trade-finance-manager-ui/server/controllers/case/cancellation/reason-for-cancelling.controller.get.test.ts
ttbarnes: Really not a fan of this approach as we're testing things that are handled by function B, in function A.

Mocking non-async style helper functions introduces unnecessary test complexity.

As with other PRs / mentioned a few times:

1. The controller simply calls `getPreviousPageUrlForCancellationFlow`.
2. The controller test doesn't need to test anything about `getPreviousPageUrlForCancellationFlow`, just that it's been consumed.
3. With the current approach, if `getPreviousPageUrlForCancellationFlow` changes - the controller test needs to be updated, as well as the `getPreviousPageUrlForCancellationFlow` test - doubling time/effort

The testing approach is inconsistent throughout, including with the current approach in this PR, helper function mockings and the simpler (and IMO more effective, less maintanance) approaches. 

I don't want to block this PR so approving.

ttbarnes (approved):

trade-finance-manager-ui/server/controllers/case/cancellation/reason-for-cancelling.controller.get.test.ts
BethThomas141: Appreciate the response, seems like there are some different views around about how best to handle external function calls in unit tests. Given the non-block I'll leave as is here and resolve but maybe one for a wider discussion outside this PR.

trade-finance-manager-ui/server/controllers/case/cancellation/reason-for-cancelling.controller.post.test.ts
BethThomas141: As per previous comment thread, resolving here (and open to further discussion outside the PR)

trade-finance-manager-ui/server/controllers/case/cancellation/reason-for-cancelling.controller.get.test.ts
AlexBramhill: Hey both,

I agree, it'd be great to have a general chat about this as we seem to be discussing it a lot! 👍  Shall we catch up tomorrow?

PR #3648, opened 2024-10-07
feat(FN-2294): send email to bank if report reconciled

Zainzzkk:
## Introduction :pencil2:
This PR sends an email to the bank when a report is marked as reconciled

* Added helper to send an email once report is reconciled
* Added helper to generate email variables
* Added external-api call to dtfs-central-api to send emails
* Added unit tests

* Moved date format functions to libs/common


dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/helpers/generate-record-reconciled-email-variables.ts
```diff
@@ -0,0 +1,39 @@
...
+ * @param report
+ * @returns emails and variables for the email
+ */
+export const generateRecordReconciledEmailVariables = async (report: UtilisationReportEntity): Promise<FeeRecordReconciledEmail> => {
+  try {
+    const bank = await getBankById(report.bankId);
+
+    if (!bank) {
+      console.error('Bank not found: %s', report.bankId);
+      throw new Error('Bank not found');
```
RLavender98, line 18: Please can we throw a `NotFoundError` error here?

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/helpers/generate-record-reconciled-email-variables.ts
```diff
@@ -0,0 +1,39 @@
...
+      emails,
+      variables: {
+        bankRecipient: teamName,
+        reportReconciledDate: formattedSubmittedDate,
+        reportPeriod: reportPeriodString,
+      },
+    };
+  } catch (error) {
+    console.error('Error getting bank - generateRecordReconciledEmailVariables %o', error);
+    throw new Error('Error getting bank - generateRecordReconciledEmailVariables');
```
RLavender98, line 37: Similar to my other comment, please can we rethrow the original error here so our controller error handling can work as intended

dtfs-central-api/src/external-api/send-email.ts
```diff
@@ -0,0 +1,38 @@
...
+        templateId,
+        sendToEmailAddress,
+        emailVariables,
+      },
+    });
+
+    return response.data;
+  } catch (error) {
+    console.error('Error sending email to %s %o', sendToEmailAddress, error);
+    return false;
```
RLavender98, line 36: Similar to my other comments around catching - let's rethrow the original error please

RLavender98 (changes requested):
The email is when the report is reconciled not just a record so have left some rename suggestions about that. Also the error handling needs a few updates since we are losing some data with the current catching and then throwing a different error. I think the ApiError pattern may have been introduced since you were last on DTFS so please do reach out if you want to chat about it!

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/helpers/send-record-reconciled-email.ts
```diff
@@ -0,0 +1,29 @@
...
+import externalApi from '../../../../../external-api/api';
+import EMAIL_TEMPLATE_IDS from '../../../../../constants/email-template-ids';
+
+/**
+ * sends the record reconciled email to the bank
+ * generates the email variables and the emails to send to
+ * sends the email/s to the bank
+ * @param report
+ */
+export const sendRecordReconciledEmail = async (report: UtilisationReportEntity) => {
```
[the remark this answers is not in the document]
ttbarnes: Should this be `Emails` rather than `Email` since it is mapping over multiple emails from `generateRecordReconciledEmailVariables`?

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/mark-fee-records-as-reconciled/mark-fee-records-as-reconciled.event-handler.test.ts
```diff
@@ -23,16 +29,22 @@ describe('handleUtilisationReportMarkFeeRecordsAsReconciledEvent', () => {
...
+    externalApi.sendEmail = sendEmailSpy;
+    jest.mocked(getBankById).mockImplementation(jest.fn().mockResolvedValue(mockGetBankByIdResponse));
   });
 
   afterEach(() => {
     jest.resetAllMocks();
   });
 
-  it('calls the fee record state machine with the MARK_AS_RECONCILED event for every fee record to reconcile', async () => {
+  describe('when every free record has to be reconciled', () => {
```
ttbarnes, line 47: Looks like a typo, `free`?

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/mark-fee-records-as-reconciled/mark-fee-records-as-reconciled.event-handler.test.ts
```diff
@@ -43,47 +55,58 @@ describe('handleUtilisationReportMarkFeeRecordsAsReconciledEvent', () => {
...
-        requestSource,
-      },
+    it('should call call externalApi.sendEmail once', () => {
+      // Assert
+      expect(sendEmailSpy).toHaveBeenCalledTimes(1);
     });
   });
 
-  it('sets report status to RECONCILED if all fee records are now reconciled', async () => {
+  describe('when all fee records are now reconciled', () => {
```
ttbarnes, line 109:
```suggestion
  describe('when all fee records are reconciled', () => {
```

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/mark-fee-records-as-reconciled/mark-fee-records-as-reconciled.event-handler.test.ts
```diff
@@ -94,30 +117,40 @@ describe('handleUtilisationReportMarkFeeRecordsAsReconciledEvent', () => {
...
+    });
+
+    it('should call call externalApi.sendEmail once', () => {
+      // Assert
+      expect(sendEmailSpy).toHaveBeenCalledTimes(1);
+    });
   });
 
-  it('does not set report status to RECONCILED if not all fee records are now reconciled', async () => {
+  describe('when not all fee records are now reconciled', () => {
```
ttbarnes, line 153:
```suggestion
  describe('when only some fee records are reconciled', () => {
```

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/mark-fee-records-as-reconciled/mark-fee-records-as-reconciled.event-handler.test.ts
```diff
@@ -128,23 +161,33 @@ describe('handleUtilisationReportMarkFeeRecordsAsReconciledEvent', () => {
...
+      expect(mockFindOneOrFail).toHaveBeenCalledWith(UtilisationReportEntity, { where: { id: report.id }, relations: { feeRecords: true } });
+      expect(mockSave).not.toHaveBeenCalled();
+      expect(report).toEqual(
+        expect.objectContaining<Partial<UtilisationReportEntity>>({
+          status: 'RECONCILIATION_IN_PROGRESS',
+        }),
+      );
+    });
+
+    it('should NOT call call externalApi.sendEmail', () => {
```
ttbarnes, line 188: `call call`

dtfs-central-api/src/v1/email/index.ts
```diff
@@ -0,0 +1,24 @@
+import { AnyObject } from '@ukef/dtfs2-common';
+import { Response } from 'express';
+import externalApi from '../../external-api/api';
+
+export const sendEmail = async (templateId: string, sendToEmailAddress: string, emailVariables: AnyObject) => {
+  try {
+    if (!templateId) {
```
ttbarnes, line 7: Is it even possible for this to happen since `templateId` is a required type/param?

libs/common/src/helpers/date.test.ts
```diff
@@ -0,0 +1,37 @@
+import { formatInTimeZone } from 'date-fns-tz';
+import { getNowAsUtcISOString, getMonthName } from './date';
+
+describe('date-validation helpers', () => {
```
ttbarnes, line 4:
```suggestion
describe('date helpers', () => {
```

ttbarnes (changes requested):
Nice work, few initial comments

dtfs-central-api/src/v1/email/index.ts
Zainzzkk: This was taken from portal-api which does it

dtfs-central-api/src/v1/email/index.ts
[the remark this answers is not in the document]
Zainzzkk: this is taken from portal-api and no tests there for this either too

dtfs-central-api/src/v1/email/index.ts
ttbarnes: Ah, I see that the version of this in portal-api is in JS though, so we don't need this check in here because of TS

dtfs-central-api/src/v1/email/index.ts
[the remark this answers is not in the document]
RLavender98: Every function should have tests and since this is a new to dtfs-central-api we should add tests please

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/mark-fee-records-as-reconciled/mark-fee-records-as-reconciled.event-handler.ts
```diff
@@ -28,29 +32,42 @@ export const handleUtilisationReportMarkFeeRecordsAsReconciledEvent = async (
...
 
-  return report;
+    if (error instanceof ApiError) {
+      throw TransactionFailedError.forApiError(error);
+    }
+    if (error instanceof Error) {
+      throw TransactionFailedError.forError(error);
+    }
+    throw TransactionFailedError.forUnknownError();
+  }
```
RLavender98, line 72: This handling is already around this function call by nature of being within execute with SQl transaction so here we should just rethrow the original error:
```suggestion
throw error;
```

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/mark-fee-records-as-reconciled/mark-fee-records-as-reconciled.event-handler.ts
```diff
@@ -28,29 +31,35 @@ export const handleUtilisationReportMarkFeeRecordsAsReconciledEvent = async (
...
+      report.updateWithStatus({ status: 'RECONCILIATION_COMPLETED', requestSource });
+      await transactionEntityManager.save(UtilisationReportEntity, report);
+      await sendRecordReconciledEmail(report);
+    }
 
-  return report;
+    return report;
+  } catch (error) {
+    console.error(`Failed to mark fee records as reconciled - handleUtilisationReportMarkFeeRecordsAsReconciledEvent %o`, error);
+    throw new Error(`Failed to mark fee records as reconciled - handleUtilisationReportMarkFeeRecordsAsReconciledEvent`);
```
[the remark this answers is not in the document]
RLavender98: I've left a fresh comment on the latest change so resolving this comment

dtfs-central-api/src/types/utilisation-reports.ts
```diff
@@ -97,3 +97,14 @@ export type ValidatedPaymentDetailsFilters = {
   paymentCurrency?: Currency;
   paymentReference?: string;
 };
+
+export type FeeRecordReconciledEmailVariables = {
```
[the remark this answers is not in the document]
RLavender98: Resolving because leaving a fresh comment

dtfs-central-api/src/types/utilisation-reports.ts
```diff
@@ -97,3 +97,14 @@ export type ValidatedPaymentDetailsFilters = {
...
+  bankRecipient: string;
+  reportReconciledDate: string;
+  reportPeriod: string;
+};
+
+export type FeeRecordReconciledEmail = {
```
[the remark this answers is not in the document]
RLavender98: Resolving and leaving a fresh comment

dtfs-central-api/src/types/utilisation-reports.ts
```diff
@@ -97,3 +97,14 @@ export type ValidatedPaymentDetailsFilters = {
...
   paymentReference?: string;
 };
+
+export type FeeReportReconciledEmailVariables = {
+  bankRecipient: string;
+  reportReconciledDate: string;
+  reportPeriod: string;
+};
+
+export type FeeReportReconciledEmail = {
```
RLavender98, line 107: Why are these prefixed by the word `Fee`? Should probably be `UtilisationReportReconciledEmail` or just `ReportReconciledEmail`

dtfs-central-api/src/v1/email/index.ts
```diff
@@ -0,0 +1,21 @@
...
+    return emailSent;
+  } catch (error) {
+    console.error('DTFS-Central API - Failed to send email %o', error);
+    if (error instanceof ApiError) {
+      throw TransactionFailedError.forApiError(error);
+    }
+    if (error instanceof Error) {
+      throw TransactionFailedError.forError(error);
+    }
+    throw TransactionFailedError.forUnknownError();
```
RLavender98, line 19: This has nothing to do with a transaction and could very well be called outside of a transaction.
```suggestion
    throw error;
```

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/helpers/generate-report-reconciled-email-variables.ts
```diff
@@ -0,0 +1,47 @@
...
+ * @param report
+ * @returns emails and variables for the email
+ */
+export const generateReportReconciledEmailVariables = async (report: UtilisationReportEntity): Promise<FeeReportReconciledEmail> => {
+  try {
+    const bank = await getBankById(report.bankId);
+
+    if (!bank) {
+      console.error('Bank not found: %s', report.bankId);
+      throw new NotFoundError('Bank not found');
```
RLavender98, line 19: Please can we include the bank id in the error message? Will be useful if we are ever investigating this error from logs

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/helpers/generate-report-reconciled-email-variables.ts
```diff
@@ -0,0 +1,47 @@
...
+  } catch (error) {
+    console.error('Error getting bank - generateReportReconciledEmailVariables %o', error);
+
+    if (error instanceof ApiError) {
+      throw TransactionFailedError.forApiError(error);
+    }
+    if (error instanceof Error) {
+      throw TransactionFailedError.forError(error);
+    }
+    throw TransactionFailedError.forUnknownError();
```
RLavender98, line 45:
```suggestion
    throw error;
```

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/helpers/generate-record-reconciled-email-variables.ts
RLavender98: Resolving as leaving a new comment on the changes

dtfs-central-api/src/external-api/send-email.ts
```diff
@@ -0,0 +1,39 @@
+/* eslint-disable @typescript-eslint/no-unsafe-return */
```
RLavender98, line 1: We shouldn't be using eslint disables

dtfs-central-api/src/external-api/send-email.ts
```diff
@@ -0,0 +1,39 @@
...
+      },
+      data: {
+        templateId,
+        sendToEmailAddress,
+        emailVariables,
+      },
+    });
+
+    return response.data;
+  } catch (error: any) {
```
RLavender98, line 34: We shouldn't use `any` as a type unless absolute last resort. Is the type even needed? If so then does `Error` work? If it is required and that doesn't work then `unknown` is better.

RLavender98 (changes requested):
Just a couple of comments about the typing

dtfs-central-api/src/external-api/send-email.ts
Zainzzkk: this is a problem across the service and cannot be fixed now

dtfs-central-api/src/external-api/send-email.ts
Zainzzkk: committed by mistake, have removed

dtfs-central-api/src/external-api/send-email.ts
RLavender98: Ah ok, what is the problem?

Zainzzkk: removed return and lint disable

RLavender98 (approved):
LGTM!

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/mark-fee-records-as-reconciled/mark-fee-records-as-reconciled.event-handler.test.ts
ttbarnes: Resolving in place of a fresh comment

libs/common/src/helpers/date.test.ts
```diff
@@ -0,0 +1,37 @@
...
+      { monthNumber: 9, expectedMonthName: 'September' },
+      { monthNumber: 10, expectedMonthName: 'October' },
+      { monthNumber: 11, expectedMonthName: 'November' },
+      { monthNumber: 12, expectedMonthName: 'December' },
+    ])("should return month name '$expectedMonthName' from month number '$monthNumber'", ({ monthNumber, expectedMonthName }) => {
+      // Act
+      const monthName = getMonthName(monthNumber);
+
+      // Assert
+      expect(monthName).toEqual(expectedMonthName);
```
ttbarnes, line 34: `expect(result)` ?

Zainzzkk: this has been moved across - can change

libs/common/src/helpers/date.test.ts
```diff
@@ -0,0 +1,37 @@
...
+    });
```
ttbarnes, line 35: I am wondering if it we should change the assertion to use `format(date, 'MMMM');` 

At the moment `expectedMonthName` is essentially hardcoded/mimicking what the `format` does. 

We could simplify the `each` array to have only monthNumber and then the test can become this:

```ts
const result = getMonthName(monthNumber);

const expected = format(monthNumber, 'MMMM');
expect(result).toEqual(expectedMonthName);
```

What do you think?

RLavender98: I reckon whoever wrote this code wasn't aware that `format(monthNumber, 'MMMM');` behaved that way since the implementation is quite a bit over complicated in that case. We should probably actually change the implementation to call that but then keep the tests the same?

RLavender98: I don't think `const expected = format(monthNumber, 'MMMM');` will actually give you the correct month, I _think_ the first parameter is unix timestamp in millisecons so if you give it say number 2 as the month number it seems to return 'January' which is not we want.

RLavender98: I don't see a why to dynamically set the expected strings that doesn't just copy the function duplication, if there was a simpler way surely we should just simplify the function implementation?

ttbarnes: Test feels like it is testing what date-fns does for us.

We should construct dates in the test and call the date-fns function, would be simpler and more realistic.

It is currently like testing a lodash function or an express method - no need to.

Since this has simply moved from one place to another in this PR, happy to resolve.

ttbarnes (approved):

PR #3649, opened 2024-10-07
feat(DTFS2-7454): redirect from cancellation pages if no cancellation in draft

MarRobSoftwire:
## Introduction :pencil2:
When a user visits certain cancellation pages, if they dont have a draft cancellation already they should be redirected to the deal cancellation page. This should only happen if they use URL navigation, or browser history, rather than clicking buttons available on screen.

- Check the deal cancellation exists when visiting:
    - bank request date
    - effective from date page
    - cancel cancellation page
    - check details page
- The reason for cancelling page is the start of the journey so does not need a draft cancellation to exist
- Update unit tests

## Misc ➕ 
- Use `toEqual` not `toBe` in cancellation unit tests
- Create new `aRequestSession` mock data helper and use in cancellation unit tests

### DIFF
--- a/trade-finance-manager-ui/server/controllers/case/cancellation/bank-request-date.controller.get.test.ts
+++ b/trade-finance-manager-ui/server/controllers/case/cancellation/bank-request-date.controller.get.test.ts
@@ -59,30 +59,73 @@ describe('getBankRequestDate', () => {
     expect(res._getRedirectUrl()).toEqual(`/not-found`);
   });
 
-  it('redirects to deal summary page if the submission type is invalid (MIA)', async () => {
-    // Arrange
-    jest.mocked(api.getDeal).mockResolvedValue({ dealSnapshot: { details: { ukefDealId }, submissionType: DEAL_SUBMISSION_TYPE.MIA } });
-
-    const { req, res } = createMocks<GetBankRequestDateRequest>({
-      params: { _id: dealId },
-      session: {
-        user: mockUser,
-        userToken: 'a user token',
-      },
+  describe(`when the deal type is ${DEAL_SUBMISSION_TYPE.MIA}`, () => {
+    beforeEach(() => {
+      jest.mocked(api.getDeal).mockResolvedValue({ dealSnapshot: { details: { ukefDealId }, submissionType: DEAL_SUBMISSION_TYPE.MIA } });
     });
 
-    // Act
-    await getBankRequestDate(req, res);
+    it('redirects to deal summary page', async () => {
+      // Arrange
+      const { req, res } = createMocks<GetBankRequestDateRequest>({
+        params: { _id: dealId },
+        session: {
+          user: mockUser,
+          userToken: 'a user token',
+        },
+      });
 
-    // Assert
-    expect(res._getRedirectUrl()).toEqual(`/case/${dealId}/deal`);
+      // Act
+      await getBankRequestDate(req, res);
+
+      // Assert
+      expect(res._getRedirectUrl()).toEqual(`/case/${dealId}/deal`);
+    });
+
+    it('does not get the cancellation', async () => {
+      // Arrange
+      const { req, res } = createMocks<GetBankRequestDateRequest>({
+        params: { _id: dealId },
+        session: {
+          user: mockUser,
+          userToken: 'a user token',
+        },
+      });
+
+      // Act
+      await getBankRequestDate(req, res);
+
+      // Assert
+      expect(api.getDealCancellation).toHaveBeenCalledTimes(0);
+    });
   });
 
   describe.each([DEAL_SUBMISSION_TYPE.AIN, DEAL_SUBMISSION_TYPE.MIN])('when the deal type is %s', (validDealType) => {
-    it('renders the bank request date page without prepopulated data when it does not exist', async () => {
+    beforeEach(() => {
+      jest.mocked(api.getDeal).mockResolvedValue({ dealSnapshot: { details: { ukefDealId }, submissionType: validDealType } });
+    });
+
+    it('redirects to deal summary page if the deal cancellation is empty', async () => {
       // Arrange
       jest.mocked(api.getDealCancellation).mockResolvedValue({});
-      jest.mocked(api.getDeal).mockResolvedValue({ dealSnapshot: { details: { ukefDealId }, submissionType: validDealType } });
+
+      const { req, res } = createMocks<GetBankRequestDateRequest>({
+        params: { _id: dealId },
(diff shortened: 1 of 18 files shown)


trade-finance-manager-ui/server/controllers/case/cancellation/bank-request-date.controller.get.test.ts
```diff
@@ -59,30 +59,73 @@ describe('getBankRequestDate', () => {
...
+    it('redirects to deal summary page if the deal cancellation is empty', async () => {
       // Arrange
       jest.mocked(api.getDealCancellation).mockResolvedValue({});
-      jest.mocked(api.getDeal).mockResolvedValue({ dealSnapshot: { details: { ukefDealId }, submissionType: validDealType } });
+
+      const { req, res } = createMocks<GetBankRequestDateRequest>({
+        params: { _id: dealId },
+        session: {
+          user: mockUser,
+          userToken: 'a user token',
```
ttbarnes, line 115: Not a blocker for this PR - I think we should create a DRY `mockSession` as i'm seeing this exact same mock session in a lot of places.

What do you think - I can create a tech ticket and do this

trade-finance-manager-ui/server/controllers/case/cancellation/bank-request-date.controller.get.test.ts
```diff
@@ -59,30 +59,73 @@ describe('getBankRequestDate', () => {
...
+        },
+      });
+
+      // Act
+      await getBankRequestDate(req, res);
+
+      // Assert
+      expect(res._getRedirectUrl()).toBe(`/case/${dealId}/deal`);
```
ttbarnes, line 123: Can we use `toEqual` instead of `toBe` please

https://github.com/UK-Export-Finance/dtfs2/pull/3654

trade-finance-manager-ui/server/controllers/case/cancellation/bank-request-date.controller.ts
```diff
@@ -37,6 +38,10 @@ export const getBankRequestDate = async (req: GetBankRequestDateRequest, res: Re
 
     const cancellation = await api.getDealCancellation(_id, userToken);
 
+    if (isEmpty(cancellation)) {
```
ttbarnes, line 41: What do you think about creating some middleware instead of having to do this in individual controllers?

The middleware could be for the `/cancellation/*` route with a condition so that if it's the "first" route we just call `next()`? I think it'd be a lot simpler

ttbarnes (changes requested):
E2E tests?

trade-finance-manager-ui/server/controllers/case/cancellation/bank-request-date.controller.ts
MarRobSoftwire: There's currently no existing middleware that calls apis but I agree this could be really nice. 

I think this would need some thought about what pattern we'd want to follow & what benefits we'd gain - we could have a call sometime to chat about this but I think its outside the scope of this PR

trade-finance-manager-ui/server/controllers/case/cancellation/bank-request-date.controller.get.test.ts
MarRobSoftwire: Good idea, it looks like `trade-finance-manager-ui\server\controllers\utilisation-reports\edit-payment\index.test.ts` has a local function that could just be commonised & exported - I could do that now quite quickly if you like?

trade-finance-manager-ui/server/controllers/case/cancellation/bank-request-date.controller.ts
ttbarnes: Sorry for the long comment - I think there'd be a few benefits:

- None of the controllers need to make the same API call
- One middleware handles it all, DRY
- If we need to update the redirect or add/change some logic, can just update in one place

I think there's only 1/2 ways to create the middleware, e.g

```ts
const canCancelDealCancellation = async (req, res, next) => {
  if (isRelevantRoute()) {
    try {
      const { _id } = req.params;

      const cancellation = await api.getDealCancellation(_id, userToken);

      if (isEmpty(cancellation)) {
        return res.redirect(`/case/${_id}/deal`);
      }

      return next();
    } catch () {
      // ... 
    }
  }

  return next();
}
```

But this made me think - the controllers also call `api.getDealCancellation` to get a cancellation object. This would be required in the controllers and the middleware. So we could also create a second middleware that gets called first and stores the cancellation object in `res.locals` e.g

```ts
const getDealCancellation = async () => {
  if (isRelevantRoute()) {
    const cancellation = await api.getDealCancellation(_id, userToken);

    res.locals.dealCancellationData = cancellation;
  }

  return next();
};
```


This way we're doing the "get cancellation data" and "redirect if need to" type logic in one single place.

What do you think - shall we create a ticket for this? Happy to get this PR without such middleware.

ttbarnes: Strange one - feels like we're kind of introducing some tech debt rather than doing it straight away, but sometimes the implementation can highlight other ways to do it.

Happy to do this in another ticket. Can we create a JIRA ticket please?

MarRobSoftwire: I think this would be a pretty large change of pattern - no other middleware makes api-requests. I've tried to find any documentation talking about this &[ Next.js explicity advise against this](https://nextjs.org/docs/app/building-your-application/routing/middleware)
> Middleware is not designed for direct data fetching or manipulation, this should be done within Route Handlers or server-side utilities instead.

Moreover, this would require quite a few pieces of new middleware
- Get the deal, return 404 if it doesnt exist, add it to the `res`
- `validateDealCanBeCancelled` Validate the deal is MIN/AIN
- get the cancellation & add it to the `res`
- `checkCancellationNotEmpty` redirects if the deal is empty

If we do want to add a new pattern, I'd want to arrange a meeting with the other devs to decide what the scope of making this change is & how much dev time we'd want to commit to it.

ttbarnes: Agree with everything but the first point @MarRobSoftwire that's for NextJS, we're not using NextJS.

Fetching data in NodeJS middleware is very common - for example for user authentication.

Having the same kind of functions in routes - being called before controllers, is pretty much doing the same thing.

ttbarnes (approved):

PR #3650, opened 2024-10-07
fix(FN-3422): calculate fixed fee from report period end date

ChristianMcc:
## Introduction :pencil2:
We were calculating the fixed fee based on report period start instead of the day after report period end (1st of the next month)

Calculate fixed fee based on day after report period end


dtfs-central-api/src/services/state-machines/fee-record/event-handlers/helpers/calculate-fixed-fee.ts
```diff
@@ -8,19 +8,16 @@ import { getDateFromMonthAndYear, MonthAndYear, ReportPeriod } from '@ukef/dtfs2
  */
 const BANK_ADMIN_FEE_ADJUSTMENT = 0.9;
 
-const getNumberOfDaysRemainingInCoverPeriod = (reportPeriodStart: MonthAndYear, coverStartDate: Date, coverEndDate: Date): number => {
-  const currentReportPeriodStartMonthStart = startOfMonth(getDateFromMonthAndYear(reportPeriodStart));
+const getNumberOfDaysRemainingInCoverPeriod = (reportPeriodEnd: MonthAndYear, coverEndDate: Date): number => {
+  const endDateOfReportPeriod = startOfDay(endOfMonth(getDateFromMonthAndYear(reportPeriodEnd)));
```
ChristianMcc, line 12: I'd like to get people's thoughts here on readability. 
endOfMonth will give something like 2024-10-31:23:59:59 and differenceInDays only does whole day difference, so if I want to find the number of days between the 1st day after the reporting period and cover end date I can either do:
```
const endDate = endOfMonth(getDateFromMonthAndYear(reportPeriodEnd))
return differenceInDays(coverEndDate, startDateOfNextReportPeriod);
```

Or the explicit startOfDay then adding a day style thing I've done here to show exactly what date I care about starting from (and not obscuring the kind of 23:59:59 behaviour of endOfMonth)

More code but maybe more readable?

dtfs-central-api/src/services/state-machines/fee-record/event-handlers/helpers/calculate-fixed-fee.ts
```diff
@@ -8,19 +8,16 @@ import { getDateFromMonthAndYear, MonthAndYear, ReportPeriod } from '@ukef/dtfs2
...
+  const startDateOfNextReportPeriod = addDays(endDateOfReportPeriod, 1);
 
-  if (isBefore(currentReportPeriodStartMonthStart, coverStartDate)) {
```
ChristianMcc, line 14: Can get rid of this as you can't have coverStartDate after reportPeriod.end

dtfs-central-api/src/services/state-machines/fee-record/event-handlers/helpers/calculate-fixed-fee.test.ts
```diff
@@ -16,59 +13,30 @@ describe('calculateFixedFeeAdjustment', () => {
     jest.resetAllMocks();
   });
 
-  describe('when the report period starts before the cover start date', () => {
```
ChristianMcc, line 19: This is an edge case we can't have anymore :)

e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/generate-keying-sheet.spec.js
```diff
@@ -45,15 +46,15 @@ context('PDC_RECONCILE users can generate keying data', () => {
     .build();
 
   const ONE_YEAR_AGO = subYears(TODAY, 1);
-  const ONE_YEAR_FROM_CURRENT_REPORT_PERIOD_START = addYears(START_OF_CURRENT_REPORT_PERIOD, 1); // days left in cover period = 365
+  const ONE_YEAR_FROM_CURRENT_REPORT_PERIOD_END = addDays(END_OF_CURRENT_REPORT_PERIOD, 365); // days left in cover period = 365
```
ChristianMcc, line 49: I didn't want to have to worry about leap years so changed this

dtfs-central-api/src/services/state-machines/fee-record/event-handlers/helpers/calculate-fixed-fee.ts
RLavender98: I think I prefer the `startOfDay` and then adding a day as you have done as otherwise I would have read it as being a day out. 

That said, in either case it might be good to leave a comment in the code since I think it is unlikely that all future maintainers will be fully aware of the specifics of this?

e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/generate-keying-sheet.spec.js
RLavender98: Ah nice catch!

RLavender98 (dismissed):

dtfs-central-api/src/services/state-machines/fee-record/event-handlers/helpers/calculate-fixed-fee.ts
ChristianMcc: Good shout, have added documentation to say it's calculating day after report period end to end of cover date (think that's enough without a bunch of info on how date-fns works)

dtfs-central-api/src/services/state-machines/fee-record/event-handlers/helpers/calculate-fixed-fee.test.ts
```diff
@@ -16,59 +13,30 @@ describe('calculateFixedFeeAdjustment', () => {
...
-        reportPeriod,
-        coverStartDate: coverStartDateBeforeReportPeriod,
-        coverEndDate,
-      });
-
-      // Assert
-      expect(result).toBe(5326.03); // 100000 * (5 / 100) * 0.9 * 432 / 365 = 5,326.02739... = 5,326.03 (ROUNDED)
-    });
+    // Assert
+    expect(result).toBe(2626.03); // 100000 * (5 / 100) * 0.9 * 213 / 365 = 2,626.0274 = 2,626.03 (ROUNDED)
```
ttbarnes, line 40: Can we use `toEqual` throughout please

dtfs-central-api/src/services/state-machines/fee-record/event-handlers/helpers/calculate-fixed-fee.test.ts
```diff
@@ -16,59 +13,30 @@ describe('calculateFixedFeeAdjustment', () => {
     jest.resetAllMocks();
   });
 
-  describe('when the report period starts before the cover start date', () => {
-    it('returns the product of the utilisation, interest percentage (divided by 100), bank admin fee percentage (fixed at 0.9) and number of days remaining in the cover period divided by the day count basis and rounds to 2 decimal places', () => {
-      // Arrange
-      const utilisation = 100000;
-      const interestPercentage = 5;
-      const dayCountBasis = 365;
+  it('returns the product of the utilisation, interest percentage (divided by 100), bank admin fee percentage (fixed at 0.9) and number of days remaining in the cover period divided by the day count basis and rounds to 2 decimal places', () => {
```
ttbarnes, line 16: Could we simplify this, I don't think we need the exact description - if the calculation changes, it'd be easy for this to become out of sync. Maybe something like this instead:

```ts
it('should return the correct calculation'
```

dtfs-central-api/src/services/state-machines/fee-record/event-handlers/helpers/calculate-fixed-fee.ts
```diff
@@ -8,19 +8,23 @@ import { getDateFromMonthAndYear, MonthAndYear, ReportPeriod } from '@ukef/dtfs2
...
-  const currentReportPeriodStartMonthStart = startOfMonth(getDateFromMonthAndYear(reportPeriodStart));
+/**
+ * Gets the number of days remaining in cover period by calculating difference
+ * between the day after the end of the report period and the cover end date
+ * @param reportPeriodEnd - The report period end
+ * @param coverEndDate - The cover end date of the facility
+ * @returns The number of days remaining in the cover period
+ */
+const getNumberOfDaysRemainingInCoverPeriod = (reportPeriodEnd: MonthAndYear, coverEndDate: Date): number => {
+  const endDateOfReportPeriod = startOfDay(endOfMonth(getDateFromMonthAndYear(reportPeriodEnd)));
```
ttbarnes, line 19:
```suggestion
  const date = getDateFromMonthAndYear(reportPeriodEnd);

  const endDateOfReportPeriod = startOfDay(endOfMonth(date));
```

ttbarnes (changes requested):
Nice work. Few minor comments

ttbarnes (dismissed):

dtfs-central-api/src/services/state-machines/fee-record/event-handlers/helpers/calculate-fixed-fee.test.ts
ChristianMcc: good shout, at some point we should just get rid of all the toBes remaining in one fel sweep, I like the idea of no overhead in worrying between toBe and toEqual!

ttbarnes: It was already done 😎  https://github.com/UK-Export-Finance/dtfs2/pull/3654

ChristianMcc: oh sweet! nice :D

ttbarnes (approved):

PR #3654, opened 2024-10-08
chore(DTFS2-7462): replace toBe assertions with toEqual

ttbarnes:
## Introduction :pencil2:
- We have 5463 instances of `toEqual`.
- We have 1353 instances of `toBe`.
- `toEqual` is better thatn `toBe`.

- Replace all instances of `toBe` with `toEqual`

### DIFF
--- a/dtfs-central-api/api-tests/v1/durable-functions/durable-functions.api-test.js
+++ b/dtfs-central-api/api-tests/v1/durable-functions/durable-functions.api-test.js
@@ -56,7 +56,7 @@ describe('DELETE /v1/portal/durable-functions', () => {
       })
       .to(`/v1/portal/durable-functions`);
 
-    expect(firstStatus).toBe(200);
-    expect(secondStatus).toBe(200);
+    expect(firstStatus).toEqual(200);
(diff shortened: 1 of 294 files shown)


dtfs-central-api/api-tests/v1/durable-functions/durable-functions.api-test.js
```diff
@@ -56,7 +56,7 @@ describe('DELETE /v1/portal/durable-functions', () => {
       })
       .to(`/v1/portal/durable-functions`);
 
-    expect(firstStatus).toBe(200);
-    expect(secondStatus).toBe(200);
+    expect(firstStatus).toEqual(200);
```
Zainzzkk, line 59: should we change these to HttpStatusCode.Ok?

Zainzzkk (approved):

dtfs-central-api/api-tests/v1/durable-functions/durable-functions.api-test.js
ttbarnes: That'd be good but I think should be done separately as there are ~400 instances that could be replaced.

PR #3655, opened 2024-10-08
fix(DTFS2-6616): facility covenant and guarantee record variable correction

abhi-markan:
## Introduction :pencil2:
Facility covenant and guarantee amendments failed due to an `undefined` facility identifier, this PR aims to resolve this by correcting the variable name which supplied to the endpoint URI.

* Amended to correct variable name.

* Minor documentation updates.

### DIFF
--- a/azure-functions/acbs-function/src/functions/activity-update-facility-guarantee.js
+++ b/azure-functions/acbs-function/src/functions/activity-update-facility-guarantee.js
@@ -1,37 +1,21 @@
-/*
- * Facility guarantee amendment DAF
- * ***************************
- * This DAF (Durable Activity Function) is never invoked directly.
- * It is invoked via DOF (Durable Orchestrator Function).
- *
- * Pre-requisites
- * --------------
- * 0. 'npm install durable-functions'
- * 1. Durable  HTTP trigger function (acbs-http)
- * 2. Durable Orchestrator function (DOF) (acbs-amend-facility)
- *
- * ------------------
- * HTTP -> DOF -> DAF
- * ------------------
- */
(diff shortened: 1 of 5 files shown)


Zainzzkk (dismissed):

azure-functions/acbs-function/src/functions/activity-update-facility-guarantee.js
```diff
@@ -1,56 +1,43 @@
...
     const submittedToACBS = getNowAsIsoString();
-    const { status, data } = await api.updateFacilityGuarantee(facilityIdentifier, acbsFacilityGuaranteeInput);
+    const { status, data } = await api.updateFacilityGuarantee(facilityId, acbsFacilityGuaranteeInput);
 
     if (isHttpErrorStatus(status)) {
       throw new Error(
         JSON.stringify(
           {
             name: 'ACBS Facility Guarantee amend error',
-            facilityIdentifier,
```
ChristianMcc, line 53: how come this isn't included as facilityId in the error message anymore?

ChristianMcc (approved):
LGTM, minor question as a comment. I'm missing a bit of context of why the changes are in current state but they all look sensible and improved documentation.

Am I correct in thinking with this current approach of the ACBS function if you do an amendment and the first steps (FLR and FMR) pass then it fails then the amendment will only be partially in ACBS?

azure-functions/acbs-function/src/functions/activity-update-facility-guarantee.js
abhi-markan: It is in `data` object.

PR #3656, opened 2024-10-08
feat(FN-2311): split premium payments and payment detail types

sw-mattw:
## Introduction :pencil2:
The premium payments and payment details tabs have two separate tables. The row data for these tables is currently passed around as the same type, despite these two tables containing slightly different data.

Now that the table on the payment details tab is filterable, this is causing some issues. The row data is currently passed through as 'fee record payment groups' which can contain multiple payments and multiple fee records. The payment details table displays one payment per row, however, the current type returns the aforementioned groups from the backend and then maps these into separate rows in the frontend. This causes issues when we filter by payment reference as we only want the backend to return individual payments (rather than the whole group it belongs to (which may include other payments with different payment references)).

The original implementation of these types requires updating, along with the mapping for these tabs.

- Split `FeeRecordPaymentGroup` into `PremiumPaymentsGroup` and `PaymentDetails`
- Updated `getPaymentDetails` function to flatten the retrieved `FeeRecordPaymentEntityGroup` objects to ensure we have at most one payment per group
- Updated mappers and templates to handle these new types
- Add e2e tests for payment details filtering

N/A


dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-report-reconciliation-details-by-id.controller/helpers/get-utilisation-report-reconciliation-details.test.ts
```diff
@@ -257,6 +264,80 @@ describe('get-utilisation-report-reconciliation-details-by-id.controller helpers
...
+
+          const paymentDetailsFilters = {
+            facilityId: '12345678',
+          };
+
+          // Act
+          await getPaymentDetails(groups, paymentDetailsFilters);
+
+          // Assert
+          expect(filterFeeRecordSpy).toHaveBeenCalledWith(expectedFlattenedGroups, paymentDetailsFilters);
```
Zainzzkk, line 336: is it worth testing how many times the spy has been called for both tests?

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-report-reconciliation-details-by-id.controller/helpers/map-to-premium-payments-groups.ts
```diff
@@ -0,0 +1,50 @@
...
+ * @param feeRecordPaymentEntityGroups - The fee record payment entity groups
+ * @returns The premium payments groups
+ * @throws {Error} If the group has no payments but has multiple fee records
+ */
+export const mapToPremiumPaymentsGroups = (feeRecordPaymentEntityGroups: FeeRecordPaymentEntityGroup[]): PremiumPaymentsGroup[] =>
+  feeRecordPaymentEntityGroups.map((group) => {
+    const groupStatus = getFeeRecordPaymentEntityGroupStatus(group);
+
+    if (group.payments.length === 0) {
+      // If there are no payments, there is only one fee record in the group
```
ttbarnes, line 18: Could/should we move the proceeding logic into it's own function? Would make `mapToPremiumPaymentsGroups` simpler

e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/filter-payment-details.spec.js
```diff
@@ -26,52 +28,693 @@ context(`users can filter payment details by facility id and payment reference a
...
+      // On first page load, the filter panel should be visible.
+      pages.utilisationReportPage.paymentDetailsTab.filters.panel().should('be.visible');
+
+      cy.assertText(pages.utilisationReportPage.paymentDetailsTab.filters.panelToggleButton(), 'Hide filters');
+      pages.utilisationReportPage.paymentDetailsTab.filters.panelToggleButton().click();
+      pages.utilisationReportPage.paymentDetailsTab.filters.panel().should('not.be.visible');
+
+      cy.assertText(pages.utilisationReportPage.paymentDetailsTab.filters.panelToggleButton(), 'Show filters');
+      pages.utilisationReportPage.paymentDetailsTab.filters.panelToggleButton().click();
+      pages.utilisationReportPage.paymentDetailsTab.filters.panel().should('be.visible');
```
ttbarnes, line 57: Could we destructure so easier to read? E.g

```js
filters.panel().should('be.visible');

cy.assertText(filters.panelToggleButton(), 'Hide filters');

filters.panelToggleButton().click();
filters.panel().should('not.be.visible');

cy.assertText(filters.panelToggleButton(), 'Show filters');

filters.panelToggleButton().click();
filters.panel().should('be.visible');
```

e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/filter-payment-details.spec.js
```diff
@@ -26,52 +28,693 @@ context(`users can filter payment details by facility id and payment reference a
...
+        pages.utilisationReportPage.paymentDetailsTab.filters.facilityIdInput().should('have.value', partialFacilityIdFilter);
+
+        pages.utilisationReportPage.paymentDetailsTab.paymentDetailsTable.row(firstPayment.id, firstFeeRecord.id).should('exist');
+        pages.utilisationReportPage.paymentDetailsTab.paymentDetailsTable.row(secondPayment.id, secondFeeRecord.id).should('not.exist');
+      });
+    });
+
+    describe('when the facility id filter is submitted with an invalid value', () => {
+      it('should display an error message for invalid facility id and persist the inputted value', () => {
+        const invalidFacilityIdFilter = '123'; // Invalid due to length.
```
ttbarnes, line 205: Are we able to use a constant here e.g MIN - 1? I think we have a "minimum" constant somewhere already

e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/filter-payment-details.spec.js
```diff
@@ -26,52 +28,693 @@ context(`users can filter payment details by facility id and payment reference a
...
+    };
+
+    describe('when a payment currency filter is submitted', () => {
+      it('should only display the payment with the supplied payment currency and persist the checked currency', () => {
+        const paymentCurrencyFilter = CURRENCY.USD;
+
+        const firstFeeRecord = FeeRecordEntityMockBuilder.forReport(utilisationReport).withId(1).withFacilityId('11111111').build();
+        const secondFeeRecord = FeeRecordEntityMockBuilder.forReport(utilisationReport).withId(2).withFacilityId('22222222').build();
+
+        const firstPayment = PaymentEntityMockBuilder.forCurrency(paymentCurrencyFilter)
```
ttbarnes, line 317: Seems to be a lot of duplication with the entity builders, are we able to make it more DRY?

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-report-reconciliation-details-by-id.controller/helpers/map-to-payment-details.test.ts
```diff
@@ -0,0 +1,161 @@
...
+  beforeEach(() => {
+    findTfmUserSpy.mockRejectedValue('Some error');
+    getFeeRecordPaymentEntityGroupReconciliationDataSpy.mockResolvedValue({});
+  });
+
+  afterEach(() => {
+    jest.resetAllMocks();
+  });
+
+  it('populates each of the groups with their respective group status', async () => {
```
Zainzzkk, line 29: can all its in this PR start it `it('should...`. please

e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/filter-payment-details.spec.js
ttbarnes: Actually if we could destructure all of the selectors in this test (to some degree so it still makes sense) I think the test will be easier to read/maintain, lots of long lines atm e.g

```js
pages.utilisationReportPage.paymentDetailsTab.paymentDetailsTable.row(firstPayment.id, firstFeeRecord.id).should('exist');
pages.utilisationReportPage.paymentDetailsTab.paymentDetailsTable.row(secondPayment.id, secondFeeRecord.id).should('not.exist');
```

e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/filter-payment-details.spec.js
```diff
@@ -26,52 +28,693 @@ context(`users can filter payment details by facility id and payment reference a
...
+      pages.utilisationReportPage.paymentDetailsTab.filters.paymentCurrencyRadioInput(paymentCurrencyFilter).should('be.checked');
+      pages.utilisationReportPage.paymentDetailsTab.filters.paymentReferenceInput().should('have.value', partialPaymentReferenceFilter);
+      pages.utilisationReportPage.paymentDetailsTab.filters.facilityIdInput().should('have.value', partialFacilityIdFilter);
+
+      pages.utilisationReportPage.paymentDetailsTab.paymentDetailsTable.row(firstPayment.id, firstFeeRecord.id).should('exist');
+      pages.utilisationReportPage.paymentDetailsTab.paymentDetailsTable.row(secondPayment.id, secondFeeRecord.id).should('not.exist');
+      pages.utilisationReportPage.paymentDetailsTab.paymentDetailsTable.row(thirdPayment.id, thirdFeeRecord.id).should('exist');
+      pages.utilisationReportPage.paymentDetailsTab.paymentDetailsTable.row(fourthPayment.id, thirdFeeRecord.id).should('not.exist');
+      pages.utilisationReportPage.paymentDetailsTab.paymentDetailsTable.row(fourthPayment.id, fourthFeeRecord.id).should('not.exist');
+    });
```
ttbarnes, line 718: This test is huge, just over 700 lines 🤯 

Can we split it up please? 

How long does it take to run?

trade-finance-manager-ui/server/controllers/utilisation-reports/helpers/reconciliation-for-report-helper.test.ts
```diff
@@ -128,15 +128,15 @@ describe('reconciliation-for-report-helper', () => {
...
         {
-          ...aFeeRecordPaymentGroup(),
+          ...aPremiumPaymentsGroup(),
           feeRecords: [firstFeeRecord, secondFeeRecord],
         },
       ];
 
       // Act
-      const viewModel = mapFeeRecordPaymentGroupsToPremiumPaymentsViewModelItems(feeRecordPaymentGroups, DEFAULT_IS_CHECKBOX_SELECTED);
+      const viewModel = mapPremiumPaymentsGroupsToPremiumPaymentsViewModelItems(premiumPaymentGroups, DEFAULT_IS_CHECKBOX_SELECTED);
```
ttbarnes, line 139: Can we shorten the function name?

ttbarnes (changes requested):

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-report-reconciliation-details-by-id.controller/helpers/get-utilisation-report-reconciliation-details.ts
```diff
@@ -26,29 +26,35 @@ export const getPremiumPayments = async (feeRecordPaymentEntityGroups: FeeRecord
...
+
+  // Flatten groups to apply filters to individual payments rather than entire groups.
+  let paymentsWithFeeRecords = feeRecordPaymentEntityGroups.flatMap((group) =>
+    group.payments.map((payment) => ({
+      payments: [payment],
+      feeRecords: group.feeRecords,
+    })),
+  );
+
+  if (facilityIdFilter || paymentCurrencyFilter || paymentReferenceFilter) {
```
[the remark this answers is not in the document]
sw-mattw: Yep, there are unit tests from another PR that covers this if statement, both when only a single filter is applied and when multiple are applied:

https://github.com/UK-Export-Finance/dtfs2/blob/0ab38f7978763ccc63cacb6605fed47aa8d08b41/dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-report-reconciliation-details-by-id.controller/helpers/get-utilisation-report-reconciliation-details.test.ts#L222-L241

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-report-reconciliation-details-by-id.controller/helpers/get-utilisation-report-reconciliation-details.test.ts
```diff
@@ -257,6 +264,80 @@ describe('get-utilisation-report-reconciliation-details-by-id.controller helpers
...
+          // Arrange
+          const mapToPaymentDetailsSpy = jest.spyOn(mapToPaymentDetailsModule, 'mapToPaymentDetails').mockResolvedValue([]);
+
+          const paymentDetailsFilters = {};
+
+          // Act
+          await getPaymentDetails(groups, paymentDetailsFilters);
+
+          // Assert
+          expect(mapToPaymentDetailsSpy).toHaveBeenCalledWith(expectedFlattenedGroups);
```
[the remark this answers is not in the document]
ttbarnes: Resolving this, covered in another comment https://github.com/UK-Export-Finance/dtfs2/pull/3656#discussion_r1796794815

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-report-reconciliation-details-by-id.controller/helpers/get-utilisation-report-reconciliation-details.test.ts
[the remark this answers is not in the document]
ttbarnes: Resolving this, covered in another comment https://github.com/UK-Export-Finance/dtfs2/pull/3656#discussion_r1796794815

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-report-reconciliation-details-by-id.controller/helpers/map-to-premium-payments-groups.ts
sw-mattw: Split this logic out into two sub cases:
- `mapGroupWithNoPaymentsToPremiumPaymentsGroup`
- `mapGroupWithPaymentsToPremiumPaymentsGroup`
The main mapping function is now a lot simpler, thanks for the suggestion

e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/filter-payment-details.spec.js
sw-mattw: This one uses an existing regex which includes the length check, but I've updated the other invalid filter value for payment reference to use the `MIN_` constant that was introduced

e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/filter-payment-details.spec.js
sw-mattw: It was around 33s, but I agree in retrospect it should've been split up as it's rather large. I've split this by filter, and kept a single file for generic checks.

e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/filter-payment-details.spec.js
sw-mattw: Simplified these a lot thanks, please see the new spec tests

e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/filter-payment-details.spec.js
ttbarnes: thanks @sw-mattw i'm not following though, i'm thinking something like this?

```
// const invalidFacilityIdFilter = '123';

const invalidFacilityIdFilter = MIN + 1;
```

dtfs-central-api/test-helpers/test-data/fee-record-payment-entity-group.ts
```diff
@@ -0,0 +1,24 @@
+import { FeeRecordStatus, Currency, FeeRecordEntityMockBuilder, UtilisationReportEntityMockBuilder } from '@ukef/dtfs2-common';
+import { FeeRecordPaymentEntityGroup } from '../../src/types/fee-record-payment-entity-group';
+
+export const createFeeRecordEntityPaymentGroupForSingleFeeRecord = (
```
ttbarnes, line 4: I'm a bit confused here: 

- File name is `fee-record-payment-entity-group`
- Function name is `...Fee Record Entity Payment Group For Single Fee Record`

As in one says "fee record entity ...", but the other says "fee record payment entity ..."

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-report-reconciliation-details-by-id.controller/helpers/map-to-payment-details.ts
```diff
@@ -0,0 +1,41 @@
...
+ * These should be flattened such that there is at most one payment in each group.
+ * @returns The payment details
+ * @throws {Error} If the group has more than one payment
+ */
+export const mapToPaymentDetails = async (paymentsWithFeeRecords: FeeRecordPaymentEntityGroup[]): Promise<PaymentDetails[]> => {
+  const mappedPaymentDetails: PaymentDetails[] = [];
+
+  for (const group of paymentsWithFeeRecords) {
+    if (group.payments.length > 1) {
+      throw new Error('Error mapping payments to payment details - groups must have at most one payment.');
```
ttbarnes, line 19:
```suggestion
      throw new Error('Error mapping group payments to payment details - group has more than one payment.');
```

dtfs-central-api/test-helpers/test-data/fee-record-payment-entity-group.ts
sw-mattw: Sorry yes, thanks for spotting, I've updated the naming to match the return type of `FeeRecordPaymentEntityGroup`

e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/filter-payment-details.spec.js
sw-mattw: Apologies, the minimum constant that was recently introduced was for the payment reference filter (`MIN_PAYMENT_REFERENCE_FILTER_CHARACTER_COUNT`) rather than the facility ID filter.

The way we're checking for a valid facility ID filter value is by using the existing `PARTIAL_FACILITY_ID: /^\d{4,10}$/,` regex which includes the length requirements of 4-10 digits in the string we're checking.

The facility ID filter value is a string and we're looking to test that an invalid value is rejected, for this case it's a string that is one character short. As we don't have a constant for the minimum number of digits I'm not sure we can do what you're suggesting (if I'm understanding correctly).

We're testing any invalid value is rejected so maybe we could update this const to be alphabetical characters (eg `aaaa`) instead? (as there's unit tests on the validation function itself to test it validates as expected so we could use any invalid case)

Hope that clarifies things, please let me know if not and we can have a chat

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-report-reconciliation-details-by-id.controller/helpers/map-to-payment-details.ts
sw-mattw: We don't at the minute - unfortunately we don't have any group or correlation ID from the point at which we get these groups in `getFeeRecordPaymentEntityGroups` - below:

https://github.com/UK-Export-Finance/dtfs2/blob/df928779cc08be8ce401bab871fdca921416af05/dtfs-central-api/src/helpers/get-fee-record-payment-entity-groups.ts#L18


I think the best logging we can add is logging the whole group just before we throw this error, with a similar error message - I've added this, open to comments on how you feel about this thanks

I've also made the update to the error message thanks

ttbarnes: Nice one thanks @sw-mattw 

I'm not sure about the object logging, but it's the only sensible thing we can log I think it's better than nothing

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-utilisation-report-reconciliation-details-by-id.controller/helpers/map-to-payment-details.test.ts
sw-mattw: Updated all of these thanks and surrounded by `describe('when...` where possible

ttbarnes (approved):

PR #3659, opened 2024-10-09
feat(dtfs2-6892): create and auto submit accept sso redirect form

AlexBramhill:
## Introduction :pencil2:
This PR sets up changes ready for work to handles the autosubmitted form and sends check to sso authority

- update naming of folders to be kebab-case
- update schema tests to have nicer indentation, present dev with tests faster through moving functions to bottom of test file


trade-finance-manager-ui/server/schemas/with-entra-id-user-schema.tests.ts
```diff
@@ -0,0 +1,149 @@
+import { ZodObject } from 'zod';
+import { withSchemaTests } from '../../test-helpers';
+import { EntraIdUser } from '../types/entra-id-user';
+import { ENTRA_ID_USER_GROUP } from '../constants/entra-id-user';
+
+type TestCasesParams = {
+  getValidObjectWithUpdatedEntraIdUserParams: (entraIdUser: unknown) => unknown;
```
ttbarnes, line 7: Can we replace `unknown` with a real type?

trade-finance-manager-ui/server/schemas/entra-id-user.schema.entra-id-user-schema.test.ts
```diff
@@ -0,0 +1,9 @@
+import { EntraIdUserSchema } from './entra-id-user.schema';
+import { withEntraIdUserSchemaTests } from './with-entra-id-user-schema.tests';
+
+describe('EntraIdUserSchema', () => {
+  withEntraIdUserSchemaTests({
+    schema: EntraIdUserSchema,
+    getValidObjectWithUpdatedEntraIdUserParams: (entraIdUser) => entraIdUser,
```
ttbarnes, line 7: What is the purpose of `getValidObjectWithUpdatedEntraIdUserParams` please? Seems to just return the same thing?

trade-finance-manager-ui/server/schemas/with-entra-id-user-schema.tests.ts
```diff
@@ -0,0 +1,149 @@
...
+};
+
+type withEntraIdUserSchemaTestsParams = {
+  schema: ZodObject<any>;
```
ttbarnes, line 11: Can we replace `any` with a real type?

trade-finance-manager-ui/server/schemas/with-entra-id-user-schema.tests.ts
```diff
@@ -0,0 +1,149 @@
...
+        const { oid, ...rest } = aValidEntraIdUser();
+        return getValidObjectWithUpdatedEntraIdUserParams(rest);
+      },
+      description: 'the oid is missing',
+    },
+    {
+      aTestCase: () => {
+        // eslint-disable-next-line @typescript-eslint/no-unused-vars, camelcase
+        const { verified_primary_email, ...rest } = aValidEntraIdUser();
+        return getValidObjectWithUpdatedEntraIdUserParams(rest);
```
ttbarnes, line 49: Why do we need to do `...rest` please, `getValidObjectWithUpdatedEntraIdUserParams` seems to return what it's given, unless i'm missing something. If we can pass the whole user without `...rest` we could remove the eslint disables

ttbarnes (changes requested):

trade-finance-manager-ui/server/schemas/entra-id-user.schema.entra-id-user-schema.test.ts
AlexBramhill: It's to allow for commonisation of `EntraIdUserSchema` tests

trade-finance-manager-ui/server/schemas/with-entra-id-user-schema.tests.ts
AlexBramhill: This is a real type -- I've kept this rather wide to allow for compatibility with zod objects, but all ZodObjects can have `.safeParse` called on it (hence no type error!)

trade-finance-manager-ui/server/schemas/with-entra-id-user-schema.tests.ts
AlexBramhill: Spreading with `...rest` is a way to delete `verified_primary_email` without mutating objects.

This is a common pattern to get properties without using `delete` and having side effects or faffing with deep cloning (that's still hard to get perfect, even in 2024).

https://medium.com/coinmonks/object-rest-and-spread-in-typescript-90d8739705ae#:~:text=Object%20rest%20and%20spread%20are%20powerful%20features%20in%20TypeScript%20that,one%20object%20into%20another%20object.

https://mariusschulz.com/blog/object-rest-and-spread-in-typescript

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax

(and used in functions)
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/rest_parameters

Our eslint rules are erroring due to `verified_primary_email` being unused. Usually, you'd simply do `_verified_primary_email: verified_primary_email` and there would be no error, but I know we lean on not having this rule set, at the cost of having to do this manually

https://www.sonarsource.com/blog/common-typescript-issues-no-3-unused-local-variables-and-functions/#:~:text=Naming%20a%20variable%20that%20is,to%20name%20the%20variable%20though.

https://www.reddit.com/r/learnjavascript/comments/qwwibu/underscores_for_unused_parameters/

https://stackoverflow.com/questions/64052318/how-to-disable-warn-about-some-unused-params-but-keep-typescript-eslint-no-un

ttbarnes: @AlexBramhill I know what the spread operator does. I'm not understand _why_ we need to pass/exclude specific properties when calling `getValidObjectWithUpdatedEntraIdUserParams` because this function seems to just return whatever is passed to it?

```ts
getValidObjectWithUpdatedEntraIdUserParams: (entraIdUser) => entraIdUser,
```

AlexBramhill: Ah sorry, I misunderstood!

It would be pointless to have `getValidObjectWithUpdatedEntraIdUserParams` if the test helpers `with-entra-id-user-schema.tests` were just used in the `entra-user-id-schema` test file, as the `entra-user-id-schema` test file just passes the object straight through. I completely agree.

We also use `with-entra-id-user-schema.tests` in `entra-id-authentication-result-schema` -- which uses a different method (not just a passthrough).

`getValidObjectWithUpdatedEntraIdUserParams` is needed to create a veritile test helper to avoid having to repeat these tests every time, even when the `EntraIdUser` is nested in the object to be validated (`entra-id-authentication-result-schema` has this, for instance, set to configure `idTokenClaims`
```
    getValidObjectWithUpdatedEntraIdUserParams: (entraIdUser) => ({
      ...aValidPayload(),
      account: { idTokenClaims: entraIdUser },
    }),
```
)

The need for this is to avoid mocking the result of the validation of `entra user id` in `entra-id-authentication-result-schema` tests through creating a reusable test object.

As the test object is resuable and veristile, it's best to reuse it in `entra-id-user-schema` tests whereever we are testing `entra user id`. This means that there is 0 chance of the tests for `entra-id-user-schema` and anywhere else that needs to validate `entra-id-user-schema` getting out of sync

trade-finance-manager-ui/server/schemas/entra-id-user.schema.entra-id-user-schema.test.ts
AlexBramhill: (sorry @ttbarnes -- might not have seen that we use them in `entra-id-authentication-result-schema` validation as well)

trade-finance-manager-ui/server/schemas/with-entra-id-user-schema.tests.ts
AlexBramhill: I've been more specific here...update inbound

trade-finance-manager-ui/server/constants/entra-id-user.ts
```diff
@@ -0,0 +1,23 @@
...
+  'AZURE_SSO_GROUP_UNDERWRITING_SUPPORT',
+  'AZURE_SSO_GROUP_UNDERWRITER_MANAGERS',
+  'AZURE_SSO_GROUP_UNDERWRITERS',
+  'AZURE_SSO_GROUP_RISK_MANAGERS',
+  'AZURE_SSO_GROUP_BUSINESS_SUPPORT',
+  'AZURE_SSO_GROUP_PIM',
+  'AZURE_SSO_GROUP_PDC_READ',
+  'AZURE_SSO_GROUP_PDC_RECONCILE',
+] as const;
+
```
ttbarnes, line 11: Should this file maybe be called `entra-id-user-groups`?

AlexBramhill: renamed

ttbarnes (approved):

PR #3662, opened 2024-10-09
feat(FN-2976): add match success notification to template

RLavender98:
## Introduction :pencil2:
The current page let's you add a payment that brings to match status nad remain on the add payment page to add another payment, but when you try you will always fail since you are not allowed to add a payment to fee records in a match state. This PR addresses that.

- Return the fee record status after adding a payment
- If the user selected they wished to add another payment when adding a payment BUT that payment took the record(s) to match status THEN redirect to the premium payments page AND display a success notification

Screenshot 2024-10-09 at 14 57 38


dtfs-central-api/src/v1/controllers/utilisation-report-service/post-payment.controller/index.ts
```diff
@@ -28,7 +35,9 @@ export const postPayment = async (req: PostPaymentRequest, res: Response) => {
     };
 
     await addPaymentToUtilisationReport(parseInt(reportId, 10), feeRecordIds, user, newPaymentDetails);
-    return res.sendStatus(HttpStatusCode.Ok);
+    const { status } = await FeeRecordRepo.findOneByOrFail({ id: feeRecordIds[0] });
```
sw-mattw, line 38: I feel a bit mixed about this being included here vs in the transaction within `addPaymentToUtilisationReport`:

If `findOneByOrFail` rejects we'll end up in the catch which will produce an error on the frontend, but the payment would've been added.

It feels like this would ideally be in the same transaction as adding the payment (at the end of the same transaction that's being executed within `addPaymentToUtilisationReport`), then we can return it from the call to that maybe?

Will mean some of the tests need shifting too - apologies

sw-mattw (commented):
Looks good, added one comment on the position of a return

dtfs-central-api/src/v1/controllers/utilisation-report-service/post-payment.controller/index.ts
RLavender98: Good shout! Have made the change 😄

dtfs-central-api/src/v1/controllers/utilisation-report-service/post-payment.controller/helpers.ts
```diff
@@ -8,12 +9,20 @@ import { executeWithSqlTransaction } from '../../../../helpers';
 
 /**
  * Adds a payment to the utilisation report with the specified id
+ * and returns the resulting status of the fee record(s) it was
+ * added against
```
Zainzzkk, line 13: could this be on the line above?

e2e-tests/tfm/cypress/e2e/pages/utilisation-reports/utilisationReportPage.js
```diff
@@ -3,6 +3,8 @@ const utilisationReportPage = {
   keyingSheetTabLink: () => cy.get('a[data-cy="bank-report-tab-keying-sheet"]'),
   paymentDetailsLink: () => cy.get('a[data-cy="bank-report-tab-payment-details"]'),
   premiumPaymentsTab: {
+    matchSuccessNotificationHeading: () => cy.get('[data-cy="match-success-notification"] h3'),
+    matchSuccessNotificationMessage: () => cy.get('[data-cy="match-success-notification"] p'),
```
Zainzzkk, line 7: could these not have their own data-cys? -heading -message?

dtfs-central-api/src/v1/controllers/utilisation-report-service/post-payment.controller/index.ts
```diff
@@ -27,8 +33,9 @@ export const postPayment = async (req: PostPaymentRequest, res: Response) => {
       reference: paymentReference,
     };
 
-    await addPaymentToUtilisationReport(parseInt(reportId, 10), feeRecordIds, user, newPaymentDetails);
-    return res.sendStatus(HttpStatusCode.Ok);
+    const feeRecordStatus = await addPaymentToUtilisationReport(parseInt(reportId, 10), feeRecordIds, user, newPaymentDetails);
```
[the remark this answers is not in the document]
RLavender98: I don't think so, simplifying to `Number(reportId)`

e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/add-payment.spec.js
```diff
@@ -149,6 +149,34 @@ context('PDC_RECONCILE users can add a payment to a report', () => {
...
+
+    cy.keyboardInput(cy.getInputByLabelText('Day'), '12');
+
+    cy.keyboardInput(cy.getInputByLabelText('Month'), '12');
+
+    cy.keyboardInput(cy.getInputByLabelText('Year'), '2023');
+
+    cy.getInputByLabelText('Yes').click();
+
+    cy.contains('button', 'Continue').click();
```
Zainzzkk, line 166: sorry missed this.  Could we use continueButton from partials here?

e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/add-payment.spec.js
```diff
@@ -185,7 +186,7 @@ context('PDC_RECONCILE users can add a payment to a report', () => {
     cy.keyboardInput(cy.getInputByLabelText('Year'), '2023');
     cy.getInputByLabelText('Yes').click();
 
-    cy.contains('button', 'Continue').click();
+    continueButton().click();
```
ttbarnes, line 189: Sorry @RLavender98 should have mentioned - we can just use this command here and avoid the import 😎 

```js
cy.clickContinueButton();
```

Zainzzkk (approved):

sw-mattw (approved):

PR #3667, opened 2024-10-11
feat(FN-3484): download report link on utilisation tab

RLavender98:
## Introduction :pencil2:
Users need to be able to download the report contents for offline viewing

- Add link to the utilisation tab under the table to download report contents as a CSV


trade-finance-manager-ui/component-tests/utilisation-reports/utilisation-report-reconciliation-for-report.component-test.ts
```diff
@@ -304,4 +305,19 @@ describe(page, () => {
     wrapper.expectElement(`${utilisationTabSelector} table`).toExist();
     wrapper.expectElement(`${utilisationTabSelector} table tr[data-cy="utilisation-table-row-${feeRecordId}"]`);
   });
+
+  it('should render the download link for the report', () => {
+    const downloadUrl = 'utilisation-reports/12345/download';
```
ttbarnes, line 310: Could we move this to the top and use in line 39 and 299? DRY

trade-finance-manager-ui/server/routes/utilisation-reports/index.ts
```diff
@@ -33,13 +33,7 @@ utilisationReportsRoutes.post(
...
 );
 
-utilisationReportsRoutes.get(
-  '/:id/download',
-  validateTfmPaymentReconciliationFeatureFlagIsNotEnabled,
-  validateUserTeam(Object.values(PDC_TEAM_IDS)),
-  validateSqlId('id'),
-  getReportDownload,
-);
+utilisationReportsRoutes.get('/:id/download', validateUserTeam(Object.values(PDC_TEAM_IDS)), validateSqlId('id'), getReportDownload);
```
ttbarnes, line 36: The `validateTfmPaymentReconciliationFeatureFlagIsNotEnabled` check has been removed, is this intentional please?

e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/view-utilisation.spec.js
```diff
@@ -119,4 +120,11 @@ context('Users can view utilisation', () => {
       cy.assertText(cy.get('[data-cy="fees-payable"]'), `${CURRENCY.GBP} 100.00`);
     });
   });
+
+  it('should be able to download report for offline viewing', () => {
```
ttbarnes, line 124: Can we say e.g "should render a link"  instead of "should be able to download" please, because (quite rightly) it's not testing the actual download, rather that the link exists

ttbarnes (changes requested):
Few minor comments otherwise LGTM

trade-finance-manager-ui/server/routes/utilisation-reports/index.ts
RLavender98: Yes, we used to be able to download on a previous page, but when we made the changes to that page that are behind the feature flag we removed the link and so we put the route behind this check for security, however that was when we didn't know there would be a requirement to allow downloads after turning on the feature flag, just on a different page. 

So it is correct to allow it with it with the flag off, because the landing page has downloads in that case, and with the flag on, as this page will have downloads in that case.

ttbarnes: OK great thanks @RLavender98

ttbarnes (approved):

sw-mattw (approved):

ttbarnes (approved):

trade-finance-manager-ui/component-tests/utilisation-reports/utilisation-report-reconciliation-for-report.component-test.ts
```diff
@@ -341,6 +343,20 @@ describe(page, () => {
...
+      ...params,
+      utilisationDetails: {
+        utilisationTableRows: [aUtilisationTableRowViewModel()],
+        downloadUrl,
+      },
+    });
+
+    const utilisationTabSelector = 'div#utilisation';
+
+    wrapper.expectLink(`${utilisationTabSelector} a`).toLinkTo(downloadUrl, 'Download the report submitted by the bank as a CSV');
```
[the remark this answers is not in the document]
ttbarnes: PR LGTM apart from this 👍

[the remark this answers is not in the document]
sw-mattw: Updated this to use the selector thanks

ttbarnes (approved):

Zainzzkk (dismissed):

ttbarnes (approved):

Zainzzkk (approved):

PR #3668, opened 2024-10-11
chore(DTFS2-NONE): fix flakey test

BethThomas141:
## Introduction :pencil2:
As reported in the flaky test reporting slack channel, this test has failed on a few different PRs recently. It relies heavily on consistent values of `new Date()`. 

Freeze the value of `new Date()` at the start of the test and unfreeze it at the end. 

### DIFF
--- a/libs/common/src/helpers/date.test.ts
+++ b/libs/common/src/helpers/date.test.ts
@@ -3,6 +3,15 @@ import { getNowAsUtcISOString, getMonthName } from './date';
 
 describe('date helpers', () => {
   describe('getNowAsUtcISOString', () => {
+    beforeAll(() => {
+      jest.useFakeTimers();
+      jest.setSystemTime(new Date());
+    });
+
+    afterAll(() => {
+      jest.useRealTimers();
+    });
+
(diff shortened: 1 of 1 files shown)


ttbarnes (approved):

libs/common/src/helpers/date.test.ts
```diff
@@ -3,6 +3,15 @@ import { getNowAsUtcISOString, getMonthName } from './date';
 
 describe('date helpers', () => {
   describe('getNowAsUtcISOString', () => {
+    beforeAll(() => {
+      jest.useFakeTimers();
+      jest.setSystemTime(new Date());
```
AlexBramhill, line 8: im not versed in this, but I'd assume this is already set as the current date when you start using fake timers -- so we dont need this line

BethThomas141: I can totally believe that, was struggling to find hard evidence of that in the docs so was erring on the side of caution but will do another lil investigation

BethThomas141: I believe it, have removed that line

ttbarnes (approved):

PR #3671, opened 2024-10-11
feat(DTFS2-6892): commonise entra user schema and refactor tests

AlexBramhill:
## Introduction :pencil2:
This PR does a lot of heavy lifting refactoring and bringing some patterns in line regarding schemas.

This includes
- Moving the EntraIdUser schema to common as it will be needed in tfm api
- Update tests to use a mock generator (`aTestCase()` as apposed to `testCase`) to avoid accidental side effects on the test object. This is a welcome addition as schema tests are all about validating objects, and we should not allow for these to be polluted by side effects.
- Update schemas in TFM UI to `SCREAMING_SNAKE_CASE`. I really dislike screaming snake case, especially for schemas as they are not constants, but this is the existing pattern in libs/common (inherited from other services prior to zod). Note - I'm not removing the `schema` reference from the file name as this could easily be confused with constants, which is definitely not ideal!


libs/common/src/schemas/portal-user.update.test.ts
```diff
@@ -0,0 +1,57 @@
...
+function getFailureTestCases() {
+  return [
+    { description: 'a string', aTestCase: () => 'string' },
+    { description: 'an object', aTestCase: () => ({ An: 'object' }) },
+    { description: 'an array', aTestCase: () => ['array'] },
+    { description: 'a user with an additional parameter', aTestCase: () => ({ ...aValidPortalUser(), invalidField: true }) },
+  ];
+}
+
+function aValidPortalUser() {
```
ttbarnes, line 38: Do you think it's worth moving this into a shared mock/fixture, since it's repeated twice 

Maybe a bigger piece as I imagine a lot of portal user objects are duplicated without the `aValidPortalUser` wrapper

ttbarnes (approved):
Nice work

libs/common/src/schemas/portal-user.update.test.ts
AlexBramhill: It's an incredibly large piece of work that would be prone to error (I've had this thought a lot) as the portal user type is incredibly transient (there are several). The first stage of this work is actually typing portal users properly in our codebase so we can work out what's what (which is an ongoing process)!

ttbarnes (approved):

PR #3672, opened 2024-10-11
chore(DTFS2-6892): commonise london timezone

AlexBramhill:
## Introduction :pencil2:
As part of sso, I will need to use the london iana timezone. It is not currently commonised.

This ticket seeks to commonise the london timezone (outside of tests which are referring to it as a string), as well as creating a IanaTimezone type.

- The `IanaTimezone` type is broad -- being a `string`. This is because calling the `Intl.supportedValuesOf('timeZone')` function returns a broad type (`string`). I do not want to overly specify this type by making a static list of iana timezones in this PR.
- I've destructured imports as per the existing convention
- I've tidied up mismatching `require` and `import` on some js files


portal/server/nunjucks-configuration/filter-localiseTimestamp.ts
```diff
@@ -1,3 +1,4 @@
+import { IanaTimezone } from '@ukef/dtfs2-common';
```
AlexBramhill, line 1: commonising this file with the other filter file is not in scope for this ticket, as this is a large epic (sso)

portal/component-tests/admin/user-edit.component-test.js
```diff
@@ -1,14 +1,14 @@
...
+} = require('@ukef/dtfs2-common');
+const pageRenderer = require('../pageRenderer');
 
 const page = 'admin/user-edit.njk';
 const render = pageRenderer(page);
 
 describe(page, () => {
   const adminUser = {
-    timezone: LONDON_TIMEZONE,
+    timezone: LONDON,
```
ttbarnes, line 11: Some files now `TIMEZONES.LONDON`, others `LONDON`, can we be consistent please? I think either works

portal/component-tests/contract/components/contract-overview-table.component-test.js
```diff
@@ -78,11 +81,7 @@ describe(component, () => {
       wrapper
         .expectText('[data-cy="submissionDate"]')
         .toRead(
-          formatInTimeZone(
-            new Date(dealWithManualInclusionApplicationSubmissionDate.details.manualInclusionApplicationSubmissionDate),
-            LONDON_TIMEZONE,
-            FULL_DATE,
-          ),
+          formatInTimeZone(new Date(dealWithManualInclusionApplicationSubmissionDate.details.manualInclusionApplicationSubmissionDate), LONDON, FULL_DATE),
```
ttbarnes, line 84: Whilst we're in here, any chance we can destructure the date so this reads nicer? 

```js
formatInTimeZone(new Date(manualInclusionApplicationSubmissionDate), LONDON, FULL_DATE),
```

We could also rename `dealWithManualInclusionApplicationSubmissionDate` to `dealWithMiaSubmissionDate`

Appreciate this isn't the goal of the PR but whilst we're here, would be great

libs/common/src/constants/date-timezones.ts
```diff
@@ -0,0 +1,8 @@
+import { IanaTimezone } from '../types';
+
+/**
+ * IANA timezones used in the codebase
+ * Note - If needing a full list, we should look to Intl.supportedValuesOf('timeZone')
+ * {@link https://www.iana.org/time-zones}
+ */
+export const IANA_TIMEZONES = { LONDON: 'Europe/London' } as const satisfies Record<string, IanaTimezone>;
```
ttbarnes, line 8: Can we match the file name with the const please

portal/server/nunjucks-configuration/filter-localiseTimestamp.test.ts
```diff
@@ -1,6 +1,8 @@
-import { LONDON_TIMEZONE } from '../constants/date';
+import { IANA_TIMEZONES } from '@ukef/dtfs2-common';
```
ttbarnes, line 1: Some are importing `TIMEZONES`, others `IANA_TIMEZONES`, can they be consistent?

ttbarnes (changes requested):

libs/common/src/constants/date-timezones.ts
AlexBramhill: updated -- kept date prefix for ease of finding in file browser

portal/component-tests/admin/user-edit.component-test.js
AlexBramhill: updated

ttbarnes (approved):

libs/common/src/constants/date-iana-timezones.ts
```diff
@@ -0,0 +1,8 @@
+import { DateIanaTimezone } from '../types';
+
+/**
+ * IANA timezones used in the codebase
+ * Note - If needing a full list, we should look to Intl.supportedValuesOf('timeZone')
+ * {@link https://www.iana.org/time-zones}
+ */
+export const DATE_IANA_TIMEZONES = { LONDON: 'Europe/London' } as const satisfies Record<string, DateIanaTimezone>;
```
abhi-markan, line 8: Can this please be picked from the env variable `TZ`.

abhi-markan (changes requested):

libs/common/src/constants/date-iana-timezones.ts
AlexBramhill: I'm slightly cautious of this -- (prior to this pr) our code currently references everywhere as `LONDON_TIMEZONE` which is referencing the IANA timezone name, which is static so shouldn't be an env var. If we're after `DEFAULT_TIMEZONE` then I think we should look to refactor the variables, which is outside this scope (and will need to be considered whereever we transform timezones -- ie if we changes the `TZ` var to Europe/Paris, would we want to change _all_ cases of where we show time for this? Would we want to save things in the database like this etc?

Basically -- I know you'll have a great reason for this -- but for now as this is such a big ticket etc, and the changes could introduce subtle bugs if not considered, I might leave it if okay.

ttbarnes: If we want to use env vars instead of importing a constant, I suggest we do this in a different PR.

This PR is purely aimed at creating a DRY/common instance of `Europe/London`, some of which were hard coded.

Also btw @AlexBramhill maybe should rename the PR title to be `chore` instead of `feature`

AlexBramhill: updated title

abhi-markan: It is a minor change with values always set to `Europe/London` reason I suggest using `TZ` is to ensure all the Timezone values are referred from a single point during microservice initialisation locally, CICD or on Azure.

@ttbarnes understand this PR aims to make it DRY what I am suggesting is within those remits by not repeating a value which is already specified in env, existing PR or a new PR is both acceptable as long as it is updated since time is being invested in keeping the code DRY.

ttbarnes: Thanks @abhi-markan makes sense to me, I think we should do it as part of this PR whilst we're here. Unless it's blocking SSO in which case i'd suggest a new PR.

I understand Alex's concern about changing the TZ var to another timezone, but this seems like an extremely rare potential future requirement to me. **UK**EF

AlexBramhill: Could I ask why we need to use an env var for it please?

AlexBramhill: I'm not sure I agree this should be an env var. We do not have checking across services that these env vars are all identical, and its important to ensure they are all consistently set. Further, we have no way to verify that all our env vars are set to a valid IANA time zone. 

Regardless:
- I've updated this pattern to use a config, as it is no longer a constant.
- I've added validation that the env var is a string (so it'll fail if missing). 
- I've renamed the timezone to `DEFAULT` to reflect that this is the `DEFAULT` timezone so it won't be incorrectly used by people looking for `LONDON` explicitly (If we ever have to use the `TZ` env var, we will most likely need to reference LONDON timezone in the future, considering UKEF is a UK business)
- I've not added a default to this env var, as adding such a fallback would allow services to easily get out of sync accidentally
- I've had a brief skim of the code, and I don't think this will have an effect on existing code, but timezone changes can be prone to error.

AlexBramhill: I would like to pull the above change, and keep it static, as it adds additional complexity for the release of SSO which is out of scope (as seen in all tests failing due to env vars not being configured properly). Mind if I revert @abhi-markan

.github/workflows/development_infrastructure.yml
```diff
@@ -1126,7 +1126,8 @@ jobs:
                 GOV_NOTIFY_API_KEY="${{ secrets.GOV_NOTIFY_API_KEY }}" \
                 MONGO_INITDB_DATABASE="${{ env.database_name }}" \
                 MONGODB_URI="$mongo_connection_string" \
-                RATE_LIMIT_THRESHOLD=${{ vars.RATE_LIMIT_THRESHOLD }}
+                RATE_LIMIT_THRESHOLD=${{ vars.RATE_LIMIT_THRESHOLD }} \
+                TZ="${{ vars.TZ }}"
```
abhi-markan, line 1130:
```suggestion
                TZ=${{ vars.TZ }}
```

abhi-markan (changes requested):
One minor commit please.

abhi-markan (approved):

PR #3674, opened 2024-10-14
feat(DTFS2-7489): create submit cancellation endpoint tfm api

MarRobSoftwire:
## Introduction :pencil2:
The user needs to be able to submit the cancellation. This endpoint provides the stub from which the database update & email can be sent from.

- Add submit cancellation endpoint with middleware validation
- Add api and unit test coverage

## Note ⚠️ 
This endpoint doesn't actually submit the cancellation yet, but merging this early will allow DTFS2-7298 and DTFS2-7364 to be completed in parallel

### DIFF
--- a/trade-finance-manager-api/api-tests/v1/deal-cancellation/submit-deal-cancellation.post.api-test.ts
+++ b/trade-finance-manager-api/api-tests/v1/deal-cancellation/submit-deal-cancellation.post.api-test.ts
@@ -1,0 +1,172 @@
+import { AnyObject, MAX_CHARACTER_COUNT, TEAM_IDS } from '@ukef/dtfs2-common';
+import { ObjectId, UpdateResult } from 'mongodb';
+import { createApi } from '../../api';
+import app from '../../../src/createApp';
+import { initialiseTestUsers } from '../../api-test-users';
+import { TestUser } from '../../types/test-user';
+import { withTeamAuthorisationTests } from '../../common-tests/with-team-authorisation.api-tests';
+import { PostSubmitDealCancellationPayload } from '../../../src/v1/middleware/validate-post-submit-deal-cancellation-payload';
+
+const updateDealCancellationMock = jest.fn() as jest.Mock<Promise<UpdateResult>>;
+
+jest.mock('../../../src/v1/api', () => ({
+  ...jest.requireActual<AnyObject>('../../../src/v1/api'),
+  updateDealCancellation: () => updateDealCancellationMock(),
+}));
+
+const originalProcessEnv = { ...process.env };
+const { as, post } = createApi(app);
+
+const validId = new ObjectId().toString();
+
+const mockUpdateResult: UpdateResult = {
+  acknowledged: true,
+  modifiedCount: 1,
+  matchedCount: 1,
+  upsertedCount: 1,
+  upsertedId: new ObjectId(validId),
+};
+
+const getSubmitTfmDealCancellationUrl = ({ id }: { id: string }) => `/v1/deals/${id}/cancellation/submit`;
+
+const aValidPayload = (): PostSubmitDealCancellationPayload => ({
+  reason: 'x'.repeat(MAX_CHARACTER_COUNT),
+  bankRequestDate: new Date().valueOf(),
+  effectiveFrom: new Date().valueOf(),
+});
+
+describe('POST /v1/deals/:id/cancellation/submit', () => {
+  let testUsers: Awaited<ReturnType<typeof initialiseTestUsers>>;
+  let aPimUser: TestUser;
+
+  beforeAll(async () => {
+    testUsers = await initialiseTestUsers(app);
+    aPimUser = testUsers().withTeam(TEAM_IDS.PIM).one();
+  });
+
+  beforeEach(() => {
+    jest.resetAllMocks();
+    jest.mocked(updateDealCancellationMock).mockResolvedValue(mockUpdateResult);
+  });
+
+  afterAll(() => {
+    jest.restoreAllMocks();
+    process.env = originalProcessEnv;
+  });
+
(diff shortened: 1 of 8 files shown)


trade-finance-manager-api/api-tests/v1/deal-cancellation/submit-deal-cancellation.post.api-test.ts
```diff
@@ -0,0 +1,154 @@
...
+        description: 'the reason is undefined',
+        payload: { ...aValidPayload(), reason: undefined },
+      },
+      {
+        description: 'the bankRequestDate is a string ',
+        payload: { ...aValidPayload(), bankRequestDate: new Date().toISOString() },
+      },
+      {
+        description: 'the bankRequestDate is undefined ',
+        payload: { ...aValidPayload(), bankRequestDate: undefined },
```
ttbarnes, line 120: I think in such POST validation tests, instead of undefined, it should be an empty string `''`, since this is what `req.body` would contain if nothing is provided - same on line 112 and 128,

ttbarnes, line 120: I think in such POST validation tests, instead of undefined, it should be an empty string `''`, since this is what `req.body` would contain if nothing is provided - same on line 112 and 128,

ttbarnes (changes requested):
Few minor comments, otherwise LGTM

trade-finance-manager-api/api-tests/v1/deal-cancellation/submit-deal-cancellation.post.api-test.ts
ttbarnes: Same in the middleware

trade-finance-manager-api/api-tests/v1/deal-cancellation/submit-deal-cancellation.post.api-test.ts
```diff
@@ -0,0 +1,154 @@
...
+      const url = getSubmitTfmDealCancellationUrl({ id: 'invalid' });
+
+      // Act
+      const response = await as(aPimUser).post(aValidPayload()).to(url);
+
+      // Assert
+      expect(response.status).toEqual(400);
+    });
+
+    const invalidPayloads = [
```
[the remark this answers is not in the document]
MarRobSoftwire: I think its nice to check the middleware is working as expected in an api test as well as the unit test - but you're right I can be less exhaustive. I'll just include a single `when the payload is invalid` test

trade-finance-manager-api/api-tests/v1/deal-cancellation/submit-deal-cancellation.post.api-test.ts
MarRobSoftwire: I dont think I quite understand what you mean
> since this is what req.body would contain if nothing is provided
Its possible for the payload to not include a property, so it would be undefined

trade-finance-manager-api/src/v1/controllers/deal-cancellation/submit-deal-cancellation.controller.test.ts
```diff
@@ -0,0 +1,38 @@
...
+const dealCancellation = {
+  reason: 'test reason',
+  bankRequestDate: 1794418807,
+  effectiveFrom: 1794418808,
+};
+
+const mockDealId = new ObjectId();
+const mockUserId = new ObjectId();
+
+describe('controllers - deal cancellation', () => {
```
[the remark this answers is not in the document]
MarRobSoftwire: I've got the differentiating describe within this `'POST - submitDealCancellation'` - I think its an existing pattern for the top level describe to refer to the group of controllers (in this case those serving endpoints `/deal/:id/cancellation`).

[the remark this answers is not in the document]
MarRobSoftwire: I can refactor this (but should probably do the same to the put/get/delete tests too for consistency) if you like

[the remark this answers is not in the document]
ttbarnes: If it's an existing pattern we should stick to it

trade-finance-manager-api/api-tests/v1/deal-cancellation/submit-deal-cancellation.post.api-test.ts
ttbarnes: A typical use case would be submitting an empty form, or the form with valid values apart from `bankRequestDate`. In this case it'd come through as `bankRequestDate: ''`

The only way AFAIK It would be `undefined` is if a user is manipulating the DOM or network request

I'd prefer this to be an empty string rather than specifically `undefined`. What do you think?

ttbarnes: Sorry @MarRobSoftwire I was thinking this was UI POST test 🤦 

Happy with this but should we add a test for if `bankRequestDate` is an empty string?

MarRobSoftwire: I've got a test
```
    {
      description: "the 'bankRequestDate' is a string",
      payload: {
        ...aValidPayload(),
        bankRequestDate: new Date().toString(),
      },
    },
```
The zod validation is looking for bankRequestDate to be a number so i dont think we gain anything new by checking for an empty string?

ttbarnes: OK makes sense, thanks!

ttbarnes (approved):

PR #3681, opened 2024-10-15
fix(DTFS2-7492): await tfm deal update

MarRobSoftwire:
## Introduction :pencil2:
When submitting a deal from portal to TFM, the deal is not shown in ACBS (on staging). The logs revealed an error that can be traced to not awaiting a deal update

Await the deal update in tfm-api

### DIFF
--- a/trade-finance-manager-api/src/v1/controllers/deal.submit.controller.js
+++ b/trade-finance-manager-api/src/v1/controllers/deal.submit.controller.js
@@ -107,7 +107,7 @@ const submitDealAfterUkefIds = async (dealId, dealType, checker, auditDetails) =
     }
 
     // Update the deal with all the above modifications
-    const tfmDeal = api.updateDeal({ dealId, dealUpdate, auditDetails });
+    const tfmDeal = await api.updateDeal({ dealId, dealUpdate, auditDetails });
 
     // Submit to ACBS
     const canSubmitDealToACBS = await canSubmitToACBS(tfmDeal);


AlexBramhill (approved):

trade-finance-manager-api/src/v1/controllers/deal.submit.controller.js
```diff
@@ -107,7 +107,7 @@ const submitDealAfterUkefIds = async (dealId, dealType, checker, auditDetails) =
     }
 
     // Update the deal with all the above modifications
-    const tfmDeal = api.updateDeal({ dealId, dealUpdate, auditDetails });
+    const tfmDeal = await api.updateDeal({ dealId, dealUpdate, auditDetails });
```
abhi-markan, line 110: This is left out on purpose to avoid sending the update deal stage.

abhi-markan (changes requested):

trade-finance-manager-api/src/v1/controllers/deal.submit.controller.js
ttbarnes: I'm also confused @abhi-markan 

All the api calls in this file are an `await`, quite surprised this was missing

If needs to _not_ await (why? - bizarre) can we add a comment about this please.

abhi-markan: Apologies both, I misunderstood this with deal status controller. Good check @MarRobSoftwire `await` should have been there inline with LOC:183.

I tested this and it appears due to an unexpected response from `addPartyUrn` which causes a snowball effect. I would have expected this to fail in last 6 months but above must have developed a race condition.

abhi-markan (dismissed):

trade-finance-manager-api/src/v1/controllers/deal.submit.controller.js
abhi-markan: @MarRobSoftwire when you get a minute can you please set the base to `2.4.0` so we can regress this on `staging` next so we can get this out on `prod`.

MarRobSoftwire:
Closing PR to point this at 2.4

PR #3683, opened 2024-10-15
feat(DTFS2-7528): submit cancellation in tfm-ui

MarRobSoftwire:
## Introduction :pencil2:
The check your details page currently does nothing when the button is pressed. It should send a submit request to tfm-api

- Send `POST` request to `cancellation/submit` in tfm-api, sending cancellation as payload.
- Refactor check-details page to take `cancellation` object, which is returned on the post request

- Take format dates in nunjucks template, not controller using new `formatUnixTimestamp` filter
- fix `delete account` typo in template

### DIFF
--- a/trade-finance-manager-ui/component-tests/case/cancellation/check-details.component-test.ts
+++ b/trade-finance-manager-ui/component-tests/case/cancellation/check-details.component-test.ts
@@ -32,7 +32,7 @@ describe(page, () => {
     wrapper.expectLink('[data-cy="return-link"]').toLinkTo(`/case/${dealId}/cancellation/cancel`, 'Return to deal summary');
   });
 
-  it('should render the delete account button', () => {
+  it('should render the cancel deal button', () => {
     // Arrange
     const checkDetailsViewModel: CheckDetailsViewModel = aCheckDetailsViewModel();
 
@@ -41,7 +41,7 @@ describe(page, () => {
 
     // Assert
     wrapper.expectElement('[data-cy="delete-button"]').toExist();
-    wrapper.expectText('[data-cy="delete-button"]').toRead('Delete account');
+    wrapper.expectText('[data-cy="delete-button"]').toRead('Cancel deal');
   });
 
   it('should display a warning message about cancelling the deal', () => {
@@ -71,7 +71,7 @@ describe(page, () => {
 
   it('should display a dash when the reason for cancelling a deal is empty', () => {
     // Arrange
-    const checkDetailsViewModel: CheckDetailsViewModel = aCheckDetailsViewModel();
+    const checkDetailsViewModel: CheckDetailsViewModel = { ...aCheckDetailsViewModel(), cancellation: {} };
 
     // Act
     const wrapper = render(checkDetailsViewModel);
@@ -84,7 +84,7 @@ describe(page, () => {
   it('should display the reason for cancelling text when it exists', () => {
     // Arrange
     const reason = 'test reason';
-    const checkDetailsViewModel: CheckDetailsViewModel = { ...aCheckDetailsViewModel(), reason };
+    const checkDetailsViewModel: CheckDetailsViewModel = { ...aCheckDetailsViewModel(), cancellation: { reason } };
 
     // Act
     const wrapper = render(checkDetailsViewModel);
@@ -96,28 +96,30 @@ describe(page, () => {
 
   it('should display the received bank request date date string', () => {
     // Arrange
-    const receivedDateString = '1 January 2024';
-    const checkDetailsViewModel: CheckDetailsViewModel = { ...aCheckDetailsViewModel(), bankRequestDate: receivedDateString };
+    const bankRequestDate = 1729007392441;
+    const checkDetailsViewModel: CheckDetailsViewModel = { ...aCheckDetailsViewModel(), cancellation: { bankRequestDate } };
 
     // Act
     const wrapper = render(checkDetailsViewModel);
 
     // Assert
+    const expectedBankRequestDateString = '15 October 2024';
     wrapper.expectElement('[data-cy="bank-request-date-response"]').toExist();
-    wrapper.expectText('[data-cy="bank-request-date-response"]').toRead(receivedDateString);
+    wrapper.expectText('[data-cy="bank-request-date-response"]').toRead(expectedBankRequestDateString);
   });
 
   it('should display the received effective from date string', () => {
     // Arrange
-    const receivedDateString = '1 January 2024';
-    const checkDetailsViewModel: CheckDetailsViewModel = { ...aCheckDetailsViewModel(), effectiveFromDate: receivedDateString };
+    const effectiveFrom = 1729007392441;
+    const checkDetailsViewModel: CheckDetailsViewModel = { ...aCheckDetailsViewModel(), cancellation: { effectiveFrom } };
 
     // Act
     const wrapper = render(checkDetailsViewModel);
 
     // Assert
+    const expectedEffectiveFromString = '15 October 2024';
     wrapper.expectElement('[data-cy="effective-from-response"]').toExist();
-    wrapper.expectText('[data-cy="effective-from-response"]').toRead(receivedDateString);
+    wrapper.expectText('[data-cy="effective-from-response"]').toRead(expectedEffectiveFromString);
   });
 
   it('should render the Change links', () => {
--- a/trade-finance-manager-ui/server/api.js
+++ b/trade-finance-manager-ui/server/api.js
@@ -1300,6 +1300,27 @@ const deleteDealCancellation = async (dealId, userToken) => {
   }
 };
 
+/**
+ * Submits the deal cancellation object on a TFM MIN or AIN deal
+ * @param {string} dealId - The deal ID
+ * @param {import('@ukef/dtfs2-common').TfmDealCancellation} cancellation - The deal cancellation object
+ * @param {string} userToken - The user token
+ * @returns {Promise<void>}
+ */
+const submitDealCancellation = async (dealId, cancellation, userToken) => {
+  try {
+    await axios({
+      method: 'post',
+      url: `${TFM_API_URL}/v1/deals/${dealId}/cancellation/submit`,
+      headers: generateHeaders(userToken),
+      data: cancellation,
+    });
+  } catch (error) {
+    console.error('Failed to submit deal cancellation', error);
+    throw error;
+  }
+};
+
 module.exports = {
   getDeal,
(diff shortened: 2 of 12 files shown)


trade-finance-manager-ui/server/controllers/case/cancellation/check-details.controller.ts
```diff
@@ -76,11 +71,15 @@ export const postDealCancellationDetails = async (req: PostDealCancellationDetai
       return res.redirect('/not-found');
     }
 
-    if (!canSubmissionTypeBeCancelled(deal.dealSnapshot.submissionType)) {
+    const submissionTypeCanBeCancelled = canSubmissionTypeBeCancelled(deal.dealSnapshot.submissionType);
+    const cancellationIncomplete = reason === undefined || !bankRequestDate || !effectiveFrom;
```
ttbarnes, line 75:
```suggestion
    const cancellationIncomplete = !reason || !bankRequestDate || !effectiveFrom;
```

trade-finance-manager-ui/server/nunjucks-configuration/filter-dashIfEmpty.js
```diff
@@ -5,7 +5,8 @@ const dashIfEmpty = (text) => {
 
   const textStr = String(text);
 
-  const isEmpty = !(textStr && textStr.trim().length > 0 && textStr !== 'NaN' && textStr !== 'Invalid date');
+  const isEmpty = !textStr || textStr.trim().length === 0 || textStr === 'NaN' || textStr === 'Invalid date';
```
ttbarnes, line 8: Can we add a unit test please, doesn't seem to be anything that checks the length specifically

trade-finance-manager-ui/component-tests/case/cancellation/check-details.component-test.ts
```diff
@@ -41,7 +41,7 @@ describe(page, () => {
 
     // Assert
     wrapper.expectElement('[data-cy="delete-button"]').toExist();
-    wrapper.expectText('[data-cy="delete-button"]').toRead('Delete account');
+    wrapper.expectText('[data-cy="delete-button"]').toRead('Cancel deal');
```
BethThomas141, line 44: Good catch... my bad on that one😅

trade-finance-manager-ui/server/nunjucks-configuration/filter-formatUnixTimestamp.ts
```diff
@@ -0,0 +1,15 @@
...
+ * @param toFormat date format using {@link https://date-fns.org/v3.3.1/docs/format | unicode tokens}
+ * @returns date formatted as toFormat or 'Invalid date' if can't parse isoDateStr
+ */
+export const formatUnixTimestamp = (timestamp: number, toFormat: string) => {
+  if (!timestamp) {
+    return 'Invalid date';
+  }
+
+  const date = toDate(timestamp);
+  return isValid(date) ? format(date, toFormat) : 'Invalid date';
```
ttbarnes, line 14: Should we use `isValidDate` from `libs/common` instead of date-fns?

trade-finance-manager-ui/templates/case/cancellation/check-details.njk
```diff
@@ -14,51 +14,51 @@
...
     Change <span class="govuk-visually-hidden">date effective from</span>
-  </a>
+</a>
 {% endset %}
-  {% set effectiveFromDate =
+{% set effectiveFromDate =
     [
       { text: 'Date effective from', attributes: { "data-cy": "effective-from-heading" }, classes: rowTextClasses },
-      { text: effectiveFromDate | dashIfEmpty, attributes: { "data-cy": "effective-from-response" }, classes: rowTextClasses },
+      { text: cancellation.effectiveFrom | formatUnixTimestamp('d MMMM yyyy') | dashIfEmpty, attributes: { "data-cy": "effective-from-response" }, classes: rowTextClasses },
```
BethThomas141, line 61: Good change, I prefer the formatting handled here - thanks!

ttbarnes (changes requested):

trade-finance-manager-ui/server/nunjucks-configuration/filter-dashIfEmpty.js
BethThomas141, line 8: That is so much easier to read😌

trade-finance-manager-ui/server/nunjucks-configuration/filter-formatUnixTimestamp.ts
BethThomas141, line 14: Aware it's the existing pattern, is there a better way to handle invalid data in nunjucks files as opposed to returning an arbitrary 'invalid date' string?

trade-finance-manager-ui/server/controllers/case/cancellation/check-details.controller.get.test.ts
```diff
@@ -129,14 +126,12 @@ describe('getDealCancellationDetails', () => {
 
       // Assert
       expect(res._getRenderView()).toEqual('case/cancellation/check-details.njk');
-      expect(res._getRenderData() as BankRequestDateViewModel).toEqual({
+      expect(res._getRenderData() as CheckDetailsViewModel).toEqual({
```
BethThomas141, line 129: Thanks for spotting all these that slipped through a previous PR😅

trade-finance-manager-ui/server/api.js
```diff
@@ -1300,6 +1300,27 @@ const deleteDealCancellation = async (dealId, userToken) => {
   }
 };
 
+/**
+ * Submits the deal cancellation object on a TFM MIN or AIN deal
+ * @param {string} dealId - The deal ID
+ * @param {import('@ukef/dtfs2-common').TfmDealCancellation} cancellation - The deal cancellation object
```
BethThomas141, line 1306: Out of interest, if no reason was submitted, would this in fact be a `Partial` of this type instead? (looks like the zod schema it's referencing doesn't have it as optional - but not sure if an empty string would still satisfy the `string` requirement)

trade-finance-manager-ui/server/controllers/case/cancellation/check-details.controller.post.test.ts
```diff
@@ -48,41 +55,128 @@ describe('postBankRequestDate', () => {
...
+        cancellation: { reason, effectiveFrom },
+      },
+      {
+        description: 'when the effectiveFrom is undefined',
+        cancellation: { reason, bankRequestDate },
+      },
+    ];
+
+    describe.each(invalidTestCases)('$description', ({ cancellation }) => {
+      it('does not submit the deal cancellation', async () => {
```
BethThomas141, line 115: 🤩 Where is this logic handled in the ui layer? 🕵️

trade-finance-manager-ui/server/nunjucks-configuration/filter-formatUnixTimestamp.ts
MarRobSoftwire: `isValid` is used in hundred of places - I dont know why `isValidDate` was added?

trade-finance-manager-ui/server/api.js
MarRobSoftwire: I tested this a while back, if you submit an empty text input it sends an empty string so the reason should always be defined

trade-finance-manager-ui/server/controllers/case/cancellation/check-details.controller.post.test.ts
MarRobSoftwire: `    const cancellationIncomplete = reason === undefined || !bankRequestDate || !effectiveFrom;` deals with this - Its obviously not a perfect check but it keeps Typescript happy & it shouldnt ever be needed

trade-finance-manager-ui/server/controllers/case/cancellation/check-details.controller.ts
MarRobSoftwire: `reason` can be an empty string

trade-finance-manager-ui/templates/case/cancellation/check-details.njk
MarRobSoftwire: Thought it made sense when I need the hidden cancellation input to only pass this data in once :)

trade-finance-manager-ui/server/nunjucks-configuration/filter-formatUnixTimestamp.ts
BethThomas141: Hmm yes, I think it would be a case of either making both: 
- This return undefined
- And `dashIfEmpty` to stop handling the strange arbitrary string cases.
at the same time. 
But I don't think that's feasible in this PR / would be inconsistent across other services. 

So maybe one for another time and leave as is👍

MarRobSoftwire: Should i remove `isValidDate` and replace it with `isValid`?

ttbarnes: Huh, that's strange. I think @Zainzzkk recently added this

I suggest we stick with `isValid` in this instance

trade-finance-manager-ui/server/controllers/case/cancellation/check-details.controller.ts
ttbarnes: `!reason` should work for undefined and empty string?

MarRobSoftwire: Exactly, I want to rule out `undefined` but allow an empty string. Using your suggestion if the user doesnt provide a reason, it will be set to `''` and then they wont be able to submit the cancellation

ttbarnes: Ah, sorry is "reason" an optional field?

MarRobSoftwire: Yep its optional, so will always have a value of at least an empty string

ttbarnes (approved):

trade-finance-manager-ui/server/controllers/case/cancellation/check-details.controller.ts
ttbarnes: Thanks @MarRobSoftwire , didn't realise, makes complete sense 😄

PR #3684, opened 2024-10-15
feat(DTFS2-7531): bump gov design library

ttbarnes:
## Introduction :pencil2:
This PR updates the `govuk-frontend` package from **4.8.0** to the latest version, **5.7.1** in all UI microservices:
  - GEF UI
  - Portal UI
  - TFM UI

This is the first PR as part of a series of PRs / changes / improvements we can make with the latest GOV design version.

- Bump `govuk-frontend` package.
- Update all `js-enabled` scripts to include a conditional `govuk-frontend-supported` class.
- Update relevant client side scripts integrity hashes.
- Update all client side scripts to have `type="module"`.
- Update express assets, nunjucks config's and SCSS GOV paths.
- Bump MOJ package.
- Update `service-unavailable` CSS.
- Fix some minor bugs seemingly caused by the GOV design updates (DTFS2-7641, DTFS2-7639, DTFS2-7640, DTFS2-7642).

- Update nunjucks component tests assertions to not check the `govuk-button--disabled` class, since it has been removed. Instead, check the `aria-disabled` attribute.
- Improve the Portal's main `index.njk` readability, indentation. 
- Update TFM UI's success banner (from [this original PR](https://github.com/UK-Export-Finance/dtfs2/pull/3904))
- Update TFM "facility pricing risk" table component to be hard coded instead of using the GOV component.
  - Unfortunately, the existing design cannot be achieved by using the GOV component, we have to hard code it. More details in the DTFS2-7531 jira ticket. Documentation has been added to the nunjucks component.
- Various conditional/validation rule tweaks.
- Add missing unit test for a currency validation test.
- Update/align all "reconciliation" status tags to _not_ be uppercase.

### DIFF
--- a/e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/report-reconciliation/delete-payment.spec.js
+++ b/e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/report-reconciliation/delete-payment.spec.js
@@ -215,7 +215,7 @@ context(`${PDC_TEAMS.PDC_RECONCILE} users can delete payments`, () => {
     pages.utilisationReportConfirmDeletePaymentPage.clickContinueButton();
 
     pages.utilisationReportPage.premiumPaymentsTab.getPaymentLink(payment.id).should('not.exist');
-    cy.get('strong[data-cy="fee-record-status"]:contains("TO DO")').should('exist');
+    cy.get('strong[data-cy="fee-record-status"]:contains("To do")').should('exist');
   });
 
   function* idGenerator() {


gef-ui/package.json
```diff
@@ -34,7 +34,7 @@
     "unit-test-ff": "jest --coverage --verbose --config=unit.ff.jest.config.js --passWithNoTests"
   },
   "dependencies": {
-    "@ministryofjustice/frontend": "1.8.0",
+    "@ministryofjustice/frontend": "3.0.1",
```
AlexBramhill, line 37: @ttbarnes can we bump to 3.0.2 as it has my fix for the issue @BethThomas141 was having with the banner class cc @MarRobSoftwire

ttbarnes: Will do - what was the banner issue?

AlexBramhill: Thewre was an issue with MoJ components not displaying class if you set `type` (see [here](https://github.com/ministryofjustice/moj-frontend/issues/886)). Its since been fixed. If I'm still blocked tomorrow I'll dig out Beth's work around for this and update it.

ttbarnes: Nice one @AlexBramhill I think we can just bump the MOJ library in this PR instead of the workaround

AlexBramhill: Sounds great yeah!

ttbarnes: I've bumped MOJ to `3.0.2`. Need to run this branch locally and check some things before it's ready

e2e-tests/portal/cypress/e2e/journeys/maker-about-supply-contract/fill-in-about-supply-contract-validation.spec.js
```diff
@@ -36,31 +36,50 @@ context('about-supply-contract', () => {
 
     contractAboutSupplier.visit(deal);
     contractAboutSupplier.supplierName().clear();
+
```
ttbarnes, line 39: The improvements in this test are due to 2x bugs found after updating the GOV design library,  DTFS2-7639, DTFS2-7640.

portal-api/src/v1/section-currency/index.js
```diff
@@ -1,15 +1,24 @@
...
-    currencyId,
-  };
+    return {
+      text,
+      id,
+      currencyId,
+    };
+  }
+
+  return {};
```
ttbarnes, line 21: After GOV design update - not sure why exactly, but this was erroring and causing a crash if no currency is found via `findOneCurrency` (DTFS2-7639, DTFS2-7640)

Therefore i've updated this to return an empty object.

portal/server/routes/contract/about/industryFields/index.js
```diff
@@ -7,21 +7,31 @@ const industryFields = (postedSubmissionDetails, industrySectors) => {
...
+      };
+    } else {
+      /**
+       * This is required because - if an industry sector is not found
+       * and we save 'industry-sector' as undefined,
+       * the validation thinks that it has a valid value.
+       * Therefore, if no industry sector is found, the object should be wiped.
+       */
+      submissionDetails['industry-sector'] = {};
+    }
```
ttbarnes, line 28: After GOV design update - not sure why exactly, but this was erroring and causing validation issues if no indusry is selected - the validation would think that "undefined" is a valid value. (DTFS2-7640)

Therefore i've updated this so that if no industry sector is found by the provided ID, an empty object with be returned.

portal-api/src/utils/string.js
```diff
@@ -7,8 +7,16 @@ const isEmptyString = (str) => {
...
+ * Check if a param has a value, i.e
+ * - the param is defined
+ * - the param is not an empty string
+ * - the param is not a 'Select value' string
+ * @param {unknown}
+ * @returns Boolean
+ */
 const hasValue = (str) => {
-  if (str && !isEmptyString(str)) {
+  if (str && !isEmptyString(str) && str !== 'Select value') {
```
ttbarnes, line 19: This is required, otherwise various validation thinks that "Select value" is a valid string.

portal-api/src/v1/validation/fields/currency-not-the-same-as-supply-contract-rules/currency.js
```diff
@@ -1,11 +1,12 @@
...
 const { orderNumber } = require('../../../../utils/error-list-order-number');
 const { currencyIsDisabled } = require('../currency');
 
+// TODO - unit test
+
 module.exports = (facility, errorList) => {
   const newErrorList = { ...errorList };
 
-  if (!hasValue(facility.currency)) {
+  if (!facility.currency?.id) {
```
ttbarnes, line 9: This validation was causing issues because `hasValue` was returning `true` for an empty object, which means no value has been selected/mapped.

gef-ui/server/controllers/select-exporters-correspondence-address/index.js
```diff
@@ -27,7 +27,7 @@ const validateSelectExportersCorrespondenceAddress = (req, res) => {
   const parseAddresses = JSON.parse(addresses);
   let selectedAddressError;
 
-  if (isEmpty(selectedAddress)) {
+  if (isEmpty(selectedAddress) || selectedAddress.includes('Addresses Found')) {
```
ttbarnes, line 30: This is required, otherwise the validation thinks that the first select option (with "X Addresses Found" is a valid option.

e2e-tests/portal/cypress/e2e/journeys/maker-about-supply-contract/fill-in-about-supply-contract-validation.spec.js
AlexBramhill: Impressive!

e2e-tests/portal/cypress/e2e/journeys/maker-bond/bond-details.spec.js
```diff
@@ -46,8 +46,8 @@ context('Bond Details', () => {
 
       partials.errorSummaryLinks().should('have.length', TOTAL_REQUIRED_FORM_FIELDS);
 
-      pages.bondDetails.bondTypeInputErrorMessage().should('be.visible');
-      pages.bondDetails.facilityStageInputErrorMessage().should('be.visible');
+      cy.assertText(pages.bondDetails.bondTypeInputErrorMessage(), 'Error: Enter the Bond type');
```
AlexBramhill, line 49: Really nice job upgrading this

gef-ui/public/js/disableFormSubmitOnSubmission.js
```diff
@@ -1,42 +1,2 @@
...
-        t.after(e),
-        e
-      );
-    })(i);
-    (e = i), (n = s);
-  });
-})(),
-  ((DTFS_GEF = void 0 === DTFS_GEF ? {} : DTFS_GEF).disableFormSubmitOnSubmission = {});
-//# sourceMappingURL=disableFormSubmitOnSubmission.js.map
+var DTFS_GEF;(()=>{const t=(()=>{const t=document.querySelectorAll("form");return t.length&&t[t.length-1]||null})();if(!t)return;let e=null,n=null;t.addEventListener("submit",(t=>{const{submitter:i}=t;if(!(i instanceof HTMLElement))return;if(e){if(i===e)return void t.preventDefault();(r=n)instanceof HTMLElement&&r.remove(),(t=>{t instanceof HTMLElement&&(t.removeAttribute("disabled"),t.removeAttribute("aria-disabled"))})(e)}var r;(t=>{t instanceof HTMLElement&&(t.setAttribute("disabled",""),t.setAttribute("aria-disabled","true"))})(i);const s=(t=>{const e=document.createElement("input");return e.setAttribute("id","resubmit-prevention-hidden-input"),e.setAttribute("type","hidden"),["name","value"].forEach((n=>{const i=t.getAttribute(n);i&&e.setAttribute(n,i)})),e.classList.add("js-hidden-input"),t.after(e),e})(i);e=i,n=s}))})(),(DTFS_GEF=void 0===DTFS_GEF?{}:DTFS_GEF).disableFormSubmitOnSubmission={};
```
AlexBramhill, line 1: You'll need to regenerate all integrity SHAs for all modified (including line breaks iirc) files

gef-ui/public/js/govukFrontend.js.LICENSE.txt
```diff
@@ -0,0 +1,97 @@
+/**
```
AlexBramhill, line 1: Is this some new autogenerated file?

gef-ui/server/controllers/select-exporters-correspondence-address/index.js
AlexBramhill: Thats a good shout as a fix -- 

Trying not to get nerd sniped by this (And I don't think we should for the sake of this ticket) -- but I think there's probably a neater way of doing this using `disabled` if we can override the first element (if this is a drop down) ie `<option selected disabled>10 Addresses Found</option>` would disable this option from ever getting sent. Could probably pass in `:first-child disabled` or some sort of custom style like that.

Side note - `:user-valid` is a new CSS class which is cool for text input fields that allows for a similar thing on text fields

gef-ui/templates/index.njk
```diff
@@ -45,11 +45,11 @@
...
-    {% block scripts %}{% endblock %}
-  </body>
-</html>
\ No newline at end of file
+    <script src="/assets/js/main.js" type="module" integrity="sha512-HEhxQSphBlHV9OM8V0qLsCylJB4c9LbFdOC7pB9Ujg7ySL8E29ifnmtF+6i5I5ieMJk9+q+WlvhPyvyn0GxdhQ==" crossorigin="anonymous"></script>
+    <script src="/assets/js/govukFrontend.js" type="module" integrity="sha512-lgazaHvhhkqqZddDJelEkMTKltMOXka6crS1q7BjVdPYHxIEZzylH47PPHYklGNX6iZFhbkp8JHwXVJp1zfQIQ==" crossorigin="anonymous"></script>
+    <script src="/assets/js/mojFrontend.js" type="module" integrity="sha512-eOl0TifTsZ/1GhH6FNfs6SoJhL0shOkKCDjvLnC9+Ewwp6n6uYwIXuiQqphBMSVenttWkRWfM2Fqs6zCvneIzw==" crossorigin="anonymous"></script>
+    <script src="/assets/js/maskedInputs.js" type="module" integrity="sha512-4XTH3IT3oQN/60lRAdBZWXMXq4/L1n5kUoGEftu+nY0Bba/7T0RDIZ62es6v+dCyNJ+uaYZTjWl3nPYv0rR6WA==" crossorigin="anonymous"></script>
+    <script src="/assets/js/disableFormSubmitOnSubmission.js" type="module" integrity="sha512-RwF5ml+CiOHtgzu/uJmiCoytVi+TMCZGAKWvD+meYlY8bxi+kxC2qMDDiAyi17CoSK5rzB9DO4lej7QRXz8sQQ==" crossorigin="anonymous"></script>
```
AlexBramhill, line 52: I'd check these integrity hashes don't need regenerating. I feel they will...

AlexBramhill (commented):
Added review for GEF-UI -- Looking good. Working on Portal UI now

e2e-tests/portal/cypress/e2e/journeys/maker-bond/bond-details.spec.js
ttbarnes: Thanks - could have done more in this file and others, but didn't want to bloat this PR even more

portal/server/routes/contract/about/industryFields/index.js
AlexBramhill: Interesting -- Where is this validation failing? It's odd we allow undefined... Is the answer to do this or is the answer to update our validation in this area (with another ticket)?

gef-ui/public/js/disableFormSubmitOnSubmission.js
ttbarnes: All appropriate scripts/integrities have been updated - just double checked `disableFormSubmitOnSubmission` and it is running without integrity mismatch

gef-ui/public/js/govukFrontend.js.LICENSE.txt
ttbarnes: Yes, auto generated by GOV/webpack copy

portal/server/routes/contract/about/industryFields/index.js
AlexBramhill: I'd be tempted to consider refactoring this -- but again maybe beyond the scope of this so we can leave this complexity for now and pick it up another time

AlexBramhill: For what its worth -- We're not deep cloning the postedSubmissionDetails -- So we are changing both, so assignment is pointless
```
const industryFields = (submissionDetails, industrySectors) => {
    const industrySectorCode = submissionDetails['industry-sector'];  
    const industrySectorObj = industrySectorCode && getIndustrySectorById(industrySectors, industrySectorCode);  

    if(!industrySectorObj) {  
        submissionDetails['industry-sector'] = {  
            code: industrySectorCode,  
            name: industrySectorObj.name,  
        };       
    } else {
        submissionDetails['industry-sector'] = {};  
    }

    return submissionDetails;  
}
```

ttbarnes: I agree, lots of things could be improved.

I cannot remember the exact details for this off the top of my head - can dive back into previous commits if you'd like, but without this change, if I remember correctly:

- The validation thought "Select value" was a valid value.
- With the above in play, `submissionDetails['industry-sector']` was being set to `undefined`, because no `industrySectorCode` is found from "Select value".

More details in this ticket: https://ukef-dtfs.atlassian.net/browse/DTFS2-7640

In other instances, `submissionDetails['industry-sector']` is set to an empty object. I think it's safest to do this instead of allowing `undefined`

ttbarnes: Will have a look into those suggestions @AlexBramhill , thanks.

I was trying more for a "quick fix" with the existing function but more than happy to improve it.

I don't think the first suggestion will work as the object is returned before `industry-class` logic happens

portal-api/src/v1/validation/fields/currency-not-the-same-as-supply-contract-rules/currency.js
AlexBramhill: I wonder if we need an isObjectEmpty function as a temp fix?

The return of empty objects is widespread and can cause a host of errors (noticeable and silent)... but thats another day's discussion. I'm not even sure where we would begin, but if we moved to an easily maintainable existing library, like zod or joi, I reckon around 90% of validation code would be gutted.

portal-api/src/v1/validation/fields/currency-not-the-same-as-supply-contract-rules/currency.test.js
```diff
@@ -0,0 +1,66 @@
...
+        currency: {
+          order: orderNumber(errorList),
+          text: 'Enter the Currency',
+        },
+      };
+
+      expect(result).toEqual(expected);
+    });
+  });
+
```
AlexBramhill, line 43: I'd consider addressing this here -- We're setting ourselves up for some future tech debt on this test file as having the mocked function in this test now makes it so this suite cannot be reordered easily. If a dev adds a new test above this test, it'll behave different to below this test.

We should therefore either:
- Mock currencyIsDisabled everywhere, have a before each to clear the mock to clean up this config
- Spy on currencyIsDisabled just for this test, and have a cleanup for it after the assertion (mockRestore)
- Spy on it in the before each of `describe('when the facility has a currency that is disabled'` -- I prefer this as its related to the describe. We then clean up in the after each using a mock restore.


Also -- it's not async, so it should be `mockReturnValueOnce`

service-unavailable/index.html
```diff
@@ -5,13 +5,7 @@
 
     <title>Sorry, this service is unavailable - UK Export Finance</title>
     <meta name="viewport" content="width=device-width, initial-scale=1, viewport-fit=cover" />
-    <meta name="robots" content="noindex, nofollow, noarchive, noimageindex, nosnippet" />
```
AlexBramhill, line 8: Any reason for removing the robots?

AlexBramhill (commented):
Looking good

portal-api/src/utils/string.js
ttbarnes: I do not think CSS is a failsafe approach

Unfortunately, some old portal pages have the validation in `portal-api`, rather than the front end.

I don't want to introduce new patterns here, I think it's better to tweak/update the existing validation - wherever it is.

Have updated this to be one line

portal-api/src/v1/section-currency/index.js
```diff
@@ -1,15 +1,24 @@
...
 };
```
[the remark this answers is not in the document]
ttbarnes: Either is fine IMO, updated

portal-api/src/v1/validation/fields/currency-not-the-same-as-supply-contract-rules/currency.js
ttbarnes: Completely agree - I did think about creating an `isEmptyObject` type helper but this should be thought about another time.

portal-api/src/v1/validation/fields/currency-not-the-same-as-supply-contract-rules/currency.test.js
ttbarnes: `currencyIsDisabled` related tests (and related logic), is lacking.

This seems to be the only place where such tests exist because I added in this PR.

portal/templates/index.njk
```diff
@@ -16,69 +16,81 @@
...
+
     {% include "_partials/skip-link.njk" %}
+
     {% if user and user
-      .roles and user
-      .roles
-      .includes('admin') %}
+    .roles and user
+    .roles
+    .includes('admin') %}
```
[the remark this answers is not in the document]
ttbarnes: Agreed - sorry didn't spot this after indentation update, i've improved it.

portal/templates/index.njk
```diff
@@ -16,69 +16,81 @@
     <link rel="apple-touch-icon" sizes="152x152" href="/assets/images/govuk-apple-touch-icon-152x152.png">
     <link rel="apple-touch-icon" href="/assets/images/govuk-apple-touch-icon.png">
     <link rel="stylesheet" href="/assets/css/styles.css"/>
-    <meta property="og:image" content="/assets/images/govuk-opengraph-image.png">
-    {% block head %}{% endblock %}
```
AlexBramhill, line 20: We need this block head for post-login-redirect.njk to operate correctly

portal-api/src/v1/validation/fields/currency-not-the-same-as-supply-contract-rules/currency.test.js
AlexBramhill: Just fyi -- this is still susceptible to the above as it's only resetting the mockfn to be an empty mockfn -- we'd want `currencyIsDisabled.mockRestore();` in the after each.

I know this is very unlikely to be extended, but I tend to try and avoid leaving tests that _can_ leak in the future, as when you find them they can be a huge time sink to realise

gef-ui/public/js/disableFormSubmitOnSubmission.js
AlexBramhill: That is very surprising! Fair enough :)

portal/templates/index.njk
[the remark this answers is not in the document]
AlexBramhill: Thanks very much!

portal-api/src/utils/string.js
AlexBramhill: Sorry I also meant with F.E (but server side) validation of 'Select value' as well -- to avoid people posting with it...

I think we might be introducing a new pattern here regardless it seems? `hasValue` is an api function that is coupled with our front end copy -- which feels slightly odd?

portal-api/src/v1/validation/fields/currency-not-the-same-as-supply-contract-rules/currency.js
AlexBramhill: We have lodash installed -- Could we use `_.isEmpty(facility.currency);` maybe?

portal/templates/index.njk
ttbarnes: Not sure what happened here, reverted

portal-api/src/v1/validation/fields/currency-not-the-same-as-supply-contract-rules/currency.test.js
ttbarnes: Updated to use `mockRestore`

service-unavailable/index.html
ttbarnes: It's not used anywhere else. As far as I know, this `service-unavailable` is not actually used anywhere.

trade-finance-manager-ui/component-tests/case/underwriting/pricing-and-risk/edit-facility-risk-profile/_macros/section-facilities/facility-pricing-risk-table.component-test.js
```diff
@@ -123,14 +123,16 @@ describe(component, () => {
       wrapper.expectText(`[data-cy="facility-${params.facility._id}-risk-profile-value"]`).toRead(paramsWithRiskProfile.facility.tfm.riskProfile);
     });
 
-    it('should NOT render change link', () => {
-      wrapper = render(params);
+    describe('when params.userCanEdit is false', () => {
+      it('should NOT render `change` link', () => {
+        wrapper = render(params);
```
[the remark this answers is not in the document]
ttbarnes: Correct, i've made this clearer

trade-finance-manager-ui/component-tests/case/underwriting/pricing-and-risk/edit-facility-risk-profile/_macros/section-facilities/facility-pricing-risk-table.component-test.js
```diff
@@ -140,7 +142,9 @@ describe(component, () => {
 
         const expectedLink = `/case/${params.caseId}/underwriting/pricing-and-risk/facility/${params.facility._id}/risk-profile`;
 
-        wrapper.expectLink(`[data-cy="facility-${params.facility._id}-change-risk-profile-link"]`).toLinkTo(expectedLink, 'Change risk profile');
+        const expectedText = `Change ${params.facility.facilitySnapshot.type} ${params.facility.facilitySnapshot.ukefFacilityId} risk profile`;
+
+        wrapper.expectLink(`[data-cy="facility-${params.facility._id}-change-link"]`).toLinkTo(expectedLink, expectedText);
```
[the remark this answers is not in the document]
ttbarnes: The table column / design was messy - it's now changed to be visually simpler so this comment is out of date.

service-unavailable/index.html
AlexBramhill: I have a recollection it's used when express fails to start (I think its statically served?) I'm actually really not sure on this -- @abhi-markan can you recall?

portal/server/routes/contract/about/industryFields/index.js
AlexBramhill: Am happy for leaving this for now as it's not departing from existing patterns in this file

gef-ui/server/controllers/select-exporters-correspondence-address/index.js
AlexBramhill: Opps -- I resolved this comment thinking it was mine, happy with this to be resolved whenever you are

AlexBramhill (commented):
lgtm with exception of a couple of floating comments from the last PR (maybe forgot to post the pending replies?) :)

portal-api/src/utils/string.js
```diff
@@ -7,12 +7,15 @@ const isEmptyString = (str) => {
...
-};
+/**
+ * Check if a param has a value, i.e
+ * - the param is defined
+ * - the param is not an empty string
+ * - the param is not a 'Select value' string
+ * @param {unknown}
+ * @returns Boolean
+ */
+const hasValue = (str) => str && !isEmptyString(str) && str !== 'Select value';
```
ttbarnes, line 18: Need to revert this one liner change @AlexBramhill - with the one liner, it can return a string instead of a boolean, this then breaks other things.

trade-finance-manager-ui/templates/utilisation-reports/_macros/report-reconciliation-status.njk
```diff
@@ -6,8 +6,8 @@
 
   {% set tagColourClassForStatus = {
     'REPORT_NOT_RECEIVED': 'govuk-tag--red',
-    'PENDING_RECONCILIATION': '',
-    'RECONCILIATION_IN_PROGRESS': 'govuk-tag--blue',
+    'PENDING_RECONCILIATION': 'govuk-tag--blue',
+    'RECONCILIATION_IN_PROGRESS': 'govuk-tag--light-blue',
```
RLavender98, line 10: There is an equivalent component in portal that also needs updating please: https://github.com/UK-Export-Finance/dtfs2/blob/main/portal/templates/utilisation-report-service/previous-reports/_macros/report-reconciliation-status.njk

portal/server/routes/contract/about/industryFields/index.js
```diff
@@ -7,21 +7,31 @@ const industryFields = (postedSubmissionDetails, industrySectors) => {
 
   let industrySectorObj;
 
-  if (industrySectorCode) {
+  if (!industrySectorCode) {
```
ttbarnes, line 10: @AlexBramhill I've made some improvements to this function and added unit test coverage.

trade-finance-manager-ui/templates/utilisation-reports/_macros/report-reconciliation-status.njk
ttbarnes: Thanks @RLavender98 great shout - updated.

FYI, pending feedback from design RE status tags. In the meantime, I've just added "light-blue" so there are 2x different "blue tags".

RLavender98 (commented):
Done a quick look through utilisation report pages to check things still look nice and work and from what I can tell all looks good except for a few changes needed to some status components:

On the premium payments tab and on the keying sheet tab the statuses are all in capslock but with the design change need to be sentence case to match the utilisation report status on the bank reports page:
status-on-keying-sheet-tab-needs-to-be-lower-case-now
status-on-premium-payments-needs-to-be-lower-case-now

Also on the premium payments tab the status 'Does not match' doesn't seem to always fit in the box. I do wonder if once in sentence this might just fit, otherwise I'm not sure what the fix is. Josh had some issues when first implementing the table so probably looking into what is already there first and adjusting somehow. 
weirdness-with-long-status-on-premium-payments-tab

ttbarnes:
> Done a quick look through utilisation report pages to check things still look nice and work and from what I can tell all looks good except for a few changes needed to some status components:
> 
> On the premium payments tab and on the keying sheet tab the statuses are all in capslock but with the design change need to be sentence case to match the utilisation report status on the bank reports page: status-on-keying-sheet-tab-needs-to-be-lower-case-now status-on-premium-payments-needs-to-be-lower-case-now
> 
> Also on the premium payments tab the status 'Does not match' doesn't seem to always fit in the box. I do wonder if once in sentence this might just fit, otherwise I'm not sure what the fix is. Josh had some issues when first implementing the table so probably looking into what is already there first and adjusting somehow. weirdness-with-long-status-on-premium-payments-tab

Thanks @RLavender98 I noticed the uppercase issue - was going to tackle in another PR, but I have updated all reconciliation statuses to _not_ be capitilised in this PR.

I have since been unable to replicate the "does not match" layout/alignment issue

portal/server/routes/contract/about/industryFields/index.js
ttbarnes: Made some updates - resolving this comment in favour of a new fresh thread. https://github.com/UK-Export-Finance/dtfs2/pull/3684/files#r1856629286

RLavender98 (commented):
Thanks for the fixes Tony! I won't give an approval because haven't reviewed all files, but giving E2ETF sign off on the utilisation report page changes 👍

portal-api/src/v1/validation/fields/currency-not-the-same-as-supply-contract-rules/currency.js
ttbarnes: I don't see a need to use lodash here, `facility.currency?.id` does the job.

It's also a failsafe to specifically check for the ID - probably edge case, butt we don't want this to error if the object contains properties other than an ID

portal-api/src/utils/string.js
AlexBramhill: Oh -- that's very fair. Could also do it with `!!str && !isEmptyString(str) && str !== 'Select value'`, but easy with revert

portal-api/src/utils/string.js
```diff
@@ -7,10 +7,19 @@ const isEmptyString = (str) => {
   return false;
 };
 
+/**
+ * Check if a param has a value, i.e
+ * - the param is defined
+ * - the param is not an empty string
+ * - the param is not a 'Select value' string
+ * @param {unknown}
+ * @returns Boolean
```
AlexBramhill, line 16:
```suggestion
 * @returns {boolean}
```

AlexBramhill (approved):
lgtm, happy to approve after suggestion

e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/report-reconciliation/add-payment.spec.js
```diff
@@ -155,7 +155,7 @@ context('PDC_RECONCILE users can add a payment to a report', () => {
 
     cy.contains('Premium payments').should('exist');
 
-    cy.assertText(premiumPaymentsTable.status(FEE_RECORD_ID_ONE), FEE_RECORD_STATUS.MATCH);
+    cy.assertText(premiumPaymentsTable.status(FEE_RECORD_ID_ONE), 'Match');
```
Zainzzkk, line 158: should this not be from a constant?

ttbarnes: That'd be great - there is some confusion around these constants IMO. `FEE_RECORD_STATUS.MATCH` is in uppercase. The actual status is not.

gef-ui/server/controllers/select-exporters-correspondence-address/index.test.js
```diff
@@ -73,9 +73,12 @@ describe('controllers/select-exporters-correspondence-address', () => {
   });
 
   describe('Validate Select Exporters Correspondence Address', () => {
-    it('returns error object if address from select dropdown has not been selected', async () => {
+    beforeEach(() => {
       mockRequest.session.postcode = 'W1 7PD';
       mockRequest.session.addresses = JSON.stringify([{ addressLine1: 'line1' }]);
+    });
+
+    it('returns error object if address from select dropdown has not been selected', async () => {
```
Zainzzkk, line 81: should return... in this file please

gef-ui/server/controllers/select-exporters-correspondence-address/index.test.js
```diff
@@ -89,9 +92,21 @@ describe('controllers/select-exporters-correspondence-address', () => {
...
 
+    it('returns error object if address from select dropdown contains `Addresses Found`', async () => {
+      mockRequest.body.selectedAddress = 'Addresses Found';
+
+      await validateSelectExportersCorrespondenceAddress(mockRequest, mockResponse);
+
+      expect(mockResponse.render).toHaveBeenCalledWith(
+        'partials/select-exporters-correspondence-address.njk',
+        expect.objectContaining({
+          errors: expect.any(Object),
```
Zainzzkk, line 103: should we be more specific with which error? (or is this how the rest of the tests are for similar?)

portal-api/src/utils/string.js
Zainzzkk: having this without return true/false would be nicer but if it is not the current pattern, then fine to leave it as is

portal-api/src/v1/section-currency/index.js
```diff
@@ -1,8 +1,17 @@
 const { findOneCurrency } = require('../controllers/currencies.controller');
 
+/**
+ * Find a currency by currency code and return a simple object
+ * @param currencyCode - The currency code
```
Zainzzkk, line 5: do we need a type here {String} for currencyCode?

portal-api/src/v1/validation/fields/currency-not-the-same-as-supply-contract-rules/currency.js
```diff
@@ -1,17 +1,17 @@
-const { hasValue } = require('../../../../utils/string');
 const { orderNumber } = require('../../../../utils/error-list-order-number');
 const { currencyIsDisabled } = require('../currency');
 
 module.exports = (facility, errorList) => {
   const newErrorList = { ...errorList };
 
-  if (!hasValue(facility.currency)) {
+  if (!facility.currency?.id) {
```
Zainzzkk, line 7: do we need optional chaining for facility?.currency? ignore if not needed

gef-ui/server/controllers/select-exporters-correspondence-address/index.test.js
ttbarnes: This is how the rests of the tests are asserted. There are other instances. Not a fan, do not want to bloat this PR even more.

portal-api/src/v1/section-currency/index.js
ttbarnes: I have used the same pattern as other `currencyCode` param documentations.

portal/server/routes/contract/about/industryFields/index.js
```diff
@@ -7,24 +7,36 @@ const industryFields = (postedSubmissionDetails, industrySectors) => {
...
-      name: getIndustrySectorById(industrySectors, industrySectorCode).name,
-    };
-  } else {
-    submissionDetails['industry-sector'] = {};
+    if (industrySectorObj) {
+      submissionDetails['industry-sector'] = {
+        code: industrySectorCode,
+        name: industrySectorObj.name,
+      };
+    } else {
```
Zainzzkk, line 20: could we do this without the else? 

`submissionDetails['industry-sector'] = {};

 if (industrySectorObj) {
      submissionDetails['industry-sector'] = {
        code: industrySectorCode,
        name: industrySectorObj.name,
      };
    }
`

gef-ui/server/controllers/select-exporters-correspondence-address/index.test.js
ttbarnes: Updated all test descriptions.

portal-api/src/v1/validation/fields/currency-not-the-same-as-supply-contract-rules/currency.test.js
```diff
@@ -0,0 +1,72 @@
...
+      const expected = {
+        currency: {
+          order: orderNumber(errorList),
+          text: 'Facility currency is no longer available',
+        },
+      };
+
+      expect(result).toEqual(expected);
+    });
+  });
```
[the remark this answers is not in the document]
ttbarnes: Good spot, have added.

portal/server/routes/contract/about/industryFields/index.js
ttbarnes: Unfortunately not - left a comment in the code on line 21, in the else.

If no industry sector object is found via `getIndustrySectorById`, `submissionDetails['industry-sector']` will be set to undefined when it should be an empty object.

portal/templates/index.njk
```diff
@@ -19,66 +19,79 @@
...
+      crossorigin="anonymous"></script>
+
     <script
-        src="/assets/js/disableFormSubmitOnSubmission.js"
-        type="text/javascript"
-        integrity="sha512-RwF5ml+CiOHtgzu/uJmiCoytVi+TMCZGAKWvD+meYlY8bxi+kxC2qMDDiAyi17CoSK5rzB9DO4lej7QRXz8sQQ=="
-        crossorigin="anonymous"></script>
+      src="/assets/js/disableFormSubmitOnSubmission.js"
+      type="module"
+      integrity="sha512-RwF5ml+CiOHtgzu/uJmiCoytVi+TMCZGAKWvD+meYlY8bxi+kxC2qMDDiAyi17CoSK5rzB9DO4lej7QRXz8sQQ=="
```
Zainzzkk, line 94: it would be nice to have a tech debt PR to move these integrities to constants like EXIPs

ttbarnes: Agree - that'd be a massive job.

Zainzzkk (approved):

Zainzzkk (approved):

service-unavailable/index.html
ttbarnes: Resolving this - made a note to ask Abhi if we plan on doing this or if we should just remove the page.

PR #3687, opened 2024-10-15
feat(DTFS2-6892): add tfm api upsert user endpoint

AlexBramhill:
## Introduction :pencil2:
We need to add an endpoint to update the user in our database when the user has passed SSO.

Add endpoint to upsert user

limit existing endpoints depending on feature flag status


libs/common/src/errors/document-not-updated.error.ts
```diff
@@ -0,0 +1,14 @@
+import { DatabaseError } from './database.error';
+
+/**
+ * This error is thrown when a document is not successfully updated.
+ * It is not thrown by default by the db-client,
+ */
+export class DocumentNotUpdatedError extends DatabaseError {
+  constructor(documentId?: string) {
+    const message = documentId ? `Document update failed for document with id: ${documentId}` : 'Document update failed';
```
ttbarnes, line 9:
```suggestion
    const message = documentId ? `Document update failed for document with id: ${documentId}` : 'Document update failed with undefined document ID';
```

libs/common/src/schemas/tfm/entra-id-user-to-tfm-upsert-request.scema.test.ts
```diff
@@ -0,0 +1,60 @@
...
+    const result = ENTRA_ID_USER_TO_TFM_UPSERT_REQUEST_SCHEMA.parse(request);
+    expect(result).toEqual({
+      azureOid: request.oid,
+      email: request.verified_primary_email[0],
+      username: request.verified_primary_email[0],
+      teams: request.roles,
+      timezone: timezoneConfig.DEFAULT,
+      firstName: request.given_name,
+      lastName: request.family_name,
+    });
```
ttbarnes, line 52: `lastLogin` is missing from this, is it passing?

libs/common/src/schemas/tfm/user-upsert-request.schema.ts
```diff
@@ -0,0 +1,13 @@
+import z from 'zod';
+import { TfmTeamSchema } from './tfm-team.schema';
+
+export const USER_UPSERT_REQUEST_SCHEMA = z.object({
```
ttbarnes, line 4: I worry the "upsert request" style naming could be confusing/misleading, what do you think about something like `SSO_USER_REQUEST_SCHEMA` or maybe `ENTRA_...`? Prefer SSO though as it's clearer, not everyone knows what Entra is

libs/common/src/test-helpers/schemas/with-entra-id-user-schema.tests.ts
```diff
@@ -18,11 +17,11 @@ export function aValidEntraIdUser(): EntraIdUser {
     verified_secondary_email: ['a-secondary-email'],
     given_name: 'a-given-name',
     family_name: 'a-family-name',
-    groups: [ENTRA_ID_USER_GROUP.AZURE_SSO_GROUP_BUSINESS_SUPPORT],
+    roles: ['BUSINESS_SUPPORT'],
```
ttbarnes, line 20: constant?

libs/common/src/types/mongo-db-models/tfm-users.ts
```diff
@@ -1,18 +1,37 @@
...
+    username: string;
+    email: string;
+    teams: TeamId[];
+    timezone: string;
+    firstName: string;
+    lastName: string;
+    status: string;
+    lastLogin?: UnixTimestamp;
+    sessionIdentifier?: string;
+    // Audit records may not exist on a user if it has not been modified after Audit Logs is released
```
ttbarnes, line 33: What do we mean by "after audit logs is released" please, is this not live already?

trade-finance-manager-api/.eslintrc.js
```diff
@@ -18,7 +18,15 @@ const baseRules = {
...
     'error',
     {
-      devDependencies: ['**/*.test.{js,ts}', '**/*.api-test{,s}.{js,ts}', '**/*.spec.{js,ts}', '**/webpack.*.{js,ts}', '**/api-test*/**', '**/__mocks__/**'],
+      devDependencies: [
+        '**/*.test.{js,ts}',
+        '**/*.api-test{,s}.{js,ts}',
+        '**/*.spec.{js,ts}',
+        '**/webpack.*.{js,ts}',
+        '**/api-test*/**',
+        '**/*.tests.{js,ts}',
```
ttbarnes, line 27: What's the reason for adding `*.tests`, can we just use the existing `*.test`? Otherwise we could end up with some files using ".test", other ".tests"

trade-finance-manager-api/package.json
```diff
@@ -44,6 +44,7 @@
     "crypto": "^1.0.1",
     "date-fns": "^2.30.0",
     "dotenv": "16.0.3",
+    "escape-string-regexp": "^4.0.0",
```
ttbarnes, line 47: Do we need to use this package, it's very small and and increases deps/security concerns etc, can we have our own implementation?

trade-finance-manager-api/src/generateApp.js
```diff
@@ -29,7 +30,7 @@ const generateApp = () => {
   app.use(removeCsrfToken);
   app.use(healthcheck);
   app.use(passport.initialize());
-  app.post('/v1/login', loginController.login);
+  app.post('/v1/login', validateSsoFeatureFlagIsOff, loginController.login);
```
ttbarnes, line 33: Can we use `isTfmSsoFeatureFlagEnabled` instead? Other feature flag checks have this convention:

```
isTfmDealCancellationFeatureFlagEnabled
isTfmFacilityEndDateFeatureFlagEnabled
isTfmPaymentReconciliationFeatureFlagEnabled
```

trade-finance-manager-api/src/v1/controllers/user/user.controller.js
```diff
@@ -106,6 +107,19 @@ exports.update = async (_id, update, auditDetails, callback) => {
   });
 };
 
+/**
+ * Creates or updates a TFM user from an Entra user.
```
ttbarnes, line 111: "SSO" instead "Entra" or maybe "Entra SSO"?

trade-finance-manager-api/src/v1/controllers/user/user.controller.upsert-tfm-user-from-entra-id-user.test.js
```diff
@@ -0,0 +1,54 @@
...
+import { upsertTfmUserFromEntraIdUser } from './user.controller';
+
+jest.mock('../../services/user.service', () => ({
+  __esModule: true,
+  UserService: {
+    upsertUserFromEntraIdUser: jest.fn(),
+  },
+}));
+
+jest.mock('./helpers/mapUserData.helper', () => ({
```
ttbarnes, line 13: Do we need to mock this?

trade-finance-manager-api/src/v1/controllers/user/user.routes.js
```diff
@@ -60,6 +65,31 @@ module.exports.createTfmUser = (req, res, next) => {
...
+      });
+    }
+    if (error instanceof ZodError) {
+      return res.status(HttpStatusCode.BadRequest).send({
+        status: HttpStatusCode.BadRequest,
+        message: 'Error validating payload',
+      });
+    }
+
+    return next(error);
```
ttbarnes, line 89: Should we be returning `next` if there is an unknown error? Maybe return 400/500 or throw a generic error instead?

trade-finance-manager-api/src/v1/middleware/validate-sso-feature-flag.ts
```diff
@@ -0,0 +1,39 @@
...
+ * Used internally by validateSsoFeatureFlagIsOn and validateSsoFeatureFlagIsOff
+ */
+const createValidateSsoFeatureFlagMiddleware =
+  ({ desiredFeatureFlagState }: { desiredFeatureFlagState: boolean }): RequestHandler =>
+  (req, res, next) => {
+    const isFeatureFlagEnabled = isTfmSsoFeatureFlagEnabled();
+
+    if ((desiredFeatureFlagState && !isFeatureFlagEnabled) || (!desiredFeatureFlagState && isFeatureFlagEnabled)) {
+      console.error(
+        'Tfm SSO feature flag is in incorrect state to access this endpoint, current feature flag state is %s, required state is %s',
```
ttbarnes, line 18:
```suggestion
        'Tfm SSO feature flag is in incorrect state to access this endpoint, current feature flag state is %s, desired state is %s',
```

trade-finance-manager-api/src/v1/middleware/validate-sso-feature-flag.ts
```diff
@@ -0,0 +1,39 @@
...
+
+import { HttpStatusCode } from 'axios';
+import { RequestHandler } from 'express';
+
+/**
+ * @private
+ * Validates that the TFM SSO feature flag is in the correct state to access the endpoint
+ * Used internally by validateSsoFeatureFlagIsOn and validateSsoFeatureFlagIsOff
+ */
+const createValidateSsoFeatureFlagMiddleware =
```
ttbarnes, line 11:
```suggestion
const validateSsoFeatureFlag =
```

trade-finance-manager-api/src/v1/repo/user.repo.upsert-user.test.ts
```diff
@@ -0,0 +1,103 @@
...
+        mockSuccessfulUpdateOneResponse();
+      });
+
+      it('calls the correct collection', async () => {
+        mockSuccessfulGenerateAuditDatabaseRecordFromAuditDetails();
+        mockSuccessfulUpdateOneResponse();
+
+        await makeRequest();
+
+        expect(getCollectionMock).toHaveBeenCalledWith(MONGO_DB_COLLECTIONS.TFM_USERS);
```
ttbarnes, line 46: Can we add a `toHaveBeenCalledTimes` assertion please

ttbarnes (changes requested):

trade-finance-manager-api/src/generateApp.js
ttbarnes: Resolving this as I can see why we want to do this now => https://github.com/UK-Export-Finance/dtfs2/pull/3687#discussion_r1814745990

trade-finance-manager-api/src/v1/repo/user.repo.upsert-user.test.ts
AlexBramhill: Opps this test file shouldnt exist anymore -- but yes, will add to others

libs/common/src/test-helpers/schemas/with-entra-id-user-schema.tests.ts
AlexBramhill: No need in TS as it's typed so will kick up a fuss if it's not correct -- but I've updated

libs/common/src/types/mongo-db-models/tfm-users.ts
AlexBramhill: It is -- this is a comment from elsewhere. Due to us not running any migration scripts, this is still technically optional (as we cannot rely on a user object in the db having been updated since it's release)

trade-finance-manager-api/.eslintrc.js
AlexBramhill: We're using `tests` to contain reusable tests that are not standalone and only run when imported (if you search for `with-` you'll find several test files which contain reusable tests that do not run when running `npm run unit-test-dev` unless imported into other files)

trade-finance-manager-api/src/v1/controllers/user/user.routes.js
AlexBramhill: It'll get handled by the built in middleware and return a 500 automatigically :)

trade-finance-manager-api/package.json
AlexBramhill: I'd agree -- oddly this package has caused a fair bit of pain. We have it elsewhere however. I'll raise a ticket to look at removing this -- We have to be considerate to write our own implimentation of escaping strings, as this is published with a MIT licience, so we cannot simply copy and paste :)

libs/common/src/errors/document-not-updated.error.ts
AlexBramhill: I'm not sure we'll ever have an undefined document id -- I was using the ternary to keep the behaviour of this error in the same pattern of other existing database errors

libs/common/src/schemas/tfm/entra-id-user-to-tfm-upsert-request.scema.test.ts
AlexBramhill: Ah no its not passing -- this was as this was a late addition.

trade-finance-manager-api/src/v1/middleware/validate-sso-feature-flag.ts
AlexBramhill: This is actually the creator function which returns the validateSsoFeatureFlagMiddleware

libs/common/src/schemas/tfm/user-upsert-request.schema.ts
AlexBramhill: I am going to try and extract it out to `create_user_request` and alias it from there -- should be a bit clearer.

This request here has nothing to do with entra however -- it's mapped to our database model

trade-finance-manager-api/src/v1/controllers/user/user.controller.upsert-tfm-user-from-entra-id-user.test.js
AlexBramhill: the tldr of this is that if it isn't mocked, we'd have to maintain an accurate `UserService.upsertUserFromEntraIdUser` response -- which is not ideal at all (as since it's not from the repo layer, it could have mappings). I'd like to set the idea that in tests specifically like this (orchestration functions that call async, and do something with an async response), mocking is used as it's less maintainence overall -- but in tests like the others I've done we can do that neat way demonstrated in this PR and by yourself.

trade-finance-manager-api/src/v1/controllers/user/user.controller.js
AlexBramhill: Am going to add this is in the SSO login process (the reason to stick to entra is that this is more in line with GIFT.)

I do think you have a great point about making things easily identifiable as SSO though, so I'll update comments to suite!

libs/common/src/types/mongo-db-models/tfm-users.ts
ttbarnes: I think we should just remove the comment @AlexBramhill I think it just adds confusion, for me at least

I.e we don't comment about other optional things like `lastLogin`

libs/common/src/errors/document-not-updated.error.ts
ttbarnes: Makes sense to be consistent

libs/common/src/schemas/tfm/create-user-request.schema.ts
```diff
@@ -0,0 +1,19 @@
...
+ */
+export const CREATE_USER_REQUEST_SCHEMA = z.object({
+  azureOid: z.string(),
+  email: z.string(),
+  username: z.string(),
+  teams: z.array(TfmTeamSchema),
+  timezone: z.string(),
+  firstName: z.string(),
+  lastName: z.string(),
+  lastLogin: z.number().int().positive(), // unix timestamp
```
ttbarnes, line 18: There's already 2x different definitions of `lastLogin` with zod - this introduces a new 3rd definition, can we update to be the same?

```ts
lastLogin: z.number().optional(), // central API
lastLogin: z.number(), // libs/common
lastLogin: z.number().int().positive(), // libs/common
```

AlexBramhill: updating! thanks :)

libs/common/src/types/mongo-db-models/tfm-users.ts
AlexBramhill: Lovely, removed

libs/common/src/schemas/tfm/create-user-request.schema.ts
AlexBramhill: Commonised this the best I can. Also added in a `UnixTimestampMilliseconds` as we should be descriptive with our units -- as currently it leads to confusion and we cannot be sure what format a timestamp is in as a result.

libs/common/src/test-helpers/schemas/with-upsert-user-request-schema.tests.ts
```diff
@@ -0,0 +1,44 @@
...
+  });
+}
+
+export function getUpsertUserRequestFailureTestCases({ getTestObjectWithUpdatedUpsertUserRequestParams }: TestCasesParams) {
+  return getCreateUserRequestFailureTestCases({ getTestObjectWithUpdatedCreateUserRequestParams: getTestObjectWithUpdatedUpsertUserRequestParams });
+}
+
+export function getUpsertUserRequestSuccessTestCases({ getTestObjectWithUpdatedUpsertUserRequestParams }: TestCasesParams) {
+  return getCreateUserRequestSuccessTestCases({ getTestObjectWithUpdatedCreateUserRequestParams: getTestObjectWithUpdatedUpsertUserRequestParams });
+}
```
ttbarnes, line 44: I'm finding this file hard to follow because of the namings, can we simplify? Maybe something like this

```
getFailureTestCases
failureTestCases

getSuccessTestCases
successTestCases
```

libs/common/src/types/mongo-db-models/tfm-users.ts
```diff
@@ -1,18 +1,41 @@
...
 
-export type TfmUser = WithId<{
-  username: string;
-  email: string;
-  teams: TeamId[];
-  timezone: string;
-  firstName: string;
-  lastName: string;
+/**
+ * These properties are only on non-SSO users,
```
ttbarnes, line 7:
```suggestion
 * These properties are only on non-SSO users (TFM users),
```

trade-finance-manager-api/src/v1/controllers/user/user.controller.upsert-tfm-user-from-entra-id-user.test.js
ttbarnes: I think it'd be much simpler to do this?

```ts
it('should return the mapped user', async () => {
  const result = await makeRequest();

  const expected = mapUserData(upsertedUserResponse);

});
```

Since the actual implemention simply calls `mapUserData` with the upsert response

And then we don't need the following:

```ts
jest.mock('./helpers/mapUserData.helper', () => ({
  mapUserData: jest.fn(),
}));

when(jest.mocked(mapUserData)).calledWith(upsertedUserResponse).mockReturnValue(mappedUserDetails);
```

trade-finance-manager-api/src/v1/controllers/user/user.routes.test.js
```diff
@@ -0,0 +1,114 @@
...
+          const { req, res, next } = getHttpMocks(validRequest);
+
+          await upsertTfmUserFromEntraIdUser(req, res, next);
+
+          expect(res._getData()).toEqual(aSuccessfulResponseFromController);
+        });
+      });
+    });
+
+    function getHttpMocks(body) {
```
ttbarnes, line 106: Not a blocker for this PR, it looks like repeated or very similar approaches for this has been introduced in 5x places, if we search for `function getHttpMocks`, should be DRY

trade-finance-manager-api/src/v1/helpers/get-escaped-regex-from-string.js
```diff
@@ -0,0 +1,5 @@
+const escapeStringRegexp = require('escape-string-regexp');
+
+const getEscapedRegexFromString = (string) => new RegExp(`^${escapeStringRegexp(string)}$`, 'i');
```
ttbarnes, line 3: Can we add some documentation please

ttbarnes: Also - can this be in TS?

trade-finance-manager-api/src/v1/middleware/validate-sso-feature-flag.test.ts
```diff
@@ -0,0 +1,86 @@
...
+  function itCallsTheNextMiddleware({ makeRequest }: { makeRequest: RequestHandler }) {
+    it('calls the next middleware', () => {
+      const { req, res } = getHttpMocks();
+      const next = jest.fn();
+      makeRequest(req, res, next);
+      expect(next).toHaveBeenCalled();
+    });
+  }
+
+  function itReturnsAnErrorResponse({ makeRequest }: { makeRequest: RequestHandler }) {
```
ttbarnes, line 77: I like the DRYness of these functions but I think it overcomplicates the test, found it a bit tricky to follow, I think something like this would be nicer

```ts
describe('validateSsoFeatureFlag', () => {
  const { req, res } = httpMocks.createMocks();

  const next = jest.fn();

  beforeEach(() => {
    jest.resetAllMocks();
  });

  describe('when the SSO feature flag is enabled', () => {
    beforeEach(() => {
      mockIsTfmSsoFeatureFlag(true);
    });

    it('should call the next middleware', () => {
      expect(next).toHaveBeenCalled();
    });
  });
});
```

trade-finance-manager-api/src/v1/repo/user.repo.update-user-by-id.test.ts
```diff
@@ -0,0 +1,121 @@
...
+      });
+
+      jest.spyOn(mongoDbClient, 'getCollection').mockImplementation(getCollectionMock);
+    }
+
+    function mockSuccessfulFindOneAndUpdateResponse() {
+      mockFindOneAndUpdateResponse({ ok: 1, value: updateUserDatabaseResponse });
+    }
+
+    function mockUnsuccessfulFindOneAndUpdateResponse() {
```
ttbarnes, line 113: Similar to previous comments, `mockSuccessfulFindOneAndUpdateResponse` and `mockUnsuccessfulFindOneAndUpdateResponse` are only called once in individual before each's, I don't see a need to abstract into another function

trade-finance-manager-api/src/v1/services/user.service.transform-entra-id-user-to-tfm-upsert-user-request.test.ts
```diff
@@ -0,0 +1,22 @@
...
+
+    afterAll(() => {
+      jest.useRealTimers();
+    });
+
+    it('transforms an entra id user to a tfm upsert user request', () => {
+      const validEntraIdUser = aValidEntraIdUser();
+      const expectedResult = ENTRA_ID_USER_TO_UPSERT_USER_REQUEST_SCHEMA.parse(validEntraIdUser);
+      const result = UserService.transformEntraIdUserToUpsertUserRequest(validEntraIdUser);
+      expect(result).toEqual(expectedResult);
```
ttbarnes, line 19: Can we have `expected` instead of `expectedResult` please

trade-finance-manager-api/src/v1/services/user.service.upsert-user-from-entra-id-user.test.ts
```diff
@@ -0,0 +1,82 @@
...
+          userUpdate: transformedUser,
+          auditDetails,
+        });
+      });
+    });
+
+    describe('when multiple users are found with the email addresses', () => {
+      beforeEach(() => {
+        jest.mocked(UserRepo.findUsersByEmailAddresses).mockResolvedValue([existingUser, existingUser]);
+      });
```
ttbarnes, line 76: New line please

trade-finance-manager-api/test-helpers/with-validate-payload.tests.ts
```diff
@@ -0,0 +1,50 @@
+import httpMocks from 'node-mocks-http';
+import { HttpStatusCode } from 'axios';
+import { NextFunction, Request, Response } from 'express';
+
+type ValidatePayloadTestCases = { aTestCase: () => unknown; description: string }[];
+
+export const withValidatePayloadTests = ({
```
ttbarnes, line 7: Documentation

ttbarnes (changes requested):

trade-finance-manager-api/src/v1/controllers/user/user.controller.upsert-tfm-user-from-entra-id-user.test.js
AlexBramhill: I'm a little confused here.

We will have to mock `upsertedUserResponse` as an accurate `UserService.upsertedUserResponse` response for this test to be accurate, and have `mapUserData` behave in a way that is meaningful for this test.

This has a problem if `UserService` response changes -- this test wont fail as a result. We will have, in this test, an out of date `upsertedUserResponse` object that is still passing tests, and could be relied on by people as an accurate `upsertedUserResponse` when it is out of date.

We'd have to conciously and proactively look to maintain a valid `UserService.upsertedUserResponse` whenever we update `UserService.upsertedUserResponse`. This would mean every time we update this function, we would instinctively have to look through where it is called by all other functions, and update these.

If that's a pattern we wish to propergate, we should look to create a test double objects library, as if any async service responses change, we will need to ensure we remember to update the test double object.

---

So for this example -- 

```
describe('user controller', () => {
  describe('upsertTfmUserFromEntraIdUser', () => {
    const entraIdUser = 'anEntraIdUser';
    const auditDetails = 'auditDetails';

    let upsertedUserResponse
    let mappedUserDetails

    beforeEach(() => {
      jest.resetAllMocks();
      resetAllWhenMocks();

      upsertedUserReponse = aValidUpsertUserReponse();
      mappedUserDetails = mapUserData(upsertedUserResponse);
      when(jest.mocked(UserService.upsertUserFromEntraIdUser)).calledWith({ entraIdUser, auditDetails }).mockResolvedValue(upsertedUserResponse);
      when(jest.mocked(mapUserData)).calledWith(upsertedUserResponse).mockReturnValue(mappedUserDetails);
    });

    it('should upsert user in the database', async () => {
      await makeRequest();

      expect(UserService.upsertUserFromEntraIdUser).toHaveBeenCalledWith({ entraIdUser, auditDetails });
    });
    
    ....

    async function makeRequest() {
      return await upsertTfmUserFromEntraIdUser({ entraIdUser, auditDetails });
    }

    // This function is tying the unit test of an orchestration controller to the implimnentation of the user service it is calling
    // If the user service return value updates, this will not be automatically updated
    // and is prone to go out of sync with the actual values
    // leading to false confidence that the tests pass with the correct values
    async function aValidUpsertUserResponse() {
        return {
            _id: new ObjectId()
            azureOid: 'an-azure-oid',
            email: 'an-email',
            username: 'a-username',
            teams: ['BUSINESS_SUPPORT'],
            timezone: 'Europe/London',
            firstName: 'a-first-name',
            lastName: 'a-last-name',
            lastLogin: Date.now(),
        }
    }
  });
});
```

AlexBramhill: I think a prime example of it falling out of sync is all our tfm users in our unit tests -- a search for `user` shows how these are in varying degrees of out-of-date-ness

AlexBramhill: I'm not sure I really agree with this as I'd argue the smallest possible unit to test in a controller is it's orchestration of functions -- but as not to be dogmatic, I'll add a test double for this

AlexBramhill: Added a test double -- as a result of this and to ensure that this ticket isn't adding tech debt in this area, a lot of changes have been made to the current ticket. Sorry in advance!

A summary explanation of why there's so much changed:
- Extracted the creation of the `UserService` `upsertUserFromEntraIdResponse` into a new `userServiceMockResponses` file
- Created a new exported type for `upsertUserFromEntraIdResponse` in `UserService` so that if this type is updated, we will get typeErrors in our tests
- Utilised the existing `tfmUser` creator from DTFS-Central-API for `upsertUserFromEntraIdResponse` as apposed to copying this file to avoid new tech debt
- This required moving this into the common libs -- So I've updated all DTFS-Central-API references in that regard
- As this move continues the pattern inside our libs common for creation of test mock data inside `test-helpers/mock-data`, I've moved other creation of test mock data there from this PR instead of schemas --such as `create-user-request` etc
- As this now is completely separate from the schema, we should call these objects `aCreateTfmUserRequest` as apposed to `aCreateUserRequest` as this would be disingenuous, and allow for accidental misuse in Portal etc, as it would be `import {aCreateUserRequest} from '@ukex/dtfs2-common`, returning a `CreateUserRequest` type, which is highly confusing and also not keeping in pattern with `aTfmUser`
- As a result, I've updated all types/schemas in this PR to reference `TfmUser` as apposed to just `User`
- To avoid creating tech debt I've continued this name change wherever we are referencing these types (as we require params to be called i.e. `UpsertTfmUserFromEntraIdUserParams` now instead of `UpsertUserFromEntraIdUserParams`, the function is called `upsertTfmUserFromEntraUserId` instead of `upsertUserFromEntraUserId` to avoid confusion.

libs/common/src/test-helpers/schemas/with-upsert-tfm-user-request-schema.tests.ts
```diff
@@ -0,0 +1,39 @@
...
+  describe('with UPSERT_TFM_USER_REQUEST_SCHEMA tests', () => {
+    withSchemaTests({
+      schema,
+      failureTestCases: getUpsertTfmUserRequestFailureTestCases({ getTestObjectWithUpdatedUpsertTfmUserRequestParams }),
+      successTestCases: getUpsertTfmUserRequestSuccessTestCases({ getTestObjectWithUpdatedUpsertTfmUserRequestParams }),
+    });
+  });
+}
+
+export function getUpsertTfmUserRequestFailureTestCases({ getTestObjectWithUpdatedUpsertTfmUserRequestParams }: TestCasesParams) {
```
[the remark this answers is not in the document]
AlexBramhill: (This does have the additional cognetive load of piggybacking off of `create` but it should do to keep tests DRY)

trade-finance-manager-api/src/v1/repo/user.repo.update-user-by-id.test.ts
AlexBramhill: This was to allow the tests to be read easier -- I'll remove these

libs/common/src/constants/tfm/team-ids.ts
```diff
@@ -1,7 +1,21 @@
+import { PdcTeamId, TeamId } from '../../types';
+
+export const ALL_PDC_TEAM_IDS = ['PDC_READ', 'PDC_RECONCILE'] as const;
+
+export const ALL_TEAM_IDS = [
+  'UNDERWRITING_SUPPORT',
```
Zainzzkk, line 6: could we not have a seperate array for ALL_UNDERWRITING_IDS? like the PDC one

libs/common/src/errors/document-not-created.error.test.ts
```diff
@@ -0,0 +1,28 @@
+import { DocumentNotCreatedError } from './document-not-created.error';
+import { DatabaseError } from './database.error';
+
+describe('DocumentNotCreatedError', () => {
+  it('exposes the message the error was created with', () => {
```
Zainzzkk, line 5: these its should start with "should" as in it('should....

libs/common/src/schemas/tfm/entra-id-user-to-upsert-tfm-user-request.schema.test.ts
```diff
@@ -0,0 +1,70 @@
...
+  });
+
+  describe('when provided an invalid entra id user with incorrect fields', () => {
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    const request = { ...aEntraIdUser(), roles: 1 };
+    itShouldThrowAnError(request);
+  });
+
+  describe('when no primary email is provided', () => {
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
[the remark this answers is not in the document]
AlexBramhill: good spot -- We don't need this

libs/common/src/errors/document-not-created.error.test.ts
AlexBramhill: We have 4651 tests which do not begin with should and 4943 with do -- I think if we want to align these it's a different PR

libs/common/src/schemas/tfm/entra-id-user-to-upsert-tfm-user-request.schema.test.ts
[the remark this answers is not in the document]
Zainzzkk: we have a few more through this PR

libs/common/src/test-helpers/schemas/with-entra-id-user-schema.tests.ts
```diff
@@ -36,88 +30,90 @@ function getFailureTestCases({ getTestObjectWithUpdatedUserParams }: TestCasesPa
...
+        const { oid: _oid, ...rest } = aEntraIdUser();
         return getTestObjectWithUpdatedUserParams(rest);
       },
       description: 'the oid is missing',
     },
     {
       aTestCase: () => {
-        // eslint-disable-next-line @typescript-eslint/no-unused-vars, camelcase
-        const { verified_primary_email, ...rest } = aValidEntraIdUser();
+        // eslint-disable-next-line camelcase
```
Zainzzkk, line 40: is there a way around this

libs/common/src/constants/tfm/team-ids.ts
AlexBramhill: it's not currently used (and if we have it, we'd want to make a type of it etc etc) -- So I'm verging on yagni until we have a use for this variable to avoid any potential issues this causes

trade-finance-manager-api/test-helpers/with-validate-payload.tests.ts
```diff
@@ -0,0 +1,57 @@
...
+import { NextFunction, Request, Response } from 'express';
+
+type ValidatePayloadTestCases = { aTestCase: () => unknown; description: string }[];
+
+/**
+ * This is a reusable test to allow for validating payloads
+ * It is intended to be used for any controller-level function that utilises
+ * `req`, `res` and `next` as parameters
+ * It allows for commonisation of test cases and commonisation of tests
+ */
```
ttbarnes, line 12: Do we intend to use this for other non-SSO controllers?

dtfs-central-api/api-tests/v1/utilisation-reports/put-keying-data-mark-as-done.api-test.ts
```diff
@@ -36,7 +43,7 @@ describe(`PUT ${BASE_URL}`, () => {
   it('returns a 400 when the fee record ids are not a valid ids', async () => {
     // Arrange
     const requestBody = {
-      user: MOCK_TFM_USER,
+      user: tfmSessionUser,
```
ttbarnes, line 46: Could this just be the following?:

```ts
user: aTfmSessionUser(),
```

instead of having to add `let tfmSessionUser: TfmSessionUser;` and a `beforeEach`

trade-finance-manager-api/src/v1/controllers/user/user.controller.upsert-tfm-user-from-entra-id-user.test.ts
```diff
@@ -0,0 +1,48 @@
...
+      expect(UserService.upsertTfmUserFromEntraIdUser).toHaveBeenCalledWith({ entraIdUser, auditDetails });
+    });
+
+    it('should return the mapped user', async () => {
+      const result = await makeRequest();
+
+      expect(result).toEqual(mappedUserDetails);
+    });
+
+    async function makeRequest() {
```
ttbarnes, line 44: Do we really need this, we could just have `await upsertTfmUserFromEntraIdUser({ entraIdUser, auditDetails });` in the `beforeEach`

trade-finance-manager-api/src/v1/repo/user.repo.create-user.test.ts
```diff
@@ -0,0 +1,107 @@
...
+      jest.spyOn(mongoDbClient, 'getCollection').mockImplementation(getCollectionMock);
+    }
+
+    function mockSuccessfulInsertOneResponse() {
+      mockInsertOneResponse({ acknowledged: true, insertedId });
+    }
+
+    function mockUnsuccessfulInsertOneResponse() {
+      mockInsertOneResponse({ acknowledged: false, insertedId });
+    }
```
ttbarnes, line 101: I think functions like this convolute the test, especially as only used once, instead of these (`mockSuccessfulInsertOneResponse`, `mockUnsuccessfulInsertOneResponse`), can we just call `mockInsertOneResponse` instead please?

trade-finance-manager-api/test-helpers/with-validate-payload.tests.ts
AlexBramhill: Yep -- Anywhere where we want to test validate payload middleware

trade-finance-manager-ui/server/schemas/entra-id.schema.entra-id-authentication-result-schema.test.ts
```diff
@@ -21,16 +21,15 @@ describe('ENTRA_ID_AUTHENTICATION_RESULT_SCHEMA', () => {
 function aValidPayload(): EntraIdAuthenticationResult {
   return {
     accessToken: 'an-access-token',
-    account: { idTokenClaims: aValidEntraIdUser() },
+    account: { idTokenClaims: aEntraIdUser() },
```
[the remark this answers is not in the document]
AlexBramhill: renamed to `anEntraIdUser` -- keeping in existing pattern to get a new object (and the reason we do this is in case of mutations in test cases)

dtfs-central-api/api-tests/v1/utilisation-reports/put-keying-data-mark-as-done.api-test.ts
AlexBramhill: The reason I did it this way was to access parameters in tests without having to constantly redeclare the `const tfmSessionUser = aTfmSessionUser()`, but its only used once in this file -- makes accessing parameters a lot easier ie line 146 (only here for now, but generally makes this more scalable as apposed to adding the same line of  `const tfmSessionUser = aTfmSessionUser()` to every test )

trade-finance-manager-api/src/v1/controllers/user/user.controller.upsert-tfm-user-from-entra-id-user.test.ts
AlexBramhill: We need this -- We're looking at the result in the second test :)

trade-finance-manager-api/src/v1/repo/user.repo.create-user.test.ts
AlexBramhill: Updated

trade-finance-manager-api/src/v1/controllers/user/user.controller.upsert-tfm-user-from-entra-id-user.test.ts
ttbarnes: I think just having `await upsertTfmUserFromEntraIdUser({ entraIdUser, auditDetails });`  in individual tests would be nicer, but i'm not too fussed about this.

trade-finance-manager-api/test-helpers/with-validate-payload.tests.ts
AlexBramhill: (the point is here is that we should always throw a bad request so we can throw in a bunch of tests for this and its very reusable)

trade-finance-manager-api/src/v1/controllers/user/user.controller.upsert-tfm-user-from-entra-id-user.test.ts
AlexBramhill: Might keep as is -- my reasoning for liking having functions like this is it takes away cognitive load when debugging (as you're not presented with a lot of unneeded information)

libs/common/src/test-helpers/schemas/with-upsert-tfm-user-request-schema.tests.ts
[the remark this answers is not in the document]
AlexBramhill: Any suggestions for this? I'm not sure how we could as I think each segment of this provides important context (and probably doesnt want changing -- for instance if we changed the part `UpsertTfmUserRequest`, it'd add abiguity as it would not be identifiable to the `UpsertTfmUserRequest` type without having to open and check this file)?

trade-finance-manager-api/src/v1/controllers/user/user.controller.upsert-tfm-user-from-entra-id-user.test.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    beforeEach(() => {
+      jest.resetAllMocks();
+      resetAllWhenMocks();
+      entraIdUser = anEntraIdUser();
+      auditDetails = generateSystemAuditDetails();
+
+      upsertTfmUserFromEntraIdUserResponse = userServiceMockResponses.anUpsertTfmUserFromEntraIdUserResponse();
+      mappedUserDetails = mapUserData(upsertTfmUserFromEntraIdUserResponse);
+      // eslint-disable-next-line @typescript-eslint/unbound-method
```
Zainzzkk, line 33: can this be removed

AlexBramhill: A very interesting question -- I've got a little time to have a look at this. The reason why this doesn't matter is because it is a static class (there is a ESLint rule we could change to allow static classes unbound methods).

It seems an odd interaction between jest and classes -- We cannot move to DI in the constructor, which would avoid this issue all together, due to the heavy use of this controller service

AlexBramhill: That was good fun -- Updated

libs/common/src/schemas/tfm/entra-id-user-to-upsert-tfm-user-request.schema.test.ts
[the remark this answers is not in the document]
AlexBramhill: They should all be gone now

libs/common/src/test-helpers/schemas/with-upsert-tfm-user-request-schema.tests.ts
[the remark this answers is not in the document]
AlexBramhill: @ttbarnes I've had a thought about this -- I'm not sure I can reduce `getTestObjectWithUpdatedCreateTfmUserRequestParams: getTestObjectWithUpdatedUpsertTfmUserRequestParams` any further either -- with `getTestObjectWithUpdatedCreateTfmUserRequestParams` being because we're need a function that gets a test object with updated `CreateTfmUserRequest`...And we need that to be a function to pass in how we set whatever param is needing to be set by the `updatedUpsertTfmUserRequest` object...

trade-finance-manager-api/test-helpers/with-validate-payload.tests.ts
ttbarnes: I can't actually see this being used anywhere in this branch?

Concern here is that this seems to be a new pattern / different to other existing payload validations in controllers and unless we create a ticket in the backlog, we could have different patterns for doing the same thing

trade-finance-manager-api/src/v1/controllers/user/user.controller.upsert-tfm-user-from-entra-id-user.test.ts
ttbarnes: IMO / conversely I find such functions can cause more effort because you have to scroll to the very bottom to see what it's doing, before you can continue debugging from the top, especially when there is only 1x async call involved - I don't see benefit to creating a function purely to call another function

trade-finance-manager-api/test-helpers/with-validate-payload.tests.ts
AlexBramhill: Ah I havent actually added it to a test -- I have done now. It's a new testing pattern that allows us to test validation through using existing schema tests -- In `users.routes.test` we now use it -- and it allows us to test our validation is correctly being called without mocks
```
  user routes
    upsertTfmUserFromEntraIdUser
      when validating the payload
        √ should respond with a '400' if the oid is missing (5 ms)  
        √ should respond with a '400' if the verified primary email is missing (1 ms)
        √ should respond with a '400' if the verified secondary email is missing (1 ms)
        √ should respond with a '400' if the given name is missing  
        √ should respond with a '400' if the family name is missing 
        √ should respond with a '400' if the roles are missing      
        √ should respond with a '400' if the oid is not a string    
        √ should respond with a '400' if the verify primary email array is empty (1 ms)
        √ should respond with a '400' if the verify primary email is not a string array (1 ms)
        √ should respond with a '400' if the verify primary email is not an array
        √ should respond with a '400' if the verify secondary email is not a string array
        √ should respond with a '400' if the verify secondary email is not an array
        √ should respond with a '400' if the given name is not a string
        √ should respond with a '400' if the family name is not a string
        √ should respond with a '400' if the roles are not an array of user roles (1 ms)
        √ should respond with a '400' if the roles are not an array (1 ms)
        √ should respond with a '400' if the object is empty        
        √ should return a success status if a complete valid payload is present
        √ should return a success status if the verified secondary email array is empty
        √ should return a success status if the roles array is empty (1 ms)
        √ should return a success status if there is an extra field 

// The below is not part of the validation tests
      when the request body is valid
        √ calls the user controller
        when the upsert is unsuccessful
          when the error is an api error
            √ returns the api error with expected formatting (1 ms) 
          when the error is not an api error
            √ passes the error to further middleware (1 ms)
        when the upsert is successful                               
          √ returns the result of the user controller      
```

trade-finance-manager-api/src/v1/controllers/user/user.controller.upsert-tfm-user-from-entra-id-user.test.ts
AlexBramhill: removed

trade-finance-manager-api/test-helpers/with-validate-payload.tests.ts
AlexBramhill: In this case -- this function is checking the `ENTRA_ID_USER_SCHEMA` and the errors/status returned

dtfs-central-api/src/v1/routes/middleware/payload-validation/schemas/tfm-session-user.schema.ts
```diff
@@ -11,5 +11,5 @@ export const TfmSessionUserSchema = z.object({
   lastName: z.string(),
   status: z.string(),
   _id: MongoObjectIdSchema,
-  lastLogin: z.number().optional(),
+  lastLogin: UNIX_TIMESTAMP_MILLISECONDS_SCHEMA.optional(),
```
abhi-markan, line 14: What is going to be the default value?

abhi-markan (changes requested):

trade-finance-manager-api/src/v1/controllers/user/user.controller.js
```diff
@@ -106,6 +107,19 @@ exports.update = async (_id, update, auditDetails, callback) => {
...
 };
 
+/**
+ * Used during the SSO login process to keep the TFM user in sync with the Entra ID user.
+ * Upserts (Creates or updates) a TFM user from an Entra ID user.
+ * @param {object} upsertTfmUserFromEntraIdUserParams
+ * @param {import('@ukef/dtfs2-common').EntraIdUser} upsertTfmUserFromEntraIdUserParams.entraIdUser
+ * @param {import('@ukef/dtfs2-common').AuditDetails} upsertTfmUserFromEntraIdUserParams.auditDetails
+ */
+exports.upsertTfmUserFromEntraIdUser = async ({ entraIdUser, auditDetails }) => {
```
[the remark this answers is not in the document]
abhi-markan: Missed this one.

libs/common/src/test-helpers/mock-data/create-tfm-user-request.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export const aCreateTfmUserRequest = (): CreateTfmUserRequest => ({
+  azureOid: 'an-azure-oid',
+  email: 'an-email',
+  username: 'a-username',
+  teams: ['BUSINESS_SUPPORT'],
+  timezone: 'Europe/London',
+  firstName: 'a-first-name',
+  lastName: 'a-last-name',
+  lastLogin: Date.now(),
```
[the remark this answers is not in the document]
abhi-markan: Missed out this one.

libs/common/src/schemas/unix-timestamp.schema.ts
```diff
@@ -0,0 +1,7 @@
+import z from 'zod';
+
+export const UNIX_TIMESTAMP_SCHEMA = z.number().int().positive();
```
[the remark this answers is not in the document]
abhi-markan: Missed out this one.

dtfs-central-api/src/v1/routes/middleware/payload-validation/schemas/tfm-session-user.schema.ts
abhi-markan: Missed out this one.

libs/common/src/errors/document-not-updated.error.ts
```diff
@@ -0,0 +1,14 @@
+import { DatabaseError } from './database.error';
+
+/**
+ * This error is thrown when a document is not successfully updated.
+ * It is not thrown by default by the db-client,
+ */
+export class DocumentNotUpdatedError extends DatabaseError {
+  constructor(documentId?: string) {
```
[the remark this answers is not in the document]
abhi-markan: Missed this one.

abhi-markan (commented):
@AlexBramhill missed few comments.

libs/common/src/test-helpers/schemas/with-upsert-tfm-user-request-schema.tests.ts
```diff
@@ -0,0 +1,43 @@
...
+  });
+}
+
+export function getUpsertTfmUserRequestFailureTestCases({
+  getTestObjectWithUpdatedUpsertTfmUserRequestParams = (upsertTfmUserRequest) => upsertTfmUserRequest,
+}: TestCasesParams) {
+  return getCreateTfmUserRequestFailureTestCases({ getTestObjectWithUpdatedCreateTfmUserRequestParams: getTestObjectWithUpdatedUpsertTfmUserRequestParams });
+}
+
+export function getUpsertTfmUserRequestSuccessTestCases({
```
ttbarnes, line 39: I think have mentioned before, finding these really hard to read/follow, very long names. Is there anything we can do to shorten?

trade-finance-manager-api/src/v1/middleware/validate-sso-feature-flag.test.ts
```diff
@@ -0,0 +1,87 @@
...
+
+      itReturnsAnErrorResponse({ makeRequest });
+    });
+  });
+
+  function mockIsTfmSsoFeatureFlag(value: boolean) {
+    jest.mocked(isTfmSsoFeatureFlagEnabled).mockReturnValue(value);
+  }
+
+  function itCallsTheNextMiddleware({ makeRequest }: { makeRequest: RequestHandler }) {
```
ttbarnes, line 71: I find this `itReturnsAnErrorResponse` harder to follow - I know it's to be DRY but I think the trade off is readability, I think it'd be clearer to have `expect(next).toHaveBeenCalled();` (same with the error assertions), repeated in the tests themselves. If there was a lot more instances I would prefer the current approach

Can we add `toHaveBeenCalledTimes(1);` for the "next" assertions please?

trade-finance-manager-api/src/v1/repo/user.repo.create-user.test.ts
```diff
@@ -0,0 +1,99 @@
...
+      createTfmUserRequest = aCreateTfmUserRequest();
+      createUserDatabaseRequest = { ...createTfmUserRequest, status: USER.STATUS.ACTIVE, auditRecord };
+
+      insertedId = new ObjectId();
+    });
+
+    afterAll(() => {
+      jest.useRealTimers();
+    });
+    describe('when the create user is successful', () => {
```
ttbarnes, line 51: New line please

trade-finance-manager-api/src/v1/services/user.service.transform-entra-id-user-to-tfm-upsert-user-request.test.ts
```diff
@@ -0,0 +1,22 @@
...
+
+    afterAll(() => {
+      jest.useRealTimers();
+    });
+
+    it('transforms an entra id user to a tfm upsert user request', () => {
+      const validEntraIdUser = anEntraIdUser();
+      const expected = ENTRA_ID_USER_TO_UPSERT_TFM_USER_REQUEST_SCHEMA.parse(validEntraIdUser);
+      const result = UserService.transformEntraIdUserToUpsertTfmUserRequest(validEntraIdUser);
+      expect(result).toEqual(expected);
```
ttbarnes, line 19: Can we add some new lines after each const?

trade-finance-manager-api/src/v1/services/user.service.upsert-user-from-entra-id-user.test.ts
```diff
@@ -0,0 +1,86 @@
...
+    describe('when one user is found with the email addresses', () => {
+      beforeEach(() => {
+        jest.spyOn(UserRepo, 'findUsersByEmailAddresses').mockResolvedValue([existingUser]);
+        updateUserByIdSpy = jest.spyOn(UserRepo, 'updateUserById');
+      });
+
+      it('updates the user in the database', async () => {
+        await UserService.upsertTfmUserFromEntraIdUser({ entraIdUser, auditDetails });
+
+        expect(updateUserByIdSpy).toHaveBeenCalledWith({
```
ttbarnes, line 68: Can we add `toHaveBeenCalledTimes(1);` assertions in this file please

AlexBramhill (commented):
Ah -- My comments didnt post

trade-finance-manager-api/src/v1/controllers/user/user.controller.js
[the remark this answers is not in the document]
AlexBramhill: Ah my comment didnt post --
Hopefully there now

libs/common/src/test-helpers/mock-data/create-tfm-user-request.ts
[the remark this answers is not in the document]
AlexBramhill: Ah my comment didnt post --
Hopefully there now

libs/common/src/test-helpers/schemas/with-upsert-tfm-user-request-schema.tests.ts
ttbarnes: Bump @AlexBramhill

trade-finance-manager-api/src/v1/services/user.service.upsert-user-from-entra-id-user.test.ts
```diff
@@ -0,0 +1,87 @@
...
+    describe('when no users are found with the email addresses', () => {
+      beforeEach(() => {
+        jest.spyOn(UserRepo, 'findUsersByEmailAddresses').mockResolvedValue([]);
+        createUserSpy = jest.spyOn(UserRepo, 'createUser');
+      });
+
+      it('creates a new user in the database', async () => {
+        await UserService.upsertTfmUserFromEntraIdUser({ entraIdUser, auditDetails });
+
+        expect(createUserSpy).toHaveBeenCalledWith({
```
ttbarnes, line 52: Can we add an `toHaveBeenCalledTimes` assertion please

libs/common/src/test-helpers/schemas/with-upsert-tfm-user-request-schema.tests.ts
AlexBramhill: My comments seem to constantly not post -- trying again now.

abhi-markan (changes requested):
All looks good just one minor clarification please. Thanks

trade-finance-manager-api/src/v1/controllers/user/user.controller.js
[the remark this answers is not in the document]
AlexBramhill: Yep -- trade-finance-manager-api\src\v1\controllers\user\user.routes.test.js loc 49 to 86 :)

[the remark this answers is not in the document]
abhi-markan: Thanks you @AlexBramhill  look good to me.

abhi-markan (approved):

ttbarnes (approved):
LGTM, great work @AlexBramhill 👏

ttbarnes (approved):

abhi-markan (approved):

PR #3690, opened 2024-10-16
feat(DTFS2-7490): send email on cancellation submission

MarRobSoftwire:
## Introduction :pencil2:
When the submit cancellation endpoint is called, tfm-api should send an email to the PIM email address

- call GOV notify to send email when submitting deal, sending a different template depending on whether the effective date is the past/present or future

## Screenshots 📸 
Past effective date:
image

Future effective date:

image

Blank reason & 1st October
image

- Move send email service out of controllers folder
- check details page wasnt sending data correctly in tfm-ui
- fix flakey cancellation repo test

## Note ⚠️ 
The dtfs-central endpoint should return the cancelled deal & risk expired facilities to save making an additional request


trade-finance-manager-api/src/constants/email-template-ids.js
```diff
@@ -19,6 +19,7 @@ const MANUAL_AMENDMENT_DECISION_DECLINED = 'd51198ad-ffd7-4ae3-8850-b8526815da01
 const MANUAL_AMENDMENT_BANK_PROCEED = 'b381f2a0-d4f8-447d-b36c-ed886b652681';
 const MANUAL_AMENDMENT_BANK_WITHDRAW = '3111dc5c-f8f4-48ec-bd5a-fa4c4f8a9d1e';
 const INTERNAL_AMENDMENT_NOTIFICATION = 'd0ead538-454b-4d92-b35e-c1508cfc6962';
+const DEAL_CANCELLED_FUTURE_EFFECTIVE_DATE = 'template-id';
```
MarRobSoftwire, line 22: This needs updating with the new template id

trade-finance-manager-api/src/v1/controllers/deal-cancellation/submit-deal-cancellation.controller.ts
```diff
@@ -15,10 +16,14 @@ export type SubmitDealCancellationRequest = CustomExpressRequest<{
...
  * @param res - response
  */
-export const submitDealCancellation = (req: SubmitDealCancellationRequest, res: Response) => {
+export const submitDealCancellation = async (req: SubmitDealCancellationRequest, res: Response) => {
+  const cancellation = req.body;
   try {
-    // TODO: DTFS2-7298 - update cancellation in database
-    // TODO: DTFS2-7490 - send email
+    // TODO: DTFS2-7298 - update cancellation in database & return cancelled deal/facility ids
+    const ukefDealId = '00123144';
```
MarRobSoftwire, line 23: get these from database

trade-finance-manager-api/src/v1/api.js
```diff
@@ -164,6 +164,10 @@ const updatePortalFacility = async (facilityId, update, auditDetails) => {
   }
 };
 
+/**
+ * @param {string} dealId
+ * @returns {Promise<import('@ukef/dtfs2-common').TfmDeal}
```
MarRobSoftwire, line 169: This isn't _strictly_ accurate since the database models will be transformed when converted to JSON. Should I add a new DTO to reflect this?

ttbarnes: What's the transformation? Might be overkill for `api.js`

trade-finance-manager-api/src/v1/services/deal-cancellation/deal-cancellation.service.test.ts
```diff
@@ -0,0 +1,118 @@
...
+
+jest.mock('./helpers/format-facility-ids', () => ({
+  formatFacilityIds: jest.fn(() => mockFormattedFacilities),
+}));
+
+const dealId = 'dealId';
+
+describe('sendDealCancellationEmail', () => {
+  beforeAll(() => {
+    jest.useFakeTimers().setSystemTime(1729089801730);
```
ttbarnes, line 32: Is this really necessary, could we use `new Date` or similar instead?

trade-finance-manager-api/src/v1/services/deal-cancellation/deal-cancellation.service.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  public static async submitDealCancellation(dealId: string, dealCancellation: TfmDealCancellation) {
+    // TODO: DTFS2-7298 - update cancellation in database & return cancelled deal/facility ids
+
+    const {
+      dealSnapshot: { ukefDealId },
+    } = await api.findOneDeal(dealId);
+    const facilities = await api.findFacilitiesByDealId(dealId);
+
+    const ukefFacilityIds = facilities.map((facility) => facility.facilitySnapshot.ukefFacilityId).filter((id) => id !== null);
```
ttbarnes, line 34: Can we move this into its own helper/mapping function?

trade-finance-manager-api/src/v1/services/deal-cancellation/helpers/format-facility-ids.ts
```diff
@@ -0,0 +1 @@
+export const formatFacilityIds = (facilityIds: string[]) => facilityIds.map((id, index) => ` ${index + 1}. Facility ID ${id}`).join(`\n`);
```
ttbarnes, line 1: Documentation?

trade-finance-manager-ui/server/controllers/case/cancellation/check-details.controller.ts
```diff
@@ -71,13 +77,6 @@ export const postDealCancellationDetails = async (req: PostDealCancellationDetai
       return res.redirect('/not-found');
     }
 
-    const submissionTypeCanBeCancelled = canSubmissionTypeBeCancelled(deal.dealSnapshot.submissionType);
-    const cancellationIncomplete = reason === undefined || !bankRequestDate || !effectiveFrom;
-
-    if (!submissionTypeCanBeCancelled || cancellationIncomplete) {
-      return res.redirect(`/case/${_id}/deal`);
-    }
```
ttbarnes, line 79: Why aren't we validating this now please?

ttbarnes (changes requested):

trade-finance-manager-api/src/v1/api.js
MarRobSoftwire: In general, everything that's sent as JSON will lose some information - e.g. `Date` and `ObjectId` properties will be converted to string. We could get around this by adding a `Jsonify` type helper (a la https://effectivetypescript.com/2020/04/09/jsonify/) 

for the deal in particular `dtfs-central-api\src\v1\controllers\tfm\deal\tfm-get-deal.controller.js` does some weird mapping, though for my particular use case it doesnt have any impact.

Keeping _as-is_ is the simplest thing to do but it could cause confusion down the line. I'd caution against trusting any of the types declared in a `js` file anyway, so its not tooo big an issue

trade-finance-manager-api/src/v1/services/deal-cancellation/deal-cancellation.service.test.ts
MarRobSoftwire: This allows me to hard code the output as `'16 October 2024'` without having to use date-fns to format it.
I have a slight preference towards this, since then the tests arent assuming i've written `dd MMMM yyyy` correctly

trade-finance-manager-ui/server/controllers/case/cancellation/check-details.controller.ts
MarRobSoftwire: I found this was giving some pretty weird behaviour: when the validation wasnt met the browser was redirected to `/case/${_id}/deal` - the same page as when the validation is met & the deal is cancelled.

Given the user can't actually change anythign on this page, it makes the most sense to send the data to the api & if the api rejects it, let the browser render `problem-with-service.njk`.
Given this is a post request, sending weird data requires using Postman, or modifying the HTML, so it should never happen in normal usage

trade-finance-manager-api/src/v1/api.js
ttbarnes: Agree @MarRobSoftwire 

I think it would be overkill at this point in time to be super specific in `api.js`, especially considering it's essentially a "middleman".

trade-finance-manager-api/src/v1/services/deal-cancellation/deal-cancellation.service.test.ts
ttbarnes: Can we use date-fns? I'm finding these mysterious; Currently if a future developer was to come into this file, the hard coded dates don't have any meaning or context with these:

```ts
setSystemTime(1729089801730);

bankRequestDate: 1729089801730,
effectiveFrom: 1729176201730
```

I.e I have no idea what these dates actually are.

trade-finance-manager-api/src/v1/services/deal-cancellation/deal-cancellation.service.test.ts
```diff
@@ -0,0 +1,118 @@
...
+  ),
+}));
+
+jest.mock('./helpers/format-facility-ids', () => ({
+  formatFacilityIds: jest.fn(() => mockFormattedFacilities),
+}));
+
+const dealId = 'dealId';
+
+describe('sendDealCancellationEmail', () => {
```
ttbarnes, line 30: Can we test the other methods as well please?

trade-finance-manager-api/src/v1/services/deal-cancellation/deal-cancellation.service.test.ts
MarRobSoftwire: Sure, I see that & its maybe more clear to use some constants at the top. I've switched to using `format` and `add` from date-fns though

trade-finance-manager-api/src/v1/services/deal-cancellation/deal-cancellation.service.test.ts
AlexBramhill: Sorry -- I knew I said I'd set up a call to chat about our testing approaches, but this week has got away with me.

Are you sure we need to be testing private functions? It feels slightly odd to expose them from the service for the sake of only testing, as they are then accessible anywhere in the service?

I generally think of private methods as 'part of the unit being tested' -- and they are only really broken from the main function for readability (as apposed to say, business logic etc)

If we were to make all our private methods public, I'd expect to mock them otherwise are we not testing the behaviour twice -- ie

```
public static anOtherwisePrivateMethod () {}

public static aPublicMethodThatUsesPrivateMethod() {}
```

Are we wanting to test `anOtherwisePrivateMethod ` standalone, as well as part of in `aPublicMethodThatUsesPrivateMethod`? Or are we mocking it in `aPublicMethodThatUsesPrivateMethod`?

I'd lean to only testing private methods once -- in the public methods they are used in as I think this is one `unit` to test?

I'll pop in a meeting for Monday now Abhi is back!

ttbarnes: I thought we agreed before that every function should be tested

Maybe we should convert this to pure functions so it's not a class with public/private methods

AlexBramhill: I think my confusion is about whether we then duplicate test efforts on this in the tests for the function where it's called as well?

ttbarnes: Not sure why we'd want to duplicate test efforts, I would expect something like this:

```ts
describe('deal-cancellation.service', () => {
  describe('sendDealCancellationEmail', () => {});

  describe('getFacilityIds', () => {});

  describe('submitDealCancellation', () => {
    describe('it should call api.findOneDeal', () => {});

    describe('it should call api.findFacilitiesByDealId', () => {});

    describe('it should call sendDealCancellationEmail', () => {});
  });
});
```

AlexBramhill: Ah -- In this case we align that `sendDealCancellationEmail` should be mocked as it's async... But if `sendDealCancellationEmail` would we still look to mock it in our tests as well as test it? I think this is where my main confusion around this lies... ie:

```
describe('deal-cancellation.service', () => {
  describe('sendDealCancellationEmail', () => {
    describe('it should do whatever send deal cancellation email does when successful', () => {});

    describe('it should do whatever send deal cancellation email does when successful (another thing the function does)', () => {});

    describe('it should do whatever error throwing deal cancellation email does when it fails', () => {});

    describe('it should do whatever error throwing deal cancellation email does when it fails in another way', () => {});
});

  describe('getFacilityIds', () => {});

  describe('submitDealCancellation', () => {
    describe('it should call api.findOneDeal', () => {});

    describe('it should call api.findFacilitiesByDealId', () => {});

    describe('it should do whatever send deal cancellation email does when successful', () => {});

    describe('it should do whatever send deal cancellation email does when successful (another thing the function does)', () => {});

    describe('it should do whatever error throwing deal cancellation email does when it fails', () => {});

    describe('it should do whatever error throwing deal cancellation email does when it fails in another way', () => {});
  });
});
```

ttbarnes: This is what i'm thinking - always open to suggestions though

```ts
describe('deal-cancellation.service', () => {
  describe('sendDealCancellationEmail', () => {
    it('should call api.findOneTeam', () => { });

    describe('when X date is Y', () => {
      it('should call sendTfmEmail with template no.1', () => { });
    });

    describe('when X date is Z', () => {
      it('should call sendTfmEmail with template no.2', () => { });
    });
  });

  describe('getFacilityIds', () => {
    it('should return an array of mapped facility IDs without null entries', () => { });
  });

  describe('submitDealCancellation', () => {
    it('should call api.findOneDeal', () => {});

    it('should call api.findFacilitiesByDealId', () => {});

    it('should call sendDealCancellationEmail', () => {});

    describe('when sendEmail errors', () => {
      it('should call throw an error', () => { });
    });
  });
});
```

Maybe the email function could/should be moved into it's own directory and then we could call/mock like `dealCancellationEmail.send();`, but don't want to introduce another new pattern.

trade-finance-manager-api/src/v1/services/deal-cancellation/deal-cancellation.service.ts
```diff
@@ -0,0 +1,58 @@
...
+   */
+  public static async submitDealCancellation(dealId: string, dealCancellation: TfmDealCancellation) {
+    // TODO: DTFS2-7298 - update cancellation in database & return cancelled deal/facility ids
+
+    const {
+      dealSnapshot: { ukefDealId },
+    } = await api.findOneDeal(dealId);
+    const facilities = await api.findFacilitiesByDealId(dealId);
+
+    const ukefFacilityIds = this.getFacilityIds(facilities);
```
abhi-markan, line 54: Add some validation to ensure they are returned, in any case `ukefFacilityIds` returns empty array we don't want to send an email with empty bullet points.

trade-finance-manager-api/src/v1/services/deal-cancellation/helpers/format-facility-ids.test.ts
```diff
@@ -0,0 +1,19 @@
+import { formatFacilityIds } from './format-facility-ids';
+
+describe('formatFacilityIds', () => {
+  it('formats the id correctly', () => {
```
abhi-markan, line 4: Can we add an assertion with an empty array please.

abhi-markan (changes requested):
Nice work, few minor comments to address please.

trade-finance-manager-api/src/v1/helpers/amendment.helpers.js
```diff
@@ -1,7 +1,7 @@
 const { format, fromUnixTime } = require('date-fns');
 const { CURRENCY, AMENDMENT_STATUS } = require('@ukef/dtfs2-common');
 const api = require('../api');
-const sendTfmEmail = require('../controllers/send-tfm-email');
+const sendTfmEmail = require('../services/send-tfm-email');
```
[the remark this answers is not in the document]
MarRobSoftwire: `sendTfmEmail` isn't a controller & i think its confusing to keep it in that folder

trade-finance-manager-api/src/v1/services/deal-cancellation/deal-cancellation.service.ts
```diff
@@ -0,0 +1,58 @@
...
+   * @param dealCancellation deal cancellation object
+   * @param facilityIds ukef facility ids
+   */
+  private static async sendDealCancellationEmail(ukefDealId: string, dealCancellation: TfmDealCancellation, facilityIds: string[]) {
+    const effectiveFromDate = toDate(dealCancellation.effectiveFrom);
+    const endOfToday = endOfDay(new Date());
+
+    const { email: pimEmail } = await api.findOneTeam(TEAMS.PIM.id);
+
+    const emailTemplateId = isAfter(effectiveFromDate, endOfToday) ? CANCEL_DEAL_FUTURE_DATE : CANCEL_DEAL_PAST_DATE;
```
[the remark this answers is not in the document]
MarRobSoftwire: If there's no pimEmail should it throw an error or default to some other value?

trade-finance-manager-api/src/v1/services/deal-cancellation/deal-cancellation.service.ts
```diff
@@ -0,0 +1,58 @@
...
+    });
+  }
+
+  /**
+   * Maps the facilities objects into their UKEF ids
+   * @param facilities Facilities
+   * @returns UKEF facility Ids
+   */
+  private static getFacilityIds(facilities: TfmFacility[]) {
+    return facilities.map((facility) => facility.facilitySnapshot.ukefFacilityId).filter((id) => id !== null);
```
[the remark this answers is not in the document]
MarRobSoftwire: If the ID is one of these the email won't be sent - is that the behaviour you'd want?

trade-finance-manager-api/src/v1/services/deal-cancellation/deal-cancellation.service.ts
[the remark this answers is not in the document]
MarRobSoftwire: I've gone for sending generic errors (so the service will 500 in this case) - this feels like the most accurate reflection? If the database is down/there's no PIM user with an email for some reason I think we should be getting 500s

trade-finance-manager-api/src/v1/emails/AIN-MIN-confirmation/gef-facilities-list.js
```diff
@@ -55,11 +55,11 @@ const facilityFieldsObj = ({
...
   hasBeenIssued: hasBeenIssued && mapIssuedValue(hasBeenIssued),
-  coverStartDate: coverStartDate && format(Number(coverStartDate), LONG_FORM_DATE_FORMAT),
-  coverEndDate: coverEndDate && format(parseISO(coverEndDate), LONG_FORM_DATE_FORMAT),
+  coverStartDate: coverStartDate && format(Number(coverStartDate), DATE_FORMATS.DO_MMMM_YYYY),
+  coverEndDate: coverEndDate && format(parseISO(coverEndDate), DATE_FORMATS.DO_MMMM_YYYY),
   isUsingFacilityEndDate: mapBooleanToYesOrNo(isUsingFacilityEndDate),
-  facilityEndDate: facilityEndDate && format(parseISO(facilityEndDate), LONG_FORM_DATE_FORMAT),
-  bankReviewDate: bankReviewDate && format(parseISO(bankReviewDate), LONG_FORM_DATE_FORMAT),
+  facilityEndDate: facilityEndDate && format(parseISO(facilityEndDate), DATE_FORMATS.DO_MMMM_YYYY),
+  bankReviewDate: bankReviewDate && format(parseISO(bankReviewDate), DATE_FORMATS.DO_MMMM_YYYY),
```
ttbarnes, line 62: Can we destructure `DO_MMMM_YYYY` throughout?

trade-finance-manager-api/src/v1/services/deal-cancellation/deal-cancellation.service.test.ts
```diff
@@ -0,0 +1,154 @@
...
+  findFacilitiesByDealId: jest.fn(() =>
+    ukefFacilityIds.map((ukefFacilityId) => ({
+      facilitySnapshot: {
+        ukefFacilityId,
+      },
+    })),
+  ),
+}));
+
+jest.mock('./helpers/format-facility-ids', () => ({
```
ttbarnes, line 25: Can we not mock the helper, test could just do this instead down below


```ts
formattedFacilitiesList: formatFacilityIds(ukefFacilityIds)
```

trade-finance-manager-api/src/v1/services/deal-cancellation/deal-cancellation.service.test.ts
```diff
@@ -0,0 +1,154 @@
...
+
+        // Act
+        await DealCancellationService.sendDealCancellationEmail(ukefDealId, aDealCancellation(), ukefFacilityIds);
+
+        // Assert
+        expect(sendTfmEmail).toHaveBeenCalledTimes(1);
+        expect(sendTfmEmail).toHaveBeenCalledWith(CANCEL_DEAL_PAST_DATE, mockPimEmailAddress, {
+          cancelReason: dealCancellation.reason,
+          bankRequestDate: format(today, 'dd MMMM yyyy'),
+          effectiveFromDate: format(today, 'dd MMMM yyyy'),
```
ttbarnes, line 129: Would be nice to use constants for `'dd MMMM yyyy'` - not a blocker, we have 42 instances of this, should be done in a different PR

trade-finance-manager-api/src/v1/services/deal-cancellation/deal-cancellation.service.ts
```diff
@@ -0,0 +1,64 @@
...
+import { formatFacilityIds } from './helpers/format-facility-ids';
+import { UKEF_ID } from '../../../constants/deals';
+
+export class DealCancellationService {
+  /**
+   * Sends email to PIM email address confirming the cancellation has been submitted
+   * @param ukefDealId ukef deal id
+   * @param dealCancellation deal cancellation object
+   * @param facilityIds ukef facility ids
+   * @private - This function is exported for unit testing only and should not be used outside of the service
```
ttbarnes, line 15: Should we be saying `private` when it's not actually private?

trade-finance-manager-api/src/v1/services/deal-cancellation/deal-cancellation.service.ts
```diff
@@ -0,0 +1,64 @@
...
+   */
+  public static async submitDealCancellation(dealId: string, dealCancellation: TfmDealCancellation) {
+    // TODO: DTFS2-7298 - update cancellation in database & return cancelled deal/facility ids
+
+    const { dealSnapshot } = await api.findOneDeal(dealId);
+    const facilities = await api.findFacilitiesByDealId(dealId);
+
+    const ukefDealId = dealSnapshot.dealType === DEAL_TYPE.BSS_EWCS ? dealSnapshot.details.ukefDealId : dealSnapshot.ukefDealId;
+
+    const ukefFacilityIds = facilities
```
ttbarnes, line 54: This is pretty heavy, can we move into it's own function e.g something like `mapAndFilterFacilityIds`

Will be nice for testing as well

trade-finance-manager-api/src/v1/services/deal-cancellation/helpers/format-facility-ids.test.ts
```diff
@@ -0,0 +1,30 @@
...
+    // Act
+    const result = formatFacilityIds(facilityIds);
+
+    // Assert
+    expect(result).toEqual(
+      ` 1. Facility ID ${facilityIds[0]}
+ 2. Facility ID ${facilityIds[1]}
+ 3. Facility ID ${facilityIds[2]}
+ 4. Facility ID ${facilityIds[3]}`,
+    );
```
ttbarnes, line 17: `const expected`?

ttbarnes (changes requested):

trade-finance-manager-api/src/v1/services/deal-cancellation/deal-cancellation.service.ts
```diff
@@ -0,0 +1,64 @@
...
+
+    const { email: pimEmail } = await api.findOneTeam(TEAMS.PIM.id);
+
+    if (!pimEmail) {
+      throw Error('Failed to send deal cancellation email to PIM');
+    }
+
+    const emailTemplateId = isAfter(effectiveFromDate, endOfToday) ? CANCEL_DEAL_FUTURE_DATE : CANCEL_DEAL_PAST_DATE;
+
+    const formattedEffectiveFromDate = format(dealCancellation.effectiveFrom, DATE_FORMATS.D_MMMM_YYYY);
```
abhi-markan, line 29:
```suggestion
    const effectiveFromDate= format(dealCancellation.effectiveFrom, DATE_FORMATS.D_MMMM_YYYY);
```

trade-finance-manager-api/src/v1/services/deal-cancellation/deal-cancellation.service.ts
```diff
@@ -0,0 +1,64 @@
...
+    const formattedBankRequestDate = format(dealCancellation.bankRequestDate, DATE_FORMATS.D_MMMM_YYYY);
+
+    await sendTfmEmail(emailTemplateId, pimEmail, {
+      ukefDealId,
+      effectiveFromDate: formattedEffectiveFromDate,
```
abhi-markan, line 34:
```suggestion
      effectiveFromDate,
```

trade-finance-manager-api/src/v1/services/deal-cancellation/deal-cancellation.service.ts
```diff
@@ -0,0 +1,64 @@
...
+      bankRequestDate: formattedBankRequestDate,
+      cancelReason: dealCancellation.reason || '-',
```
abhi-markan, line 36: Add this above please.

```ts
const cancelReason = dealCancellation?.reason ?? '-'
```


```suggestion
      cancelReason,
```

trade-finance-manager-api/src/v1/services/deal-cancellation/deal-cancellation.service.ts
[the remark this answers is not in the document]
abhi-markan: Yes please to avoid email being sent out, this is an extremely edge case but if in case deal does makes it way to TFM with those IDs then cancellation should not send email with those IDs. Thanks

abhi-markan (changes requested):
Few minor comments to address please. Thanks

trade-finance-manager-api/src/v1/services/deal-cancellation/deal-cancellation.service.ts
```diff
@@ -0,0 +1,64 @@
...
+
+export class DealCancellationService {
+  /**
+   * Sends email to PIM email address confirming the cancellation has been submitted
+   * @param ukefDealId ukef deal id
+   * @param dealCancellation deal cancellation object
+   * @param facilityIds ukef facility ids
+   * @private - This function is exported for unit testing only and should not be used outside of the service
+   */
+  public static async sendDealCancellationEmail(ukefDealId: string, dealCancellation: TfmDealCancellation, facilityIds: string[]): Promise<void> {
```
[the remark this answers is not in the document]
MarRobSoftwire: What would the try catch contain? We need the errors to bubble up to `trade-finance-manager-api\src\v1\controllers\deal-cancellation\submit-deal-cancellation.controller.ts` in order to send the api responses accordingly.

I dont think adding the following would be very helpful, especially since it would hide the location of the original error when its logged a second time in the controller.
```
try {

} catch (e){
  console.error('an error', e)
  throw e
}
```

trade-finance-manager-api/src/v1/services/deal-cancellation/deal-cancellation.service.ts
MarRobSoftwire: `const effectiveFromDate = toDate(dealCancellation.effectiveFrom);` is defined at the top of the function in order to test whether the date is in the future or not. In order to keep these clearly distinct I think adding `formatting` is useful

trade-finance-manager-api/src/v1/services/deal-cancellation/deal-cancellation.service.ts
MarRobSoftwire: We need to use `||` since we want it to default to `'-'` when the reason is an empty string. The dealCancellation is definitely defined here too (using zod payload validation) so I dont think using optional chaining is necessary

trade-finance-manager-api/src/v1/services/deal-cancellation/deal-cancellation.service.ts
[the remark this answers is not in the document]
ttbarnes: I was thinking it could contain a `console.info` and throw an error in the catch. Makes sense to let the error bubble up.

Can we add a `console.info` at the top?

trade-finance-manager-api/src/v1/services/deal-cancellation/deal-cancellation.service.ts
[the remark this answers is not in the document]
MarRobSoftwire: Should the email still be sent? Thats the current behaviour as written

trade-finance-manager-api/src/v1/services/deal-cancellation/deal-cancellation.service.ts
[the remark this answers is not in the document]
ttbarnes: Bump, can we add a `console.info`?

trade-finance-manager-api/src/v1/services/deal-cancellation/deal-cancellation.service.ts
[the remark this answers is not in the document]
abhi-markan: I would not expect an email to be sent to the user with facility IDs as `PENDING` and/or `00000000`.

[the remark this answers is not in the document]
MarRobSoftwire: If one facility has ID `PENDING` and another is valid, should the email be sent with only one ID?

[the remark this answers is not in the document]
MarRobSoftwire: Have changed the getUkefFacilityIds function & throw an error now if any facility has one of these Id

abhi-markan (approved):
LGTM 👍🏼

ttbarnes (approved):

PR #3692, opened 2024-10-16
feat(FN-3493): persist selected records on add payment screen

sw-mattw:
## Introduction :pencil2:
At present, clicking the back/cancel buttons on the add payment screen does not persist the selected fee records. This behaviour does not align with that of the 'add to an existing payment' screen.

- Update add payment screen template to accept a back link href
- Update add payment view model to accept back link href
- Update the add payment controller to use the existing back link href helper function
- Updates to the relevant unit/component/E2E tests

N/A

### DIFF
--- a/e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/add-payment.spec.js
+++ b/e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/add-payment.spec.js
@@ -9,6 +9,7 @@ import {
 import pages from '../../pages';
 import USERS from '../../../fixtures/users';
 import { NODE_TASKS } from '../../../../../e2e-fixtures';
+import relative from '../../relativeURL';
 import { getMatchingTfmFacilitiesForFeeRecords } from '../../../support/utils/getMatchingTfmFacilitiesForFeeRecords';
 
 context('PDC_RECONCILE users can add a payment to a report', () => {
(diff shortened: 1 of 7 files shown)


e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/add-payment.spec.js
```diff
@@ -193,4 +198,88 @@ context('PDC_RECONCILE users can add a payment to a report', () => {
...
+      cy.task(NODE_TASKS.INSERT_UTILISATION_REPORTS_INTO_DB, [report]);
+
+      const feeRecords = [feeRecordOne, feeRecordTwo, feeRecordThree, feeRecordFour];
+      cy.task(NODE_TASKS.INSERT_FEE_RECORDS_INTO_DB, feeRecords);
+
+      const matchingTfmFacilities = getMatchingTfmFacilitiesForFeeRecords(feeRecords);
+      cy.task(NODE_TASKS.INSERT_TFM_FACILITIES_INTO_DB, matchingTfmFacilities);
+    });
+
+    describe('by clicking the back button', () => {
```
ttbarnes, line 229: Can we add "with XYZ checkboxes checked" please?

ttbarnes (changes requested):
1x minor comment otherwise LGTM.

ttbarnes (approved):

e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/add-payment.spec.js
sw-mattw: Just pushed a change for this thanks, please let me know if you're not sure on how it's named (I'm not 100% sure on it but can't think of a better alternative)

RLavender98 (approved):

PR #3693, opened 2024-10-16
feat(dtfs2-6892): fix docker config and update routers to avoid issue with unbound methods

AlexBramhill:
## Introduction :pencil2:
update routers to correctly bind `this`
update docker to include env vars

### DIFF
--- a/trade-finance-manager-ui/server/routes/login/configs/login-sso.ts
+++ b/trade-finance-manager-ui/server/routes/login/configs/login-sso.ts
@@ -10,10 +10,9 @@ export const getLoginSsoRouter: GetRouter = () => {
   const entraIdApi = new EntraIdApi({ entraIdConfig });
   const entraIdService = new EntraIdService({ entraIdConfig, entraIdApi });
   const loginController = new LoginController({ entraIdService });
-
   const loginSsoRouter = express.Router();
-  // eslint-disable-next-line @typescript-eslint/no-misused-promises, @typescript-eslint/unbound-method
-  loginSsoRouter.get('/', loginController.getLogin);
+  // eslint-disable-next-line @typescript-eslint/no-misused-promises
+  loginSsoRouter.get('/', (req, res) => loginController.getLogin(req, res));
 
   return loginSsoRouter;
 };
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -125,6 +125,11 @@ services:
       - FF_TFM_FACILITY_END_DATE_ENABLED
       - FF_TFM_DEAL_CANCELLATION_ENABLED
       - FF_TFM_SSO_ENABLED
+      - ENTRA_ID_CLIENT_ID
+      - ENTRA_ID_CLOUD_INSTANCE
+      - ENTRA_ID_TENANT_ID
+      - ENTRA_ID_CLIENT_SECRET
+      - ENTRA_ID_REDIRECT_URL
 
   trade-finance-manager-api:
     build:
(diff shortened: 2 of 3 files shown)


trade-finance-manager-ui/server/routes/login/configs/login-sso.ts
```diff
@@ -10,10 +10,9 @@ export const getLoginSsoRouter: GetRouter = () => {
   const entraIdApi = new EntraIdApi({ entraIdConfig });
   const entraIdService = new EntraIdService({ entraIdConfig, entraIdApi });
   const loginController = new LoginController({ entraIdService });
-
   const loginSsoRouter = express.Router();
-  // eslint-disable-next-line @typescript-eslint/no-misused-promises, @typescript-eslint/unbound-method
-  loginSsoRouter.get('/', loginController.getLogin);
+  // eslint-disable-next-line @typescript-eslint/no-misused-promises
```
ttbarnes, line 14: Nice to remove the `unbound-method` disabling, but not 100% happy replacing it with `no-misused-promises`

Seems so strange, would like to understand where it orignates from so we can fix the issue

Have this throughout the codebase so not a blocker, some day it'd be good to fix in one go

ttbarnes (approved):

trade-finance-manager-ui/server/routes/login/configs/login-sso.ts
abhi-markan: Agree nice to remove unbound-method, can we please look into removing inline Eslint disable rules they are piling up and adding to tech-debt.

abhi-markan (changes requested):

trade-finance-manager-ui/server/routes/login/configs/login-sso.ts
AlexBramhill: The issue is that we are returning a `promise<void>` and not just `void`.

We can either add a new dependancy to address this (https://www.npmjs.com/package/express-async-errors) and update some eslint rules to allow this, or we could have an Immediately Invoked Async Function which is rather ugly

example:
```
  loginSsoRouter.get('/', (req, res) => {
    (async () => {
      await loginController.getLogin(req, res);
    })();
  });
```

Any preference please?

ttbarnes: Express 5 seems the most sensible, but looks like it could be a huge job.

I don't think we should use the `express-async-errors` package as it hasn't been updated in a few years and could become redundant quickly, especially if we migrate to express 5.

I think we'd all like to not have any eslint disabling in this route - and any other places, but IMO this should be tackled on its own. This PR also seems to blocking SSO which has larger benefits to the business.

ttbarnes: I would be interested to know the differences with this between DTFS and EXIPS. EXIPS calls routes very simply and no eslint disabling is required, e.g:

```ts
insuranceAccountRouter.get('/a-route', yourDetailsGet);
```

We use custom types for express, maybe that's what it is.

ttbarnes: Interesting thanks @AlexBramhill - I thought we aligned the eslint config with DTFS, but I maybe wrong.

I'm happy for this PR to be merged as per my comments above 

@abhi-markan

abhi-markan: Thanks both, I'll raise a PR to align DTFS and EXIP Eslint rules.
I'll approve this for now as I don't want this to block SSO any further and we can look into removing inline disabled rules.

abhi-markan (approved):

PR #3695, opened 2024-10-17
feat(FN-3504): apply cover percentage to initial utilisation

RLavender98:
## Introduction :pencil2:
We're not saving the correct value for the initial utilisation because we are returning the value without the cover percentage applied

- Use calculate drawn amount function to calculate initial utilisation which applies cover percentage

### DIFF
--- a/dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/helpers/calculate-initial-utilisation-and-fixed-fee.test.ts
+++ b/dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/helpers/calculate-initial-utilisation-and-fixed-fee.test.ts
@@ -1,9 +1,15 @@
-import { calculateInitialUtilisation } from '@ukef/dtfs2-common';
+import * as dtfsCommon from '@ukef/dtfs2-common';
 import { calculateInitialUtilisationAndFixedFee, parseDate, hasRequiredValues, RequiredParams } from './calculate-initial-utilisation-and-fixed-fee';
 import { TfmFacilitiesRepo } from '../../../../../repositories/tfm-facilities-repo';
 import { aTfmFacility } from '../../../../../../test-helpers';
-import { calculateInitialFixedFee } from './calculate-initial-fixed-fee';
-import { calculateUkefShareOfUtilisation } from '../../../../../helpers';
+import * as fixedFeeHelpers from './calculate-initial-fixed-fee';
+
+jest.mock('./calculate-initial-fixed-fee');
+// eslint-disable-next-line @typescript-eslint/no-unsafe-return
+jest.mock('@ukef/dtfs2-common', () => ({
+  ...jest.requireActual('@ukef/dtfs2-common'),
+  calculateDrawnAmount: jest.fn(),
+}));
 
 describe('helpers/calculate-initial-utilisation-and-fixed-fee', () => {
   describe('parseDate', () => {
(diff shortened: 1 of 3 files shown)


dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/helpers/calculate-initial-utilisation-and-fixed-fee.ts
```diff
@@ -70,12 +70,10 @@ export const calculateInitialUtilisationAndFixedFee = async (facilityId: string)
     throw new Error(`TFM facility values for ${facilityId} are missing`);
   }
 
-  const utilisation = calculateInitialUtilisation(value);
-
-  const ukefShareOfUtilisation = calculateUkefShareOfUtilisation(utilisation, coverPercentage);
+  const ukefShareOfInitialUtilisation = new Big(calculateDrawnAmount(value, coverPercentage)).round(2).toNumber();
```
RLavender98, line 73: `calculateDrawnAmount` doesn't use Big or any rounding but we need to match what is in ACBS so we have to use the same function and then do the rounding ourselves after here

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/helpers/calculate-initial-utilisation-and-fixed-fee.test.ts
```diff
@@ -1,9 +1,15 @@
...
+import * as dtfsCommon from '@ukef/dtfs2-common';
 import { calculateInitialUtilisationAndFixedFee, parseDate, hasRequiredValues, RequiredParams } from './calculate-initial-utilisation-and-fixed-fee';
 import { TfmFacilitiesRepo } from '../../../../../repositories/tfm-facilities-repo';
 import { aTfmFacility } from '../../../../../../test-helpers';
-import { calculateInitialFixedFee } from './calculate-initial-fixed-fee';
-import { calculateUkefShareOfUtilisation } from '../../../../../helpers';
+import * as fixedFeeHelpers from './calculate-initial-fixed-fee';
+
+jest.mock('./calculate-initial-fixed-fee');
+// eslint-disable-next-line @typescript-eslint/no-unsafe-return
```
ChristianMcc, line 8: Is this something that we just think will happen every time we want to mock something?
@abhi-markan if this is an expected piece of code for mocking in tests would you prefer we disable this rule across all test.ts files or leave this disable in?

ChristianMcc (commented):
Thank you for picking this up so quick! One comment for Abhi to take a quick look at re eslint

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/helpers/calculate-initial-utilisation-and-fixed-fee.test.ts
```diff
@@ -1,9 +1,15 @@
...
+jest.mock('@ukef/dtfs2-common', () => ({
```
Zainzzkk, line 9: if we mock the file only, then we do not need this eslint disable.  And do we need to mock it? it's doing a simple calculation and we could let the function calculate it for us and test the end result

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/helpers/calculate-initial-utilisation-and-fixed-fee.test.ts
RLavender98: I copied this from another file but have since seen some other files are doing this:
```
jest.mock('@ukef/dtfs2-common', () => ({
  ...jest.requireActual<object>('@ukef/dtfs2-common'),
  calculateDrawnAmount: jest.fn(),
}));
```
I don't really know enough about this eslint rule to know if that is preferable or not

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/helpers/calculate-initial-utilisation-and-fixed-fee.ts
RLavender98: Roughly, we want to change to this because using the same function as what we send to ACBS means we will always be in sync rather than relying on the methods lining up. 

The actual real change and fix is actually further down in what we return.

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/helpers/calculate-initial-utilisation-and-fixed-fee.test.ts
ChristianMcc: Ah got you, okay i'm going to resolve and approve. Let's look at this with our tech debt time next week :)

ChristianMcc (approved):

PR #3697, opened 2024-10-17
feat(DTFS2-7553): add status to deal cancellation

MarRobSoftwire:
## Introduction :pencil2:
The deal cancellation objects need a status to be marked as draft, scheduled or completed

- Add new `TfmDealCancellationWithStatus` type and rename `TfmDealCancellation` to`TfmDealCancellationWithoutStatus`
- Add `status: 'DRAFT'` when updating cancellation object
- Add constants for `SCHEDULED` and `COMPLETED` for use in future tickets


libs/common/src/types/index.ts
```diff
@@ -23,7 +23,7 @@ export * from './api-error-code';
 export * from './facility-type';
 export * from './deal-type';
 export * from './any-object';
-export * from './tfm-deal-cancellation';
+export * from './deal-cancellation';
```
ttbarnes, line 26: What's the reason for the rename please, deal cancellation is TFM specific?

ttbarnes, line 26: What's the reason for the rename please, deal cancellation is TFM specific?

libs/common/src/types/mongo-db-models/tfm-deal-cancellation.ts
```diff
@@ -0,0 +1,7 @@
+import { TFM_DEAL_CANCELLATION_STATUS } from '../../constants';
+import { TfmDealCancellationWithoutStatus } from '../deal-cancellation';
+import { ValuesOf } from '../types-helper';
+
+export type TfmDealCancellationWithStatus = TfmDealCancellationWithoutStatus & {
```
ttbarnes, line 5: Could `TfmDealCancellation` just have an optional `status` instead of `TfmDealCancellationWithoutStatus` and `TfmDealCancellationWithStatus`?

Or perhaps it should initiate with a default `DRAFT` status and not optional?

ttbarnes (changes requested):

libs/common/src/types/mongo-db-models/tfm-deal-cancellation.ts
MarRobSoftwire: The `TfmDealCancellationWithoutStatus` is what the user submits and is the payload on the PUT and POST request.
The `TfmDealCancellationWithStatus` is what's stored in the database & returned on GET requests.
These are two distinct types so I dont want them to the same. I'll edit the names though

libs/common/src/types/index.ts
MarRobSoftwire: Oops, I did this when changing the name of the type and forgot to put it back

libs/common/src/types/mongo-db-models/tfm-deal-cancellation.ts
ttbarnes: Thanks @MarRobSoftwire , makes sense - thanks for the rename, much better than having `TfmDealCancellationWithoutStatus` and `TfmDealCancellationWithStatus` 😄

ttbarnes (approved):

PR #3698, opened 2024-10-18
fix(DTFS2-7440): acbs docker file reference

abhi-markan:
## Introduction :pencil2:
Deployment script is unable to reference the Dockerfile from correct context.

* Working directory changed to root.
* docker build will be executed with `-f` explicit dockerfile argument.

### DIFF
--- a/.github/workflows/deployment.yml
+++ b/.github/workflows/deployment.yml
@@ -279,10 +279,10 @@ jobs:
           password: ${{ secrets.ACR_PASSWORD }}
 
       - name: Artifacts 🗃️
-        working-directory: ./azure-functions/${{ matrix.functions }}
+        working-directory: .
         run: |
           # Build images
-          docker build . \
+          docker build -f ./azure-functions/${{ matrix.functions }}/Dockerfile . \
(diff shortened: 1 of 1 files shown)


Zainzzkk (approved):

.github/workflows/deployment.yml
```diff
@@ -279,10 +279,10 @@ jobs:
           password: ${{ secrets.ACR_PASSWORD }}
 
       - name: Artifacts 🗃️
-        working-directory: ./azure-functions/${{ matrix.functions }}
+        working-directory: .
```
ChristianMcc, line 282: does this mean we don't even need this is we're just setting it to '.'? Unsure if it's resetting something, either way LGTM I'll approve!

ChristianMcc (approved):

.github/workflows/deployment.yml
abhi-markan: Do you mean we don't need `working-directory` argument?

ChristianMcc: yeah

PR #3752, opened 2024-10-21
feat(DTFS2-7555): map cancellation activity in tfm-ui

MarRobSoftwire:
## Introduction :pencil2:
Tfm-ui renders activities all the same, this wont work if we want to display a [govuk-tag component](https://design-system.service.gov.uk/components/tag/) on the cancelled activity.

Move `mapActivities` method into its own file & add mapping for cancellation type.

### BSS/EWCS
image

### GEF
image

### Empty Reason
image

- Add `TfmActivity` type and associated `ACTIVITY_TYPES` constants to differentiate cancellations from other activity
- Remove `'OTHER'` activity type as I cannot see any prod code using it


e2e-tests/tfm/cypress/e2e/journeys/case/activities/activities.spec.js
```diff
@@ -28,7 +29,7 @@ context('Users can create and submit comments', () => {
       const otherActivity = {
         tfm: {
           activities: {
-            type: 'OTHER',
+            type: ACTIVITY_TYPES.OTHER,
```
MarRobSoftwire, line 32: I can't tell if `"Other"` is a real activity type - it only seems to exist in tests (but the same is true of `"ACTIVITY"` and I've found that in the dev database!)

trade-finance-manager-ui/server/controllers/case/activity/helpers/map-activities.ts
```diff
@@ -0,0 +1,50 @@
...
+    };
+
+    switch (activity.type) {
+      case ACTIVITY_TYPES.CANCELLATION:
+        return {
+          label: {
+            text: 'Deal cancelled',
+          },
+          html: `
+          <p> Deal stage:
```
MarRobSoftwire, line 25: I don't really like having the html generated here but Im not sure how else to effectively do this

trade-finance-manager-ui/server/controllers/case/activity/helpers/map-activities.ts
```diff
@@ -0,0 +1,50 @@
...
+            <strong class="govuk-tag govuk-tag--red">
+              Cancelled
+            </strong>
+            <br/><br/>
+            Bank request date: ${format(activity.bankRequestDate, 'd MMMM yyyy')}
+            <br/><br/>
+            Date effective from: ${format(activity.effectiveFrom, 'd MMMM yyyy')}
+            <br/><br/>
+            Comments: ${activity.reason || '-'}
```
MarRobSoftwire, line 34: This is a security issue!!!

libs/common/src/types/mongo-db-models/tfm-activity.ts
```diff
@@ -0,0 +1,27 @@
+import { ACTIVITY_TYPES } from '../../constants';
+import { UnixTimestamp, UnixTimestampSeconds } from '../date';
+
+export type TfmActivity =
+  | {
+      type: typeof ACTIVITY_TYPES.ACTIVITY | typeof ACTIVITY_TYPES.COMMENT;
```
ttbarnes, line 6: Can we destructure `ACTIVITY_TYPES` in here, I think it makes sense to _not_ do this in the previousf files, but in here it's a tongue twister 😄 

```ts
type: typeof ACTIVITY_TYPES.ACTIVITY | typeof
```

trade-finance-manager-api/src/v1/rest-mappings/filters/filterActivities.test.js
```diff
@@ -8,28 +9,28 @@ const MOCK_AUTHOR = {
...
-    type: 'COMMENT',
+    type: ACTIVITY_TYPES.COMMENT,
     timestamp: 13345665,
     text: 'test2',
     author: MOCK_AUTHOR,
     label: 'Comment added',
   },
   {
-    type: 'OTHER',
+    type: ACTIVITY_TYPES.ACTIVITY,
```
ttbarnes, line 33: Same as above

What do you think about this being `ACTIVITY_TYPES.GENERIC_ACTIVITY` instead of `ACTIVITY_TYPES.ACTIVITY`, I think it'd read easier 

Maybe could then destructure these constants as well

trade-finance-manager-ui/server/controllers/case/activity/helpers/map-activities.test.ts
```diff
@@ -0,0 +1,109 @@
...
+        byline: {
+          text: `${comment.author.firstName} ${comment.author.lastName}`,
+        },
+      },
+    ]);
+  });
+
+  it('returns empty array if activities is undefined', () => {
+    // Act
+    const result = mapActivities(undefined);
```
ttbarnes, line 104: `mapActivities();` please

trade-finance-manager-ui/server/controllers/case/activity/helpers/map-activities.ts
```diff
@@ -0,0 +1,43 @@
+import { ACTIVITY_TYPES, TfmActivity } from '@ukef/dtfs2-common';
+import { fromUnixTime } from 'date-fns';
+
+export const mapActivities = (activities: TfmActivity[] | undefined) => {
```
ttbarnes, line 4: Can we make the activities param optional instead of having to specificallty declare undefined?

ttbarnes, line 4: Documentation? Not clear why this mapping exists and without going into the mapping and switch/case, what it does

ttbarnes (changes requested):
Nice work, some minor comments

libs/common/src/types/mongo-db-models/tfm-activity.ts
```diff
@@ -0,0 +1,27 @@
+import { ACTIVITY_TYPES } from '../../constants';
+import { UnixTimestamp, UnixTimestampSeconds } from '../date';
+
+export type TfmActivity =
```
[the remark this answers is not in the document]
MarRobSoftwire: This is a part of the TfmDeals model (stored in the database) so it could be used everywhere - in my next PR I'll use this type to update the database with a cancellation

trade-finance-manager-api/src/v1/rest-mappings/filters/filterActivities.test.js
```diff
@@ -8,28 +9,28 @@ const MOCK_AUTHOR = {
...
-    type: 'COMMENT',
+    type: ACTIVITY_TYPES.COMMENT,
     timestamp: 13345665,
     text: 'test1',
     author: MOCK_AUTHOR,
     label: 'Comment added',
   },
   {
-    type: 'OTHER',
+    type: ACTIVITY_TYPES.ACTIVITY,
```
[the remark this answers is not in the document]
MarRobSoftwire: "OTHER" was added in tests only & isnt anywhere in production code - I've replaced all these instances with a real value for the type

trade-finance-manager-api/src/v1/rest-mappings/filters/filterActivities.test.js
MarRobSoftwire: I want to keep the ACTIVITY_TYPES clear so that `ACTIVITY_TYPES.ACTIVITY = 'ACTIVITY'` and similar
I think this is a lot clearer than `` ACTIVITY_TYPES.GENERIC_ACTIVITY  = 'ACTIVITY'`. 'ACTIVITY' but is legacy and would need a migration to change this enum

trade-finance-manager-ui/server/controllers/case/activity/helpers/map-activities.test.ts
MarRobSoftwire: Typescript views `mapActivities()` and `mapActivities(undefined)` as two different signatures (even though it will behave the same) - I've updated the mapActivites signature to allow this.

ttbarnes: Feels strange, I don't see a need to specifically pass in `undefined`

Not sure why it's not picking up here, but in another PR, this was flagged in SonarCloud as unnecessary

MarRobSoftwire: I've updated this now anyway :)

ttbarnes (approved):

PR #3754, opened 2024-10-21
feat(FN-3519): support effective date in unix time in seconds

RLavender98:
## Introduction :pencil2:
Effective date is stored in unix epoch time in seconds rather than milliseconds so `new Date(effectiveDate)` was always returning a date in 1970 instead of the correct date, so all amendments were counting as effective.

- Use `convertTimestampToDate` helper to read effective dates to ensure we support the date whether stored as seconds or milliseconds.

### DIFF
--- a/dtfs-central-api/api-tests/v1/utilisation-reports/post-keying-data.api-test.ts
+++ b/dtfs-central-api/api-tests/v1/utilisation-reports/post-keying-data.api-test.ts
@@ -401,6 +401,59 @@ describe(`POST ${BASE_URL}`, () => {
     expect(entities[0].fixedFee).toEqual(525.6);
   });
 
+  it('calculates the ukef share of utilisation and saves to facility utilisation table', async () => {
+    // Arrange
+    const facilityId = '11111111';
+
+    const tfmFacility: TfmFacility = {
+      ...aTfmFacility(),
+      facilitySnapshot: {
+        ...aFacility(),
+        ukefFacilityId: facilityId,
+        coverPercentage: 80,
+      },
+    };
+
+    const tfmFacilitiesCollection = await mongoDbClient.getCollection('tfm-facilities');
+    await tfmFacilitiesCollection.insertOne(tfmFacility);
+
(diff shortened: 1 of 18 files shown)


ChristianMcc (dismissed):
LGTM! :)

dtfs-central-api/api-tests/v1/utilisation-reports/post-keying-data.api-test.ts
```diff
@@ -342,15 +342,17 @@ describe(`POST ${BASE_URL}`, () => {
           ...aTfmFacilityAmendment(),
           value: 350000,
           status: AMENDMENT_STATUS.COMPLETED,
-          effectiveDate: dateAfterReportPeriodEnd.getTime(),
+          // Effective dates are stored in unix epoch time in seconds not milliseconds.
+          effectiveDate: Math.round(dateAfterReportPeriodEnd.getTime() / 1000),
```
ttbarnes, line 346: Could we create a helper for `Math.round( ... / 1000)`? Repeated 11x times. Maybe in "common"?

ttbarnes (changes requested):

libs/common/src/test-helpers/convert-milliseconds-to-seconds.test.ts
```diff
@@ -0,0 +1,27 @@
...
+
+    it('divides the passed in number by 1000 and rounds the result - rounding up', () => {
+      // Arrange
+      const milliseconds = 1729521500;
+
+      // Act
+      const seconds = convertMillisecondsToSeconds(milliseconds);
+
+      // Assert
+      expect(seconds).toEqual(1729522);
```
ttbarnes, line 24: Feels like we're kind of testing the native `Math.round` method here, I think just 1x assertion with `Math.round(milliseconds / 1000);` would suffice. But there's lots of conflicting opinions/preferences with unit tests so happy to approve.

ttbarnes (approved):

PR #3756, opened 2024-10-21
feat(DTFS2-7298): hide cancel deal button when deal cancelled

BethThomas141:
## Introduction :pencil2:
- add `canDealStillBeCancelled` to check that the cancellation stage isn't COMPLETED or SCHEDULED (ie. check it hasn't already be cancelled. 
- Impact of this is that the 'Cancel deal' button will be hidden when the deal is already cancelled


libs/common/src/constants/index.ts
```diff
@@ -24,6 +24,5 @@ export * from './url-character-allowances';
 export * from './max-character-count';
 export * from './deal-submission-type';
 export * from './floating-point-rounding-decimal-places';
-export * from './cancellation-status';
```
BethThomas141, line 27: I moved the deal cancellation status constants inside the `tfm` folder which is why these changes are showing up

trade-finance-manager-ui/server/controllers/helpers/can-deal-be-cancelled.helper.ts
```diff
@@ -16,14 +23,19 @@ export const canSubmissionTypeBeCancelled = (submissionType: DealSubmissionType)
  * Checks if deal cancellation is enabled for a deal and user type
  * @param submissionType - the deal submission type
  * @param user - the session user
+ * @param cancellationStatus - the current deal cancellation status if it exists
  * @returns true or false depending on the feature flag, submission type and user type
  */
-export const isDealCancellationEnabled = (submissionType: DealSubmissionType, user: TfmSessionUser): boolean => {
-  // TODO: DTFS2-7298: also check that the deal hasn't already been cancelled.
+export const canDealBeCancelled = (submissionType: DealSubmissionType, user: TfmSessionUser, cancellationStatus?: TfmDealCancellationStatus): boolean => {
```
BethThomas141, line 29: I've renamed this as I think the new name is a tad clearer (makes it clear it's not just based on the feature flag being enabled)

trade-finance-manager-ui/server/controllers/case/index.js
```diff
@@ -45,13 +44,11 @@ const getCaseDeal = async (req, res) => {
...
-  let hasDraftCancellation = false;
+  const dealCancellation = await api.getDealCancellation(dealId, userToken);
 
-  if (dealCancellationIsEnabled) {
-    const cancellation = await api.getDealCancellation(dealId, userToken);
-    hasDraftCancellation = !isEmpty(cancellation);
-  }
+  const dealCancellationPermitted = canDealBeCancelled(submissionType, user, dealCancellation.status);
+
+  const hasDraftCancellation = dealCancellationPermitted && dealCancellation?.status === TFM_DEAL_CANCELLATION_STATUS.DRAFT;
```
ttbarnes, line 51: Maybe could move into a helper directory, like `canDealBeCancelled`?

trade-finance-manager-ui/server/controllers/case/index.test.js
```diff
@@ -99,12 +99,12 @@ describe('controllers - case', () => {
       it('should check whether deal cancellation is enabled', async () => {
         await caseController.getCaseDeal(req, res);
 
-        expect(isDealCancellationEnabled).toHaveBeenCalledTimes(1);
+        expect(canDealBeCancelled).toHaveBeenCalledTimes(1);
```
ttbarnes, line 102: Should we add `calledWith()` assertions?

trade-finance-manager-ui/server/controllers/helpers/can-deal-be-cancelled.helper.ff-test.ts
```diff
@@ -0,0 +1,93 @@
...
+import { TfmSessionUser } from '../../types/tfm-session-user';
+
+const pimUser = { teams: [TEAM_IDS.PIM] } as TfmSessionUser;
+const nonPimUser = { teams: [TEAM_IDS.UNDERWRITERS] } as TfmSessionUser;
+
+const { AIN, MIN, MIA } = DEAL_SUBMISSION_TYPE;
+const { COMPLETED, SCHEDULED, DRAFT } = TFM_DEAL_CANCELLATION_STATUS;
+
+describe('canDealBeCancelled', () => {
+  describe('when `FF_TFM_FACILITY_END_DATE_ENABLED` is set to true', () => {
```
ttbarnes, line 12: We have a lot of inconsistent test descriptions for feature flag tests:

```ts
describe('when deal cancellation is enabled', () => {
describe('when deal cancellation is disabled', () => {
  it('does not call next when deal cancellation is enabled', () => {

describe('when FF_TFM_DEAL_CANCELLATION_ENABLED is disabled', () => {
describe('when FF_TFM_DEAL_CANCELLATION_ENABLED is enabled', () => {

it("redirects to '/not-found' if the feature flag is not enabled", () => {
it('calls the next function if the feature flag is enabled', () => {
```

Not sure what to suggest here, we should align these in a different PR

trade-finance-manager-ui/server/controllers/helpers/can-deal-be-cancelled.helper.ts
```diff
@@ -16,14 +23,19 @@ export const canSubmissionTypeBeCancelled = (submissionType: DealSubmissionType)
  * Checks if deal cancellation is enabled for a deal and user type
  * @param submissionType - the deal submission type
  * @param user - the session user
+ * @param cancellationStatus - the current deal cancellation status if it exists
```
ttbarnes, line 26:
```suggestion
 * @param cancellationStatus - optional current deal cancellation status
```

trade-finance-manager-ui/server/controllers/helpers/can-deal-be-cancelled.helper.ts
```diff
@@ -16,14 +23,19 @@ export const canSubmissionTypeBeCancelled = (submissionType: DealSubmissionType)
...
+export const canDealBeCancelled = (submissionType: DealSubmissionType, user: TfmSessionUser, cancellationStatus?: TfmDealCancellationStatus): boolean => {
+  const isDealCancellationFeatureFlagEnabled = isTfmDealCancellationFeatureFlagEnabled();
 
   const isUserAllowedToCancelDeal = userIsInTeam(user, [TEAM_IDS.PIM]);
-  const isDealCancellationFeatureFlagEnabled = isTfmDealCancellationFeatureFlagEnabled();
+
   const isAcceptableSubmissionType = canSubmissionTypeBeCancelled(submissionType);
 
-  return isUserAllowedToCancelDeal && isDealCancellationFeatureFlagEnabled && isAcceptableSubmissionType;
+  const isDealNotAlreadyCancelled = !(
```
ttbarnes, line 36: Finding `isDealNotAlreadyCancelled` a little confusing especially in the final return, how about something like this?

```ts
const dealIsAlreadyCancelled = Object.values(TFM_DEAL_CANCELLATION_STATUS).includes(cancellationStatus);
```

Actually `Object.values` probably won't work here, maybe just remove the `!` and rename the existing const?

Then the return can do `!dealIsAlreadyCancelled`

ttbarnes (changes requested):
Nice work - few minor comments

Can we add an E2E test?

libs/common/src/constants/index.ts
MarRobSoftwire: Good spot - thanks! Tbf the name should probably be tfm-cancellation-status too oops

libs/common/src/types/tfm/cancellation-status.ts
```diff
@@ -0,0 +1,4 @@
+import { TFM_DEAL_CANCELLATION_STATUS } from '../../constants';
+import { ValuesOf } from '..';
+
+export type TfmDealCancellationStatus = ValuesOf<typeof TFM_DEAL_CANCELLATION_STATUS>;
```
MarRobSoftwire, line 4: I probably should have done this too oops! Could you use this type in `TfmDealCancellationWithStatus` please :)

trade-finance-manager-ui/server/controllers/case/index.js
```diff
@@ -45,13 +44,11 @@ const getCaseDeal = async (req, res) => {
...
   const { submissionType } = deal.dealSnapshot;
 
-  const dealCancellationIsEnabled = isDealCancellationEnabled(submissionType, user);
-  let hasDraftCancellation = false;
+  const dealCancellation = await api.getDealCancellation(dealId, userToken);
 
-  if (dealCancellationIsEnabled) {
-    const cancellation = await api.getDealCancellation(dealId, userToken);
-    hasDraftCancellation = !isEmpty(cancellation);
-  }
```
MarRobSoftwire, line 54: Checking the status rather than using `isEmpty` is soooo much nicer - maybe too much for this PR but there's a few other places I added this check that could now look for draft (and that would prevent the pages being accessed once the cancellations submitted)

trade-finance-manager-ui/server/controllers/helpers/can-deal-be-cancelled.helper.ts
MarRobSoftwire: Agree `dealIsAlreadyCancelled` is much easier to read :) And since we're negating here anyway it's probably simpler to see whats going on anyway

trade-finance-manager-ui/server/controllers/helpers/can-deal-be-cancelled.helper.test.ts
```diff
@@ -0,0 +1,79 @@
...
+const nonPimUser = { teams: [TEAM_IDS.UNDERWRITERS] } as TfmSessionUser;
+
+const { AIN, MIN, MIA } = DEAL_SUBMISSION_TYPE;
+const { COMPLETED, SCHEDULED, DRAFT } = TFM_DEAL_CANCELLATION_STATUS;
+
+describe('canDealBeCancelled', () => {
+  describe('when `FF_TFM_FACILITY_END_DATE_ENABLED` is set to false', () => {
+    describe('when user is a PIM user', () => {
+      it.each([MIA, MIN, AIN])('should return false if the deal type is %s and the cancellation status is completed', (type) => {
+        const result = canDealBeCancelled(type, pimUser, COMPLETED);
```
MarRobSoftwire, line 15: could these `it.each`'s be `describe.each`?

```suggestion
      describe.each([MIA, MIN, AIN])('when the deal type is %s', (type) => {
        it('should return false if the cancellation status is completed', () => {
          const result = canDealBeCancelled(type, pimUser, COMPLETED);

          expect(result).toEqual(false);
        });

        it('should return false if the cancellation status is scheduled', (type) => {
          const result = canDealBeCancelled(type, pimUser, SCHEDULED);

          expect(result).toEqual(false);
        });

        // etc...
      });
```

trade-finance-manager-ui/server/controllers/case/index.js
BethThomas141: This logic will only be used here, so keen to keep it as is if that's ok!

trade-finance-manager-ui/server/controllers/helpers/can-deal-be-cancelled.helper.test.ts
MarRobSoftwire: actually is this just copied over & git being weird? feel free to ignore

trade-finance-manager-ui/server/controllers/case/index.test.js
BethThomas141: Have added, thank you!

trade-finance-manager-ui/server/controllers/helpers/can-deal-be-cancelled.helper.ff-test.ts
BethThomas141: yes good spot - seems like there's no clear precedent here so maybe one for another PR

trade-finance-manager-ui/server/controllers/helpers/can-deal-be-cancelled.helper.ts
BethThomas141: changed, thanks!

trade-finance-manager-ui/server/controllers/helpers/can-deal-be-cancelled.helper.ts
BethThomas141: Have changed! Thanks both

libs/common/src/constants/index.ts
BethThomas141: Have updated the name!

trade-finance-manager-ui/server/controllers/helpers/can-deal-be-cancelled.helper.test.ts
BethThomas141: Good spot, that is a valid and better change - have updated the test files. Thanks!

libs/common/src/types/tfm/cancellation-status.ts
BethThomas141: Yes good idea! Have changed, thanks

trade-finance-manager-ui/server/controllers/case/index.js
BethThomas141: Yes I noticed that! Let's do that as a separate PR

trade-finance-manager-ui/server/controllers/case/index.test.js
```diff
@@ -147,7 +149,7 @@ describe('controllers - case', () => {
 
       describe('when deal cancellation is disabled', () => {
         it('should render the template with showDealCancelButton=false', async () => {
-          jest.mocked(isDealCancellationEnabled).mockReturnValueOnce(false);
+          jest.mocked(canDealBeCancelled).mockReturnValueOnce(false);
```
ttbarnes, line 152: Can we update the test descriptions please @BethThomas141 , they're currently:

- when deal cancellation is enabled
- when deal cancellation is disabled

(etc)

But the condition is now "is deal cancellation permitted"

ttbarnes, line 152: Also, unless i'm missing it, could we add a test for "deal cancellation is permitted, but the deal cancellation status is _not_ `TFM_DEAL_CANCELLATION_STATUS.DRAFT` - for the `hasDraftCancellation` page/nunjucks variable

The alternative would be to extract the `hasDraftCancellation` const into it's own helper function and then that could just be called in the test once

BethThomas141: Great spot, these tests needed refactoring after updating the logic in the index file (explained why I had a failing test😅). Have now rewritten to reflect the draft cancellation status check

BethThomas141: Great spot, have changed! Thanks!

ttbarnes (dismissed):

ttbarnes (approved):

ttbarnes (approved):

trade-finance-manager-ui/server/controllers/helpers/deal-cancellation-enabled.helper.ts
```diff
@@ -19,11 +26,29 @@ export const canSubmissionTypeBeCancelled = (submissionType: DealSubmissionType)
...
 
   return isUserAllowedToCancelDeal && isDealCancellationFeatureFlagEnabled && isAcceptableSubmissionType;
 };
+
+/**
+ * Checks if a deal can be cancelled; if it has already been cancelled then this returns false
+ * @param cancellationStatus - optional current deal cancellation status
+ * @returns true or false depending on if the deal has already been cancelled
+ */
+export const canDealStillBeCancelled = (cancellationStatus?: TfmDealCancellationStatus): boolean => {
```
ttbarnes, line 41: What's the reason for `canDealStillBeCancelled` please, how about `canDealBeCancelled`?

trade-finance-manager-ui/server/controllers/helpers/deal-cancellation-enabled.helper.ts
```diff
@@ -19,11 +26,29 @@ export const canSubmissionTypeBeCancelled = (submissionType: DealSubmissionType)
...
+  const isDealAlreadyCancelled = cancellationStatus === TFM_DEAL_CANCELLATION_STATUS.COMPLETED || cancellationStatus === TFM_DEAL_CANCELLATION_STATUS.SCHEDULED;
+
+  return !isDealAlreadyCancelled;
```
ttbarnes, line 44: What do you think about this instead? More lines, but I think it's clearer

```ts
const isCompleted = cancellationStatus === TFM_DEAL_CANCELLATION_STATUS.COMPLETED;
const isScheduled = cancellationStatus === TFM_DEAL_CANCELLATION_STATUS.SCHEDULED;

const canBeCancelled = !isCompleted && !isScheduled;

return canBeCancelled;
```

trade-finance-manager-ui/server/controllers/helpers/deal-cancellation-enabled.helper.ts
BethThomas141: Fair point, have changed! Thank you

ttbarnes (approved):

PR #3762, opened 2024-10-23
feat(FN-3507): display selected filters on payment details tab

RLavender98:
## Introduction :pencil2:
We want users to be able to see selected filters and remove them individually, when the filter panel is open and closed.

- Populate selected filters in payment details tab filter pane
- Populate action bar with selected filters in payment details tab

Screenshot 2024-10-23 at 15 18 56

Screenshot 2024-10-23 at 15 19 24


e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/filter-payment-details/filter-payment-details.spec.js
```diff
@@ -166,5 +184,124 @@ context(`users can filter payment details by facility id and payment reference a
...
+      cy.assertText(filters.selectedFilter(paymentCurrencyFilter), `Remove this filter ${paymentCurrencyFilter}`);
+      cy.assertText(filters.selectedFilter(paymentReferenceFilter), `Remove this filter ${paymentReferenceFilter}`);
+      cy.assertText(filters.selectedFilter(facilityIdFilter), `Remove this filter ${facilityIdFilter}`);
+
+      cy.assertText(filters.actionBarItem(facilityIdFilter), `Remove this filter ${facilityIdFilter}`);
+      cy.assertText(filters.actionBarItem(paymentCurrencyFilter), `Remove this filter ${paymentCurrencyFilter}`);
+      cy.assertText(filters.actionBarItem(paymentReferenceFilter), `Remove this filter ${paymentReferenceFilter}`);
+    });
+
+    it('should remove all filters when clear filters button clicked', () => {
```
ttbarnes, line 230: There's a lot going on in each "it", could we split into "its" and "describess" so it's easier to read/update in the future? E.g something like this

```js
describe('when clear filters button clicked', () => {
  before(() => {
    // login, click XYZ etc
  });

  it('should remove all filters', () => {
    // assertions
  });
});
```

e2e-tests/tfm/cypress/e2e/pages/utilisation-reports/utilisationReportPage.js
```diff
@@ -63,6 +63,10 @@ const utilisationReportPage = {
       paymentCurrencyRadioInput: (currencyCode) => cy.get(`input[type="radio"][data-cy="currency-${currencyCode}"]`),
       paymentCurrencyError: () => cy.get('[data-cy="payment-details--filter-payment-currency-error"]'),
       noRecordsMatchingFiltersText: () => cy.get('[data-cy="payment-details-no-records-matching-filters-text"]'),
+      actionBar: () => cy.get('[data-cy="payment-details--filters-action-bar"]'),
+      actionBarItem: (value) => cy.get(`[data-cy="payment-details--filters-action-bar"] a:contains(${value})`),
+      clearFiltersLink: () => cy.get('a:contains("Clear filters")'),
```
ttbarnes, line 68: Can we use a `data-cy` attribute instead of `a:contains`?

trade-finance-manager-ui/component-tests/utilisation-reports/utilisation-report-reconciliation-for-report.component-test.ts
```diff
@@ -186,6 +187,22 @@ describe(page, () => {
 
       wrapper.expectElement(`div#premium-payments input[type="checkbox"]`).notToExist();
     });
+
+    it('should not display match success notification when param is false', () => {
```
ttbarnes, line 191: Can we wrap "match success notification" in some quotes for these tests please

trade-finance-manager-ui/component-tests/utilisation-reports/utilisation-report-reconciliation-for-report.component-test.ts
```diff
@@ -314,60 +333,46 @@ describe(page, () => {
...
+        },
+      });
 
-    wrapper.expectElement('[data-cy="match-success-notification"]').toExist();
-    wrapper.expectText('[data-cy="match-success-notification-heading"]').toRead('Match payment recorded');
-    wrapper
-      .expectText('[data-cy="match-success-notification-message"]')
-      .toRead('The fee(s) are now at a Match state. Further payments cannot be added to the fee record.');
+      wrapper.expectLink('[data-cy="download-report-link"]').toLinkTo(downloadUrl, 'Download the report submitted by the bank as a CSV');
+    });
```
ttbarnes, line 376: Sorry had to comment here as it's not possible to comment on an image file - it looks like `icon-tag-remove-cross.svg` has been added but is not being used, unless i'm missing something?

trade-finance-manager-ui/server/controllers/utilisation-reports/utilisation-report-reconciliation-for-report/get-payment-details-tab-href.test.ts
```diff
@@ -0,0 +1,85 @@
...
+          paymentCurrency: paymentCurrencyFilter,
+          paymentReference: paymentReferenceFilter,
+        };
+
+        // Act
+        const href = getPaymentDetailsTabHref(activeFilters, reportId);
+
+        // Assert
+        expect(href).toEqual(
+          `/utilisation-reports/${reportId}?paymentDetailsFacilityId=${facilityIdFilter}&paymentDetailsPaymentCurrency=${paymentCurrencyFilter}&paymentDetailsPaymentReference=${paymentReferenceFilter}#payment-details`,
```
ttbarnes, line 80: Can we have a `const expected` please - finding this hard to follow, could we look into splitting up some params into other consts, something like this?

```ts
const expected = `/utilisation-reports/${reportId}${expectedXParams}${expectedYParams}`;
```

trade-finance-manager-ui/server/controllers/utilisation-reports/utilisation-report-reconciliation-for-report/map-to-selected-payment-details-filters-view-model.test.ts
```diff
@@ -0,0 +1,142 @@
...
+        .mockReturnValue('/remove-payment-reference-filter');
+
+      // Act
+      const result = mapToSelectedPaymentDetailsFiltersViewModel(filters, reportId);
+
+      // Assert
+      expect(result!.paymentReference!.removeHref).toEqual('/remove-payment-reference-filter');
+    });
+  });
+});
```
ttbarnes, line 142: Can we test the full return pleae?

trade-finance-manager-ui/server/controllers/utilisation-reports/utilisation-report-reconciliation-for-report/map-to-selected-payment-details-filters-view-model.test.ts
```diff
@@ -0,0 +1,142 @@
+import { CURRENCY } from '@ukef/dtfs2-common';
+import { when } from 'jest-when';
+import { mapToSelectedPaymentDetailsFiltersViewModel } from './map-to-selected-payment-details-filters-view-model';
+import { getPaymentDetailsTabHref } from './get-payment-details-tab-href';
+
+jest.mock('./get-payment-details-tab-href');
```
ttbarnes, line 6: Can we call the helper please instead of mocking and hard coding the assertion? i.e

```ts
const expected = getPaymentDetailsTabHref({ ... });

expect(result!.paymentCurrency!.removeHref).toEqual(expected);
```

instead of:

```ts
expect(result!.paymentCurrency!.removeHref).toEqual('/remove-payment-currency-filter');
```

trade-finance-manager-ui/server/controllers/utilisation-reports/utilisation-report-reconciliation-for-report/map-to-selected-payment-details-filters-view-model.ts
```diff
@@ -0,0 +1,29 @@
...
+ * @returns the selected payment details filters view model
+ */
+export const mapToSelectedPaymentDetailsFiltersViewModel = (
+  filters: PaymentDetailsFilters,
+  reportId: string,
+): SelectedPaymentDetailsFiltersViewModel | null => {
+  if (!filters.facilityId && !filters.paymentCurrency && !filters.paymentReference) {
+    return null;
+  }
+  return {
```
ttbarnes, line 18: Finding the conditions/names a little hard to follow, what do you think about doing something like this instead?

```ts
  const mapped = {
    facilityId: null,
    paymentCurrency: null,
    paymentReference: null,
  };

  if (filters.facilityId) {
    mapped.facilityId = {
      value: filters.facilityId
      removeHref: getPaymentDetailsTabHref({ ...filters, facilityId: undefined }, reportId)
    };
  }

  if (filters.paymentCurrency) {
    // mapped.
  }

  return mapped;
```

trade-finance-manager-ui/server/types/view-models/utilisation-report-reconciliation-for-report-view-model.ts
```diff
@@ -109,9 +109,18 @@ export type PaymentDetailsRowViewModel = {
   };
 };
 
+export type SelectedFilter = { value: string; removeHref: string };
+
+export type SelectedPaymentDetailsFiltersViewModel = {
+  facilityId: SelectedFilter | null;
+  paymentCurrency: SelectedFilter | null;
+  paymentReference: SelectedFilter | null;
```
ttbarnes, line 117: Is it possible to do `?SelectedFilter` instead of `SelectedFilter | null`?

ttbarnes (changes requested):
Nice work, few comments to address please

trade-finance-manager-ui/component-tests/utilisation-reports/utilisation-report-reconciliation-for-report.component-test.ts
RLavender98: It is used by the moj filter component, without it the selected filters have no crosses (see screenshots in PR description to see cross showing up)

trade-finance-manager-ui/server/controllers/utilisation-reports/utilisation-report-reconciliation-for-report/map-to-selected-payment-details-filters-view-model.test.ts
RLavender98: I've split it so that we have separate assertions in separate tests for the different things that can individually go wrong, we lose that specificity if we test the full return in one.

trade-finance-manager-ui/server/controllers/utilisation-reports/utilisation-report-reconciliation-for-report/map-to-selected-payment-details-filters-view-model.test.ts
RLavender98: Has this been agreed as the pattern going forwards? I think this makes it harder to see the error going forwards if this test were ever to fail because the test output will display a long string that was generated by a function not in this file?

e2e-tests/tfm/cypress/e2e/pages/utilisation-reports/utilisationReportPage.js
ttbarnes: The nunjucks documentation isn't great for MOJ components - can we try adding a `data-cy` to the nunjucks - if it doesn't work then happy with the current selector

```
clearLink: {
  text: "Clear filters",
  attributes: {
    "data-cy": "testing"
  }
}
```

trade-finance-manager-ui/server/types/view-models/utilisation-report-reconciliation-for-report-view-model.ts
RLavender98: I did this because on previous PRs you have preferred us to not explicitly set things to undefined and as writing it I realised I was explicitly setting undefined so used null instead. I guess with the above refactor I can switch to undefined by starting with `const mapped = {}` so I can change this.

Just to be clear for going forwards, is the preference to have fields be optional and then if the logic would require us to explicitly set to undefined to keep the code readable, then in that case the preference switches to not be optional but allow null instead?

ttbarnes: Ah sorry @RLavender98 so changing this would mean we have to specifically set `undefined` values? If so I think the above implementation (`| null`) is better 👍

trade-finance-manager-ui/server/controllers/utilisation-reports/utilisation-report-reconciliation-for-report/map-to-selected-payment-details-filters-view-model.test.ts
RLavender98: Hmm I kind of see your point, we do have a return type on the function however so we would get a linting error if we accidentally threw in extra fields:
Screenshot 2024-10-23 at 17 35 28

trade-finance-manager-ui/server/controllers/utilisation-reports/utilisation-report-reconciliation-for-report/map-to-selected-payment-details-filters-view-model.test.ts
RLavender98: Ok will change.

trade-finance-manager-ui/templates/utilisation-reports/_macros/payment-details-filters-panel.njk
```diff
@@ -8,79 +8,86 @@
   {% set errors = params.errors %}
   {% set reportId = params.reportId %}
 
-  <div class="moj-filter-layout__filter" data-filtered-table-id="payment-details" data-cy="payment-details--filters-panel">
+  {% if object.field2 %}
```
ChristianMcc, line 11: what's this and the field3 stuff below?

RLavender98: Ah whoops, that was left in by accident!

trade-finance-manager-ui/server/controllers/utilisation-reports/utilisation-report-reconciliation-for-report/map-to-selected-payment-details-filters-view-model.test.ts
```diff
@@ -0,0 +1,127 @@
...
+
+      // Act
+      const result = mapToSelectedPaymentDetailsFiltersViewModel(filters, reportId);
+
+      // Assert
+      const expected = getPaymentDetailsTabHref({ facilityId: '12345', paymentReference: 'REF123' }, reportId);
+      expect(result!.paymentCurrency!.removeHref).toEqual(expected);
+    });
+
+    it.only('should set paymentReference remove href to link to report with other active filters still applied', () => {
```
ttbarnes, line 115: `.only`?

trade-finance-manager-ui/server/controllers/utilisation-reports/utilisation-report-reconciliation-for-report/map-to-selected-payment-details-filters-view-model.test.ts
```diff
@@ -0,0 +1,127 @@
...
+      // Assert
+      expect(result!.paymentReference!.value).toEqual('REF123');
+    });
+
+    it('should set facilityId remove href to link to report with other active filters still applied', () => {
+      // Arrange
+      const filters = { facilityId: '12345', paymentReference: 'REF123', paymentCurrency: CURRENCY.USD };
+
+      when(getPaymentDetailsTabHref)
+        .calledWith({ paymentReference: 'REF123', paymentCurrency: CURRENCY.USD }, reportId)
```
ttbarnes, line 92: Looks like we shouldn't need this

trade-finance-manager-ui/server/controllers/utilisation-reports/utilisation-report-reconciliation-for-report/map-to-selected-payment-details-filters-view-model.test.ts
RLavender98: Oh sorry, thanks for spotting

ttbarnes (approved):

PR #3765, opened 2024-10-24
feat(DTFS2-7573): call submit cancellation in tfm-api

MarRobSoftwire:
## Introduction :pencil2:
The tfm-api submit cancellation endpoint needs to call the dtfs-central-api endpoint.

- Add new api method `submitDealCancellation`
- Call `api.submitDealCancellation` in `DealCancellationService.submitDealCancellation`.
- Update unit test coverage

### DIFF
--- a/e2e-tests/tfm/cypress/e2e/journeys/deal-cancellation/check-details.spec.js
+++ b/e2e-tests/tfm/cypress/e2e/journeys/deal-cancellation/check-details.spec.js
@@ -1,6 +1,6 @@
 import relative from '../../relativeURL';
 import MOCK_DEAL_AIN from '../../../fixtures/deal-AIN';
-import { ADMIN, BANK1_MAKER1, PIM_USER_1, T1_USER_1 } from '../../../../../e2e-fixtures';
+import { ADMIN, BANK1_MAKER1, PIM_USER_1 } from '../../../../../e2e-fixtures';
 import caseDealPage from '../../pages/caseDealPage';
 import { backLink, errorSummary } from '../../partials';
 import { today, threeMonthsOneDay, yesterday } from '../../../../../e2e-fixtures/dateConstants';
@@ -29,38 +29,6 @@ context('Deal cancellation - check details', () => {
     cy.deleteDeals(dealId, ADMIN);
     dealFacilities.forEach((facility) => {
       cy.deleteFacility(facility._id, BANK1_MAKER1);
-    });
-  });
-
-  describe('when logged in as a non-PIM user', () => {
-    beforeEach(() => {
-      cy.login(PIM_USER_1);
-      cy.visit(relative(`/case/${dealId}/deal`));
-
-      caseDealPage.cancelDealButton().click();
-
-      cy.url().should('eq', relative(`/case/${dealId}/cancellation/reason`));
-      cy.clickContinueButton();
-
-      cy.url().should('eq', relative(`/case/${dealId}/cancellation/bank-request-date`));
-
-      cy.completeDateFormFields({ idPrefix: 'bank-request-date' });
-
-      cy.clickContinueButton();
-
-      cy.url().should('eq', relative(`/case/${dealId}/cancellation/effective-from-date`));
-
-      cy.completeDateFormFields({ idPrefix: 'effective-from-date', date: yesterday.date });
-
-      cy.clickContinueButton();
-
-      cy.login(T1_USER_1);
-
-      cy.visit(`/case/${dealId}/cancellation/check-details`);
-    });
-
-    it('should redirect when visiting the check details page', () => {
-      cy.url().should('eq', relative('/deals/0'));
     });
   });
 
--- a/e2e-tests/tfm/cypress/e2e/journeys/deal-cancellation/check-details.spec.js
+++ b/e2e-tests/tfm/cypress/e2e/journeys/deal-cancellation/check-details.spec.js
@@ -32,31 +32,31 @@ context('Deal cancellation - check details', () => {
     });
   });
 
-  beforeEach(() => {
-    cy.login(PIM_USER_1);
-    cy.visit(relative(`/case/${dealId}/deal`));
-
-    caseDealPage.cancelDealButton().click();
-
-    cy.url().should('eq', relative(`/case/${dealId}/cancellation/reason`));
-    cy.clickContinueButton();
-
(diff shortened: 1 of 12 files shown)


e2e-tests/tfm/cypress/e2e/journeys/deal-cancellation/check-details.spec.js
```diff
@@ -245,15 +249,13 @@ context('Deal cancellation - check details', () => {
         });
       });
     });
-  });
 
-  describe('when logged in as a non-PIM user', () => {
-    beforeEach(() => {
-      cy.login(T1_USER_1);
-    });
+    describe('"delete deals" button', () => {
```
MarRobSoftwire, line 253: This describe block is needed to ensure this test executes after the rest (clicking the button cannot be undone). 

Annoyingly if the `it` existed outside of this `describe` block it is hoisted above the describes & the tests fail

MarRobSoftwire: E.g, even when the tests are written in the following order
image

They execute in this order
image

and the `reason for cancelling - change link` tests fail as the deal cant be cancelled again

trade-finance-manager-api/src/v1/controllers/deal-cancellation/submit-deal-cancellation.controller.test.ts
```diff
@@ -1,18 +1,19 @@
...
+import { AnyObject, TestApiError } from '@ukef/dtfs2-common';
+import { generateTfmAuditDetails } from '@ukef/dtfs2-common/change-stream';
 import { submitDealCancellation, SubmitDealCancellationRequest } from './submit-deal-cancellation.controller';
 
 const submitDealCancellationMock = jest.fn() as jest.Mock<Promise<void>>;
 
 jest.mock('../../services/deal-cancellation/deal-cancellation.service', () => ({
   DealCancellationService: {
-    submitDealCancellation: (dealId: string, dealCancellation: TfmDealCancellation) => submitDealCancellationMock(dealId, dealCancellation),
+    submitDealCancellation: (params: AnyObject) => submitDealCancellationMock(params),
```
ttbarnes, line 12: Are we able to use specific params here?

trade-finance-manager-api/src/v1/controllers/deal-cancellation/submit-deal-cancellation.controller.test.ts
```diff
@@ -1,18 +1,19 @@
...
   },
 }));
 
-const dealCancellation = {
+const cancellation = {
```
ttbarnes, line 16: I think `dealCancellation` is nicer - if we add other cancellation functionality that is _not_ for a deal this could cause confusion.

Equally we're in a `deal-cancellation` directory so maybe it's ok

Thoughts?

trade-finance-manager-api/src/v1/services/deal-cancellation/deal-cancellation.service.ts
```diff
@@ -50,16 +56,12 @@ export class DealCancellationService {
    * @param dealId the Deal ID
    * @param dealCancellation the deal cancellation object
    */
-  public static async submitDealCancellation(dealId: string, dealCancellation: TfmDealCancellation) {
+  public static async submitDealCancellation({ dealId, cancellation, auditDetails }: SubmitDealCancellationParams) {
```
ttbarnes, line 59: Can we update the documentation please

ttbarnes (changes requested):

trade-finance-manager-api/src/v1/controllers/deal-cancellation/submit-deal-cancellation.controller.test.ts
MarRobSoftwire: No real preference - I just updated this so I could use `cancellation: dealCancellation` in the params. Have changed now

e2e-tests/tfm/cypress/e2e/journeys/deal-cancellation/check-details.spec.js
```diff
@@ -32,29 +32,39 @@ context('Deal cancellation - check details', () => {
...
+      cy.login(T1_USER_1);
+    });
 
-      cy.clickContinueButton();
+    it('should redirect when visiting the check details page', () => {
+      cy.url().should('eq', relative('/deals/0'));
     });
+  });
 
+  describe('when logged in as a PIM user', () => {
```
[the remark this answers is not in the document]
MarRobSoftwire: It was doing from the top level `beforeEach` but I've refactored to make it clearer whats happening now. Previously the `'when logged in as a non-PIM user'` test wasnt actually visiting the correct page (and wouldnt be able to unless a PIM user has previously created a cancellation)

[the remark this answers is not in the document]
MarRobSoftwire: The dificulty in testing the `non-pim user` does make me think this should just be covered by api-tests (though we dont have any for tfm-ui)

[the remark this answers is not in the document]
ttbarnes: Thanks @MarRobSoftwire I think the non-pim user should be in it's own E2E test so we can test it fully. I do not think it should only be covered by unit tests since there is different functionality for different user types

e2e-tests/tfm/cypress/e2e/journeys/deal-cancellation/check-details.spec.js
```diff
@@ -50,7 +50,7 @@ context('Deal cancellation - check details', () => {
 
       cy.url().should('eq', relative(`/case/${dealId}/cancellation/effective-from-date`));
 
-      cy.completeDateFormFields({ idPrefix: 'effective-from-date', date: tomorrow.date });
+      cy.completeDateFormFields({ idPrefix: 'effective-from-date', date: yesterday.date });
```
ttbarnes, line 53: What's the reason for changing from yesterday to tomorrow please?

MarRobSoftwire: The future date hasnt been implemented yet & will need a separate e2e test

ttbarnes (approved):

PR #3768, opened 2024-10-24
feat(FN-3328): screenreader ux improvements to payment details

RLavender98:
## Introduction :pencil2:
The payment details tab has a complciated table with grouped rows. These grouped rows mean that standard navigation with a screenreader does not convey the same meaning as the visual display. This PR improves that by making the screenreader experience line up with the order of reading of the table with the intended visual flow.

- Hide sub rows of grouped rows from screenreaders
- Hide facility id and exporter columns from screenreaders
- Add hidden column containing the facility id and exporter in a screenreader friendly format

### DIFF
--- a/trade-finance-manager-ui/component-tests/utilisation-reports/_macros/payment-details-table-row.component-test.ts
+++ b/trade-finance-manager-ui/component-tests/utilisation-reports/_macros/payment-details-table-row.component-test.ts
@@ -8,7 +8,7 @@ const component = '../templates/utilisation-reports/_macros/payment-details-tabl
 const render = componentRenderer(component, true);
 
 describe(component, () => {
-  const screenreaderAlternativeCellSelector = '[data-cy="screenreader-alternative-facility-id-and-expoter"]';
+  const hiddenCellSelector = '[data-cy="hidden-facility-id-and-expoter"]';
 
   const aPaymentDetailsPayment = (): PaymentDetailsPaymentViewModel => ({
     id: 1,
@@ -80,7 +80,7 @@ describe(component, () => {
     wrapper.expectElement(`tr td:contains("Some exporter")`).toHaveAttribute('aria-hidden', 'true');
   });
 
-  it('renders screenreader alternative combined facility ID and exporter cell', () => {
+  it('renders visually hidden alternative combined facility ID and exporter cell', () => {
     const paymentDetailsRow: PaymentDetailsRowViewModel = {
       ...aPaymentDetailsTableRow(),
       feeRecords: [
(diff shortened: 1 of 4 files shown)


trade-finance-manager-ui/templates/utilisation-reports/_macros/payment-details-table-row.njk
```diff
@@ -7,21 +7,33 @@
...
+        {% endfor %}
+      </ul>
+    </span>
+  {% endset %}
+
   {% for feeRecord in feeRecords %}
     {% set isFirstLineOfRow = loop.index === 1 %}
     {% set isLastLineOfRow = loop.index === feeRecords.length %}
-    <tr class="govuk-table__row" data-cy="payment-details-row--paymentId-{{ payment.id }}-feeRecordId-{{ feeRecord.id }}">
+    <tr class="govuk-table__row" aria-hidden="{{ false if isFirstLineOfRow else true }}"data-cy="payment-details-row--paymentId-{{ payment.id }}-feeRecordId-{{ feeRecord.id }}">
```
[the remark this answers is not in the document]
RLavender98: Ah yes it can be simplified, have changed to `aria-hidden="{{ not isFirstLineOfRow }}"` for the logic to be the same.

ttbarnes (approved):

trade-finance-manager-ui/component-tests/utilisation-reports/_macros/payment-details-table-row.component-test.ts
```diff
@@ -175,6 +197,18 @@ describe(component, () => {
...
     });
 
+    it('should visually hide all rows other than first row', () => {
+      const paymentDetailsRow: PaymentDetailsRowViewModel = {
+        ...aPaymentDetailsTableRow(),
+        feeRecords: [aFeeRecord(), aFeeRecord(), aFeeRecord()],
+      };
+      const wrapper = getWrapper({ paymentDetailsRow });
+
+      wrapper.expectElement('tr:eq(0)').toHaveAttribute('aria-hidden', 'false');
```
ChristianMcc, line 207: Not something to block merging this PR at all so will approve, could something like this be a loop so the test doesn't need to know how many <tr> blocks there will be? 
Wondering if we can loop over everything matching 'tr', from a quick play it looks like we'd have to write our own helper so the answer is probably not easily sadly

ChristianMcc (approved):
LGTM, left a minor nice to have that isn't for this PR

trade-finance-manager-ui/component-tests/utilisation-reports/_macros/payment-details-table-row.component-test.ts
RLavender98: Yeah I'm not aware of a simple way of achieving this so possibly not worth it, but agree that would be nicer if done in a nice way. I might have a play later if I have some down time.

PR #3769, opened 2024-10-24
feat(DTFS2-7556): add activity on deal cancellation

MarRobSoftwire:
## Introduction :pencil2:
When the user submits a deal cancellation the activity and comments page should be updated

- Optionally update activity on `TfmDealCancellationRepo.submitDealCancellation` 
- Create activity object in `DealCancellationService.cancelDeal` from users auditDetails, throwing `InvalidAuditDetailsError` if this fails.
- Update api & unit tests

- Add createTfmUser test helper
- Add a generic to `AuditDetails<T>` type to allow filtering by `userType` - update `validateAuditDetailsAndUserType` to use this


libs/common/src/change-stream/validate-audit-details.ts
```diff
@@ -53,7 +53,7 @@ export function validateAuditDetails(auditDetails: unknown): asserts auditDetail
   }
 }
 
-export function validateAuditDetailsAndUserType(auditDetails: unknown, requiredUserType: AuditUserTypes): asserts auditDetails is AuditDetails {
+export function validateAuditDetailsAndUserType<T extends AuditUserTypes>(auditDetails: unknown, requiredUserType: T): asserts auditDetails is AuditDetails<T> {
```
ttbarnes, line 56: Do we really need `<T`?

ttbarnes (changes requested):
1x minor comment otherwise LGTM

libs/common/src/change-stream/validate-audit-details.ts
MarRobSoftwire: In `dtfs-central-api\src\v1\controllers\tfm\deal-cancellation\tfm-post-submit-deal-cancellation.controller.ts` there's the lines 
```    
validateAuditDetailsAndUserType(auditDetails, AUDIT_USER_TYPES.TFM);

const cancelDealResponse = await DealCancellationService.cancelDeal(dealId, cancellation, auditDetails);
```

This change means `auditDetails` is asserted to be type `AuditDetails<AUDIT_USER_TYPES.TFM>` which means, its passed into `DealCancellationService.cancelDeal` without throwing any type errors & without needing to add any further type assertions. Its all backwards compatible too & works behind the scenes without any `validateAuditDetailsAndUserType` calls & `AuditDetails` types needing to be modified

MarRobSoftwire: I dont think I'd say its cleaner - I'd probably say the opposite. My concern here is that it wont stay up to date if i changed the first line to validate as `validateAuditDetailsAndUserType(auditDetails, AUDIT_USER_TYPES.PORTAL);` I'd need to remember to change the second line to use `auditDetails as PortalAuditDetails`. I

My understanding is its best to avoid using `as` where possible since this can overwrite with incorrect typing if you're not careful.

I appreciate the generic has an increased mental load when looking at the function definition `validateAuditDetailsAndUserType` but I think it reduces it when looking at places its used.

Do you think it would be helpful if I gave a more descriptive name here?

```suggestion
export function validateAuditDetailsAndUserType<TRequiredUserType extends AuditUserTypes>(auditDetails: unknown, requiredUserType: TRequiredUserType ): asserts auditDetails is AuditDetails<TRequiredUserType > {
```

ttbarnes: `TRequiredUserType` seems like a good compromise

ttbarnes (dismissed):

ttbarnes (approved):

PR #3770, opened 2024-10-25
feat(DTFS2-7298): update facility stage in the db on cancel

BethThomas141:
## Introduction :pencil2:
When we cancel a deal we want to add a new field to every associated facility to the `facility.tfm` object called `facilityStage` which will take the value 'Risk Expired'. 

- Update the api call in dtfs-central-api to update all the facilities associated with the deal with the new facility stage
- Return the ids of the updated facilities (these will be passed into tfm-api to use when sending the email, this will be in a separate PR)

- I'm now calling this new field `facilityStage` as opposed to just `stage` on the tfm object as I think it makes it clearer
- In the TFM API layer where this gets mapped onto the facility snapshot, I've added unit tests to check this gets mapped over correctly


dtfs-central-api/src/repositories/tfm-deals-repo/tfm-deal-cancellation.repo.ts
```diff
@@ -167,6 +173,20 @@ export class TfmDealCancellationRepo {
...
+      { 'facilitySnapshot.dealId': { $eq: new ObjectId(dealId) } },
+      flatten({
+        'tfm.facilityStage': TFM_FACILITY_STAGE.RISK_EXPIRED,
+        auditRecord: generateAuditDatabaseRecordFromAuditDetails(auditDetails),
+      }),
+    );
+
+    const updatedFacilities = await facilityCollection.find({ 'facilitySnapshot.dealId': { $eq: new ObjectId(dealId) } }).toArray();
+
+    const updatedFacilityUkefIds = updatedFacilities.map((facility) => facility.facilitySnapshot.ukefFacilityId).filter((id) => id !== null);
```
ttbarnes, line 188: I think we may already have a helper for this @MarRobSoftwire created in another PR, can we use this?

dtfs-central-api/src/repositories/tfm-deals-repo/tfm-deal-cancellation.repo.ts
```diff
@@ -167,6 +173,20 @@ export class TfmDealCancellationRepo {
...
+    await facilityCollection.updateMany(
+      { 'facilitySnapshot.dealId': { $eq: new ObjectId(dealId) } },
+      flatten({
+        'tfm.facilityStage': TFM_FACILITY_STAGE.RISK_EXPIRED,
+        auditRecord: generateAuditDatabaseRecordFromAuditDetails(auditDetails),
+      }),
+    );
+
+    const updatedFacilities = await facilityCollection.find({ 'facilitySnapshot.dealId': { $eq: new ObjectId(dealId) } }).toArray();
+
```
ttbarnes, line 187: I haven't worked with MongoDB for a while - is it possible to return the facilities from `.updateMany` instead of having to do a `.find` afterwards?

trade-finance-manager-api/src/v1/rest-mappings/mappings/facilities/mapFacilitySnapshot.test.js
```diff
@@ -141,4 +142,14 @@ describe('mapFacility', () => {
 
     expect(result).toEqual(expected);
   });
+
+  describe('when a tfm facilityStage exists', () => {
```
ttbarnes, line 146: I think we can just add the call to the main test above as it's not really conditional in the main function, it's just calling `mapBssEwcsFacilityStage` - `mapBssEwcsFacilityStage` handles if `facilityStage` exists or not and we're just consuming that here

trade-finance-manager-api/src/v1/rest-mappings/mappings/gef-facilities/mapGefFacility.test.js
```diff
@@ -65,4 +66,14 @@ describe('mapGefFacility', () => {
 
     expect(result).toEqual(expected);
   });
+
+  describe('when a tfm facilityStage exists', () => {
```
ttbarnes, line 70: Same as above RE combine into the main test above

ttbarnes (changes requested):
Nice work, few minor comments

dtfs-central-api/src/repositories/tfm-deals-repo/tfm-deal-cancellation.repo.ts
MarRobSoftwire: That's in tfm-api currently but it will be removed once this code is in, I think makes sense to copy the helper for now & remove the current one later

ttbarnes: Thanks @MarRobSoftwire  - can/should we create a subtask for this so we don't forget?

BethThomas141: Thanks both! If OK with you @ttbarnes now @MarRobSoftwire's latest PR is merged in, I'm happy to do another commit here, moving the helper to dtfs-central and removing it from tfm-api (and updating the logic in tfm-api respectively). Will add a few more file changes to the diff here, and need a re-review but I think it will be the neatest option!

ttbarnes: Sounds good to me @BethThomas141 👍

dtfs-central-api/src/repositories/tfm-deals-repo/tfm-deal-cancellation.repo.ts
BethThomas141: Unfortunately it seems like you can't - there's `findOneAndUpdate` which allows you to update a single facility and return the updated facility in the response, but as soon as you update multiple facilities at a time you can't return them all 😞

trade-finance-manager-api/src/v1/rest-mappings/mappings/facilities/mapFacilitySnapshot.test.js
BethThomas141: Sure, have removed this separate test👍

trade-finance-manager-api/src/v1/rest-mappings/mappings/gef-facilities/mapGefFacility.test.js
BethThomas141: Done, thanks!

dtfs-central-api/src/helpers/get-ukef-facility-ids.test.ts
```diff
@@ -1,11 +1,9 @@
 import { TfmFacility } from '@ukef/dtfs2-common';
-import { MOCK_FACILITY_SNAPSHOT } from '../../../__mocks__/mock-facility-snapshot';
 import { getUkefFacilityIds } from './get-ukef-facility-ids';
 
 const aFacility = ({ ukefFacilityId }: { ukefFacilityId: unknown }): TfmFacility =>
   ({
     facilitySnapshot: {
-      ...MOCK_FACILITY_SNAPSHOT,
```
BethThomas141, line 8: Moved this helper file from tfm-api to dtfs-central (this mock snapshot is a tfm-api object, so removed here and is not needed for the tests)

dtfs-central-api/src/repositories/tfm-deals-repo/tfm-deal-cancellation.repo.submitDealCancellation.test.ts
```diff
@@ -28,15 +29,15 @@ const auditDetails = generateTfmAuditDetails(tfmUserId);
...
 
-  beforeAll(() => {
-    jest.useFakeTimers();
-  });
+  const matchingFacilityId1 = new ObjectId();
+  const matchingFacilityId2 = new ObjectId();
 
-  afterAll(() => {
-    jest.useRealTimers();
-  });
```
BethThomas141, line 39: Accidentally two sets of using fake timers was added to this file, so removing one set here (further down in the file we still do this)

e2e-tests/tfm/cypress/e2e/journeys/deal-cancellation/submit-deal-cancellation-in-past.spec.js
```diff
@@ -0,0 +1,80 @@
...
+  });
+
+  after(() => {
+    cy.deleteDeals(dealId, ADMIN);
+    dealFacilities.forEach((facility) => {
+      cy.deleteFacility(facility._id, BANK1_MAKER1);
+    });
+  });
+
+  describe('when logged in as a PIM user', () => {
```
ttbarnes, line 39: Do we have a test for a non-PIM user?

MarRobSoftwire: I dont think this is testable as a non-PIM user - we check the page doesnt render in another file, so checking they cant POST is impossible in a e2e test

BethThomas141: Yep, there are tests in another file to check the deal cancellation button itself doesn't appear for a non-PIM user, and I don't _think_ there's anything helpful we can test in an E2E test here for non-PIM.

ttbarnes: Makes sense, thanks - sorry I was just reading the `describe` without realising the context of the actual test 😄

ttbarnes (dismissed):

libs/common/src/types/tfm-deal-cancellation.ts
```diff
@@ -6,5 +6,5 @@ export type TfmDealCancellation = z.infer<typeof DEAL_CANCELLATION>;
 
 export type TfmDealCancellationResponse = {
   cancelledDealUkefId: string | ObjectId;
-  riskExpiredFacilityUkefIds?: (string | ObjectId)[];
```
MarRobSoftwire, line 9: Actually ive jsut realised what youre doing I think this makes sense as is

MarRobSoftwire (dismissed):

ttbarnes (approved):

PR #3771, opened 2024-10-25
fix(DTFS2-7567): acbs container initilisation

abhi-markan:
## Introduction :pencil2:
ACBS microservice container did not initialise properly despite no errors on the logs, it was unable to accept any request and returned `404` status code. 

* Removed `libs` library due to lack of Typescript support in current Azure functions base image.
* Removed root `package-lock.json` to ensure only `acbs-function` `package.json` is included in the container and replaced `ci` with `i` (since no package-lock.json exists due to workspaces)

* Moved few test pertinent dependencies to `devDependencies`.

### DIFF
--- a/azure-functions/acbs-function/Dockerfile
+++ b/azure-functions/acbs-function/Dockerfile
@@ -10,9 +10,7 @@ ENV AzureWebJobsFeatureFlags=EnableWorkerIndexing
 # Directories setup
 WORKDIR /app
 
-COPY libs libs
 COPY azure-functions/acbs-function .
 
-COPY package*.json .
-RUN npm ci
+RUN npm i
 RUN npm cache clean --force
--- a/azure-functions/acbs-function/package.json
+++ b/azure-functions/acbs-function/package.json
@@ -32,17 +32,16 @@
     "axios": "^1.7.7",
     "date-fns": "^2.30.0",
     "dotenv": "^16.4.5",
-    "durable-functions": "^3.1.0",
-    "lodash": "^4.17.21"
+    "durable-functions": "^3.1.0"
   },
   "devDependencies": {
-    "@ukef/dtfs2-common": "1.0.0",
     "babel-jest": "29.5.0",
     "babel-loader": "9.1.2",
     "eslint": "^8.57.1",
     "eslint-config-airbnb-base": "^15.0.0",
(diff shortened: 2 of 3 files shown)


azure-functions/acbs-function/package.json
```diff
@@ -32,21 +32,20 @@
     "axios": "^1.7.7",
     "date-fns": "^2.30.0",
     "dotenv": "^16.4.5",
-    "durable-functions": "^3.1.0",
-    "lodash": "^4.17.21"
+    "durable-functions": "^3.1.0"
   },
   "devDependencies": {
-    "@ukef/dtfs2-common": "1.0.0",
```
ChristianMcc, line 39: ah we were not actually using anything in common we had just imported it in?

ttbarnes (approved):

azure-functions/acbs-function/Dockerfile
```diff
@@ -10,9 +10,7 @@ ENV AzureWebJobsFeatureFlags=EnableWorkerIndexing
 # Directories setup
 WORKDIR /app
 
-COPY libs libs
 COPY azure-functions/acbs-function .
 
-COPY package*.json .
-RUN npm ci
+RUN npm i
```
ChristianMcc, line 15: how come we're going from ci -> i? This would worry me a little bit in terms of the package-lock.json not necessarily mapping to what's actually being installed. What people see locally in their package-lock.json compared to what gets built in the dockerfile could be different with this I think

azure-functions/acbs-function/package.json
abhi-markan: We were but due to base image lack of TS support this has been moved to a separate ticket where we will use `libs/common` contants.

azure-functions/acbs-function/Dockerfile
ChristianMcc: ah I see, that sounds like something to be sorted out but not here, approving!

ChristianMcc (approved):

azure-functions/acbs-function/Dockerfile
abhi-markan: There is a ticket `DTFS2-7456` associated with TS upgrade.

PR #3772, opened 2024-10-28
fix(DTFS2-7440): acbs microservice docker communication

abhi-markan:
## Introduction :pencil2:
Locally `external-api` is unable to communicate with `acbs` microservice, the issue is due to recent `acbs` docker service update inline with other microservices however the ACBS URL has not been updated.

* Set `AZURE_ACBS_FUNCTION_URL` to `http://acbs` when executing `acbs` microservice locally using `docker compose` command. This does not have any impact of Azure RG.

* Disabled SOF `acbs-amend-facility-covenant-record` and `acbs-amend-facility-guarantee-record`.

### DIFF
--- a/azure-functions/acbs-function/src/functions/acbs-amend-facility.js
+++ b/azure-functions/acbs-function/src/functions/acbs-amend-facility.js
@@ -139,9 +139,9 @@ df.app.orchestration('acbs-amend-facility', function* amendFacility(context) {
     });
 
     /**
-     * Below have disabled due to BR.
-     * Enable below two alongside with investor and fixed fee records.
      * TODO: DTFS2-7581
+     * Below have been disabled for release 2.4.0.
+     * Enable below two alongside with investor and fixed fee records.
      */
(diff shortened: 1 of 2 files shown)


azure-functions/acbs-function/src/functions/acbs-amend-facility.js
```diff
@@ -138,23 +138,28 @@ df.app.orchestration('acbs-amend-facility', function* amendFacility(context) {
       amendments,
     });
 
+    /**
```
ChristianMcc, line 141: are we happy for this to go back to main commented out and then be fixed as part of 7581?

abhi-markan: No we want this to be commented in `main` until next release which will enable it. Once an existing PR to merge `release-2.4.0` is merged to `main` we should not need to do anything more on `main`.

Zainzzkk (approved):

PR #3774, opened 2024-10-28
feat(FN-1528): add test for bank holiday backup update reminder

RLavender98:
## Introduction :pencil2:
We have backup data stored for the UK bank holidays.
This data needs updating regularly to include more future dates.

- Add a test that will fail in July 2026 to prompt an update of the backup data
- Update the backup data


external-api/src/external-api/bank-holidays.test.ts
```diff
@@ -0,0 +1,18 @@
...
+describe('bank-holidays', () => {
+  /**
+   * If this test fails it means it is time to update the bank holiday backup data.
+   * The backup should be retrieved from https://www.gov.uk/bank-holidays.json.
+   *
+   * After updating the backup, the deadline in the test should be extended
+   * to be 6 months before the date of the final bank holiday in the backup.
+   */
+  it('should have up to date backup data for bank holidays', () => {
+    const deadlineForUpdatingBackup = new Date('2026-07-01');
```
RLavender98, line 12: This is a year in the future compared to the ticket AC because I updated the back up data before realising the ticket didn't actually require me to do that but then I thought might as well since I had done it so extended the deadline also. 

@ChristianMcc is this the correct periodicity for updating?

RLavender98, line 12: The deadline is a year in the future compared to the ticket AC because I updated the back up data before realising the ticket didn't actually require me to do that, but then I thought might as well since I had done it so extended the deadline also. 

@ChristianMcc is this the correct periodicity for updating?

external-api/src/external-api/bank-holidays.test.ts
```diff
@@ -0,0 +1,18 @@
+import { isBefore } from 'date-fns';
+
+describe('bank-holidays', () => {
+  /**
+   * If this test fails it means it is time to update the bank holiday backup data.
+   * The backup should be retrieved from https://www.gov.uk/bank-holidays.json.
+   *
+   * After updating the backup, the deadline in the test should be extended
+   * to be 6 months before the date of the final bank holiday in the backup.
+   */
```
ttbarnes, line 10: I think a small CRON job that sends emails and/or maybe a Teams notification would be the ideal solution

But this seems to be the most sensible approach for now

ttbarnes (approved):

external-api/src/external-api/bank-holidays.test.ts
ChristianMcc: Yep this is good, thank you and good we could already update it!

ChristianMcc (approved):

external-api/src/external-api/bank-holidays.test.ts
RLavender98: FWIW I think we decided on this solution since it was easy to implement, ties it to the code and also is not ignorable. It would be fairly easy to accidentally ignore an email and not update it, especially if the people working on the project/channel of communication changed.

We did consider caching the api responses which would remove the need for manual updates but as is detailed on the ticket it was not deemed worth the development time.

ttbarnes: Thanks @RLavender98 Concern is that if no-one is working on the project it could be missed

Not a blocker, this is another piece, this solution is great

RLavender98: Ah true that is an angle I hadn't considered!

PR #3775, opened 2024-10-28
feat(DTFS2-7298): update deal cancellation in progress logic

BethThomas141:
## Introduction :pencil2:
On the deal cancellation flow pages we want to check if a deal cancellation is in progress, and if it's not redirect away from them (e.g. you shouldn't be able to get to the check answers page if you haven't already started a cancellation, likewise with the bank request date page which should only be accessible if you've entered a reason for cancelling).

Previously, we were checking this by checking if the deal cancellation object was empty or not (and if it was empty it meant it hadn't yet been started). 

Since then, we've introduced a new field `status` on a cancellation that has the value DRAFT if it's been started. 

This PR updates the logic instead to base the redirection on the cancellation `status` instead of whether or not the cancellation is empty or not. (Which has the bonus benefit of meaning you are redirected away from these pages if the cancellation has already been submitted). 

- Replace `isEmpty(cancellation)` checks with `!isDealCancellationInDraft(cancellation.status)`
- Update the tests to check additional case where the cancellation has already been submitted


MarRobSoftwire (approved):
LGTM - thanks for doing this :)

ttbarnes (approved):

PR #3777, opened 2024-10-28
feat(FN-3518): hide select all when there is nothing to select

RLavender98:
## Introduction :pencil2:
Currently we always display the select all checkbox on the premium payments tab, even if there is nothing to select.

- Hide select all checkbox and it's corresponding column when there are no selectable premium payments

- Nest premium payments tab only fields of view model to match pattern for other tabs

### DIFF
--- a/trade-finance-manager-ui/component-tests/utilisation-reports/_macros/premium-payments-table.component-test.ts
+++ b/trade-finance-manager-ui/component-tests/utilisation-reports/_macros/premium-payments-table.component-test.ts
@@ -19,6 +19,7 @@ type ComponentRendererParams = {
   userCanEdit: boolean;
   feeRecordPaymentGroups: PremiumPaymentsViewModelItem[];
   enablePaymentsReceivedSorting: boolean;
+  displaySelectAllCheckbox: boolean;
 };
 
 const render = componentRenderer<ComponentRendererParams>(component);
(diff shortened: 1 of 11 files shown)


trade-finance-manager-ui/server/controllers/utilisation-reports/utilisation-report-reconciliation-for-report/index.ts
```diff
@@ -92,7 +93,17 @@ export const getUtilisationReportReconciliationByReportId = async (req: GetUtili
 
     const enablePaymentsReceivedSorting = premiumPaymentsGroupsHaveAtLeastOnePaymentReceived(premiumPayments);
 
-    const premiumPaymentsViewModel = mapPremiumPaymentsToViewModelItems(premiumPayments, isCheckboxChecked);
+    const premiumPaymentsItems = mapPremiumPaymentsToViewModelItems(premiumPayments, isCheckboxChecked);
+
+    const premiumPaymentsViewModel = {
```
sw-mattw, line 98: Nice, this is a lot tidier thanks

trade-finance-manager-ui/server/types/view-models/utilisation-report-reconciliation-for-report-view-model.ts
```diff
@@ -149,17 +150,22 @@ export type UtilisationDetailsViewModel = {
...
 export type UtilisationReportReconciliationForReportViewModel = BaseViewModel & {
   bank: SessionBank;
   formattedReportPeriod: string;
   reportId: string;
-  premiumPayments: PremiumPaymentsViewModelItem[];
-  premiumPaymentsFilters?: PremiumPaymentsFilters;
-  premiumPaymentsFilterError?: ErrorSummaryViewModel;
-  premiumPaymentsTableDataError?: ErrorSummaryViewModel;
-  enablePaymentsReceivedSorting: boolean;
+  premiumPayments: PremiumPaymentsViewModel;
```
sw-mattw, line 167: Lovely

trade-finance-manager-ui/component-tests/utilisation-reports/_macros/premium-payments-table.component-test.ts
```diff
@@ -137,13 +124,18 @@ describe(component, () => {
...
     wrapper.expectElement(`${tableSelector} thead td input[type="checkbox"]#select-all-checkbox`).toExist();
   });
 
   it('should not render the select all checkbox in the table headings row when userCanEdit is false', () => {
-    const wrapper = render({ ...defaultRendererParams(), userCanEdit: false });
+    const wrapper = render({ ...defaultRendererParams(), userCanEdit: false, hasSelectableRows: true });
+    wrapper.expectElement(`${tableSelector} thead th input[type="checkbox"]#select-all-checkbox`).notToExist();
+  });
+
+  it('should not render the select all checkbox in the table headings row when hasSelectableRows is false', () => {
```
sw-mattw, line 137: Slightly picky - should we reference userCanEdit being true in the naming too?

e.g. 
```suggestion
  it('should not render the select all checkbox in the table headings row when userCanEdit is true and hasSelectableRows is false', () => {
```

Happy to hear thoughts, this would also apply to the above test case

sw-mattw (approved):
Nice work, particularly a fan of the grouped premium payments data - thanks

trade-finance-manager-ui/component-tests/utilisation-reports/_macros/premium-payments-table.component-test.ts
RLavender98: I left it out to try and keep the name shorter but also precise to the assertion. I have set the userCanEdit to true explicitly so that the test actually fails if you change it, but I don't think we're really too interested in that in this test? What do you think?

sw-mattw (approved):

ttbarnes (changes requested):
1x minor comment otherwise LGTM

RLavender98:
I have added an e2e test also for checking it does not show up when no actions

sw-mattw (approved):

ttbarnes (approved):

PR #3779, opened 2024-10-29
feat(FN-3441): make report entity mock builder constructor public

RLavender98:
## Introduction :pencil2:
The UtilisationReportEntityMockBuilder was created with a private constructor and it was required to pass in a status when creating one. Since then we have found cases where the status of the report is not important and rather than make the constructor public we created an anti-pattern helper named 'aUtilisationReport'.

- Make constructor of UtilisationReportEntityMockBuilder so can be created without a status
- Remove usages of `aUtilisationReport` and replace with `new UtilisationReportEntityMockBuilder().build()` to consistently use the builder pattern for mock entity creation.


dtfs-central-api/src/mapping/fee-record-mapper.test.ts
```diff
@@ -132,7 +133,9 @@ describe('fee record mapper', () => {
 
     it('maps the fee record entity fees paid to the fee record reported fees amount', () => {
       // Arrange
-      const feeRecordEntity = FeeRecordEntityMockBuilder.forReport(aUtilisationReport()).withFeesPaidToUkefForThePeriod(314.59).build();
+      const feeRecordEntity = FeeRecordEntityMockBuilder.forReport(new UtilisationReportEntityMockBuilder().build())
+        .withFeesPaidToUkefForThePeriod(314.59)
```
Zainzzkk, line 137: can this number be a const, and the same for rest in this file.  const feesPaid = 314.59.  that way it only has to be changed once if we change the value for the test.

libs/common/src/test-helpers/mock-data/utilisation-report.entity.mock-builder.ts
```diff
@@ -1,13 +1,13 @@
...
 import { UtilisationReportReconciliationStatus } from '../../types';
 import { MOCK_AZURE_FILE_INFO } from './azure-file-info.mock';
 
 export class UtilisationReportEntityMockBuilder<ReportStatus extends UtilisationReportReconciliationStatus> {
   private readonly report: UtilisationReportEntity;
 
-  private constructor(report: UtilisationReportEntity) {
-    this.report = report;
+  public constructor(report?: UtilisationReportEntity) {
+    this.report = report ?? UtilisationReportEntityMockBuilder.forStatus(UTILISATION_REPORT_RECONCILIATION_STATUS.RECONCILIATION_IN_PROGRESS).report;
```
Zainzzkk, line 10: this line is pretty long.  Could either RECONCILIATION_IN_PROGRESS be dereferenced above or put in a const

RLavender98: This will be covered by other PR I have open, would like to merge this first please

dtfs-central-api/src/mapping/fee-record-mapper.test.ts
RLavender98: That is out of scope of this ticket, I'm just replacing `aUtilisationReport()`

Zainzzkk (dismissed):

ttbarnes (approved):

Zainzzkk (approved):

PR #3780, opened 2024-10-29
feat(FN-3456): use first visible worksheet of bank upload

sw-mattw:
## Introduction :pencil2:
At present, we use only the first worksheet in the uploaded workbook. This leads to confusing validation errors when the first worksheet is hidden.

- Skip any hidden worksheets and use the first visible worksheets content.
- Added an e2e test with a workbook with a hidden first worksheet and visible second worksheet.

Background: An Excel file must contain at least 1 visible worksheet at all times.


e2e-tests/portal/cypress/fixtures/valid-utilisation-report-hidden-first-sheet-February_2023_monthly.xlsx
sw-mattw, line 1: The first sheet of this file is hidden and the second sheet is visible with valid data

portal/server/utils/csv-utils.js
```diff
@@ -236,7 +236,8 @@ const extractCsvData = async (file) => {
       // Read the .xlsx file using exceljs
       const workbook = new ExcelJS.Workbook();
       await workbook.xlsx.load(file.buffer, { sheetStubs: true }).then(async () => {
-        const worksheet = workbook.worksheets[0]; // Assume the utilisation report data is on the first sheet
+        // An Excel file must contain at least 1 visible worksheet at all times.
+        const worksheet = workbook.worksheets.find((sheet) => sheet.state === 'visible');
```
ttbarnes, line 240: Can/should we add a log and/or return an error of some kind if there is no visible worksheet?

Seems otherwise we'll be creating an XLSX and proceeding with the rest of the journey when no data is available/visible?

sw-mattw: Find will return a worksheet object or undefined - in theory we should never hit the undefined case as we can't have an xlsx file without at least 1 visible worksheet, if we did there would be an error in the function call below this which will be caught by the try-catch block in this function.

We could add a check after this line to throw a more specific error though if you think that may be beneficial, e.g.:
```
if (!worksheet) {
  throw new Error('No visible worksheet found in the Excel file');
}
```

ttbarnes: Sounds good yes please @sw-mattw just thinking for debugging purposes, if we could have a `console.error` at least - not sure off the top of my head if throwing an error would bubble up and be handled gracefully, rather than crashing anything(?)

sw-mattw: No worries, I've added the error code block from the above comment. This should be caught by the try-catch within this function which calls `console.error` itself with the error object and doesn't re-throw - so this should all be okay thanks

ttbarnes: Nice one thanks @sw-mattw

ttbarnes (dismissed):

portal/server/utils/csv-utils.js
```diff
@@ -236,7 +236,11 @@ const extractCsvData = async (file) => {
       // Read the .xlsx file using exceljs
       const workbook = new ExcelJS.Workbook();
       await workbook.xlsx.load(file.buffer, { sheetStubs: true }).then(async () => {
-        const worksheet = workbook.worksheets[0]; // Assume the utilisation report data is on the first sheet
+        // An Excel file must contain at least 1 visible worksheet at all times.
```
RLavender98, line 239: Please can we still also have the comment about the report data being on the first sheet just reworded:
`Assume the utilisation report data is on the first visible sheet.`

sw-mattw: Fair point - updated to this thanks

Also removed the comment that a `file must contain at least 1 visible worksheet at all times` as I feel the error message in the `if` block now covers that

sw-mattw: Added the comment back now thanks

RLavender98 (approved):

ttbarnes (approved):

Zainzzkk (approved):

PR #3781, opened 2024-10-29
feat(DTFS2-7298): display deal cancellation success message

BethThomas141:
## Introduction :pencil2:
When a user submits a cancellation in the past (or present) then they should be shown a 1 time banner saying the deal cancellation was successful. When they refresh the page the banner should disappear. 

- Add a success message nunjucks template that is conditionally displayed on the case page
- use `flash` to store a one time message (this is how we handle similar messages in `portal-ui`)
- Add unit, component & e2e tests

**Screen recording to show implementation:**
In the following I submit a deal cancellation in the past, see the success banner and then refresh the page to show it subsequently disappears. 

https://github.com/user-attachments/assets/5ce6485a-0d7b-4a2e-b577-9c16871d2756


trade-finance-manager-ui/server/controllers/case/cancellation/check-details.controller.post.test.ts
```diff
@@ -9,14 +9,16 @@ jest.mock('../../../api', () => ({
...
+
 const dealId = 'dealId';
 const ukefDealId = 'ukefDealId';
 
 const reason = 'reason';
 const bankRequestDate = new Date().valueOf().toString();
 const effectiveFrom = new Date().valueOf().toString();
 
-describe('postBankRequestDate', () => {
+describe('postDealCancellationDetails', () => {
```
BethThomas141, line 21: This test was incorrectly named before

trade-finance-manager-ui/server/helpers/getFlashSuccessMessage.js
```diff
@@ -0,0 +1,3 @@
+const getFlashSuccessMessage = (req) => req.flash('successMessage')[0];
```
BethThomas141, line 1: File copied exactly over from `portal-ui`

trade-finance-manager-ui/templates/_partials/case_sub_navigation.njk
```diff
@@ -1,6 +1,6 @@
 {% from "moj/components/primary-navigation/macro.njk" import mojPrimaryNavigation %}
   <div class="govuk-grid-row">
-    <div class="govuk-grid-column-full govuk-!-padding-top-8">
+    <div class="govuk-grid-column-full  {{ 'govuk-!-padding-top-8' if not successMessage }} ">
```
BethThomas141, line 3: This is a stylistic change to remove the padding above the tabs on the case page when the success message is visible to keep in line with the designs

e2e-tests/tfm/cypress/e2e/journeys/deal-cancellation/submit-deal-cancellation-in-past.spec.js
```diff
@@ -36,45 +38,63 @@ context('Deal cancellation - effective from date in past', () => {
...
   describe('when logged in as a PIM user', () => {
     before(() => {
       cy.login(PIM_USER_1);
-
       cy.visit(relative(`/case/${dealId}/deal`));
+    });
 
-      caseDealPage.cancelDealButton().click();
+    describe('after submitting the cancellation', () => {
+      it('should redirect you to the deal summary page with the success banner visible once', () => {
```
BethThomas141, line 66: I have to do all this in one test - if it's split out into multiple tests then the window get's refreshed/reloaded as part of starting up the new test, which means the success banner disappears

e2e-tests/tfm/cypress/e2e/journeys/deal-cancellation/submit-deal-cancellation-in-past.spec.js
```diff
@@ -36,45 +38,63 @@ context('Deal cancellation - effective from date in past', () => {
...
-      checkDetailsPage.dealDeletionButton().click();
-    });
+      it('should not show the "Cancel Deal" button', () => {
+        caseDealPage.cancelDealButton().should('not.exist');
+      });
 
-    it('should redirect you to the deal summary page', () => {
-      cy.url().should('eq', relative(`/case/${dealId}/deal`));
-    });
+      it('should show the deal status as "Cancelled"', () => {
```
ttbarnes, line 82: Can we use a constant here instead of hard coded "Cancelled"?

e2e-tests/tfm/cypress/e2e/pages/caseDealPage.js
```diff
@@ -10,6 +10,7 @@ const caseDealPage = {
   bankSecurityFacilityText: () => cy.get('[data-cy="bank-security-facility-text"]'),
   cancelDealButton: () => cy.get('[data-cy="cancel-deal-button"]'),
   dealStage: () => cy.get('[data-cy="ukef-deal-stage-value"]'),
+  successBanner: () => cy.get('[data-cy="success-banner"]'),
```
ttbarnes, line 13: Are we able to move `successBanner` into `partials` please? I can only see one other instance of `successBanner` in `unissuedFacilityTable`, so could remove that as well

trade-finance-manager-api/src/v1/rest-mappings/mappings/gef-facilities/mapGefFacilitySnapshot.js
```diff
@@ -44,7 +44,7 @@ const mapGefFacilitySnapshot = (facility, dealSnapshot) => {
 
   facilitySnapshot.facilityProduct = mapFacilityProduct(type);
 
-  facilitySnapshot.facilityStage = mapGefFacilityStage(hasBeenIssued);
+  facilitySnapshot.facilityStage = mapGefFacilityStage(hasBeenIssued, facilityTfm?.facilityStage);
```
ttbarnes, line 47: Bit concerning if everything was passing before without this change 🤔

trade-finance-manager-ui/server/controllers/case/cancellation/check-details.controller.post.test.ts
```diff
@@ -82,12 +87,31 @@ describe('postBankRequestDate', () => {
       );
     });
 
+    it('adds a successMessage to flash storage', async () => {
```
ttbarnes, line 90:
```suggestion
    it('adds a successMessage to req.flash' async () => {
```

trade-finance-manager-ui/server/controllers/case/cancellation/check-details.controller.ts
```diff
@@ -79,6 +79,10 @@ export const postDealCancellationDetails = async (req: PostDealCancellationDetai
 
     await api.submitDealCancellation(_id, { reason, bankRequestDate, effectiveFrom }, userToken);
 
+    const { ukefDealId } = deal.dealSnapshot.details;
+
+    req.flash('successMessage', `Deal ${ukefDealId} cancelled`);
```
ttbarnes, line 84: What does `api.submitDealCancellation` return please? I'm wondering if `api.submitDealCancellation` fails, does the flash still occur?

trade-finance-manager-ui/server/controllers/case/index.test.js
```diff
@@ -87,6 +94,7 @@ describe('controllers - case', () => {
           tfm: mockDeal.tfm,
           activePrimaryNavigation: 'manage work',
           activeSubNavigation: 'deal',
+          successMessage: mockSuccessBannerMessage,
```
ttbarnes, line 97: Could this be simpler perhaps? i.e:

```ts
successMessage: getFlashSuccessMessage(mockReq),
```

trade-finance-manager-ui/server/helpers/getFlashSuccessMessage.js
ttbarnes: Would be awesome to add some documentation to both functions whilst we're here 🙏

ttbarnes (changes requested):
Nice work, few minor comments

trade-finance-manager-api/src/v1/rest-mappings/mappings/gef-facilities/mapGefFacilitySnapshot.js
BethThomas141: Agreed - it was why I was keen to have this unit test in previously https://github.com/UK-Export-Finance/dtfs2/pull/3770/files/179736077bd2ddc592ab1f8f5cf9b624b869049e#r1816714044 - do you reckon I add it back in?

BethThomas141: Seperately, it was passing E2E tests as this only impacted non-Gef facilities, and (concerningly) a lot of our E2E tests just check non-GEF facilities. Possibly we should duplicate our E2E deal cancellation tests for GEF facilities? 🤷‍♀️

trade-finance-manager-ui/server/controllers/case/cancellation/check-details.controller.ts
BethThomas141: It throws an error, which means here it will skip the `req.flash` section of code and go straight to the error handling block in this file

trade-finance-manager-api/src/v1/rest-mappings/mappings/gef-facilities/mapGefFacilitySnapshot.js
ttbarnes: > Seperately, it was passing E2E tests as this only impacted non-Gef facilities, and (concerningly) a lot of our E2E tests just check non-GEF facilities. Possibly we should duplicate our E2E deal cancellation tests for GEF facilities? 🤷‍♀️

Highly recommend adding more tests for non-GEF facilities - sounds like most likely this would have affected non-deal cancellation related tests/scenarios/journeys 😢 

Can we create a ticket in the backlog please?

trade-finance-manager-ui/server/controllers/case/cancellation/check-details.controller.ts
ttbarnes: Brilliant thanks @BethThomas141

trade-finance-manager-ui/server/controllers/case/index.test.js
BethThomas141: I've made it a tad simpler, have a look and see what you think :) I quite like keeping it as checking the expected string gets passed in, as opposed to calling a function - I think it's easier to read the test as and see what object we're expecting without having to work out what `getFlashSuccessMessage` does.

trade-finance-manager-ui/server/controllers/case/cancellation/check-details.controller.post.test.ts
BethThomas141: Changed

trade-finance-manager-api/src/v1/rest-mappings/mappings/gef-facilities/mapGefFacilitySnapshot.js
BethThomas141: Good idea, I've refactored both `mapGefFacilitySnapshot` and `mapFacilitySnapshot` tests to do this now. Take a look and let me know what you think!
And yes sure, I'll create one

trade-finance-manager-api/src/v1/rest-mappings/mappings/facilities/mapFacilitySnapshot.test.js
```diff
@@ -86,8 +86,6 @@ describe('mapFacility', () => {
 
     const expectedFacilityValueExportCurrency = `${mockCurrency.id} ${formattedFacilityValue}`;
 
-    const facilityStage = mapBssEwcsFacilityStage(mockFacilityStage, mockFacilityTfm.facilityStage);
```
MarRobSoftwire, line 89: How come this has been removed?

trade-finance-manager-ui/server/helpers/getFlashSuccessMessage.js
MarRobSoftwire: Alternatively what do we think about putting it in libs/common? its very front-end oriented but I think thats ok?

trade-finance-manager-ui/templates/case/case.njk
```diff
@@ -20,6 +21,10 @@
     {{caseSummary.render(componentData)}}
   {% endif %}
 
+  {% if successMessage %}
+   {{ successBanner.render(message = successMessage) }}
```
MarRobSoftwire, line 25: I always forget how nunjucks macros work - can we simplify this? If this is a stupid suggestion ignore & resolve :) 

```suggestion
   {{ successBanner.render(successMessage) }}
```

trade-finance-manager-ui/package.json
```diff
@@ -37,14 +37,15 @@
     "@azure/msal-node": "^2.15.0",
     "@babel/polyfill": "^7.12.1",
     "@ministryofjustice/frontend": "1.8.0",
+    "@types/connect-flash": "^0.0.40",
```
MarRobSoftwire, line 40: This should be a dev dependency

trade-finance-manager-api/src/v1/rest-mappings/mappings/facilities/mapFacilitySnapshot.test.js
BethThomas141: Removed following a discussion with Tony on a separate thread about the unit tests for this file. But as a consequence of that discussion, I've realised that this is not needed at all (either in the test or in the original `mapFacilitySnapshot` file): 

- In `mapFacilitySnapshot` we clone the facility snapshot and add additional properties to the cloned object. This `clonedFacility` is then passed into several functions where the fields on the cloned facility are used and accessed. However `facilityStage` is not used / accessed in any of these functions, it's only used as the value for the new `facilityStage` field on the `mappedFacilitySnapshot`. So I've refactored here to calculate the `facilityStage` just when it's needed on the mappedSnapshot & not add it as a temporary property to the cloned snapshot. 

Not sure if that's a clear explanation or not, happy to hop on a call to show you!

trade-finance-manager-ui/server/helpers/getFlashSuccessMessage.js
BethThomas141: Documentation added! I'm going to leave it in their respective directories if that's ok @MarRobSoftwire - if I move it to libs/common we'd want the helper in TS which would mean installing extra dev dependencies with the `flash` types in libs/common

trade-finance-manager-ui/templates/case/case.njk
BethThomas141: It can, good spot! For some reason I thought nunjucks didn't like that, but seems all good!

trade-finance-manager-ui/package.json
BethThomas141: Good spot, have moved thanks!

e2e-tests/tfm/cypress/e2e/pages/caseDealPage.js
BethThomas141: I can do, which `partials` file do you expect it to live in? `unissuedFacilityTable` is in GEF as opposed to TFM-UI so we'd have to make sure it's a partial file that both services use. Possibly we need a new file, but then it would just have `successBanner` in it and nothing else, and not sure what to call it - what do you think? 

image

ttbarnes: I was just thinking in the index, where we have backLink, cancelLink, errorSummary etc 

```
e2e-tests/tfm/cypress/e2e/partials/index.js
```

trade-finance-manager-ui/server/helpers/getFlashSuccessMessage.js
BethThomas141: Resolving as docs added👍

trade-finance-manager-ui/server/controllers/case/cancellation/check-details.controller.post.test.ts
```diff
@@ -82,12 +87,31 @@ describe('postBankRequestDate', () => {
...
+        session,
+        body: { reason, bankRequestDate, effectiveFrom },
+        flash: flashMock,
+      });
+
+      // Act
+      await postDealCancellationDetails(req, res);
+
+      // Assert
+      expect(flashMock).toHaveBeenCalledWith('successMessage', `Deal ${ukefDealId} cancelled`);
```
ttbarnes, line 105: Can we add a `toHaveBeenCalledTimes` test please

e2e-tests/tfm/cypress/e2e/pages/caseDealPage.js
BethThomas141: Got it, good idea thanks. I've moved that over. I won't change the other instances in this PR in `unissuedFacilityTable` as that's used in a lot of touched files so will show up as a big diff in this PR.

ttbarnes (approved):

PR #3782, opened 2024-10-29
feat(DTFS2-7578): clone gef deals to v1

MarRobSoftwire:
## Introduction :pencil2:
When GEF_DEAL_VERSION is set to 1, a version 0 deal should be cloned to this version. Version 1 deals should also have the facility end date values wiped (https://ukef-dtfs.atlassian.net/browse/DTFS2-7579)

- set deal version to default version when cloning
- set facility end date values to null on cloned facilities
- Add e2e tests for v0 -> v1 and v1 -> v1

- Separate out cloning logic into service methods & refactor
- Add unit test coverage
- Further additions to `Facility` and `Deal` mongodb models
- Add linting rule for `no-param-reassign`


e2e-tests/gef/cypress/e2e/clone/clone-gef-deal.spec.js
```diff
@@ -1,5 +1,5 @@
 import relative from '../relativeURL';
-import { form, mainHeading, submitButton } from '../partials';
```
MarRobSoftwire, line 2: I dont really like this `form` - we should probably replace this with `cy.clickFooButton`

portal-api/.eslintrc.js
```diff
@@ -41,6 +41,7 @@ const baseRules = {
       consistent: true,
     },
   ],
+  'no-param-reassign': ['error', { props: true, ignorePropertyModificationsFor: ['draft', 'req'] }],
```
MarRobSoftwire, line 44: in order to use immer this is needed

.github/workflows/test.yml
```diff
@@ -538,7 +538,7 @@ jobs:
             '*.spec.js',
             'application/**/*.spec.js',
             'checker/**/*.spec.js',
-            'clone/**/*.spec.js',
+            'clone/*.spec.js',
```
AlexBramhill, line 541: are we happy with this only getting the tests on the root of this folder

e2e-tests/gef/cypress/e2e/clone/clone-gef-deal.spec.js
AlexBramhill: I think this is a wider question -- so nothing to change in this PR. 

It's whether we want to differentiate between submitting a form and just clicking continue button (either way we should look to follow the `cy.` pattern)... I'm of no strong opinion on this

libs/common/src/types/mongo-db-models/deals.ts
```diff
@@ -8,9 +8,11 @@ type BssEwcsDeal = AnyObject & {
   };
 };
 
-type GefDeal = AnyObject & {
+export type GefDeal = AnyObject & {
   dealType: typeof DEAL_TYPE.GEF;
-  ukefDealId: string;
+  ukefDealId: string | null;
+  eligibility: AnyObject;
+  exporter: AnyObject;
```
AlexBramhill, line 15: Good job, this slow build out of these models as we understand them more is exactly what we need

libs/common/src/types/mongo-db-models/facility.ts
```diff
@@ -39,7 +39,7 @@ export type Facility = {
   shouldCoverStartOnSubmission: boolean;
   coverStartDate: UnixTimestampString | Date | null;
   coverEndDate: UnixTimestampString | Date | null;
-  issueDate: UnixTimestampString | Date | null;
+  issueDate?: UnixTimestampString | Date | null;
```
AlexBramhill, line 42: This can be undefined and null? (same with hasBeenIssuedAndAcknowledged?)

portal-api/.eslintrc.js
AlexBramhill: I agree here, although I know its been taken out before -- Once we have sign off that this is good, could we keep the other `.eslintrc`s updated with this change please.

The alternative for this is a lot of eslint ignores which are not good for the code health

portal-api/src/v1/gef/services/clone-deal.service.ts
```diff
@@ -0,0 +1,118 @@
...
+ * @param params.bankInternalRefName - the new deals internal ref name
+ * @param params.additionalRefName - new new deals additional ref name
+ * @param params.maker - the user making the clone
+ * @param params.userId - the makers id
+ * @param params.bank - the users bank
+ * @param params.auditDetails - the users audit details
+
+ * @returns the inserted deal id
+ */
+export const cloneDeal = async ({
```
AlexBramhill, line 54: `cloneDealToLatestVersion`?

AlexBramhill (commented):
A quick pass with a few comments

.github/workflows/test.yml
MarRobSoftwire: I think so - there arent any fed-disabled tests here (I could add a test that v0 goes to v0 but feels unnecessary given the FF is being removed soon)

libs/common/src/types/mongo-db-models/facility.ts
MarRobSoftwire: Ooh thats a good point - I think it might always be null (on GEF deals at least, on BSS/EWCS it might be undefined)

e2e-tests/gef/cypress/e2e/clone/facility-end-date-enabled/clone-v0-to-v1.spec.js
```diff
@@ -0,0 +1,86 @@
...
+      applicationDetails.addContingentFacilityButton().click();
+      facilities.hasBeenIssuedRadioYesRadioButton().click();
+      cy.clickContinueButton();
+
+      cy.visit(relative(`/gef/application-details/${version0DealId}`));
+
+      cloneGEFDeal.cloneGefDealLink().click();
+      mandatoryCriteria.trueRadio().click();
+
+      cy.clickContinueButton();
```
MarRobSoftwire, line 43: The `clone-gef-deal.spec` tests use `form().submit();` rather than these commands - @ttbarnes should these be replaced everywhere? I dont really like that this manipulates the `form` element directly rather than clicking a button on screen

ttbarnes: Wowsa, didn't realise, sorry I didn't spot this before.

I completely agree, can we replace all instaces of `form().submit()` please, there are 25x instances in 4x files

MarRobSoftwire: Have made a ticket https://ukef-dtfs.atlassian.net/browse/DTFS2-7589 but I probably wont have time this sprint to take a look at this.

portal-api/src/v1/gef/controllers/clone-gef-deal.controller.js
```diff
@@ -1,182 +1,46 @@
...
-    bankInternalRefName,
-    additionalRefName,
-    maker: req.user,
-    userId,
-    bank,
-    auditDetails,
-  });
+    return res.status(200).send({ dealId: insertedId });
+  } catch (error) {
+    console.error('Failed to clone deal, %o', error);
```
ttbarnes, line 38: Are we able to add a `dealId` in here please

portal-api/src/v1/gef/services/clone-deal.service.test.ts
```diff
@@ -0,0 +1,210 @@
...
+  },
+  auditRecord: {},
+  ukefDecision: 'ukefDecision',
+  ukefDecisionAccepted: 'ukefDecisionAccepted',
+  checkerMIN: 'checkerMIN',
+  manualInclusionNoticeSubmissionDate: 'manualInclusionNoticeSubmissionDate',
+  comments: 'comments',
+  previousStatus: 'previousStatus',
+  dataMigration: 'dataMigration',
+};
```
ttbarnes, line 66: Can we use a mock/fixture? Rather than ~30 lines of mocks in the test.

portal-api/src/v1/gef/services/clone-deal.service.test.ts
```diff
@@ -0,0 +1,210 @@
...
+      maker,
+      bank: mockBank,
+      ukefDealId: null,
+      checkerId: null,
+      editedBy: [maker._id],
+      exporter: cloneExporter(existingDeal.exporter),
+      portalActivities: [],
+      supportingInformation: {},
+      clonedDealId: existingDealId,
+      auditRecord: generateAuditDatabaseRecordFromAuditDetails(mockAuditDetails),
```
ttbarnes, line 207: Can we use a fixture and then assert only the expected changed properties?

portal-api/src/v1/gef/services/clone-deal.service.ts
```diff
@@ -0,0 +1,118 @@
...
+  bankInternalRefName: string;
+  additionalRefName: string;
+  maker: PortalSessionUser;
+  userId: string | ObjectId;
+  bank: Bank;
+  auditDetails: AuditDetails;
+};
+
+/**
+ * Creates a copy of existing GEF deal & inserts a clone into the database, with certain values updated/removed.
```
ttbarnes, line 41:
```suggestion
 * Creates a copy of existing GEF deal.
 * Inserts a clone into the database as the latest version.
 * Updates/removes certain values.
```

portal-api/src/v1/gef/services/clone-deal.service.ts
```diff
@@ -0,0 +1,118 @@
...
+    draft.bank = bank;
+    draft.ukefDealId = null;
+    draft.checkerId = null;
+    draft.editedBy = [userId];
+    draft.exporter = cloneExporter(draft.exporter);
+    draft.portalActivities = [];
+    draft.supportingInformation = {};
+    draft.clonedDealId = dealId;
+    draft.auditRecord = generateAuditDatabaseRecordFromAuditDetails(auditDetails);
+  });
```
ttbarnes, line 111: Can we move this into it's own helper/function please maybe mentioning "sanitised"

portal-api/src/v1/gef/services/clone-exporter.service.ts
```diff
@@ -0,0 +1,16 @@
+import { AnyObject } from '@ukef/dtfs2-common';
+import { exporterStatus } from '../controllers/validation/exporter';
+
+/**
+ * Clones the exporter
+ * @param currentExporter the existing exporter
```
ttbarnes, line 6: current vs existing, can we use just one of the terms

portal-api/src/v1/gef/services/clone-exporter.service.ts
```diff
@@ -0,0 +1,16 @@
...
+ * @returns the exporter with updated at set to now
```
ttbarnes, line 7:
```suggestion
 * @returns the exporter with updatedAt set to now
```

portal-api/src/v1/gef/services/clone-facilities.service.test.ts
```diff
@@ -0,0 +1,200 @@
...
+  ukefFacilityId: '10000012',
+  feeType: 'In advance',
+  feeFrequency: 'Monthly',
+  dayCountBasis: 365,
+  coverDateConfirmed: true,
+  canResubmitIssuedFacilities: null,
+  hasBeenIssuedAndAcknowledged: null,
+  issueDate: null,
+};
+
```
ttbarnes, line 91: Same as previous comments - can we use mocks/fixtures?

portal-api/src/v1/gef/services/clone-facilities.service.ts
```diff
@@ -0,0 +1,61 @@
...
+    .find({
+      dealId: { $eq: new ObjectId(currentDealId) },
+    })
+    .toArray()) as Partial<Facility>[];
+
+  if (!existingFacilities.length) {
+    return;
+  }
+
+  const clonedFacilities = existingFacilities.map((facility) => {
```
ttbarnes, line 29: Same as the deal level cloning, can this be in its own file/function please and mention "sanitise"?

portal-api/src/v1/users/transform-database-user.js
```diff
@@ -1,4 +1,3 @@
-/* eslint-disable no-param-reassign */
```
ttbarnes, line 1: yay

e2e-tests/gef/cypress/e2e/clone/facility-end-date-enabled/clone-v0-to-v1.spec.js
```diff
@@ -0,0 +1,84 @@
...
+import mandatoryCriteria from '../../pages/mandatory-criteria';
+import nameApplication from '../../pages/name-application';
+import applicationDetails from '../../pages/application-details';
+import facilities from '../../pages/facilities';
+import { submitButton } from '../../partials';
+import aboutFacilityUnissued from '../../pages/unissued-facilities-about-facility';
+
+/**
+ * NOTE: These tests check the backwards compatibility with in-flight version 0 deals.
+ * A migration may need to be run on production if this test is updated.
```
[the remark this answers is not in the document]
MarRobSoftwire: The idea is this protects us against breaking v0 functionality. If this test stops working we should double check this isnt because our version 0 deals need updating. For this test in particular its unlikely to cause issues, but I've added this comment to all tests concerning v0 deals

e2e-tests/gef/cypress/e2e/clone/clone-gef-deal.spec.js
```diff
@@ -111,7 +111,6 @@ context('Clone GEF (AIN) deal', () => {
 
         applicationDetails.bankRefName().contains('Cloned AIN deal');
         applicationDetails.automaticCoverStatus().contains('Not started');
-        applicationDetails.facilityStatus().contains('Completed');
```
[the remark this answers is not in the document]
MarRobSoftwire: The facility status wont be `Completed` since the FED values are removed (when cloning to version 0) - It also assumes that the cover start date wasnt removed too (which will happen if it was in the past)

portal-api/src/v1/users/transform-database-user.js
MarRobSoftwire: I think there will be lots of other places this can be done now I've update the linting service-wide. That said I dontthink its part of this PR

portal-api/src/v1/gef/services/clone-deal.service.test.ts
MarRobSoftwire: I thought a lot about this & I dont think it's ideal to do that - the tests overwrite ~25 properties on the deals so we really care about what value they were initially and what value they end up being.

I think if this is hidden by `const existingDeal = MOCK_GEF_DEAL`  then its a bit less intentional that we want, for example, the `dataMigration` value to be removed.

If I went down the route of
```
const existingDeal = {
  ...MOCK_GEF_DEAL,
  dataMigration: 'dataMigration',
  // etc...
}
```
Then I'd still need to add 25 values to check that they are being updated/overwritten

I also cant actually see any mock gef deal constants or helpers in portal-api

portal-api/src/v1/gef/services/clone-deal.service.test.ts
MarRobSoftwire: Similiar to above, about 25 properties are changed

e2e-tests/gef/cypress/e2e/clone/clone-gef-deal.spec.js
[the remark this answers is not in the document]
ttbarnes: Thanks - can we replace with the expected status i.e `IN_PROGRESS`?

e2e-tests/gef/cypress/e2e/clone/facility-end-date-enabled/clone-v0-to-v1.spec.js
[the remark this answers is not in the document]
ttbarnes: Thanks @MarRobSoftwire I still find the comment a bit confusing though, could we say something like this instead, what do you think?

```js
/**
 * NOTE: These tests check the backwards compatibility with in-flight version 0 deals.
 * If migration is updated for version 0 deals/facilities
 * and therefore this test is updated,
 * we need to check if migration needs to be run in production.
 */
```

portal-api/src/v1/gef/services/clone-deal.service.ts
ttbarnes: Agree, thanks

portal-api/src/v1/users/transform-database-user.js
ttbarnes: Brilliant, would be great to create a ticket for this

portal-api/src/v1/gef/services/clone-deal.service.test.ts
```diff
@@ -0,0 +1,245 @@
...
+        editedBy: [maker._id],
+        exporter: cloneExporter(existingDeal.exporter),
+        portalActivities: [],
+        supportingInformation: {},
+        clonedDealId: existingDealId,
+        auditRecord: generateAuditDatabaseRecordFromAuditDetails(mockAuditDetails),
+      });
+    });
+  });
+});
```
ttbarnes, line 245: There's lots going on in the service, can we split into new directories? E.g

```ts
clone-deal.service/index.ts
clone-deal.service/generate-cloned-deal.ts
```

Concerned if either of the functions/tests grow

portal-api/src/v1/gef/services/clone-deal.service.test.ts
```diff
@@ -0,0 +1,245 @@
...
+        }),
+        _id: expect.any(ObjectId) as ObjectId,
+      });
+    });
+  });
+
+  describe('generateCloneDeal', () => {
+    it('returns the expected clone deal', () => {
+      // Act
+      const cloneDeal = generateCloneDeal({
```
ttbarnes, line 202:
```suggestion
      const result = generateCloneDeal({
```

portal-api/src/v1/gef/services/clone-deal.service.test.ts
```diff
@@ -0,0 +1,245 @@
...
+        maker,
+        bank: mockBank,
+        ukefDealId: null,
+        checkerId: null,
+        editedBy: [maker._id],
+        exporter: cloneExporter(existingDeal.exporter),
+        portalActivities: [],
+        supportingInformation: {},
+        clonedDealId: existingDealId,
+        auditRecord: generateAuditDatabaseRecordFromAuditDetails(mockAuditDetails),
```
ttbarnes, line 241: The original deal has the properties listed in `propertiesToRemove` - I know the test is checking the entire returned deal object, but could we add another test to be clear that XYZ fields should be null? e.g

```ts
it('should NOT include fields from propertiesToRemove', () => {
  expect(result.x).toBeBeUndefined();
  expect(result.y).toBeBeUndefined();
});
```

portal-api/src/v1/gef/services/clone-deal.service.ts
```diff
@@ -0,0 +1,170 @@
...
+  DealNotFoundError,
+  AnyObject,
+} from '@ukef/dtfs2-common';
+import { generateAuditDatabaseRecordFromAuditDetails } from '@ukef/dtfs2-common/change-stream';
+import { mongoDbClient } from '../../../drivers/db-client';
+import { getLatestEligibilityCriteria } from '../controllers/eligibilityCriteria.controller';
+import { DEAL } from '../../../constants';
+import { cloneExporter } from './clone-exporter.service';
+
+const propertiesToRemove: (keyof GefDeal)[] = [
```
ttbarnes, line 21: I think we should add a test for this - currently, it looks like if a field is removed from here, the tests will pass

e2e-tests/gef/cypress/e2e/clone/facility-end-date-enabled/clone-v1-to-v1.spec.js
```diff
@@ -0,0 +1,138 @@
...
+import applicationDetails from '../../pages/application-details';
+import { submitButton } from '../../partials';
+import aboutFacilityUnissued from '../../pages/unissued-facilities-about-facility';
+import facilityEndDate from '../../pages/facility-end-date';
+import bankReviewDate from '../../pages/bank-review-date';
+import { MOCK_APPLICATION_AIN } from '../../../fixtures/mocks/mock-deals';
+import { anUnissuedCashFacility, anIssuedCashFacility } from '../../../../../e2e-fixtures/mock-gef-facilities';
+import { DEAL_STATUS, FACILITY_TYPE } from '../../../fixtures/constants';
+
+context('Clone version 1 deal to version 1', () => {
```
ttbarnes, line 15: How/why is it possible to clone a version 1 deal to the same version?

e2e-tests/gef/cypress/e2e/clone/facility-end-date-enabled/clone-v0-to-v1.spec.js
```diff
@@ -0,0 +1,109 @@
...
+
+    it('bank review date page does not show pre-populated values', () => {
+      applicationDetails.facilitySummaryListTable(0).bankReviewDateAction().click();
+
+      bankReviewDate.bankReviewDateDay().should('have.value', '');
+      bankReviewDate.bankReviewDateMonth().should('have.value', '');
+      bankReviewDate.bankReviewDateYear().should('have.value', '');
+    });
+  });
+});
```
ttbarnes, line 109: Are we able to add an E2E test to check that eligibility has been updated to the latest version?

It would be ideal to check that all properties modified during the cloning are as expected in E2E

portal-api/src/v1/gef/services/clone-exporter.service.test.ts
```diff
@@ -0,0 +1,31 @@
...
+  it('updates the updatedAt and status fields', () => {
+    // Arrange
+    const mockExporter = {
+      updatedAt: 17090930940932,
+      status: 'An exporter status',
+      anotherField: 'anotherField',
+    };
+
+    // Act
+    const response = cloneExporter(mockExporter);
```
ttbarnes, line 22:
```suggestion
    const result = cloneExporter(mockExporter);
```

portal-api/src/v1/gef/services/clone-facilities.service.test.ts
ttbarnes: It looks like facilities are very similar with the only difference being `name` and `hasBeenIssued`. Can we create a DRY `mockGenericFacility` / `mockBaseFacility` or similar?

portal-api/src/v1/gef/services/clone-deal.service.test.ts
ttbarnes: Thanks @MarRobSoftwire , seems it makes sense on the deal level at least

e2e-tests/gef/cypress/e2e/clone/facility-end-date-enabled/clone-v1-to-v1.spec.js
MarRobSoftwire: If `GEF_DEAL_VERSION=1`, the user creates a deal & then clones it. The deal version will be 1 for both.

Cloning is a functionality for makers, deal versioning behaves like a feature flag but keeping compatibility with existing in-flight deals

portal-api/src/v1/users/transform-database-user.js
MarRobSoftwire: https://ukef-dtfs.atlassian.net/browse/DTFS2-7603 done

e2e-tests/gef/cypress/e2e/clone/facility-end-date-enabled/clone-v0-to-v1.spec.js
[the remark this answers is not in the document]
ttbarnes: Sorry @MarRobSoftwire I like the new comment but I'm finding it a bit strange, unless i'm missing a point:

```js
/**
 * NOTE: These tests check the backwards compatibility with in-flight version 0 deals.
 * If this test needs updating because version 0 deals are no longer being supported,
 * we need to check if migration needs to be run in production.
 */
```

This reads to me that it's "If we update the test because version 0 is no longer supported ... we need to check production migration"

But if version 0 is no longer supported, the test would just be removed, no?

I feel it should say "if migration is updated (and therefore this test is updated), we need to check production migration"

portal-api/src/v1/gef/services/clone-deal.service.test.ts
ttbarnes: Bump @MarRobSoftwire can't see a test for this?

e2e-tests/gef/cypress/e2e/clone/facility-end-date-enabled/clone-v0-to-v1.spec.js
MarRobSoftwire: I'm not exactly sure how I'd check that? Moreover it would fall under the existing `e2e-tests\gef\cypress\e2e\clone\clone-gef-deal.spec.js` test.

portal-api/src/v1/gef/services/clone-deal.service.test.ts
MarRobSoftwire: Its in portal-api\src\v1\gef\services\clone\helpers\generate-cloned-deal.test.ts now :)

e2e-tests/gef/cypress/e2e/clone/facility-end-date-enabled/clone-v0-to-v1.spec.js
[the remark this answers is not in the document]
MarRobSoftwire: We're using mongodb & dont have automatic migrations that would be 'updated' - maybe theres a better word to make it clearer?

Thoughts on
``` 
 * If this test needs updating then version 0 deals may no longer be supported and
 * we need to update these deals in production.
```

e2e-tests/gef/cypress/e2e/clone/facility-end-date-enabled/clone-v0-to-v1.spec.js
ttbarnes: Unless i'm missing something, it looks like we don't have full E2E coverage for GEF eligibility criiteria I.E criteria 5 should render X text 😞 could improve in the future.

Can we add some coverage for the other cloned fields though please i.e deal name etc?

portal-api/src/v1/gef/services/clone/helpers/generate-cloned-deal.test.ts
```diff
@@ -0,0 +1,131 @@
...
+    information: 'information',
+  },
+  auditRecord: {},
+  ukefDecision: 'ukefDecision',
+  ukefDecisionAccepted: 'ukefDecisionAccepted',
+  checkerMIN: 'checkerMIN',
+  manualInclusionNoticeSubmissionDate: 'manualInclusionNoticeSubmissionDate',
+  comments: 'comments',
+  previousStatus: 'previousStatus',
+  dataMigration: 'dataMigration',
```
ttbarnes, line 51: Looks like this should be an object?

portal-api/src/v1/gef/services/clone/helpers/generate-cloned-deal.ts
```diff
@@ -0,0 +1,85 @@
+import { ObjectId } from 'mongodb';
+import { produce } from 'immer';
+import { AuditDetails, Bank, getCurrentGefDealVersion, PortalSessionUser, GefDeal, AnyObject } from '@ukef/dtfs2-common';
+import { generateAuditDatabaseRecordFromAuditDetails } from '@ukef/dtfs2-common/change-stream';
+import { DEAL } from '../../../../../constants';
+import { generateClonedExporter } from './generate-cloned-exporter';
+
+const propertiesToRemove: (keyof GefDeal)[] = [
```
ttbarnes, line 8: Can we export this and add a unit test please?

e2e-tests/gef/cypress/e2e/clone/facility-end-date-enabled/clone-v0-to-v1.spec.js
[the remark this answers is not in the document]
ttbarnes: Looks good @MarRobSoftwire - could/should we add the same kind of comment in non-E2E level code?

portal-api/src/v1/gef/services/clone/clone-facilities.service.test.ts
```diff
@@ -0,0 +1,143 @@
...
+  name: 'Facility one',
+  coverStartDate: new Date(1638403200000),
+  ukefFacilityId: '10000011',
+  coverDateConfirmed: null,
+};
+
+const unissuedFacility: Facility = {
+  ...facilityDefaults,
+  _id: new ObjectId(),
+  dealId: new ObjectId(existingDealId),
```
ttbarnes, line 61: Can we move the `dealId` into `facilityDefaults`?

portal-api/src/v1/gef/services/clone/helpers/generate-cloned-facility.test.ts
```diff
@@ -0,0 +1,133 @@
...
+  dayCountBasis: 365,
+  canResubmitIssuedFacilities: null,
+  hasBeenIssuedAndAcknowledged: null,
+  issueDate: null,
+};
+
+const issuedFacility: Facility = {
+  ...facilityDefaults,
+  _id: new ObjectId(),
+  dealId: new ObjectId(existingDealId),
```
ttbarnes, line 48: Can we move `dealId` into `facilityDefaults` please

portal-api/src/v1/gef/services/clone/helpers/generate-cloned-facility.ts
```diff
@@ -0,0 +1,52 @@
...
+    draft.auditRecord = generateAuditDatabaseRecordFromAuditDetails(auditDetails);
+
+    draft.isUsingFacilityEndDate = null;
+    draft.facilityEndDate = null;
+    draft.bankReviewDate = null;
+
+    const currentTime = new Date();
+    currentTime.setHours(0, 0, 0, 0);
+
+    if (draft.coverStartDate) {
```
ttbarnes, line 47: Can we add a comment please as to why we need to conditionally set the date to null?

portal-api/src/v1/gef/services/clone/helpers/generate-cloned-facility.ts
```diff
@@ -0,0 +1,52 @@
...
+      if (isBefore(new Date(draft.coverStartDate), startOfDay(new Date()))) {
+        draft.coverStartDate = null;
+      }
```
ttbarnes, line 50: I think something this would be easier to read

```ts
if (draft.coverStartDate) {
  const coverStartDate = new Date(draft.coverStartDate);
  const today = startOfDay(new Date());

  if (isBefore(coverStartDate, today)) {
    draft.coverStartDate = null;
  }
}
```

portal-api/src/v1/gef/services/clone/helpers/generate-cloned-deal.test.ts
MarRobSoftwire: I dont think these types actually matter (I've taken a similar attitude for all the values that are deleted) since its a unit test. but will update this

portal-api/src/v1/gef/services/clone/helpers/generate-cloned-deal.ts
MarRobSoftwire: This is a constant - what sort of unit test are you expecting? These are tested by `it('removes the expected properties', () => {`

ttbarnes: Concern here is that a property could be added here and the tests would still pass. If we test this specifically, it'll fail if a property is added.

e2e-tests/gef/cypress/e2e/clone/facility-end-date-enabled/clone-v0-to-v1.spec.js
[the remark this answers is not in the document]
MarRobSoftwire: I'll add it to the gef-deal-versioning file

e2e-tests/gef/cypress/e2e/clone/facility-end-date-enabled/clone-v0-to-v1.spec.js
ttbarnes: Bump, can we make sure we have tests for as many fields as possible (that are modified during cloning)?

https://github.com/UK-Export-Finance/dtfs2/pull/3782/files/1403404a951cb4617c6164e29e2c3cea47f2d415#diff-a467928018f518e2c357f605eab63127ac4b8e9d7e04dc1c8a2989a45d02cd13R66

portal-api/src/v1/gef/services/clone/helpers/generate-cloned-deal.ts
ttbarnes: Bump

e2e-tests/gef/cypress/e2e/clone/facility-end-date-enabled/clone-v0-to-v1.spec.js
MarRobSoftwire: There's no real existing tests at all for the actual ECs that are used in a deal thats just been created, nevermind cloned.

(e.g e2e-tests\gef\cypress\e2e\eligibility\automatic-cover.spec.js just checks validation, e2e-tests\gef\cypress\e2e\eligibility\eligible-automatic-cover.spec.js doesnt check much at all)

Testing this in `clone-gef-deal.spec.js` would hugely increase the size of the PR needing me to work out what values should be visible & how to view them & then adding data-cy and new selectors to find these values on the screen. I don't think thats really within the scope of this ticket?

ttbarnes: OK, not ideal - can we create a ticket in the backlog please to add this E2E coverage?

We should also have at least 1x test to ensure that a deal can be submitted after cloning.

e2e-tests/gef/cypress/e2e/clone/facility-end-date-enabled/clone-v0-to-v1.spec.js
```diff
@@ -0,0 +1,110 @@
...
+
+      cloneGEFDeal.cloneGefDealLink().click();
+      mandatoryCriteria.trueRadio().click();
+
+      cy.clickContinueButton();
+
+      cy.keyboardInput(nameApplication.internalRef(), 'Cloned of v0 deal');
+      cy.clickContinueButton();
+
+      cy.get('table.govuk-table tr').eq(1).find('td').eq(1).find('.govuk-link').click();
```
ttbarnes, line 51: What link is this, are we able to use a selector instead? Looks like this was added in `clone-gef-deal.spec.js` and then copied over to these new E2E tests

e2e-tests/gef/cypress/e2e/clone/facility-end-date-enabled/clone-v0-to-v1.spec.js
MarRobSoftwire: Done https://ukef-dtfs.atlassian.net/browse/DTFS2-7610

e2e-tests/gef/cypress/e2e/clone/facility-end-date-enabled/clone-v0-to-v1.spec.js
MarRobSoftwire: I copied this across & couldnt find a good alternative. Its pretty niche and only needed when cloning a deal.

Basically this gets the first deal from the deals page. In portal tests there's some additional selectors (`e2e-tests\portal\cypress\e2e\pages\dashboardDeals.js`) but you need to know the id in advance - e.g.
```
    link: (id) => cy.get(`[data-cy="deal__link--${id}"]`),
```

I could add something to the sparse `dashboad.js` page in gef
```
  facility: (row) => ({
    link: () => cy.get('table.govuk-table tr').eq(1).find('td').eq(row).find('.govuk-link').click(),
  }),
```
It probably needs some tweaking & it still has to use the slightly ugly `.eq` and `.find` syntax because the data-cy selector contains the deal id.

libs/common/src/types/mongo-db-models/deals.ts
```diff
@@ -8,9 +8,11 @@ type BssEwcsDeal = AnyObject & {
   };
 };
 
-type GefDeal = AnyObject & {
+export type GefDeal = AnyObject & {
   dealType: typeof DEAL_TYPE.GEF;
-  ukefDealId: string;
+  ukefDealId: string | null;
```
Zainzzkk, line 13: `ukefDealId?` will not work right?

portal-api/src/v1/gef/controllers/clone-gef-deal.controller.js
```diff
@@ -1,182 +1,46 @@
...
-  const response = await cloneDeal({
-    dealId: existingDealId,
-    bankInternalRefName,
-    additionalRefName,
-    maker: req.user,
-    userId,
-    bank,
-    auditDetails,
-  });
+    return res.status(200).send({ dealId: insertedId });
```
Zainzzkk, line 36: could we use HttpStatusCode here? for the 200 and 422?

portal-api/src/v1/gef/services/clone/helpers/generate-cloned-deal.test.ts
```diff
@@ -0,0 +1,146 @@
...
+const existingDealId = new ObjectId().toString();
+const mockAuditDetails = generateSystemAuditDetails();
+const mockBank = {
+  id: 'bank id',
+} as Bank;
+const bankInternalRefName = 'bankInternalRefName';
+const additionalRefName = 'additionalRefName';
+const maker = aPortalSessionUser();
+
+const existingDeal: Deal = {
```
Zainzzkk, line 20: could this come from test-mocks? (please ignore if this has been moved across)

portal-api/src/v1/gef/services/clone/helpers/generate-cloned-deal.test.ts
```diff
@@ -0,0 +1,146 @@
...
+      maker,
+      bank: mockBank,
+      ukefDealId: null,
+      checkerId: null,
+      editedBy: [maker._id],
+      exporter: generateClonedExporter(existingDeal.exporter),
+      portalActivities: [],
+      supportingInformation: {},
+      clonedDealId: existingDealId,
+      auditRecord: generateAuditDatabaseRecordFromAuditDetails(mockAuditDetails),
```
Zainzzkk, line 104: how different is this from existingDeal? could we do { ...existingDeal, property: ... }?

portal-api/src/v1/gef/services/clone/helpers/generate-cloned-facility.test.ts
```diff
@@ -0,0 +1,131 @@
...
+    jest.useRealTimers();
+  });
+
+  it('clones an issued facility correctly', () => {
+    // Act
+    const result = generateClonedFacility({ facility: issuedFacility, newDealId, auditDetails: mockAuditDetails });
+
+    // Assert
+    expect(result).toEqual({
+      ...issuedFacility,
```
Zainzzkk, line 81: same comment as the deal one, can this come from facilityDefaults? if not const expected =

portal-api/src/v1/gef/services/clone/helpers/generate-cloned-facility.test.ts
```diff
@@ -0,0 +1,131 @@
...
+    });
+  });
+
+  it('clones an unissued facility correctly', () => {
+    // Act
+    const result = generateClonedFacility({ facility: unissuedFacility, newDealId, auditDetails: mockAuditDetails });
+
+    // Assert
+    expect(result).toEqual({
+      ...unissuedFacility,
```
Zainzzkk, line 110: const expected please

portal-api/src/v1/gef/services/clone/helpers/generate-cloned-facility.test.ts
```diff
@@ -0,0 +1,131 @@
...
+import { ObjectId } from 'mongodb';
+import { add } from 'date-fns';
+import { generateAuditDatabaseRecordFromAuditDetails, generateSystemAuditDetails } from '@ukef/dtfs2-common/change-stream';
+import { generateClonedFacility } from './generate-cloned-facility';
+
+const existingDealId = new ObjectId().toString();
+const newDealId = new ObjectId();
+const mockAuditDetails = generateSystemAuditDetails();
+
+const facilityDefaults = {
```
Zainzzkk, line 11: is there a mock for this?

e2e-tests/gef/cypress/e2e/clone/facility-end-date-enabled/clone-v0-to-v1.spec.js
ttbarnes: Thanks @MarRobSoftwire 

It looks like this needs refactoring completely. We should have something like this

```js
table: {
  body: {
    row: (dealId) => ({
      status: () => cy.get(`[data-cy="ref-${dealId}-status"]`),
    });
  },
},
```

Then tests can just do this 

```
table.body.row(dealId).status().click();
```

With `dealId` and/or `index`

Not for this PR though, this could break a lot of things. Also, this could conflict with an upcoming PR regarding some E2E refactoring for GEF deals. Resolving

libs/common/src/types/mongo-db-models/deals.ts
MarRobSoftwire: Nope, this is set in `portal-api\src\v1\gef\models\application.js` as `null` rather than `undefined`

portal-api/src/v1/gef/services/clone/helpers/generate-cloned-deal.test.ts
MarRobSoftwire: I replied to Tony's comment here about this https://github.com/UK-Export-Finance/dtfs2/pull/3782/files/bb651262b4edbd25b55bb131b309bbfcef2c354f#r1824772073 

The TLDR is that I cant see any good mock deal fixtures & cloning a deal changes a lot of these values so I want to be explicit in these tests

portal-api/src/v1/gef/services/clone/helpers/generate-cloned-deal.test.ts
MarRobSoftwire: Spreading `existingDeal` wont work because of the removed properties (see the next test down). Will use `expected`

portal-api/src/v1/gef/services/clone/helpers/generate-cloned-facility.test.ts
MarRobSoftwire: Its not ideal, for similar reasoning to the deals

libs/common/.eslintrc.js
```diff
@@ -51,6 +51,7 @@ const baseRules = {
       consistent: true,
     },
   ],
+  'no-param-reassign': ['error', { props: true, ignorePropertyModificationsFor: ['draft', 'req', 'res'] }],
```
Zainzzkk, line 54: just need approval from @abhi-markan on this

ttbarnes (dismissed):

libs/common/.eslintrc.js
ttbarnes: Actually sorry @MarRobSoftwire - with fresh eyes - and forgetting previous comments, was this eslint change added for the `produce` function? If so, I think we can change the function and remove the need for this `no-param-reassign` rule entirely? E.g:

```ts
produce(existingDeal, (draft) => {
  const cloned = draft;  // <=====================

  propertiesToRemove.forEach((property) => {
    if (cloned[property]) {
      delete cloned[property];
    }
  });

  cloned._id = new ObjectId();
  cloned.version = getCurrentGefDealVersion();
  cloned.x = '...'

```

I don't think we should have `'draft'`  in the eslintrc, feels strange as it's very specific to cloning rather than a generic thing.

MarRobSoftwire: This isnt specific to cloning but actually for the package `immer` - there's quite a few places that are using this package & disabling this rule. (see https://github.com/immerjs/immer/issues/189)

This rule is also violated whenever express middleware adds something to the `req` or `res` objects 
image

Your example tony is still editing the parameters & feels like its hiding this. I quite like our linting to explictly allow in this case, rather than setting a precedent of using `// eslint-ignore` or reassigning the object.

Adding this rule in will allow us to remove lots of `// eslint-ignores` (https://ukef-dtfs.atlassian.net/browse/DTFS2-7603) which I'd love to get done

ttbarnes: My concern is, the rule is there for a reason - the example I provided is valid and is not editing the parameters; It creates a new instance.

Would be great to remove as many eslint ignore's as possible, but I don't think we should completely disable the rule.

I.e instead of `req.routePath` (not sure what this is used for), maybe using e.g `req.session` or `req.headers` would be better.

AlexBramhill: Afaik, this rule is to stop mutability of objects. Immer uses a draft object that is mutated to create a resultant object that is immutable (I think it actually object.freezes). This rule can be relaxed for immer as a result.

portal-api/src/v1/gef/controllers/clone-gef-deal.controller.js
MarRobSoftwire: @Zainzzkk are you happy to resolve this?

libs/common/.eslintrc.js
MarRobSoftwire: Yep, if we want to mutate anything on the `req` (e.g. session or headers) we need to disable the rule for these variables. The `res` object is needed for setting `res.locals` (e.g. gef-ui\server\middleware\csrf\index.js or gef-ui\server\middleware\sessionUserToLocals\index.js)

(I've just checked and we're disabling this rule anyway for `req` and `res` [by using Airbnb](https://github.com/airbnb/javascript/blob/master/packages/eslint-config-airbnb-base/rules/best-practices.js#L237) )

AlexBramhill: Ah nice -- So it's just for `draft` which is used in immer which is in discussion.

In which case is this controversial still as this is a key point to how [immer functions](https://immerjs.github.io/immer/) which is key for us `work with immutable state in a more convenient way`? 

This eslint rule is trying to achieve `immutable state` as well -- but in this case it's conflicting as immer requires it off to allow for an immutable state object to be created from another immutable state object through a convenient way. Having `draft` as an exception to the eslint rule shouldn't be an issue at all as it is only used by immer.

@ttbarnes any concerns?

libs/common/src/helpers/gef-deal-versioning.ts
```diff
@@ -26,9 +26,13 @@ type Feature = 'FACILITY_END_DATE';
 /**
  * GEF deals are versioned according to what features they include and support
  *
+ *
  * | Version number | Features introduced |
  * |----------------|---------------------|
  * | 1              | Facility end date   |
+ *
+ * NOTE: If features are added to previous versions then the existing in-flight deals may not
+ * supported & they will need updating.
```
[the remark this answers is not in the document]
MarRobSoftwire: Added :)

libs/common/.eslintrc.js
ttbarnes: Sounds good to me.

libs/common/src/helpers/gef-deal-versioning.ts
```diff
@@ -26,9 +26,13 @@ type Feature = 'FACILITY_END_DATE';
 /**
  * GEF deals are versioned according to what features they include and support
  *
+ *
  * | Version number | Features introduced |
  * |----------------|---------------------|
  * | 1              | Facility end date   |
+ *
+ * NOTE: If features are added to previous versions then the existing in-flight deals may not
```
AlexBramhill, line 34:
```suggestion
 * NOTE: If features are added to previous versions then the existing in-flight deals on the previous versions may not
```

maybe?

portal-api/src/v1/gef/services/clone/clone-deal.service.ts
```diff
@@ -0,0 +1,86 @@
...
+ * @param params.bankInternalRefName - the new deals internal ref name
+ * @param params.additionalRefName - new new deals additional ref name
+ * @param params.maker - the user making the clone
+ * @param params.userId - the makers id
+ * @param params.bank - the users bank
+ * @param params.auditDetails - the users audit details
+
+ * @returns the inserted deal id
+ */
+export const cloneDealToLatestVersion = async ({
```
AlexBramhill, line 42: Is this not being used anywhere correct right now?

MarRobSoftwire: This is used in dtfs2\portal-api\src\v1\gef\controllers\clone-gef-deal.controller.js

AlexBramhill: I did not pull. Thanks!

AlexBramhill (approved):
lgtm, great job

abhi-markan (approved):
LGTM 👍🏼

PR #3793, opened 2024-10-31
feat(DTFS2-7469): simplify facility type mappings

BethThomas141:
## Introduction :pencil2:
I'm trying to reduce dependency on the facility snapshot clone in `mapFacility` so that we can ultimately (in a few refactors time) remove the cloned object. 
I hope this will make the mappings easier to understand on first read. 

- Convert the `mapFacilityType` function to TS
- Split it out for GEF and non-GEF facilities
- Only pass in the needed params to calculate the mapped facility type


trade-finance-manager-api/src/v1/rest-mappings/mappings/facilities/mapFacilityProduct.js
```diff
@@ -1,34 +1,37 @@
...
     };
   }
 
-  if (type === FACILITY_TYPE.CONTINGENT) {
+  if (type === CONTINGENT) {
     mapped = {
-      code: CONSTANTS.FACILITIES.FACILITY_PRODUCT_CODE.GEF,
-      name: CONSTANTS.FACILITIES.FACILITY_PRODUCT_NAME.GEF,
+      code: FACILITY_PRODUCT_CODE.GEF,
+      name: FACILITY_PRODUCT_NAME.GEF,
```
BethThomas141, line 34: I've just simplified this file a bit by destructuring at the top.

trade-finance-manager-api/src/v1/rest-mappings/mappings/facilities/mapFacilitySnapshot.js
```diff
@@ -55,7 +55,7 @@ const mapFacilitySnapshot = (facility, dealSnapshot) => {
 
   clonedSnapshot.facilityProduct = mapFacilityProduct(type);
 
-  clonedSnapshot.type = mapFacilityType(clonedSnapshot);
+  clonedSnapshot.type = mapFacilityType(type, facilitySnapshot);
```
BethThomas141, line 58: I'm aiming to reduce the need to use this 'clonedSnapshot' so that we can ultimately remove it

trade-finance-manager-api/src/v1/rest-mappings/mappings/facilities/mapFacilityType.ts
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * @example
+ * // returns 'Cash Facility'
+ * mapGefFacilityType('Cash')
+ *
+ * @example
+ * // returns 'Contingent Facility'
+ * mapGefFacilityType('Contingent')
+ */
+export const mapGefFacilityType = (type: FacilityType): string | null => {
```
BethThomas141, line 44: Brand new file showing in the diff as I've changed it from JS to TS. I've split out the Gef facility type mappings as GEF facilities will only ever have Cash or Contingent types

trade-finance-manager-api/src/v1/rest-mappings/mappings/facilities/mapFacilityType.js
```diff
@@ -1,49 +0,0 @@
...
- * mapFacilityType({...facility, ukefFacilityType: 'Cash'})
- *
- * @example
- * // returns 'Contingent Facility'
- * mapFacilityType({...facility, ukefFacilityType: 'Contingent'})
- */
-const mapFacilityType = (facility) => {
-  const { facilityProduct, ukefFacilityType } = facility;
-
-  if (facilityProduct.code === CONSTANTS.FACILITIES.FACILITY_PRODUCT_CODE.BOND) {
```
BethThomas141, line 30: Previously it was using the derived 'facilityProduct' on the `clonedFacilitySnapshot` to determine if it was a BOND or a LOAN facility. However if you look at the `mapFacilityProduct` file (also shows as one of the changed files so hopefully easy to locate), this is precisely equivalent to the original facility snapshot's `type` being 'Bond' or 'Loan'. 

So in the new TS file I've just based this logic of the `type` instead of using the `code` as it's doing the same things but easier to read and reduces dependency on the `clonedFacilitySnapshot`

libs/common/src/types/mongo-db-models/facility.ts
```diff
@@ -52,6 +52,7 @@ export type Facility = {
   paymentType: string;
   createdAt: UnixTimestamp;
   updatedAt: UnixTimestamp;
+  bondType?: string; // only featured in BOND facilities
   ukefExposure: number;
   guaranteeFee: number;
   submittedAsIssuedDate: UnixTimestampString | null;
```
ttbarnes, line 58: I know this isn't part of your PR, but I noticed in this file we have `dataMigration.drupalFacilityId`

I think this can be removed as it is very very old and there are no instances of `drupalFacilityId`, can we remove please whilst we're here?

trade-finance-manager-api/src/v1/rest-mappings/mappings/facilities/mapFacilityType.ts
```diff
@@ -0,0 +1,54 @@
+import { Facility, FacilityType, FACILITY_TYPE, MAPPED_FACILITY_TYPE } from '@ukef/dtfs2-common';
+
+const { BOND, LOAN, CASH, CONTINGENT } = FACILITY_TYPE;
+
+/**
+ * Maps type to a different facility type used in TFM-UI and TFM-API.
```
ttbarnes, line 6:
```suggestion
 * Maps a facility type to a differently worded facility type,
 * to ultimately be rendered in the UI, as per design.
```

Or similar - just so it's clear _why_ we're doing this

ttbarnes (changes requested):
2x minor comments otherwise LGTM

libs/common/src/types/mongo-db-models/facility.ts
BethThomas141: Ah yes, I'm nervous to remove it as I feel like it might have been added to this file for a reason - it's commented with:
`  // Legacy data from migrating old GEF Facilities into DTFS` - is there any chance instances of this can therefore still exist in the database (despite being super old)?

ttbarnes: I've asked Abhi/Zain to check the production database to see if any facilities have `drupalFacilityId`

ttbarnes: There are some old deals with this @BethThomas141 so we should keep it to be safe.

Sorry to muck around!

BethThomas141: No problem at all, interesting to check! Happy to mark as resolved?

ttbarnes (dismissed):

trade-finance-manager-api/src/v1/rest-mappings/mappings/facilities/mapFacilityType.ts
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * @example
+ * // returns 'Bid Bond'
+ * mapFacilityType('Bond', {...facilitySnapshot, code: 'BSS', bondType: 'Bid Bond'})
+ *
+ * @example
+ * // returns 'Loan'
+ * mapFacilityType('Loan', facilitySnapshot)
+ */
+export const mapFacilityType = (type: FacilityType, facilitySnapshot: Facility): string | null => {
```
MarRobSoftwire, line 19: Do we want an explicit `mapBssEwcsFacilityType` name?

trade-finance-manager-api/src/v1/rest-mappings/mappings/facilities/mapFacilityType.test.ts
```diff
@@ -0,0 +1,59 @@
+import { Facility, FACILITY_TYPE, MAPPED_FACILITY_TYPE } from '@ukef/dtfs2-common';
+import { mapFacilityType, mapGefFacilityType } from './mapFacilityType';
+
+const { BOND, LOAN, CASH, CONTINGENT } = FACILITY_TYPE;
+
+describe('mapFacilityType', () => {
+  describe(`when the facility type is ${BOND}`, () => {
+    it('should return the facility bondType', () => {
+      const mockBondFacility = {
+        bondType: 'Bid bond',
```
MarRobSoftwire, line 10: Assume we dont have any constants for `Bid bond`?

BethThomas141: No, the `bondType` is still a bit of a mystery to me, but it does exist. Unsure what other values it could take

trade-finance-manager-api/src/v1/rest-mappings/mappings/facilities/mapFacilityType.ts
BethThomas141: Good idea!

trade-finance-manager-api/src/v1/rest-mappings/mappings/facilities/mapFacilityType.test.ts
ttbarnes: We do actually have some constants for `bondType` but they're only in E2E. 

@BethThomas141 If you search for `BSS_FACILITY_TYPE =` and `bondType: ` - might help to demystify? `bondType` is a field in the UI that a maker can select in the Portal `(bond-details.njk)`

I will create a ticket to update instances of `bondType` to use constants

ttbarnes (approved):

libs/common/src/types/mongo-db-models/facility.ts
```diff
@@ -52,6 +52,7 @@ export type Facility = {
   paymentType: string;
   createdAt: UnixTimestamp;
   updatedAt: UnixTimestamp;
+  bondType?: string; // only featured in BOND facilities
```
AlexBramhill, line 55: @BethThomas141 @MarRobSoftwire  do we want to start extracting these out into a bond facility type and union the types here??

MarRobSoftwire: I think we should look to do this at some point but it might be quite a bit of work to separate out all the values. Needs some thought about how to do this as an iterative approach rather than all in one

ttbarnes: I've started updating all instances of `bondType` to use constants, I could do this at the same time.

I don't think such a PR would be _too_ bad, there are ~90 instances of hard coded `bondType` in ~60 files that'd be replaced, would just be repetitive

BethThomas141: @AlexBramhill agree we should do this, let's do it as a separate PR!

BethThomas141: Will resolve this here, but a conversation to pick up in another PR

ttbarnes: I've created DTFS2-7607 for this 👍

PR #3796, opened 2024-11-01
feat(FN-3391): display tolerance info text on add payment page

RLavender98:
## Introduction :pencil2:
We want to display the tolerance if there is one in place to PDC inputters.

- Display inset text explaining tolerance and containing GBP tolerance on add payment page

## Screenshot
Screenshot 2024-11-01 at 09 00 59


dtfs-central-api/src/v1/controllers/utilisation-report-service/get-selected-fee-records-details.controller/index.ts
```diff
@@ -58,11 +59,14 @@ export const getSelectedFeeRecordDetails = async (req: GetSelectedFeeRecordDetai
 
     const canAddToExistingPayment = await canFeeRecordsBeAddedToExistingPayment(reportId, selectedFeeRecords);
 
+    const gbpTolerance = await PaymentMatchingToleranceService.getGbpPaymentMatchingTolerance();
```
ttbarnes, line 62: Should we/do we need to throw an error here if no `gbpTolerance` is found? Other checks above throw an error

ttbarnes (commented):
One comment otherwise LGTM

dtfs-central-api/src/v1/controllers/utilisation-report-service/get-selected-fee-records-details.controller/index.ts
RLavender98: `PaymentMatchingToleranceService.getGbpPaymentMatchingTolerance` will either return a value, or throw an error. It is not possible for it to just not return a value. If it throws because there is no tolerance, then that gets caught by the try catch block surrounding this block of code.

ttbarnes: Makes sense, thanks @RLavender98

ttbarnes (approved):

Zainzzkk (approved):

sw-mattw (approved):
Lovely

PR #3844, opened 2024-11-01
feat(DTFS2-7595): add deal cancellation success banner across all pages

BethThomas141:
## Introduction :pencil2:
The deal cancellation banner should be shown on all pages when the cancellation is scheduled

- Add new `getSuccessBannerMessage` helper & use on all tabs
- Update unit test coverage
- Add e2e coverage

### Tasks (logged in as PIM)
image

### Parties (logged in as Underwriter Manager)
image

### Documents (Logged in as business support user)
image

### Activity and Comments (Logged in as PIM)
image

### Underwriting (logged in as PDC Reconcile)
image

- Add getUkefDealId helper
- The `activities/index.test.js` unit test was not properly clearing mocks between tests. This has been updated & the tests fixed
- Add `ff-test` files to dev dependencies for linting


e2e-tests/tfm/cypress/e2e/journeys/deal-cancellation/submit-deal-cancellation-effective-from-in-future.spec.js
```diff
@@ -119,6 +122,28 @@ context('Deal cancellation - submit cancellation with "effectiveFrom" in future'
...
+        caseSubNavigation.documentsLink().click();
+        successBanner().should('exist');
+        cy.assertText(successBanner(), expectedSuccessBannerText);
+
+        caseSubNavigation.activityLink().click();
+        successBanner().should('exist');
+        cy.assertText(successBanner(), expectedSuccessBannerText);
+
+        caseSubNavigation.underwritingLink().click();
+        successBanner().should('exist');
```
ttbarnes, line 144: No need for the `.should('exist');` checks here - if it doesn't exist, `cy.assertText` will fail because it can't find the text 😎

trade-finance-manager-ui/server/controllers/helpers/get-success-banner-message.helper.ts
MarRobSoftwire, line 1: TODO add unit tests

dtfs-central-api/src/repositories/tfm-deals-repo/tfm-deal-cancellation.repo.ts
MarRobSoftwire, line 1: Sorry for the scope creep here! I came to use the `getUkefDealId` and realised the return types were incorrect in the first place.

trade-finance-manager-ui/server/controllers/case/activity/index.test.js
```diff
@@ -18,354 +18,341 @@ const session = {
...
-        },
-        body: {
-          comment: normalComment,
-        },
-        session,
-      };
-      const activities = mapActivities(mockDeal.tfm.activities);
-
-      await activityController.postComment(req, res);
-      expect(res.render).toHaveBeenCalledWith('case/activity/activity.njk', {
```
MarRobSoftwire, line 308: These were only passing because the mocks weren't being cleared. This controller should never have called res.render

trade-finance-manager-ui/server/controllers/helpers/get-success-banner-message.helper.ts
```diff
@@ -0,0 +1,40 @@
...
+ * @returns the success message to be shown or null
+ */
+export const getSuccessBannerMessage = async (
+  submissionType: DealSubmissionType,
+  user: TfmSessionUser,
+  userToken: string,
+  dealId: string,
+  ukefDealId: string,
+): Promise<string | null> => {
+  const dealCancellationIsEnabled = isDealCancellationEnabled(submissionType, user);
```
MarRobSoftwire, line 23: Just noticed the ACs actually need all users to be able to see the banner - need to update this test

dtfs-central-api/src/repositories/tfm-deals-repo/tfm-deal-cancellation.repo.ts
```diff
@@ -207,7 +208,7 @@ export class TfmDealCancellationRepo {
 
     const updatedFacilityUkefIds = getUkefFacilityIds(updatedFacilities);
 
-    return { cancelledDealUkefId: dealId, riskExpiredFacilityUkefIds: updatedFacilityUkefIds };
+    return { cancelledDealUkefId: getUkefDealId(deal.dealSnapshot), riskExpiredFacilityUkefIds: updatedFacilityUkefIds };
```
ttbarnes, line 211: How about this instead?

```ts
const cancelledDealUkefId = getUkefDealId(deal.dealSnapshot);
const riskExpiredFacilityUkefIds = getUkefFacilityIds(updatedFacilities);

return { cancelledDealUkefId, riskExpiredFacilityUkefIds };
```
```

File: trade-finance-manager-ui/server/controllers/case/activity/index.js [hunk 5]
```diff
@@ -36,12 +38,18 @@ const getActivity = async (req, res) => {
 
   const activities = mapActivities(deal.tfm.activities);
 
+  const { submissionType } = deal.dealSnapshot;
+
+  const successMessage =
+    (await getSuccessBannerMessage(submissionType, user, userToken, dealId, getUkefDealId(deal.dealSnapshot))) ?? getFlashSuccessMessage(req);
```
ttbarnes, line 44: How about this instead, I think would be easier to read/maintain

```ts
const successBannerMessage = await getSuccessBannerMessage(submissionType, user, userToken, dealId, getUkefDealId(deal.dealSnapshot));
const flashSuccessMessage = getFlashSuccessMessage(req);

const successMessage = successBannerMessage || flashSuccessMessage;
```

trade-finance-manager-ui/server/controllers/case/parties/index.js
```diff
@@ -33,11 +36,17 @@ const getAllParties = async (req, res) => {
 
     const canEdit = userCanEdit(user);
 
+    const { submissionType } = deal.dealSnapshot;
+
+    const successMessage =
```
ttbarnes, line 41: Can we use middleware to do this instead of repeating the same thing in 7 controllers?

Middleware can check if the current route is in array of relevant routes, if so, can do e.g

```ts
const successBannerMessage = await getSuccessBannerMessage(submissionType, user, userToken, dealId, getUkefDealId(deal.dealSnapshot));

const flashSuccessMessage = getFlashSuccessMessage(req);

req.locals.successMessage = successBannerMessage || flashSuccessMessage;

return next();

File: trade-finance-manager-ui/server/controllers/helpers/get-success-banner-message.helper.ts [hunk 7]
```diff
@@ -0,0 +1,40 @@
...
+
+  const cancellation = (await api.getDealCancellation(dealId, userToken)) as TfmDealCancellationWithStatus;
+
+  const dealIsScheduledToBeCancelled = cancellation.status === TFM_DEAL_CANCELLATION_STATUS.SCHEDULED && cancellation.effectiveFrom;
+
+  if (!dealIsScheduledToBeCancelled) {
+    return null;
+  }
+
+  const formattedDate = format(new Date(cancellation.effectiveFrom), 'd MMMM yyyy');
```
ttbarnes, line 37: Do we have constants we can use for the date format?

ttbarnes (changes requested):

libs/common/src/helpers/get-ukef-deal-id.ts
```diff
@@ -0,0 +1,15 @@
+import { DEAL_TYPE } from '../constants';
+import { Deal } from '../types';
+
+/**
+ * Returns the ukef deal id on a deal depending on the deal type
+ * @param deal a BSS/EWCS or GEF deal
+ * @returns the ukef deal id
+ */
+export const getUkefDealId = (deal: Deal): string => {
```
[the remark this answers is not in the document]
MarRobSoftwire: https://ukef-dtfs.atlassian.net/browse/DTFS2-7611

trade-finance-manager-ui/server/controllers/case/parties/index.js
MarRobSoftwire: The `getSuccessBannerMessage` parameters come from the deal object & it makes a request to the cancellation endpoint to determine if the deal is scheduled for cancellation or not.

We could add lots more middleware to commonise this (as has been discussed before) but it would be a large change of pattern & definitely out of scope for this ticket

ttbarnes: I am concerned that we are otherwise introducing tech debt by repeating the same thing 7x times.

Appreciate we've had discussions about this before - would be great to move e.g `await api.getDeal` into middleware.

I propose that as an interim step we introduce some middleware purely for the success banner - for these routes, that calls `api.getDeal` and handles the success banner - then we can eventually create some `getDeal` middleware to do this for all routes. I am thinking something like this:

```ts
const RELEVANT_ROUTES = [
  '/summary',
  '/x-route/activity',
  '/etc',
];

const shouldExecute = (url: string) => {
  if (url.includes('/case') && RELEVANT_ROUTES.includes(url)) {
    return true;
  }

  return false;
};

const cancellationSuccessBannerMiddleware = async (req, res, next) => {
  if (shouldExecute(req.originalUrl)) {

    const { dealId, dealSnapshot, submissionType } = await api.getDeal(req.params._id, userToken);

    const successBannerMessage = await getSuccessBannerMessage(submissionType, user, userToken, dealId, getUkefDealId(dealSnapshot));

    res.locals.successMessage = successBannerMessage || flashSuccessMessage;

    return next();
  }

  return next();
};
```

Then:

1. We don't need `api.getDeal` calls in these controllers.
2. We don't have the same logic repeated 7 times in each controller and unit test.
3. If we need to update the logic, we only need to do it in one place.
4. We don't introduce tech debt.
  - I know we'd eventually need to remove `api.getDeal` from the proposed interim middleware, but this is miniscule.
6. For each relevant controller (that's currently updated in this PR), they can have `api.getDeal` removed and simply consume `res.locals.successMessage`.

I think it is cleaner and more scalable to do this as part of this PR for the reasons above. What do you think?

trade-finance-manager-ui/server/controllers/helpers/get-success-banner-message.helper.test.ts
```diff
@@ -0,0 +1,95 @@
...
+        reason: '',
+        status: TFM_DEAL_CANCELLATION_STATUS.SCHEDULED,
+      });
+
+      // Act
+      const response = await getSuccessBannerMessage({ submissionType, userToken, dealId, ukefDealId });
+
+      // Assert
+      const expectedFormattedDate = format(effectiveFromDate, DATE_FORMATS.D_MMMM_YYYY);
+      expect(response).toEqual(`Deal ${ukefDealId} scheduled for cancellation on ${expectedFormattedDate}`);
```
ttbarnes, line 81: `const expected` please?

dtfs-central-api/src/repositories/tfm-deals-repo/tfm-deal-cancellation.repo.ts
```diff
@@ -274,6 +277,6 @@ export class TfmDealCancellationRepo {
 
     const cancelledFacilityUkefIds = getUkefFacilityIds(cancelledFacilities);
 
-    return { cancelledDealUkefId: dealId, riskExpiredFacilityUkefIds: cancelledFacilityUkefIds };
+    return { cancelledDealUkefId: getUkefDealId(deal.dealSnapshot), riskExpiredFacilityUkefIds: cancelledFacilityUkefIds };
```
ttbarnes, line 280: Can we do this instead? Simpler

```ts
const cancelledDealUkefId = getUkefDealId(deal.dealSnapshot);
const riskExpiredFacilityUkefIds = getUkefFacilityIds(updatedFacilities);

return { cancelledDealUkefId, riskExpiredFacilityUkefIds };
```

dtfs-central-api/src/repositories/tfm-deals-repo/tfm-deal-cancellation.repo.ts
ttbarnes: Resolving in favour of a newer comment

trade-finance-manager-ui/server/controllers/case/parties/index.js
MarRobSoftwire: This feels like a large change & I've explicitly been told not to introduce new patterns. I'd want to have a chat with @abhi-markan  & the rest of the devs before making db calls in middleware.

I think in the meantime the techdebt incurred by having the same 2 function calls repeated in each controller isnt too large

ttbarnes: Thanks all.

@AlexBramhill I don't entirely agree RE middleware. I've worked on plenty of applications that have business logic in middleware. A simple example would be user/auth checks - this is business logic and should absolutely be in middleware.

I know this could seem "magical" as you describe, or confusing, for new devs, but having the same logic repeated in various places, instead of having 1x DRY instance, can be equally confusing and also unscalable - it is a bad practice.

A new helper like you suggested seems great and much preferred to the current implementation.

I would like us to explore and implement a middleware to get a deal though and other common api fetches; One for another day.

azure-functions/acbs-function/.eslintrc.js
```diff
@@ -18,7 +18,15 @@ const baseRules = {
   'import/no-extraneous-dependencies': [
     'error',
     {
-      devDependencies: ['**/*.test.{js,ts}', '**/*.api-test.{js,ts}', '**/*.spec.{js,ts}', '**/webpack.*.{js,ts}', '**/api-test*/**', '**/__mocks__/**'],
+      devDependencies: [
+        '**/*.test.{js,ts}',
+        '**/*.ff-test.{js,ts}',
```
MarRobSoftwire, line 23: Have added `.ff-test.` to this array - noticed this was missing when importing `import { createMocks } from 'node-mocks-http';`

libs/common/src/helpers/get-ukef-deal-id.test.ts
```diff
@@ -0,0 +1,39 @@
+import { BssEwcsDeal, GefDeal } from '../types';
+import { getUkefDealId } from './get-ukef-deal-id';
+import { DEAL_TYPE } from '../constants';
+
+describe('getUkefDealId', () => {
+  it('returns ukefDealId on a GEF deal', () => {
+    // Arrange
+    const ukefDealId = 'ukefDealId';
+
+    const deal = {
```
AlexBramhill, line 10: Not for this PR, but my PR starts to look at pulling out these objects from test files into `aBla` functions...

libs/common/src/helpers/get-ukef-deal-id.ts
```diff
@@ -0,0 +1,15 @@
...
+  if (deal.dealType === DEAL_TYPE.GEF) {
+    return deal.ukefDealId;
+  }
+
+  return deal.details.ukefDealId;
```
AlexBramhill, line 14: Could probably oneline this, or do you find this better for clarity

libs/common/src/types/mongo-db-models/deals.ts
```diff
@@ -1,16 +1,22 @@
+import { ObjectId } from 'mongodb';
 import { DEAL_TYPE } from '../../constants';
 import { AnyObject } from '../any-object';
+import { DealSubmissionType } from '..';
 
-type BssEwcsDeal = AnyObject & {
+export type BssEwcsDeal = AnyObject & {
+  _id: ObjectId;
```
AlexBramhill, line 7: I wonder if we want a base type? What do you think?

trade-finance-manager-ui/server/controllers/case/underwriting/index.test.js
```diff
@@ -53,18 +55,27 @@ describe('GET getUnderwriterPage', () => {
         user: MOCKS.MOCK_USER_UNDERWRITER_MANAGER,
         userCanEdit: true,
       },
+      successMessage: mockSuccessBannerMessage,
```
AlexBramhill, line 58: I have a lot of thoughts about these tests, but it isnt a discussion for this PR. Good job managing to work with them though.

trade-finance-manager-ui/server/controllers/helpers/deal-cancellation-enabled.helper.ff-test.ts
```diff
@@ -1,37 +1,69 @@
...
 import { TfmSessionUser } from '../../types/tfm-session-user';
 
 const pimUser = { teams: [TEAM_IDS.PIM] } as TfmSessionUser;
 const nonPimUser = { teams: [TEAM_IDS.UNDERWRITERS] } as TfmSessionUser;
 
 const { AIN, MIN, MIA } = DEAL_SUBMISSION_TYPE;
 
-describe('dealCancellationEnabled', () => {
+describe('isDealCancellationEnabledForUser', () => {
   describe('when `FF_TFM_FACILITY_END_DATE_ENABLED` is set to true', () => {
```
AlexBramhill, line 11: Lovely tests, nice job

trade-finance-manager-ui/server/helpers/get-flash-success-message.test.ts
```diff
@@ -1,5 +1,5 @@
 import { createMocks } from 'node-mocks-http';
-import { getFlashSuccessMessage } from './getFlashSuccessMessage';
+import { getFlashSuccessMessage } from './get-flash-success-message';
```
AlexBramhill, line 2: Can we document if no flash stored for `successMessage` please

trade-finance-manager-ui/server/controllers/helpers/get-success-banner-message.helper.getDealSuccessBannerMessage.ff-test.ts
```diff
@@ -0,0 +1,98 @@
...
+  getDealCancellation: (params: AnyObject) => getDealCancellationMock(params),
+}));
+
+const { AIN, MIN, MIA } = DEAL_SUBMISSION_TYPE;
+
+const userToken = 'userToken';
+const dealId = 'dealId';
+const ukefDealId = 'ukefDealId';
+
+describe('getDealSuccessBannerMessage', () => {
```
AlexBramhill, line 18: Could we document if no success message on either please

trade-finance-manager-ui/server/controllers/helpers/get-success-banner-message.helper.ts
```diff
@@ -0,0 +1,53 @@
...
+  return `Deal ${ukefDealId} scheduled for cancellation on ${formattedDate}`;
+};
+
+/**
+ * Determines if a success message should be shown and returns the highest priority message.
+ * @param params - the deal
+ * @param params.dealSnapshot - the deal
+ * @param params.userToken - the user token
+ * @param params.req the express request object
+ * @returns the success message to be shown or null
```
AlexBramhill, line 43: Does this ever return null? It depends on the behaviour of `getFlashSuccessMessage` on an empty array...

AlexBramhill (commented):
a few changes but looking good

dtfs-central-api/src/repositories/tfm-deals-repo/tfm-deal-cancellation.repo.ts
```diff
@@ -189,7 +190,7 @@ export class TfmDealCancellationRepo {
       update,
     );
 
-    if (!updateDeal?.matchedCount) {
+    if (!deal) {
```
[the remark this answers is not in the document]
MarRobSoftwire: I want to use `findOneAndUpdate` so that the function can return the `ukefDealId` - it was previously returnign the deal object id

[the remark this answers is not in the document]
MarRobSoftwire: (UpdateOne returns an UpdateResult, findOneAndUpdate returns a ModifyResult. )

trade-finance-manager-ui/server/controllers/helpers/get-success-banner-message.helper.getDealSuccessBannerMessage.ff-test.ts
MarRobSoftwire: I've gone for `expect(response).toEqual(getFlashSuccessMessage(req));` to document it defaults to that value & added undefined tests for `getFlashSuccessMessage`

trade-finance-manager-api/api-tests/v1/deal-cancellation/deal-cancellation.get.api-test.ts
```diff
@@ -68,13 +67,6 @@ describe('/v1/deals/:id/cancellation', () => {
         jest.resetAllMocks();
       });
 
-      withTeamAuthorisationTests({
```
ttbarnes, line 71: Why has this been removed please?

trade-finance-manager-ui/server/controllers/case/activity/index.test.js
```diff
@@ -18,354 +18,341 @@ const session = {
...
+            _id: mockDeal._id,
+          },
+          session,
+          flash: jest.fn(() => [mockSuccessMessage]),
+        };
+
+        const activities = mapActivities(mockDeal.tfm.activities);
+
+        await activityController.getActivity(req, res);
+        expect(res.render).toHaveBeenCalledWith('case/activity/activity.njk', {
```
ttbarnes, line 69: Can we add a new line before the `expect` please

trade-finance-manager-ui/server/controllers/case/activity/index.test.js
```diff
@@ -18,354 +18,341 @@ const session = {
...
-      api.getAmendmentsByDealId = () => Promise.resolve({ data: [] });
+      it('should render template with data', async () => {
+        const mockSuccessMessage = 'mockSuccessMessage';
+
+        const req = {
+          params: {
+            _id: mockDeal._id,
+          },
+          session,
+          flash: jest.fn(() => [mockSuccessMessage]),
```
ttbarnes, line 63: Instead of this and `successMessage: mockSuccessMessage` below, I think the tests should simply call `getDealSuccessBannerMessage`

trade-finance-manager-api/api-tests/v1/deal-cancellation/deal-cancellation.get.api-test.ts
MarRobSoftwire: The GET endpoints don't need to be behind RBAC, since all users need to know if a deal has been scheduled

trade-finance-manager-ui/server/controllers/case/activity/index.js
```diff
@@ -76,12 +86,21 @@ const filterActivities = async (req, res) => {
 
   const activities = mapActivities(deal.tfm.activities);
 
+  const { dealSnapshot } = deal;
+
+  const successMessage = await getDealSuccessBannerMessage({
```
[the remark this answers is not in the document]
MarRobSoftwire: I tried this & unfortunately it seems to need the context of the `req` object- I asume req.flash calls `this.session` or something similar :(

trade-finance-manager-ui/server/controllers/case/parties/index.js
AlexBramhill: Completely agree re middleware -- I probably oversimplified too much. My general thoughts align with [nexts](https://nextjs.org/docs/pages/building-your-application/routing/middleware).

If we are making the pattern that middleware sources all data we need for an endpoint that's fine (and feels okay tbh) -- but I think we need to clearly define scope as not to end up with everything in middleware, and figure out how we're going to do nice types for it.

ttbarnes: Agreed @AlexBramhill I think only "general" or "common" things like getting a deal or getting a facility should be in middleware and locked down but a list of specific routes and methods. For example a lot of GET routes will have similar POST routes for form validation, some POSTs might only need deal/facility data if there are validation errors, so that should be handled on the controller level

One for another day 😄

trade-finance-manager-ui/server/controllers/case/activity/index.test.js
ttbarnes: Sorry, didn't realise it was async. I would prefer to mock `getDealSuccessBannerMessage` in the test so it's more realistic, but maybe overkill. What do you think?

trade-finance-manager-ui/server/controllers/case/activity/index.js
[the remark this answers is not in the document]
ttbarnes: Did you try updating `getFlashSuccessMessage` to this?

```ts
export const getFlashSuccessMessage = (reqFlash) => 

// getFlashSuccessMessage(req.flash);
```

Really want to avoid passing the `req` around functions as it is not a good practice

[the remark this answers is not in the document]
MarRobSoftwire: Yep in commit 501917d47574d95c7ce89f7bfe8f95c2a6571809 I tried this, but it was throwing `'req.flash() requires sessions'` - looking [at the connect flash package](https://github.com/jaredhanson/connect-flash/blob/master/lib/flash.js) it uses `this.session`. I could probably get around this by passing in `flash.bind(req)` but I think thats replacing one code smell with another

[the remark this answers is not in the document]
MarRobSoftwire: Tbh the more I look at connect-flash the more I dislike it - it was last updated 11 years ago :'( https://www.npmjs.com/package/connect-flash

dtfs-central-api/src/repositories/tfm-deals-repo/tfm-deal-cancellation.repo.ts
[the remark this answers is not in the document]
MarRobSoftwire: Update: Have undone this as it'l cause conflicts with another PR

trade-finance-manager-ui/server/controllers/case/activity/index.js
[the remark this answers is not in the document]
ttbarnes: Completely agree, not ideal :/ Thinking of cleaner ways to do this, what do you think about this?

```ts
  const successMessage = await getDealSuccessBannerMessage({
    dealSnapshot,
    userToken,
    // req,
    flashSuccessMessage: getFlashSuccessMessage(req),
  });
```

Or maybe even this - it just repeats `getFlashSuccessMessage` - it is not DRY, but the positive is we're not passing `req` to another function and is most likely more efficient. If we were to do this I think removing `getFlashSuccessMessage` to be consistent would be better.

```ts
  const successMessage = await getDealSuccessBannerMessage({
    dealSnapshot,
    userToken,
    // req,
    flashSuccessMessage: req.flash('successMessage')[0],
  });
```

Just really want to avoid passing `req` around, it passes _so_ many things into a function when the function typically only needs 1x thing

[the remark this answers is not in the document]
MarRobSoftwire: Have updated, I do feel that `getDealSuccessBannerMessage` is doing something slightly odd now by returning a parameter passed into it but happy to leave & get this merged in.
Also added some more constants and types so that ` flashSuccessMessage: req.flash(FLASH_TYPES.SUCCESS_MESSAGE)[0],` feels a bit safer. Tried to get declaration merging working to override the flash type & only accept `FlashType` rather than string, but wasn't quite working & non-essential

trade-finance-manager-ui/server/controllers/case/activity/index.test.js
MarRobSoftwire: Misunderstood this comment! Have updated to mock all the async calls of `getDealSuccessBannerMessage`

trade-finance-manager-ui/server/controllers/case/activity/index.js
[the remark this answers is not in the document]
ttbarnes: Great work @MarRobSoftwire 

I do agree it's a bit odd, but it seems to be the cleanest approach we can do without passing `req` around in different functions

trade-finance-manager-ui/server/controllers/helpers/deal-cancellation-enabled.helper.ts
```diff
@@ -15,22 +15,28 @@ const { AIN, MIN } = DEAL_SUBMISSION_TYPE;
  * @param submissionType - the deal submission type
  * @returns true if the deal is AIN or MIN
  */
-export const canSubmissionTypeBeCancelled = (submissionType: DealSubmissionType): boolean => {
+export const canSubmissionTypeBeCancelled = (submissionType: DealSubmissionType | null): boolean => {
   return submissionType === AIN || submissionType === MIN;
 };
 
+export const isDealCancellationEnabled = (submissionType: DealSubmissionType | null): boolean => {
```
ttbarnes, line 22: Documentation please

trade-finance-manager-ui/server/controllers/helpers/get-success-banner-message.helper.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  const dealCancellationIsEnabled = isDealCancellationEnabled(submissionType);
+
+  if (!dealCancellationIsEnabled) {
+    return null;
+  }
+
+  const cancellation = (await api.getDealCancellation(dealId.toString(), userToken)) as TfmDealCancellationWithStatus;
+
+  const dealIsScheduledToBeCancelled = cancellation.status === TFM_DEAL_CANCELLATION_STATUS.SCHEDULED && cancellation.effectiveFrom;
```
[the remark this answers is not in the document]
MarRobSoftwire: Good spot - if the cancellation has been scheduled the `effectiveFrom` has to exist - have removed that check

trade-finance-manager-ui/server/controllers/helpers/deal-cancellation-enabled.helper.ts
MarRobSoftwire: Updated :)

trade-finance-manager-ui/server/controllers/helpers/get-success-banner-message.helper.ts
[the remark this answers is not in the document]
ttbarnes: Even better 😄

ttbarnes (approved):

PR #3847, opened 2024-11-01
fix(DTFS2-7590): facility loan amount amendment

abhi-markan:
## Introduction :pencil2:
When facility amendment is applied on facility type `Loan`, with only `amount` attribute execution should be terminated.

* Facility loan record mapping validation.

## Miscellaneous ➕ 
* Updated `azure/cli` to `v2.1.0`.

### DIFF
--- a/azure-functions/acbs-function/src/functions/acbs-amend-facility-loan-record.js
+++ b/azure-functions/acbs-function/src/functions/acbs-amend-facility-loan-record.js
@@ -42,13 +42,13 @@ df.app.orchestration('acbs-amend-facility-loan-record', function* amendFacilityL
     let facilityLoanRecordAmendments;
 
     // 1.1. Facility Loan Record (FLR) amendment mapping
-    const flrMApped = mappings.facility.facilityLoanAmend(amendments, facility, fmr);
+    const flrMapped = mappings.facility.facilityLoanAmend(amendments, facility, fmr);
 
     /**
      * When facility amendment is applied on facility type `Loan`,
      * with only `amount` attribute execution should be terminated.
      */
-    if (!flrMApped?.length) {
+    if (!flrMapped?.length) {
       facilityLoanRecordAmendments = 'Facility loan amount only amendment, aborting FLR amendment';
 
       return facilityLoanRecordAmendments;
@@ -69,7 +69,7 @@ df.app.orchestration('acbs-amend-facility-loan-record', function* amendFacilityL
         const amount = yield context.df.callActivityWithRetry('update-facility-loan-amount', retryOptions, {
           loanId,
           facilityId,
-          acbsFacilityLoanInput: flrMApped,
+          acbsFacilityLoanInput: flrMapped,
         });
 
         facilityLoanRecordAmendments = {
@@ -83,7 +83,7 @@ df.app.orchestration('acbs-amend-facility-loan-record', function* amendFacilityL
         const coverEndDate = yield context.df.callActivityWithRetry('update-facility-loan', retryOptions, {
           loanId,
           facilityId,
-          acbsFacilityLoanInput: flrMApped,
+          acbsFacilityLoanInput: flrMapped,
         });
 
         facilityLoanRecordAmendments = {
--- a/azure-functions/acbs-function/src/functions/acbs-amend-facility-loan-record.js
+++ b/azure-functions/acbs-function/src/functions/acbs-amend-facility-loan-record.js
@@ -45,8 +45,11 @@ df.app.orchestration('acbs-amend-facility-loan-record', function* amendFacilityL
     const flrMapped = mappings.facility.facilityLoanAmend(amendments, facility, fmr);
 
     /**
-     * When facility amendment is applied on facility type `Loan`,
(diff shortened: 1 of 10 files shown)


azure-functions/acbs-function/src/functions/acbs-amend-facility-loan-record.js
```diff
@@ -38,11 +38,22 @@ df.app.orchestration('acbs-amend-facility-loan-record', function* amendFacilityL
...
     let facilityLoanRecordAmendments;
 
     // 1.1. Facility Loan Record (FLR) amendment mapping
     const flrMApped = mappings.facility.facilityLoanAmend(amendments, facility, fmr);
 
+    /**
+     * When facility amendment is applied on facility type `Loan`,
+     * with only `amount` attribute execution should be terminated.
+     */
+    if (!flrMApped?.length) {
```
ttbarnes, line 51: Is this correct - It looks like `mappings.facility.facilityLoanAmend` returns facilities regardless of if there is an `amount`?

azure-functions/acbs-function/src/functions/acbs-amend-facility-loan-record.js
```diff
@@ -38,11 +38,22 @@ df.app.orchestration('acbs-amend-facility-loan-record', function* amendFacilityL
...
+      facilityLoanRecordAmendments = 'Facility loan amount only amendment, aborting FLR amendment';
```
ttbarnes, line 52: I'm not understanding the need/issue here, "with only amount attribute" - above comment will probably answer this and might suggest different wording here

azure-functions/acbs-function/src/functions/acbs-amend-facility-loan-record.js
ttbarnes, line 51: Unless `flrMApped` is intentional, whilst we're here, can we rename to `flrMApped` to `flrMapped`?

ttbarnes (changes requested):

azure-functions/acbs-function/src/functions/acbs-amend-facility-loan-record.js
abhi-markan: Yes, no that's incorrect it will return an `empty` object if and only if it is a facility loan amendment with `amount` attribute.

abhi-markan: Ah apologies, good spot.

azure-functions/acbs-function/src/functions/acbs-amend-facility-loan-record.js
abhi-markan: The above comment states the behaviour but I can elaborate here.
We only accepts two attributes at the moment on a facility amendment `cover end date` and `amount`, business requirements states that we should not process a loan (facility type) with `amount` only amendment.

Due to above requirement I have added a validation as without this validation mapping will be empty and APIM validation error will be thrown.

azure-functions/acbs-function/src/functions/acbs-amend-facility-loan-record.js
ttbarnes: Bump @abhi-markan

azure-functions/acbs-function/src/functions/acbs-amend-facility-loan-record.js
```diff
@@ -38,11 +38,22 @@ df.app.orchestration('acbs-amend-facility-loan-record', function* amendFacilityL
...
     const { amendment } = amendments;
+
     let facilityLoanRecordAmendments;
 
     // 1.1. Facility Loan Record (FLR) amendment mapping
     const flrMApped = mappings.facility.facilityLoanAmend(amendments, facility, fmr);
 
+    /**
+     * When facility amendment is applied on facility type `Loan`,
+     * with only `amount` attribute execution should be terminated.
```
ttbarnes, line 49: i think this would be clearer

```js
/**
 * When a facility amendment is applied on a `Loan` facility type,
 * if there is no `amount` attribute,
 * execution should be terminated.
 */
```

abhi-markan: Above is incorrect as it is only applied when only the `amount` attribute is present.
Above suggestion is incorrect as if there is `coverEndDate` attribute then FLR record should be mapped.

abhi-markan: I have reworded the documentation.

ttbarnes (approved):

Zainzzkk (approved):

azure-functions/acbs-function/src/functions/acbs-amend-facility-loan-record.js
ChristianMcc: Is this a short term fix? Are we allowing a user in TFM to create an amendment with just amount but then it fails when it gets to ACBS? Might be misunderstanding exactly how this is working so apologies if that's the case

PR #3849, opened 2024-11-04
chore(DTFS2-7608): add rerun to the pipeline

Zainzzkk:
## Introduction :pencil2:
This PR adds a retry module to our GHA test suite to retry failing test steps 3 times.  This should hopefully account for the flaky tests failing on PRs
* Uses github composite action/function
* Consumes env variables, command, retry-limit and working directory - including multiline support
* Retries failing steps up to 3 times until they pass
* Will break out of the retry loop if the test/step passes
* Will retry execute step 3 times as well.

* Added `rerun` module
* Added call to `rerun` module to test.yml
* Added documentation

## Miscellaneous ➕ 
* Dependencies updates.


.github/workflows/test.yml
```diff
@@ -492,20 +502,28 @@ jobs:
...
+        uses: ./.github/modules/retry
+        with:
+          command: |
+            sleep 100s
+            npm run load
+          working-directory: ./utils
+          run: |
+            sleep 100s
+            npm run load
+          env-vars: |
```
abhi-markan, line 524: Can we rename `env-vars` to `env`?

.github/workflows/test.yml
```diff
@@ -492,20 +502,28 @@ jobs:
...
-        working-directory: ./utils
-        env:
-          MONGODB_URI: ${{ secrets.MONGODB_URI_LOCALHOST }}
-        run: |
-          sleep 100s
-          npm run load
-
+        uses: ./.github/modules/retry
+        with:
+          command: |
```
abhi-markan, line 517: Can we not repeat `command` and `run` please.

.github/modules/retry/action.yml
```diff
@@ -0,0 +1,26 @@
...
+    required: false
+    default: "./"
+  env-vars:
+    required: false
+    default: ""
+runs:
+  using: "composite"
+  steps:
+    - run: |
+        limit=3
```
abhi-markan, line 15: Set this to GitHub variables please.

.github/modules/retry/action.yml
```diff
@@ -0,0 +1,26 @@
...
+  command:
+    required: true
+  working-directory:
+    required: false
+    default: "./"
+  env-vars:
+    required: false
+    default: ""
+runs:
+  using: "composite"
```
abhi-markan, line 12: Add some documentation why we are using `composite` in this custom GHA module please.

.github/modules/retry/action.yml
```diff
@@ -0,0 +1,26 @@
...
+  steps:
+    - run: |
+        limit=3
+        run=1
+        env_vars=${{ inputs.env-vars }}
+        until [ $run -gt $limit ]
+        do
+          cd ${{ inputs.working-directory }} && eval "$env_vars" ${{ inputs.command }} && break
```
abhi-markan, line 20: Can these commands be individual lines.

.github/modules/retry/action.yml
```diff
@@ -0,0 +1,26 @@
...
+    required: false
+    default: ""
+runs:
+  using: "composite"
+  steps:
+    - run: |
+        limit=3
+        run=1
+        env_vars=${{ inputs.env-vars }}
+        until [ $run -gt $limit ]
```
abhi-markan, line 18: Can we refactor the loop and conditional branching into one please.

abhi-markan (changes requested):

.github/modules/retry/action.yml
Zainzzkk: added at top of file

.github/modules/retry/action.yml
```diff
@@ -0,0 +1,40 @@
+# Composite function to retry a command for a maximum of 3 times
```
ttbarnes, line 1: Should we maybe start with 2 times?

Concerned that we could be eating up minutes.

Zainzzkk: we're using 3 on EXIPs and it will only go 3 times if something keeps failing.  Most of the time it will retry twice and pass on the second attempt.  I think 2 should be fine too.  @abhi-markan opinion?

AlexBramhill: > Concerned that we could be eating up minutes.

This is a concern of mine too -- especially if it blocks out runners.  I'm not sure if adding retries is the way forward? 

Sorry I'm out the loop here -- I'm assuming this retry the entire test suite, which will massively increase network calls due to existing flakeyness

Is there not a way using jest.retryTimes to only run the failed tests as apposed to the entire suite?

.github/modules/rerun/action.yml
```diff
@@ -38,16 +38,19 @@ runs:
         # Specify environment variables
         eval "$env"
 
+        # Exit immediately upon failure
+        set -e
```
AlexBramhill, line 42: Do we only want this to early exit if its not the last run?

ttbarnes (dismissed):
LGTM

abhi-markan (approved):

AlexBramhill (approved):

ttbarnes (approved):
LGTM great work

PR #3850, opened 2024-11-04
feat(FN-3587): add fee record correction feature flag

RLavender98:
## Introduction :pencil2:
As we build out the fee record correction journey we need to hide the behaviour behind a feature flag.

- Add FF_FEE_RECORD_CORRECTION_ENABLED feature flag
- Reconfigure utilisation reporting e2e-tests for feature flagged testing with the new flag going forwards


trade-finance-manager-ui/component-tests/utilisation-reports/utilisation-report-reconciliation-for-report.component-test.ts
```diff
@@ -179,6 +180,24 @@ describe(page, () => {
...
+    });
+
+    it('should render the create record correction request button when the fee record correction feature flag is enabled', () => {
+      const wrapper = getWrapper({
+        ...params,
+        isFeeRecordCorrectionFeatureFlagEnabled: true,
+      });
+
+      wrapper.expectElement('[data-cy="create-record-correction-request-button"]').toExist();
+    });
```
ttbarnes, line 199: Can we split these into describes please? i.e

```ts
describe('when X feature flag is enabled', () => {
  it('should ...
});

describe('when X feature flag is disabled', () => {
  it('should ...
});
```

trade-finance-manager-ui/server/controllers/utilisation-reports/utilisation-report-reconciliation-for-report/index.test.ts
```diff
@@ -291,6 +299,54 @@ describe('controllers/utilisation-reports/utilisation-report-reconciliation-for-
...
+      // Act
+      await getUtilisationReportReconciliationByReportId(req, res);
+
+      // Assert
+      expect(res._getRenderView()).toEqual('utilisation-reports/utilisation-report-reconciliation-for-report.njk');
+      const viewModel = res._getRenderData() as UtilisationReportReconciliationForReportViewModel;
+      expect(viewModel.isFeeRecordCorrectionFeatureFlagEnabled).toEqual(true);
+    });
+
+    it("renders the page with 'isFeeRecordCorrectionFeatureFlagEnabled' set to false if the feature flag is not enabled", async () => {
```
ttbarnes, line 326: Same as above RE describes

trade-finance-manager-ui/component-tests/utilisation-reports/utilisation-report-reconciliation-for-report.component-test.ts
RLavender98: I'd be against doing this because I think it just splits the test in half and makes it less readable especially as there is only one test per describe in your suggestion so there is also no added benefit to having a commonised test up. It's also not the pattern for the other tests in this file.

trade-finance-manager-ui/server/controllers/utilisation-reports/utilisation-report-reconciliation-for-report/index.test.ts
RLavender98: Left a reply on the your other comment

trade-finance-manager-ui/component-tests/utilisation-reports/utilisation-report-reconciliation-for-report.component-test.ts
ttbarnes: Other instances have this approach, would like to be consitent. I think it's more readable and also helps if other conditions are added for the feature flag.

Screenshot 2024-11-06 at 13 54 31

.github/workflows/test.yml
```diff
@@ -795,10 +798,10 @@ jobs:
     needs: [setup]
     env:
       # Various feature flags
-      FF_TFM_PAYMENT_RECONCILIATION_ENABLED: 'true'
```
sw-mattw, line 798: A bit unsure on why this is being removed - it seems there are still a lot of uses of this flag in the codebase (accessed through `isTfmPaymentReconciliationFeatureFlagEnabled`)

Do we not want all FF set to true here?

trade-finance-manager-ui/server/controllers/utilisation-reports/utilisation-report-reconciliation-for-report/index.test.ts
```diff
@@ -291,6 +299,54 @@ describe('controllers/utilisation-reports/utilisation-report-reconciliation-for-
...
 
+    it("renders the page with 'isFeeRecordCorrectionFeatureFlagEnabled' set to true if the feature flag is enabled", async () => {
+      // Arrange
+      const { req, res } = httpMocks.createMocks({
+        session,
+        params: {
+          reportId,
+        },
+        query: {
+          matchSuccess: 'true',
```
sw-mattw, line 310: Is this needed? Can we omit this query param?

Same as the FF not enabled case below

sw-mattw (commented):
Nice, few queries

.github/workflows/test.yml
RLavender98: Ah sorry, it is being set to 'true' as default on the pipeline since we don't have any off test cases for this specific feature flag we might as well do this now, since want a test with this flag on and the new flag off. (These just override flags that are off by default)

trade-finance-manager-ui/component-tests/utilisation-reports/utilisation-report-reconciliation-for-report.component-test.ts
RLavender98: Added.

sw-mattw (approved):

trade-finance-manager-ui/component-tests/utilisation-reports/utilisation-report-reconciliation-for-report.component-test.ts
RLavender98: I do think this might be good to chat about in dev friday because I personally find it much harder to work with tests with lots of nested describe blocks - I think there is an interesting trade off around maintainability and readability when it comes to describes. The worst times I have had writing tests have all been making large changes to units of code where the tests have been in nested describe blocks.

ttbarnes (approved):

PR #3892, opened 2024-11-05
feat(FN-3163): create new dtfs customer in sf feature branch

natdeanlewissoftwire:
## Introduction :pencil2:
As part of the CCM project, if a customer does exist in Salesforce, we are automatically creating it rather than relying on the manual process currently undertaken by Business Support Group. This change sends a request to a new createParty endpoint that will be merged in to the MDM codebase.

Tech docs of the full process are available [here](https://ukef.sharepoint.com/:w:/r/sites/pmo-PRO211/Shared%20Documents/1.%20TO%20USE%20New%20file%20plan/2.%20Tranche%202/03.%20Customer%20and%20Credit%20Migration/Documents/DTFS%20%26%20Salesforce%20Integration/DTFS%20-%20SF%20Integration%20Technical%20Details.docx?d=w75ec8ca455ec492db2a912cc20cea7e7&csf=1&web=1&e=FO1VI2)

Feature flagged for now:
If the customer lookup fails, send a creation request to MDM.

We also replace the current GET customer in SF (that uses Informatica) with a new one that goes direct (hence the new url with /direct/ in) via the SF API.


external-api/src/v1/controllers/party-db.controller.ts
```diff
@@ -32,3 +34,30 @@ export const lookup = async (req: Request, res: Response) => {
...
+export const createParty = async (req: CustomExpressRequest<{ reqBody: { companyName: string } }>, res: Response) => {
+  const { partyDbCompanyRegistrationNumber: companyReg } = req.params;
+  const { companyName } = req.body;
+
+  if (!isValidCompanyRegistrationNumber(companyReg)) {
+    console.error('Invalid company registration number provided %s', companyReg);
+    return res.status(HttpStatusCode.BadRequest).send({ status: HttpStatusCode.BadRequest, data: 'Invalid company registration number' });
+  }
+
+  const response: { status: number; data: unknown } = await axios({
```
natdeanlewissoftwire, line 47: the response needs a status (following the pattern elsewhere in dtfs) so that we can distinguish between a 404 and an actual failure when querying SF - we don't want to create customers unless we definitely have a 404

trade-finance-manager-api/src/v1/api.js
```diff
@@ -820,13 +821,35 @@ const getPartyDbInfo = async ({ companyRegNo }) => {
       url: `${EXTERNAL_API_URL}/party-db/${encodeURIComponent(companyRegNo)}`,
       headers: headers.external,
     });
-    return response.data;
+    return AUTOMATIC_SF_CUSTOMER_CREATION_ENABLED ? { status: 200, data: response.data } : response.data;
```
natdeanlewissoftwire, line 824: the response needs a status (following the pattern elsewhere in dtfs) so that we can distinguish between a 404 and an actual failure when querying SF - we don't want to create customers unless we definitely have a 404

trade-finance-manager-api/src/v1/controllers/deal.party-db.js
```diff
@@ -60,7 +77,7 @@ const addPartyUrns = async (deal, auditDetails) => {
       ...deal.tfm,
       parties: {
         exporter: {
-          partyUrn: await getPartyUrn({ companyRegNo: deal.exporter.companiesHouseRegistrationNumber }),
+          partyUrn: await getPartyUrn({ companyRegNo: deal.exporter.companiesHouseRegistrationNumber, companyName: deal.exporter.companyName }),
```
natdeanlewissoftwire, line 80: I've not flagged this as that seems unnecessarily fiddly, let me know if I should!

external-api/src/v1/controllers/party-db.controller.ts
```diff
@@ -21,7 +21,9 @@ export const lookup = async (req: Request, res: Response) => {
 
   const response: { status: number; data: unknown } = await axios({
     method: 'get',
-    url: `${APIM_MDM_URL}customers?companyReg=${companyReg}`,
+    url: AUTOMATIC_SF_CUSTOMER_CREATION_ENABLED
+      ? `${APIM_MDM_URL}customers/direct?companyRegistrationNumber=${companyReg}`
```
natdeanlewissoftwire, line 25: This uses the new direct SF api in MDM, old route is Informatica

gef-ui/server/services/api.js
```diff
@@ -113,7 +113,7 @@ const setApplicationStatus = async ({ dealId, status, userToken }) => {
       {
         status,
       },
-      { ...config(userToken), timeout: 10000 },
```
natdeanlewissoftwire, line 116: timeout required as sometimes this takes just over 10 seconds (at least on my local build, wouldn't be good if this only worked on prod). Without this the user sees a "something wrong with the service page" even though there actually isn't, we've just got a lot of chained API calls on submission, to which this PR adds one. According to the root README:

> When a deal is submitted to TFM, many external API calls are currently made in the TFM submission controller. This process can be slow, resource-intensive, and prone to failures if one of the API calls encounters an issue. Retries are not currently configured.

> To address these issues, the plan is to move all these API calls into background processes with automatic retries. This will enhance the user experience, make the process fail-safe, and improve the development workflow.

However, that work isn't in the scope of this PR of course

trade-finance-manager-api/src/v1/controllers/deal.party-db.js
```diff
@@ -28,11 +28,28 @@ const getCompany = async (req, res) => {
...
+      }
+
+      partyDbInfo = await api.createParty({ companyRegNo, companyName });
+    }
+
+    if (partyDbInfo.status !== 200) {
+      return '';
+    }
+
+    return partyDbInfo.data[0].partyUrn || '';
```
ChristianMcc, line 51: This is accessing data from getPartyDbInfo differently to line 58 which is calling the same thing, is this going to cause any issues?

trade-finance-manager-api/src/v1/api.js
```diff
@@ -820,13 +821,35 @@ const getPartyDbInfo = async ({ companyRegNo }) => {
       url: `${EXTERNAL_API_URL}/party-db/${encodeURIComponent(companyRegNo)}`,
       headers: headers.external,
     });
-    return response.data;
+    return AUTOMATIC_SF_CUSTOMER_CREATION_ENABLED === 'true' ? { status: 200, data: response.data } : response.data;
```
ChristianMcc, line 824: So while this means there will be a change that isn't FF'd, is it worth just making this behaviour change regardless of the FF? It's a little surprising to have different return types like this depending on a state of the FF

trade-finance-manager-api/src/v1/controllers/deal.party-db.js
ChristianMcc: Ah I saw the other part of this and left a comment there instead

gef-ui/server/services/api.js
ChristianMcc: Okay makes sense, I think given we are already close to the timeout can we make this 20 seconds instead of 30? We might want to add some extra documentatoin in the form of a JS-doc to the top of this function explaining that this goes off to do lots of different (long) API calls and that's why the timeout is very long

external-api/src/v1/controllers/party-db.controller.ts
ChristianMcc: Could we call this route `customers/salesforce` rather than `customers/direct`, makes it a little clearer/less tied to the fact that an endpoint already exists

ChristianMcc (commented):
Nice work for first PR! Left a few comments and will go through in some more detail soon :)

external-api/src/v1/controllers/party-db.controller.ts
natdeanlewissoftwire: Sure, I agree /salesforce is slightly clearer, will update my MDM work to match also

trade-finance-manager-api/src/v1/api.js
natdeanlewissoftwire: I don't think it should really matter, I'd rather not change the behaviour of the existing endpoint in a feature flagged PR, but I'll add tests to ensure the update in response type depending on the flag if you're happy with that compromise?

ChristianMcc: Mm, yeah I'm sold on that that sounds fine to me - thanks :)

trade-finance-manager-api/src/v1/controllers/deal.party-db.js
```diff
@@ -28,11 +28,28 @@ const getCompany = async (req, res) => {
   }
 };
 
-const getPartyUrn = async ({ companyRegNo }) => {
+const getPartyUrn = async ({ companyRegNo, companyName }) => {
```
ChristianMcc, line 31: testing wise could we add some tests for this spying on what is called and making sure the right things are called for FF on and off please?

external-api/src/v1/controllers/party-db.controller.ts
```diff
@@ -21,11 +21,40 @@ export const lookup = async (req: Request, res: Response) => {
...
-    return { data: 'Failed to call Party DB API', status: error?.response?.status || HttpStatusCode.InternalServerError };
+    return { data: 'Error calling Party DB API', status: error?.response?.status || HttpStatusCode.InternalServerError };
+  });
+
+  const { status, data } = response;
+
+  return res.status(status).send(data);
+};
+
+export const createParty = async (req: CustomExpressRequest<{ reqBody: { companyName: string } }>, res: Response) => {
```
ChristianMcc, line 38: aware the testing in this area is not great atm so you might have to set a couple of things up yourself, but testing different inputs and the validation of company registration number etc. would be good here!

ChristianMcc (commented):
Couple of extra comments around tests to add, little worried about what testing is currently in place so let me know if you're having to do a lot of setup to add a few simple tests

.env.sample
```diff
@@ -116,5 +116,6 @@ CLAMAV_SCANNING_ENABLED=false
 FF_TFM_PAYMENT_RECONCILIATION_ENABLED=false
 FF_TFM_FACILITY_END_DATE_ENABLED=false
 FF_TFM_DEAL_CANCELLATION_ENABLED=false
+AUTOMATIC_SF_CUSTOMER_CREATION_ENABLED=false
```
ttbarnes, line 119: I think we should have `SALESFORCE` instead of `SF`, initially, I didn't know what SF meant

external-api/api-tests/v1/party-db.api-test.ts
```diff
@@ -19,12 +19,69 @@ import { api } from '../api';
...
+  axiosMock.onGet(`${APIM_MDM_URL}customers?companyReg=${VALID_WITH_LETTERS}`).reply(HttpStatusCode.Ok, {});
+  axiosMock.onGet(`${APIM_MDM_URL}customers/salesforce?companyRegistrationNumber=${VALID}`).reply(HttpStatusCode.Ok, {});
+  axiosMock.onGet(`${APIM_MDM_URL}customers/salesforce?companyRegistrationNumber=${VALID_WITH_LETTERS}`).reply(HttpStatusCode.Ok, {});
+});
+
+afterEach(() => {
+  axiosMock.resetHistory();
+});
+
+describe('when AUTOMATIC_SF_CUSTOMER_CREATION_ENABLED is not true', () => {
```
ttbarnes, line 37: Can we update throughout to be more aligned with other feature flags tests e.g

```ts
describe('when TFM Facility end date feature flag is enabled', () => { ... }
describe('when TFM Facility end date feature flag is disabled', () => { ... }
```

I do like the use of the env var here, much better IMO, but better to be consistent

external-api/api-tests/v1/party-db.api-test.ts
```diff
@@ -19,12 +19,69 @@ import { api } from '../api';
...
+  beforeAll(() => {
+    process.env.AUTOMATIC_SF_CUSTOMER_CREATION_ENABLED = 'false';
+  });
```
ttbarnes, line 40: Can we add a new line after this please

external-api/src/v1/controllers/party-db.controller.ts
```diff
@@ -21,11 +21,40 @@ export const lookup = async (req: Request, res: Response) => {
...
```
ttbarnes, line 38: Can we add some documentation please

external-api/src/v1/routes/external-apis.route.ts
```diff
@@ -358,6 +358,43 @@ apiRoutes.post('/acbs/facility/:id/amendments', acbs.amendAcbsFacilityPost);
...
+ *     tags: [APIM, Salesforce]
+ *     description: We only consume the Companies House number and company name. Not all fields are in the response example.
+ *     parameters:
+ *       - in: path
+ *         name: partyDbCompanyRegistrationNumber
+ *         schema:
+ *           type: string
+ *           example: '12341234'
+ *         required: true
+ *         description: Companies House Registration Number to create the UKEF Party
```
ttbarnes, line 375:
```suggestion
 *         description: Companies House Registration Number for UKEF Party creation
```

gef-ui/server/services/api.js
ttbarnes: The timeout increase concerns me - how long does deal submission currently take with a production build?

I am wondering if we should create a CRON job for this functionality instead of increasing submission time.

Depending on the timings and if the increase isn't that bad, perhaps this could be something we could look into as a future iteration (maybe an "improve submission time/performance" epic pending business priorities etc.

trade-finance-manager-api/src/v1/__mocks__/api.js
```diff
@@ -180,14 +180,27 @@ module.exports = {
...
+    }
+    return companyRegNo === 'NO_MATCH'
       ? false
       : [
           {
             partyUrn: 'testPartyUrn',
           },
-        ],
+        ];
+  },
```
ttbarnes, line 203: The returns here are inconsistent, what's the reason for this please?

From line 196 we don't have `status` or `data` properties?

trade-finance-manager-api/src/v1/controllers/deal.party-db.js
```diff
@@ -28,11 +28,27 @@ const getCompany = async (req, res) => {
...
```
ttbarnes, line 31: Can we add some documentation please

trade-finance-manager-api/src/v1/controllers/deal.party-db.js
```diff
@@ -28,11 +28,27 @@ const getCompany = async (req, res) => {
...
   if (!companyRegNo) {
     return '';
   }
 
+  if (process.env.AUTOMATIC_SF_CUSTOMER_CREATION_ENABLED === 'true') {
```
ttbarnes, line 36: Are we able to use `isFeatureFlagEnabled` (in `libs/common/src/helpers/is-feature-flag-enabled.ts`) throughout?

trade-finance-manager-api/src/v1/controllers/deal.party-db.js
ttbarnes: What do you mean by flagged please?

trade-finance-manager-api/src/v1/controllers/deal.party-db.test.js
```diff
@@ -53,3 +53,174 @@ describe('getCompany returns company', () => {
...
+  
+    it('should call getPartyDbInfo and return urn', async () => {
+      const { getPartyDbInfo, createParty } = require('../api.js');
+      getPartyDbInfo.mockResolvedValue([{ partyUrn: 'TEST_URN' }]);
+  
+      const companyData = { companyRegNo: '12345678' };
+      
+      const result = await api.getPartyUrn(companyData);
+  
+      expect(getPartyDbInfo).toHaveBeenCalledWith(companyData);
```
ttbarnes, line 86: Can we add a `toHaveBeenCalledTimes` assertion please - same in other tests

trade-finance-manager-api/src/v1/controllers/deal.party-db.test.js
```diff
@@ -53,3 +53,174 @@ describe('getCompany returns company', () => {
...
+  
+      expect(createParty).toHaveBeenCalledTimes(0);
```
ttbarnes, line 88: Can we move this into it's own test please i.e "should not call ..." - same in other tests

ttbarnes (changes requested):
Nice work 👍

trade-finance-manager-api/src/v1/controllers/deal.party-db.js
abhi-markan, line 31: Why this function is being changed, only change I am expecting is at `APIM` end none at DTFS for a `GET` endpoint. Also why would we need to send `companyName` in argument?

abhi-markan (changes requested):

trade-finance-manager-api/src/v1/api.js
```diff
@@ -820,13 +821,35 @@ const getPartyDbInfo = async ({ companyRegNo }) => {
       url: `${EXTERNAL_API_URL}/party-db/${encodeURIComponent(companyRegNo)}`,
       headers: headers.external,
     });
-    return response.data;
+    return process.env.AUTOMATIC_SF_CUSTOMER_CREATION_ENABLED === 'true' ? { status: 200, data: response.data } : response.data;
```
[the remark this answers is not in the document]
ttbarnes: Would be ideal to use `isFeatureFlagEnabled` here for the if statement

trade-finance-manager-api/src/v1/api.js
```diff
@@ -820,13 +821,35 @@ const getPartyDbInfo = async ({ companyRegNo }) => {
...
   } catch (error) {
     console.error('Unable to get party DB info %o', error);
+    if (process.env.AUTOMATIC_SF_CUSTOMER_CREATION_ENABLED === 'true') {
```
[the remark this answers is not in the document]
ttbarnes: `isFeatureFlagEnabled` would be ideal

trade-finance-manager-api/src/v1/__mocks__/api.js
natdeanlewissoftwire: So previously we didn't need to return a status code because the old GET method that checks for a customer existing in Salesforce via Informatica didn't distinguish between e.g. server error and a 404. In either case the BSG would have to check if a new customer had to be created, and create it if so. 

In the new implementation, we need to be able to tell if we receive a 404 when looking for the customer in Salesforce, because only in that case do we want to create a new customer. (We don't want to create a new one if we've just got a server error as the customer may well exist in Salesforce already).

gef-ui/server/services/api.js
natdeanlewissoftwire: I think there's a plan to move the deal submission to a background task long-term, and I agree that's a good idea. I'm not sure if falls in the scope of this PR though, all the deal submission tasks should be included in that, not just this one. I'm not sure how long this will take on a production build, though I'd hope it would come in under 10 seconds (it's only slightly over on my local build).

trade-finance-manager-api/src/v1/api.js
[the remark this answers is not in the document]
natdeanlewissoftwire: Sure, have done - though worth noting that every other status code in this file is just a raw number, might be worth a tech debt ticket to refactor the rest of it. Feels messy for me to do that here as it would add to the diff quite substantially

trade-finance-manager-api/src/v1/controllers/deal.party-db.js
natdeanlewissoftwire: The company name is a required field for Salesforce account creation, so we pass it through here. this means we're not dependent on doing a later check by companies house number to pull the company name. We _could_ get it from a check MDM runs to the Dun and Bradstreet API, but that felt needless to me when we already have it in DTFS.

trade-finance-manager-api/src/v1/api.js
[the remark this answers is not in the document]
natdeanlewissoftwire: Sure, have done - though worth noting that every other status code in this file is a raw number. Might be worth a tech debt ticket to refactor the rest of them at some point, though it doesn't feel like a good idea for me to do that in this PR as that would expand the diff considerably. Let me know if you disagree!

trade-finance-manager-api/src/v1/api.js
```diff
@@ -820,13 +821,35 @@ const getPartyDbInfo = async ({ companyRegNo }) => {
...
+      if (error?.status === 404) {
+        return { status: 404, data: 'Party not found' };
+      }
+      return { status: error?.status || 500, data: 'Failed to get party' };
+    }
     return false;
   }
 };
 
+const createParty = async ({ companyRegNo, companyName }) => {
```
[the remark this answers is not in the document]
abhi-markan: Thanks

trade-finance-manager-api/src/v1/controllers/deal.party-db.js
abhi-markan: Understood, will discuss other questions separately.

trade-finance-manager-api/src/v1/__mocks__/api.js
ttbarnes: Thanks @natdeanlewissoftwire - seems to make sense.

What do you think about changing the function to be more if/else based? I know it's more lines but personally I think would be easier to read/maintain, e.g

```js
getPartyDbInfo: ({ companyRegNo }) => {
  const noCompanyMatch = companyRegNo === 'NO_MATCH';

  if (isAutomaticSalesforceCustomerCreationFeatureFlagEnabled()) {
    if (noCompanyMatch) {
      return { status: 404, data: 'Party not found' };
    }
    
    return {
      status: 200,
      data: [
        {
          partyUrn: 'testPartyUrn',
        },
      ],
    };
  } else {
    if (noCompanyMatch) {
      return false;
    }

    return {
      partyUrn: 'testPartyUrn',
    };
  }
}
```

trade-finance-manager-api/src/v1/controllers/deal.party-db.js
```diff
@@ -28,11 +29,27 @@ const getCompany = async (req, res) => {
   }
 };
 
-const getPartyUrn = async ({ companyRegNo }) => {
+const getPartyUrn = async ({ companyRegNo, companyName }) => {
   if (!companyRegNo) {
     return '';
   }
 
+  if (isAutomaticSalesforceCustomerCreationFeatureFlagEnabled()) {
```
abhi-markan, line 37: Is this required for get Pary URN?

gef-ui/server/services/api.js
abhi-markan: I agree with @ChristianMcc  we should increase this to `30,000ms` for fail-safe and in my current PR which tackles this issue I will speak to the UCD to temporarily show a processing page.

Long-term I would refer to Azure functions like approach where multiple external API calls can be handled and if either fails an alert is dispatched and reverting the deal status.

Thanks @natdeanlewissoftwire for a precise and concise explanation and agree with @ChristianMcc  if above can be added to DocString please.

abhi-markan (changes requested):
Nice work, few minor comments to address please.

trade-finance-manager-api/src/v1/api.js
```diff
@@ -820,10 +825,43 @@ const getPartyDbInfo = async ({ companyRegNo }) => {
       url: `${EXTERNAL_API_URL}/party-db/${encodeURIComponent(companyRegNo)}`,
       headers: headers.external,
     });
-    return response.data;
+    if (isAutomaticSalesforceCustomerCreationFeatureFlagEnabled()) {
```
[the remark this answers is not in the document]
natdeanlewissoftwire: Good thought - and yep it's needed, same reason as in this thread https://github.com/UK-Export-Finance/dtfs2/pull/3892#discussion_r1856522670

external-api/api-tests/v1/party-db.api-test.ts
```diff
@@ -12,19 +12,82 @@
...
 const { get } = api(app);
+let axiosMock: MockAdapter;
+
+jest.mock('@ukef/dtfs2-common', () => ({
+  ...jest.requireActual('@ukef/dtfs2-common'),
+  isAutomaticSalesforceCustomerCreationFeatureFlagEnabled: jest.fn(),
+}));
+
+beforeEach(() => {
+  axiosMock = new MockAdapter(axios);
```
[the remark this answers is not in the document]
natdeanlewissoftwire: No problem, just inherited the mocking from the existing (admittedly minimal) testing in this file. Is there a standard process in DTFS for setting up objects externally for testing?
If I want to create a customer in Salesforce to test that the GET returns 200 for a customer that exists, I'm assuming I should be setting up that customer within the test, and deleting it afterwards. There's not currently a way of deleting customers in SF via APIM, so that feels like a lot of extra code for this small change. Should I instead just set up a long-running customer manually in the QA/test environment to GET and just hope nobody comes along and deletes it? Wouldn't work in prod of course so then we'd have to xit this test in prod - all of which is to say it feels like maybe this wouldn't provide much value and mocking might be the right approach after all, what do you think @abhi-markan ?

PR #3893, opened 2024-11-05
feat(FN-3569): add error state to page titles

sw-mattw:
## Introduction :pencil2:
The title of bank report pages in TFM should be prefixed by `Error -` when there is an error state on those pages.

- Added nunjucks filter to add the error prefix if errors are defined.
- Updated all utilisation report templates to make use of this filter, passing in a flag to indicate if there are page errors.

N/A


trade-finance-manager-ui/component-tests/assertions.ts
```diff
@@ -132,4 +132,9 @@ export const assertions = <TParams extends object>(wrapper: CheerioAPI, html: st
       expect(wrapper(selector).attr('aria-sort')).toEqual(text);
     },
   }),
+  expectPageTitle: () => ({
+    toRead: (text: string) => {
+      expect(wrapper('title').text().replace(', Trade Finance Manager, gov.uk', '').trim()).toEqual(text);
```
sw-mattw, line 137: Adding in the `replace` allows us to not repeat the given string (which is added in the nunjucks index template) in each call to this assertion.

Alternatively we could change this to be a `contains` assertion.

Open to thoughts

trade-finance-manager-ui/component-tests/utilisation-reports/add-payment.component-test.ts
sw-mattw, line 1: The odd changes further down this file relate to the JS to TS migration

trade-finance-manager-ui/component-tests/assertions.ts
```diff
@@ -107,7 +107,7 @@ export const assertions = <TParams extends object>(wrapper: CheerioAPI, html: st
     },
   }),
   expectInput: (selector: string) => ({
-    toHaveValue: (value: string) => {
+    toHaveValue: (value?: string) => {
```
ttbarnes, line 110: Can the param remain required?

trade-finance-manager-ui/component-tests/assertions.ts
ttbarnes: Strict `toEqual` is better, contains is risky

Could we create a `result` const please? Easier to read IMO

```ts
expect(result).toEqual(text);
```

trade-finance-manager-ui/server/nunjucks-configuration/index.js
```diff
@@ -57,6 +57,7 @@ const configureNunjucks = (opts) => {
   nunjucksEnvironment.addFilter('userIsInTeam', (user, teamIdList) => userIsInTeam(user, teamIdList));
   nunjucksEnvironment.addFilter('userIsOnlyInTeams', (user, teamIdList) => userIsOnlyInTeams(user, teamIdList));
   nunjucksEnvironment.addFilter('removeCommasAndFloat', (value) => parseFloat(value.replace(/,/g, '')));
+  nunjucksEnvironment.addFilter('formatPageTitleWithError', (title, pageHasErrors) => (pageHasErrors ? `Error - ${title}` : title));
```
ttbarnes, line 60:
```suggestion
  nunjucksEnvironment.addFilter('formatPageTitle', (title, pageHasErrors) => (pageHasErrors ? `Error - ${title}` : title));
```

trade-finance-manager-ui/templates/utilisation-reports/add-payment.njk
```diff
@@ -9,9 +9,14 @@
 {% import "./_macros/payment-date-input.njk" as paymentDateInput %}
 {% extends "index.njk" %}
 {% set backLink = { "href": backLinkHref } %}
-{% block pageTitle %}Add a payment{% endblock %}
+{% set pageHasErrors = errors.errorSummary | length %}
```
ttbarnes, line 12: Would this work?
```suggestion
{% set validationErrors = errors.errorSummary %}
```

Then we could replace instances of `errors.errorSummary` e.g

```
{{ govukErrorSummary({
  titleText: "There is a problem",
  errorList: validationErrors
```

ttbarnes (changes requested):

trade-finance-manager-ui/component-tests/assertions.ts
sw-mattw: This is from the JS->TS migration, tried leaving this as is initially, but the input value the component tests return when no value is given is `undefined`.

We could have an additional function on `expectInput`, e.g. `toHaveAnUndefinedValue`, but I'm not super keen on that. 

Some debug after changing the expected values to empty string to match the typing (instead of undefined):
Screenshot 2024-11-05 at 13 54 50
Screenshot 2024-11-05 at 13 54 55

trade-finance-manager-ui/component-tests/assertions.ts
sw-mattw: Pulled this out thanks

trade-finance-manager-ui/templates/utilisation-reports/add-payment.njk
ttbarnes: Thanks @sw-mattw I think we should keep it simple and go with the following instead - existing approach in this PR but renamed to this:
```
{% set hasValidationErrors = errors.errorSummary | length %}
```

There's some misalignment with naming e.g `errors.errorSummary`, `validationErrors.summary`, `errorSummary` so would like to keep it simple. What do you think?

trade-finance-manager-ui/component-tests/utilisation-reports/add-payment.component-test.ts
```diff
@@ -1,10 +1,27 @@
...
+  });
+
+  it('should not add error prefix to page title when there are no errors', () => {
+    const addPaymentViewModel = anAddPaymentViewModel();
+    addPaymentViewModel.errors.errorSummary = [];
+    const wrapper = render(addPaymentViewModel);
+
+    wrapper.expectPageTitle().toRead('Add a payment');
+  });
+
```
AlexBramhill, line 24: (just as a knowledge share, resolve once read, no action needed)

We could have these as `withErrorPrefixInPageTitleTests.ts`
```
export const withErrorPrefixInPageTitleTests = ({aViewModel, defaultPageTitle}) => {
  it('should add error prefix to page title when there are errors', () => {
    const viewModel = aViewModel();
    viewModel.errors.errorSummary = [{ text: 'an error', href: 'error-href' }];
    const wrapper = render(viewModel);

    wrapper.expectPageTitle().toRead(`Error - ${defaultPageTitle}`);
  });

  it('should not add error prefix to page title when there are no errors', () => {
    const viewModel = aViewModel();
    viewModel.errors.errorSummary = [];
    const wrapper = render(viewModel);

    wrapper.expectPageTitle().toRead(defaultPageTitle);
  });
}
``` 

and then for these tests, we wouldnt have to repeat this and we could just have
```
withErrorPrefixInPageTitleTests({
    aViewModel: anAddPaymentViewModel,
    defaultPageTitle: 'Add a payment'
})
```
in our 4 test files -- removing a fair chunk of repeated tests

trade-finance-manager-ui/server/nunjucks-configuration/index.js
```diff
@@ -57,6 +57,7 @@ const configureNunjucks = (opts) => {
   nunjucksEnvironment.addFilter('userIsInTeam', (user, teamIdList) => userIsInTeam(user, teamIdList));
   nunjucksEnvironment.addFilter('userIsOnlyInTeams', (user, teamIdList) => userIsOnlyInTeams(user, teamIdList));
   nunjucksEnvironment.addFilter('removeCommasAndFloat', (value) => parseFloat(value.replace(/,/g, '')));
+  nunjucksEnvironment.addFilter('formatPageTitle', (title, pageHasErrors) => (pageHasErrors ? `Error - ${title}` : title));
```
AlexBramhill, line 60: Nice job here -- Right tool for the job!

trade-finance-manager-ui/templates/utilisation-reports/add-payment.njk
AlexBramhill: Just to add to Tony's comment -- I tend to like renaming like this [to make it clear at a glance it's a boolean](https://dev.to/michi/tips-on-naming-boolean-variables-cleaner-code-35ig), although if there's a reason for this to need the word `page` i'd be interested to hear it :) `hasValidationErrorsOnPage` could be an alt name if we need it -- but I'm not sure we do

AlexBramhill (approved):
lgtm after tony's renaming request

trade-finance-manager-ui/templates/utilisation-reports/add-payment.njk
ttbarnes: Thing is, I don't think there'd ever be an instance where there are validation errors, _not_ on the page, in page

sw-mattw: Simplified to `hasErrors` as we should add the prefix when there are any errors, and at the minute all errors on these pages are likely to only be validation errors

ttbarnes (approved):

PR #3851, opened 2024-11-04
chore(DTFS2-7607): replace hard coded bondType strings with constants

ttbarnes:
## Introduction :pencil2:
Various tests and functions had a hard coded `bondType` string.

This PR replaces these with constants.

## Resolution :heavy_check_mark:
- Remove unused/misleading  `BSS_FACILITY_TYPE` `e2e` constant.
- Remove unused/misleading  `TYPE` `portal-api` constant.
- Add `BOND_TYPE` constnats to `libs/common`.
- Replace all hard coded `bondType` strings with constants.

## Miscellaneous :heavy_plus_sign:
- Add missing environment variable to `.env.sample`.
- Update Portal constants to be ordered alphabetically.
- Replace an import of `roundNumber` from `portal-api` into E2E tests.
- Remove an E2E test directory that has been skipped for a very long time, due to number-generator/GHA issues.

PR #3896, opened 2024-11-05
feat(FN-3608): use utilisation report event type constant

sw-mattw:
## Introduction :pencil2:
Historically we have used string literals for utilisation report state machine event types, e.g. `'REPORT_UPLOADED'`. Instead we should use the `UTILISATION_REPORT_EVENT_TYPE` constant to represent these, e.g. `UTILISATION_REPORT_EVENT_TYPE.REPORT_UPLOADED`.

- Update string literal uses to use the `UTILISATION_REPORT_EVENT_TYPE` constant.

N/A


RLavender98 (approved):

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/delete-payment/delete-payment.event-handler.ts
```diff
@@ -4,14 +4,15 @@ import { NotFoundError } from '../../../../../errors';
...
+import { UTILISATION_REPORT_EVENT_TYPE } from '../../event/utilisation-report.event-type';
 
 type DeletePaymentEventPayload = {
   transactionEntityManager: EntityManager;
   paymentId: number;
   requestSource: DbRequestSource;
 };
 
-export type UtilisationReportDeletePaymentEvent = BaseUtilisationReportEvent<'DELETE_PAYMENT', DeletePaymentEventPayload>;
+export type UtilisationReportDeletePaymentEvent = BaseUtilisationReportEvent<typeof UTILISATION_REPORT_EVENT_TYPE.DELETE_PAYMENT, DeletePaymentEventPayload>;
```
ttbarnes, line 15: Could we update `report.event-type.ts` to export each constant so in each instance it can be e.g the following?:

```ts
import { DELETE_PAYMENT } from '../../event/utilisation-report.event-type';

export type UtilisationReportDeletePaymentEvent = BaseUtilisationReportEvent<typeof DELETE_PAYMENT, DeletePaymentEventPayload>;

await stateMachine.handleEvent({
  type: ADD_A_PAYMENT,
});

await stateMachine.handleEvent({
  type: MANUALLY_SET_COMPLETED,
});
```

ttbarnes (changes requested):

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/delete-payment/delete-payment.event-handler.ts
RLavender98: What is the pattern we want for constants going forwards generally? e.g. saw on your PR the other day Tony for bondtype you didn't do this: https://github.com/UK-Export-Finance/dtfs2/pull/3851/files. Just keen to understand what is expected going forwards so that we can do the correct thing upfront and not have to rework in review!

ttbarnes: I think the preference is to have the usages as simple as possible so we don't have long lines of constants.

My PR is a bit different because we already have other facility constants:

```ts
export const GEF_FACILITY_TYPE = {};
export const BSS_EWCS_FACILITY_TYPE = {};
```

I added `BOND_FACILITY_TYPE` with the same structure, to be consistent, otherwise they will have different patterns and I worry it could be confusing and losing context, especially as it's a very specific type of bond that isn't technically a `BSS_EWCS_FACILITY_TYPE`.

ChristianMcc: I'm just going to be re-iterating my old comment so if it lands flat it lands flat. 
If we have a pattern of using grouping of constants and then referencing them from their grouping to make it clear what it's coming from is that not a good thing to keep that context? Keeping some context in return for some extra characters in a line of code just seems like a good trade off to me

ttbarnes: Why do we want to differentiate seemingly generic event types? I don't see why these for example should be specific to X state machine or Y state machine, surely they'd have the same value?

```ts
ADD_A_PAYMENT: 'ADD_A_PAYMENT',
DELETE_PAYMENT: 'DELETE_PAYMENT',
REPORT_UPLOADED: 'REPORT_UPLOADED',
EDIT_PAYMENT: 'EDIT_PAYMENT',
```

As in, state machine Y surely wouldn't be this for example:
```ts
DELETE_PAYMENT: 'DELETE_A_PAYMENT'
```

sw-mattw: Yep the values would be the same for shared event types

If we're grouping things up anyway, is it not beneficial to have the additional context that the named constant offers still (while remaining at a reasonable length)? I feel like it's nicer having that context, allowing you to immediately know which event types are available for that state machine rather than having to look around for all the individually exported event types

AlexBramhill (commented):
Looking good -- This review I've just done checks there were no errors with the change to constants (ie swapping one out for another) so no further reviews needed to check that.

AlexBramhill (approved):

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/delete-payment/delete-payment.event-handler.ts
sw-mattw: @ttbarnes would it be okay to keep this naming as is?

ttbarnes: I am happy with either approach if @Zainzzkk is.

I would prefer shorter usage of constants, but appreciate there is concern that this could cause confusion when there are multiple state machines in play.

Zainzzkk: I am happy with how it is.  
Like tony said, it would be nicer if the names were shorter as UTILISATION_REPORT_EVENT_TYPE.DELETE_PAYMENT is quite long and maybe overdescriptive

ttbarnes (approved):

PR #3897, opened 2024-11-06
fix(DTFS2-7604): access _id deal property

abhi-markan:
## Introduction :pencil2:
When dispatching first amendment task email, deal ID in the email is `undefined`.

* Access `_id` for Mongo Object deal ID.
* Added validation for imperative mandatory arguments.
* Added exceptional handling to `TFM-UI` case render.

* Improved test coverage with additional mocks.
* Improved documentation.

### DIFF
--- a/trade-finance-manager-api/src/v1/__mocks__/amendmentVariables.js
+++ b/trade-finance-manager-api/src/v1/__mocks__/amendmentVariables.js
@@ -11,6 +11,7 @@ const conditions = 'ABCD';
 const declined = 'EFGH';
 
 const approvedWithoutConditionsBothAmendments = {
+  _id: 'test',
   user: {
     firstname: 'Bob',
     surname: 'Smith',
@@ -217,6 +218,20 @@ const noTaskVariables = {
   },
 };
 
+const noCompanyNameDeal = {
+  ...firstTaskVariables,
+  dealSnapshot: {
+    ukefDealId,
+    exporter: {},
+    bankInternalRefName: 'UKEF',
+  },
+};
+
+const noDealId = {
+  ...firstTaskVariables,
+  _id: null,
+};
+
 module.exports = {
   approvedWithoutConditionsBothAmendments,
   approvedWithoutConditionsBothAmendmentsBSS,
@@ -234,4 +249,6 @@ module.exports = {
   wrongAmendments,
   firstTaskVariables,
   noTaskVariables,
+  noDealId,
+  noCompanyNameDeal,
 };
--- a/trade-finance-manager-api/src/v1/helpers/amendment.helpers.js
+++ b/trade-finance-manager-api/src/v1/helpers/amendment.helpers.js
@@ -20,6 +20,8 @@ const dealTypeCoverStartDate = require('./dealTypeCoverStartDate.helper');
 const { formattedNumber } = require('../../utils/number');
 const { decimalsCount, roundNumber } = require('./number');
 
+const { TFM_UI_URL } = process.env;
+
 // checks if amendment exists and if eligible to send email
 const amendmentEmailEligible = (amendment) =>
   amendment &&
@@ -310,24 +312,50 @@ const firstTaskEmailConfirmation = async (facilityId, amendmentId, auditDetails)
   await api.updateFacilityAmendment(facilityId, amendmentId, payload, auditDetails);
 };
 
-// sends email for first amendment task when pim submit amendment
-const sendFirstTaskEmail = async (taskVariables, auditDetails) => {
-  const { amendment, dealSnapshot, facilityId, amendmentId } = taskVariables;
+/**
+ * Sends an email notification for the first task of an amendment.
+ *
+ * This function performs the following operations:
+ * 1. Extracts necessary details from the deal and amendment.
+ * 2. Retrieves the first task from the amendment tasks.
+ * 3. Constructs the email variables and sends the email using the TFM email service.
+ * 4. Updates the flag to indicate that the email has been sent if the email is successfully sent.
+ *
+ * @param {Object} deal - The deal object containing deal and amendment details.
+ * @param {Object} deal.amendment - The amendment details.
+ * @param {Object} deal.dealSnapshot - The snapshot of the deal details.
+ * @param {string} deal.facilityId - The ID of the facility.
+ * @param {string} deal.amendmentId - The ID of the amendment.
+ * @param {Object} deal.dealSnapshot.exporter - The exporter details.
+ * @param {string} deal.dealSnapshot.exporter.companyName - The name of the exporter company.
+ * @param {string} [deal.dealSnapshot.ukefDealId] - The UKEF deal ID for GEF deals.
+ * @param {Object} [deal.dealSnapshot.details] - The details object for BSS deals.
+ * @param {string} [deal.dealSnapshot.details.ukefDealId] - The UKEF deal ID for BSS deals.
+ * @param {Object} auditDetails - The audit details for logging purposes.
+ * @returns {Promise<void>} - A promise that resolves when the email has been sent and the flag has been updated.
+ * @throws {Error} - Logs the error if any error occurs during the email sending process.
+ * */
+const sendFirstTaskEmail = async (deal, auditDetails) => {
+  const { amendment, dealSnapshot, facilityId, amendmentId } = deal;
   const { tasks } = amendment;
-  const { dealId, exporter } = dealSnapshot;
+  const { _id: dealId } = deal;
+  const { exporter } = dealSnapshot;
+  const { companyName } = exporter;
 
   // dealId in snapshot for gef and details for bss
-  const ukefDealId = dealSnapshot.ukefDealId || dealSnapshot.details.ukefDealId;
-
+  const ukefDealId = dealSnapshot.ukefDealId || dealSnapshot?.details?.ukefDealId;
   const firstTask = getFirstTask(tasks);
-  const urlOrigin = process.env.TFM_UI_URL;
   const templateId = EMAIL_TEMPLATE_IDS.TASK_READY_TO_START;
 
   try {
     const { team } = firstTask;
     const { email: sendToEmailAddress } = await api.findOneTeam(team.id);
 
-    const emailVariables = generateTaskEmailVariables(urlOrigin, firstTask, dealId, exporter.companyName, ukefDealId);
+    if (!dealId || !ukefDealId || !firstTask || !companyName) {
+      throw new Error('Invalid imperative arguments %s %s %o %s', dealId, ukefDealId, firstTask, companyName);
+    }
+
+    const emailVariables = generateTaskEmailVariables(TFM_UI_URL, firstTask, dealId, companyName, ukefDealId);
 
     const emailResponse = await sendTfmEmail(templateId, sendToEmailAddress, emailVariables);
 
--- a/trade-finance-manager-api/src/v1/helpers/amendment.helpers.test.js
+++ b/trade-finance-manager-api/src/v1/helpers/amendment.helpers.test.js
@@ -16,9 +16,7 @@ const {
 const CONSTANTS = require('../../constants');
 const { AMENDMENT_UW_DECISION, AMENDMENT_BANK_DECISION } = require('../../constants/deals');
 const { formattedNumber } = require('../../utils/number');
-
 const amendmentVariables = require('../__mocks__/amendmentVariables');
-
 const MOCK_NOTIFY_EMAIL_RESPONSE = require('../__mocks__/mock-notify-email-response');
 const MOCK_NOTIFY_EMAIL_BAD_RESPONSE = require('../__mocks__/mock-notify-email-bad-response');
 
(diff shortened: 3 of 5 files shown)


trade-finance-manager-api/src/v1/helpers/amendment.helpers.js
```diff
@@ -310,24 +312,50 @@ const firstTaskEmailConfirmation = async (facilityId, amendmentId, auditDetails)
...
   const firstTask = getFirstTask(tasks);
-  const urlOrigin = process.env.TFM_UI_URL;
   const templateId = EMAIL_TEMPLATE_IDS.TASK_READY_TO_START;
 
   try {
     const { team } = firstTask;
     const { email: sendToEmailAddress } = await api.findOneTeam(team.id);
 
-    const emailVariables = generateTaskEmailVariables(urlOrigin, firstTask, dealId, exporter.companyName, ukefDealId);
+    if (!dealId || !ukefDealId || !firstTask || !companyName) {
```
Zainzzkk, line 354: should this not go on line 351? if firstTask does not exist, it is being dereferenced before the check?
Also do we need to check if the object is populated rather than does the object exist?

trade-finance-manager-api/src/v1/helpers/amendment.helpers.test.js
```diff
@@ -575,6 +573,32 @@ describe('sendFirstTaskEmail()', () => {
...
+      'Invalid imperative arguments %s %s %o %s',
+      undefined,
+      amendmentVariables.firstTaskVariables.dealSnapshot.ukefDealId,
+      {},
+      amendmentVariables.firstTaskVariables.dealSnapshot.exporter.companyName,
+    );
+    expect(console.error).toHaveBeenCalledWith('Error sending first amendment task email %o', error);
+  });
+
+  it('should throw an error if company name is missing', async () => {
```
Zainzzkk, line 590: test for missing firstTask?

trade-finance-manager-api/src/v1/helpers/amendment.helpers.js
```diff
@@ -310,24 +312,50 @@ const firstTaskEmailConfirmation = async (facilityId, amendmentId, auditDetails)
...
+ * @param {string} deal.facilityId - The ID of the facility.
+ * @param {string} deal.amendmentId - The ID of the amendment.
+ * @param {Object} deal.dealSnapshot.exporter - The exporter details.
+ * @param {string} deal.dealSnapshot.exporter.companyName - The name of the exporter company.
+ * @param {string} [deal.dealSnapshot.ukefDealId] - The UKEF deal ID for GEF deals.
+ * @param {Object} [deal.dealSnapshot.details] - The details object for BSS deals.
+ * @param {string} [deal.dealSnapshot.details.ukefDealId] - The UKEF deal ID for BSS deals.
+ * @param {Object} auditDetails - The audit details for logging purposes.
+ * @returns {Promise<void>} - A promise that resolves when the email has been sent and the flag has been updated.
+ * @throws {Error} - Logs the error if any error occurs during the email sending process.
```
AlexBramhill, line 336: I think this function never throws an error that isn't caught and then silently ignored (well, technically `getFirstTask` could throw and that would not be caught by the `catch` block)?

trade-finance-manager-api/src/v1/helpers/amendment.helpers.js
ChristianMcc: line 351 shouldn't deference firstTask, it just pulls out team from it into a separate variable. Improved object validation doesn't sound like a bad idea, haven't looked into the implementation of `getFirstTask` on if there is anything in there

trade-finance-manager-ui/server/controllers/case/index.js
```diff
@@ -20,52 +20,57 @@ const {
...
+      dealId,
+      user: req.session.user,
+      amendments,
+      amendmentsInProgress,
+      hasAmendmentInProgress,
+      showDealCancelButton: dealCancellationIsEnabled,
+      hasDraftCancellation,
+    });
+  } catch (error) {
+    console.error('Unable to render deal %o', error);
```
AlexBramhill, line 71: Just a question, not a comment -- If we have an uncaught exception, it'll hit our middleware and show the problem with service page. It'll also log the error, which will include the stack trace through this function etc ie
```
err:
  controllers/case/index.js at getCaseDeal at LoC Bla
// goes further into depth about the stack trace
```

It seems the only thing this try catch is doing is appending `unable to render deal` -- Does this add value to debugging that justifies the additional indentations considering the stack trace will contain this information anyway?

trade-finance-manager-api/src/v1/__mocks__/amendmentVariables.js
```diff
@@ -217,6 +218,20 @@ const noTaskVariables = {
   },
 };
 
+const noCompanyNameDeal = {
```
ttbarnes, line 221:
```suggestion
const dealWithNoCompanyName = {
```

trade-finance-manager-api/src/v1/__mocks__/amendmentVariables.js
```diff
@@ -217,6 +218,20 @@ const noTaskVariables = {
...
+  ...firstTaskVariables,
+  dealSnapshot: {
+    ukefDealId,
+    exporter: {},
+    bankInternalRefName: 'UKEF',
+  },
+};
+
+const noDealId = {
```
ttbarnes, line 230:
```suggestion
const dealWithNoId = {
```

trade-finance-manager-api/src/v1/helpers/amendment.helpers.test.js
```diff
@@ -507,8 +505,8 @@ describe('sendManualDecisionAmendmentEmail()', () => {
 
 describe('sendFirstTaskEmail()', () => {
   const sendEmailApiSpy = jest.fn(() => Promise.resolve(MOCK_NOTIFY_EMAIL_RESPONSE));
-
   const updateFacilityAmendmentSpy = jest.fn(() => Promise.resolve({}));
+  console.error = jest.fn();
```
ttbarnes, line 509: Do we really need to do this

trade-finance-manager-ui/server/controllers/case/index.js
abhi-markan: Good question, due to lack of exceptional handling this will cause `tfm-ui` microservice to crash thus requiring an auto reboot.

With above not only we gracefully handle any such exception i.e. `dealId` being invalid as parameter being provided is `undefined` but also helps pinpoints the issue and aide in quick debugging.

trade-finance-manager-api/src/v1/helpers/amendment.helpers.test.js
abhi-markan: Yes to assert `console.error` is being executed with the correct message.

ttbarnes: I see - I think it's overkill to test console logs but happy to leave in if you'd prefer.

trade-finance-manager-api/src/v1/helpers/amendment.helpers.js
```diff
@@ -310,24 +312,49 @@ const firstTaskEmailConfirmation = async (facilityId, amendmentId, auditDetails)
...
+ * @param {Object} deal.amendment - The amendment details.
+ * @param {Object} deal.dealSnapshot - The snapshot of the deal details.
+ * @param {string} deal.facilityId - The ID of the facility.
+ * @param {string} deal.amendmentId - The ID of the amendment.
+ * @param {Object} deal.dealSnapshot.exporter - The exporter details.
+ * @param {string} deal.dealSnapshot.exporter.companyName - The name of the exporter company.
+ * @param {string} [deal.dealSnapshot.ukefDealId] - The UKEF deal ID for GEF deals.
+ * @param {Object} [deal.dealSnapshot.details] - The details object for BSS deals.
+ * @param {string} [deal.dealSnapshot.details.ukefDealId] - The UKEF deal ID for BSS deals.
+ * @param {Object} auditDetails - The audit details for logging purposes.
```
ttbarnes, line 334: It looks like the params could be simpler perhaps:

```ts
* @param {import('@ukef/dtfs2-common').Deal}
* @param {import("@ukef/dtfs2-common").AuditDetails}
```

trade-finance-manager-api/src/v1/__mocks__/amendmentVariables.js
```diff
@@ -217,6 +218,30 @@ const noTaskVariables = {
...
+
+const noUkefDealId = {
+  ...firstTaskVariables,
+  dealSnapshot: {
+    exporter: {
+      companyName: 'Tester',
+    },
+    bankInternalRefName: 'UKEF',
+  },
+};
```
AlexBramhill, line 243: This is somewhere that immer could be used to increase readability while keeping it all as one object --

```
const noUkefDealId = produce(firstTaskVariables, draft => {
    delete draft.dealSnapshot.ukefDealId;
});
```

But happy as is, as I think we should probably just look at replacing these with builders eventually and use this file as the director. But that's for another ticket/PR

AlexBramhill, line 243: This is somewhere that immer could be used to increase readability while keeping it all as one object --

```
const noUkefDealId = produce(firstTaskVariables, draft => {
    delete draft.dealSnapshot.ukefDealId;
});
```

But happy as is, as I think we should probably just look at replacing these with builders eventually and use this file as the director. But that's for another ticket/PR

trade-finance-manager-api/src/v1/__mocks__/amendmentVariables.js
```diff
@@ -217,6 +218,30 @@ const noTaskVariables = {
...
+
+const dealWithNoId = {
+  ...firstTaskVariables,
+  _id: null,
+};
+
+const noUkefDealId = {
```
AlexBramhill, line 235: `dealWithNoUkefDealId`?

trade-finance-manager-api/src/v1/helpers/amendment.helpers.js
AlexBramhill: This might kick up a stink with `Deal` as it may need bulking out to contain these items (but I'm not sure how much JSDocs cares about these things in reality -- so worth trying!

trade-finance-manager-api/src/v1/helpers/amendment.helpers.test.js
```diff
@@ -507,12 +505,13 @@ describe('sendManualDecisionAmendmentEmail()', () => {
...
 
   beforeEach(() => {
     sendEmailApiSpy.mockClear();
     updateFacilityAmendmentSpy.mockClear();
+    jest.clearAllMocks();
```
AlexBramhill, line 514: It looks to me as if `sendEmailApiSpy` and `updateFacilityAmendmentSpy` are not actually spies. As a result, I think that `jest.clearAllMocks();` should clear everything for us?

trade-finance-manager-ui/server/controllers/case/index.test.js
```diff
@@ -180,6 +180,26 @@ describe('controllers - case', () => {
         expect(res.redirect).toHaveBeenCalledWith('/not-found');
       });
     });
+
+    describe('should catch any thrown error', () => {
```
AlexBramhill, line 184: This should probably be 

```suggestion
    describe('when an error is thrown', () => {
```

to be in keeping with the naming of describe blocks in this file?

AlexBramhill (commented):
looking good, few minor things

ChristianMcc (approved):
Approving this to get this merged but think there are a couple of outstanding comments that we're going to need to address in retrospect after this is merged please!

Zainzzkk (approved):

trade-finance-manager-api/src/v1/helpers/amendment.helpers.js
ttbarnes: Need to get this PR asap so can look into this after merge.

ttbarnes (approved):

abhi-markan:
Thanks all will raise a new PR straight away, this is in order to get staging deployment initiated.

trade-finance-manager-api/src/v1/helpers/amendment.helpers.test.js
abhi-markan: Apologies, this should have been removed.
I have removed this in a new PR - https://github.com/UK-Export-Finance/dtfs2/pull/3899/files

trade-finance-manager-api/src/v1/helpers/amendment.helpers.js
abhi-markan: This a good shout @ttbarnes unfortunately not all of these properties exist so might be prudent to leave as it, however I have updated for `AuditDetails`.

abhi-markan: https://github.com/UK-Export-Finance/dtfs2/pull/3899/files

PR #3899, opened 2024-11-06
fix(DTFS2-7604): residual comments for PR #3897

abhi-markan:
## Introduction :pencil2:
Few comments were not addressed in the PR #3897 

* Added fallback assignment using logical OR operator for `_id` (Mongo deal ID) property access, pertinent to diverse deal types.

## Miscellaneous ➕ 
* Removed `jest.clearAllMocks()`;
* Updated JS Docs.
* Reworded Jest `describe` label.
* Updated no Mongo deal ID mock.

### DIFF
--- a/trade-finance-manager-api/src/v1/__mocks__/amendmentVariables.js
+++ b/trade-finance-manager-api/src/v1/__mocks__/amendmentVariables.js
@@ -11,6 +11,7 @@ const conditions = 'ABCD';
 const declined = 'EFGH';
 
 const approvedWithoutConditionsBothAmendments = {
+  _id: 'test',
   user: {
     firstname: 'Bob',
     surname: 'Smith',
@@ -217,6 +218,20 @@ const noTaskVariables = {
   },
 };
 
+const noCompanyNameDeal = {
+  ...firstTaskVariables,
+  dealSnapshot: {
+    ukefDealId,
+    exporter: {},
+    bankInternalRefName: 'UKEF',
+  },
+};
+
+const noDealId = {
+  ...firstTaskVariables,
+  _id: null,
+};
+
 module.exports = {
   approvedWithoutConditionsBothAmendments,
   approvedWithoutConditionsBothAmendmentsBSS,
@@ -234,4 +249,6 @@ module.exports = {
   wrongAmendments,
   firstTaskVariables,
   noTaskVariables,
+  noDealId,
+  noCompanyNameDeal,
 };
--- a/trade-finance-manager-api/src/v1/__mocks__/amendmentVariables.js
+++ b/trade-finance-manager-api/src/v1/__mocks__/amendmentVariables.js
@@ -232,6 +232,16 @@ const noDealId = {
   _id: null,
 };
 
+const noUkefDealId = {
+  ...firstTaskVariables,
+  dealSnapshot: {
+    exporter: {
+      companyName: 'Tester',
+    },
+    bankInternalRefName: 'UKEF',
+  },
+};
+
 module.exports = {
   approvedWithoutConditionsBothAmendments,
   approvedWithoutConditionsBothAmendmentsBSS,
@@ -251,4 +261,5 @@ module.exports = {
   noTaskVariables,
   noDealId,
   noCompanyNameDeal,
+  noUkefDealId,
 };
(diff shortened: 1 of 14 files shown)


ttbarnes (dismissed):
Great work!

ttbarnes (dismissed):

trade-finance-manager-api/src/v1/helpers/amendment.helpers.js
```diff
@@ -331,19 +331,21 @@ const firstTaskEmailConfirmation = async (facilityId, amendmentId, auditDetails)
...
- * @param {Object} auditDetails - The audit details for logging purposes.
+ * @param {import("@ukef/dtfs2-common").AuditDetails}
  * @returns {Promise<void>} - A promise that resolves when the email has been sent and the flag has been updated.
  */
 const sendFirstTaskEmail = async (deal, auditDetails) => {
   try {
     const { amendment, dealSnapshot, facilityId, amendmentId } = deal;
     const { tasks } = amendment;
-    const { _id: dealId } = deal;
+    const { _id } = deal;
```
ChristianMcc, line 341: you could keep this re-assignment to dealId to make it clear, maybe even dealMongoId, but ye good catch the assignment was wrong way round

`const { dealMongoId: _id } = deal;`

ChristianMcc: probably not necessary just thought that's probably where the error is more than anything else! we can keep the logic on line 346 of using the dealSnapshot and just getting whichever one we can if we need

trade-finance-manager-api/src/v1/helpers/amendment.helpers.js
```diff
@@ -331,19 +331,21 @@ const firstTaskEmailConfirmation = async (facilityId, amendmentId, auditDetails)
...
     const { companyName } = exporter;
 
-    // dealId in snapshot for gef and details for bss
-    const ukefDealId = dealSnapshot.ukefDealId || dealSnapshot?.details?.ukefDealId;
+    // BSS/EWCS or GEF Mongo deal ID
+    const dealId = dealSnapshot?._id || _id;
+    // BSS/EWCS or GEF UKEF deal ID
+    const ukefDealId = dealSnapshot?.details?.ukefDealId || dealSnapshot.ukefDealId;
     const firstTask = getFirstTask(tasks);
     const templateId = EMAIL_TEMPLATE_IDS.TASK_READY_TO_START;
```
ChristianMcc, line 351: it's not part of this PR but just fwiw something like:

`Missing required parameters to send email for ${dealId}` or `Missing required arguments provided for ${dealId}` would be a slightly clearer error message on line 353

ChristianMcc (approved):
LGTM, this code is absolutely fine so approving but left a couple of comments if you want to change anything before merging or just in retrospect

trade-finance-manager-ui/server/controllers/case/index.test.js
```diff
@@ -181,7 +181,7 @@ describe('controllers - case', () => {
       });
     });
 
-    describe('should catch any thrown error', () => {
+    describe('when an error is thrown', () => {
```
AlexBramhill, line 184: Thank you!

AlexBramhill (dismissed):
lgtm, thanks!

trade-finance-manager-api/src/v1/helpers/amendment.helpers.js
abhi-markan: Good shout I have reworded it.

trade-finance-manager-api/src/v1/helpers/amendment.helpers.js
abhi-markan: Good shout, have moved the logic to LOC:346 should have thought about it.

trade-finance-manager-api/src/v1/__mocks__/amendmentVariables.js
```diff
@@ -226,10 +226,17 @@ const dealWithNoCompanyName = {
     bankInternalRefName: 'UKEF',
   },
 };
-
-const dealWithNoId = {
+const dealWithNoMongoId = {
```
abhi-markan, line 229: To be explicit on which Id as `dealWithNoId` can be confused for either MongoId or UKEF Deal ID.

ChristianMcc (dismissed):
LGTM! :)

PR #3908, opened 2024-11-07
feat(FN-3573): add create record correction request page

sw-mattw:
## Introduction :pencil2:
This PR implements the first stage of the new record correction journey. A button (out of scope) will perform a GET request to this page, passing through a single fee record ID which the user wishes to create a record correction request for.

Note that:
- E2E tests will be added by a later ticket for this journey (as the form is split across a number of pages which will be implemented over multiple tickets)
- The fee record is hardcoded at present (a follow up PR under this ticket will add a new endpoint and call this from the controller)

- Added a template and controller for the initial step in this journey
- Added fee record summary macro template to show a summary of the fee record details (this is to be shared with other pages on the journey)

- Added a new middleware function `validateTfmFeeRecordCorrectionFeatureFlagIsEnabled` to verify the new record correction feature flag is enabled

Screenshot of the new page (with hardcoded facility ID data):

UKEF_FN-3573_PR-1_v3


trade-finance-manager-ui/public/js/govukFrontend.js.map
sw-mattw, line 1: These were recompiled when `npm run build` was run - no changes were made to the source files

trade-finance-manager-ui/server/controllers/utilisation-reports/record-corrections/create-record-correction-request/index.test.ts
```diff
@@ -0,0 +1,27 @@
...
+    user: aTfmSessionUser(),
+  };
+
+  it('should render create record correction request page', () => {
+    // Arrange
+    const { req, res } = httpMocks.createMocks<CreateRecordCorrectionRequestRequest>({
+      session: requestSession,
+      params: { reportId: '123', feeRecordId: '456' },
+    });
+
```
sw-mattw, line 20: There's no mocking at present as the data is hardcoded - this will be changed in a subsequent PR, and this test file will be updated

trade-finance-manager-ui/server/controllers/utilisation-reports/record-corrections/create-record-correction-request/index.test.ts
```diff
@@ -0,0 +1,27 @@
...
+    // Act
+    createRecordCorrectionRequest(req, res);
+
+    // Assert
+    expect(res._getRenderView()).toEqual('utilisation-reports/record-corrections/create-record-correction-request.njk');
```
ttbarnes, line 25: Can we add an assertion for the params/variables please e.g

```ts
expect(res._getRenderData() as BankRequestDateViewModel).toEqual({ ... });
```

trade-finance-manager-ui/styles/block-header.scss
```diff
@@ -0,0 +1,11 @@
+.block-header {
```
ttbarnes, line 1: How does the case summary have a blue background etc, can we re-use anything from there?

trade-finance-manager-ui/styles/block-header.scss
```diff
@@ -0,0 +1,11 @@
...
+  background-color: $govuk-brand-colour;
+  padding: 10px;
+  color: govuk-colour("white");
+  margin-bottom: 40px;
+
+  p {
+    margin-top: 10px;
+    margin-bottom: 10px;
+  }
```
ttbarnes, line 10: Can we use the GOV class names in nunjucks, `govuk-!-margin-bottom-X` etc instead of in the styles?

trade-finance-manager-ui/templates/utilisation-reports/record-corrections/_macros/fee-record-summary.njk
```diff
@@ -0,0 +1,40 @@
...
+        </div>
+      </div>
+    </div>
+
+    <div class="govuk-grid-row">
+      <div class="govuk-grid-column-one-half">
+        <div class="govuk-grid-column-one-third" data-cy="requested-by-label">
+          <p><strong>Requested by</strong></p>
+        </div>
+        <div class="govuk-grid-column-two-thirds" data-cy="requested-by">
```
ttbarnes, line 34: Can we move the `data-cy`'s into the actual text please

ttbarnes (changes requested):
Nice work, few comments

trade-finance-manager-ui/styles/block-header.scss
sw-mattw: The base styling looks fairly similar - I've commonised this selector (`block-header`) and added a new selector `fee-record-summary` which extends this, and also updated `case-summary` to also extend this commonised header

trade-finance-manager-ui/styles/block-header.scss
sw-mattw: Note: this has now moved to `fee-record-summary.scss` as a result of the changes for the above comment

This is to achieve spacing between the rows in the blue banner, so in retrospect it would be better if we apply the `govuk-body` class to the paragraph elements and apply the margin to the rows themselves.

We could achieve the row margin in nunjucks using `govuk-!-margin-` or I've just found that we can do the following in the scss file:

```
.govuk-grid-row {
  margin-top: govuk-spacing(3);
}

.govuk-grid-row:last-child {
  margin-bottom: govuk-spacing(3);
}
```

Do you have a preference?

ttbarnes: Thanks @sw-mattw I think we should do it in nunjucks beacuse the majority of the codebase does this vs `govuk-spacing` in scsss, can be consistent

trade-finance-manager-ui/server/middleware/validateTfmFeeRecordCorrectionFeatureFlagIsEnabled/index.test.ts
```diff
@@ -0,0 +1,43 @@
+import httpMocks from 'node-mocks-http';
```
ttbarnes, line 1: Not for this PR but this file name is very long 😄 `validateTfmFeeRecordCorrectionFeatureFlagIsEnabled`

I think a structure like this would be better - but we should look into this in another PR

```
middleware/validate-feature-flag-enabled/record-correction
middleware/validate-feature-flag-enabled/reconcilliation
middleware/validate-feature-flag-disabled/reconcilliation
```

trade-finance-manager-ui/styles/block-header.scss
sw-mattw: Sounds good thanks, updated to use `govuk-!-margin-` in the template

trade-finance-manager-ui/styles/fee-record-summary.scss
```diff
@@ -0,0 +1,16 @@
+.fee-record-summary {
+  @extend .block-header;
+  padding: govuk-spacing(4);
+  margin-bottom: govuk-spacing(8);
+
+  .govuk-body,
+  .govuk-link,
+  .govuk-back-link {
```
ttbarnes, line 8: Can only see `govuk-body` in `fee-record-summary`, in nunjucks, seems we only need govuk-body

trade-finance-manager-ui/server/middleware/validateTfmFeeRecordCorrectionFeatureFlagIsEnabled/index.test.ts
sw-mattw: I've raised a ticket on Jira for this to be addressed thanks - FN-3617 (`Tech Debt - Reorganise feature flag middleware files`)

ttbarnes: Amazing thanks 😄

ttbarnes (dismissed):

trade-finance-manager-ui/templates/utilisation-reports/record-corrections/create-record-correction-request.njk
```diff
@@ -0,0 +1,107 @@
...
+              attributes: {"data-cy": "reason-utilisation-incorrect"}
+            },
+            {
+              value: "other",
+              text: "Other",
+              hint: { text: "Something else" },
+              attributes: {"data-cy": "reason-other"}
+            }
+          ]
+        }) }}
```
ttbarnes, line 72: Can we indent nested properties please e.g instead of this:

```
hint: { text: "The fee is higher or lower than expected" },
attributes: { "data-cy": "reason-reported-fee-incorrect" }
```

Can be aligned =>

```
hint: {
  text: "The fee is higher or lower than expected"
}

attributes: {
  "data-cy": "reason-reported-fee-incorrect"
}
```

trade-finance-manager-ui/server/controllers/utilisation-reports/record-corrections/create-record-correction-request/index.ts
```diff
@@ -0,0 +1,50 @@
...
+      },
+      reportPeriod: {
+        start: { month: 1, year: 2024 },
+        end: { month: 1, year: 2024 },
+      },
+      facilityId: '0012345678',
+      exporter: 'Sample Company Ltd',
+    };
+
+    return renderCreateRecordCorrectionRequestPage(res, {
```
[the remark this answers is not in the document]
RLavender98: This is the pattern we have for all pages using viewModels, it's nice because it is consistent and also because `res.render` doesn't force you to pass in an object of any particular type, by adding this function and then calling it we are adding in some nice type safety on the object of variables we pass to the template.

trade-finance-manager-ui/templates/utilisation-reports/record-corrections/create-record-correction-request.njk
RLavender98: Done. I thought it looked nicer personally before so just wondering if there is a ticket to look into adding linting/code styling akin to prettier to nunjucks files so that we don't have to rely on communicating the desired styling across the team who will each have different opinions and also to save time and dev effort and improve consistency by having it happen automatically?

trade-finance-manager-ui/server/controllers/utilisation-reports/record-corrections/create-record-correction-request/index.ts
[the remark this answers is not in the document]
ttbarnes: Thanks @RLavender98 nice pattern - happy to keep for reports, would be great to apply in other areas as there are only 8x instances of `Page = (res`

ttbarnes (approved):

PR #3912, opened 2024-11-08
fix(FN-3616): fix end of row payment currency bug

RLavender98:
## Introduction :pencil2:
There is a bug where if the report is uploaded with the payment currency as the final column, then any rows where the payment currency is blank have the payment currency saved as '' instead of the fees paid to UKEF currency.

- Pad rows from xlsx file with less values than number of headers with null values instead of empty strings to match what happens if there is an empty not at the end of a row
- Update mapping from csv row to fee record entity to allow for empty strings to match validation and correct typing
- Add e2e-test that includes uploading in portal and then checking values have been persisted correctly in tfm
	- The report includes a row with empty payment currency after final value of cell
	- Include both xlsx and csv reports to check these are parsed identically


e2e-tests/tfm/cypress/e2e/pages/utilisation-reports/utilisationReportPage.js
```diff
@@ -28,7 +28,16 @@ const utilisationReportPage = {
...
+      rows: () => cy.get(`tr[data-cy^="premium-payments-table-row--feeRecordId-"]`),
+      facilityIdByRowIndex: (rowIndex) =>
+        cy.get('tr[data-cy^="premium-payments-table-row--feeRecordId-"]').eq(rowIndex).find('[data-cy="facility-id"]').first(),
+      exporterIdByRowIndex: (rowIndex) => cy.get('tr[data-cy^="premium-payments-table-row--feeRecordId-"]').eq(rowIndex).find('[data-cy="exporter"]').first(),
+      reportedFeesByRowIndex: (rowIndex) =>
+        cy.get('tr[data-cy^="premium-payments-table-row--feeRecordId-"]').eq(rowIndex).find('[data-cy="reported-fees"]').first(),
+      reportedPaymentsByRowIndex: (rowIndex) =>
+        cy.get('tr[data-cy^="premium-payments-table-row--feeRecordId-"]').eq(rowIndex).find('[data-cy="reported-payments"]').first(),
+      totalReportedPaymentsByRowIndex: (rowIndex) =>
+        cy.get('tr[data-cy^="premium-payments-table-row--feeRecordId-"]').eq(rowIndex).find('[data-cy="total-reported-payments"]').first(),
```
ttbarnes, line 40: Can we have a `rowIndex` object instead?

```js
rowIndex: {
  facilityId: () => '...',
  exporterId: () => '...',
  // etc
}
```

e2e-tests/ukef/cypress/e2e/journeys/utilisation-reports/utilisation-report-submission.spec.js
```diff
@@ -0,0 +1,235 @@
...
+    );
+  };
+
+  /**
+   * Asserts the values in a given row of the utilisation table match the expected
+   * values from the report as defined in february2023ExpectedValues
+   * @param {number} utilisationTableRowIndex - The index of the row of utilisation table
+   * @param {object} expectedParsedValuesForReportRow - The expected values to be found in the row
+   */
+  const assertUtilisationTableRowContainsExpectedValues = (utilisationTableRowIndex, expectedParsedValuesForReportRow) => {
```
ttbarnes, line 110: Can we rename `expectedParsedValuesForReportRow` to something simpler like `expectedValues`?

e2e-tests/ukef/cypress/e2e/journeys/utilisation-reports/utilisation-report-submission.spec.js
```diff
@@ -0,0 +1,235 @@
...
+    cy.assertText(
+      tfmPages.utilisationReportPage.premiumPaymentsTab.premiumPaymentsTable.reportedFeesByRowIndex(premiumPaymentsRowIndex),
+      expectedParsedValuesForReportRow.reportedFees,
+    );
+    cy.assertText(
+      tfmPages.utilisationReportPage.premiumPaymentsTab.premiumPaymentsTable.reportedPaymentsByRowIndex(premiumPaymentsRowIndex),
+      expectedParsedValuesForReportRow.reportedPayments,
+    );
+    cy.assertText(
+      tfmPages.utilisationReportPage.premiumPaymentsTab.premiumPaymentsTable.totalReportedPaymentsByRowIndex(premiumPaymentsRowIndex),
```
ttbarnes, line 99: Can we destructure the selectors (whilst keeping some level of context) so that it reads nicer? e.g

```js
cy.assertText(rowIndex.reportedFees(rowIndex), expectedValues.reportedFees);
```

e2e-tests/ukef/cypress/e2e/journeys/utilisation-reports/utilisation-report-submission.spec.js
```diff
@@ -0,0 +1,235 @@
...
+    portalPages.confirmAndSend.confirmAndSendButton().click();
+
+    //---------------------------------------------------------------------------------
+    // PDC_RECONCILE user can login to TFM and view the data from the submitted report
+    //---------------------------------------------------------------------------------
+    cy.clearCookie('dtfs-session');
+    cy.clearCookie('_csrf');
+    cy.getCookies().should('be.empty');
+
+    cy.forceVisit(TFM_URL);
```
ttbarnes, line 140: Why do we need `forceVisit`?

e2e-tests/ukef/cypress/e2e/journeys/utilisation-reports/utilisation-report-submission.spec.js
```diff
@@ -0,0 +1,235 @@
...
+
+    /**
+     * Each line of the report will have an entry in the premium payments table,
+     * but they will not be in the same order so first we find the index within
+     * the premium payments tab table for each of the report rows.
+     */
+    tfmPages.utilisationReportPage.premiumPaymentsTab.premiumPaymentsTable.rows().each(($row, index) => {
+      cy.wrap($row)
+        .find('[data-cy="exporter"]')
+        .eq(0)
```
ttbarnes, line 155: Are we able to use selectors and `rowIndex` RE a previous comment?

e2e-tests/ukef/cypress/support/portal/logIn.js
```diff
@@ -15,5 +16,9 @@ module.exports = (opts) => {
     signInLink.visit({ token: signInToken, userId: _id });
   });
 
-  cy.url().should('eq', relative('/dashboard/deals/0'));
+  if (roles?.length === 1 && roles?.at(0) === ROLES.PAYMENT_REPORT_OFFICER) {
```
ttbarnes, line 19: Is `roles` an array of strings? If so I think this would be better, just incase the ordering changes:

```js
if (roles.includes(ROLES.PAYMENT_REPORT_OFFICER)) { ... }
```

What do you think?

ttbarnes (changes requested):

.github/workflows/test.yml
```diff
@@ -723,26 +723,27 @@ jobs:
...
+            'portal/submit-MIN-changes-tfm-stage-to-confirmed/**/*.spec.js',
+            'portal/submit-MIN-deal-does-not-create-credit-rating-in-tfm/**/*.spec.js',
+            'portal/submit-MIN-deal-updates-portal-deal-status/**/*.spec.js',
+            'portal/submit-MIN-with-unissued-facilities-and-resubmit-issued-facilities-updates-facilities-in-tfm/**/*.spec.js',
+            'portal/submit-deal-adds-risk-profile-to-facilities/**/*.spec.js',
+            'portal/submit-deal-converts-contract-value-to-gbp/**/*.spec.js',
+            'portal/submit-deal-creates-loss-given-and-probability-defaults-in-tfm/**/*.spec.js',
+            'portal/submit-deal-with-issued-facilities/**/*.spec.js',
+            'portal/submit-deal-with-non-gbp-facilities/**/*.spec.js',
+            'portal/submit-deal-with-unissued-facilities-and-resubmit-as-issued/**/*.spec.js',
```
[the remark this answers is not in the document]
RLavender98: These are the 'ukef' e2e-tests, they were already in a 'portal' folder, it was referenced lower down, but I added a new folder 'utilisation-reports' inside the 'ukef' tests, so to run those, I had to move the base path one directory up and then change these.

dtfs-central-api/src/helpers/fee-record-csv-row-mapper.ts
```diff
@@ -13,7 +20,7 @@ type FeeRecordCsvRowToSqlEntityParams = {
  * @param defaultNumber - The number to default the value to
  * @returns Either the `maybeString` as a number or the default number
  */
-const asNumberOrDefault = (maybeString: string | undefined, defaultNumber: number) => (maybeString ? Number(maybeString) : defaultNumber);
+const asNumberOrDefault = (maybeString: string | undefined | null, defaultNumber: number) => (maybeString ? Number(maybeString) : defaultNumber);
```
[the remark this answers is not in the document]
RLavender98: Yes they are

e2e-tests/tfm/cypress/e2e/pages/utilisation-reports/utilisationReportPage.js
RLavender98: We have to pass the row index in as a parameter and I dislike:`rowIndex.facilityId(rowIndex)` since the parameter is removed from the statement that claims to care about it. 

I guess I could try and define it as:
```
rowIndex: (index) => ({
  facilityId: () => '...some selector with index in it...',
  exporterId: () => '...some selector with index in it...',
  // etc
}),
```
So then would be called like `rowIndex().facilityId()` but this feels like a different pattern to everything I have seen until now so keen to know what you think before I try it?

e2e-tests/ukef/cypress/e2e/journeys/utilisation-reports/utilisation-report-submission.spec.js
RLavender98: I didn't destructure because we have pages for both portal and tfm in this file and I felt it was quite useful to see at a glance which route we are currently in throughout the test. I think this is the pattern in the other 'ukef' e2e-tests

e2e-tests/ukef/cypress/e2e/journeys/utilisation-reports/utilisation-report-submission.spec.js
RLavender98: I'm not sure, I copied it from the other tests in the 'ukef' set of tests since they all do it so I just assumed it was necessary. I'll have a see what happens without it.

e2e-tests/ukef/cypress/e2e/journeys/utilisation-reports/utilisation-report-submission.spec.js
RLavender98: Sorry, I'm not sure what you mean? I didn't put this in pages since it is quite complicated and I'm not sure I see this being relevant to many other tests since in most of our tests we want to get the row by the data-cy which uses the feeRecordId since in most tests that is the most usual way to get it, otherwise we would have to do essentially this in all those tests and that would be in a lot more tests. Would be nice if we could have two different data-cy on one element so we can get them by different properties in different tests but I haven't seen a way to do that.

RLavender98: The problem is most of our e2e-tests get the rows by feeRecordId and that is generally much more useful than by exporter since exporter is not guaranteed to be unique, so feerecordId is used in the data-cy. Here I want to get by exporter because I have picked exporter name that are unique to allow for this testing.

It would be nice if we could have multiple data-cy per element but as far as I can see that is not possible and I think this is the simplest I can manage this.

e2e-tests/ukef/cypress/support/portal/logIn.js
ChristianMcc: FWIW I think the issue here is we want to test the redirect when your role is ONLY payment_report_officer, not when it's one of your multiple roles and one of your roles in `payment_report_officer` which means we need the extra checks

portal/server/utils/csv-utils.js
```diff
@@ -111,15 +111,15 @@ const parseXlsxToCsvArrays = (worksheet) => {
 
     if (!firstRow) {
       // If the row has no data in the final columns of the row, we need to fill in the empty cells with
-      // empty strings and cell addresses so that the csv parser library can parse the data correctly
+      // null values and cell addresses so that the csv parser library can parse the data correctly
       while (columnCount < headerCount) {
-        rowData.push('');
+        rowData.push(`${null}`);
```
ChristianMcc, line 116: so when this gets parsed, is it once it gets to JSON from the `csv-parser` that this then gets parsed as 'null' in the json data?

e2e-tests/ukef/cypress/e2e/journeys/utilisation-reports/utilisation-report-submission.spec.js
```diff
@@ -0,0 +1,235 @@
...
+import portalPages from '../../../../../portal/cypress/e2e/pages';
+import tfmPages from '../../../../../tfm/cypress/e2e/pages';
+
+import MOCK_USERS from '../../../../../e2e-fixtures/portal-users.fixture';
+import { tfmFacilityForReport } from '../../../fixtures/tfm-facility';
+import { february2023ExpectedValues } from '../../../fixtures/february-2023-expected-values';
+
+const { BANK1_PAYMENT_REPORT_OFFICER1 } = MOCK_USERS;
+
+context('Portal to TFM utilisation report submission', () => {
```
ChristianMcc, line 20: Nice test, I'll add a ticket to do an end to end one up to keying sheets where we create the csv file to test it :)

ChristianMcc (commented):
LGTM, thanks for fixing the bug and the long test. Pretty tricky selectors on some of these pages with so much data but I think it's partly unavoidable

e2e-tests/tfm/cypress/e2e/pages/utilisation-reports/utilisationReportPage.js
ttbarnes: I think the current selectors are too long and not very readable, e.g `totalReportedPaymentsByRowIndex` or `xyzByRowIndex`

There's only 1x existing selector that has `xyzByRowIndex` (facilityIdByRowIndex)

I think a `rowIndex` function like you mentioned is a lot cleaner. What do you think? e.g

```js
rowIndex: (index) => ({
  facilityId: () => cy.get(`[data-cy="row-${index}-facility-id"]`),
  exporterId: () => cy.get(`[data-cy="row-${index}-exporter-id"]`),
})
```

e2e-tests/ukef/cypress/e2e/journeys/utilisation-reports/utilisation-report-submission.spec.js
ttbarnes: bump @RLavender98

e2e-tests/ukef/cypress/e2e/journeys/utilisation-reports/utilisation-report-submission.spec.js
ttbarnes: I think it's possible to have multiple data-cy values separated by a space and a different cypress selector - trying this out in another PR like this:

```
"data-cy": "deal__exporter--index--1 deal__exporter--abc12345678"
```

So we can get a row by index, or deal ID.  E.g by ID, the "*" after `data-cy` is for if `data-cy` contains the value

```js
exporter: (id) => cy.get(`[data-cy*="deal__exporter--${id}"]`)
```

e2e-tests/ukef/cypress/support/portal/logIn.js
ttbarnes: Can we add a comment for this please, it's not obvious otherwise

My concern was that if the `roles` param changes in some relevant test, to have a different first item or multiple roles, this'd fail

portal/server/utils/csv-utils.js
RLavender98: Ahh looked into where this is used more and I hadn't tested the csv download and it gives me spreadsheet with lots of strings of the word null in it which we don't want. 

It's the with cell addresses where using null in the string matches the empty cells not at the end of the row, but given I have now added handling to the mapping for empty strings since we need it for csv uploads anyway I will simply revert this part back.

e2e-tests/ukef/cypress/e2e/journeys/utilisation-reports/utilisation-report-submission.spec.js
RLavender98: Ooh interesting, I'll give this a go to see if it can help me simplify!

e2e-tests/tfm/cypress/e2e/pages/utilisation-reports/utilisationReportPage.js
RLavender98: I have switched to this pattern but combined with the double selector trick you shared to use the exporter directly, this is now greatly simplified! Thanks for the suggestions, especially the double data-cy, makes this so much nicer!

e2e-tests/ukef/cypress/e2e/journeys/utilisation-reports/utilisation-report-submission.spec.js
RLavender98: This is now simplified greatly, as it removed the need to loop through the rows and store the indexes.

e2e-tests/tfm/cypress/e2e/pages/utilisation-reports/utilisationReportPage.js
RLavender98: Applied the same pattern to the utilisation tab tests for consistency whilst I was at it as well

e2e-tests/ukef/cypress/support/portal/logIn.js
RLavender98: Added a comment 👍

e2e-tests/tfm/cypress/e2e/pages/utilisation-reports/utilisationReportPage.js
```diff
@@ -75,19 +90,48 @@ const utilisationReportPage = {
...
     table: {
-      row: (feeRecordId) => cy.get(`tr[data-cy="utilisation-table-row-${feeRecordId}"]`),
+      // NOTE: Using the ^= selector to match all table rows starting with the given data-cy value (regardless of the actual fee record id value).
+      rows: () => cy.get(`tr[data-cy^="utilisation-table-row-"]`),
+      /**
+       * NOTE: this selector should only be used with test data with unique exporters.
+       * If fee record id is known then the preference is to use that selector as it has guaranteed uniqueness.
+       */
+      rowWithExporter: (exporter) => ({
+        facilityId: () => cy.get(`tr[data-cy*="utilisation-table-row--exporter-${exporter}"]`).find('[data-cy="facility-id"]').first(),
```
ttbarnes, line 100: Is `.first()` required? it looks like it might be needed 

Assuming the `exporter` is an ID, can we rename to `exporterId`?

e2e-tests/tfm/cypress/e2e/pages/utilisation-reports/utilisationReportPage.js
```diff
@@ -75,19 +90,48 @@ const utilisationReportPage = {
...
+        facilityId: () => cy.get(`tr[data-cy*="utilisation-table-row-${feeRecordId}"]`).find('[data-cy="facility-id"]').first(),
+        exporter: () => cy.get(`tr[data-cy*="utilisation-table-row-${feeRecordId}"]`).find('[data-cy="exporter"]').first(),
+        baseCurrency: () => cy.get(`tr[data-cy*="utilisation-table-row-${feeRecordId}"]`).find('[data-cy="base-currency"]').first(),
+        value: () => cy.get(`tr[data-cy*="utilisation-table-row-${feeRecordId}"]`).find('[data-cy="value"]').first(),
+        utilisation: () => cy.get(`tr[data-cy*="utilisation-table-row-${feeRecordId}"]`).find('[data-cy="utilisation"]').first(),
+        coverPercentage: () => cy.get(`tr[data-cy*="utilisation-table-row-${feeRecordId}"]`).find('[data-cy="cover-percentage"]').first(),
+        exposure: () => cy.get(`tr[data-cy*="utilisation-table-row-${feeRecordId}"]`).find('[data-cy="exposure"]').first(),
+        feesAccrued: () => cy.get(`tr[data-cy*="utilisation-table-row-${feeRecordId}"]`).find('[data-cy="fees-accrued"]').first(),
+        feesPayable: () => cy.get(`tr[data-cy*="utilisation-table-row-${feeRecordId}"]`).find('[data-cy="fees-payable"]').first(),
+      }),
```
ttbarnes, line 131: Not for this PR, we should split this file up later, lots going on

e2e-tests/ukef/cypress/e2e/journeys/utilisation-reports/utilisation-report-submission.spec.js
```diff
@@ -0,0 +1,235 @@
...
+    tfmPages.utilisationReportPage.utilisationTab.table.rowByIndex(utilisationTableRowIndex).within(() => {
+      cy.assertText(tfmPages.utilisationReportPage.utilisationTab.table.facilityId(), expectedParsedValuesForReportRow.facilityId);
+      cy.assertText(tfmPages.utilisationReportPage.utilisationTab.table.exporter(), expectedParsedValuesForReportRow.exporter);
+      cy.assertText(tfmPages.utilisationReportPage.utilisationTab.table.feesPayable(), expectedParsedValuesForReportRow.reportedFees);
+      cy.assertText(tfmPages.utilisationReportPage.utilisationTab.table.feesAccrued(), expectedParsedValuesForReportRow.feesAccrued);
+      cy.assertText(tfmPages.utilisationReportPage.utilisationTab.table.utilisation(), expectedParsedValuesForReportRow.utilisation);
+    });
+  };
+
+  const testFebruary2023ReportUploadWithFileType = (filetype) => {
```
ttbarnes, line 100:
```suggestion
  const testFebruary2023ReportUploadWithFileType = (fileType) => {
```

e2e-tests/tfm/cypress/e2e/pages/utilisation-reports/utilisationReportPage.js
ttbarnes: Should it be `exporterName`? Just the current param name looks like it would be an object

Do we need the `.first()`'s here?

RLavender98: It shouldn't be `exporterName` because we use `exporter` extensively in utilisation reports and it always is just a string not an object in this context so it would be inconsistent to do it differently here and a very large effort task to rename it generally to `exporterName`.

I thought the `first`s were necessary as thought it returned an array but seems it id not required if there is only one to find so have removed 👍

ttbarnes: Thanks @RLavender98 

It's not clear what `exporter` actually is, would like to rename throughout - if it's not an exporter name or ID, curious what it actually is 🤔 

But not a blocker for this PR, better to be consistent

ttbarnes (dismissed):

portal/server/utils/csv-utils.test.js
```diff
@@ -62,7 +62,7 @@ describe('csv-utils', () => {
       expect(parsedData).toEqual(expectedParsedData);
     });
 
-    it('Parses an excelJS workbook and adds in addresses for any missing cells in a row', async () => {
+    it('Parses an excelJS workbook and adds in null values with addresses for any missing cells in a row', async () => {
```
ChristianMcc, line 65: is null technically right now? It's not far off an empty space is basically null in a string :D

ChristianMcc (approved):
LGTM, nit pick on test name but I don't think it's wrong technically

ttbarnes (approved):

PR #3915, opened 2024-11-08
feat(FN-2870): send email if create utilisation report job fails

Zainzzkk:
## Introduction :pencil2:
This PR introduces code to send an email if the create utilisation cron job fails

* Added env
* Added try catch and call to external api email if catch triggered
* Added test
* Added env to docker-compose
* Added env to env.sample
* Added env to test.yml


dtfs-central-api/src/cron-scheduler-jobs/create-utilisation-reports/index.test.ts
```diff
@@ -28,29 +33,94 @@ describe('scheduler/jobs/create-utilisation-reports', () => {
...
+      isVisibleInTfmUtilisationReports: true,
+      utilisationReportPeriodSchedule: MONTHLY_REPORT_PERIOD_SCHEDULE,
+    },
+    {
+      id: '3',
+      name: 'Bank 3',
+      isVisibleInTfmUtilisationReports: true,
+      utilisationReportPeriodSchedule: MONTHLY_REPORT_PERIOD_SCHEDULE,
+    },
+  ] as Bank[];
```
ttbarnes, line 82: Some repetition here, can do this please


```ts
const banks = [
  {
    ...bank,
    id: '1',
    name: 'Bank 1',
  },
  {
    ...bank,
    id: '2',
    name: 'Bank 2',
  },
] as Bank[];
```

dtfs-central-api/src/cron-scheduler-jobs/create-utilisation-reports/index.test.ts
```diff
@@ -28,29 +33,94 @@ describe('scheduler/jobs/create-utilisation-reports', () => {
...
+    { startMonth: 6, endMonth: 6 },
+    { startMonth: 7, endMonth: 7 },
+    { startMonth: 8, endMonth: 8 },
+    { startMonth: 9, endMonth: 9 },
+    { startMonth: 10, endMonth: 10 },
+    { startMonth: 11, endMonth: 11 },
+    { startMonth: 12, endMonth: 12 },
+  ];
+
+  const bank = {
```
ttbarnes, line 56:
```suggestion
  const mockBank = {
```

dtfs-central-api/src/cron-scheduler-jobs/create-utilisation-reports/index.test.ts
```diff
@@ -137,9 +180,9 @@ describe('scheduler/jobs/create-utilisation-reports', () => {
       banks.forEach(({ id }) => expect(findOneByBankIdAndReportPeriodSpy).toHaveBeenCalledWith(id, mockReportPeriod));
       expect(saveUtilisationReportSpy).toHaveBeenCalledTimes(banks.length);
 
-      banks.forEach((bank) => {
+      banks.forEach((eachBank) => {
```
ttbarnes, line 183:
```suggestion
      banks.forEach((bank) => {
```

ttbarnes (changes requested):

ChristianMcc (changes requested):
@abhi-markan I was imagining this working as more of an alert in azure based of error logs similar to the ones we already have that look for errors. This feels like the kind of thing that we would want to keep within azure as error/alerting/monitoring. Is this something we can set up instead of a custom email?

abhi-markan:
Spoke to @ChristianMcc this actually needs to be an Azure alert, closing this PR.

.github/workflows/test.yml
```diff
@@ -140,6 +140,7 @@ env:
   AUDIT_API_USERNAME: ${{ secrets.AUDIT_API_USERNAME }}
   AUDIT_API_PASSWORD: ${{ secrets.AUDIT_API_PASSWORD }}
   UTILISATION_REPORT_CREATION_FOR_BANKS_SCHEDULE: ${{ vars.UTILISATION_REPORT_CREATION_FOR_BANKS_SCHEDULE }}
+  UTILISATION_REPORT_CREATION_FAILURE_EMAIL_ADDRESS: ${{ secrets. UTILISATION_REPORT_CREATION_FAILURE_EMAIL_ADDRESS }}
```
ChristianMcc, line 143: Does this need to be a secret or can it just be a variable?

dtfs-central-api/src/cron-scheduler-jobs/create-utilisation-reports/index.test.ts
```diff
@@ -28,29 +33,91 @@ describe('scheduler/jobs/create-utilisation-reports', () => {
...
+  const mockReportPeriod = {} as ReportPeriod;
+
+  describe('errors on save', () => {
+    beforeEach(() => {
+      jest.mocked(getAllBanks).mockResolvedValue(banks);
+      sendEmailSpy = jest.fn(() => Promise.resolve({}));
+      externalApi.sendEmail = sendEmailSpy;
+    });
+
+    it('should send an email when UtilisationReportRepo.save fails', async () => {
```
ChristianMcc, line 90: Would it be worth simplifying this to just mock the implementation of UtilisationReportRepo.save to reject a promise rather than mocking the findOneByBankIdAndReportPeriodSpy to have custom behaviour to cause it to fail (which is then intertwined with the logic inside the .save function)?
line 95-102 could then just be a single mock on the .save function to make it reject

ChristianMcc (changes requested):
Nice one, looking good. Minor comments on a test and variable and another comment to follow up on separate to this PR about the email address to use

.github/workflows/test.yml
Zainzzkk: CONTACT_US_EMAIL_ADDRESS is stored as a secret so thought should keep this the same

ChristianMcc: absolutely fair for this, I'd argue contact us email is pretty public (accessible by anyone using the system) and so shouldn't be a secret but happy to leave this as a secret because it's minor anyway - it's just slightly easier to manage in the future as a variable. Will resolve this comment :)

dtfs-central-api/src/cron-scheduler-jobs/create-utilisation-reports/index.test.ts
Zainzzkk: changed and had to move the test to the bottom

ChristianMcc: Thank you, looks cleaner! Re moving test to the bottom does that imply we aren't resetting our mocks properly and need to update them?

ChristianMcc (approved):
Nice work, LGTM :)

dtfs-central-api/src/cron-scheduler-jobs/create-utilisation-reports/index.test.ts
```diff
@@ -70,29 +107,23 @@ describe('scheduler/jobs/create-utilisation-reports', () => {
...
+
+      const existingReport = UtilisationReportEntityMockBuilder.forStatus(REPORT_NOT_RECEIVED).withBankId(mockBank.id).build();
       findOneByBankIdAndReportPeriodSpy.mockResolvedValue(existingReport);
 
       // Act
       await createUtilisationReportForBanksJob.task(new Date());
 
       // Assert
-      expect(findOneByBankIdAndReportPeriodSpy).toHaveBeenCalledWith(bank.id, mockReportPeriod);
+      expect(findOneByBankIdAndReportPeriodSpy).toHaveBeenCalledWith(mockBank.id, mockReportPeriod);
```
ttbarnes, line 119: Can we add a `toHaveBeenCalledTimes(1)` assertion please

dtfs-central-api/src/cron-scheduler-jobs/create-utilisation-reports/index.test.ts
```diff
@@ -183,4 +189,26 @@ describe('scheduler/jobs/create-utilisation-reports', () => {
       expect(saveUtilisationReportSpy).toHaveBeenCalledWith(newReportForBankWithoutReport);
     });
   });
+
+  describe('errors on save', () => {
```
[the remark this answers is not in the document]
Zainzzkk: done

dtfs-central-api/src/cron-scheduler-jobs/create-utilisation-reports/index.test.ts
Zainzzkk: done

ttbarnes (approved):
LGTM

PR #3917, opened 2024-11-11
chore(FN): deploy main to feature before SSO is deployed

ChristianMcc:
## Introduction :pencil2:
Deploy main to feature temporarily while dev is fixed and SSO isn't ready


ttbarnes (approved):
Great work

PR #3920, opened 2024-11-11
feat(DTFS2-6892): add controllers and error handler wrapper to start routes

AlexBramhill:
## Introduction :pencil2:
updates start routes in portal api to use error wrapper

extracts out start routes to controllers

### DIFF
--- a/portal/server/controllers/start/before-you-start-bank-deal.controller.js
+++ b/portal/server/controllers/start/before-you-start-bank-deal.controller.js
@@ -1,0 +1,38 @@
+import { constructPayload, errorHref, postToApi, requestParams } from '../../helpers';
+import { MANDATORY_CRITERIA } from '../../routes/api-data-provider';
+import api from '../../api';
+
+export const getBeforeYouStartBankDealPage = async (req, res) => {
+  res.render('before-you-start/before-you-start-bank-deal.njk', {
+    user: req.session.user,
+  });
+};
+
+export const postBeforeYouStartBankDealPage = async (req, res) => {
+  const { userToken } = requestParams(req);
+
+  const allowedFields = ['bankInternalRefName', 'additionalRefName'];
+  const sanitizedBody = constructPayload(req.body, allowedFields);
+
+  const newDeal = {
+    ...sanitizedBody,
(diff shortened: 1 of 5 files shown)


portal/server/controllers/start/before-you-start-bank-deal.controller.js
```diff
@@ -0,0 +1,38 @@
+import { constructPayload, errorHref, postToApi, requestParams } from '../../helpers';
```
AlexBramhill, line 1: I am not adding tests as this is only moving code around -- and theres a lot to do!

portal/server/controllers/start/before-you-start.controller.js
```diff
@@ -0,0 +1,29 @@
+import { errorHref, generateErrorSummary } from '../../helpers';
+import beforeYouStartValidation from '../../validation/before-you-start';
+
+export const getBeforeYouStartPage = async (req, res) => {
+  const { mandatoryCriteria } = req.apiData;
+  res.render('before-you-start/before-you-start.njk', {
```
abhi-markan, line 6: Can these be inside `template` directories to keep it consistent please.

portal/server/controllers/start/before-you-start.controller.js
```diff
@@ -0,0 +1,29 @@
+import { errorHref, generateErrorSummary } from '../../helpers';
+import beforeYouStartValidation from '../../validation/before-you-start';
+
+export const getBeforeYouStartPage = async (req, res) => {
```
abhi-markan, line 4: Docstring please.

abhi-markan (changes requested):

portal/server/controllers/start/before-you-start.controller.js
AlexBramhill: sorry -- These are just me pulling out controller functions into controllers -- If this is the method i'm going to go with, there'll be a huge effort in doing this, so I'm not looking to touch the actual content if thats okay?

AlexBramhill:
Moved to drafts due to reassessment of deliverables

PR #3923, opened 2024-11-12
feat(DTFS2-7612): update portal facility status to risk expired

MarRobSoftwire:
## Introduction :pencil2:
When a deal is cancelled in tfm, the portal facilities should show `risk expired`

- Modify the repo return type to send facilities
- Add new facilities service to update BSS and GEF facilities
- Update facilities in deal cancellation service

## Screenshots 📸 
BSS/EWCS Facilities
image

- Add `FacilityStatus` type
- Modify `TfmDealCancellationResponse` type back to a previous version & update controllers/tests to handle this

## Note ⚠️ 
This PR doesnt cover front end changes, gef-ui will need some refactoring to render the new status

### DIFF
--- a/dtfs-central-api/api-tests/v1/tfm/deal-cancellation/tfm-submit-deal-cancellation-post.api-test.ts
+++ b/dtfs-central-api/api-tests/v1/tfm/deal-cancellation/tfm-submit-deal-cancellation-post.api-test.ts
@@ -1,6 +1,14 @@
 import { ObjectId } from 'mongodb';
 import { HttpStatusCode } from 'axios';
-import { MONGO_DB_COLLECTIONS, AnyObject, TFM_DEAL_STAGE, API_ERROR_CODE, TfmAuditDetails, TfmDeal } from '@ukef/dtfs2-common';
+import {
+  MONGO_DB_COLLECTIONS,
+  AnyObject,
+  TFM_DEAL_STAGE,
+  API_ERROR_CODE,
+  TfmAuditDetails,
+  TfmDealCancellationResponse,
+  FACILITY_TYPE,
+} from '@ukef/dtfs2-common';
 import { generatePortalAuditDetails, generateTfmAuditDetails } from '@ukef/dtfs2-common/change-stream';
 import { withMongoIdPathParameterValidationTests } from '@ukef/dtfs2-common/test-cases-backend';
 import wipeDB from '../../../wipeDB';
@@ -17,7 +25,7 @@ const originalProcessEnv = { ...process.env };
 
 describe('/v1/tfm/deals/:dealId/cancellation/submit', () => {
   let dealId: string;
-  let ukefFacilityId: string;
+  const ukefFacilityId = 'ukefFacilityId';
   let submitDealCancellationUrl: string;
   let auditDetails: TfmAuditDetails;
   let tfmUserId: string;
@@ -28,13 +36,16 @@ describe('/v1/tfm/deals/:dealId/cancellation/submit', () => {
     effectiveFrom: new Date().valueOf(),
   };
 
+  const ukefDealId = 'ukefDealId';
+
   const newDeal = aDeal({
     dealType: DEALS.DEAL_TYPE.BSS_EWCS,
     submissionType: DEALS.SUBMISSION_TYPE.AIN,
+    details: { ukefDealId },
   }) as AnyObject;
 
   beforeAll(async () => {
-    await wipeDB.wipe([MONGO_DB_COLLECTIONS.TFM_DEALS, MONGO_DB_COLLECTIONS.DEALS]);
+    await wipeDB.wipe([MONGO_DB_COLLECTIONS.TFM_DEALS, MONGO_DB_COLLECTIONS.DEALS, MONGO_DB_COLLECTIONS.FACILITIES, MONGO_DB_COLLECTIONS.TFM_FACILITIES]);
 
     const tfmUser = await createTfmUser();
     tfmUserId = tfmUser._id;
@@ -42,16 +53,17 @@ describe('/v1/tfm/deals/:dealId/cancellation/submit', () => {
 
   beforeEach(async () => {
     const createDealResponse: { body: { _id: string } } = await createDeal({ deal: newDeal, user: aPortalUser() });
-    const createFacilityResponse: { body: { ukefFacilityId: string } } = await createFacility({
+    dealId = createDealResponse.body._id;
+
+    await createFacility({
       facility: {
         dealId,
-        ukefFacilityId: 'ukefFacilityId',
+        type: FACILITY_TYPE.BOND,
+        ukefFacilityId,
       },
       user: aPortalUser(),
     });
 
-    dealId = createDealResponse.body._id;
-    ukefFacilityId = createFacilityResponse.body.ukefFacilityId;
     auditDetails = generateTfmAuditDetails(tfmUserId);
     submitDealCancellationUrl = `/v1/tfm/deals/${dealId}/cancellation/submit`;
 
@@ -88,7 +100,7 @@ describe('/v1/tfm/deals/:dealId/cancellation/submit', () => {
 
   afterEach(async () => {
     process.env = { ...originalProcessEnv };
-    await wipeDB.wipe([MONGO_DB_COLLECTIONS.TFM_DEALS, MONGO_DB_COLLECTIONS.DEALS]);
+    await wipeDB.wipe([MONGO_DB_COLLECTIONS.TFM_DEALS, MONGO_DB_COLLECTIONS.DEALS, MONGO_DB_COLLECTIONS.FACILITIES, MONGO_DB_COLLECTIONS.TFM_FACILITIES]);
   });
 
   describe('POST /v1/tfm/deals/:dealId/cancellation/submit', () => {
@@ -126,9 +138,9 @@ describe('/v1/tfm/deals/:dealId/cancellation/submit', () => {
         const submitCancellationResponse = await testApi.post({ cancellation, auditDetails }).to(submitDealCancellationUrl);
 
         expect(submitCancellationResponse.body).toEqual({
-          cancelledDeal: expect.objectContaining({ _id: dealId }) as TfmDeal,
+          cancelledDealUkefId: ukefDealId,
           riskExpiredFacilityUkefIds: [ukefFacilityId],
-        });
+        } as TfmDealCancellationResponse);
         expect(submitCancellationResponse.status).toEqual(HttpStatusCode.Ok);
       });
 
--- a/dtfs-central-api/src/repositories/tfm-deals-repo/tfm-deal-cancellation.repo.ts
+++ b/dtfs-central-api/src/repositories/tfm-deals-repo/tfm-deal-cancellation.repo.ts
@@ -11,7 +11,6 @@ import {
   TFM_FACILITY_STAGE,
   TfmDeal,
   TfmDealCancellation,
-  TfmDealCancellationResponse,
   TfmDealCancellationWithStatus,
   TfmDealWithCancellation,
   TfmFacility,
@@ -19,7 +18,6 @@ import {
 import { generateAuditDatabaseRecordFromAuditDetails } from '@ukef/dtfs2-common/change-stream';
 import { flatten } from 'mongo-dot-notation';
 import { mongoDbClient } from '../../drivers/db-client';
-import { getUkefFacilityIds } from '../../helpers/get-ukef-facility-ids';
 
 export class TfmDealCancellationRepo {
   private static async getDealCollection(): Promise<Collection<WithoutId<TfmDeal>>> {
@@ -156,7 +154,7 @@ export class TfmDealCancellationRepo {
     cancellation: TfmDealCancellation;
     activity?: TfmActivity;
     auditDetails: AuditDetails;
-  }): Promise<TfmDealCancellationResponse> {
+  }): Promise<{ cancelledDeal: TfmDeal; riskExpiredFacilities: TfmFacility[] }> {
     if (!ObjectId.isValid(dealId)) {
       throw new InvalidDealIdError(dealId.toString());
     }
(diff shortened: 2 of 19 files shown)


dtfs-central-api/test-helpers/test-data/tfm-facility.ts
```diff
@@ -4,16 +4,13 @@ import { generateMockPortalUserAuditDatabaseRecord } from '@ukef/dtfs2-common/ch
...
-const facility = aFacility();
-const { dayCountBasis, interestPercentage, coverPercentage, value } = facility;
-
 export const aTfmFacility = (): TfmFacility => {
   const tfmFacilityId = new ObjectId();
 
   return {
     _id: tfmFacilityId,
     facilitySnapshot: {
-      ...facility,
```
MarRobSoftwire, line 16: This was not instantiating the facility within the `aTfmFacility` function. An object on this facility could be modified & it would affect all instances of `aTfmFacility`

trade-finance-manager-api/src/v1/services/deal-cancellation/deal-cancellation.service.test.ts
MarRobSoftwire, line 1: I dont think this should have changed!

libs/common/src/types/tfm-deal-cancellation.ts
```diff
@@ -1,10 +1,10 @@
 import z from 'zod';
+import { ObjectId } from 'mongodb';
 import { DEAL_CANCELLATION } from '../schemas/deal-cancellation';
-import { TfmDeal } from './mongo-db-models';
 
 export type TfmDealCancellation = z.infer<typeof DEAL_CANCELLATION>;
 
 export type TfmDealCancellationResponse = {
-  cancelledDeal: TfmDeal;
+  cancelledDealUkefId: string | ObjectId;
```
MarRobSoftwire, line 8: I've reverted this return type back to how it was before my previous PR
-we're only sending the data we need to tfm-api
- I've decoupled the response type from the api from the type returned in the repo layer
Happy to change this based on feedback, this feels cleanest IMO

dtfs-central-api/src/repositories/tfm-deals-repo/tfm-deal-cancellation.repo.ts
```diff
@@ -156,7 +154,7 @@ export class TfmDealCancellationRepo {
     cancellation: TfmDealCancellation;
     activity?: TfmActivity;
     auditDetails: AuditDetails;
-  }): Promise<TfmDealCancellationResponse> {
+  }): Promise<{ cancelledDeal: TfmDeal; riskExpiredFacilities: TfmFacility[] }> {
```
MarRobSoftwire, line 157: As I said in another comment, im decoupling this from the DTO which feels cleaner

dtfs-central-api/src/services/tfm/deal-cancellation.service.ts
```diff
@@ -52,16 +57,21 @@ export class DealCancellationService {
     };
 
     if (dealCancellationIsInFuture) {
-      return await TfmDealCancellationRepo.scheduleDealCancellation({ dealId, cancellation, activity, auditDetails });
+      const { cancelledDeal, riskExpiredFacilities } = await TfmDealCancellationRepo.scheduleDealCancellation({
```
MarRobSoftwire, line 60: update unit test coverage

dtfs-central-api/src/services/tfm/deal-cancellation.service.ts
```diff
@@ -100,6 +114,23 @@ export class DealCancellationService {
...
+    await Promise.all(
+      riskExpiredFacilities.map(({ _id: facilityId }) =>
+        PortalFacilityService.updateStatus({ facilityId, status: FACILITY_STATUS.RISK_EXPIRED, dealType, auditDetails }),
+      ),
+    );
+
+    return this.getTfmDealCancellationResponse({ cancelledDeal, riskExpiredFacilities });
+  }
+
+  private static getTfmDealCancellationResponse({ cancelledDeal, riskExpiredFacilities }: { cancelledDeal: TfmDeal; riskExpiredFacilities: TfmFacility[] }) {
```
MarRobSoftwire, line 126: add unit tests

dtfs-central-api/src/services/tfm/deal-cancellation.service.ts
```diff
@@ -100,6 +114,23 @@ export class DealCancellationService {
       dealType,
     });
 
-    return response;
+    await Promise.all(
```
MarRobSoftwire, line 117: update unit test coverage

dtfs-central-api/src/services/tfm/deal-cancellation.service.ts
MarRobSoftwire: Add docs

dtfs-central-api/api-tests/v1/tfm/deal-cancellation/tfm-submit-deal-cancellation-post.api-test.ts
```diff
@@ -126,9 +138,9 @@ describe('/v1/tfm/deals/:dealId/cancellation/submit', () => {
         const submitCancellationResponse = await testApi.post({ cancellation, auditDetails }).to(submitDealCancellationUrl);
 
         expect(submitCancellationResponse.body).toEqual({
-          cancelledDeal: expect.objectContaining({ _id: dealId }) as TfmDeal,
+          cancelledDealUkefId: ukefDealId,
```
ttbarnes, line 141: Can we add in the PR description why it's changed to return only an ID instead of the whole cancelled deal please?

dtfs-central-api/src/services/portal/facility.service.ts
```diff
@@ -0,0 +1,39 @@
...
+    auditDetails,
+    dealType,
+  }: {
+    facilityId: ObjectId | string;
+    status: FacilityStatus;
+    auditDetails: AuditDetails;
+    dealType: DealType;
+  }): Promise<void> {
+    if (dealType === DEAL_TYPE.GEF) {
+      await updateFacility({ facilityId, facilityUpdate: { status }, auditDetails });
```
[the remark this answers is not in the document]
ttbarnes: `status` field seems good to me. I didn't realise GEF facilities had no `status` field 😮

dtfs-central-api/src/services/tfm/deal-cancellation.service.ts
```diff
@@ -70,7 +80,13 @@ export class DealCancellationService {
...
+      riskExpiredFacilities.map(({ _id: facilityId }) =>
+        PortalFacilityService.updateStatus({ facilityId, status: FACILITY_STATUS.RISK_EXPIRED, dealType, auditDetails }),
```
ttbarnes, line 85: I haven't worked with Mongo for a while. Are we able to execute a bulk operation instead of mapping for individual calls?

dtfs-central-api/src/services/tfm/deal-cancellation.service.ts
```diff
@@ -100,6 +114,29 @@ export class DealCancellationService {
...
+   * @returns Tfm deal cancellation response DTO
+   */
+  public static getTfmDealCancellationResponse({ cancelledDeal, riskExpiredFacilities }: { cancelledDeal: TfmDeal; riskExpiredFacilities: TfmFacility[] }) {
+    const cancelledDealUkefId = getUkefDealId(cancelledDeal.dealSnapshot) as string;
+
+    return {
+      cancelledDealUkefId,
+      riskExpiredFacilityUkefIds: riskExpiredFacilities
+        .map(({ facilitySnapshot }) => facilitySnapshot.ukefFacilityId)
+        .filter((facilityId) => facilityId !== null),
```
ttbarnes, line 139: At this stage, are we ever going to have facilities that do not have a `ukefFacilityId`?

dtfs-central-api/src/v1/controllers/portal/facility/update-facility-status.controller.js
```diff
@@ -12,12 +12,12 @@ const withoutId = (obj) => {
 };
 
 /**
- * Update a facility status
+ * Update a BSS/EWCS facility status
```
ttbarnes, line 15: Can we rename the function and the logs as well please

trade-finance-manager-api/src/v1/services/deal-cancellation/deal-cancellation.service.ts
```diff
@@ -61,13 +61,7 @@ export class DealCancellationService {
   public static async submitDealCancellation({ dealId, cancellation, auditDetails }: SubmitDealCancellationParams) {
     console.info(`Submitting deal cancellation for ${dealId}`);
 
-    const { cancelledDeal, riskExpiredFacilityUkefIds } = await api.submitDealCancellation({ dealId, cancellation, auditDetails });
-
-    const ukefDealId = getUkefDealId(cancelledDeal.dealSnapshot);
-
-    if (!ukefDealId) {
-      throw new Error(`Failed to find deal Id on deal ${dealId} when submitting deal cancellation. No email has been sent`);
```
ttbarnes, line 69: Is it now impossible for this to happen?

ttbarnes (changes requested):
Nice work

trade-finance-manager-api/src/v1/services/deal-cancellation/deal-cancellation.service.ts
MarRobSoftwire: This should never occur (and was just added in as a type checking measure) since the deal is in TFM, it should have an ID. I can add this to `getTfmDealCancellationResponse` rather than using `as string` if you prefer?

dtfs-central-api/src/services/tfm/deal-cancellation.service.ts
MarRobSoftwire: We shouldn't, but otherwise typescript isn't happy with this. I guess theres a couple alternatives
- throw an error in the `map` if they're null (note that this means no email will be sent, but the deal will have been cancelled)
- assert with an `as string` that these are never  `null`

dtfs-central-api/src/services/tfm/deal-cancellation.service.ts
MarRobSoftwire: Its possible using a `updateMany` call but I cant find any existing functions doing this. I'd like to keep the scope of this smaller & reuse existing functions, rather than creating a new `updateFacilitiesByDealId` function. Whats your thoughts?

dtfs-central-api/api-tests/v1/tfm/deal-cancellation/tfm-submit-deal-cancellation-post.api-test.ts
MarRobSoftwire: I added
> Modify `TfmDealCancellationResponse` type back to a previous version & update controllers/tests to handle this
This was done as I realised the previous changes Id made werent entirely necessary & it made more sense to decouple the Repository response type from the DTO sent to tfm-api from the controller.

dtfs-central-api/src/services/portal/facility.service.ts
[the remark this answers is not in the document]
MarRobSoftwire: I'm considering changing this to `stage` since thats how its rendered in `gef-ui` (screenshot attatched). 
The pros of this
- makes whats rendered to the user consistent with whats stored in the database
The cons
- inconsistent with the `status` in BSS/EWCS facilities (though these two objects are completely different already)


image

dtfs-central-api/src/services/tfm/deal-cancellation.service.ts
MarRobSoftwire: Thinking about it, I definitely want to call the existing update BSS/EWCS facility function, since it adds a few extra fields & Id want to keep that logic DRY

ttbarnes: Do you know what the typical average and maximum amount of facilities we've seen?

dtfs-central-api/src/services/tfm/deal-cancellation.service.ts
ttbarnes: String assertion sounds nicest to me open to other approaches

trade-finance-manager-api/src/v1/services/deal-cancellation/deal-cancellation.service.ts
ttbarnes: I'm wondering if we need it at all. Like you say, this service will only be called once a deal is in TFM - when it'll have a `ukefDealId`. Seems it would just be adding unnecessary noise.

Also, if it does end up having no `ukefDealId`, I presume it'd fail somewhere else and throw an error anyway.

MarRobSoftwire: It wouldn't fail (on this endpoint call at least) & could send an email with `ukefDealId: undefined` - I think it might fail on other calls though & feels like such an edge case as to be ok?

dtfs-central-api/src/services/tfm/deal-cancellation.service.ts
MarRobSoftwire: Asking on slack :)

MarRobSoftwire: Abhi reckons 10 is about the most that is practical

dtfs-central-api/src/services/portal/facility.service.ts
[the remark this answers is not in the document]
MarRobSoftwire: Think i'll just stick with `status`

dtfs-central-api/src/services/tfm/deal-cancellation.service.ts
MarRobSoftwire: Raised a subtask for this https://ukef-dtfs.atlassian.net/browse/DTFS2-7625

dtfs-central-api/src/services/portal/facility.service.ts
```diff
@@ -0,0 +1,39 @@
...
```
ttbarnes, line 28: Not for this PR, at some point it'd be great to rename `updateFacility` to `updateGefFacility`

ttbarnes (dismissed):

PR #3932, opened 2024-11-15
feat(FN-3633): only send relevant data from utilisation report via api

ChristianMcc:
## Introduction :pencil2:
We are sending the whole contents of the file even if there are lots of extra headers, we want to only send the data we are going to validate to the API so the payload is smaller

Filter out irrelevant data from JSON sent to the API once we have converted xlsx to csv to json.


portal/server/controllers/utilisation-report-service/utilisation-report-upload/index.js
```diff
@@ -131,7 +132,10 @@ const postUtilisationReportUpload = async (req, res) => {
       return renderPageWithError(req, res, extractDataErrorSummary, extractDataError, dueReportPeriods);
     }
 
-    const { csvValidationErrors } = await api.generateValidationErrorsForUtilisationReportData(csvJson, bankId, userToken);
+    // We filter down the JSON to only send the necessary data to the API for validation
+    const filteredCsvJson = filterReportJsonToRelevantKeys(csvJson);
+
+    const { csvValidationErrors } = await api.generateValidationErrorsForUtilisationReportData(filteredCsvJson, bankId, userToken);
```
ttbarnes, line 138: Can we add a unit test for this please

portal/server/helpers/filterReportJson.test.js
```diff
@@ -0,0 +1,31 @@
...
+    const reportDataWithReportHeaderKeysOnly = [
+      {
+        [UTILISATION_REPORT_HEADERS.BASE_CURRENCY]: 'GBP',
+        [UTILISATION_REPORT_HEADERS.FACILITY_UTILISATION]: '100000',
+      },
+      {
+        [UTILISATION_REPORT_HEADERS.BASE_CURRENCY]: 'GBP',
+        [UTILISATION_REPORT_HEADERS.FACILITY_UTILISATION]: '200000',
+      },
+    ];
```
ttbarnes, line 28: Can we add a new line after each const please, would find easier to read

ttbarnes (changes requested):
Nice work, couple of minor comments

portal/server/controllers/utilisation-report-service/utilisation-report-upload/index.js
ChristianMcc: Sure, what are you expecting - would you be happy with a test which spys on what we pass to the generateValidationerrors api call to check its filtered down?

portal/server/controllers/utilisation-report-service/utilisation-report-upload/index.js
```diff
@@ -131,7 +132,10 @@ const postUtilisationReportUpload = async (req, res) => {
       return renderPageWithError(req, res, extractDataErrorSummary, extractDataError, dueReportPeriods);
     }
 
-    const { csvValidationErrors } = await api.generateValidationErrorsForUtilisationReportData(csvJson, bankId, userToken);
+    // We filter down the JSON to only send the necessary data to the API for validation
+    const filteredCsvJson = filterReportJsonToRelevantKeys(csvJson);
```
abhi-markan, line 136: Can we ensure `filteredCsvJson` is not empty before invoking `generateValidationErrorsForUtilisationReportData` function.

Should we also check `bankId` and `userToken`?

portal/server/controllers/utilisation-report-service/utilisation-report-upload/index.js
abhi-markan: Can we perform some validation on the response before de-structuring the object, wondering in case API call fails this de-structure will fail.

ttbarnes: Exactly my thoughts 😄

portal/server/helpers/filterReportJson.js
```diff
@@ -0,0 +1,23 @@
+import { UTILISATION_REPORT_HEADERS } from '@ukef/dtfs2-common';
+
+/**
+ * Iterates through an array of objects and removes any key value pairs
+ * from each object where the key isn't in the UTILISATION_REPORT_HEADERS const
+ * @param {Array<Object>} reportJson auth token
+ * @returns {Object} Filtered down object
+ */
+export const filterReportJsonToRelevantKeys = (reportJson) => {
```
abhi-markan, line 9: `map` could be slower when dealing with large number of rows I would suggest something like below to improve execution efficiency:

```javascript
export const filterReportJsonToRelevantKeys = (reportJson) => {
  // Convert relevant keys into a Set for O(1) lookups
  const relevantKeysSet = new Set(Object.values(UTILISATION_REPORT_HEADERS));

  // Map through the report rows and filter keys
  return reportJson.map((reportRow) => {
    return Object.fromEntries(
      Object.entries(reportRow).filter(([key]) => relevantKeysSet.has(key))
    );
  });
};
```

abhi-markan (changes requested):

portal/server/controllers/utilisation-report-service/utilisation-report-upload/index.js
ChristianMcc: Added a test which hopefully covers this @ttbarnes 

The destructuring error would be caught by the try/catch and handled as a problem with service error

portal/server/controllers/utilisation-report-service/utilisation-report-upload/index.test.js
```diff
@@ -161,5 +162,49 @@ describe('controllers/utilisation-report-service/utilisation-report-upload', ()
...
+        error: false,
+      });
+
+      jest.spyOn(api, 'generateValidationErrorsForUtilisationReportData').mockReturnValueOnce({ csvValidationErrors: [] });
+
+      // Act
+      await postUtilisationReportUpload(req, res);
+
+      // Assert
+      expect(api.generateValidationErrorsForUtilisationReportData).toHaveBeenCalledWith(expectedFilteredReportJson, expect.any(String), expect.any(String));
```
ttbarnes, line 207: This could be:

```ts
expect(api.generateValidationErrorsForUtilisationReportData).toHaveBeenCalledTimes(1);

const expected = filterReportJsonToRelevantKeys(reportJsonWithExtraKeys);

expect(api.generateValidationErrorsForUtilisationReportData).toHaveBeenCalledWith(expected);
```

portal/server/controllers/utilisation-report-service/utilisation-report-upload/index.js
ChristianMcc: For bankId and userToken this comes from session where this isn't explicitly tested in a lot of places (and you can't really get to this page without having a a real session) so I think just having an error being caught later on and getting problem with service would be fine here.

I'll add an explicit check that csvJson isn't empty and throw an error if it is completely empty

ChristianMcc: Thinking about this a bit more, all I'll be doing is adding throwing an error which will get caught and show the problem with service page, which is the same as the current behaviour. We don't have any requirements to show a special page to tell the user that the report can't be empty but I'm happy to bring this as a new requirement to UCD to think about?

portal/server/helpers/filterReportJson.js
ChristianMcc: For reference I ran the script with 5 iterations across both functions (original and proposed) with 1000 rows in the reportJson and 15 columns getting filtered down to 5 columns, I can do 31 -> 11 which is our real world example to double check

portal/server/controllers/utilisation-report-service/utilisation-report-upload/index.js
abhi-markan: Agree it will be but we wouldn't know what was the API call response in the try-catch.
Perhaps something like below:

```javascript
const response = ...
if (!response) {
throw new Error(`Empty response received from utilisation report validation ${response}`);
}
```

Above will speed up de-bugging.

portal/server/controllers/utilisation-report-service/utilisation-report-upload/index.js
ChristianMcc: Sure, I'll throw an error and throw an error earlier on before we even filter the json to check it's not an empty array, an array of empty objects post filtering is actually still a valid payload

portal/server/controllers/utilisation-report-service/utilisation-report-upload/index.js
ChristianMcc: FWIW I can add this if we expect the API to not error but still return a completely empty response - I'm just trying to minimise lines of code that I don't think could ever be hit (and reduce complexity of the function). If we want to cover this case anyway then I'm happy to just chuck in the above lines of code!

portal/server/controllers/utilisation-report-service/utilisation-report-upload/index.js
abhi-markan: Thank you, out of curiosity it is valid because it is an empty data set returned after removing frivolous dataset? If yes then should it proceed further or should we graceful throw an error back to the user?

portal/server/controllers/utilisation-report-service/utilisation-report-upload/index.js
abhi-markan: All good just wanted to make sure this case is covered but happy to leave it as it is if we can get the empty `response` in the catch trace stack.

portal/server/controllers/utilisation-report-service/utilisation-report-upload/index.js
ChristianMcc: It's valid because we'll send it off to the API and it will return that there are missing required columns in the spreadsheet which is a valid user journey (and will show as missing column errors on the page back to the user)

portal/server/helpers/filterReportJsonToRelevantKeys.js
```diff
@@ -0,0 +1,23 @@
+import { UTILISATION_REPORT_HEADERS } from '@ukef/dtfs2-common';
+
+/**
+ * Iterates through an array of objects and removes any key value pairs
+ * from each object where the key isn't in the UTILISATION_REPORT_HEADERS const
+ * @param {Array<Object>} reportJson auth token
```
ttbarnes, line 6: auth token?

ChristianMcc: how on earth did I get that in there, nice catch aha

ttbarnes (dismissed):

portal/server/controllers/utilisation-report-service/utilisation-report-upload/index.js
abhi-markan: Understood, could you please add a line stating that a valid `[]` should be send to the below API to ensure validation errors are shown to the user at the front-end please.

Apart from this all looks good. Thanks

abhi-markan (changes requested):
One minor documentation request please.

AlexBramhill (approved):
lgtm

abhi-markan (dismissed):

abhi-markan (approved):

ttbarnes (approved):

dtfs-central-api/src/services/utilisation-report-data-validator/utilisation-report-cell-validators/validate-csv-data/rows/index.ts
```diff
@@ -28,9 +28,8 @@ export const validateRows = (csvData: UtilisationReportCsvRowData[]): Utilisatio
     const ukefFacilityId = row['ukef facility id']?.value;
     const baseCurrencyValue = row['base currency']?.value;
     const facilityUtilisationValue = row['facility utilisation']?.value;
-    const exporterName = row['bank facility reference']?.value;
 
-    if (!baseCurrencyValue || !facilityUtilisationValue || !ukefFacilityId || !exporterName) {
+    if (!baseCurrencyValue || !facilityUtilisationValue || !ukefFacilityId) {
```
ttbarnes, line 32: Is there a unit test for this? Worrying if it's passing without any changes

ChristianMcc: there was no unit test for this, so going to add some as part of the https://ukef-dtfs.atlassian.net/browse/FN-3643 as well as using the const for pulling out the values instead of hard coded strings!

ttbarnes (approved):

PR #3934, opened 2024-11-19
feat(DTFS2-7637): replace scheduled status with pending

MarRobSoftwire:
## Introduction :pencil2:
Since we're adding a new `Pending Cancellation` status, the functions & existing status should be updated to reflect this new wording.

- Replace `TFM_DEAL_CANCELLATION_STATUS.SCHEDULED` with `PENDING`
- Update references to `scheduled for cancellation` to `pending cancellation`
- I have not updated: 
    -  `TfmDealCancellationRepo.scheduleDealCancellation` - I think this is better than `markCancellationAsPending` or something similar
    - `getScheduledCancellationBannerMessage` in the front end - the banner phrasing is `Deal ${ukefDealId} scheduled for cancellation on ${formattedDate}` so it makes sense to leave as is


dtfs-central-api/src/repositories/tfm-deals-repo/tfm-deal-cancellation.repo.ts
```diff
@@ -208,7 +208,7 @@ export class TfmDealCancellationRepo {
 
   /**
    * Schedules a deal cancellation (occurs when a deal cancellation is submitted but the effectiveFrom is in the future).
-   * In this instance, the deal and facility statuses remain the same, but the tfm cancellation object 'status' is updated to 'Scheduled'.
+   * In this instance, the deal and facility statuses remain the same, but the tfm cancellation object 'status' is updated to 'pending'.
```
MarRobSoftwire, line 211: I've left this function as is because I prefer the phrase `scheduleDealCancellation` to `markCancellationAsPending` or similar. Happy to hear any alternatives

trade-finance-manager-ui/server/controllers/helpers/get-success-banner-message.helper.ts
```diff
@@ -4,7 +4,7 @@ import api from '../../api';
 import { isDealCancellationEnabled } from './deal-cancellation-enabled.helper';
 
 /**
- * Returns deal cancellation message if the deal is scheduled for cancellation
+ * Returns deal cancellation message if the deal is pending cancellation
```
MarRobSoftwire, line 7: I've left this function the same name since it returns a message about the scheduled date

ttbarnes (approved):
Nice work and completely agree RE namings

ttbarnes (approved):

PR #3935, opened 2024-11-19
feat(DTFS2-7638): update portal status to pending

MarRobSoftwire:
## Introduction :pencil2:
When a deal is cancelled with the effective date in the future, the deal status should be updated to `pending cancellation`

- Update portal deal status to `Pending Cancellation` in `DealCancellationService.submitDealCancellation`

- Add new e2e tests


dtfs-central-api/src/services/tfm/deal-cancellation.service.submitDealCancellation.test.ts
```diff
@@ -274,12 +274,13 @@ describe('DealCancellationService', () => {
           expect(dealCancellationResponse).toEqual(DealCancellationService.getTfmDealCancellationResponse(mockRepositoryResponse));
         });
 
-        it('does not call PortalDealService.updateStatus', async () => {
+        it(`it calls PortalDealService.updateStatus with ${DEAL_STATUS.PENDING_CANCELLATION} status`, async () => {
```
ttbarnes, line 257: Can we/do we have a test for when `PortalDealService.updateStatus` errors?

e2e-tests/ukef/cypress/e2e/journeys/portal/deal-cancellation/BSS-EWCS/AIN-future-cancellation-updates-portal.spec.js
```diff
@@ -0,0 +1,78 @@
...
+      cy.clearCookie('dtfs-session');
+      cy.clearCookie('_csrf');
+      cy.getCookies().should('be.empty');
+      cy.forceVisit(TFM_URL);
+      cy.tfmLogin(PIM_USER_1);
+
+      cy.submitDealCancellation({ dealId, effectiveDate: tomorrow.date });
+    });
+
+    it(`displays deal status ${DEAL_STATUS.PENDING_CANCELLATION} on deal summary page`, () => {
```
ttbarnes, line 68: Can we split up the 2x status assertions into their own `it`'s please

e2e-tests/ukef/cypress/e2e/journeys/portal/deal-cancellation/BSS-EWCS/MIN-future-cancellation-updates-portal.spec.js
```diff
@@ -0,0 +1,81 @@
...
+      cy.clearCookie('dtfs-session');
+      cy.clearCookie('_csrf');
+      cy.getCookies().should('be.empty');
+      cy.forceVisit(TFM_URL);
+      cy.tfmLogin(PIM_USER_1);
+
+      cy.submitDealCancellation({ dealId, effectiveDate: tomorrow.date });
+    });
+
+    it(`displays deal status ${DEAL_STATUS.PENDING_CANCELLATION} on deal summary page`, () => {
```
ttbarnes, line 71: Same as above

ttbarnes (changes requested):
Nice work, few minor comments

dtfs-central-api/src/services/tfm/deal-cancellation.service.submitDealCancellation.test.ts
MarRobSoftwire: There's no logic in the service to catch the errors - I can add it but I dont think its necessary in the way we want to check controllers convert `ApiErrors` into a status and a body

ttbarnes: Thanks @MarRobSoftwire - had a look and understand now there's no e.g try/catch handling here and the error will bubble up to a controller 👍

ttbarnes (approved):

dtfs-central-api/src/services/tfm/deal-cancellation.service.ts
```diff
@@ -64,6 +64,17 @@ export class DealCancellationService {
         auditDetails,
       });
```
AlexBramhill, line 66: @MarRobSoftwire this function is starting to get rather large -- when we look to update tfm status could we refactor this down into

```
submitDealCancellation(){
    // Can this logic for audit details be abstracted at all?
    const user = await TfmUsersRepo.findOneUserById(auditDetails.id);

    if (!user) {
      throw new InvalidAuditDetailsError(`Supplied auditDetails 'id' ${auditDetails.id.toString()} does not correspond to a valid user`);
    }}
    // end audit detail abstraction comment

    if(DealCancellationService.isDealCancellationInFuture(cancellation)) {
        DealCancellationService.handleFutureDealCancellation() // this would contain all teh logic of updating portal/tfm deal status as well as scheduling the cancellation (which, actually should just be driven by the deal status & cancellation date?) 
    } else {
        DealCancellationService.handleCancelDeal()
    } // The else here might add clarity as to the condition of the code below it running

```

AlexBramhill (approved):
Just a comment to note for future work -- otherwise lgtm :)

dtfs-central-api/src/services/tfm/deal-cancellation.service.ts
MarRobSoftwire: This sounds good - had a little play (wont do now to keep PR slim) but think this could work
```
generateCancellationActivityFromAuditDetails
submitDealCancellationInFuture
submitDealCancellationInPast
```

leaving something like
```
  public static async submitDealCancellation(
    dealId: ObjectId | string,
    cancellation: TfmDealCancellation,
    auditDetails: TfmAuditDetails,
  ): Promise<TfmDealCancellationResponse> {
    const effectiveFromDate = toDate(cancellation.effectiveFrom);
    const endOfToday = endOfDay(new Date());

    const dealCancellationIsInFuture = isAfter(effectiveFromDate, endOfToday);

    if (dealCancellationIsInFuture) {
      return await this.submitDealCancellationInFuture(dealId, cancellation, auditDetails);
    }
    return await this.submitDealCancellationInPast(dealId, cancellation, auditDetails);
  }
```

AlexBramhill: Any reason we're not using `isDealCancellationInFuture` for the boolean naming?

Looking a lot nicer! Great job.

I'm wondering about `submitDealCancellationInFuture` and `submitDealCancellationInPast` as names -- I'm not overly keen but maybe it is better than `handleFutureDealCancellation`? I think `submitFutureDealCancellation` does add ambiguity as to what is in the future, so maybe this is a good name!

PR #3936, opened 2024-11-19
Feat(DTFS2-7467): add risk expired filter

MarRobSoftwire:
## Introduction :pencil2:


portal-api/src/v1/helpers/escapeOperators.js
```diff
@@ -73,6 +74,11 @@ const recursivelyReplaceEscapeOperators = (filters, result = {}) => {
       // These last two if statements handle the lowest level cases
     } else if (key === 'KEYWORD') {
       result.$regex = escapeStringRegexp(filters[key]);
+    } else if (key === 'hasBeenIssued') {
```
MarRobSoftwire, line 77: Update this logic?

portal/server/controllers/dashboard/facilities/index.js
```diff
@@ -70,8 +69,7 @@ exports.allFacilities = async (req, res) => {
   let activeSortByOrder = req.session.sortBy ?? CONSTANTS.SORT_BY.DEFAULT;
 
   if (Object.keys(req.body).length) {
-    const sanitisedBody = sanitiseBody(req.body);
-    req.session.dashboardFilters = sanitisedBody;
```
MarRobSoftwire, line 74: This was just mapping the `hasBeenIssued` values from `'true'`/`'false'`

portal/server/controllers/dashboard/facilities/selected-filters.js
```diff
@@ -31,10 +16,11 @@ const mapIssuedValueToText = (hasBeenIssued) => {
  * @example ( 'Bank facility stage', 'hasBeenIssued', [ true, false ] )
  * @returns generateSelectedFiltersObjectWithMappedValues('Facility stage', 'hasBeenIssued', [ {value: true, mappedValue: 'Issued' }])
  */
-const selectedHasBeenIssuedFilters = (heading, fieldName, submittedFilters) => {
+const selectedStageFilters = (heading, fieldName, submittedFilters) => {
   const mappedFilters = submittedFilters.map((value) => ({
     value,
-    mappedValue: mapIssuedValueToText(value),
+    // TODO: work out if this is needed
+    mappedValue: value,
```
MarRobSoftwire, line 23: Is this necessary?

portal-api/src/v1/helpers/escapeOperators.js
```diff
@@ -70,9 +71,15 @@ const recursivelyReplaceEscapeOperators = (filters, result = {}) => {
...
-      // These last two if statements handle the lowest level cases
+      // These last three if statements handle the lowest level cases
     } else if (key === 'KEYWORD') {
       result.$regex = escapeStringRegexp(filters[key]);
+    } else if (key === 'hasBeenIssued') {
+      result.hasBeenIssued = {
+        $eq: filters.hasBeenIssued,
+      };
+      // When filtering by `hasBeenIssued`, want to exclude cases where `facilityStage` is'Risk expired'
+      result.facilityStage = { $ne: FACILITY_STATUS.RISK_EXPIRED };
```
MarRobSoftwire, line 82: I don't like this being necessary but I cant think of a better way to ignore `Risk expired`/`Abandoned` facilities. I think it has to be in portal-api without a big refactor of this function (to map `NE` into `$ne`) but this feels like a can of worms...

portal/server/controllers/dashboard/facilities/facilities-filters-query.js
```diff
@@ -51,15 +55,28 @@ const dashboardFacilitiesFiltersQuery = (filters, user) => {
...
+          OR: [],
+        };
+        filterValue.forEach((value) => {
+          if (value === FACILITY_STATUS.RISK_EXPIRED) {
+            fieldFilter.OR.push({ facilityStage: FACILITY_STATUS.RISK_EXPIRED });
+          } else {
+            fieldFilter.OR.push({
+              hasBeenIssued: value === BESPOKE_FILTER_VALUES.FACILITIES.ISSUED,
+            });
+          }
```
MarRobSoftwire, line 69: this isnt the most extensible, when `Abandoned` is added in might want to refactor this

MarRobSoftwire: I think leave for now though

portal/server/constants/field-names.js
```diff
@@ -11,14 +11,14 @@ const DEAL = {
...
   TYPE: 'type',
-  HAS_BEEN_ISSUED: 'hasBeenIssued',
   NAME: 'name',
   UKEF_FACILITY_ID: 'ukefFacilityId',
   CURRENCY: 'currency',
   VALUE: 'value',
   DEAL_SUBMISSION_TYPE: 'deal.submissionType',
   EXPORTER_COMPANY_NAME: 'deal.exporter.companyName',
   CREATED_BY: 'createdBy',
+  STAGE: 'stage',
```
AlexBramhill, line 21: This entire area of the code is really rather confusing. Would it be possible to add a comment to the top of this file as to the use of this file (at a glance, it seems to be used in collab with content-strings?)

Not for this PR - This file adds an additional layer of (in my opinion) probably unneeded abstraction. We're tying our types to unchecked strings -- ie if we changed companyName, we'd probably miss updating it here... This is really rather confusing, great job navigating this.

portal/server/controllers/dashboard/facilities/facilities-filters-query.js
AlexBramhill: I was thinking the same thing about refactor -- Happy to think about it during abandon. Even if its just breaking down this large function into smaller ones, and revisiting the comments to make this easier for someone with no context dropping into this file!

portal/server/controllers/dashboard/facilities/selected-filters.js
```diff
@@ -78,12 +38,12 @@ const selectedFilters = (submittedFilters) => {
     selected.push(obj);
   }
 
-  if (submittedFilters[CONSTANTS.FIELD_NAMES.FACILITY.HAS_BEEN_ISSUED]) {
+  if (submittedFilters.stage) {
```
AlexBramhill, line 41: Should this not reference the constants file for field names (I don't really think this adds much to this file and happy to avoid this request if there's logic to this)

portal/server/controllers/dashboard/facilities/template-filters.js
```diff
@@ -26,20 +26,27 @@ const typeFilters = (submittedFilters) => {
...
+      value: BESPOKE_FILTER_VALUES.FACILITIES.ISSUED,
     },
     {
       text: BESPOKE_FILTER_VALUES.FACILITIES.UNISSUED,
-      value: FACILITY_HAS_BEEN_ISSUED.UNISSUED,
+      value: BESPOKE_FILTER_VALUES.FACILITIES.UNISSUED,
     },
   ];
 
+  if (isTfmDealCancellationFeatureFlagEnabled()) {
```
AlexBramhill, line 43: Is this new `risk_expired` the only additional status to issued/unissued?

I know `Issued` and `unissued` status are derived from the database as `hasBeenIssued` (or similar)... Is this the first new status apart from these two?

If it is -- We should really think about a migration. Or at least raise a ticket to migrate this data...Or consider our options...

portal/server/controllers/dashboard/filters/generate-template-filters.js
```diff
@@ -12,17 +12,7 @@ const { SUBMISSION_TYPE } = require('../../../constants');
...
-  const hasSubmittedFilters = Object.keys(submittedFilters).length;
-
-  if (hasSubmittedFilters) {
-    const filterHasBeenSubmitted = submittedFilters[field] && String(submittedFilters[field]).includes(String(value));
-
-    if (filterHasBeenSubmitted) {
-      checked = true;
-    }
-  }
+  const checked = !!(Object.keys(submittedFilters).length && submittedFilters[field] && String(submittedFilters[field]).includes(String(value)));
```
AlexBramhill, line 15: `isChecked` please

AlexBramhill: Also could we consider making this a little clearer?
```
const isFieldInSubmittedFilters = Object.keys(submittedFilters).length &&!!submittedFilters[field]; // does this have any issue with falsey things?
const isValueInField = isFieldInSubmittedFilters && String(submittedFilters[field]).includes(String(value));

const isChecked = isFieldInSubmittedFilters && isValueInField;
```
or something like this? Let me know what you think -- I'm finding this review a tad hard to dive into as I've never seen this code before

portal/server/controllers/dashboard/filters/generate-template-filters.js
```diff
@@ -12,17 +12,7 @@ const { SUBMISSION_TYPE } = require('../../../constants');
  * @returns { text: 'GEF deals', value: 'GEF', checked: true }
```
AlexBramhill, line 12: Could you update the params of this jsdoc please to be correct to the parameters below?

portal-api/src/v1/helpers/escapeOperators.js
AlexBramhill: Yeah, this is probably a larger conversation with the concept of the `hasBeenIssued` boolean... Left a comment elsewhere about it

portal/server/controllers/dashboard/facilities/facilities-filters-query.js
```diff
@@ -40,7 +43,8 @@ const dashboardFacilitiesFiltersQuery = (filters, user) => {
       const fieldName = Object.keys(filterObj)[0];
       const filterValue = filterObj[fieldName];
 
-      const isKeywordField = fieldName === CONTENT_STRINGS.DASHBOARD_FILTERS.BESPOKE_FIELD_NAMES.KEYWORD;
+      const isKeywordField = fieldName === BESPOKE_FIELD_NAMES.KEYWORD;
+      const isStageField = fieldName === BESPOKE_FIELD_NAMES.STAGE;
```
AlexBramhill, line 47: Once we start doing things like this, we should start thinking about a constant and a switch maybe (the constant can just be declared locally and not exported), but not for this PR

AlexBramhill (commented):
This is some really tough code. You've done a really good job sinking your teeth into it, and also looking to add value through small improvements when possible. Let me know when it's ready to re-review

portal-api/src/v1/helpers/escapeOperators.js
MarRobSoftwire: Oh yeah 100%, and thats an even bigger can of worms

portal/server/controllers/dashboard/facilities/facilities-filters-query.js
MarRobSoftwire: i think i can quite easily make this a switch which would be a bit nicer

MarRobSoftwire: I want to make it use `map` rather than `forEach` and `push` but had a play and was quite fiddly

portal/server/controllers/dashboard/facilities/facilities-filters-query.test.js
```diff
@@ -54,7 +54,7 @@ describe('controllers/dashboard/facilities - filters query', () => {
           OR: [{ [CONSTANTS.FIELD_NAMES.FACILITY.TYPE]: mockFilters[0].type[0] }, { [CONSTANTS.FIELD_NAMES.FACILITY.TYPE]: mockFilters[0].type[1] }],
         },
         {
-          OR: [{ [CONSTANTS.FIELD_NAMES.FACILITY.HAS_BEEN_ISSUED]: mockFilters[1].hasBeenIssued[0] }],
+          OR: [{ [CONSTANTS.FIELD_NAMES.FACILITY.HAS_BEEN_ISSUED]: true }],
```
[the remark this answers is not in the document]
MarRobSoftwire: I'd done this incorrectly - good spot! But this needs updating because we're mapping `stage` into `hasBeenIssued` and `facilityStage`

portal/server/controllers/dashboard/facilities/selected-filters.js
MarRobSoftwire: I think I did this before I'd made that constant - as you say Im not the keenest on it but I'll add it back to preserve the pattern

portal/server/controllers/dashboard/facilities/template-filters.js
MarRobSoftwire: I think currently we dont need a migration, but yes I'd love to migrate all these deals from `hasBeenIssued` to just using `facilityStage`. This could be a bit trickier on bss/ewcs deals since they already have `facilityStage` and it takes some other values that I dont understand.
Happy to chat this through if you're concerned about this but (hopefully) I've not done anything needing a migration (yet!)

portal/server/controllers/dashboard/filters/generate-template-filters.js
MarRobSoftwire: on the `isChecked` point, there is some precedence to keep this the same since it will be passed straight into a nunjucks filter

MarRobSoftwire: But i think id prefer this to be more readable than more compact, given the weirdness in this area

portal-api/src/v1/helpers/escapeOperators.js
```diff
@@ -70,9 +71,15 @@ const recursivelyReplaceEscapeOperators = (filters, result = {}) => {
     } else if (typeof filters[key] === 'object' && filters[key] !== null) {
       result[key] = {};
       recursivelyReplaceEscapeOperators(filters[key], result[key]);
-      // These last two if statements handle the lowest level cases
+      // These last three if statements handle the lowest level cases
     } else if (key === 'KEYWORD') {
       result.$regex = escapeStringRegexp(filters[key]);
+    } else if (key === 'hasBeenIssued') {
```
ttbarnes, line 77: Can we add a unit test please

portal/server/controllers/dashboard/facilities/facilities-filters-query.js
```diff
@@ -40,38 +43,55 @@ const dashboardFacilitiesFiltersQuery = (filters, user) => {
       const fieldName = Object.keys(filterObj)[0];
       const filterValue = filterObj[fieldName];
 
-      const isKeywordField = fieldName === CONTENT_STRINGS.DASHBOARD_FILTERS.BESPOKE_FIELD_NAMES.KEYWORD;
+      switch (fieldName) {
+        case BESPOKE_FIELD_NAMES.KEYWORD: {
```
ttbarnes, line 47: How about something like this instead, whilst we're here, easer to read/understand

```js
case BESPOKE_FIELD_NAMES.KEYWORD: {
  const keywordValue = filterValue[0];

  const keywordFilter = {
    OR: keywordQuery(keywordValue),
  };

  query.AND.push(keywordFilter);

  break;
}
```

portal/server/controllers/dashboard/facilities/template-filters.ff-test.js
```diff
@@ -0,0 +1,36 @@
+import { FACILITY_STATUS } from '@ukef/dtfs2-common';
+import { stageFilters } from './template-filters';
+import { FIELD_NAMES } from '../../../constants';
+import { DASHBOARD_FILTERS } from '../../../content-strings';
+import { generateFiltersArray } from '../filters/generate-template-filters';
+
+describe('controllers/dashboard/facilities - template-filters', () => {
+  describe('stageFilters - ff enabled', () => {
```
ttbarnes, line 8: Can we say "feature flag" instead of `ff` please

Not for this PR but I think in all feature flag related tests/descriptions/documentation, we need to mention explicitly _which_ feature flag it is.

E.g looking at this, I have no idea which feature flag affects this functionality

ttbarnes: In other tests we have this 

```
```

ttbarnes: Seems the existing feature flag tests/descriptions are mixed and matched 😞

ttbarnes (changes requested):
Fantastic work @MarRobSoftwire really nice. Appreciate the filtering/querying functionality is very in-depth!

portal/server/controllers/dashboard/facilities/template-filters.ff-test.js
MarRobSoftwire: I think `describe('when `FF_TFM_FACILITY_END_DATE_ENABLED` is set to true', () => {` is from before Abhi introduced the `ff-test` tests, they typically contain a `beforeEach` setting the FF to true

portal-api/src/v1/helpers/escapeOperators.test.js
MarRobSoftwire, line 1: I've done a bit of refactoring here - improving the test descriptions & making it more clearly in `arrange/act/assert` format

ttbarnes: Nice 😄

portal/server/controllers/dashboard/facilities/template-filters.ff-test.js
```diff
@@ -0,0 +1,36 @@
+import { FACILITY_STATUS } from '@ukef/dtfs2-common';
+import { stageFilters } from './template-filters';
+import { FIELD_NAMES } from '../../../constants';
+import { DASHBOARD_FILTERS } from '../../../content-strings';
+import { generateFiltersArray } from '../filters/generate-template-filters';
+
+describe('controllers/dashboard/facilities - template-filters', () => {
+  describe('stageFilters - FF_TFM_FACILITY_END_DATE_ENABLED enabled', () => {
```
ttbarnes, line 8: Can this be "facility stage filters"  instead of "stageFilters", threw me a little bit

.github/workflows/test.yml
```diff
@@ -815,6 +815,41 @@ jobs:
...
           [
+            '.portal/cypress/e2e/journeys/*.spec.js',
+            '.portal/cypress/e2e/journeys/homepage/**/*.spec.js',
+            '.portal/cypress/e2e/journeys/admin/**/*.spec.js',
+            '.portal/cypress/e2e/journeys/checker/**/*.spec.js',
+            '.portal/cypress/e2e/journeys/dashboard/dashboard-deals/filter-by/*.spec.js',
+            '.portal/cypress/e2e/journeys/dashboard/dashboard-deals/filters/*.spec.js',
+            '.portal/cypress/e2e/journeys/dashboard/dashboard-deals/sort/*.spec.js',
+            '.portal/cypress/e2e/journeys/dashboard/dashboard-facilities/filter-by/*.spec.js',
+            '.portal/cypress/e2e/journeys/dashboard/dashboard-facilities/filters/*.spec.js',
```
MarRobSoftwire, line 826: These are affected by the feature flag but it probably makes sense to test all of them to prevent any issues appearing once the FF is turned on

portal/server/controllers/dashboard/filters/generate-template-filters.js
AlexBramhill: Happy with whatever works for this area of the code

portal/server/controllers/dashboard/facilities/selected-filters.js
AlexBramhill: Thanks!

portal/server/controllers/dashboard/facilities/facilities-filters-query.js
AlexBramhill: This wsa amazing good job. Thanks. A reducer would probably be better? Resolve once youve read -- just a future suggestion

AlexBramhill (commented):
Super minor comments, let me know when you're ready for me to approve

.github/workflows/test.yml
MarRobSoftwire: the two facility filtering ones are, the rest arent

portal/server/constants/field-names.js
```diff
@@ -9,16 +13,20 @@ const DEAL = {
   BSS_EWCS_UKEF_DEAL_ID: 'details.ukefDealId',
 };
 
+/**
+ * These are the field names used on the facility filter query object
+ * This is sent to portal-api to query the database
```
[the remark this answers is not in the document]
MarRobSoftwire: Maybe stage isnt supposed to be here, I'm confused now

these ones are ( I think)
```
  DEAL_SUBMISSION_TYPE: 'deal.submissionType',
  EXPORTER_COMPANY_NAME: 'deal.exporter.companyName',
```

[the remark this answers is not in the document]
MarRobSoftwire: portal\server\controllers\dashboard\facilities\template-filters.js uses `deal.submissionType` so it looks like these are used for the templating. I'm still a bit confused but I'd really like to just get rid of this stuff - portal-api should really be handling all this :/

.github/workflows/test.yml
```diff
@@ -872,6 +872,40 @@ jobs:
...
+            './portal/cypress/e2e/journeys/dashboard/dashboard-facilities/filter-by/*.spec.js',
+            './portal/cypress/e2e/journeys/dashboard/dashboard-facilities/filters/*.spec.js',
+            './portal/cypress/e2e/journeys/dashboard/dashboard-facilities/sort/*.spec.js',
+            './portal/cypress/e2e/journeys/feedback/**/*.spec.js',
+            './portal/cypress/e2e/journeys/login/**/*.spec.js',
+            './portal/cypress/e2e/journeys/maker/check-deal-details/*.spec.js',
+            './portal/cypress/e2e/journeys/maker/clone-deal/*.spec.js',
+            './portal/cypress/e2e/journeys/maker/dashboard/**/*.spec.js',
+            './portal/cypress/e2e/journeys/maker/submit-a-deal-for-review/*.spec.js',
+            './portal/cypress/e2e/journeys/maker/submit-a-deal-for-review-with-issued-and-unissued-facilities-and-resubmit-after-checker-returns/*.spec.js',
```
BethThomas141, line 891: Out of interest, is there a reason to not just do e.g.
`./portal/cypress/e2e/journeys/maker/`
instead of all of the subsets like: 
```
'./portal/cypress/e2e/journeys/maker/clone-deal/*.spec.js',
'./portal/cypress/e2e/journeys/maker/dashboard/**/*.spec.js',
'./portal/cypress/e2e/journeys/maker/submit-a-deal-for-review/*.spec.js',
```
(Aware this is likely copied from the portal structure above)

.github/workflows/test.yml
```diff
@@ -872,6 +872,40 @@ jobs:
...
+            './portal/cypress/e2e/journeys/maker/submit-issued-facilities-for-review/*.spec.js',
+            './portal/cypress/e2e/journeys/maker/bank-details/*.spec.js',
+            './portal/cypress/e2e/journeys/maker/eligibility/*.spec.js',
+            './portal/cypress/e2e/journeys/maker/abandon-a-deal/*.spec.js',
+            './portal/cypress/e2e/journeys/maker/before-you-start/*.spec.js',
+            './portal/cypress/e2e/journeys/maker/bss-mandatory-criteria/*.spec.js',
+            './portal/cypress/e2e/journeys/maker/edit-deal-name/*.spec.js',
+            './portal/cypress/e2e/journeys/maker/proceed-to-review/*.spec.js',
+            './portal/cypress/e2e/journeys/maker/select-scheme/*.spec.js',
```
BethThomas141, line 900: Out of interest, is there a reason we don't do e.g.:

`'./portal/cypress/e2e/journeys/maker/**/*.spec.js',`

instead of e.g. 

```
'./portal/cypress/e2e/journeys/maker/eligibility/*.spec.js',
'./portal/cypress/e2e/journeys/maker/abandon-a-deal/*.spec.js',
'./portal/cypress/e2e/journeys/maker/before-you-start/*.spec.js',
...
```

(I might not understand the **/* correctly though)

AlexBramhill (dismissed):

.github/workflows/test.yml
MarRobSoftwire: I believe these are some of the tests @ttbarnes split up to reduce e2e run time - its starting to look a bit ugly and maybe there is a better way of dealing with this 🤷

ttbarnes: Correct - they are split up to reduce E2E time. Prior to this change, some individual E2E runners took I think 20-30mins to run 😱 Last time I checked, I think the entire suite took ~10-12mins, quicker than the API tests.

The trade off as you've alluded to, is extra maintenance with the `test.yml`, but I think it's worth it - just need to remember to update it if there is a new E2E directory, or if a directory becomes busy.

I am not sure how the `test.yml` could become prettier, it's all about the directories and wildcards for those directories could be uglier. Open to any ideas as always though 😊

.github/workflows/test.yml
```diff
@@ -873,10 +873,41 @@ jobs:
       matrix:
         spec:
           [
+            './portal/cypress/e2e/journeys/*.spec.js',
```
ttbarnes, line 876: Can you add the reason for this change in the PR description please

portal-api/src/v1/helpers/escapeOperators.js
```diff
@@ -70,9 +71,15 @@ const recursivelyReplaceEscapeOperators = (filters, result = {}) => {
     } else if (typeof filters[key] === 'object' && filters[key] !== null) {
       result[key] = {};
       recursivelyReplaceEscapeOperators(filters[key], result[key]);
-      // These last two if statements handle the lowest level cases
+      // These last three if statements handle the lowest level cases
```
ttbarnes, line 74:
```suggestion
      // These last few statements handle the lowest level cases
```

Could be very easy to miss this if a statement is added or removed (good spot)

portal-api/src/v1/helpers/escapeOperators.test.js
```diff
@@ -298,20 +417,23 @@ describe('escapeOperators function', () => {
         },
       ],
     };
-    const result = escapeOperators(filter);
     expect(result).toEqual(expected);
   });
 
-  // Tests that the AND operator key is deleted after escaping
-  it('should test delete and operator key after escaping', () => {
+  it('deletes AND operator key after escaping', () => {
```
ttbarnes, line 423: Can we stick to the "it should" naming please?

```
it('should = 4712 results
it('should return = 1796 results
it('returns = 861 results
```

portal/server/controllers/dashboard/facilities/facilities-filters-query.test.js
```diff
@@ -113,6 +113,69 @@ describe('controllers/dashboard/facilities - filters query', () => {
...
+      ],
+    };
+
+    expect(result).toEqual(expected);
+  });
+
+  it(`should return hasBeenIssued filter when stage is ${DASHBOARD_FILTERS.BESPOKE_FILTER_VALUES.FACILITIES.UNISSUED}`, () => {
+    const mockFilters = [
+      {
+        [CONSTANTS.FIELD_NAMES.FACILITY.STAGE]: [[DASHBOARD_FILTERS.BESPOKE_FILTER_VALUES.FACILITIES.UNISSUED]],
```
ttbarnes, line 140: Are we able to destructure so we don't have `DASHBOARD_FILTERS.BESPOKE_FILTER_VALUES.FACILITIES.UNISSUED` in several places?

ttbarnes (changes requested):
Great work, few minor comments

.github/workflows/test.yml
MarRobSoftwire: Done :)

ttbarnes (approved):

PR #3937, opened 2024-11-19
feat(FN-3638): turn off fixed fee adjustments

RLavender98:
## Introduction :pencil2:
Fixed fee adjustments rely on cover end dates existing which we discovered is not the case for all facilities banks are reporting on. This was preventing upload of reports and complete use of the utilisation report service for certain banks. However, the fixed fee calculation has not been finalised and so is not being used by PDC.

In this PR we disable fixed fee completely. Removing any dependency on facility values other than 'value' and 'coverPercentage' from the utilisation reporting service. This should make the service usable for all banks, just without the fixed fee functionality. Fixed fee adjustments will be reimplemented when the calculation is finalised with support for facilities with no cover end date.

- Remove usages of facility fields other than 'coverPercentage' and 'value' from utilisation report related code
- Turn off fixed fee by:
	- defaulting fixed fee to zero for new facilities
	- storing fixed fee as zero when generating keying data
	- directly setting the adjustment to always be zero instead of using a calculation
- Add test skips to tests that will be reenabled pretty much as is when we turn fixed fee back on
- Add comments to unused functions which are left in as will be used pretty much as is when we turn fixed fee back on

### DIFF
--- a/dtfs-central-api/api-tests/v1/utilisation-reports/post-keying-data.api-test.ts
+++ b/dtfs-central-api/api-tests/v1/utilisation-reports/post-keying-data.api-test.ts
@@ -372,8 +372,9 @@ describe(`POST ${BASE_URL}`, () => {
   });
 
   /**
-   * This skip should be removed when the fixed fee adjustments are turned back on
-   * and can take non-zero values.
+   * This test is skipped because fixed fee adjustments are temporarily turned off.
+   *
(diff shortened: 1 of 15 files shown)


ttbarnes (approved):
LGTM

ChristianMcc (approved):
LGTM!

PR #3939, opened 2024-11-19
fix(DTFS2-6616): variable consistency

abhi-markan:
## Introduction :pencil2:
Ensure `facilityIdentifier` is used as a standard name for facility ID.


ChristianMcc (approved):

ttbarnes (commented):
Nice work

PR #3942, opened 2024-11-20
fix(FN-3641): improve report upload reliability

RLavender98:
## Introduction :pencil2:
When errors were being thrown as part of the report upload on production the logs were not displaying the originating error because we were running lots of queries in parallel which caused rolling the transaction back to fail and so the surfaced error message was a generic transaction error message.

In this PR we resolve that by adding in an additional console.error of the thrown error so we can view the original error if rolling back the transaction ever fails. As well as serialising the facility utilisation data initialisation so that if an error is thrown in a similar way again the transaction will roll back happily.

To ensure serialising the queries does not have a performance hit I timed the implementation before and after these changes for the facility utilisation data initialisation:
- For a 300 row report with all newly reported on facilities (a lot more than we ever expect to actually happen):
	- before: 1119.7767919999897 milliseconds
	- after: 595.2127079998609 milliseconds
- For a 303 row report with 3 newly reported on facilities (this is a more realistic use case):
	- before: 475.7242509999778 milliseconds
	- after: 95.49562500001048 milliseconds

Overall these changes are actually a fair amount quicker and more robust.

- Add `console.error` to `executeWithSqlTransaction` error handling before attempting to rollback
- Create `FacilityUtiliastionDataEntity`s for new facilities in a for loop rather than a map so the db queries don't try and run in parallel when there is a single connection
- Refactor the SQL existence check into just one db query to make the insertion more performant

### DIFF
--- a/dtfs-central-api/src/constants/batch-saving.ts
+++ b/dtfs-central-api/src/constants/batch-saving.ts
@@ -1,0 +1,1 @@
+export const CHUNK_SIZE_FOR_SQL_BATCH_SAVING = 100;
--- a/dtfs-central-api/src/helpers/execute-with-sql-transaction.test.ts
+++ b/dtfs-central-api/src/helpers/execute-with-sql-transaction.test.ts
@@ -144,7 +144,7 @@ describe('executeWithSqlTransaction', () => {
     await expect(executeWithSqlTransaction(functionToExecute)).rejects.toThrow(TransactionFailedError.forApiError(customError));
   });
 
-  it('logs error even if rolling back the transaction fails', async () => {
+  it('logs error if rolling back the transaction fails', async () => {
     // Arrange
     const originalError = new Error('Some error');
     const functionToExecute = jest.fn().mockRejectedValue(originalError);
--- a/dtfs-central-api/src/helpers/execute-with-sql-transaction.test.ts
+++ b/dtfs-central-api/src/helpers/execute-with-sql-transaction.test.ts
@@ -144,6 +144,20 @@ describe('executeWithSqlTransaction', () => {
     await expect(executeWithSqlTransaction(functionToExecute)).rejects.toThrow(TransactionFailedError.forApiError(customError));
   });
 
+  it('logs error even if rolling back the transaction fails', async () => {
+    // Arrange
+    const originalError = new Error('Some error');
+    const functionToExecute = jest.fn().mockRejectedValue(originalError);
+    const consoleErrorSpy = jest.spyOn(console, 'error');
+    const secondaryError = new Error('Cannot rollback transaction');
+    mockRollbackTransaction.mockRejectedValueOnce(secondaryError);
+
+    // Act / Assert
+    await expect(executeWithSqlTransaction(functionToExecute)).rejects.toThrow(secondaryError);
+    expect(consoleErrorSpy).toHaveBeenCalledTimes(1);
+    expect(consoleErrorSpy).toHaveBeenCalledWith('Error thrown within SQL transaction: %s', originalError);
+  });
+
   it('returns the return value of the supplied function', async () => {
     // Arrange
     const returnValue = {
--- a/dtfs-central-api/src/helpers/execute-with-sql-transaction.test.ts
+++ b/dtfs-central-api/src/helpers/execute-with-sql-transaction.test.ts
@@ -155,7 +155,7 @@ describe('executeWithSqlTransaction', () => {
     // Act / Assert
     await expect(executeWithSqlTransaction(functionToExecute)).rejects.toThrow(secondaryError);
     expect(consoleErrorSpy).toHaveBeenCalledTimes(1);
-    expect(consoleErrorSpy).toHaveBeenCalledWith('Error thrown within SQL transaction: %s', originalError);
+    expect(consoleErrorSpy).toHaveBeenCalledWith('Error thrown within SQL transaction: %o', originalError);
   });
 
   it('returns the return value of the supplied function', async () => {
--- a/dtfs-central-api/src/helpers/execute-with-sql-transaction.ts
+++ b/dtfs-central-api/src/helpers/execute-with-sql-transaction.ts
@@ -71,7 +71,7 @@ export const executeWithSqlTransaction = async <ReturnValue>(functionToExecute: 
 
     return result;
   } catch (error) {
-    console.error('Error thrown within SQL transaction: %s', error);
+    console.error('Error thrown within SQL transaction: %o', error);
 
     await queryRunner.rollbackTransaction();
     if (error instanceof ApiError) {
--- a/dtfs-central-api/src/helpers/execute-with-sql-transaction.ts
+++ b/dtfs-central-api/src/helpers/execute-with-sql-transaction.ts
@@ -71,6 +71,8 @@ export const executeWithSqlTransaction = async <ReturnValue>(functionToExecute: 
 
     return result;
   } catch (error) {
+    console.error('Error thrown within SQL transaction: %s', error);
+
     await queryRunner.rollbackTransaction();
     if (error instanceof ApiError) {
       throw TransactionFailedError.forApiError(error);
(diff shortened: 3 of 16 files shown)


dtfs-central-api/src/helpers/execute-with-sql-transaction.ts
```diff
@@ -71,6 +71,8 @@ export const executeWithSqlTransaction = async <ReturnValue>(functionToExecute:
 
     return result;
   } catch (error) {
+    console.error('Error thrown within SQL transaction: %s', error);
```
sw-mattw, line 74: Picky - we seem to be using `%o` in the majority of places where we're taking this approach

```suggestion
    console.error('Error thrown within SQL transaction: %o', error);
```

dtfs-central-api/src/services/facility-utilisation-data/facility-utilisation-data.service.ts
```diff
@@ -0,0 +1,57 @@
...
+   * @param entityManager - entity manager
+   */
+  public static async initialiseFacilityUtilisationData(
+    facilityIds: Set<string>,
+    bankId: string,
+    reportPeriod: ReportPeriod,
+    requestSource: DbRequestSource,
+    entityManager: EntityManager,
+  ) {
+    const facilityIdsToInitialise = await this.filterOutFacilitiesWithExistingUtilisationData(facilityIds, entityManager);
```
sw-mattw, line 24: Nice

dtfs-central-api/src/services/facility-utilisation-data/facility-utilisation-data.service.test.ts
```diff
@@ -0,0 +1,203 @@
...
+            fixedFee: secondFixedFee,
+          }),
+        ];
+        expect(mockSave).toHaveBeenCalledWith(FacilityUtilisationDataEntity, expectedEntities, { chunk: 100 });
+      });
+    });
+  });
+
+  describe('filterOutFacilitiesWithExistingUtilisationData', () => {
+    describe('when all facility ids are associated with existing FacilityUtilisationDataEntities', () => {
```
sw-mattw, line 147: Should we test that `mockFindBy` is called correctly for these test cases?

`toHaveBeenCalledTimes`
`toHaveBeenCalledWith`

dtfs-central-api/src/services/facility-utilisation-data/facility-utilisation-data.service.ts
```diff
@@ -0,0 +1,57 @@
...
+          utilisation,
+          fixedFee,
+        }),
+      );
+    }
+
+    await entityManager.save(FacilityUtilisationDataEntity, entities, { chunk: CHUNK_SIZE_FOR_SQL_BATCH_SAVING });
+  }
+
+  public static async filterOutFacilitiesWithExistingUtilisationData(facilityIds: Set<string>, entityManager: EntityManager): Promise<Set<string>> {
```
sw-mattw, line 50: Could we get some tsdoc on this please as it's been added to the method above

sw-mattw (commented):
Minor comments, looks good thanks

sw-mattw (dismissed):
Looks good, thanks

dtfs-central-api/src/helpers/execute-with-sql-transaction.test.ts
```diff
@@ -144,6 +144,20 @@ describe('executeWithSqlTransaction', () => {
     await expect(executeWithSqlTransaction(functionToExecute)).rejects.toThrow(TransactionFailedError.forApiError(customError));
   });
 
+  it('logs error even if rolling back the transaction fails', async () => {
```
ttbarnes, line 147:
```suggestion
  it('logs error if rolling back the transaction fails', async () => {
```

dtfs-central-api/src/helpers/execute-with-sql-transaction.ts
```diff
@@ -71,6 +71,8 @@ export const executeWithSqlTransaction = async <ReturnValue>(functionToExecute:
 
     return result;
   } catch (error) {
+    console.error('Error thrown within SQL transaction: %o', error);
```
ttbarnes, line 74: Thoughts on adding a `console.info` at the top as well?

dtfs-central-api/src/services/facility-utilisation-data/facility-utilisation-data.service.ts
```diff
@@ -0,0 +1,63 @@
...
+  public static async initialiseFacilityUtilisationData(
+    facilityIds: Set<string>,
+    bankId: string,
+    reportPeriod: ReportPeriod,
+    requestSource: DbRequestSource,
+    entityManager: EntityManager,
+  ) {
+    const facilityIdsToInitialise = await this.filterOutFacilityIdsWithExistingUtilisationData(facilityIds, entityManager);
+
+    if (facilityIdsToInitialise.size === 0) {
```
ttbarnes, line 26: Can we add a comment as to why we need to return here please

dtfs-central-api/src/services/facility-utilisation-data/facility-utilisation-data.service.test.ts
```diff
@@ -0,0 +1,209 @@
...
+          }),
+          FacilityUtilisationDataEntity.create({
+            id: secondFacilityId,
+            reportPeriod: previousReportPeriod,
+            requestSource,
+            utilisation: secondUtilisation,
+            fixedFee: secondFixedFee,
+          }),
+        ];
+        expect(mockSave).toHaveBeenCalledWith(FacilityUtilisationDataEntity, expectedEntities, { chunk: 100 });
```
ttbarnes, line 141: Can we use the `CHUNK_SIZE_FOR_SQL_BATCH_SAVING` constant throughout in these tests?

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/report-uploaded/report-uploaded.event-handler.test.ts
```diff
@@ -76,68 +65,38 @@ describe('handleUtilisationReportReportUploadedEvent', () => {
...
-        utilisation,
-        fixedFee,
-      });
-    });
-
-    const expectedEntities = await Promise.all(createdFacilityUtilisationDataEntities);
-
-    expect(mockSave).toHaveBeenCalledWith(FacilityUtilisationDataEntity, expectedEntities, { chunk: 100 });
+    expect(initialiseFacilityUtilisationDataSpy).toHaveBeenCalledTimes(1);
+    expect(initialiseFacilityUtilisationDataSpy).toHaveBeenCalledWith(new Set(allFacilityIds), bankId, reportPeriod, requestSource, mockEntityManager);
```
ttbarnes, line 100: Can we put `new Set` into an `expected` const please for easier reading?

ttbarnes (changes requested):
Nice work, few minor comments otherwise LGTM

dtfs-central-api/src/helpers/execute-with-sql-transaction.ts
RLavender98: I don't think that would add much value

ttbarnes (approved):

dtfs-central-api/src/constants/batch-saving.ts
```diff
@@ -0,0 +1 @@
+export const CHUNK_SIZE_FOR_SQL_BATCH_SAVING = 100;
```
ChristianMcc, line 1: If we're making this commonised across central-api I wonder if we need specify what this chunk is saving for. One of the main reasons we do this is SQL parameter query limits which is very dependent on how large the table you are saving to (and how many columns there are). Maybe as there is one we don't need to care about this now but can imagine in catching people out in future

ChristianMcc (commented):
Nice work, reviewed this in retrospect and just a minor comment to think about in future :) very happy we've got this into release-2.6.0!

PR #3953, opened 2024-11-25
fix(FN-3631): enable deletion of payments against fees below tolerance

RLavender98:
## Introduction :pencil2:
Bug description: If a fee has reported payments less than the tolerance then once a payment was added, an error was being thrown when trying to delete the payment, since there was a check for matching that was not updated when tolerance's were introduced.

This PR fixes this by removing the check and always setting the status of a fee to TO_DO if it has no payments after a payment deletion.

- Remove incorrect check for error state that is actually a valid state in the system.


sw-mattw (approved):

ttbarnes (changes requested):
Nice work - can we add an E2E test please to make sure the functionality works as expected?

ttbarnes:
> > Nice work - can we add an E2E test please to make sure the functionality works as expected?
> 
> @ttbarnes I considered adding an e2e-test but this is an really niche edge case and the unit tests provide good coverage so I don't think it warrants it.

I think it's definitely warranted - it's a feature / user journey that a user should be able to complete? Additionally, there was a bug that prevented a user from doing this. Unit tests are great, but it doesn't cover the real world user journey

ttbarnes (approved):

PR #3966, opened 2024-11-29
feat(DTFS2-7668): added ICICI and ABC bank

abhi-markan:
## Introduction :pencil2:
Add new banks to portal.

* Added `ICICI` and `ABC` bank.

* Added `ICICI` to cspell.json.
* Added both the banks to bank mock data loader.

### DIFF
--- a/utils/mock-data-loader/banks/index.js
+++ b/utils/mock-data-loader/banks/index.js
@@ -159,6 +159,36 @@ const MOCK_BANKS = [
     ],
     isVisibleInTfmUtilisationReports: true,
   },
+  {
+    id: '12',
+    name: 'ICICI',
+    mga: ['Test.pdf'],
+    emails: ['maker1@ukexportfinance.gov.uk'],
+    companiesHouseNo: 'FC011846',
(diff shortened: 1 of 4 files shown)


ChristianMcc (approved):

utils/mock-data-loader/banks/index.js
```diff
@@ -159,6 +159,36 @@ const MOCK_BANKS = [
...
+    hasGefAccessOnly: false,
+    paymentOfficerTeam: {
+      teamName: 'Payment Reporting Team',
+      emails: ['payment-officer4@ukexportfinance.gov.uk'],
+    },
+    utilisationReportPeriodSchedule: MONTHLY_REPORT_PERIOD_SCHEDULE,
+    isVisibleInTfmUtilisationReports: true,
+  },
+  {
+    id: '12',
```
RLavender98, line 178: Utilisation reports tests are failing because we have to insert a report for each bank opted in to utilisation reporting and we make the assumption that bank ids will be unique across banks. I think that assumption is correct and this needs to be different to the id of the bank above.

PR #3968, opened 2024-11-30
feat(DTFS2-7673): create get portal amendment endpoint

MarRobSoftwire:
## Introduction :pencil2:
We need to fetch portal amendments in portal api. We want new endpoints to separate this from tfm amendments

- add get amendment controller
- add api tests
- add `FacilityAmendment` & `PortalFacilityAmendment` types & refactor existing `TfmFacilityAmendment`

- Update (and rename) `findOneAmendmentByFacilityIdAndAmendmentId` to use `findOne` rather than an aggregation pipeline & add unit tests


dtfs-central-api/src/repositories/tfm-facilities-repo/tfm-facilities.repo.ts
```diff
@@ -145,42 +145,43 @@ export class TfmFacilitiesRepo {
...
       .toArray();
   }
 
   /**
-   * Finds amendments by the facility id and amendment id
-   * @param facilityId - The facility id
-   * @param amendmentId - The amendment id
-   * @returns The found amendments
-   */
-  public static async findAmendmentsByFacilityIdAndAmendmentId(facilityId: string | ObjectId, amendmentId: string | ObjectId): Promise<Document[]> {
```
MarRobSoftwire, line 162: This method was unused

dtfs-central-api/src/repositories/tfm-facilities-repo/tfm-facilities.repo.ts
```diff
@@ -145,42 +145,43 @@ export class TfmFacilitiesRepo {
...
    * @returns The found amendment
    */
-  public static async findAmendmentByFacilityIdAndAmendmentId(facilityId: string | ObjectId, amendmentId: string | ObjectId): Promise<Document | null> {
+  public static async findOneAmendmentByFacilityIdAndAmendmentId(
+    facilityId: string | ObjectId,
+    amendmentId: string | ObjectId,
+  ): Promise<(FacilityAmendment & { ukefFacilityId: string | null }) | undefined> {
     const collection = await this.getCollection();
-    const results = await collection
-      .aggregate(aggregatePipelines.amendmentsByFacilityIdAndAmendmentId(facilityId, amendmentId))
```
MarRobSoftwire, line 179: I think its more clear to use `findOne` - since there should only be one matching facility. This also means the types are inferred automatically

dtfs-central-api/src/v1/routes/portal-routes.js
```diff
@@ -801,6 +804,42 @@ portalRouter.route('/gef/facilities').get(getGefFacilitiesController.findAllFaci
...
+ *         required: true
+ *         description: Amendment ID to get
+ *     responses:
+ *       200:
+ *         description: OK
+ *         content:
+ *           application/json:
+ *             schema:
+ *               allOf:
+ *                 - $ref: '#/definitions/PortalAmendment' // TODO add a new definition for GEF amendments
```
MarRobSoftwire, line 834: TODO - add a definition

libs/common/src/types/mongo-db-models/tfm-facilities.ts
```diff
@@ -66,12 +63,32 @@ export type TfmFacilityAmendment = {
...
+}
+
+/**
+ * Amendments created in TFM
+ *
+ * This type is likely incomplete and should be added
+ * to as and when new properties are discovered
+ */
+export interface TfmFacilityAmendment extends BaseAmendment {
+  type?: typeof AMENDMENT_TYPES.TFM;
```
MarRobSoftwire, line 88: I want to make this a required type but we'll need to migrate amendments in prod - any thoughts?

dtfs-central-api/api-tests/v1/portal-facility-amendments/amendment-get.api-test.ts
```diff
@@ -0,0 +1,135 @@
...
+        const { status, body } = await testApi.get<FacilityAmendmentResponse>(generateUrl(facilityId, aValidButNonExistentAmendmentId));
+
+        expect(status).toEqual(HttpStatusCode.NotFound);
+        expect(body).toEqual({
+          status: HttpStatusCode.NotFound,
+          message: `Amendment not found: ${aValidButNonExistentAmendmentId}`,
+        });
+      });
+
+      // TODO: DTFS2-7674 - add success tests after making a PUT request
```
MarRobSoftwire, line 132: I only realised when it came to writing these api tests that I cant check the endpoints working without another endpoint to create an entry. I think its best to get this merged in & in the PUT PR I'll update this test (since I need the types from this PR I want it merged in first)

dtfs-central-api/api-tests/test-api.ts
```diff
@@ -39,30 +39,30 @@ class TestApi {
...
     this.assertIsInitialised();
-    const to = async (url: string): Promise<Response> => await request(this.app).delete(url).send(data).set(headers);
+    const to = async <T extends Response>(url: string): Promise<T> => (await request(this.app).delete(url).send(data).set(headers)) as T;
     return { to };
   }
 
   public patch(data: object) {
     this.assertIsInitialised();
-    const to = async (url: string): Promise<Response> => await request(this.app).patch(url).send(data).set(headers);
+    const to = async <T extends Response>(url: string): Promise<T> => (await request(this.app).patch(url).send(data).set(headers)) as T;
```
ttbarnes, line 65: What's the reason for using `T extends Response` instead of just `Response` please?

dtfs-central-api/src/repositories/tfm-facilities-repo/tfm-facilities.repo.ts
```diff
@@ -145,42 +145,43 @@ export class TfmFacilitiesRepo {
    * @param status - The amendment status
    * @returns The found amendments
    */
-  public static async findAmendmentsByFacilityIdAndStatus(facilityId: string | ObjectId, status: AmendmentStatus): Promise<TfmFacilityAmendment[]> {
+  public static async findAmendmentsByFacilityIdAndStatus(facilityId: string | ObjectId, status: AmendmentStatus): Promise<FacilityAmendment[]> {
```
ttbarnes, line 148: It feels odd that in "TFM facilities repo - find one amendment ...", we use `FacilityAmendment` type instead of `TfmFacilityAmendment`, and `TfmFacilityAmendment` still exists?

dtfs-central-api/src/repositories/tfm-facilities-repo/tfm-facilities.repo.findOneAmendmentByFacilityIdAndAmendmentId.test.ts
```diff
@@ -0,0 +1,131 @@
...
+      const result = await TfmFacilitiesRepo.findOneAmendmentByFacilityIdAndAmendmentId(facilityId, amendmentId.toString());
+
+      // Assert
+      expect(result).toEqual({
+        ...amendment,
+        ukefFacilityId: facility.facilitySnapshot.ukefFacilityId,
+      });
+    });
+
+    it('returns undefined if the amendmentId does not correspond to an amendment', async () => {
```
ttbarnes, line 107:
```suggestion
    describe('when an amendment is not found', () => {
      it('should return undefined', () => {}
    });
=> {
```

dtfs-central-api/src/repositories/tfm-facilities-repo/tfm-facilities.repo.ts
```diff
@@ -145,42 +145,43 @@ export class TfmFacilitiesRepo {
...
+    });
+
+    const { ukefFacilityId } = facility.facilitySnapshot;
+
+    return (
+      amendment && {
+        ...amendment,
+        ukefFacilityId,
+      }
+    );
```
ttbarnes, line 184: This reads strange to me, could we just do i.e

```ts
return {
  ...amendment,
  ukefFacilityId,
};
```

dtfs-central-api/src/v1/controllers/portal/facility/get-amendment.controller.ts
```diff
@@ -0,0 +1,32 @@
...
+    if (!amendment || amendment.type !== AMENDMENT_TYPES.PORTAL) {
+      return res.status(HttpStatusCode.NotFound).send({ status: HttpStatusCode.NotFound, message: `Amendment not found: ${amendmentId}` });
+    }
+
+    return res.status(HttpStatusCode.Ok).send(amendment);
+  } catch (error) {
+    if (error instanceof ApiError) {
+      const { status, message, code } = error;
+      return res.status(status).send({ status, message, code });
+    }
```
ttbarnes, line 24: Can we add a new line after this please

dtfs-central-api/src/v1/routes/portal-routes.js
```diff
@@ -27,6 +29,8 @@ const getGefFacilitiesController = require('../controllers/portal/gef-facility/g
 const createGefFacilityController = require('../controllers/portal/gef-facility/create-gef-facility.controller');
 const updateGefFacilityController = require('../controllers/portal/gef-facility/update-facility.controller');
 
+const getGefPortalFacilityAmendmentController = require('../controllers/portal/facility/get-amendment.controller');
```
ttbarnes, line 32:
```suggestion
const getGefFacilityAmendmentController = require('../controllers/portal/facility/get-amendment.controller');
```

libs/common/src/types/mongo-db-models/tfm-facilities.ts
ttbarnes: I think we should enforce it, otherwise it will get messy. cc @abhi-markan

libs/common/src/types/mongo-db-models/tfm-facilities.ts
```diff
@@ -66,12 +63,32 @@ export type TfmFacilityAmendment = {
...
+  createdBy?: {
+    username: string;
+    name: string;
+    email: string;
+  };
+  leadUnderwriter?: {
+    _id: ObjectId | 'Unassigned';
+    firstName: string;
+    lastName: string;
+  };
```
ttbarnes, line 78: Isn't `leadUnderwriter` TFM specific?

ttbarnes (changes requested):

dtfs-central-api/api-tests/test-api.ts
MarRobSoftwire: I think this makes our tests cleaner, but just a preference thing & happy to reset it

allows me to write 
```
        const { status, body } = await testApi.get<FacilityAmendmentResponse>(generateUrl(facilityId, anInvalidAmendmentId));
```
rather than
```
```
        const { status, body } = await testApi.get(generateUrl(facilityId, anInvalidAmendmentId)) as FacilityAmendmentResponse;
```

I looked into this in more detail & it looks like sadly supertest doesn't support using generics without using an `as` somewhere along the way (https://github.com/ladjs/supertest/issues/720)

File: dtfs-central-api/src/repositories/tfm-facilities-repo/tfm-facilities.repo.ts (hunk 7 above)
Comment (MarRobSoftwire) [reply dtfs-central-api/src/repositories/tfm-facilities-repo/tfm-facilities.repo.ts:148]: We decided it would be best to store the portal amendments in the same place as TFM facilities (since any amalgamation into `current values` needs to include amendments from portal as well as TFM). This does mean some of these methods & endpoints will need updating, which I/Beth will do once the new endpoints are included.

I'd be happy to chat more about this because I do appreciate its a bit weird but I think makes the most sense, rather than risking creating a second source of truth in portal facilities

File: dtfs-central-api/src/repositories/tfm-facilities-repo/tfm-facilities.repo.ts (hunk 9 above)
Comment (MarRobSoftwire) [reply dtfs-central-api/src/repositories/tfm-facilities-repo/tfm-facilities.repo.ts:184]: If the amendment is undefined that returns `{ ukefFacilityId }`.

I could do this instead if you prefer

```
  if (!amendment) {
    return amendment
  }

  return {  
    ...amendment,  
    ukefFacilityId,  
  }
```

File: dtfs-central-api/src/v1/swagger-definitions/portal/portal-amendments.js
Comment (MarRobSoftwire) [inline dtfs-central-api/src/v1/swagger-definitions/portal/portal-amendments.js:1]: This was pretty painful to write & will be worse to maintain. This feels like a second source of truth & will very quickly become out of date

File: dtfs-central-api/src/repositories/tfm-facilities-repo/tfm-facilities.repo.ts (hunk 9 above)
Comment (ttbarnes) [reply dtfs-central-api/src/repositories/tfm-facilities-repo/tfm-facilities.repo.ts:184]: I think `if (!amendment) {` is clearer

Or maybe

```ts
if (amendment) {
  return {
    ...amendment,
    ukefFacilityId,
  };
}
```

dtfs-central-api/src/repositories/tfm-facilities-repo/tfm-facilities.repo.findOneAmendmentByFacilityIdAndAmendmentId.test.ts
```diff
@@ -0,0 +1,133 @@
...
+      const result = await TfmFacilitiesRepo.findOneAmendmentByFacilityIdAndAmendmentId(facilityId, amendmentId);
+
+      // Assert
+      expect(result).toEqual({
+        ...amendment,
+        ukefFacilityId: facility.facilitySnapshot.ukefFacilityId,
+      });
+    });
+
+    it('should return the correct amendment with the ukefFacilityId when the amendmentId is given as a string', async () => {
```
ttbarnes, line 74: Can we move the "when the..." into a `describe`

dtfs-central-api/api-tests/test-api.ts
ttbarnes: I think we should keep as is as we don't seem to be doing this anywhere else and want to avoid introducing new approaches/patterns.

dtfs-central-api/src/v1/controllers/portal/facility/get-amendment.controller.test.ts
```diff
@@ -0,0 +1,139 @@
...
+const amendmentId = new ObjectId().toString();
+
+describe('getAmendment', () => {
+  beforeEach(() => {
+    jest.resetAllMocks();
+
+    jest.spyOn(TfmFacilitiesRepo, 'findOneAmendmentByFacilityIdAndAmendmentId').mockImplementation(mockFindOneAmendmentByFacilityIdAndAmendmentId);
+  });
+
+  it('calls TfmFacilitiesRepo.findOneAmendmentByFacilityIdAndAmendmentId with the facility and amendment id', async () => {
```
ttbarnes, line 20:
```suggestion
  it('should call TfmFacilitiesRepo.findOneAmendmentByFacilityIdAndAmendmentId with the facility and amendment id', async () => {
```

ttbarnes: `it('should` throughout these tests please

dtfs-central-api/src/v1/controllers/portal/facility/get-amendment.controller.ts
```diff
@@ -0,0 +1,33 @@
+import { AMENDMENT_TYPES, ApiError, CustomExpressRequest } from '@ukef/dtfs2-common';
+import { HttpStatusCode } from 'axios';
+import { Response } from 'express';
+import { TfmFacilitiesRepo } from '../../../../repositories/tfm-facilities-repo';
+
+type GetAmendmentRequestParams = { facilityId: string; amendmentId: string };
+export type GetAmendmentRequest = CustomExpressRequest<{ params: GetAmendmentRequestParams }>;
+
+export const getAmendment = async (req: GetAmendmentRequest, res: Response) => {
```
ttbarnes, line 9: Documentation?

dtfs-central-api/src/v1/routes/portal-routes.js
```diff
@@ -27,6 +29,8 @@ const getGefFacilitiesController = require('../controllers/portal/gef-facility/g
 const createGefFacilityController = require('../controllers/portal/gef-facility/create-gef-facility.controller');
 const updateGefFacilityController = require('../controllers/portal/gef-facility/update-facility.controller');
 
+const getGefFacilityAmendmentController = require('../controllers/portal/facility/get-amendment.controller');
```
ttbarnes, line 32: Does this need GEF in the name, doesn't seem to be anything GEF specific? Looks like it should be `getFacilityAmendmentController`

libs/common/src/types/mongo-db-models/tfm-facilities.ts
MarRobSoftwire: I think yes but I'll raise a future ticket for this - the PR could bloat quite a lot fixing this otherwise

MarRobSoftwire: https://ukef-dtfs.atlassian.net/browse/DTFS2-7686

ttbarnes: Nice one @MarRobSoftwire

PR #4020, opened 2024-12-02
feat(FN-3578): confirmation screen data

RLavender98:
## Introduction :pencil2:
When a user has filled in the details of the record correction request we need to display their answers back to them to check before sending the email and creating the correction.

In a previous PR the template for the page and the controller were added but using hard coded data. In this PR we display actual real data on the page now that the persistence work is in part finished.

- New dtfs-central-api endpoint `GET utilisation-reports/:reportId/fee-records/:feeRecordId/correction-request-review/:userId` and corresponding tfm api endpoint to call it.
- check the information page controller calls the new endpoint to fetch the review of the record correction and uses to populate view model

## Remaining work on ticket
- I still need to add e2e-tests for the cancel and back buttons on this page. Matt has some e2e-test changes I would like to write this off of so leaving this until his PR is in.


trade-finance-manager-api/src/v1/api-response-types/fee-record-correction-request-review-response-body.ts
```diff
@@ -0,0 +1,13 @@
+import { ReportPeriod, SessionBank } from '@ukef/dtfs2-common';
+
+export type FeeRecordCorrectionRequestReviewResponseBody = {
+  bank: SessionBank;
+  reportPeriod: ReportPeriod;
+  correctionRequestDetails: {
+    facilityId: string;
+    exporter: string;
+    reasons: string[];
```
sw-mattw, line 9: Should we type this as `RecordCorrectionReason[]` from `libs/common/src/types/record-correction-reason.ts` ?

trade-finance-manager-api/src/v1/controllers/utilisation-reports/fee-record-correction/get-fee-record-correction-request-review.controller.test.ts
```diff
@@ -0,0 +1,124 @@
...
+      // Act
+      await getFeeRecordCorrectionRequestReview(req, res);
+
+      // Assert
+      expect(res._getData()).toEqual(responseBody);
+      expect(api.getFeeRecordCorrectionRequestReview).toHaveBeenCalledTimes(1);
+      expect(api.getFeeRecordCorrectionRequestReview).toHaveBeenCalledWith(reportId, feeRecordId, user);
+    });
+
+    it('should respond with a 200', async () => {
```
sw-mattw, line 58: `HttpStatusCode.Ok` please, similar below with other HTTP status codes

trade-finance-manager-ui/server/controllers/utilisation-reports/record-corrections/check-the-information/index.test.ts
```diff
@@ -22,23 +31,44 @@ describe('controllers/utilisation-reports/record-corrections/check-the-informati
...
+      };
+
+      jest.mocked(api.getFeeRecordCorrectionRequestReview).mockResolvedValue(mockApiResponse);
+
       // Act
-      getRecordCorrectionRequestInformation(req, res);
+      await getRecordCorrectionRequestInformation(req, res);
 
       // Assert
+      const expectedReasons = mapReasonsToDisplayValues(correctionRequestDetails.reasons).join(', ');
```
sw-mattw, line 56: How would you feel about pulling these expectation variables up into the arrange section?

sw-mattw (commented):
Nice, minor comments

dtfs-central-api/src/v1/controllers/utilisation-report-service/fee-record-correction/get-fee-record-correction-request-review.controller/index.test.ts
```diff
@@ -0,0 +1,208 @@
...
+    afterEach(() => {
+      jest.resetAllMocks();
+    });
+
+    it(`should respond with a ${HttpStatusCode.NotFound} when no form data is found for the fee record and user combination`, async () => {
+      // Arrange
+      const { req, res } = getHttpMocks();
+
+      findFormDataSpy.mockResolvedValue(null);
+      findFeeRecordSpy.mockResolvedValue(FeeRecordEntityMockBuilder.forReport(new UtilisationReportEntityMockBuilder().build()).build());
```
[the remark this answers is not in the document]
RLavender98: The reason to do this is to keep the test separate from the implementation details. I don't think there is a requirement on which order these are fetched and so my test shouldn't enforce that or rely upon that.

dtfs-central-api/src/v1/controllers/utilisation-report-service/fee-record-correction/get-fee-record-correction-request-review.controller/index.test.ts
```diff
@@ -0,0 +1,208 @@
...
+      expect(findFormDataSpy).toHaveBeenCalledWith(userId, feeRecordId);
+    });
+
+    it(`should respond with a ${HttpStatusCode.NotFound} when the fee record cannot be found`, async () => {
+      // Arrange
+      const { req, res } = getHttpMocks();
+
+      findFormDataSpy.mockResolvedValue(new FeeRecordCorrectionTransientFormDataEntityMockBuilder().build());
+      findFeeRecordSpy.mockResolvedValue(null);
+      findBankSpy.mockResolvedValue(aBank());
```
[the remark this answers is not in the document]
RLavender98: Same reply!

trade-finance-manager-api/src/v1/api-response-types/fee-record-correction-request-review-response-body.ts
RLavender98: Ah yes, not sure why I didn't! 😅

trade-finance-manager-api/src/v1/controllers/utilisation-reports/fee-record-correction/get-fee-record-correction-request-review.controller.test.ts
sw-mattw: @RLavender98 thanks for changing this instance, there's one below for `500` too

sw-mattw (approved):

dtfs-central-api/api-tests/v1/utilisation-reports/fee-record-corrections/get-fee-record-correction-request-review.api-test.ts
```diff
@@ -0,0 +1,159 @@
...
+import { aBank } from '../../../../test-helpers';
+import { mongoDbClient } from '../../../../src/drivers/db-client';
+import { wipe } from '../../../wipeDB';
+
+console.error = jest.fn();
+
+const BASE_URL = '/v1/utilisation-reports/:reportId/fee-records/:feeRecordId/correction-request-review/:userId';
+
+describe(`GET ${BASE_URL}`, () => {
+  const getUrl = (reportId: number | string, feeRecordId: number | string, userId: string) => {
```
Zainzzkk, line 24: could this be used anywhere else? if so it could be a test helper?

dtfs-central-api/src/v1/controllers/utilisation-report-service/fee-record-correction/get-fee-record-correction-request-review.controller/index.test.ts
```diff
@@ -0,0 +1,208 @@
...
+    const findFormDataSpy = jest.spyOn(FeeRecordCorrectionTransientFormDataRepo, 'findByUserIdAndFeeRecordId');
+    const findBankSpy = jest.spyOn(BanksRepo, 'getBankById');
+
+    afterEach(() => {
+      jest.resetAllMocks();
+    });
+
+    it(`should respond with a ${HttpStatusCode.NotFound} when no form data is found for the fee record and user combination`, async () => {
+      // Arrange
+      const { req, res } = getHttpMocks();
```
Zainzzkk, line 41: this can be called on line 34/35 so it is not repeated in every test?

dtfs-central-api/src/v1/controllers/utilisation-report-service/fee-record-correction/get-fee-record-correction-request-review.controller/index.test.ts
```diff
@@ -0,0 +1,208 @@
...
+      const bankId = '12356';
+
+      const formData = new FeeRecordCorrectionTransientFormDataEntityMockBuilder()
+        .withFormData({
+          reasons: [RECORD_CORRECTION_REASON.FACILITY_ID_INCORRECT],
+          additionalInfo: 'this is more info',
+        })
+        .build();
+      const feeRecord = FeeRecordEntityMockBuilder.forReport(new UtilisationReportEntityMockBuilder().withBankId(bankId).build())
```
Zainzzkk, line 108: new line please

e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/fee-record-correction/feature-flag-enabled/create-fee-record-correction.spec.js
```diff
@@ -88,7 +89,20 @@ context('When fee record correction feature flag is enabled', () => {
 
       cy.clickContinueButton();
 
-      cy.url().should('eq', relative(`/utilisation-reports/${reportId}/create-record-correction-request/${feeRecordAtToDoStatus.id}/check-the-information`));
+      //---------------------------------------------------------------
+      // "Check the information" page
+      //---------------------------------------------------------------
+      cy.assertText(mainHeading(), 'Check the information before submitting the record correction request');
```
Zainzzkk, line 95: can this be in a different it statement? easier to follow if something fails

trade-finance-manager-ui/server/controllers/utilisation-reports/record-corrections/check-the-information/index.test.ts
```diff
@@ -22,23 +31,44 @@ describe('controllers/utilisation-reports/record-corrections/check-the-informati
         params: { reportId, feeRecordId: feeRecordId.toString() },
       });
 
+      const bank = { name: 'Test bank', id: '129' };
```
Zainzzkk, line 34: could this to line 48 go outside the it, so it is a shorter it test then

dtfs-central-api/api-tests/v1/utilisation-reports/fee-record-corrections/get-fee-record-correction-request-review.api-test.ts
RLavender98: I don't think it makes sense to do so because it is extremely dependent on the BASE_URL which will be different for each api-test file. There might be some urls that happen to have the same parameters but then we would have to pass in BASE_URL as well and have lots of variations for each different set of parameters. I think best to leave it as is as probably not worth the time it would take to commonise.

e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/fee-record-correction/feature-flag-enabled/create-fee-record-correction.spec.js
RLavender98: I find that running cypress tests locally, the ui is really clear which step of the test failed and why so I don't think this is a concern.

sw-mattw (approved):

e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/fee-record-correction/feature-flag-enabled/create-fee-record-correction.spec.js
Zainzzkk: Can it still be split up please, long it blocks are hard to manage in the future as well.  Since the test is asserting values on a page, it should have its own block.  It won't take too much longer to run, can keep getting to the page in a beforeEach.

e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/fee-record-correction/feature-flag-enabled/create-fee-record-correction.spec.js
```diff
@@ -88,7 +89,20 @@ context('When fee record correction feature flag is enabled', () => {
...
+
+      summaryList().should('exist');
+      summaryList().should('contain', feeRecordAtToDoStatus.facilityId);
```
Zainzzkk, line 98: Can any of these use cy.assertText?

RLavender98: I can't add data-cy directly to the items so can't access them directly via a data-cy. I could get the exact element and then use cy.assertText, by using a complex html selector but those selectors would be quite brittle, so I thought it would be best to have a slightly less strict assertion, but if you would prefer it that way I can change it.

e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/fee-record-correction/feature-flag-enabled/create-fee-record-correction.spec.js
RLavender98: We don't just have to 'get to the page', we have to wipe the database and then rerun the seeding and then navigate to the page again. Then if something early in the flow fails you will have loads of failing tests not just one and they will be failing in the before each rather than even in the test which is very far from ideal.

RLavender98: What you are suggesting is actually an anti-pattern when it comes to cypress best practices: https://docs.cypress.io/app/core-concepts/best-practices#Creating-Tiny-Tests-With-A-Single-Assertion

RLavender98: > long it blocks are hard to manage in the future as well

Lots of nested describe blocks with set up split across multiple beforeEach's is harder to maintain

RLavender98: I have split this up into several smaller tests in what I think are logical splits of things that could fail. Let me know if this is better!

Zainzzkk (approved):

Zainzzkk (approved):

PR #3961, opened 2024-11-27
feat(FN-3577): persist transient record correction values (PUT)

sw-mattw:
## Introduction :pencil2:
The values entered in the "create record correction request" form need to be stored so that they are available for display and reuse later on in the journey.

We have decided to save this transient form data within the SQL database. A PUT endpoint is required to save these values to the database. Form data should be saved against a user ID and fee record ID.

## Resolution :heavy_check_mark:
- Added a new entity to store the transient form data
- Added a `PUT` endpoint for updating the transient form data
- Added payload validation to validate the form data
- Added unit tests

Note that E2E tests will follow in a subsequent PR under this ticket once the GET endpoint is implemented.

Also note that this PR is based on the validation work (PR: https://github.com/UK-Export-Finance/dtfs2/pull/3948), so this PR should not be merged until the base branch has been merged into main.

## Miscellaneous :heavy_plus_sign:
- Pulled the `MAX_RECORD_CORRECTION_ADDITIONAL_INFO_LENGTH` constant out of TFM UI into the common library as `MAX_RECORD_CORRECTION_ADDITIONAL_INFO_CHARACTER_COUNT` for use in the other services

PR #4022, opened 2024-12-03
feat(FN-3577): persist transient record correction values (GET)

sw-mattw:
## Introduction :pencil2:
The values entered in the "create record correction request" form are stored in the SQL database (PR: #3961) as for reuse later on in the "create record correction request" journey. These transient values are stored against the user ID and fee record ID.

A GET endpoint is required to retrieve these values from the database. This data should be used to populate the form fields when we return to the initial creation form.

- Added a GET endpoint for retrieving the transient form data
- Updated get controller for the "create record correction request" page to make a call to the GET endpoint and prefill the form field values
- Added unit tests
- Added e2e tests

- Updated the e2e tests to check the character limit on the additional info field to use the new char limit constant in `libs/common`


dtfs-central-api/src/v1/controllers/utilisation-report-service/fee-record-correction/get-fee-record-correction-transient-form-data.controller/index.test.ts
```diff
@@ -0,0 +1,182 @@
...
+
+    let req: GetFeeRecordCorrectionTransientFormDataRequest;
+    let res: MockResponse<Response>;
+
+    beforeEach(() => {
+      FeeRecordRepo.existsByIdAndReportId = mockFeeRecordExists;
+      mockFeeRecordExists.mockReturnValue(true);
+
+      FeeRecordCorrectionTransientFormDataRepo.findByUserIdAndFeeRecordId = mockTransientFormDataFind;
+      mockTransientFormDataFind.mockResolvedValue(null);
```
RLavender98, line 40: I'd prefer we don't have default mocks and instead mock in every test or in a describe with it's own before each that states the condition, since we are sometimes relying on these and sometimes not and it makes the test reading a little more confusing.

dtfs-central-api/src/v1/routes/utilisation-reports-routes.js
```diff
@@ -832,4 +828,53 @@ utilisationReportsRouter
...
+ *         description: the id for the user
+ *     responses:
+ *       200:
+ *         description: OK
+ *         content:
+ *           application/json:
+ *             schema:
+ *               oneOf:
+ *                 - type: object
+ *                   $ref: '#/definitions/FeeRecordCorrectionResponse'
```
RLavender98, line 865: I imagine we will soon have a get end point for an actual correction and this name might conflict so best to keep the reference to 'transient' in it please

trade-finance-manager-api/src/v1/api.js
```diff
@@ -1706,6 +1705,23 @@ const updateFeeRecordCorrectionTransientFormData = async (reportId, feeRecordId,
       user,
     },
   });
+};
+
+/**
+ * Gets the fee record by report id, fee record id and user id
```
RLavender98, line 1711: Copying error here, this gets correction form data, not a fee record

RLavender98 (commented):
Just a few minor comments!

dtfs-central-api/src/v1/controllers/utilisation-report-service/fee-record-correction/get-fee-record-correction-transient-form-data.controller/index.test.ts
sw-mattw: Fair point, moved into the individual test cases where required thanks

RLavender98 (approved):
LGTM!

RLavender98 (dismissed):

dtfs-central-api/src/v1/controllers/utilisation-report-service/fee-record-correction/get-fee-record-correction-transient-form-data.controller/index.test.ts
```diff
@@ -0,0 +1,187 @@
...
+
+    it(`should respond with a '${HttpStatusCode.Ok}' and the retrieved form data if a transient form data entity exists`, async () => {
+      // Arrange
+      const formData: RecordCorrectionTransientFormData = {
+        reasons: [RECORD_CORRECTION_REASON.OTHER],
+        additionalInfo: 'Some additional information',
+      };
+      const transientFormDataEntity = new FeeRecordCorrectionTransientFormDataEntityMockBuilder().withFormData(formData).build();
+
+      mockFeeRecordExists.mockReturnValue(true);
```
Zainzzkk, line 57: this could probably go in the beforeEach and then you only need it in the tests where it is not true

dtfs-central-api/src/v1/controllers/utilisation-report-service/fee-record-correction/get-fee-record-correction-transient-form-data.controller/index.test.ts
```diff
@@ -0,0 +1,187 @@
...
+      mockFeeRecordExists.mockRejectedValue(new Error('Some error'));
+
+      // Act
+      await getFeeRecordCorrectionTransientFormData(req, res);
+
+      // Assert
+      expect(res._getData()).toEqual(`Failed to get fee record correction transient form data`);
+    });
+
+    it('should call fee record exists once', async () => {
```
Zainzzkk, line 149: could these maybe go in a describe or have a more descriptive if? Bit confused since it's gone from error to calling fee record exists

trade-finance-manager-api/src/v1/validation/route-validators/route-validators.js
```diff
@@ -56,8 +54,6 @@ exports.partyUrnValidation = [partyURNValidation];
 
 exports.bankIdValidation = [bankIdValidation];
 
-exports.mongoIdValidation = [mongoIdValidation];
```
Zainzzkk, line 59: is this not being used anywhere else?

trade-finance-manager-ui/server/controllers/utilisation-reports/record-corrections/create-record-correction-request/index.test.ts
```diff
@@ -94,6 +103,18 @@ describe('controllers/utilisation-reports/record-corrections/create-record-corre
       expect(api.getFeeRecord).toHaveBeenCalledTimes(1);
       expect(api.getFeeRecord).toHaveBeenCalledWith(reportId, feeRecordId, userToken);
     });
+
+    it('should fetch the fee record correction transient form data using the reportId, feeRecordId, and user', async () => {
+      // Arrange
+      const { req, res } = getHttpMocks();
```
Zainzzkk, line 109: this could go above the beforeEach and can use the req and res from there?

trade-finance-manager-ui/server/controllers/utilisation-reports/record-corrections/create-record-correction-request/index.ts
```diff
@@ -32,6 +32,7 @@ export const getCreateRecordCorrectionRequest = async (req: GetCreateRecordCorre
     const { reportId, feeRecordId } = req.params;
 
     const feeRecord = await api.getFeeRecord(reportId, feeRecordId, userToken);
+    const formValues = await api.getFeeRecordCorrectionTransientFormData(reportId, feeRecordId, user, userToken);
```
Zainzzkk, line 35: do we need to check that anything has been returned before rendering?

trade-finance-manager-api/src/v1/validation/route-validators/route-validators.js
sw-mattw: Yep, no usages so removed

dtfs-central-api/src/v1/controllers/utilisation-report-service/fee-record-correction/get-fee-record-correction-transient-form-data.controller/index.test.ts
sw-mattw: This is related to one of Ruby's comments for another mock - https://github.com/UK-Export-Finance/dtfs2/pull/4022/files/ec1f46f38511503b471924a97ede4396e76c4c0b#r1867570291

I think as only 4 of the 11 tests rely on `mockFeeRecordExists` returning true, it's good to have it only in those individual test cases

trade-finance-manager-ui/server/controllers/utilisation-reports/record-corrections/create-record-correction-request/index.ts
sw-mattw: I don't believe we're doing this handling for other controllers from a quick search. The API returns either an empty object or an object with form values. If it returns something like `null` or `undefined` (for some reason) the page should handle these still and present an empty form.

I feel conflicted about adding a `!formValues` check in given we're also controlling the API it's calling. Further, we wouldn't be able to add any tests for such a check (due to the typing on the api function docs), it'd just be tested through the other functions (as not throwing).

Zainzzkk (approved):

PR #4023, opened 2024-12-03
Feat/dtfs2 7670/create amend value page

MarRobSoftwire:
## Introduction :pencil2:


gef-ui/templates/partials/amendments/facility-value.njk
```diff
@@ -0,0 +1,78 @@
...
+        {{ govukInput({
+          id: "facility-value",
+          classes: "govuk-input--width-10",
+          label: {
+            html: headingHtml,
+            isPageHeading: true
+          },
+          value: facilityValue,
+          prefix: {
+            text: "£"
```
MarRobSoftwire, line 59: This needs to modify depending on the currency chosen

gef-ui/server/controllers/amendments/facility-value/getCurrencySymbol.ts
MarRobSoftwire, line 1: I'm not sure where this helper should live (or if this kind of logic is repeated somewhere else) - it might be useful to have in libs common but I think for now makes sense to leave as local as possible

gef-ui/server/controllers/amendments/facility-value/get-facility-value.ts
```diff
@@ -0,0 +1,37 @@
...
+    const facility = await api.getFacility({ facilityId, userToken });
+
+    const currencySymbol = getCurrencySymbol(facility.details.currency?.id ?? CURRENCY.GBP);
+
+    const viewModel: FacilityValueViewModel = {
+      dealId,
+      facilityId,
+      amendmentId,
+      exporterName: deal.exporter.companyName,
+      facilityValue: 0,
```
MarRobSoftwire, line 27: These values must come from the existing amendment & should be calculated once the GET endpoint exists

gef-ui/api-tests/amendments/facility-value.api-test.ts
```diff
@@ -0,0 +1,161 @@
...
+      // Act
+      const response = await getWithSessionCookie(sessionCookie);
+
+      // Assert
+      expect(response.status).toEqual(302);
+      expect(response.headers.location).toEqual('/not-found');
+    });
+  });
+
+  describe('with portal facility amendments enabled', () => {
```
ttbarnes, line 71:
```suggestion
  describe('when portal facility amendments feature flag is enabled', () => {
```

gef-ui/server/controllers/amendments/facility-value/get-facility-value.ts
```diff
@@ -0,0 +1,46 @@
...
+export type GetFacilityValueRequest = CustomExpressRequest<{
+  params: { dealId: string; facilityId: string; amendmentId: string };
+}>;
+
+export const getFacilityValue = async (req: GetFacilityValueRequest, res: Response) => {
+  try {
+    const { dealId, facilityId, amendmentId } = req.params;
+    const { userToken, user } = asLoggedInUserSession(req.session);
+
+    const deal = await api.getApplication({ dealId, userToken });
```
ttbarnes, line 18: Can we add a unit test for this please

gef-ui/server/controllers/amendments/facility-value/getCurrencySymbol.ts
```diff
@@ -0,0 +1,16 @@
...
+    case CURRENCY.USD:
+      return '$';
+    case CURRENCY.EUR:
+      return '€';
+    case CURRENCY.GBP:
+      return '£';
+    case CURRENCY.JPY:
+      return '¥';
+    default:
+      return '';
```
ttbarnes, line 14: Can we add a test for `default` please

gef-ui/server/controllers/application-details/index.js
```diff
@@ -24,7 +24,7 @@ const getUserAuthorisationLevelsToApplication = require('../../utils/user-author
 const { FACILITY_TYPE, AUTHORISATION_LEVEL, DEAL_SUBMISSION_TYPE, STAGE } = require('../../constants');
 const Application = require('../../models/application');
 const { MAKER } = require('../../constants/roles');
-const { canUserAmendIssuedFacilities } = require('../../utils/canUserAmendIssuedFacilities');
+const { canUserAmendIssuedFacilities } = require('../../utils/facility-amendments.helper');
```
ttbarnes, line 27: Is there a unit test for this? Would be surprised if the test is passing with this change

gef-ui/server/types/facility.ts
```diff
@@ -18,7 +18,9 @@ export type Facility = {
   monthsOfCover?: number | null;
   details?: AnyObject | null;
   detailsOther?: string | null;
-  currency?: AnyObject | null;
+  currency?: {
+    id: Currency;
+  } | null;
```
ttbarnes, line 23: This reads funny to me, could/should it be `currency?: Currency;`?

gef-ui/server/types/view-models/amendments/facility-value-view-model.ts
```diff
@@ -0,0 +1,12 @@
+import { ViewModelErrors } from '../view-model-errors';
+
+export type FacilityValueViewModel = {
+  dealId: string;
+  facilityId: string;
+  amendmentId: string;
+  facilityValue: number | undefined;
```
ttbarnes, line 7:
```suggestion
  facilityValue?: number;
```

gef-ui/server/utils/facility-amendments.helper.canUserAmendFacility.test.ts
```diff
@@ -0,0 +1,104 @@
...
+    submissionType: MIN,
+    status: UKEF_ACKNOWLEDGED,
+  }) as Deal;
+
+const generateIssuedFacility = () =>
+  ({
+    hasBeenIssued: true,
+  }) as Facility;
+
+describe('canUserAmendFacility', () => {
```
ttbarnes, line 22: Should we call this `userCan...` instead of `canUser...`? `userCan` is more common throughout the codebase

gef-ui/server/utils/facility-amendments.helper.ts
```diff
@@ -23,3 +25,18 @@ export const canUserAmendIssuedFacilities = (submissionType: DealSubmissionType,
...
+ * Determines if a user can amend a facility.
+ *
+ * @param facility - The facility to check.
+ * @param deal - The deal associated with the facility.
+ * @param userRoles - The roles of the user.
+ * @returns - Returns true if the user can amend the facility, otherwise false.
+ */
+export const canUserAmendFacility = (facility: Facility, deal: Deal, userRoles: Role[]) => {
+  const userCanAmendIssuedFacilities = canUserAmendIssuedFacilities(deal.submissionType, deal.status, userRoles);
+  const isFacilityIssued = facility.hasBeenIssued === true;
```
ttbarnes, line 39:
```suggestion
  const facilityIsIssued = facility.hasBeenIssued === true;
```

gef-ui/server/utils/facility-amendments.helper.ts
```diff
@@ -23,3 +25,18 @@ export const canUserAmendIssuedFacilities = (submissionType: DealSubmissionType,
...
+
+/**
+ * Determines if a user can amend a facility.
+ *
+ * @param facility - The facility to check.
+ * @param deal - The deal associated with the facility.
+ * @param userRoles - The roles of the user.
+ * @returns - Returns true if the user can amend the facility, otherwise false.
+ */
+export const canUserAmendFacility = (facility: Facility, deal: Deal, userRoles: Role[]) => {
```
ttbarnes, line 37: Should this be `userCanAmendIssuedFacility`?

The naming feels a bit off with the below function in this context, `canUserAmendIssuedFacilities` because we're only consuming one facility here?

ttbarnes (changes requested):
Nice work, some minor comments

gef-ui/api-tests/amendments/facility-value.api-test.ts
```diff
@@ -0,0 +1,161 @@
+// eslint-disable-next-line import/no-extraneous-dependencies
```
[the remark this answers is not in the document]
MarRobSoftwire: Oops! Not sure how that made it in

gef-ui/server/controllers/amendments/facility-value/get-facility-value.ts
MarRobSoftwire: I think it already exists - gef-ui\server\controllers\amendments\facility-value\get-facility-value.test.ts is there anything else you think should be covered?

gef-ui/server/controllers/application-details/index.js
MarRobSoftwire: The unit tests exist but they use `Object.containing` so are still passing without checking for `canUserAmendIssuedFacilities`

gef-ui/server/controllers/amendments/facility-value/get-facility-value.ts
ttbarnes: I can't see a test for it - can we add one with `toHaveBeenCalledTimes(1)`, `CalledWith()` etc?

gef-ui/server/controllers/amendments/facility-value/getCurrencySymbol.ts
ttbarnes: Doesn't look like there is anything existing that does this, unless it's ambiguously named

gef-ui/server/controllers/application-details/index.js
ttbarnes: i remember, thanks

gef-ui/server/types/facility.ts
ttbarnes: Does the currency object have anything else other than id?

This is bizarre, `currency.id = { currency, amount }`

Are we sure it's correct?

MarRobSoftwire: Just checked staging and it's the same there too

MarRobSoftwire: Ahhhh I've just noticed - you've looked at the `CurrencyAndAmount` type - this is just `Currency`

ttbarnes: Ah got it, thanks @MarRobSoftwire !

gef-ui/server/controllers/amendments/facility-value/get-facility-value.test.ts
```diff
@@ -0,0 +1,158 @@
...
+
+  it('should call getFacility with the correct dealId and userToken', async () => {
+    // Arrange
+    const { req, res } = getHttpMocks();
+
+    // Act
+    await getFacilityValue(req, res);
+
+    // Assert
+    expect(getFacilityMock).toHaveBeenCalledWith({ facilityId, userToken: req.session.userToken });
```
ttbarnes, line 71: Can we add `toHaveBeenCalledTimes` to this and getApplication please

gef-ui/templates/partials/amendments/facility-value.njk
```diff
@@ -0,0 +1,84 @@
+{% extends "index.njk" %}
+{% extends "index.njk" %}
```
rkontogianni, line 2: Minor. Is the second extends a duplicate and redundant?

MarRobSoftwire: Thanks nice spot :)

gef-ui/api-tests/common-tests/role-validation-api-tests.js
```diff
@@ -22,9 +22,9 @@ const withRoleValidationApiTests = ({
   makeRequestWithHeaders,
   whitelistedRoles,
   successCode,
-  successHeaders,
+  successHeaders = undefined,
   disableHappyPath = false, // TODO DTFS2-6697: remove and test happy paths.
-  redirectUrlForInvalidRoles,
+  redirectUrlForInvalidRoles = undefined,
```
ttbarnes, line 27: Why are these undefined now?

gef-ui/component-tests/partials/amendments/facility-value.component-test.ts
```diff
@@ -0,0 +1,61 @@
...
+    wrapper.expectLink('[data-cy="cancel-link"]').toLinkTo(`/case/${dealId}/facilities/${facilityId}/amendment/${amendmentId}/cancel`, 'Cancel');
+  });
+
+  it('renders the exporter name in the heading caption', () => {
+    const wrapper = render(params);
+
+    wrapper.expectText('[data-cy="heading-caption"]').toRead(exporterName);
+  });
+
+  it('should render the curency symbol', () => {
```
ttbarnes, line 56: Can we have all test descriptions with `should render`, they're mixed and matched atm

gef-ui/server/types/view-models/amendments/facility-value-view-model.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export type FacilityValueViewModel = {
+  dealId: string;
+  facilityId: string;
+  amendmentId: string;
+  facilityValue?: number;
+  currencySymbol: string;
+  exporterName: string;
+  previousPage: string;
+  errors?: ViewModelErrors | null;
```
ttbarnes, line 11: Do we need `null` if errors is optional?

gef-ui/templates/partials/amendments/facility-value.njk
```diff
@@ -0,0 +1,83 @@
...
+{% block pageTitle -%}
+  New facility value
+{%- endblock %}
+
+{% set headingHtml %}
+  <header>
+    <span class="govuk-caption-l" data-cy="heading-caption">
+      {{ exporterName }}
+    </span>
+    <h1 class="govuk-label-wrapper">
```
Zainzzkk, line 16: data-cy for the heading?

gef-ui/api-tests/amendments/facility-value.api-test.ts
```diff
@@ -0,0 +1,160 @@
...
+      whitelistedRoles: [MAKER],
+      successCode: 200,
+    });
+
+    it('should render `New facility value` page', async () => {
+      // Act
+      const response = await getWithSessionCookie(sessionCookie);
+
+      // Assert
+      expect(response.status).toEqual(200);
```
Zainzzkk, line 86: HttpStatusCode please

gef-ui/templates/partials/amendments/facility-value.njk
```diff
@@ -0,0 +1,83 @@
...
+      <label class="govuk-label govuk-label--xl govuk-!-padding-bottom-3" for="facility-value">
```
Zainzzkk, line 17: Could this not be part of the govukInput component by using isPageHeading?

`label: {
              text: 'New facility value',
              isPageHeading: true,
              classes: "govuk-label--xl",
              attributes: {
                'data-cy': 'label-heading'
              }
            },`

Like in gef-ui\templates\partials\facilities.njk which also has a heading caption

gef-ui/server/controllers/amendments/facility-value/getCurrencySymbol.ts
Zainzzkk: it would be nice to have this in commons for futureproofing but it is fine here otherwise

gef-ui/server/controllers/amendments/facility-value/get-facility-value.ts
```diff
@@ -0,0 +1,45 @@
...
+      facilityId,
+      amendmentId,
+      exporterName: deal.exporter.companyName,
+      previousPage: `/gef/application-details/${dealId}/facility/${facilityId}/amendments/${amendmentId}/bank-review-date`,
+      currencySymbol,
+    };
+
+    return res.render('partials/amendments/facility-value.njk', viewModel);
+  } catch (error) {
+    console.error(error);
```
Zainzzkk, line 42: could you add a message too please? eg `Error rendering ... %o `

gef-ui/api-tests/common-tests/role-validation-api-tests.js
MarRobSoftwire: The successHeaders is optional & since this is consumed in typescript it needs to be marked as such - this is the neatest method I'm aware of (I have modified `redirectUrlForInvalidRoles` to take a value now)

gef-ui/server/types/view-models/amendments/facility-value-view-model.ts
MarRobSoftwire: This type exists because `validationErrorHandler` returns `ViewModelErrors | null` and is used in the post controllers. On the get controllers there's no errors & the parameter isn't passed in at all. This is widely used across `gef-ui` (and the principle is across portal and tfm-ui). I think its best to be consistent, but possibly something to look at long term

gef-ui/templates/partials/amendments/facility-value.njk
MarRobSoftwire: AFAIK the `exporterName` needs to be part of the same `header` group as the `h1` to be semantically accurate - I'm not aware of any way of including the caption otherwise

gef-ui/templates/partials/amendments/facility-value.njk
```diff
@@ -0,0 +1,83 @@
...
+
+        <div class="govuk-button-group">
+          {{ govukButton({
+            text: "Continue",
+            attributes: {
+              'data-cy': 'continue-button'
+            }
+          }) }}
+
+          <a href= {{"/case/" + dealId + "/facilities/" + facilityId + "/amendment/" + amendmentId + "/cancel"}} data-cy="cancel-link" class="govuk-link">Cancel</a>
```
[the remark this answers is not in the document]
MarRobSoftwire: It looks to be working for me

gef-ui/server/types/view-models/amendments/facility-value-view-model.ts
ttbarnes: Makes sense, thanks

gef-ui/templates/partials/amendments/facility-value.njk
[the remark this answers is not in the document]
ttbarnes: Learnt something new about nunjucks! 

I am wondering if we should change it though to be consistent, there are no other instances I can see that have such href constructions.

Equally this is cleaner.

An alternative which i'd much prefer, but would be a new pattern (want to be consistent), would be to construct the URL in the controller so the nunjucks templates just consume a variable e.g `cancelUrl`.

What do you think?

gef-ui/api-tests/common-tests/role-validation-api-tests.js
MarRobSoftwire: The problem comes from gef-ui\api-tests\amendments\facility-value.api-test.ts being a typescript file importing this function - I've added jsdoc now instead to mark these as optional & give better type safety anyway

gef-ui/templates/partials/amendments/facility-value.njk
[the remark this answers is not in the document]
MarRobSoftwire: I've gone for the latter - we're already passing in `previousPage` because the logic is quite complex, I therefore dont think its too big a break in pattern to pass in the cancelUrl too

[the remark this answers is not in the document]
ttbarnes: Nice one @MarRobSoftwire , didn't realise 👍

ttbarnes (approved):
Nice work

rkontogianni (approved):

gef-ui/server/controllers/amendments/facility-value/get-facility-value.ts
```diff
@@ -0,0 +1,43 @@
...
+    const viewModel: FacilityValueViewModel = {
+      exporterName: deal.exporter.companyName,
+      cancelUrl: `/gef/application-details/${dealId}/facility/${facilityId}/amendments/${amendmentId}/cancel`,
+      previousPage: `/gef/application-details/${dealId}/facility/${facilityId}/amendments/${amendmentId}/bank-review-date`,
+      currencySymbol,
+    };
+
+    return res.render('partials/amendments/facility-value.njk', viewModel);
+  } catch (error) {
+    console.error('Error getting amendments facility value page', error);
```
Zainzzkk, line 40: sorry small thing, need %o here

PR #4027, opened 2024-12-03
feat(DTFS2-6892): add object id schema for tfm sso

AlexBramhill:
## Introduction :pencil2:
As part of SSO, we're going to want to use the objectId schema.

Currently, it is actually a schema for `ObjectId | string` which is rather annoying to use.

Create specific schemas that transform objectId when parsing an object ie:

```
    const stringObjectId = new ObjectId().toString();
    const result = OBJECT_ID_OR_OBJECT_ID_STRING.parse(stringObjectId);
    expect(result).toEqual(stringObjectId);
```
and
```
    const objectId = new ObjectId();
    const result = OBJECT_ID_STRING.parse(objectId);
    expect(result).toEqual(objectId.toString());
```


ttbarnes (approved):
LGTM

abhi-markan (approved):

PR #4031, opened 2024-12-04
feat(FN-3575): decode html entities in additional info field

sw-mattw:
## Introduction :pencil2:
The browser encodes some special characters as HTML entities when the "create record correction request" form is submitted. This can result in the additional info input length validation failing when we are close to the limit and special character such as ampersand (&) are used (see below).

Before submitting (note presence of ampersands):
before submission

After submission (note that the ampersands have been encoded as HTML entities):
after submission

- Added a new helper to decode HTML entities in a string
- Decode HTML entities when extracting the additional info field from the request
- Additional unit and e2e tests

N/A


e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/fee-record-correction/feature-flag-enabled/create-fee-record-correction.spec.js
```diff
@@ -84,7 +85,7 @@ context('When fee record correction feature flag is enabled', () => {
 
       createFeeRecordCorrectionRequestPage.reasonCheckbox(RECORD_CORRECTION_REASON.FACILITY_ID_INCORRECT).check();
       createFeeRecordCorrectionRequestPage.reasonCheckbox(RECORD_CORRECTION_REASON.OTHER).check();
-      cy.keyboardInput(createFeeRecordCorrectionRequestPage.additionalInfoInput(), 'Some additional info.');
+      cy.keyboardInput(createFeeRecordCorrectionRequestPage.additionalInfoInput(), 'Some additional info & some more!');
```
sw-mattw, line 88: Updated so we can check these special characters aren't encoded as HTML entities later on in the journey on the "check the information" screen (e2e test additions for this screen in a different PR)

RLavender98 (dismissed):
Nice fix! Just a couple of comments naming improvement suggestions.

RLavender98 (approved):
LGTM

trade-finance-manager-ui/server/helpers/decode-html-entities.ts
```diff
@@ -0,0 +1,15 @@
...
+ * @param input - The string containing HTML entities to decode.
+ * @returns The decoded string with HTML entities converted to their
+ * corresponding characters, or undefined if input undefined.
+ */
+export const decodeHtmlEntities = (input?: string): string | undefined => {
+  if (!input) {
+    return input;
+  }
+
+  return cheerio.load(input).text();
```
abhi-markan, line 14: 1. Nice work, can we standardise this per EXIP https://github.com/UK-Export-Finance/exip/blob/main/src/api/helpers/replace-character-codes-with-characters/index.ts please thought might need to expand on the special characters please.

2. Can we add this to `libs/common` as it will be super useful for other microserivces.

sw-mattw: 1. I agree that generally we should try to standardise things, but I think the right approach here, and best practice, is to make use of a library to do this rather than trying to handroll something ourselves.
\
The implementation in this PR uses cheerio which I used as it is already in the TFM UI packages and has ~9m weekly downloads, is well maintained, and should be well tested.
\
Should we not be using a package like this that already handles all the possible HTML entities (rather than a subset) and already has its own tests for these?
\
I think the linked EXIP function should also adopt this approach of using a package longer term

2. Yep good point, will make that change thanks
\
Cheerio isn't a dependency in libs/common at present and the [html-entities package](https://www.npmjs.com/package/html-entities) (~18m weekly downloads) is more suitable for this decoding so I'd prefer to use that (but this is dependent on the above).

e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/fee-record-correction/feature-flag-enabled/create-fee-record-correction.spec.js
sw-mattw: This is the only use of the `govukCharacterCount` textarea component in the E2ETF work I believe so I don't think this would be a concern for this PR

There are some other usages in other parts of TFM UI (e.g. cases 'add a comment') that I've just tested and these have the same issue, but these are out of scope of this feature ticket and might fall under DTFS. I've raised this with Christian about opening a bug ticket that could use the helper implemented here and I'll follow up on this separate to this PR

sw-mattw: The above has been raised as a bug ticket by the DTFS team thanks - `DTFS2-7687`

trade-finance-manager-ui/server/helpers/decode-html-entities.ts
abhi-markan: Thanks @sw-mattw lets go with point 2 please using `html-entities-package`. Thanks

e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/fee-record-correction/feature-flag-enabled/create-fee-record-correction.spec.js
abhi-markan: That's great thanks @sw-mattw we will take care of the rest of the `govukCharacterCount` components as long as the helper function is in `libs/common` implementation across DTFS should be swift. Thanks

trade-finance-manager-ui/server/helpers/decode-html-entities.ts
sw-mattw: Just found this in `utils/data-migration/bss-ewcs/migrate-deals.js` thanks - I'll update that to use the libs/common function and remove the package from utils

abhi-markan (changes requested):
One minor comment.

abhi-markan (approved):
LGTM 👍🏼

PR #4033, opened 2024-12-04
feat(DTFS2-7653): scheduled deal cancellation - gef activity entries

ttbarnes:
## Introduction :pencil2:
This PR updates the "deal cancellation" functionality so that when a deal is scheduled for cancellation, the following happens:

- A new "scheduled for cancellation" activity is added to the GEF deal.
- When the deal is cancelled (via CRON job), A "deal cancelled" activity is added to the GEF deal.

Note - E2E tests will be in another PR as part of DTFS2-7656 and DTFS2-7656. Currently blocked by another PR.

- Update `PortalDealService.addGefDealCancelledActivity` to:
  - Conditionally create a `type` and `label` depending on if the cancellation is in the future.
  - Create `author._id` as an empty string, if no ID is available.
- Update `DealCancellationService.processPendingCancellation` to call `PortalDealService.addGefDealCancelledActivity`.
- Update unit tests.

- Minor unit test improvements.
- Minor logging improvements.

## Screenshot example

Screenshot 2024-12-04 at 11 28 52


dtfs-central-api/src/services/portal/deal.service.add-deal-cancelled-activity.test.ts
```diff
@@ -38,32 +42,90 @@ describe('PortalDealService - addGefDealCancelledActivity', () => {
...
   describe(`when dealType is ${DEAL_TYPE.GEF}`, () => {
-    it('should call addPortalActivity', async () => {
-      // Arrange
-      addPortalActivityMock.mockResolvedValue({ matchedCount: 1 });
-
-      jest.spyOn(PortalActivityRepo, 'addPortalActivity').mockImplementation(addPortalActivityMock);
-
-      const mockDeal = deal;
-      mockDeal.dealSnapshot.dealType = DEAL_TYPE.GEF;
+    jest.spyOn(PortalActivityRepo, 'addPortalActivity').mockImplementation(addPortalActivityMock);
```
MarRobSoftwire, line 45: should this be in a `beforeAll` or `beforeEach`?

dtfs-central-api/src/services/portal/deal.service.ts
```diff
@@ -47,28 +47,35 @@ export class PortalDealService {
...
+      const label = cancellationIsInFuture ? PORTAL_ACTIVITY_LABEL.DEAL_CANCELLATION_SCHEDULED : PORTAL_ACTIVITY_LABEL.DEAL_CANCELLED;
+
       const newActivity = {
-        type: PORTAL_ACTIVITY_TYPE.DEAL_CANCELLED,
+        type,
+        label,
         timestamp: getUnixTime(new Date()),
         author: {
-          _id: author._id,
+          _id: author._id ?? '',
```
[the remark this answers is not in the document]
MarRobSoftwire: Also should we be storing a TFM user id in portal without any explanation - this seems surprising & I think I'd be confused when looking at this.

Given the users data is stored in a TFM activity anyway (and we should have audit logs on whenever the QA is finished) I dont think we need this id here at all

libs/common/src/constants/portal/activity.ts
```diff
@@ -4,10 +4,12 @@ export const PORTAL_ACTIVITY_LABEL = {
   AIN_SUBMISSION: 'Automatic inclusion notice submitted to UKEF',
   FACILITY_CHANGED_ISSUED: 'Bank facility stage changed',
   DEAL_CANCELLED: 'Deal cancelled',
+  DEAL_CANCELLATION_SCHEDULED: 'Deal cancellation scheduled',
 } as const;
 
 export const PORTAL_ACTIVITY_TYPE = {
   DEAL_CANCELLED: 'DEAL_CANCELLED',
+  DEAL_CANCELLATION_SCHEDULED: 'DEAL_CANCELLATION_SCHEDULED',
```
MarRobSoftwire, line 12: Is `scheduled` from the designs - I thought we'd moved to using the word `Pending`

dtfs-central-api/src/services/portal/deal.service.ts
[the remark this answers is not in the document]
ttbarnes: Conversely I thought I/others could be confusing if there was only a `firstName` field and no `_id`. I envisioned that an `_id` with an empty string would be clear.

I will update this to simply not have an `_id` - as you say the TFM action/activity is in TFM anyway.

libs/common/src/constants/portal/activity.ts
ttbarnes: Good shout - I have asked design. Currently I can only see some screenshots in the ticket.

dtfs-central-api/src/services/tfm/deal-cancellation.service.processPendingCancellation.test.ts
```diff
@@ -62,19 +74,20 @@ describe('DealCancellationService', () => {
...
       reason: 'a reason',
       bankRequestDate: new Date().valueOf(),
       effectiveFrom: new Date().valueOf(),
     });
-    const auditDetails = generateTfmAuditDetails(aTfmUser()._id);
 
-    it('calls submitDealCancellation with the correct params', async () => {
-      // Arrange
-      const cancellation = aDealCancellation();
+    const cancellation = aDealCancellation();
```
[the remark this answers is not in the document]
ttbarnes: Whoops good shout, have moved into a `let` and `beforeEach`.

libs/common/src/constants/portal/activity.ts
ttbarnes: Confirmed with design, it should be "pending". Updated.

dtfs-central-api/src/services/portal/deal.service.add-deal-cancelled-activity.test.ts
```diff
@@ -38,32 +42,90 @@ describe('PortalDealService - addGefDealCancelledActivity', () => {
...
+          cancellationIsInFuture: false,
+          auditDetails,
+        });
+
+        // Assert
+        expect(addPortalActivityMock).toHaveBeenCalledTimes(1);
+
+        const expectedActivity = {
+          type: PORTAL_ACTIVITY_TYPE.DEAL_CANCELLED,
+          label: PORTAL_ACTIVITY_LABEL.DEAL_CANCELLED,
```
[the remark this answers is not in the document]
ttbarnes: Need to look into this, it is confusing - I think have this approach in other places. I might have confused with non-portal activites. Will investgiate and confirm/simplify.

dtfs-central-api/src/services/portal/deal.service.ts
```diff
@@ -47,28 +47,35 @@ export class PortalDealService {
...
    */
   public static async addGefDealCancelledActivity({
     deal,
     author,
+    cancellationIsInFuture,
     auditDetails,
   }: {
     deal: TfmDeal;
     author: ActivityAuthor;
+    cancellationIsInFuture?: boolean;
```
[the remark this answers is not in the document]
MarRobSoftwire: I think its a bit of a code smell to pass in a boolean flag, we could instead pass in `activityType`. 

It does make me wonder if this function is actually doing two things & should be split in two

dtfs-central-api/src/services/tfm/deal-cancellation.service.processPendingCancellation.test.ts
```diff
@@ -55,26 +67,33 @@ describe('DealCancellationService', () => {
...
+      bankRequestDate: 0,
+      effectiveFrom: 0,
+    };
+
+    const aDealCancellation = (): TfmDealCancellation => ({
+      ...cancellation,
+      bankRequestDate: new Date().valueOf(),
+      effectiveFrom: new Date().valueOf(),
+    });
+
```
MarRobSoftwire, line 81: This is still could be affected by side effects - if one of the functions being tested was written badly & modified `cancellation`, all subsequent tests would be affected by this

dtfs-central-api/src/services/portal/deal.service.add-deal-cancelled-activity.test.ts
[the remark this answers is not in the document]
MarRobSoftwire: It feels like the text to display could just be rendered in gef-ui using a mapper, rather than storing it in the database & effectively saving the same information twice

dtfs-central-api/src/services/portal/deal.service.ts
```diff
@@ -47,28 +47,35 @@ export class PortalDealService {
...
     auditDetails: AuditDetails;
   }): Promise<void> {
     if (deal.dealSnapshot.dealType === DEAL_TYPE.GEF) {
       const { _id: dealId } = deal.dealSnapshot;
 
+      const type = cancellationIsInFuture ? PORTAL_ACTIVITY_TYPE.DEAL_CANCELLATION_PENDING : PORTAL_ACTIVITY_TYPE.DEAL_CANCELLED;
+
+      const label = cancellationIsInFuture ? PORTAL_ACTIVITY_LABEL.DEAL_CANCELLATION_PENDING : PORTAL_ACTIVITY_LABEL.DEAL_CANCELLED;
```
Zainzzkk, line 69: since these 2 lines are the same, could `cancellationIsInFuture ? PORTAL_ACTIVITY_LABEL.DEAL_CANCELLATION_PENDING : PORTAL_ACTIVITY_LABEL.DEAL_CANCELLED;` be a function?

dtfs-central-api/src/services/tfm/deal-cancellation.service.processPendingCancellation.test.ts
```diff
@@ -83,39 +102,47 @@ describe('DealCancellationService', () => {
...
+        auditDetails,
+      });
+    });
+
+    it('should return the deal cancellation response object', async () => {
+      // Act
+      const dealCancellationResponse = await DealCancellationService.processPendingCancellation(dealId, cancellation, auditDetails);
+
+      // Assert
+      expect(dealCancellationResponse).toEqual(DealCancellationService.getTfmDealCancellationResponse(mockRepositoryResponse));
```
Zainzzkk, line 145: could `DealCancellationService.getTfmDealCancellationResponse(mockRepositoryResponse)` be const expected?

dtfs-central-api/src/services/tfm/deal-cancellation.service.ts
```diff
@@ -77,6 +80,13 @@ export class DealCancellationService {
         dealType,
       });
 
+      await PortalDealService.addGefDealCancelledActivity({
```
Zainzzkk, line 83: should any of these be in a try catch in this file since they are async calls

dtfs-central-api/src/services/tfm/deal-cancellation.service.processPendingCancellation.test.ts
ttbarnes: I don't think we need to change this for a potentially badly written test in the future.

dtfs-central-api/src/services/portal/deal.service.add-deal-cancelled-activity.test.ts
```diff
@@ -38,32 +42,90 @@ describe('PortalDealService - addGefDealCancelledActivity', () => {
...
+          deal: mockDeal,
+          author,
+          cancellationIsInFuture: true,
+          auditDetails,
+        });
+
+        // Assert
+        expect(addPortalActivityMock).toHaveBeenCalledTimes(1);
+
+        const expectedActivity = {
```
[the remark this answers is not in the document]
ttbarnes: I think the opposite because `expectedActivity` is used after line 88 only?

dtfs-central-api/src/services/portal/deal.service.ts
ttbarnes: They're not the same, one uses `PORTAL_ACTIVITY_TYPE`, the other uses `PORTAL_ACTIVITY_LABEL`.

dtfs-central-api/src/services/tfm/deal-cancellation.service.ts
ttbarnes: None of the existing services have try/catch handlers - would like them to be aligned.

I imagine this decision was made so that errors bubble up.

@MarRobSoftwire ?

dtfs-central-api/src/services/portal/deal.service.ts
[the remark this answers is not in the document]
ttbarnes: I don't agree, boolean is a native type, very common?

Using `activityType` instead won't work - we need to know if the cancellation is before or after "now" to set different activity values.

dtfs-central-api/src/services/portal/deal.service.add-deal-cancelled-activity.test.ts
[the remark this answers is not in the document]
ttbarnes: I have removed `type` so we use only `label` now.

I think mapper to render text is overkill, instead we can simply have 1x property saved in the DB that can be consumed.

dtfs-central-api/src/services/portal/deal.service.ts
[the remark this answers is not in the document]
MarRobSoftwire: My understanding is _Clean code_ encourages generally avoiding boolean arguments where possible because its a sign the function is doing more than one thing. It's not too big of an issue here but I think it'd be cleaner to either:
- create a second function `addGefCancellationPendingActivity`
- pass in a `label` parameter taking value `DEAL_CANCELLATION_PENDING` or `DEAL_CANCELLED` accordingly

dtfs-central-api/src/services/tfm/deal-cancellation.service.processPendingCancellation.test.ts
MarRobSoftwire: Sorry I dont think I was clear, this might not work if `DealCancellationService.processPendingCancellation` is modifying the paramenters, not if future tests are written badly

e.g. if the method  did things like
```
  public static async processPendingCancellation(
    dealId: ObjectId | string,
    cancellation: TfmDealCancellation,
    auditDetails: AuditDetails,
  ): Promise<TfmDealCancellationResponse> {

    // Does some stuff modifying the object
    cancellation.reason = mapReason(cancellation.reason);

    if (!cancellation.auditRecord) {
      cancellation.auditRecord = generateAuditRecordFromAuditDetails(auditDetails)
    }
  }

```

Then we'd be passing in the modified `reason` & `auditRecord` fields in on each subsequent test giving unexpected behaviour.

dtfs-central-api/src/services/tfm/deal-cancellation.service.ts
MarRobSoftwire: Yep thats the idea, because we want errors thrown in the repositories/services to be handled at the controller level we dont want a try/catch here.

If we were going to add one we'd need to write

```
catch (error) {
  throw error
}
```
which es-lint prevents you from doing anyway

dtfs-central-api/src/services/portal/deal.service.add-deal-cancelled-activity.test.ts
```diff
@@ -38,43 +42,105 @@ describe('PortalDealService - addGefDealCancelledActivity', () => {
...
-      addPortalActivityMock.mockResolvedValue({ matchedCount: 1 });
-
+    beforeEach(() => {
       jest.spyOn(PortalActivityRepo, 'addPortalActivity').mockImplementation(addPortalActivityMock);
+    });
 
-      const mockDeal = deal;
-      mockDeal.dealSnapshot.dealType = DEAL_TYPE.GEF;
+    const mockDeal = deal;
+    mockDeal.dealSnapshot.dealType = DEAL_TYPE.GEF;
```
MarRobSoftwire, line 50: This is modifying `deal` since `const mockDeal = deal;` copies by reference.
As I commented elsewhere, we probably want to re-instantiate this in every test to avoid the mock deal object beign modified

libs/common/src/types/portal/activity.ts
```diff
@@ -14,7 +14,7 @@ export type PortalActivityLabel = ValuesOf<typeof PORTAL_ACTIVITY_LABEL>;
  * Ideally, we would run some data migration to rename one or both of these fields.
  */
 export type PortalActivity = {
-  type: PortalActivityType;
+  type?: PortalActivityType;
```
MarRobSoftwire, line 17: This feels inconsistent to sometimes not have a `type` - whilst I agree it doesnt feel necessary for deal cancellation we should probably keep to the current pattern & keep this mandatory

dtfs-central-api/src/services/portal/deal.service.ts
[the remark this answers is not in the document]
ttbarnes: Thanks @MarRobSoftwire I don't entirely agree that a function shouldn't do more one thing, but do agree a `addGefCancellationPendingActivity` method would be cleaner, will update.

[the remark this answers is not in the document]
ttbarnes: @MarRobSoftwire I've improved this so we now have the following:

```ts
addGefDealCancellationPendingActivity
addGefDealCancelledActivity
```

And they both accept `dealId` and `dealType` instead of having to consume the whole deal object.

dtfs-central-api/src/services/portal/deal.service.add-deal-cancelled-activity.test.ts
ttbarnes: This is now outdated, deal object is no longer required.

dtfs-central-api/src/services/portal/deal.service.ts
ttbarnes: This is now outdated - resolving

dtfs-central-api/src/services/tfm/deal-cancellation.service.processPendingCancellation.test.ts
ttbarnes: Have updated and simplified the creation of the `cancellation` object

libs/common/src/types/portal/activity.ts
ttbarnes: I'm on both sides of the fence here:

- If we keep it in, we stick to the existing pattern, but we have additional, useless data being created along with unnecessary extra lines of code. This could be confusing I think we've all had a fair amount of confusion RE activities.
- If we make `type` optional and do not add `type` to all activities moving forward, it's misaligned to other activites, but we have one less confusing thing in the code and database.

Ideally we could remove `type` entirely and run data migration for any existing deals.

Also, It seems rare that new activities will be added to the GEF timeline - in which case I imagine/hope the "deal cancelled" activity would be the most recent example to be found

Thoughts?

@Zainzzkk What do you think?

MarRobSoftwire (approved):

Zainzzkk (approved):

libs/common/src/types/portal/activity.ts
ttbarnes: I really don't think we should enforce `type` here. In other comments/PRs, people have been confused about the difference between `type` and `label` - moving forward, we don't need both, the only reason for keeping both would be to keep in aligned with legacy activities.

I think having it optional has more benefit than enforcing it.

We shouldn't have every property in `PortalActivityType` as optional - maybe as well use `AnyObject` in that case.

ttbarnes: PR is approved so marking this as resolved. Happy to revert if there are any strong objections

PR #4034, opened 2024-12-04
feat(DTFS2-6892): add schema tests for commonised tfm user

AlexBramhill:
## Introduction :pencil2:
We need to add new tests for tfm user

The existing needed to be improved to be more scalable, this has been refactored to allow ease of use and time savings

It creates a series of chainable `withSomethingTests` that can be called easily via `withSchemaValidationTests`. It then automatically creates all required tests based on the test cases, and the supplied test case `type`.

It is easily extendable (all existing patterns are established), and easy to use to write tests.

```
  withSchemaValidationTests({
    schema: DECODED_AUTH_CODE_REQUEST_STATE_SCHEMA,
    aValidPayload,
    testCases: [
      {
        parameterPath: 'csrfToken',
        type: 'string',
      },
      {
        parameterPath: 'successRedirect',
        type: 'string',
        options: { isOptional: true },
      },
    ],
  });

```


libs/common/src/schemas/object-id.ts
rkontogianni, line 1: This file I think could rename to 'libs/common/src/scemas/object-id.schema.ts'

libs/common/src/schemas/portal-user.ts
rkontogianni, line 1: this could rename to portal-user.schema.ts

libs/common/src/schemas/portal-amendment.ts
```diff
@@ -14,20 +14,12 @@ export const PORTAL_FACILITY_AMENDMENT_USER_VALUES = z
     coverEndDate: UNIX_TIMESTAMP_SECONDS_SCHEMA.optional(),
```
rkontogianni, line 14: PORTAL_FACILITY_AMENDMENT_USER_VALUES to PORTAL_FACILITY_AMENDMENT_USER_SCHEMA

libs/common/src/schemas/portal-amendment.ts
rkontogianni, line 1: this file could rename to portal-amendment.schema.ts

libs/common/src/schemas/object-id.ts
AlexBramhill: great spot!

libs/common/src/schemas/portal-amendment.ts
AlexBramhill: This is a great comment, but beyond the scope of this commonisation

trade-finance-manager-api/src/v1/controllers/user/user.routes.test.js
```diff
@@ -113,3 +112,113 @@ describe('user routes', () => {
     }
   });
 });
+
+function getEntraIdUserFailureTestCases({ getTestObjectWithUpdatedEntraIdUserParams = (entraIdUser) => entraIdUser }) {
```
AlexBramhill, line 116: Theres no need for our unit tests to have so much testing of validation and would happily reduce this down if okay'd -- We test the schemas extensively, so here we should only have to test one bad and one good route in my opinion

rkontogianni: @AlexBramhill which is the difference with that one from libs/common ?

AlexBramhill: Good question -- I'm going to give a long answer just to provide as much information as possible. I'll note if things are opinionated -- as its always good to understand what is 'right and wrong' and whats just my formed opinions which can be disagreed with.

tldr:
We've removed this from libs common and replaced with an easier syntax for testing schemas -- this is because they were incredibly time consuming and repetitive to write. In reality, we don't need all of these, but since we have them I've kept them around

long:

We currently have a variety of concepts that come to a head in validation middleware tests.

These are:
- Schemas 
- Validation middleware creation 
- Calling validation in endpoints
- Approach to testing

# Schemas
- We currently test schemas -- In my opinion, we shouldn't need to do this, as we're relying on a third party library for validation on this, and we should trust it to do it's job. If we don't -- we shouldn't use it.
- The existing method for testing schemas is incredibly time consuming to write, error prone, and hard to maintain.
- This PR updates this with a hand rolled approach which will speed up 95%+ of test writing for schemas, while also making the remaining 5% no longer to write than existing (if that!).

# Validation
- Our schemas are used for validation. This is either through a curried middleware function (`createValidationMiddlewareForSchema`) or calling a validate function, or simply doing `schema.parse(objectToCheck)`.
- In all of the above, we are already certain that the schema works correctly (and, even if it wasn't, we do currently have tests for it which I'm not sure are needed).
- In the above where there is an extra function (middleware or separate validate function), we have tests to ensure these functions create the middleware or call the `schema.parse` and handle it as expected
- We're therefore sure at this stage that we have a working schema and working validation

# Adding to an endpoint
- We then add the validation to an endpoint, and we know this validation works and is extensively tested
- Opinion: At this stage, we should not look to test validation again -- as it's already been tested multiple times completely and holistically (opinion: we probably test these far too extensively -- I'd probably look at just testing the validation middlewares call `.parse` on schemas, and handles the results as expected as we should be able to be confident that `zod` behaves how it should)
- Here, in this specific file, we had tests to test the validation -- which i think is overkill to have so many. What we need to test is that `given validation fails, x response is sent` as that's what the unit of this test is handling in terms of logic
- However, due to our existing testing strategy, we do not mock async calls, meaning we cannot just mock the pretested validation to fail, and we have to supply some properly failing test cases (which isn't the end of the world). In practice, I'd be happy only testing a handful, however as we already have all these, I've decided to keep them in.

rkontogianni: @AlexBramhill Thank you for your comment. It's very informative. When I left this comment though I hadn't noticed that has been removed from libs/common?

libs/common/src/schemas/iso-date-time-stamp.schema.ts
```diff
@@ -4,4 +4,4 @@ import z from 'zod';
  * The following regex complies with whatgetNowAsUtcISOString returns
```
rkontogianni, line 4: @AlexBramhill
```suggestion
 * mixed implementation
```

libs/common/src/schemas/audit-database-record.test.ts
```diff
@@ -1,54 +1,10 @@
-import { ObjectId } from 'mongodb';
-import {
-  generateNoUserLoggedInAuditDatabaseRecord,
-  generatePortalUserAuditDatabaseRecord,
-  generateSystemAuditDatabaseRecord,
-  generateTfmUserAuditDatabaseRecord,
-} from '../change-stream';
-import { AUDIT_DATABASE_RECORD } from './audit-database-record';
-import { withSchemaTests } from '../test-helpers';
+import { withAuditDatabaseRecordSchemaTests } from '../test-helpers/schemas/schema-tests/with-audit-database-record-schema.tests';
```
rkontogianni, line 1: @AlexBramhill I think you can rename this file to audit-database-record.schema.test.ts to be consistent with the other test files

libs/common/src/test-helpers/schemas/custom-types-tests/index.ts
```diff
@@ -0,0 +1,12 @@
...
+ * ie these schemas represent a single field, not a whole object
+ */
+export * from './with-iso-date-time-stamp-schema.tests';
+export * from './with-object-id-or-object-id-string-schema.tests';
+export * from './with-object-id-schema.tests';
+export * from './with-object-id-string-schema.tests';
+export * from './with-unix-timestamp-schema.tests';
+export * from './with-unix-timestamp-milliseconds-schema.tests';
+export * from './with-unix-timestamp-seconds-schema.tests';
+export * from './with-tfm-team-schema.tests';
```
rkontogianni, line 12: @AlexBramhill Should not be here with-currency-schema.test.ts ?

libs/common/src/schemas/audit-database-record.test.ts
AlexBramhill: Great catch!

libs/common/src/schemas/iso-date-time-stamp.schema.ts
AlexBramhill: Thank you!

libs/common/src/test-helpers/schemas/custom-types-tests/with-unix-timestamp-schema.tests.ts
```diff
@@ -0,0 +1,37 @@
...
+      getUpdatedParameterFromParsedTestObject,
+    });
+
+    it('should fail parsing if the parameter is not positive number', () => {
+      const { success } = schema.safeParse(getTestObjectWithUpdatedParameter(-1));
+      expect(success).toBe(false);
+    });
+
+    it('should fail parsing if the parameter is not an int number', () => {
+      const { success } = schema.safeParse(getTestObjectWithUpdatedParameter(-1));
```
rkontogianni, line 33: @AlexBramhill I think you should have a not int number here rather than -1

libs/common/src/test-helpers/schemas/custom-types-tests/index.ts
AlexBramhill: great spot!

libs/common/src/schemas/portal-user.schma.create.test.ts
```diff
@@ -0,0 +1,92 @@
+import { ObjectId } from 'mongodb';
```
rkontogianni, line 1: @AlexBramhill there's a misspelling in the file name

libs/common/src/schemas/portal-amendment.schema.test.ts
```diff
@@ -0,0 +1,65 @@
+import { PORTAL_FACILITY_AMENDMENT_USER_VALUES } from './portal-amendment.schema';
```
rkontogianni, line 1: @AlexBramhill same here some properties needs to be removed as we no longer using them into portal-amendments schema

trade-finance-manager-api/src/v1/middleware/validate-put-tfm-user-payload.test.ts
```diff
@@ -1,8 +1,64 @@
...
+    validateTfmPutUserPayload(req, res, next);
+
+    // Assert
+    expect(res._getStatusCode()).toEqual(HttpStatusCode.BadRequest);
+    expect(res._isEndCalled()).toEqual(true);
+    expect(next).not.toHaveBeenCalled();
+  });
+
+  it("calls the 'next' function if the payload is valid", () => {
+    // Arrange
```
rkontogianni, line 51: it("should call the 'next' function)

trade-finance-manager-api/src/v1/middleware/validate-put-tfm-user-payload.test.ts
```diff
@@ -1,8 +1,64 @@
...
+      },
+    },
+    {
+      description: 'the payload is an empty object',
+      payload: {},
+    },
+  ];
+
+  it.each(invalidPayloads)(`responds with a '${HttpStatusCode.BadRequest}' if $description`, ({ payload }) => {
+    // Arrange
```
rkontogianni, line 35: @AlexBramhill  it.each(invalidPayloads)(`should respond with a

libs/common/src/schemas/portal-amendment.schema.test.ts
AlexBramhill: sigh, added. At least these tests make it easier to add things like this!

rkontogianni (dismissed):

rkontogianni (approved):

libs/common/src/test-helpers/schemas/with-schema-validation.tests.ts
```diff
@@ -0,0 +1,129 @@
...
+import { WithTestsForTestCaseProps } from './types/with-tests-for-test-case';
+import { BaseTestCase } from './test-cases/base-test-case';
+
+/**
+ * This function orchestrates a schema's test cases.
+ * It applies schema test options to all test cases, as well as adding and schema-specific tests as required.
+ * @param params.schema The schema to test
+ * @param params.schemaTestOptions Options that are specific to the schema as a whole, for instance, if the schema is a partial, or strict
+ * @param params.aValidPayload A function that returns a valid payload for the schema
+ * @param params.testCases Test cases to test
```
rkontogianni, line 15: @AlexBramhill I think missing the param withTestsForTestCases here?

rkontogianni (approved):

PR #4042, opened 2024-12-06
feat(DTFS2-6892): update test api error constructor

AlexBramhill:
## Introduction :pencil2:
Our test api error currently is missing a few fields needed for sso tests

Add these missing fields, update constructor to take object

### DIFF
--- a/libs/common/src/test-helpers/test-api-error.ts
+++ b/libs/common/src/test-helpers/test-api-error.ts
@@ -1,8 +1,18 @@
 import { HttpStatusCode } from 'axios';
 import { ApiError } from '../errors';
+import { ApiErrorCode } from '../types';
+
+type TestApiErrorParams = {
+  status?: HttpStatusCode;
+  message?: string;
+  code?: ApiErrorCode;
+  cause?: unknown;
+};
 
 export class TestApiError extends ApiError {
-  constructor(status?: number, message?: string) {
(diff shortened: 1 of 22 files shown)


sw-mattw (approved):
Looks good

libs/common/src/test-helpers/test-api-error.ts
```diff
@@ -1,8 +1,18 @@
...
+  code?: ApiErrorCode;
+  cause?: unknown;
+};
 
 export class TestApiError extends ApiError {
-  constructor(status?: number, message?: string) {
-    super({ status: status ?? HttpStatusCode.InternalServerError, message: message ?? '' });
+  constructor(testApiErrorParams: TestApiErrorParams = {}) {
+    const defaults = { status: HttpStatusCode.InternalServerError, message: '' };
+    const apiErrorConstructorParams = { ...defaults, ...testApiErrorParams };
```
sw-mattw, line 15: My only comment would be that we can pass explicitly undefined values through which would then replace the defaults (vs the coalescing operator above)

AlexBramhill: I think this should be okay -- If we're explicitly overriding this now, it's because we want to explicitly override I'd think

ttbarnes (approved):

PR #3594, opened 2024-09-30
chore(DTFS2-7431): createBssEwcsDeal cypress command - individual deals

ttbarnes:
## Introduction :pencil2:
- BSS/EWCS E2E tests were creating deals by calling the API directly.
- We want to prevent specific user roles (like maker, checker), from calling Portal API directly.
- This PR updates all E2E tests that create individual BSS/EWCS deals, to be created via the UI, mimicking a true user journey.

## Resolution :heavy_check_mark:
- Create `createBssEwcsDeal` cypress command.
- Update all BSS/EWCS E2E tests that create single deals.

## Miscellaneous :heavy_plus_sign:
List any additional fixes or improvements.

PR #4043, opened 2024-12-06
chore(DTFS2-7688-7577): increase GEF axios timeout and add sleep to test.yml

Zainzzkk:
## Introduction :pencil2:
This PR adds a longer timeout to GEF axios calls and adds sleeps between mock data loader and database migration for GHA

* Changed timeout to 60s in `gef-ui/server/services/axios.js`
* Added 60s wait after `npm run load` and after `npm run db:migrate` in `test.yml`
* Changed Initialise step to also do db migration


portal-api/src/v1/api.js
```diff
@@ -245,6 +245,7 @@ const tfmDealSubmit = async (dealId, dealType, checker) => {
         dealType,
         checker,
       },
+      timeout: 100000,
```
abhi-markan, line 248: Is this 100s, what is the default value 60s?

abhi-markan (changes requested):

e2e-tests/ukef/cypress/e2e/journeys/portal/deal-cancellation/BSS-EWCS/MIN-future-cancellation-updates-portal.spec.js
```diff
@@ -13,16 +13,24 @@ describe('Deal Cancellation status updates', () => {
...
+        cy.createFacilities(dealId, mockFacilities, BANK1_MAKER1)
+          .then((createdFacilities) => {
+            dealFacilities.push(...createdFacilities);
+          })
+          .then(() => {
+            cy.wrap(dealFacilities).should('not.be.empty');
+          });
+      })
+      .then(() => {
+        cy.wrap(deal).should('not.be.empty');
```
ttbarnes, line 32: Do we need to make this change, can we revert? This will make https://github.com/UK-Export-Finance/dtfs2/pull/3594 harder to merge which completely removes the need to use `cy.insertOneDeal`

Zainzzkk: Just testing out

PR #4054, opened 2024-12-10
fix(DTFS2-7694): argument as an object

abhi-markan:
## Introduction :pencil2:
Facility issuance and amendments are failing due to invalid property name and argument type being supplied to DAF pertinent to extracting facility master record and facility loan identifier.

* Corrected durable activity function argument as an object.
* Removed object destructuring aliasing, since `facilityIdentifier` is the supplied property name.


ChristianMcc (approved):
LGTM, nice one

PR #4064, opened 2024-12-11
feat(FN-3582): add valid emails for PDC users

RLavender98:
## Introduction :pencil2:
Sending a record correction sends an email to the requesting user. With the email `test@testing.com` sending the email was throwing an error since the email is invalid which makes it impossible to send a record correction request. This PR prevents devs encountering this problem by setting the PDC email addresses to valid payment officer email addresses.

- Switch invalid PDC user emails for payment officer email addresses


utils/mock-data-loader/tfm/mocks/users.js
```diff
@@ -193,7 +193,7 @@ module.exports = {
   PIM_PDC_RECONCILE: {
     username: 'PIM_PDC_RECONCILE',
     password: 'AbC!2345',
-    email: 'test@testing.com',
+    email: 'payment-officer2@ukexportfinance.gov.uk',
```
ttbarnes, line 196: I don't think these emails should be in the mock data loader - all other emails are "test" emails

Seems there could be some risk with this and interrupting QA

RLavender98: These emails are already used in the banks file so if that is a risk I think it is likely already happening. 

If not these emails what would you suggest we use instead?

RLavender98: Our portal users and bank emails all have valid email addresses in mock data loader and all the emails so far for E2ETF have gone to portal users or bank emails or emails stored in env vars. Recently I merged the first work that users TFM users email to send an email.

Since I wanted to receive the email locally myself when doing the dev, I changed the email temporarily to my own email so I would receive it. 

What I didn't notice was that these tfm emails are so invalid that you can't go through the flow without having to manually come and change the email, since if the email fails to send, it is a requirement of the original email ticket that we render the problem with service page. 

This is annoying as would mean you would have to remember to come and change this every time you ran mock-data-loader which is a lot of the time, even if not working on the actual email as you wouldn't be able to progress with the user journey otherwise.

Now that I have noticed that we even have these emails against portal users in mock-data-loader, I do think it should be fine to include these emails here as is.

Zainzzkk (approved):

utils/mock-data-loader/tfm/mocks/users.js
ttbarnes: Makes sense thanks @RLavender98

ttbarnes (approved):

Zainzzkk (commented):
Will this be going in soon? after main is merged into the branch

PR #4065, opened 2024-12-11
feat(FN-3578): display line breaks in additional info

RLavender98:
## Introduction :pencil2:
When the user enters line breaks in the additional info for a record correction request these should be displayed to the user on the check the information page.

- Replace new lines with <br> tags in additional info and pass to the template as html so that the line breaks are displayed.

- Refactor e2e-tests into page focussed groupings and files to reduce nesting whilst minimising repetition
- Add e2e-test for clicking cancel button on check the information page


trade-finance-manager-ui/templates/utilisation-reports/record-corrections/check-the-information.njk
```diff
@@ -107,7 +107,7 @@
         href: cancelLink,
         classes: "govuk-button--secondary",
         attributes: {
-          'data-cy': 'cancel-button'
+          'data-cy': 'cancel-link'
```
Zainzzkk, line 110: is this a button or a link on the page? given the class is still govuk-button?

trade-finance-manager-ui/templates/utilisation-reports/record-corrections/check-the-information.njk
```diff
@@ -63,7 +63,7 @@
           text: "Provide more information"
         },
         value: {
-          text: additionalInfo
+          html: (additionalInfo | replace('\n', '<br>'))
```
Zainzzkk, line 66: could this come from the UI/helper so we don't do logic in nunjucks files?

e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/fee-record-correction/feature-flag-enabled/create-fee-record-correction-form.spec.js
```diff
@@ -0,0 +1,146 @@
...
+    });
+
+    it('should let the user enter additional info equal to the character limit containing special characters', () => {
+      premiumPaymentsTab.premiumPaymentsTable.checkbox([feeRecordAtToDoStatus.id], feeRecordAtToDoStatus.paymentCurrency, feeRecordAtToDoStatus.status).click();
+
+      premiumPaymentsTab.createRecordCorrectionRequestButton().click();
+
+      createFeeRecordCorrectionRequestPage.reasonCheckbox(RECORD_CORRECTION_REASON.OTHER).check();
+
+      const specialCharactersToTest = '&!?$£¥€¢^*()_+=-%:;@~/><,.';
```
Zainzzkk, line 136: is it possible to check that these characters are displayed when coming back to the page.
If handling for special characters breaks for this page, we will be able to catch it in a test

trade-finance-manager-ui/templates/utilisation-reports/record-corrections/check-the-information.njk
sw-mattw: I think we should keep this as `button` - I think `link` makes sense for the other pages whose 'Cancel' behaviour is through a hyperlink

RLavender98: I named it this just to use the common cypress selector, have seen Matt has added a new one for cancel button so will revert to that change when I merge in main.

e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/fee-record-correction/feature-flag-enabled/create-fee-record-correction-form.spec.js
RLavender98: I will add some special characters to the additional info in the check the info page tests

trade-finance-manager-ui/templates/utilisation-reports/record-corrections/check-the-information.njk
RLavender98: 👍 Have made it a custom nunjucks filter so that it can be reused, following the pattern of pre-existing replace filters.

sw-mattw (approved):
Nice work, thanks for reorganising the e2e tests

Zainzzkk (approved):

PR #4072, opened 2024-12-11
fix(DTFS2-7694): facility issuance reconciliation with TFM

abhi-markan:
## Introduction :pencil2:
When a facility is issued payload response received from ACBS is not being reconciliated back to `tfm-facilities` acbs object.

* Updated to `facilityIdentifier` as similar variables were updated in `abcs` microservice.

## Miscellaneous ➕ 
* Added unit test cases.
* Lint fixes.
* Documentation improvement.


trade-finance-manager-api/src/v1/controllers/acbs.controller.js
```diff
@@ -140,8 +140,8 @@ const updateDealAcbs = async (taskOutput) => {
  * @param {Object} FFR Facility fixed fee record
  * @returns {Object} ACBS returned response
  */
-const updateIssuedFacilityAcbs = ({ facilityId, issuedFacilityMaster, facilityLoan, facilityFee }) =>
-  tfmController.updateFacilityAcbs(facilityId, {
+const updateIssuedFacilityAcbs = ({ facilityIdentifier, issuedFacilityMaster, facilityLoan, facilityFee }) =>
```
Zainzzkk, line 143: do we need to update a test?

abhi-markan: Have added.

trade-finance-manager-api/src/v1/controllers/acbs.controller.test.js
```diff
@@ -39,8 +49,117 @@ const mockACBSTaskLink = {
   purgeHistoryDeleteUri: 'acbs',
 };
 
+const mockAcbsFacilityIssueTask = {
```
Zainzzkk, line 52: could this be a test mock?

abhi-markan: Added.

ttbarnes (approved):

PR #4077, opened 2024-12-13
chore(DTFS2-7708): updated dependabot.yml

Zainzzkk:
## Introduction :pencil2:
This PR updates the dependabot.yml file

* Grouped PRs for each directory in dependabot.yml


.github/dependabot.yml
```diff
@@ -16,6 +16,10 @@ updates:
       time: '00:00'
     labels:
       - 'chore'
+    groups:
```
abhi-markan, line 19: Can these grouped same as DockerFile?

abhi-markan (changes requested):

.github/dependabot.yml
Zainzzkk: You want them all in 1 PR?

abhi-markan (approved):

PR #4094, opened 2024-12-20
feat(FN-3581): fix "request sent" email line break

sw-mattw:
## Introduction :pencil2:
The email address on the TFM UI record correction request sent screen is wrapping over multiple lines. We want to prevent this happening. This should only be a UKEF email, so we can make fair assumptions on the length.

- Add new CSS class for collapsing any white space (naming in line with existing class for preserving white space)
- Applied this new class to the relevant copy
- Updated component test to ensure class is applied to copy

N/A

**Screenshots**
Before fix:
FN-3581_PR-2_record-correction-req-line-break-fix_BEFORE

After fix:
FN-3581_PR-2_record-correction-req-line-break-fix


trade-finance-manager-ui/public/css/styles.css
sw-mattw, line 1: This file has just been regenerated with `npm run build` due to the CSS changes

Zainzzkk: do any of the integrities need to be updated?

trade-finance-manager-ui/styles/typography.scss
Zainzzkk, line 1: what will happen if someone looks at this screen on mobile? will it wrap on a smaller screen?

sw-mattw: The new approach using a new full width row addresses this - it will now break to avoid horizontal scrolling 'off the page'. This is in line with the requirement that "copy of email sent has to be fixed into a single line".

**Large**

large

**Medium**

medium

**Small**

small

sw-mattw: Avoiding this overflow behaviour (reasonable length name for example):

overflow

trade-finance-manager-ui/public/css/styles.css
sw-mattw: I think it's just our `script` tags that need integrity hashes - so should be okay as it's just a css file that's been 'rebuilt'

Zainzzkk (approved):
Let's run this by Clara when she is back, but lets put it in main for now

PR #4095, opened 2024-12-20
fix(DTFS2-7616): added ms notifications web hook

abhi-markan:
## Introduction :pencil2:
Ensure automatic notifications are dispatched to selected channel when a deployment has finished successfully.

* Added `notify` step, this step will invoke a cURL POST request to a selected channel configured for an incoming webhook.

* Removed `notify.yml`.
* Updated TZ to `TIMEZONE` organisation variable.

### DIFF
--- a/.github/workflows/deployment.yml
+++ b/.github/workflows/deployment.yml
@@ -36,7 +36,6 @@ on:
     branches:
       # Manual deployment
       - dev
-      - fix/DTFS2-7616-efficient-deployment
 
 env:
   PRODUCT: dtfs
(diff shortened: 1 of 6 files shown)


.github/workflows/deployment.yml
```diff
@@ -156,3 +153,67 @@ jobs:
...
+      - name: MS Teams
+        run: |
+          curl --location '${{ env.WEBHOOK }}' \
+          --header 'Content-Type: application/json' \
+          --data-raw '{
+          "@type": "MessageCard",
+          "@context": "http://schema.org/extensions",
+          "themeColor": "00703c",
+          "title": "🚀 main : ${{ env.ENVIRONMENT }}",
+          "summary": "Deployed '\''main'\'' branch to '\''${{ env.ENVIRONMENT }}'\'' DTFS environment",
```
Zainzzkk, line 181: should the branch name come from github? what if deploying release?

abhi-markan: In this case it is always deploying from `main`.

Zainzzkk (approved):

PR #4097, opened 2024-12-23
chore(deps): dependencies updates

abhi-markan:
## Introduction :pencil2:
Dependencies updates.

## Miscllenous ➕ 
* Fixed `axios` version to `1.7.8` due to breaking change associated with `1.7.9`.


azure-functions/acbs-function/package.json
```diff
@@ -29,7 +29,7 @@
     "unit-test-ff": "jest --coverage --verbose --config=unit.ff.jest.config.js --passWithNoTests"
   },
   "dependencies": {
-    "axios": "^1.7.8",
+    "axios": "1.7.8",
```
Zainzzkk, line 32: dependabot may not listen to this, it did not on EXIPs but worth to see if it does

Zainzzkk (approved):

PR #4098, opened 2024-12-23
fix(DTFS2-7728): only active users are mapped

abhi-markan:
## Introduction :pencil2:
TFM users mapping function maps all the users regardless of their status, this can show invalid users in the user selection liste where only an active user should be displayed.

* Refactored TFM user mapping function to only show users with `status` property set to `active` value.

* Documentation improvement.
* DRY code refactorisation.
* Added `status` and `id` attributes to mock TFM users.
* Added unit test cases.
* Refactored TFM API constants with `@ukef/dtfs2-common` for `USER_STATUS`.
* Dependencies updates.
* Updates `azure/cli` from 1.0.0 to 2.1.0.
* Updates `docker/build-push-action` from 1 to 6.
* Updates `codecov/codecov-action` from 4 to 5.
* Updates `nginx` from 1.27.1 to 1.27.3.
* Updates `node` from 22.8.0-alpine3.20 to 23.4.0-alpine3.20.


trade-finance-manager-api/src/v1/controllers/team.controller.js
```diff
@@ -1,8 +1,43 @@
...
+      const { status, data } = members;
+      return res.status(status).send({ status, data });
+    }
+
+    // Only active users
+    const activeMembers = members.filter(({ status }) => status === USER_STATUS.ACTIVE);
 
-exports.findTeamMembers = findTeamMembers;
+    // Expected response
+    const teamMembers = activeMembers.map(({ _id, firstName, lastName }) => ({
```
ttbarnes, line 32: Looks like we could just do a single filter with a a condition instead of having to filter and then map:

```ts
const activeMembers = members.filter(({
  status,
  _id,
  firstName,
  lastName,
}) => {
  if (status === USER_STATUS.ACTIVE) {
    return {
      _id,
      firstName,
      lastName,
    };
  }

  return null;
});
```

libs/common/src/constants/tfm/users.ts
rkontogianni, line 1: It might be better rename this file to user-status.ts to better reflect its purpose?

trade-finance-manager-api/src/v1/controllers/team.controller.test.js
```diff
@@ -0,0 +1,75 @@
...
+
+    // Act
+    await mapTeamMembers(req, res);
+
+    // Assert
+    expect(api.findTeamMembers).toHaveBeenCalledWith(req.params.teamId);
+    expect(console.error).toHaveBeenCalledWith('An error occurred while mapping TFM team members %o', new Error('Mock error'));
+    expect(res.statusCode).toBe(HttpStatusCode.InternalServerError);
+  });
+});
```
ttbarnes, line 75: Do we have a test for when `api.findTeamMembers` doesn't return a status?

ttbarnes (changes requested):
Couple of minor comments, otherwise LGTM

libs/common/src/constants/tfm/users.ts
abhi-markan: Good shout, have renamed.

trade-finance-manager-api/src/v1/controllers/team.controller.js
abhi-markan: Good point, however in this instance I have explicity segreated `filter` and `map` as there will be more filter conditions to add which will make it easier to read and add/remove conditions in future, for example:

```js
// Only active users
    const activeMembers = members
      .filter(({ status }) => status === USER_STATUS.ACTIVE)
      .filter(({ disabled }) => !disabled);
    
```

With the advent of SSO this will be refactored to fetch other details.

trade-finance-manager-api/src/v1/controllers/team.controller.test.js
abhi-markan: Good spot, have added two assertions using `.not.toHaveProperty()` function to ensure neitehr `status` nor `data` exists when a successfull `200` expected response is received.

rkontogianni (approved):

trade-finance-manager-api/api-tests/v1/teams/get-team-members.api-test.js
```diff
@@ -27,96 +31,108 @@ describe('GET /teams/:teamId/members', () => {
...
+  ];
 
-    const { status, body } = await as(tokenUser).get(`/v1/teams/${unexpectedTeamId}/members`);
+  describe.each(tfmTeams)('for teamId %s', (teamId) => {
+    it(`should returns a ${HttpStatusCode.Ok} response with only the _id, first name, and last name of active team members only returned by DTFS Central`, async () => {
+      // Arrange
+      const usersReturnedByDtfsCentral = MOCK_USERS.map(({ _id, firstName, lastName, status }, index) => ({
+        _id,
+        username: `${firstName}.${lastName}.${index}@example.com`,
+        email: `${firstName}.${lastName}.${index}@example.com`,
```
Zainzzkk, line 49: not a blocker, could these both be a const since repeated

Zainzzkk (dismissed):

trade-finance-manager-api/api-tests/v1/teams/get-team-members.api-test.js
abhi-markan: Good shout, have approved.

Zainzzkk (approved):

trade-finance-manager-api/src/v1/controllers/team.controller.js
ttbarnes: Makes sense if more things will be added - could do the above in one filter btw e.g

```ts
.filter(({ disabled, status }) => !disabled && status === USER_STATUS.ACTIVE)
```

PR #4103, opened 2025-01-02
feat(FN-3163): update tests and fix null bug

natdeanlewissoftwire:
## Introduction :pencil2:
Some tests weren't accurate to reality in the current implementation, this addresses that.

Update tests to mock exporter name, and mock both get party db functions to return an array.

Null bug due to incorrect parsing of exporter in deal.party-db.js.


trade-finance-manager-api/src/v1/__mocks__/api.js
```diff
@@ -183,27 +181,28 @@ module.exports = {
...
     }
+
+    return [
+      {
+        partyUrn: 'testPartyUrn',
+      },
+    ];
+  },
+  getOrCreatePartyDbInfo: ({ companyRegNo }) => {
+    const noCompanyMatch = companyRegNo === 'NO_MATCH';
```
abhi-markan, line 195: Can `NO_MATCH` be referred from a constant please?

abhi-markan (changes requested):
One minor comment please.

PR #4105, opened 2025-01-03
fix(FN-3731): fix assertion dependent on current year

sw-mattw:
## Introduction :pencil2:
One of the "find reports by year" unit test assertions has a fixed report period year value, whilst the report period used is dependent on the current year.

This led the test to fail when the current year changed from the fixed string in the assertion (2024):

```
***     "formattedReportPeriod": "January 2024",
  +     "formattedReportPeriod": "January 2025",
```

Related failing runs:
https://github.com/UK-Export-Finance/dtfs2/actions/runs/12584970329/job/35110286115?pr=4089#logs

- Fixed the year used for the report period

N/A


abhi-markan (approved):
LGTM, please add the ticket number in the title and then merge. Thanks

PR #4086, opened 2024-12-17
feat(FN-3667): pending corrections list on portal

RLavender98:
## Introduction :pencil2:
When a PDC user has requested a correction we need to give the bank users a way to view the request and action it. This will happen by when they go to the report upload page if there are any correction requests that need a response displaying them in a table instead of displaying the report upload. 

## Resolution :heavy_check_mark:
- Add a new central-api endpoint to fetch all pending corrections from the oldest report with pending corrections for a bank.
- Update the get upload controller to check if the feature flag is enabled and if so check for pending corrections and if there are any render the pending corrections template instead of the report upload template.
- New pending corrections template for the new view of the page.

## Miscellaneous :heavy_plus_sign:
- Rewrite some existing controllers and tests written in javascript in typescript
- e2e-test set up for record correction tests in portal
- Rename `validateTfmFeeRecordCorrectionsFeatureFlagEnabled` validation middleware to `validateFeeRecordCorrectionsFeatureFlagEnabled`

## Out of scope of PR
I have not include the reported fees and fees paid to ukef in the reported period columns because there is an ongoing discussion about what data should be displayed in these columns. A follow up PR will be raised once the requirement has been finalised.

## Screenshot
With report due for upload:
Screenshot 2024-12-18 at 15 08 00

Without a report due for upload:
Screenshot 2024-12-18 at 15 05 29

PR #4106, opened 2025-01-03
feat(FN-3668): add record correction entry screen e2e tests

sw-mattw:
## Introduction :pencil2:
The record correction entry screen requires e2e tests.

The bank journey has only been partially implemented at present. e2e test logic/assertions for the complete flow will be added under subsequent tickets.

At present the correction entry screen must be navigated to by changing the URL, rather than clicking through to a correction (awaiting merge of https://github.com/UK-Export-Finance/dtfs2/pull/4086) - this behavior will be updated under the validation work for this screen (FN-3688).

Further, the review screen is yet to be implemented - once this is in place these e2e tests will be expanded to input data into the fields and test button behaviour.

- Adds initial e2e tests for the correction entry screen, without testing the complete flow

- Adds `insertFeeRecordCorrectionsIntoDb` cypress task (picked from https://github.com/UK-Export-Finance/dtfs2/pull/4086)
- Moves `validateTfmFeeRecordCorrectionFeatureFlagIsEnabled` middleware out of TFM to `libs/common` (as this is being applied on the portal route introduced by this ticket)

### DIFF
--- a/e2e-tests/portal/cypress/e2e/journeys/payment-report-officer/record-corrections/feature-flag-enabled/provide-correction.spec.js
+++ b/e2e-tests/portal/cypress/e2e/journeys/payment-report-officer/record-corrections/feature-flag-enabled/provide-correction.spec.js
@@ -11,7 +11,7 @@ import {
 } from '@ukef/dtfs2-common';
 import { NODE_TASKS, BANK1_PAYMENT_REPORT_OFFICER1 } from '../../../../../../../e2e-fixtures';
 import relative from '../../../../relativeURL';
-import { provideCorrection } from '../../../../pages';
+import { provideCorrection, pendingCorrections } from '../../../../pages';
 import { correctionRequestDetails } from '../../../../partials';
 
 context('Provide correction - Fee record correction feature flag enabled', () => {
@@ -74,9 +74,9 @@ context('Provide correction - Fee record correction feature flag enabled', () =>
       context('and when a user has initiated the "provide correction" journey', () => {
         beforeEach(() => {
           cy.login(BANK1_PAYMENT_REPORT_OFFICER1);
+          cy.visit(relative(`/utilisation-report-upload`));
 
-          // TODO FN-3688: Update this to navigate through from the pending corrections list - dependent on #4086 being merged.
-          cy.visit(relative(`/utilisation-reports/provide-correction/${pendingCorrectionDetails.id}`));
+          pendingCorrections.row(1).correctionLink().click();
         });
 
         it('should be able to view the record correction request summary on the "provide correction" screen', () => {
--- a/e2e-tests/portal/cypress/e2e/pages/utilisation-report-service/record-corrections/pendingCorrections.js
+++ b/e2e-tests/portal/cypress/e2e/pages/utilisation-report-service/record-corrections/pendingCorrections.js
@@ -5,6 +5,7 @@ const page = {
     facilityId: () => cy.get(`[data-cy="pending-corrections-table"] tr:nth-child(${index}) [data-cy="pending-corrections-row--facility-id"]`),
     exporter: () => cy.get(`[data-cy="pending-corrections-table"] tr:nth-child(${index}) [data-cy="pending-corrections-row--exporter"]`),
     errorSummary: () => cy.get(`[data-cy="pending-corrections-table"] tr:nth-child(${index}) [data-cy="pending-corrections-row--error-summary"]`),
+    correctionLink: () => cy.get(`[data-cy="pending-corrections-table"] tr:nth-child(${index}) [data-cy="correction-link"]`),
   }),
   correctionsHeading: () => cy.get('[data-cy="pending-corrections-heading"]'),
   nextReportDueHeading: () => cy.get('[data-cy="next-report-due-heading"]'),


e2e-tests/portal/cypress/e2e/journeys/payment-report-officer/record-corrections/feature-flag-enabled/provide-correction.spec.js
```diff
@@ -0,0 +1,107 @@
...
+
+          cy.task(NODE_TASKS.INSERT_UTILISATION_REPORTS_INTO_DB, [report]);
+          cy.task(NODE_TASKS.INSERT_FEE_RECORDS_INTO_DB, [feeRecord]);
+          cy.task(NODE_TASKS.INSERT_FEE_RECORD_CORRECTIONS_INTO_DB, [pendingCorrection]);
+        });
+      });
+
+      afterEach(() => {
+        cy.task(NODE_TASKS.DELETE_ALL_FROM_SQL_DB);
+      });
```
RLavender98, line 76: Could we have this as a before each in the same level as the 'after' and then remove the 'before'?

RLavender98 (commented):
Just a few minor comments.

There isn't a change to the test.yml which I would expect is needed for this to run in the pipeline. Is the plan to hold off on merging this until the pending corrections work is in and then rerun the pipeline and merge? If not I think it needs adding as part of this.

e2e-tests/portal/cypress/e2e/journeys/payment-report-officer/record-corrections/feature-flag-enabled/provide-correction.spec.js
Zainzzkk, line 1: is there a way to do this manually through the UI to test user flow?

e2e-tests/portal/cypress/e2e/journeys/payment-report-officer/record-corrections/feature-flag-enabled/provide-correction.spec.js
```diff
@@ -0,0 +1,102 @@
...
+  mapReasonsToDisplayValues,
+  getFormattedCurrencyAndAmount,
+} from '@ukef/dtfs2-common';
+import { NODE_TASKS, BANK1_PAYMENT_REPORT_OFFICER1 } from '../../../../../../../e2e-fixtures';
+import relative from '../../../../relativeURL';
+import { provideCorrection } from '../../../../pages';
+import { correctionRequestDetails } from '../../../../partials';
+
+context('Provide correction - Fee record correction feature flag enabled', () => {
+  context('Report GEF Utilisation and fees page', () => {
```
Zainzzkk, line 18: utilisation please (without capital u)

e2e-tests/portal/cypress/e2e/journeys/payment-report-officer/record-corrections/feature-flag-enabled/provide-correction.spec.js
Zainzzkk: actually, my PR once it is in has the manual journey and some commands to do it

e2e-tests/portal/cypress/e2e/pages/utilisation-report-service/record-corrections/provideCorrection.js
```diff
@@ -0,0 +1,16 @@
+const page = {
+  facilityIdInput: () => cy.get('input[data-cy="FACILITY_ID_INCORRECT-input"]'),
```
Zainzzkk, line 2: could these be lowercase to be consistent?

sw-mattw: The first part is using the record correction reason enum value so the component test assertions can be simpler (eg when we want to check that only the input for a single reason is showing we can get the difference to get a list of reasons that shouldn't show then iterate over those, passing in the enum value to the data-cy attribute).

Can change these to lowercase and refactor if needed, or maybe change to use the constant (below)? No strong preference

```suggestion
  facilityIdInput: () => cy.get(`input[data-cy="${RECORD_CORRECTION_REASON.FACILITY_ID_INCORRECT}-input"]`),
```

Thanks

e2e-tests/portal/cypress/e2e/journeys/payment-report-officer/record-corrections/feature-flag-enabled/provide-correction.spec.js
sw-mattw: I think ideally it'd be a manual flow yeah - on the bank side there's a few bits not implemented/merged yet so planning for a future ticket to add the manual flow in once the journey is more complete, there's a few extra details in the PR description, I've added a TODO comment with a ticket number to address this too ([this comment](https://github.com/UK-Export-Finance/dtfs2/pull/4106/files/af332600cda81febd1056c35c9798f1a9bac1150#r1901864546))

RLavender98: > my PR once it is in has the manual journey and some commands to do it

Creating a correction is on the TFM side, and providing the correction is on the portal side. To keep the tests running quickly and without a large amount of overhead I suggest we do the following:
- TFM specific tests that create corrections
- Portal specific tests which insert corrections into the db directly as Matt is doing here to avoid having to do the entire TFM login and creation flow for every single test
- A couple of fully e2e-tests in the 'ukef' folder that test the entire correction journey from start to finish in a single test, manually doing both the TFM then bank side and then back in TFM side and checking the correction has actually taken effect. This will need to added when the journey is pretty much complete, so when we do the ticket that actually applies the new values to the fee record is implemented

e2e-tests/portal/cypress/e2e/pages/utilisation-report-service/record-corrections/provideCorrection.js
Zainzzkk: Just worried that if we change the ENUM of the reason, then this will be out of date.  maybe we can tackle this later with a change in the njk.  This is fine for now

e2e-tests/portal/cypress/e2e/journeys/payment-report-officer/record-corrections/feature-flag-enabled/provide-correction.spec.js
Zainzzkk: let's explore improving this later if we have time and once everything is in place.  Would be good to have a ticket for it

portal/api-tests/utilisation-report-routes.api-test.js
```diff
@@ -46,11 +46,34 @@ describe('utilisation-report routes', () => {
   describe('GET /utilisation-reports/provide-correction/:correctionId', () => {
     const getUrl = ({ correctionId }) => `/utilisation-reports/provide-correction/${correctionId}`;
 
-    withRoleValidationApiTests({
-      makeRequestWithHeaders: (headers) => get(getUrl({ correctionId: 1 }), {}, headers),
-      whitelistedRoles: [ROLES.PAYMENT_REPORT_OFFICER],
-      successCode: HttpStatusCode.Ok,
-      disableHappyPath: true,
+    describe('when FF_FEE_RECORD_CORRECTION_ENABLED is set to `true`', () => {
+      beforeAll(() => {
```
Zainzzkk, line 50: do we have an after in this test to reset the process.env?

RLavender98 (commented):
Looks good, just the comment from Zain about env vars - happy for you to merge once you have his approval

portal/api-tests/utilisation-report-routes.api-test.js
sw-mattw: Added thanks, now resets to the original env vars

Zainzzkk (dismissed):

trade-finance-manager-ui/server/middleware/validateFeeRecordCorrectionFeatureFlagIsEnabled/index.ts
sw-mattw, line 1: Moved to `libs/common`

Zainzzkk (dismissed):

sw-mattw:
Just resolved a TODO on the E2E tests now #4086 has been merged - tests more of the flow

PR #4108, opened 2025-01-03
feat(FN-3578): improve email format for pdc correction request

sw-mattw:
## Introduction :pencil2:
Email addresses on the "check the information" screen of the PDC "create fee record correction request" journey are comma-separated and break on dashes.

A comment in demo highlighted that is difficult to read when there are several email addresses.

Further, the key should have the suffix `(es)` to handle the case of multiple emails.

- `(es)` suffix added to emails key
- Emails separated with a comma followed by a new line, breaking on any character when the containers max width is reached
  - This is the designers preference

### Screenshots

FN-3578_PR_one-line-email

N/A


RLavender98 (approved):
Nice improvement - thanks!

Zainzzkk (approved):

PR #4112, opened 2025-01-06
fix(DTFS2-10543): updated test mocks

abhi-markan:
## Introduction :pencil2:
Updated test mocks to reflect change of year.

* Updated to `2025`.

### DIFF
--- a/trade-finance-manager-ui/server/controllers/utilisation-reports/find-reports-by-year/index.test.ts
+++ b/trade-finance-manager-ui/server/controllers/utilisation-reports/find-reports-by-year/index.test.ts
@@ -341,7 +341,7 @@ describe('controllers/utilisation-reports/find-reports-by-year', () => {
         dateUploaded: '2024-02-15 10:38:01.4033333',
         totalFeesReported: 3,
         reportedFeesLeftToReconcile: 3,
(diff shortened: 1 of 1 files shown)


trade-finance-manager-ui/server/controllers/utilisation-reports/find-reports-by-year/index.test.ts
```diff
@@ -341,7 +341,7 @@ describe('controllers/utilisation-reports/find-reports-by-year', () => {
         dateUploaded: '2024-02-15 10:38:01.4033333',
         totalFeesReported: 3,
         reportedFeesLeftToReconcile: 3,
-        formattedReportPeriod: 'January 2024',
+        formattedReportPeriod: 'January 2025',
```
Zainzzkk, line 344: could we not maybe have a const on the top to get the year `new Date().getFullYear()` and then construct this string using the year const?

abhi-markan: Good shout, have added `formattedReportPeriod`. Thanks

Zainzzkk (approved):

PR #4113, opened 2025-01-06
feat(DTFS2-6892): update mock builder, enforce complete defaults, document mocked class methods

AlexBramhill:
After reverting commits -- this is a copy of the following PR https://github.com/UK-Export-Finance/dtfs2/pull/4067

## Introduction :pencil2:
We sometimes need to include sync methods in our mock builders that are the same implementation as the service being mocked.

This is because of the existing policy of not mocking sync methods in tests

Make mock builders require all params of the mocked class/type to be specified
Add documentation and example for how to pass through existing implementation of mocked class to mock builder
Make mock builders automatically instantiate defaults instead of having to use `withDefaults`
Update documentation with examples


abhi-markan (approved):

libs/common/src/test-helpers/mock-builders/mock-builder.ts
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  /**
+   * Set fields on the class instance when you want to override the defaults
+   *
+   * @example
+   * ```ts
+   * const user = new UserMockBuilder()
+   *   .with({
+   *     email: 'new.email@ukef.gov.uk',
```
abhi-markan, line 94:
```suggestion
   *     email: 'new.email@ukexportfinance.gov.uk',
```

abhi-markan (changes requested):
A minor comment to address please.

abhi-markan (approved):

PR #4116, opened 2025-01-06
feat(FN-3689): add endpoint for persisting bank correction data

RLavender98:
## Introduction :pencil2:
As the bank user navigates through the provide correction flow we need to store their answers between pages. This is an initial PR for this ticket adding an endpoint to save the transient form data.

- Add new PUT endpoint for saving correction transient form data


libs/common/src/sql-db-connection/migrations/1734616272101-AddFeeRecordCorrectionTransientFormDataTable.ts
```diff
@@ -0,0 +1,26 @@
...
+export class AddFeeRecordCorrectionTransientFormDataTable1734616272101 implements MigrationInterface {
+  name = 'AddFeeRecordCorrectionTransientFormDataTable1734616272101';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`
+            CREATE TABLE "FeeRecordCorrectionTransientFormData" (
+                "userId" nvarchar(255) NOT NULL,
+                "correctionId" int NOT NULL,
+                "formDataSerialized" nvarchar(1000) NOT NULL,
+                "lastUpdatedAt" datetime2 NOT NULL CONSTRAINT "DF_6bf1a666c5c54f26a80a93cb477" DEFAULT getdate(),
```
RLavender98, line 12: @ChristianMcc these not null constraints were generated with the same name used on the request transient form data table, I assume since they are generated using table name and column name at the time of generation. I simply replaced the random looking part with the starts of new guids, this works locally and I think is fine, but since we are determining this ourselves we could also go for more descriptive names. Just wanted to flag in case you think there could be any issues with this?

sw-mattw (approved):

dtfs-central-api/src/v1/controllers/utilisation-report-service/fee-record-correction/put-fee-record-correction-transient-form-data.controller/index.test.ts
```diff
@@ -0,0 +1,172 @@
...
+    it('should call the correction repo to fetch the correction with the correct parameters', async () => {
+      // Arrange
+      mockFindCorrection.mockResolvedValue(new FeeRecordCorrectionEntityMockBuilder().build());
+
+      // Act
+      await putFeeRecordCorrectionTransientFormData(req, res);
+
+      // Assert
+      expect(mockFindCorrection).toHaveBeenCalledWith(correctionId, bankId);
+    });
```
Zainzzkk, line 106: these can be combined if you want

dtfs-central-api/src/v1/controllers/utilisation-report-service/fee-record-correction/put-fee-record-correction-transient-form-data.controller/index.ts
```diff
@@ -0,0 +1,66 @@
...
+    const userId = user.id;
+
+    const correction = await FeeRecordCorrectionRepo.findByIdAndBankId(correctionId, bankId);
+
+    if (!correction) {
+      throw new NotFoundError(`Failed to find a correction with id '${correctionId}' for bank id '${bankId}'`);
+    }
+
+    // TODO: FN-3688 - validate the form data has the fields expected for the given reasons
+    const validatedFormData = formData as unknown as RecordCorrectionTransientFormData;
```
Zainzzkk, line 43: can this line be removed but keep the todo comment.  Should not be adding unknown and this line will be added soon anyways.  same with any other unknowns in this PR

dtfs-central-api/src/v1/controllers/utilisation-report-service/fee-record-correction/put-fee-record-correction-transient-form-data.controller/index.ts
```diff
@@ -0,0 +1,66 @@
...
+        userId,
+      },
+    });
+
+    await FeeRecordCorrectionTransientFormDataRepo.save(newTransientFormData);
+
+    return res.sendStatus(HttpStatusCode.Ok);
+  } catch (error) {
+    const errorMessage = `Failed to put fee record correction transient form data`;
+    console.error(errorMessage, error);
```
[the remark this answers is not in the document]
Zainzzkk: keeps it aligned with the rest of DTFS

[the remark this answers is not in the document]
Zainzzkk: sorry actually console.error('%s %o', errorMessage, error); since errorMessage is being used again

libs/common/src/sql-db-connection/migrations/1734616272101-AddFeeRecordCorrectionTransientFormDataTable.ts
```diff
@@ -0,0 +1,26 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class AddFeeRecordCorrectionTransientFormDataTable1734616272101 implements MigrationInterface {
```
Zainzzkk, line 3: could we leave the number out of the class?

libs/common/src/sql-db-connection/migrations/1734616272101-AddFeeRecordCorrectionTransientFormDataTable.ts
Zainzzkk, line 12: could hardcoding the id not cause issues later on if the id changes for any reason? or if duplicated.  Ignore me if not an issue

ChristianMcc: Hey, yeah a while ago we did think about naming these as they are just generated guids to be unique in name. I think naming them is a good idea and no issue with it - would recommend keeping the same identifier at the start though (e.g. a default keeping DF or unique constraint keeping UQ etc.)

dtfs-central-api/src/v1/controllers/utilisation-report-service/fee-record-correction/put-fee-record-correction-transient-form-data.controller/index.ts
RLavender98: Without this line I get a ts error and this PR would not be mergable because actually in here we need a function that parses the data. This is fine to merge because this endpoint isn't being called anywhere so no data will get created on a deployed env even on dev where the feature flag is on until the work is done to replace this type assertion with a parsing function

dtfs-central-api/src/v1/controllers/utilisation-report-service/fee-record-correction/put-fee-record-correction-transient-form-data.controller/index.ts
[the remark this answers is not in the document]
RLavender98: I'll raise a ticket!

libs/common/src/sql-db-connection/migrations/1734616272101-AddFeeRecordCorrectionTransientFormDataTable.ts
RLavender98: This is automatically generated by typeorm and matches the number in the file name. I don't think we should change this.

libs/common/src/sql-db-connection/migrations/1734616272101-AddFeeRecordCorrectionTransientFormDataTable.ts
RLavender98: I don't think it does cause issues, I tagged Christian to double check and he has confirmed this is something we are fine to set ourselves, it's just a name for the constraint

PR #4117, opened 2025-01-07
feat(DTFS2-7735): refactor Facility End Date validation

MarRobSoftwire:
## Introduction :pencil2:
The new amendments page should use the commonised date validation in libs/common. It makes sense to revisit the Facility End Date and Bank Review Date validation & commonise this too

- Update Faciltity End Date and Bank Review Date validation in gef-ui to use comonised date validation
- Add new `mapValidationError` function to handle converting tfm template errors to gef-ui template error
- Add `TfmValidationError` to distinguish from gef-ui validation error
- Update `applyStandardValidationAndParseDateInput` to handle undefined values 
- Delete obsolete `validateAndParseDayMonthYear` function from gef-ui

- Update about facility validation to use commonised date validation & mapping
- Update tests to reflect new validation errors


e2e-tests/gef/cypress/e2e/facility/pre-submission/about-facility.spec.js
```diff
@@ -114,19 +114,19 @@ context('About Facility Page  - feature flag enabled', () => {
       cy.completeDateFormFields({ idPrefix: 'cover-end-date', day: `${today.dayLong}-`, month: today.monthLong, year: today.year });
 
       cy.clickContinueButton();
-      errorSummary().contains('The day for the cover start date must include 1 or 2 numbers');
-      errorSummary().contains('The day for the cover end date must include 1 or 2 numbers');
-      aboutFacility.coverStartDateError().contains('The day for the cover start date must include 1 or 2 numbers');
-      aboutFacility.coverEndDateError().contains('The day for the cover end date must include 1 or 2 numbers');
+      errorSummary().contains('Cover start date must be a real date');
```
MarRobSoftwire, line 117: These error messages (and throughout the PR) been replaced with the standard ones generated by `applyStandardValidationAndParseDateInput` - these are based on Beths work looking at the [govuk date validation guidance](https://design-system.service.gov.uk/components/date-input/)

gef-ui/server/controllers/about-facility/validation.js
```diff
@@ -70,56 +69,68 @@ const validateAboutFacility = ({
...
-      if (coverEndDateFormattingErrors) {
-        aboutFacilityErrors.push(...coverEndDateFormattingErrors);
+      if (coverEndDateFormattingError) {
+        coverEndDateValid = false;
+
+        aboutFacilityErrors.push({
+          errRef: coverEndDateErrRef,
+          errMsg: coverEndDateFormattingError.message,
+          subFieldErrorRefs: coverEndDateFormattingError.fieldRefs,
+        });
```
MarRobSoftwire, line 133: Add helper to map this

gef-ui/server/utils/map-validation-error.test.ts
```diff
@@ -0,0 +1,26 @@
+import { GenericValidationError } from '@ukef/dtfs2-common';
+import { ValidationError } from '../types/validation-error';
+import { mapValidationError } from './map-validation-error';
+
+describe('mapValidationError', () => {
+  it('should map the validation error', () => {
+    // Arrange
+    const tfmValidationError: GenericValidationError = {
```
BethThomas141, line 8: think the `tfmValidationError` needs updating here

gef-ui/server/utils/map-validation-error.ts
```diff
@@ -0,0 +1,16 @@
+import { GenericValidationError } from '@ukef/dtfs2-common';
+import { ValidationError } from '../types/validation-error';
+
+/**
+ * The commonised validation `applyStandardValidationAndParseDateInput` returns a `GenericValidationError` object but
+ * in gef-ui a different type `ValidationError` is used. This function maps the `GenericValidationError` to a `ValidationError`.
+ * @param validationError - a tfm validation error
```
BethThomas141, line 7: Also remove the `tfm` here

libs/common/src/types/tfm/validation-error.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * This validation error is used in trade-finance-manager-ui but can be mapped for
```
BethThomas141, line 2: I guess another option here is for this to not live in the `tfm` subfolder and have a comment like

>  This validation error is returned by common validation functions such as the date validation. trade-finance-manager-ui also uses this type within the service, for other services this type can be mapped to the desired validation error structure

I think I'd prefer it like that, but it doesn't matter too much so happy to leave as is as well

gef-ui/server/utils/map-validation-error.test.ts
```diff
@@ -0,0 +1,26 @@
+import { GenericValidationError } from '@ukef/dtfs2-common';
+import { ValidationError } from '../types/validation-error';
+import { mapValidationError } from './map-validation-error';
+
+describe('mapValidationError', () => {
+  it('should map the validation error', () => {
```
abhi-markan, line 6: Can you please add an assertion where `errCode` is also returned from type `ValidationError`.

libs/common/src/helpers/date-validation.ts
```diff
@@ -184,7 +178,7 @@ export const applyStandardValidationAndParseDateInput = (inputtedDate: DayMonthY
     year: `${valueRef}-year`,
   };
 
-  const trimmedInputtedDate = { day: inputtedDate.day.trim(), month: inputtedDate.month.trim(), year: inputtedDate.year.trim() };
+  const trimmedInputtedDate = { day: inputtedDate.day?.trim() || '', month: inputtedDate.month?.trim() || '', year: inputtedDate.year?.trim() || '' };
```
abhi-markan, line 181: Can we perform conditional check in a variable and pass just the variable.

```ts
const day = inputtedDate.day?.trim() || '';
const trimmedInputtedDate = {day...);
```

abhi-markan (changes requested):
Nice work, minor comments please.

gef-ui/server/utils/map-validation-error.test.ts
MarRobSoftwire: `GenericValidationError` doesn't contain an errCode so I dont think its possible to test this

libs/common/src/types/tfm/validation-error.ts
MarRobSoftwire: Good shout, updated

BethThomas141 (approved):
LGTM🚀

abhi-markan (approved):

PR #4128, opened 2025-01-10
chore(FN-3742): remove all instances of res being passed to a function

Zainzzkk:
## Introduction :pencil2:
This PR removes all instances of res being passed to a downstream function

* Removed res being passed to another function to render a template

* Added correct string substitution where it was missing for console.errors

### DIFF
--- a/trade-finance-manager-ui/server/controllers/utilisation-reports/edit-payment/index.ts
+++ b/trade-finance-manager-ui/server/controllers/utilisation-reports/edit-payment/index.ts
@@ -18,8 +18,6 @@ import { getEditPaymentsCheckboxIdsFromObjectKeys } from '../../../helpers/edit-
 import { EditPaymentsTableCheckboxId } from '../../../types/edit-payments-table-checkbox-id';
 import { ReconciliationForReportTab } from '../../../types/reconciliation-for-report-tab';
 
-const renderEditPaymentPage = (res: Response, viewModel: EditPaymentViewModel) => res.render('utilisation-reports/edit-payment.njk', viewModel);
-
 type GetEditPaymentRequest = CustomExpressRequest<{
   query: {
     redirectTab?: ReconciliationForReportTab;
@@ -38,7 +36,7 @@ export const getEditPayment = async (req: GetEditPaymentRequest, res: Response) 
     const paymentDetails = await api.getPaymentDetailsWithFeeRecords(reportId, paymentId, userToken);
 
     if (formValues) {
-      const editPaymentViewModel = getEditPaymentViewModelWithFormValues(
+      const editPaymentViewModel: EditPaymentViewModel = getEditPaymentViewModelWithFormValues(
         paymentDetails,
         reportId,
         paymentId,
@@ -47,13 +45,15 @@ export const getEditPayment = async (req: GetEditPaymentRequest, res: Response) 
         redirectTab,
         errors,
       );
-      return renderEditPaymentPage(res, editPaymentViewModel);
+
+      return res.render('utilisation-reports/edit-payment.njk', editPaymentViewModel);
     }
 
-    const editPaymentViewModel = getEditPaymentViewModel(paymentDetails, reportId, paymentId, isCheckboxChecked, redirectTab, errors);
-    return renderEditPaymentPage(res, editPaymentViewModel);
+    const editPaymentViewModel: EditPaymentViewModel = getEditPaymentViewModel(paymentDetails, reportId, paymentId, isCheckboxChecked, redirectTab, errors);
+
+    return res.render('utilisation-reports/edit-payment.njk', editPaymentViewModel);
   } catch (error) {
-    console.error('Error updating utilisation report status:', error);
+    console.error('Error updating utilisation report status: %o', error);
     return res.render('_partials/problem-with-service.njk', { user });
   }
 };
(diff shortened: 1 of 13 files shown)


trade-finance-manager-ui/server/controllers/utilisation-reports/edit-payment/index.ts
```diff
@@ -47,13 +45,15 @@ export const getEditPayment = async (req: GetEditPaymentRequest, res: Response)
...
-      return renderEditPaymentPage(res, editPaymentViewModel);
+
+      return res.render('utilisation-reports/edit-payment.njk', editPaymentViewModel);
     }
 
-    const editPaymentViewModel = getEditPaymentViewModel(paymentDetails, reportId, paymentId, isCheckboxChecked, redirectTab, errors);
-    return renderEditPaymentPage(res, editPaymentViewModel);
+    const editPaymentViewModel: EditPaymentViewModel = getEditPaymentViewModel(paymentDetails, reportId, paymentId, isCheckboxChecked, redirectTab, errors);
+
+    return res.render('utilisation-reports/edit-payment.njk', editPaymentViewModel);
```
RLavender98, line 54: Could we extract the template path into a const and use that on line 49, line 99 and this line to avoid duplication please

trade-finance-manager-ui/server/controllers/utilisation-reports/find-reports-by-year/index.ts
```diff
@@ -107,24 +93,28 @@ export const getFindReportsByYear = async (req: Request, res: Response) => {
...
       user,
       activePrimaryNavigation: PRIMARY_NAVIGATION_KEYS.UTILISATION_REPORTS,
       bankName,
       year,
       reports: reportSummaryItemsViewModel,
       isTfmPaymentReconciliationFeatureFlagEnabled: isTfmPaymentReconciliationFeatureEnabled,
-    });
+    };
+
+    return res.render('utilisation-reports/utilisation-reports-by-bank-and-year-results.njk', viewModel);
```
RLavender98, line 115: Similarly can we pull this path into a const at the top of the file and use in lines 76 and 98 and here please

trade-finance-manager-ui/server/controllers/utilisation-reports/record-corrections/create-record-correction-request/index.ts
```diff
@@ -101,7 +100,9 @@ export const postCreateRecordCorrectionRequest = async (req: PostCreateRecordCor
       formValues,
       errors,
       backLinkHref: getLinkToPremiumPaymentsTab(reportId, [Number(feeRecordId)]),
-    });
+    };
+
+    return res.render('utilisation-reports/record-corrections/create-record-correction-request.njk', viewModel);
```
RLavender98, line 105: Same for this file

RLavender98 (changes requested):

trade-finance-manager-ui/server/controllers/utilisation-reports/record-corrections/create-record-correction-request/index.ts
Zainzzkk: done

trade-finance-manager-ui/server/controllers/utilisation-reports/find-reports-by-year/index.ts
Zainzzkk: done

trade-finance-manager-ui/server/controllers/utilisation-reports/edit-payment/index.ts
Zainzzkk: done

trade-finance-manager-ui/server/controllers/utilisation-reports/edit-payment/index.ts
```diff
@@ -24,6 +24,8 @@ type GetEditPaymentRequest = CustomExpressRequest<{
   };
 }>;
 
+const template = 'utilisation-reports/edit-payment.njk';
```
RLavender98, line 27: Please can we use a more descriptive name. We also refer to the problem with service template so there are multiple templates going around, will also make the render line more readable.

```suggestion
const editPaymentPageTemplate = 'utilisation-reports/edit-payment.njk';
```

RLavender98: Same for other similar usages of the word 'template'

ttbarnes: I think a specific name in controllers would make things harder to maintain moving forward. E.g if we call this `editPaymentPageTemplate` and the page renames to "modify payment" we'd have to rename all such variables, or they could be missed

I think `template` is succinct enough as the parent directory is called `edit-template`. There should never be more than one template referenced in a controller other than "problem with service" which is self explanatory. Maybe in another PR all template references could be moved into constants and imported so e.g

```
TEMPLATES.EDIT_PAYMENT
TEMPLATES.PROBLEM_WITH_SERVICE
```

Zainzzkk: It would be good to have all the templates in a constants file but that will be for later. 
Where there are multiple templates in a file I have made sure the const is more descriptive

RLavender98 (approved):

PR #4134, opened 2025-01-13
feat(FN-3689): handle saved values of zero on correction form

RLavender98:
## Introduction :pencil2:
We noticed that if a user entered zero as the utilisation or reported fee amount for a correction then went to the review page and then went back the field would be empty as if they had not entered a value.

- Add handling for zero as a possible value for a persisted monetary correction value.

Screenshot 2025-01-13 at 14 43 22

### DIFF
--- a/portal/server/controllers/utilisation-report-service/record-correction/provide-utilisation-report-correction/helpers.ts
+++ b/portal/server/controllers/utilisation-report-service/record-correction/provide-utilisation-report-correction/helpers.ts
@@ -96,9 +96,14 @@ export const getAdditionalCommentsFieldLabels = (correctionReasons: RecordCorrec
  */
 export const mapToProvideCorrectionFormValuesViewModel = (
   savedFormValues: GetFeeRecordCorrectionTransientFormDataResponseBody,
-): ProvideCorrectionFormValuesViewModel => ({
-  facilityId: savedFormValues.facilityId ?? null,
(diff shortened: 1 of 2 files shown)


sw-mattw (dismissed):

portal/server/controllers/utilisation-report-service/record-correction/provide-utilisation-report-correction/helpers.ts
```diff
@@ -98,7 +98,7 @@ export const mapToProvideCorrectionFormValuesViewModel = (
   savedFormValues: GetFeeRecordCorrectionTransientFormDataResponseBody,
 ): ProvideCorrectionFormValuesViewModel => ({
   facilityId: savedFormValues.facilityId ?? null,
-  utilisation: savedFormValues.utilisation ? getFormattedMonetaryValue(savedFormValues.utilisation) : null,
-  reportedFee: savedFormValues.reportedFee ? getFormattedMonetaryValue(savedFormValues.reportedFee) : null,
+  utilisation: savedFormValues.utilisation !== undefined ? getFormattedMonetaryValue(savedFormValues.utilisation) : null,
+  reportedFee: savedFormValues.reportedFee !== undefined ? getFormattedMonetaryValue(savedFormValues.reportedFee) : null,
```
Zainzzkk, line 102: would these be better as consts above for readability?

Zainzzkk: or a helper

Zainzzkk (approved):

PR #4145, opened 2025-01-15
feat(FN-3736): remaining delete transient form data scenarios

RLavender98:
## Introduction :pencil2:
When a user abandons the correction journey without clicking cancel they might start the journey again with form data already saved. When they start a new journey we therefore need to make an api call to delete any old saved data.

Also in this PR, add deletion of correction transient form data to the endpoint that will be responsible for saving the values and completes the journey so any unused old data is not hanging around.

- Add repo delete call to the PUT fee record correction endpoint
- Add call to the transient form data delete endpoint when user is starting the provide correction journey anew
- Add e2e-tests for the changes in this PR and the initial PR for this ticket which covers deleting the form data when the user clicks cancel on the review page

### DIFF
--- a/portal/server/controllers/utilisation-report-service/record-correction/provide-utilisation-report-correction/index.test.ts
+++ b/portal/server/controllers/utilisation-report-service/record-correction/provide-utilisation-report-correction/index.test.ts
@@ -64,9 +64,21 @@ describe('controllers/utilisation-reports/record-corrections/create-record-corre
       jest.resetAllMocks();
     });
 
-    describe('when there are NOT any saved form values', () => {
+    describe('when the user visits the page from a link on the Report GEF utilisation and fees page', () => {
       beforeEach(() => {
-        jest.mocked(api.getFeeRecordCorrectionTransientFormData).mockResolvedValue({});
+        req.headers.referer = 'utilisation-report-upload';
+      });
+
+      it('should clear any saved form data for the correction', async () => {
+        // Arrange
+        jest.mocked(api.getFeeRecordCorrection).mockResolvedValue(aGetFeeRecordCorrectionResponseBody());
+
+        // Act
+        await getProvideUtilisationReportCorrection(req, res);
+
+        // Assert
+        expect(api.deleteFeeRecordCorrectionTransientFormData).toHaveBeenCalledTimes(1);
+        expect(api.deleteFeeRecordCorrectionTransientFormData).toHaveBeenCalledWith(userToken, bankId, correctionId);
       });
 
       it('should render the "provide utilisation report correction" page', async () => {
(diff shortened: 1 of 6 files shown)


portal/server/controllers/utilisation-report-service/record-correction/provide-utilisation-report-correction/index.ts
```diff
@@ -27,7 +28,20 @@ export const getProvideUtilisationReportCorrection = async (req: GetProvideUtili
...
 
-    const savedFormValues = await api.getFeeRecordCorrectionTransientFormData(userToken, bankId, correctionId);
+    let savedFormValues: GetFeeRecordCorrectionTransientFormDataResponseBody = {};
+
+    /**
+     * If the user is coming from a link on the Report GEF utilisation and fees page
+     * then we want to clear any saved form data for the correction as they are starting
+     * a new journey.
+     */
+    const isNewJourney = req.headers.referer?.includes('utilisation-report-upload');
```
Zainzzkk, line 38: is it possible to double check this returns the 'utilisation-report-upload' locally?
Locally this did not return the last part of the url for me

Zainzzkk (approved):

portal/server/controllers/utilisation-report-service/record-correction/provide-utilisation-report-correction/index.ts
RLavender98: Yep I have tested this locally and it works correctly and is captured by and e2e test case also

portal/server/controllers/utilisation-report-service/record-correction/provide-utilisation-report-correction/index.test.ts
```diff
@@ -64,9 +64,21 @@ describe('controllers/utilisation-reports/record-corrections/create-record-corre
       jest.resetAllMocks();
     });
 
-    describe('when there are NOT any saved form values', () => {
+    describe('when the user visits the page from a link on the Report GEF utilisation and fees page', () => {
```
sw-mattw, line 67: Should we add another test to this describe block to check that `getFeeRecordCorrectionTransientFormData` is not called (to test that it's in an else branch)?

portal/server/controllers/utilisation-report-service/record-correction/provide-utilisation-report-correction/index.test.ts
```diff
@@ -105,50 +117,97 @@ describe('controllers/utilisation-reports/record-corrections/create-record-corre
...
-          user: mockUser,
-          primaryNav: PRIMARY_NAV_KEY.UTILISATION_REPORT_UPLOAD,
-          correctionRequestDetails: expectedCorrectionRequestDetails,
-          paymentCurrencyOptions: expectedPaymentCurrencyOptions,
-          additionalComments: {
-            label: expectedAdditionalCommentsLabel,
-            hint: expectedAdditionalCommentsHint,
-          },
-          formValues: mapToProvideCorrectionFormValuesViewModel(savedFormValues),
+      describe('when there are saved form values', () => {
```
sw-mattw, line 166: Should we be checking that `getFeeRecordCorrectionTransientFormData` has been called with correct params?

sw-mattw (approved):

PR #4146, opened 2025-01-16
feat(FN-3670): correction sent page e2e-tests

RLavender98:
## Introduction :pencil2:
The correction sent page was added before there was a complete journey for the user to get to the page. This PR adds those e2e-tests now that the user journey is complete enough for them.

- Add e2e-tests for the correction sent page

- Update data-cy that was just 'paragraph' to a more descriptive name

### DIFF
--- a/e2e-tests/portal/cypress/e2e/pages/utilisation-report-service/record-corrections/correctionSent.js
+++ b/e2e-tests/portal/cypress/e2e/pages/utilisation-report-service/record-corrections/correctionSent.js
@@ -1,6 +1,6 @@
 const page = {
   emailText: () => cy.get('[data-cy="email-text"]'),
-  emailListItem: (email) => cy.get(`[data-cy="email-list"] li:contains(${email})`),
+  emailList: () => cy.get('[data-cy="email-list"]'),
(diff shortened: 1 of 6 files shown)


e2e-tests/portal/cypress/e2e/pages/utilisation-report-service/record-corrections/correctionSent.js
```diff
@@ -0,0 +1,7 @@
+const page = {
+  emailText: () => cy.get('[data-cy="email-text"]'),
+  emailListItem: (email) => cy.get(`[data-cy="email-list"] li:contains(${email})`),
```
Zainzzkk, line 3: Pages file should just be selectors and not assertions. can this just be the data-cy.  the contains logic should happen in the test -  `correctionSent.emailList().contains(email)`

RLavender98: This is technically a selector - it gets the list item which contains the email if there is one and then in the test the assertion checks the element exists. This just returns null if there is not a li with the content. That said I can change it if you still want me to?

Zainzzkk (approved):

PR #4150, opened 2025-01-16
feat(FN-3675): implement custom typeorm naming strategy

RLavender98:
## Introduction :pencil2:
When I started working on the correction columns for storing old and new values using embedded entities I noticed the typeorm naming strategy was not very good for embedded columns:
     `previous` + `feesPaidToUkefForThePeriod` was becoming `previousFeespaidtoukeffortheperiod`
Which is not very readable. 

This initial PR addresses this so that these columns can be introduced with nice names.

- Implement custom naming strategy so that 'prefix' + 'embeddedProperty' becomes the column name 'prefixEmbeddedProperty'
- Rename two existing columns with incorrect casing in the names

### DIFF
--- a/libs/common/package.json
+++ b/libs/common/package.json
@@ -54,6 +54,7 @@
     "dotenv": "^16.4.5",
     "express": "^4.21.2",
     "html-entities": "^2.5.2",
+    "lodash": "^4.17.21",
     "lodash.difference": "^4.5.0",
     "mongodb": "^4.17.2",
     "mssql": "^11.0.1",
--- a/libs/common/package.json
+++ b/libs/common/package.json
@@ -54,7 +54,6 @@
     "dotenv": "^16.4.5",
     "express": "^4.21.2",
     "html-entities": "^2.5.2",
-    "lodash": "^4.17.21",
     "lodash.difference": "^4.5.0",
     "mongodb": "^4.17.2",
     "mssql": "^11.0.1",
--- a/libs/common/src/sql-db-connection/custom-naming-strategy.test.ts
+++ b/libs/common/src/sql-db-connection/custom-naming-strategy.test.ts
@@ -1,4 +1,4 @@
-import { CustomNamingStrategy, firstLetterCase } from './custom-naming-strategy';
+import { CustomNamingStrategy } from './custom-naming-strategy';
 
 describe('custom-naming-strategy', () => {
   describe('CustomNamingStrategy', () => {
@@ -15,10 +15,13 @@ describe('custom-naming-strategy', () => {
           ${'custom_name'} | ${['prefix_word']}        | ${'prefixWordCustomName'} | ${'the inputs are in snake case'}
           ${'custom name'} | ${['prefix word']}        | ${'prefixWordCustomName'} | ${'the inputs are in space separated'}
         `(
-          'should return the customName prefixed by all the prefixes in camel case $description',
+          'should return the customName prefixed by all the prefixes in camel case when $description',
           ({ customName, embeddedPrefixes, expected }: { customName: string; embeddedPrefixes: string[]; expected: string }) => {
+            // Act
             const result = strategy.columnName('propertyName', customName, embeddedPrefixes);
-            expect(result).toBe(expected);
+
+            // Assert
+            expect(result).toEqual(expected);
           },
         );
       });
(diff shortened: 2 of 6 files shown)


libs/common/package.json
```diff
@@ -54,6 +54,7 @@
     "dotenv": "^16.4.5",
     "express": "^4.21.2",
     "html-entities": "^2.5.2",
+    "lodash": "^4.17.21",
```
ChristianMcc, line 57: where is this being used?

ChristianMcc (approved):
LGTM! Nice fix for the issue, small question about lodash package

libs/common/src/sql-db-connection/custom-naming-strategy.test.ts
```diff
@@ -0,0 +1,60 @@
...
+        it.each`
+          customName       | embeddedPrefixes          | expected                  | description
+          ${'custom'}      | ${[]}                     | ${'custom'}               | ${'there are no embedded prefixes'}
+          ${'custom'}      | ${['prefix']}             | ${'prefixCustom'}         | ${'there is one embedded prefixes'}
+          ${'custom'}      | ${['prefix1', 'prefix2']} | ${'prefix1Prefix2Custom'} | ${'there are multiple embedded prefixes'}
+          ${'customName'}  | ${['prefixWord']}         | ${'prefixWordCustomName'} | ${'the inputs are in camel case'}
+          ${'custom_name'} | ${['prefix_word']}        | ${'prefixWordCustomName'} | ${'the inputs are in snake case'}
+          ${'custom name'} | ${['prefix word']}        | ${'prefixWordCustomName'} | ${'the inputs are in space separated'}
+        `(
+          'should return the customName prefixed by all the prefixes in camel case $description',
```
sw-mattw, line 18: Should this be the following?
```suggestion
          'should return the customName prefixed by all the prefixes in camel case when $description',
```

libs/common/src/sql-db-connection/custom-naming-strategy.test.ts
```diff
@@ -0,0 +1,60 @@
...
+          ({ customName, embeddedPrefixes, expected }: { customName: string; embeddedPrefixes: string[]; expected: string }) => {
+            const result = strategy.columnName('propertyName', customName, embeddedPrefixes);
+            expect(result).toBe(expected);
```
sw-mattw, line 21:
```suggestion
            expect(result).toEqual(expected);
```

libs/common/src/sql-db-connection/custom-naming-strategy.test.ts
sw-mattw: Same for the test case below this (line 36 at present)

libs/common/src/sql-db-connection/custom-naming-strategy.ts
```diff
@@ -0,0 +1,33 @@
+import { DefaultNamingStrategy } from 'typeorm';
+import { camelCase } from 'typeorm/util/StringUtils';
+
+/**
+ * Capitalises the first letter of a string without changing
+ * the casing of the rest of the string.
+ * @param name the name to capitalise
```
sw-mattw, line 7:
```suggestion
 * @param name - The name to capitalise
```

libs/common/src/sql-db-connection/custom-naming-strategy.ts
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * The DefaultNamingStrategy does not handle embedded columns correctly,
+ * removing the casing of the embedded column names.
+ *
+ * e.g. if we have an embedded property 'fooBar', and the prefix 'bazQux'
+ * the default strategy would return 'bazQuxFoobar' instead of 'bazQuxFooBar'.
+ *
+ * Based on discussion here: @link{https://github.com/typeorm/typeorm/issues/7307}.
+ */
+export class CustomNamingStrategy extends DefaultNamingStrategy {
```
sw-mattw, line 23: I suspect this will be the only naming strategy we'll add, but should we add a prefix to this class' name?

RLavender98: We'll only ever have one so I don't think it is necessary

Zainzzkk: ColumnCamelCaseNamingStrategy? bit long though actually

PR #4152, opened 2025-01-17
feat(FN-3675): refactor fee record correction mock builder

RLavender98:
## Introduction :pencil2:
As part of FN-3675 we will introduce completed corrections. To ensure test data consistency and accurate representation of real states of data, it will be useful for the record correction entity mock builder to enforce the specification of 'isCompleted' so that we can create a mock of the correct shape in an easy way.

Since the builder is used in a lot of places putting this up as an initial refactor to keep the changes in the subsequent PR for this ticket not too large. In the next PR there will be logic in the constructor setting other fields based on the values of 'isCompleted' which will look something like this (hence why this refactor needed):
```
if (isCompleted) {
  data.isCompleted = true;
  data.dateReceived = new Date();
  data.bankCommentary = 'some commentary';

  data.previousValues = somePreviousValues;
  data.correctedValues = someCorrectedValues;
} else {
  data.isCompleted = false;
  data.dateReceived = null;
  data.bankCommentary = null;

  data.previousValues = null;
  data.correctedValues = null;
}
```

- Replace 'forFeeRecord' with 'forFeeRecordAndIsCompleted'
- Replace public constructor that doesn't require specification of 'isCompleted' with public static method 'forIsCompleted'
- Remove 'withIsCompleted' since this field is now specified up front


libs/common/src/types/record-correction-values.ts
```diff
@@ -0,0 +1,8 @@
+import { Currency } from './currency';
+
+export type RecordCorrectionValues = {
```
sw-mattw, line 3: Assuming this going to be used for the entity changes in the subsequent PR

sw-mattw (approved):

libs/common/src/types/record-correction-values.ts
RLavender98: Ah yes left in this PR by accident - it will be used though

libs/common/src/test-helpers/mock-data/fee-record-correction.entity.mock-builder.ts
```diff
@@ -6,11 +6,15 @@ import { RecordCorrectionReason, RequestedByUser } from '../../types';
...
   private readonly correction: FeeRecordCorrectionEntity;
 
-  public constructor(correction?: FeeRecordCorrectionEntity) {
-    this.correction = correction ?? FeeRecordCorrectionEntityMockBuilder.forFeeRecord(new FeeRecordEntityMockBuilder().build()).correction;
+  private constructor(correction: FeeRecordCorrectionEntity) {
+    this.correction = correction;
   }
 
-  public static forFeeRecord(feeRecord: FeeRecordEntity): FeeRecordCorrectionEntityMockBuilder {
+  public static forIsCompleted(isCompleted: boolean): FeeRecordCorrectionEntityMockBuilder {
```
Zainzzkk, line 13: this name doesn't make full sense? should be isCompleted?

RLavender98: The preoperty on the entity is called 'isCompleted' and then the convention we have is an instantiation method starts with 'for' followed by the name of the property and then a method overwriting the default value the 'for' method sets for a specific property has the convention of starting with 'with'.

I agree this appears a little weird, and we could have an alternative convention for boolean fields if we wanted, but I don't think it should be just the property name, and I can't think of anything better than 'for' personally.

RLavender98: This is always called as follows: `FeeRecordCorrectionEntityMockBuild.forIsCompleted(true)....` or `FeeRecordCorrectionEntityMockBuild.forIsCompleted(false)....` which I do think reads ok?

Zainzzkk: I think they should all be changed at a later stage

Zainzzkk (approved):

sw-mattw (approved):

PR #4159, opened 2025-01-21
fix(DTFS2-7771): added date helper functions

abhi-markan:
## Introduction :pencil2:
`security.txt` expiry date should be set to 1 year in future to conform with [RFC-9116](https://www.rfc-editor.org/rfc/rfc9116).

* Added `addYear` helper function.
* Updated API test case assertion.

* Added `getEpochMs` helper function to get EPOCH with ms.
* Added unit test cases.

### DIFF
--- a/libs/common/src/helpers/date.test.ts
+++ b/libs/common/src/helpers/date.test.ts
@@ -76,6 +76,67 @@ describe('date helpers', () => {
       expect(result).toContain('T');
       expect(result).toContain('Z');
     });
+
+    it('should return the current date and time in ISO 8601 format', () => {
+      // Act
+      const result = getISO8601();
+      const year = now().getFullYear();
+      const month = (now().getMonth() + 1).toString().padStart(2, '0');
+      const date = now().getDate();
+
+      // Assert
+      expect(result).toContain(`${year}-${month}-${date}`);
+      expect(result).toContain('T');
+      expect(result).toContain('Z');
+    });
+
+    it('should return the provided date in ISO 8601 format', () => {
+      // Arrange
+      const date = new Date('2023-01-01T00:00:00Z');
+      const expected = date.toISOString();
+
+      // Act
+      const result = getISO8601(date);
+
+      // Assert
+      expect(result).toEqual(expected);
+    });
+
+    it('should handle leap year dates correctly', () => {
+      // Arrange
+      const date = new Date('2020-02-29T00:00:00Z');
+      const expected = date.toISOString();
+
+      // Act
+      const result = getISO8601(date);
+
+      // Assert
+      expect(result).toEqual(expected);
+    });
+
+    it('should handle dates before 1970 correctly', () => {
+      // Arrange
+      const date = new Date('1969-12-31T23:59:59Z');
+      const expected = date.toISOString();
+
+      // Act
+      const result = getISO8601(date);
+
+      // Assert
+      expect(result).toEqual(expected);
+    });
+
+    it('should handle future dates correctly', () => {
+      // Arrange
+      const date = new Date('2100-01-01T00:00:00Z');
+      const expected = date.toISOString();
+
+      // Act
+      const result = getISO8601(date);
+
+      // Assert
+      expect(result).toEqual(expected);
+    });
   });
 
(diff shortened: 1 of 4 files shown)


rkontogianni (dismissed):

portal/api-tests/static-routes.api-test.js
```diff
@@ -39,10 +39,12 @@ describe('/.well-known/security.txt', () => {
       // Act
       const response = await get('/.well-known/security.txt');
       // Only get the YYYY-MM-DD
-      const todayDate = getISO8601().split('T')[0];
+      const addOneYear = addYear(1);
```
BethThomas141, line 42: Would this be clearer called `oneYearInFutureDate`?

portal/api-tests/static-routes.api-test.js
```diff
@@ -39,10 +39,12 @@ describe('/.well-known/security.txt', () => {
...
+      const expiry = getISO8601(addOneYear);
+      const oneYearFromToday = expiry.split('T')[0];
```
BethThomas141, line 44: I'm slightly confused as in the `static.js` file we're displaying the date as an ISO string, whereas in this test file we seem to just be displaying YYYY-MM-DD - is there a reason for the difference?

Either way, this could also be written as:

```
const oneYearInFutureDate = addYear(1);
const oneYearInFutureFormatted = format(oneYearInFutureDate, "yyyy-MM-dd")
```

portal/server/routes/static.js
```diff
@@ -1,15 +1,17 @@
 import express from 'express';
-import { getISO8601 } from '@ukef/dtfs2-common';
+import { addYear, getISO8601 } from '@ukef/dtfs2-common';
 
 const router = express.Router();
+const addOneYear = addYear(1);
+const expiry = getISO8601(addOneYear);
```
BethThomas141, line 6: Can we rename these to: 
```
const oneYearInFutureDate = addYear(1);
const expiryDateIsoString = getISO8601(addOneYear)
```

libs/common/src/helpers/date.test.ts
```diff
@@ -91,4 +91,73 @@ describe('date helpers', () => {
       expect(result).toEqual(expected);
     });
   });
+
+  describe('addYear', () => {
+    it('should add one year provided date object', () => {
```
BethThomas141, line 96:
```suggestion
    it('should add one year to the provided date object', () => {
```

libs/common/src/helpers/date.test.ts
```diff
@@ -91,4 +91,73 @@ describe('date helpers', () => {
...
+
+      const result = addYear(year, date);
+
+      expect(result.getFullYear()).toBe(tenYears);
+    });
+
+    it('should add two years to now', () => {
+      const year = 2;
+      const date = now();
+      const twoYears = date.getFullYear() + year;
```
BethThomas141, line 119: I think the names for the consts are a tad confusing, could it instead be: 

const yearsToAdd = 2
const currentDate = now()
const twoYearsInFutureYear = date.getFullYear() + year

We could also have it in a for.each loop where we test the multiple test cases with different years (e.g. 1, 2 and 10) if we want to be more succinct?

libs/common/src/helpers/date.test.ts
```diff
@@ -91,4 +91,73 @@ describe('date helpers', () => {
...
+
+  describe('addYear', () => {
+    it('should add one year provided date object', () => {
+      const year = 1;
+      const date = new Date('20-09-1989');
+      const oneYear = date.getFullYear() + year;
+
+      const result = addYear(year, date);
+
+      expect(result.getFullYear()).toBe(oneYear);
```
BethThomas141, line 103: I think this isn't a full unit test - it's just testing that the year of the result is the same as the expected year - we also want to check that the day/month is correct (or that the full date object is as expected).

BethThomas141 (changes requested):
Just a few comments!

BethThomas141 (approved):

BethThomas141 (dismissed):

rkontogianni (dismissed):

BethThomas141 (approved):

PR #4164, opened 2025-01-22
feat(FN-3611): browser back button behaviour after correction request submitted

Zainzzkk:
## Introduction :pencil2:
This PR handles the back button behaviour after the correction request is submitted.  Previously a cached version of the page was loaded without calling the relevant get controller for check-the-information.  This PR handles that by setting 'Cache-Control' to 'no-store' and redirecting to page-not-found if submitted already

* Added errorKey to be returned from [get-fee-record-correction-request-review.controller/index.ts](https://github.com/UK-Export-Finance/dtfs2/compare/feat/FN-3611?expand=1#diff-0aa44088f416e88719f74f0c1836460503422c302e6f22f972f2c0786009a8c1) if status set to PENDING_CORRECTION
* Added check to [record-corrections/check-the-information/index.ts](https://github.com/UK-Export-Finance/dtfs2/compare/feat/FN-3611?expand=1#diff-7712d01eb936b329da477b6edf185c60a1715f0f2ee836b63cab217fdd4a7cd7) if errorKey is returned from api call, then to redirect to page-not-found
* Added helper to check if errorKey is set to INVALID
* Added helper to check that responseBody is correct type
* Added middleware to set `Cache-Control` to `no-store`
* Added middleware to `create-record-correction-request` and `check-the-information` get routes only
* Updated and added tests
* Added e2e tests


trade-finance-manager-api/src/v1/controllers/utilisation-reports/fee-record-correction/get-fee-record-correction-request-review.controller.test.ts
```diff
@@ -20,20 +20,7 @@ describe('get-fee-record-correction-request-review.controller', () => {
...
-      },
-      correctionRequestDetails: {
-        facilityId: '0012345678',
-        exporter: 'A sample exporter',
-        reasons: [RECORD_CORRECTION_REASON.FACILITY_ID_INCORRECT],
-        additionalInfo: 'this is the reason',
-        contactEmailAddresses: ['test@test.com'],
-      },
-    });
+    const aResponseBody = (): FeeRecordCorrectionRequestReviewResponseBody => feeRecordCorrectionRequestReviewResponseBodyMock;
```
RLavender98, line 23: Please so people don't try and modify object used across multiple tests
```suggestion
    const aResponseBody = (): FeeRecordCorrectionRequestReviewResponseBody => { ...feeRecordCorrectionRequestReviewResponseBodyMock };
```

trade-finance-manager-ui/server/api-response-types/fee-record-correction-request-review-response-body.ts
```diff
@@ -11,3 +11,9 @@ export type FeeRecordCorrectionRequestReviewResponseBody = {
     contactEmailAddresses: string[];
   };
 };
+
+export type FeeRecordCorrectionRequestReviewErrorKey = {
+  errorKey: string;
+};
```
RLavender98, line 17:
```suggestion
export type FeeRecordCorrectionRequestReviewErrorKeyResponse = {
  errorKey: string;
};
```

trade-finance-manager-ui/server/controllers/utilisation-reports/helpers/record-correction-request-already-submitted.test.ts
```diff
@@ -0,0 +1,28 @@
...
+  describe('when errorKey with a different status is provided', () => {
+    it('should return false', () => {
+      const result = recordCorrectionRequestAlreadySubmitted({ errorKey: 'abc' });
+
+      expect(result).toEqual(false);
+    });
+  });
+
+  describe(`when errorKey with status ${STATUS.INVALID} is provided`, () => {
+    it('should return false', () => {
```
RLavender98, line 22:
```suggestion
    it('should return true', () => {
```

trade-finance-manager-ui/server/controllers/utilisation-reports/record-corrections/check-the-information/index.ts
```diff
@@ -15,7 +16,29 @@ export const getRecordCorrectionRequestInformation = async (req: Request, res: R
...
+
+    /**
+     * if record correction is already submitted
+     * should render page-not-found with custom reason
+     */
+    if (recordCorrectionRequestAlreadySubmitted(bankCorrectionRequestReviewResponse)) {
+      return res.render('utilisation-reports/page-not-found.njk', {
+        reason: 'The record correction request has been sent to the bank. You cannot make any changes to the request',
+        reportId,
+      });
```
RLavender98, line 29: We have been using `ViewModel` types to define the shape of render params for different templates. It is good for type safety and understanding the interface between controllers and templates. I think it would be nice to have one for this.

libs/common/src/constants/status.ts
```diff
@@ -0,0 +1,3 @@
+export const STATUS = {
+  INVALID: 'INVALID_STATUS',
+};
```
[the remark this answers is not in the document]
Zainzzkk: I was thinking this would be a generic one where if anything else needed an invalid status, it would use this const which was why the name is so generic

dtfs-central-api/src/v1/controllers/utilisation-report-service/fee-record-correction/get-fee-record-correction-request-review.controller/index.test.ts
```diff
@@ -249,6 +261,42 @@ describe('get-fee-record-correction-request-review.controller', () => {
...
+          const { req, res } = getHttpMocks();
+
+          // Act
+          await getFeeRecordCorrectionRequestReview(req, res);
+
+          // Assert
+          expect(res._getStatusCode()).toEqual(HttpStatusCode.Ok);
+        });
+
+        it('should respond with the correction request details', async () => {
```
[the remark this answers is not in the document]
Zainzzkk: whoops changed

trade-finance-manager-api/src/v1/controllers/utilisation-reports/fee-record-correction/get-fee-record-correction-request-review.controller.test.ts
Zainzzkk: done

trade-finance-manager-ui/server/api-response-types/fee-record-correction-request-review-response-body.ts
Zainzzkk: done

trade-finance-manager-ui/server/controllers/utilisation-reports/record-corrections/check-the-information/index.ts
Zainzzkk: done

libs/common/src/constants/status.ts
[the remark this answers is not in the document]
RLavender98: I think that is harder to maintain since say multiple endpoints start using the same const and then more values get added and they are all just typed as being allowed to take any value from the STATUS const then it is not obvious which values are possible and which need to be handled by each endpoint.

Also if anyone ever moves a pre-existing `STATUS` const to libs/common then there will be an overlap and they will have to do work to either merge or change the names, perhaps on code they are not so familiar with and don't understand the purpose of because the naming is vague.

trade-finance-manager-ui/server/types/view-models/problem-with-service-view-model.ts
```diff
@@ -0,0 +1,4 @@
+export type ProblemWithServiceViewModel = {
```
RLavender98, line 1: Should we add the user to this and pass in to the template so they still have the username and sign out options in the top bar?

dtfs-central-api/src/v1/routes/utilisation-reports-routes.js
```diff
@@ -1063,9 +1063,9 @@ utilisationReportsRouter
  * @openapi
  * /utilisation-reports/:reportId/fee-records/:feeRecordId/correction-request-review/:userId:
  *   get:
- *     summary: Get correction request review information to check before sending
+ *     summary: Get correction request review information to check before sending or error key if already submitted
  *     tags: [Utilisation Report]
- *     description: Get correction request review information to check before sending
+ *     description: Get correction request review information to check before sending or error key if already submitted
```
RLavender98, line 1068: Can't comment on the line itself but we need to update the response schema on lines 1093-1095

RLavender98 (changes requested):
Just a few small changes still needed

dtfs-central-api/src/v1/controllers/utilisation-report-service/fee-record-correction/get-fee-record-correction-request-review.controller/index.ts
```diff
@@ -17,17 +17,19 @@ export type GetFeeRecordCorrectionRequestReviewRequest = CustomExpressRequest<{
...
+      reportPeriod: ReportPeriod;
+      correctionRequestDetails: {
+        facilityId: string;
+        exporter: string;
+        reasons: RecordCorrectionReason[];
+        additionalInfo: string;
+        contactEmailAddresses: string[];
+      };
+    }
+  | { errorKey: string };
```
RLavender98, line 32: Should we introduce a type that extracts 'INVALID_STATUS' from the const and then use that here so only that specific string can be set?

dtfs-central-api/src/v1/routes/utilisation-reports-routes.js
Zainzzkk: removed and added comment in correct place

trade-finance-manager-ui/server/types/view-models/problem-with-service-view-model.ts
Zainzzkk: whoops, added

libs/common/src/constants/status.ts
RLavender98, line 3: If we are keeping this as a generic name then please can we go with the generic naming:
```suggestion
export const ERROR_KEY = {
  INVALID_STATUS: 'INVALID_STATUS',
};
```
`INVALID` isn't actually a status in this context so I don't think STATUS.INVALID is particularly clear

RLavender98 (commented):
Looks good, just need to come to a place of agreement on the naming of the new const

libs/common/src/constants/status.ts
Zainzzkk: this has been changed to INVALID_STATUS

RLavender98 (approved):

abhi-markan (changes requested):

libs/common/src/middleware/set-no-store-cache-control.ts
```diff
@@ -0,0 +1,14 @@
...
+/**
+ * middleware to set Cache-Control to no-store
+ * stops cached version of a page being rendered when going back to a page (using the browser back button)
+ * ensures the get controller for the page is called (if cached, get controller for a page is not called)
+ * @param _req
+ * @param res
+ * @param next
+ */
+export const setNoStoreCacheControl = (_req: Request, res: Response, next: NextFunction) => {
+  res.set('Cache-Control', 'no-store');
```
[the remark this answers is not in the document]
Zainzzkk: done

libs/common/src/middleware/set-no-store-cache-control.ts
```diff
@@ -0,0 +1,14 @@
+import { Request, Response, NextFunction } from 'express';
```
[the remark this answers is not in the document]
Zainzzkk: added

libs/common/src/middleware/set-cache-control.test.ts
```diff
@@ -0,0 +1,21 @@
...
+afterEach(() => {
+  jest.resetAllMocks();
+});
+
+describe('set-no-store-cache-control', () => {
+  it('should set Cache-Control headers correctly', async () => {
+    const response = await request(app).get('/');
+
+    expect(response.headers['cache-control']).toBe('no-cache, no-store, must-revalidate, max-age=0');
+    expect(response.status).toBe(HttpStatusCode.Ok);
```
RLavender98, line 19: Also I think we should be testing that `next` gets called rather than asserting on the status no?

libs/common/src/middleware/set-cache-control.ts
```diff
@@ -0,0 +1,15 @@
...
+
+/**
+ * middleware to set Cache-Control to no-store, no-cache, must-revalidate and max-age=0
+ * stops cached version of a page being rendered when going back to a page (using the browser back button)
+ * ensures the get controller for the page is called
+ * @param _req
+ * @param res
+ * @param next
+ */
+export const setCacheControl = (_req: Request, res: Response, next: NextFunction) => {
```
RLavender98, line 11: Can we change this back to the old name? This is less descriptive and doesn't tell the reader the behaviour that is being set

libs/common/src/middleware/set-no-store-cache-control.test.ts
```diff
@@ -0,0 +1,31 @@
...
+  it('should set Cache-Control headers correctly', async () => {
+    const middleware = (req: Request, res: Response, next: NextFunction) => {
+      setNoStoreCacheControl(req, res, next);
+    };
+
+    const next = jest.fn((_req: Request, res: Response) => res.send());
+
+    const testApp = express();
+    testApp.use(middleware);
+    testApp.get('/', next);
```
RLavender98, line 23: We want to test that the middleware itself is calling next, not the route controller. Would this work?
```suggestion
    testApp.get('/', middleware);
```

RLavender98 (approved):

abhi-markan (approved):

PR #4166, opened 2025-01-22
feat(dtfs2-6892): fixes to sso following local running

AlexBramhill:
## Introduction :pencil2:
Following running locally, there's a few fixed that were needed to make sure SSO worked as intended

- Add ff to tfm-api docker compose
- update entraId to represent ukef data format (removal of secondary email)
- update autosubmit hash, typo
- update routing
e2e tests will come in another review


docker-compose.yml
```diff
@@ -182,6 +182,7 @@ services:
       - ENTRA_ID_TENANT_ID
       - ENTRA_ID_CLIENT_SECRET
       - ENTRA_ID_REDIRECT_URL
+      - FF_TFM_SSO_ENABLED
```
abhi-markan, line 185: Can you add same to `.env.sample` please.

trade-finance-manager-api/src/v1/repo/user.repo.ts
```diff
@@ -45,6 +45,21 @@ export class UserRepo {
...
+   * @returns the matched TFM user or null if not found
+   */
+  public static async findUserByEmailAddress(email: string): Promise<TfmUser | null> {
+    const collection = await UserRepo.getCollection();
+
+    const emailRegex = getEscapedRegexFromString(email);
+
+    const query = { email: emailRegex };
+
+    return await collection.findOne(query);
```
abhi-markan, line 60:
```suggestion
    return collection.findOne(query);
```

trade-finance-manager-ui/server/generateApp.js
```diff
@@ -54,21 +54,33 @@ const generateApp = () => {
...
+    }),
+  );
+
+  app.use('/assets', express.static('node_modules/govuk-frontend/dist/govuk/assets'), express.static(path.join(__dirname, '..', 'public')));
+
+  app.use(createRateLimit());
+
+  // We add a conditional check here as there are no auth routes for the non sso journey, and
+  // we cannot call app.use with './', undefined.s
+  if (unauthenticatedAuthRoutes) {
```
abhi-markan, line 77: Can this be FF under SSO, wondering once SSO is enabled we will not need this?

abhi-markan (changes requested):

docker-compose.yml
AlexBramhill: It's already there

trade-finance-manager-ui/server/generateApp.js
AlexBramhill: This is featureflagged under SSO through how we gather these routes -- see `~/trade-finance-manager-ui/server/routes/auth/configs/index.ts` for an example of how we do this.

We do it this way as we sometimes want to reassign certain routes -- so we need to get these routes based on an env var

trade-finance-manager-ui/server/routes/auth/configs/unauthenticated-auth-sso.ts
```diff
@@ -1,13 +1,25 @@
...
 export const getUnauthenticatedAuthSsoRouter: GetRouter = () => {
   const unauthenticatedAuthSsoController = new UnauthenticatedAuthController();
-
   const unauthenticatedAuthSsoRouter = express.Router();
 
-  unauthenticatedAuthSsoRouter.post('auth/accept-sso-redirect', unauthenticatedAuthSsoController.postSsoRedirect.bind(unauthenticatedAuthSsoController));
+  const loginService = new LoginService();
+  const userSessionService = new UserSessionService();
+  const loginController = new LoginController({ loginService, userSessionService });
+
```
vamshidharukef, line 15: Please add below for better understanding
// Route for SSO redirect

trade-finance-manager-ui/server/routes/auth/configs/unauthenticated-auth-sso.ts
```diff
@@ -1,13 +1,25 @@
...
+  unauthenticatedAuthSsoRouter.post('/auth/sso-redirect', (req, res) => {
+    unauthenticatedAuthSsoController.postSsoRedirect(req, res);
+  });
+
```
vamshidharukef, line 19: Please add below for better understanding 
// Route for handling SSO redirect form

trade-finance-manager-ui/server/generateApp.js
```diff
@@ -54,21 +54,33 @@ const generateApp = () => {
...
+    morgan('dev', {
+      skip: (req) => req.url.startsWith('/assets') || req.url.startsWith('/main.js'),
+    }),
+  );
+
+  app.use('/assets', express.static('node_modules/govuk-frontend/dist/govuk/assets'), express.static(path.join(__dirname, '..', 'public')));
+
+  app.use(createRateLimit());
+
+  // We add a conditional check here as there are no auth routes for the non sso journey, and
```
vamshidharukef, line 75: Can you re-phrase to 
// Conditionally add unauthenticated auth routes if they exist

trade-finance-manager-ui/server/generateApp.js
vamshidharukef, line 1: Can the comment on line 99 be re-phrased to below
// Catch-all error handler for CSRF token errors and other unexpected errors

vamshidharukef, line 1: It would be helpful to add a brief comment explaining what the generateApp function does and what it returns. Something like 
/**
 * Generates the Express app with middleware and routes set up
 * @returns {Express} The configured Express app
 */

AlexBramhill: Agree, but beyond the scope of this ticket as it'll have to be added to GEF/Portal too

trade-finance-manager-ui/server/routes/auth/configs/unauthenticated-auth-sso.ts
AlexBramhill: I'm not sure these are needed here -- The names of these functions define what the suggested comments do. I'll add a comment to `postSsoRedirect` however, so we have a comment much like `handleSsoRedirectForm`

trade-finance-manager-ui/server/generateApp.js
AlexBramhill: I've updated this a bit more to be a bit more descriptive of why we're doing what we're doing -- I think the code is documenting the `what we're doing`

rkontogianni (dismissed):

trade-finance-manager-api/src/v1/repo/user.repo.ts
```diff
@@ -45,6 +45,21 @@ export class UserRepo {
     return await collection.find(query).toArray();
   }
 
+  /**
+   * Finds a user by email address
+   * @param email email address to search for
+   * @returns the matched TFM user or null if not found
+   */
+  public static async findUserByEmailAddress(email: string): Promise<TfmUser | null> {
+    const collection = await UserRepo.getCollection();
```
abhi-markan, line 54: try-catch please.

abhi-markan (changes requested):

trade-finance-manager-api/src/v1/repo/user.repo.ts
AlexBramhill: We catch these errors at the controller level -- We want this to throw an error if something goes wrong (it returns null if user is not found, but if it throws that means something more sinister has happened, and we handle that at the controller level, logging and sending the appropriate response)

abhi-markan (approved):

abhi-markan (approved):

abhi-markan (approved):

rkontogianni (approved):

PR #4170, opened 2025-01-23
feat(DTFS2-7620): add e2e tests to amendments FED flow

BethThomas141:
## Introduction :pencil2:
- Add E2E tests that cover the 'facility end date' and 'do you have a facility end date?' page

- Add to the amendments E2E tests

### DIFF
--- a/e2e-tests/ukef/cypress/e2e/journeys/portal/amendments/change-cover-end-date-journey.happy-path.spec.js
+++ b/e2e-tests/ukef/cypress/e2e/journeys/portal/amendments/change-cover-end-date-journey.happy-path.spec.js
@@ -64,8 +64,7 @@ context('Change cover end date journey - happy path', () => {
   });
 
   beforeEach(() => {
-    cy.clearSessionCookies();
-    cy.login(BANK1_MAKER1);
+    cy.saveSession();
   });
 
   it('should navigate through the journey correctly', () => {
--- a/e2e-tests/ukef/cypress/e2e/journeys/portal/amendments/change-cover-end-date-journey.happy-path.spec.js
+++ b/e2e-tests/ukef/cypress/e2e/journeys/portal/amendments/change-cover-end-date-journey.happy-path.spec.js
@@ -64,7 +64,8 @@ context('Change cover end date journey - happy path', () => {
   });
 
   beforeEach(() => {
-    cy.saveSession();
+    cy.clearSessionCookies();
+    cy.login(BANK1_MAKER1);
   });
 
   it('should navigate through the journey correctly', () => {
--- a/e2e-tests/ukef/cypress/e2e/journeys/portal/amendments/change-facility-end-date-journey.unhappy-path.spec.js
+++ b/e2e-tests/ukef/cypress/e2e/journeys/portal/amendments/change-facility-end-date-journey.unhappy-path.spec.js
@@ -61,7 +61,8 @@ context('Amendments - change facility end date - unhappy path', () => {
   });
 
   beforeEach(() => {
-    cy.saveSession();
+    cy.clearSessionCookies();
+    cy.login(BANK1_MAKER1);
   });
 
   it('should render an error if nothing is selected to change', () => {
@@ -76,7 +77,7 @@ context('Amendments - change facility end date - unhappy path', () => {
     whatDoYouNeedToChange.amendmentOptionsInlineError().contains('Select if you need to change the facility cover end date, value or both');
   });
 
-  it('should navigate to the "Do you have a facility end date?" page', () => {
+  it('should render an error if nothing is selected on the "Do you have a facility end date?" page', () => {
     cy.visit(relative(`/gef/application-details/${dealId}/facilities/${facilityId}/amendments/${amendmentId}/what-do-you-need-to-change`));
 
     whatDoYouNeedToChange.coverEndDateCheckbox().click();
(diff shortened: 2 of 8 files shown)


e2e-tests/ukef/cypress/e2e/journeys/portal/amendments/change-cover-end-date-journey.happy-path.spec.js
```diff
@@ -64,8 +64,7 @@ context('Change cover end date journey - happy path', () => {
   });
 
   beforeEach(() => {
-    cy.clearSessionCookies();
-    cy.login(BANK1_MAKER1);
+    cy.saveSession();
```
BethThomas141, line 67: This means that each `it` case continues where it left of in the previous test (instead of having to log in the user etc. each time) so it's more reflective of a real user journey

e2e-tests/ukef/cypress/e2e/journeys/portal/amendments/change-facility-end-date-journey.unhappy-path.spec.js
```diff
@@ -0,0 +1,162 @@
...
+    whatDoYouNeedToChange.coverEndDateCheckbox().click();
+    cy.clickContinueButton();
+
+    cy.url().should('eq', relative(`/gef/application-details/${dealId}/facilities/${facilityId}/amendments/${amendmentId}/cover-end-date`));
+
+    cy.completeDateFormFields({ idPrefix: 'cover-end-date' });
+    cy.clickContinueButton();
+
+    cy.url().should('eq', relative(`/gef/application-details/${dealId}/facilities/${facilityId}/amendments/${amendmentId}/do-you-have-a-facility-end-date`));
+  });
```
rkontogianni, line 91: @BethThomas141 I think this is not considered as unhappy path. I had the same in my PR, but then I removed it

e2e-tests/ukef/cypress/e2e/journeys/portal/amendments/change-facility-end-date-journey.unhappy-path.spec.js
```diff
@@ -0,0 +1,162 @@
...
+      expectedErrorMessage: 'Facility end date cannot be greater than 6 years in the future',
+    },
+    {
+      description: 'the facility end date is before the cover start date',
+      dateFieldInput: { date: sub(now(), { days: 2 }) },
+      expectedErrorMessage: 'Facility end date cannot be before the cover start date',
+    },
+  ];
+
+  facilityEndDateErrorTestCases.forEach(({ description, dateFieldInput, expectedErrorMessage }) => {
```
rkontogianni, line 148: @BethThomas141 this is really nice!

e2e-tests/ukef/cypress/e2e/journeys/portal/amendments/change-facility-end-date-journey.unhappy-path.spec.js
BethThomas141: Ah, you are correct, but the test description here just tells us that we're clicking onto the next page. So while it's not an unhappy path on it's own, it's a needed step to test the unhappy paths (we need to progress to the next page). The file title of 'unhappy-path' is more reflective of the entire journey across the file, which tests all the possible unhappy cases we could come across. I think it makes the test clearer if we include this step, so when reading through we know we've gone to the next page. What do you think? :)

e2e-tests/ukef/cypress/e2e/journeys/portal/amendments/change-facility-end-date-journey.unhappy-path.spec.js
BethThomas141: Thank you!

e2e-tests/ukef/cypress/e2e/journeys/portal/amendments/change-facility-end-date-journey.unhappy-path.spec.js
rkontogianni: @BethThomas141 My opinion is that each it() block should be an independent assertion and not depend on the previous steps. However, if you check the checkbox once, it will remain checked for all subsequent tests (I am not sure if that is a behavior we expect to occur). 
you can check once the checkbox in the beforeEach and then focus on testing only the facility end date. The what-do-you-need-to-change page will be tested separately in another test. For the cover end date tests, I just combined this test to the next within the same it() blocks, but I don't think that this was the ideal choice

utils/mock-data-loader/gef/eligibilityCriteriaAmendments.js
```diff
@@ -13,7 +13,7 @@ const ELIGIBILITY_CRITERIA_AMENDMENTS = [
       },
       {
         id: 2,
-        text: 'Neither the Exporter, not its UK Parent Obligor is an Affected Person',
+        text: 'Neither the Exporter, nor its UK Parent Obligor is an Affected Person',
```
BethThomas141, line 16: This is just a quick typo fix that I'm putting through in this PR :)

e2e-tests/ukef/cypress/e2e/journeys/portal/amendments/change-facility-end-date-journey.unhappy-path.spec.js
```diff
@@ -0,0 +1,162 @@
...
+  });
+
+  it('should navigate to the facility end date page', () => {
+    cy.visit(relative(`/gef/application-details/${dealId}/facilities/${facilityId}/amendments/${amendmentId}/do-you-have-a-facility-end-date`));
+
+    doYouHaveAFacilityEndDate.yesRadioButton().click();
+    cy.clickContinueButton();
+
+    cy.url().should('eq', relative(`/gef/application-details/${dealId}/facilities/${facilityId}/amendments/${amendmentId}/facility-end-date`));
+  });
```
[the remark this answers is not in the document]
BethThomas141: Now changed, thanks

rkontogianni (approved):

PR #4173, opened 2025-01-23
feat(FN-3163): create new dtfs customers in salesforce

natdeanlewissoftwire:
## Introduction :pencil2:
As part of the CCM project, we need to automatically create in Salesforce an exporter on a DTFS deal that doesn't yet exist in Salesforce. There are accompanying changes to APIM that allow this. DTFS just needs to call the correct APIM endpoint (feature flagged in this PR) to get an existing customer and create it in Salesforce if it doesn't exist. To achieve this, we pass the customer name, registration number and probability of default as that's all we need to create the customer in Salesforce (the rest of the data we need is either inferred from the Dun & Bradstreet API, NUMGEN, or defaulted (in the case of risk data)).

This PR is a feature branch merge of previously approved changes to accomplish the above.
Also includes one small tweak to return a 400 rather than a 500 when company registration number is null in getOrCreateParty.


.env.sample
```diff
@@ -119,4 +119,5 @@ FF_TFM_DEAL_CANCELLATION_ENABLED=false
 FF_PORTAL_FACILITY_AMENDMENTS_ENABLED=false
 FF_TFM_SSO_ENABLED=false
 FF_FEE_RECORD_CORRECTION_ENABLED=false
+AUTOMATIC_SALESFORCE_CUSTOMER_CREATION_ENABLED=false
```
abhi-markan, line 122: I should have picked this before, can this be changed to `FF_SALESFORCE_CUSTOMER_CREATION` please. Apologies.

abhi-markan (changes requested):

abhi-markan (commented):
Would it be possible to add following missing tests please:

1. E2E
2. FF API test

.github/workflows/test.yml
```diff
@@ -847,9 +850,12 @@ jobs:
     needs: [setup]
     env:
       # Various feature flags
+      FF_TFM_PAYMENT_RECONCILIATION_ENABLED: 'true'
       FF_FEE_RECORD_CORRECTION_ENABLED: 'true'
       FF_TFM_DEAL_CANCELLATION_ENABLED: 'true'
       FF_PORTAL_FACILITY_AMENDMENTS_ENABLED: 'true'
+      FF_TFM_SSO_ENABLED: 'false'
```
abhi-markan, line 857: @AlexBramhill I have changed this to `false` as it is failing all the FF E2E tests, you will have to set it to `true` in your final PR please.

abhi-markan (approved):

PR #4174, opened 2025-01-24
feat(DTFS2-7498): add eligibility and bank review date e2e tests

BethThomas141:
## Introduction :pencil2:
Add E2E tests for the eligibility page and bank review date page, as well as refactor the tests to be in line with what we talked about in dev Friday (split into journeys and pages)

- Add to E2E tests and refactor existing ones


e2e-tests/ukef/cypress/e2e/journeys/portal/amendments/change-facility-end-date-journey.happy-path.spec.js
```diff
@@ -105,6 +106,20 @@ context('Amendments - Change facility end date journey - happy path', () => {
...
+    eligibility.allTrueRadioButtons().should('not.be.checked');
+    eligibility.allFalseRadioButtons().should('not.be.checked');
+
+    eligibility.criterionRadiosText(1).contains('The Facility is not an Affected Facility');
+    eligibility.criterionRadiosText(2).contains('Neither the Exporter, nor its UK Parent Obligor is an Affected Person');
+
+    eligibility.allTrueRadioButtons().click({ multiple: true });
+    cy.clickContinueButton();
+
+    cy.url().should('eq', relative(`/gef/application-details/${dealId}/facilities/${facilityId}/amendments/${amendmentId}/effective-date`));
```
BethThomas141, line 122: I've added the 'happy path' tests to the facility value & facility end date test files as the intention of these happy path files is to test the full user journey from start to finish when updating an amendment with, e.g. a facility value & so the 'happy path' for eligibility can live here, instead of a separate file.

rkontogianni: I think you are right here that the happy path files are meant to test the full user journey from start to finish. We could possibly create a single happy path file and define helper functions happyCoverEndDate, happyFacilityEndDate, etc., to encapsulate the different parts of the journey? and then import all these helper functions together in a single change-amendment-journey.happy-path.spec.js file?

e2e-tests/ukef/cypress/e2e/journeys/portal/amendments/journeys/change-both-cover-end-date-and-facility-value.spec.js
```diff
@@ -0,0 +1,139 @@
...
+   */
+  let facilityId;
+  /**
+   * @type {string}
+   */
+  let amendmentId;
+  /**
+   * @type {Date}
+   */
+  const mockFacility = anIssuedCashFacility({ facilityEndDateEnabled: true });
```
abhi-markan, line 32:
```suggestion
  const mockFacility = issuedCashFacility({ facilityEndDateEnabled: true });
```

e2e-tests/ukef/cypress/e2e/journeys/portal/amendments/journeys/change-both-cover-end-date-and-facility-value.spec.js
```diff
@@ -0,0 +1,139 @@
...
+    cy.clearSessionCookies();
+  });
+
+  beforeEach(() => {
+    cy.clearSessionCookies();
+    cy.login(BANK1_MAKER1);
+  });
+
+  it('should navigate through the journey correctly', () => {
+    cy.visit(relative(`/gef/application-details/${dealId}/facilities/${facilityId}/amendments/${amendmentId}/what-do-you-need-to-change`));
```
abhi-markan, line 72: Might be prudent to break this into smaller assertions to ease debuging if the tests fails.

abhi-markan (changes requested):

e2e-tests/ukef/cypress/e2e/journeys/portal/amendments/journeys/change-both-cover-end-date-and-facility-value.spec.js
BethThomas141: This is used in many many places (not named in this file) so not changing here (as it will cause many files changed in this PR).

e2e-tests/ukef/cypress/e2e/journeys/portal/amendments/journeys/change-both-cover-end-date-and-facility-value.spec.js
BethThomas141: I'm keen to keep this as is - my plan is to have long tests within the 'journeys' folder to mimic realistic user journeys. As per the Cypress docs, they are generally fine with large tests: https://docs.cypress.io/app/core-concepts/best-practices#Creating-Tiny-Tests-With-A-Single-Assertion and say it's easy to debug regardless as Cypress tests tell you at which point in the test they fail when you run them.
image

In particular, if we break this down further, it won't be as reflective of a real user journey, especially if we want each `it` block to be able to run independently as we discussed in dev friday. As we'd need each `it` block to repeat the previous `it` block to get the user to the correct place in the journey.

On the other hand, the 'pages' folder will contain more broken down `it` tests. 
Are you happy for me to leave as is?

e2e-tests/ukef/cypress/e2e/journeys/portal/amendments/journeys/change-both-cover-end-date-and-facility-value.spec.js
abhi-markan: Thanks I'll raise a separate PR to rename this.

abhi-markan (dismissed):

abhi-markan (approved):

PR #4175, opened 2025-01-24
fix(DTFS2-7748): created facility stage helper

abhi-markan:
## Introduction :pencil2:
Facility stage on Portal facilities tab does not represent the facility which has been risk expired (due to deal cancellation).

* Introduced a new mapping function `mapFacilityProperties`, responsible for executing various mappers on portal facilities array:
  * Currently limited to `facilityStage` property.
* Introduced a new `FacilityDashboard` type for portal dashboard results pertinent to facilities.

* Minor UI margin fix for a success banner, presented to the user upon deal cancellation.
* Updated success banner class to `margin`, accomponied with an updated test cases.
* Updated `facilityId` to `facilityIdentifier` to triage [facility isuance bug](https://ukef-dtfs.atlassian.net/browse/DTFS2-7804).
* Added additional unit test cases.
* Added durable functions log for a deal and facility ACBS submission.

## Screenshot 📸

image


portal/component-tests/_macros/facility-stage.component-test.js
```diff
@@ -0,0 +1,40 @@
+const { FACILITY_STAGE } = require('@ukef/dtfs2-common');
+const componentRender = require('../componentRenderer');
+
+const component = '_macros/facility-stage.njk';
+const render = componentRender(component);
+const facilities = require('../fixtures/deal-fully-completed');
```
MarRobSoftwire, line 6:
```suggestion
const deal = require('../fixtures/deal-fully-completed');
```

portal/component-tests/_macros/facility-stage.component-test.js
```diff
@@ -0,0 +1,40 @@
...
+
+describe('_macros/facility-stage.njk', () => {
+  it(`should render facility stage as ${FACILITY_STAGE.ISSUED}, where facility has been issued.`, () => {
+    // Act
+    const wrapper = render(facilities.loanTransactions.items[0]);
```
MarRobSoftwire, line 11: I think I'd prefer this to be more explicit, something like

```suggestion
    const wrapper = render({ ...facilities.loanTransactions.items[0], facilityStage: FACILITY_STAGE.ISSUED });
```

MarRobSoftwire: or does it need `hasBeenIssued: true` - it's not clear here whats the significant part

portal/component-tests/_macros/facility-stage.component-test.js
```diff
@@ -0,0 +1,40 @@
...
+
+  it(`should render facility stage as ${FACILITY_STAGE.UNISSUED}, with an empty facility object`, () => {
+    // Act
+    const wrapper = render(facilities.loanTransactions.items[3]);
+
+    // Assert
+    expect(wrapper.html).toContain(FACILITY_STAGE.UNISSUED);
+  });
+
+  it(`should render facility stage as ${FACILITY_STAGE.RISK_EXPIRED}`, () => {
```
MarRobSoftwire, line 33: can you add a `... when facility stage is risk expired` or simialr to description

portal/component-tests/fixtures/deal-fully-completed.js
```diff
@@ -138,6 +138,39 @@ const deal = {
...
+        coveredPercentage: '30',
+        minimumQuarterlyFee: '200',
+        ukefExposure: '3,703,703.40',
+        premiumFrequency: 'Semi-annually',
+        premiumType: 'In advance',
+        dayCountBasis: '365',
+        status: 'Completed',
+      },
+      {
+        _id: '789123',
```
MarRobSoftwire, line 172: should this ID be distinct & a real objectId?

portal/templates/_macros/facility-stage.njk
MarRobSoftwire, line 1: should this be handled in Type/Javascript before passing it into the template?

trade-finance-manager-ui/component-tests/_macros/success-banner.component-test.js
```diff
@@ -12,7 +12,15 @@ describe(component, () => {
     wrapper = render(message);
   });
 
+  it('should exist', () => {
+    wrapper.expectElement('[data-cy="success-banner"]').toExist();
+  });
```
MarRobSoftwire, line 17: Is this not tested when checking the message/class?

trade-finance-manager-ui/templates/_macros/success-banner.njk
MarRobSoftwire, line 1: Could you include a screenshot of this change please :)

portal/component-tests/_macros/facility-stage.component-test.js
abhi-markan: This does not represent the true state, in the code the whole faciliy object is supplied. This is done explicity to keep it future proof.

trade-finance-manager-ui/component-tests/_macros/success-banner.component-test.js
abhi-markan: Not directly.

portal/templates/_macros/facility-stage.njk
```diff
@@ -0,0 +1,22 @@
+{#
+    Macro: render
+    Description: This macro determines and renders the stage of a facility based on its properties.
+    Parameters:
+        - facility: An object representing the facility with the following properties:
+            - facilityStage: A string indicating the current stage of the facility.
+            - hasBeenIssued: A boolean indicating whether the facility has been issued.
+    Returns: A string representing the stage of the facility, which can be 'Risk expired', 'Issued', or 'Unissued'.
+#}
+{% macro render(facility) %}
```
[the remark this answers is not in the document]
AlexBramhill: We also have `map-cash-contingent-facility.js` which further complicates matters with assigning facilityStage in tfm api to 
```

const mapFacilityStage = (hasBeenIssued) => {
  if (hasBeenIssued) {
    return CONSTANTS.FACILITIES.FACILITY_STAGE.ISSUED;
  }

  return CONSTANTS.FACILITIES.FACILITY_STAGE.COMMITMENT;
};
```

while also for gef deals:
```
export const mapGefFacilityStage = (isIssued: boolean, tfmFacilityStage?: TfmFacilityStage): TfmFacilityStage => {
  if (tfmFacilityStage) {
    return tfmFacilityStage;
  }

  return isIssued ? TFM_FACILITY_STAGE.ISSUED : TFM_FACILITY_STAGE.COMMITMENT;
};
```

All these logics are different, which is scary. If these get applied one after each other for any reason (maybe not in one journey, but could be after being transformed/saved to database/pulled out of db/transformed again), results will be different than just applying straight to a database call. It'll also be order specific.

portal-api/src/v1/controllers/facilities.controller.js
```diff
@@ -117,6 +117,8 @@ const queryAllFacilities = async (filters = {}, sort = {}, start = 0, pagesize =
           exporter: '$deal.exporter.companyName',
           // exporter in lowercase for sorting
           lowerExporter: { $toLower: '$deal.exporter.companyName' },
+          // Facility stage - specifically used for Risk expired
+          facilityStage: '$facilityStage',
```
AlexBramhill, line 121: Just a thought -- if we only want devs to use this for risk expired, should we throw an error if someones searching for something other than risk expired to stop misuse?

portal/component-tests/fixtures/deal-fully-completed.js
AlexBramhill: (it shouldn't unless we're mapping it back to objectId after making the API call, which we shouldnt really care about!)

portal/templates/_macros/facility-stage.njk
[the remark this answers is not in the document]
ttbarnes: Not sure how feasible this is - I would prefer nunjucks to simply consume a `facilityStage` variable and have all logic in a helper, called in controllers

trade-finance-manager-ui/component-tests/_macros/success-banner.component-test.js
```diff
@@ -12,7 +12,15 @@ describe(component, () => {
...
+
   it('should render the success banner containing the message', () => {
     wrapper.expectText('[data-cy="success-banner"]').toRead(message);
   });
+
+  it('should render the success banner containing the message, with the correct class', () => {
```
ttbarnes, line 23:
```suggestion
  it('should render the correct success banner class ', () => {
```

portal/templates/_macros/facility-stage.njk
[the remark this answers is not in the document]
abhi-markan: Ideally a help function in a controller, as mentioned above but not without scope creeping this PR (ideally a new ticket on its own as it will effect multiple controllers and micro-services).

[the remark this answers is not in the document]
ttbarnes: I agree @AlexBramhill , I don't think this is a pattern we should introduce - concerned this could be confusing and misleading further down the line.

Fully appreciate this is a larger piece of tech debt we should tackle. However, in this PR, the new macro only seems to be used for the Dashboard facilities table.

Can we instead add or update mapping for this controller? This way we don't have the logic in nunjucks and could be an initial step to applying it for other areas(?)

portal/component-tests/fixtures/deal-fully-completed.js
```diff
@@ -103,7 +104,7 @@ const deal = {
   loanTransactions: {
     items: [
       {
-        _id: '123456',
+        _id: new ObjectId(),
```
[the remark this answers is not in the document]
MarRobSoftwire: Ah sorry, dont think I was clear enough - just thought it would be more accurate to have this mock with a string `abcdef123456abcdef123456`  but yep agree we shouldn't be using `ObjectId`s in the front end

portal/templates/_macros/facility-stage.njk
MarRobSoftwire: I think Alex & Tony's comment in the other thread summarises my thoughts - this doesn't really feel like logic nunjucks should be handling

portal/component-tests/_macros/facility-stage.component-test.js
MarRobSoftwire: I think the entire true state can be passed in but its hard from looking at the tests to determine which is the significant value - is it `hasBeenIssued` or is it `facilityStage`. If this is clearly shown in the mock data it will be easier to maintain this test

portal/server/helpers/map-facility-properties.ts
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * The function checks the `facilityStage` property of the input `facility`.
+ * If the `facilityStage` is `FACILITY_STAGE.RISK_EXPIRED`, it sets the `facilityStage`
+ * of the returned facility to `FACILITY_STAGE.RISK_EXPIRED`.
+ * Otherwise, it sets the `facilityStage` based on the `hasBeenIssued` property:
+ * - If `hasBeenIssued` is `true`, it sets the `facilityStage` to `FACILITY_STAGE.ISSUED`.
+ * - If `hasBeenIssued` is `false`, it sets the `facilityStage` to `FACILITY_STAGE.UNISSUED`.
+ */
+const mapFacilityStage = (facility: FacilityDashboard): FacilityDashboard => {
+  const mappedFacility = facility;
```
MarRobSoftwire, line 17: Should clone this to avoid modifying `facility` object
```suggestion
  const mappedFacility = cloneDeep(facility);
```

libs/common/src/types/mongo-db-models/facility.ts
```diff
@@ -74,3 +75,23 @@ export type Facility = {
   previousStatus?: string; // BSS/EWCS facilities only
   facilityStage?: string;
 } & FacilityEndDateProperties;
+
+/**
+ * Portal dashboard facilities
+ */
+export type FacilityDashboard = {
```
MarRobSoftwire, line 82: This shouldnt be in the `mongo-db-models` directory -  it's probably specific enough to `portal` that it should live in the front end rather than `libs/common`

libs/common/src/types/mongo-db-models/facility.ts
```diff
@@ -74,3 +75,23 @@ export type Facility = {
...
+  _id: string | ObjectId;
```
MarRobSoftwire, line 83: things in portal will never be an `ObjectId`

libs/common/src/types/mongo-db-models/facility.ts
```diff
@@ -74,3 +75,23 @@ export type Facility = {
...
+  dealId: string | ObjectId;
+  name: string | null;
+  currency: { id: Currency } | null;
+  value: number | string;
```
MarRobSoftwire, line 87: How come this is `number | string` - is it not guaranteed to be one or the other?

portal/component-tests/fixtures/deal-fully-completed.js
```diff
@@ -136,8 +136,41 @@ const deal = {
...
       },
       {
-        _id: '789123',
+        _id: '6597dffeb5ef5ff4267e5045',
         facilityStage: 'Unconditional',
+        hasBeenIssued: false,
+        'requestedCoverStartDate-day': `${format(now, 'dd')}`,
+        'requestedCoverStartDate-month': `${format(now, 'MM')}`,
+        'requestedCoverStartDate-year': `${format(now, 'yyyy')}`,
+        'coverEndDate-day': `${format(nextMonth, 'dd')}`,
```
natdeanlewissoftwire, line 145: any reason `requestedCoverStart...` and `coverEnd...` aren't named consistently?

libs/common/src/types/mongo-db-models/facility.ts
```diff
@@ -74,3 +75,23 @@ export type Facility = {
...
+  type: FacilityType;
+  submissionType?: DealSubmissionType;
+  hasBeenIssued: boolean;
+  updatedAt: UnixTimestamp;
```
MarRobSoftwire, line 91: Is this seconds or miliseconds? Could you update to the specific type

portal/server/helpers/map-facility-properties.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  if (facility.facilityStage === FACILITY_STAGE.RISK_EXPIRED) {
+    mappedFacility.facilityStage = FACILITY_STAGE.RISK_EXPIRED;
```
natdeanlewissoftwire, line 20: This sets a value to what it already is, can be removed.

libs/common/src/types/mongo-db-models/facility.ts
```diff
@@ -74,3 +75,23 @@ export type Facility = {
...
+  lowerExporter: string;
+  ukefFacilityId?: string;
+  submittedAsIssuedDate?: UnixTimestampString | null;
+  exporter?: string;
+  facilityStage?: string;
```
MarRobSoftwire, line 96: is there a facility stage type? If not could you add one equal to `ValuesOf<typeof FACILITY_STAGE>`

portal/server/helpers/map-facility-properties.ts
```diff
@@ -0,0 +1,38 @@
+import { FacilityDashboard, FACILITY_STAGE } from '@ukef/dtfs2-common';
+
+/**
+ * Maps the facility stage based on the current facility properties.
+ *
+ * @param {Facility} facility - The facility object to map the stage for.
+ * @returns {Facility} The facility object with the mapped stage.
```
MarRobSoftwire, line 7: This contradicts the other type & its also not importing correctly
```suggestion
 * @param facility - The facility object to map the stage for.
 * @returnsThe facility object with the mapped stage.
```

portal/server/helpers/map-facility-properties.ts
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * The function checks the `facilityStage` property of the input `facility`.
+ * If the `facilityStage` is `FACILITY_STAGE.RISK_EXPIRED`, it sets the `facilityStage`
+ * of the returned facility to `FACILITY_STAGE.RISK_EXPIRED`.
+ * Otherwise, it sets the `facilityStage` based on the `hasBeenIssued` property:
+ * - If `hasBeenIssued` is `true`, it sets the `facilityStage` to `FACILITY_STAGE.ISSUED`.
+ * - If `hasBeenIssued` is `false`, it sets the `facilityStage` to `FACILITY_STAGE.UNISSUED`.
+ */
+const mapFacilityStage = (facility: FacilityDashboard): FacilityDashboard => {
```
MarRobSoftwire, line 16: I don't know what type the input is (it'll be mapped somewhere and not the same as the mongo-db-model) but `FacilityDashboard` is inacurrate

portal/server/helpers/map-facility-properties.ts
```diff
@@ -0,0 +1,38 @@
...
+const mappings = [mapFacilityStage];
+
+/**
+ * Maps the properties of an array of facilities using a series of mapping functions.
+ *
+ * @param {Facility[]} facilities - The array of facilities to be mapped.
+ * @returns {Facility[]} The array of facilities with mapped properties.
+ */
+export const mapFacilityProperties = (facilities: FacilityDashboard[]): FacilityDashboard[] =>
+  facilities.map((facility) => mappings.reduce((mappedFacilities, mapper) => mapper(mappedFacilities), facility));
```
MarRobSoftwire, line 38: I think combining a `map` and a reduce is a bit confusing. Would this not be simpler

```suggestion
  facilities.map((facility) => {
    const facilityDashboard = cloneDeep(facility);

    facilityDashboard.facilityStage = getFacilityStage(facility);

    return facilityDashboard;
  });
```

with
```
const getFacilityStage = (facility: FacilityDashboard): string => {
  if (facility.facilityStage === FACILITY_STAGE.RISK_EXPIRED) {
    return FACILITY_STAGE.RISK_EXPIRED;
  }

  return facility.hasBeenIssued ? FACILITY_STAGE.ISSUED : FACILITY_STAGE.UNISSUED;
};
```

portal/server/helpers/map-facility-properties.test.ts
```diff
@@ -0,0 +1,115 @@
...
+    hasBeenIssued: true,
+    submittedAsIssuedDate: '1738937285806',
+    updatedAt: 1738937285814,
+    exporter: 'TEST LTD',
+    lowerExporter: 'test ltd',
+    facilityStage: FACILITY_STAGE.RISK_EXPIRED,
+  },
+];
+
+describe('mapFacilityProperties', () => {
```
MarRobSoftwire, line 47: Are these unit tests supposed to be for `mapFacilityProperties` or `mapFacilityStage`

portal/component-tests/fixtures/deal-fully-completed.js
abhi-markan: This is a historic structure to accomodate for different deal types.

portal/server/helpers/map-facility-properties.test.ts
```diff
@@ -0,0 +1,115 @@
...
+      // Assert
+      expect(result[0].facilityStage).toBe(FACILITY_STAGE.ISSUED);
+      expect(result[1].facilityStage).toBe(FACILITY_STAGE.UNISSUED);
+      expect(result[2].facilityStage).toBe(FACILITY_STAGE.RISK_EXPIRED);
+    });
+
+    it(`should return ${FACILITY_STAGE.RISK_EXPIRED} for all facilities`, () => {
+      // Arrange
+      const riskExpiredFacilities = [...mockFacilities];
+
```
MarRobSoftwire, line 62: This isnt properly cloning -
```
      riskExpiredFacilities[0].facilityStage = FACILITY_STAGE.RISK_EXPIRED;
```
is equivalent to
```      
      mockFacilities[0].facilityStage = FACILITY_STAGE.RISK_EXPIRED;
```

portal/server/helpers/map-facility-properties.test.ts
```diff
@@ -0,0 +1,115 @@
...
+      expect(result[2].facilityStage).toBe(FACILITY_STAGE.RISK_EXPIRED);
+    });
+
+    it('should return correct facility stages', () => {
+      // Arrange
+      const noFacilityStageFacilities = [...mockFacilities];
+
+      noFacilityStageFacilities[0].facilityStage = '';
+      noFacilityStageFacilities[1].facilityStage = '';
+      noFacilityStageFacilities[2].facilityStage = '';
```
MarRobSoftwire, line 82:
```suggestion
      delete noFacilityStageFacilities[0].facilityStage;
      delete noFacilityStageFacilities[1].facilityStage;
      delete noFacilityStageFacilities[2].facilityStage;
```

portal/server/helpers/map-facility-properties.test.ts
```diff
@@ -0,0 +1,115 @@
...
+      noFacilityStageFacilities[2].hasBeenIssued = false;
+
+      // Act
+      const result = mapFacilityProperties(noFacilityStageFacilities);
+
+      // Assert
+      expect(result[0].facilityStage).toBe(FACILITY_STAGE.UNISSUED);
+      expect(result[1].facilityStage).toBe(FACILITY_STAGE.UNISSUED);
+      expect(result[2].facilityStage).toBe(FACILITY_STAGE.UNISSUED);
+    });
```
MarRobSoftwire, line 113: These unit tests dont feel particularly useful - doesnt the first test capture this?

I think it would be a lot clearer if rather than passing an array into `mapFacilityProperties` individual cases are passed into `mapFacilityStage`

trade-finance-manager-api/src/v1/__mocks__/mock-durable-tasks.js
```diff
@@ -9,6 +9,284 @@ const MOCK_ACBS_TASK_LINK = {
   purgeHistoryDeleteUri: 'acbs',
 };
 
+const MOCK_ACBS_DEAL_LINK = {
```
MarRobSoftwire, line 12: What is this change for? This doesnt seem to be related to deal cancellation?

portal/server/helpers/map-facility-properties.ts
abhi-markan: Good point, the function returns `mappedFacility` object at the end of the function.

trade-finance-manager-api/src/v1/controllers/acbs.controller.test.js
```diff
@@ -504,3 +505,29 @@ describe('updateAmendedFacilityAcbs', () => {
     expect(updateFacilityAcbs).toHaveBeenCalledWith(_id, mockAcbsUpdate);
   });
 });
+
+describe('updateDealAcbs', () => {
```
MarRobSoftwire, line 509: Can this test be in its own file

trade-finance-manager-api/src/v1/__mocks__/mock-durable-tasks.js
MarRobSoftwire: These tfm-api changes should be in their own PR

trade-finance-manager-api/src/v1/controllers/acbs.controller.test.js
```diff
@@ -504,3 +505,29 @@ describe('updateAmendedFacilityAcbs', () => {
...
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+
+  it('should update the deal tfm.acbs and facility tfm.acbs object when successfully executed in ACBS', async () => {
```
MarRobSoftwire, line 514: can you add tests for the `filter` logic

trade-finance-manager-api/src/v1/helpers/all-parties-have-urn.js
```diff
@@ -4,7 +4,6 @@ const allPartiesHaveUrn = ({ dealSnapshot, tfm }) => {
   const facilitiesHaveUrn = dealSnapshot.facilities.every(({ facilitySnapshot, tfm: facilityTfm }) => {
     const hasRequiredIssuerUrn = !facilitySnapshot?.bondIssuer || facilityTfm?.bondIssuerPartyUrn;
     const hasRequiredBeneficiaryUrn = !facilitySnapshot?.bondBeneficiary || facilityTfm?.bondBeneficiaryPartyUrn;
```
MarRobSoftwire, line 6: I think this new line made the function more readable

.env.sample
```diff
@@ -64,9 +64,7 @@ UTILISATION_REPORT_DUE_EMAIL_SCHEDULE='0 8 * * *'
 UTILISATION_REPORT_OVERDUE_EMAIL_SCHEDULE='0 8 * * *'
 UTILISATION_REPORT_CREATION_FOR_BANKS_SCHEDULE='0 2 * * *'
 UTILISATION_REPORT_CREATION_FAILURE_EMAIL_ADDRESS=
-
 DEAL_CANCELLATION_SCHEDULE='* 2 * * *'
```
MarRobSoftwire, line 67: `DEAL_CANCELLATION_SCHEDULE` and `RECORD_CORRECTION_TRANSIENT_FORM_DATA_DELETE_SCHEDULE` dont fall under `# UTILISATION REPORT`. I'm not sure where makes sense but I think these new lines are needed

portal/server/helpers/map-facility-properties.ts
```diff
@@ -0,0 +1,38 @@
...
+ * Otherwise, it sets the `facilityStage` based on the `hasBeenIssued` property:
+ * - If `hasBeenIssued` is `true`, it sets the `facilityStage` to `FACILITY_STAGE.ISSUED`.
+ * - If `hasBeenIssued` is `false`, it sets the `facilityStage` to `FACILITY_STAGE.UNISSUED`.
+ */
+const mapFacilityStage = (facility: FacilityDashboard): FacilityDashboard => {
+  const mappedFacility = facility;
+
+  if (facility.facilityStage === FACILITY_STAGE.RISK_EXPIRED) {
+    mappedFacility.facilityStage = FACILITY_STAGE.RISK_EXPIRED;
+  } else {
```
natdeanlewissoftwire, line 21:
```suggestion
  if (facility.facilityStage !== FACILITY_STAGE.RISK_EXPIRED) {
```

portal/server/helpers/map-facility-properties.ts
natdeanlewissoftwire: It does indeed, have put a suggestion of what I mean below, won't make any functional difference just think it's a bit cleaner conceptually but up to you :)

libs/common/src/types/mongo-db-models/facility.ts
abhi-markan: Number of `GEF` and string for `BSS/EWCS`.

portal/server/helpers/map-facility-properties.ts
abhi-markan: Above snippet does allow for an effecient addition of additional mappers. With current approach all I will have to do is add or remove mappers from an array `const mappings = [mapFacilityStage];`.

We also do not need to deep clone the facility as we should be in position to run multiple mappers in future and save them to an accumulated array.

`reduce` provides an ability to do so in a succint pattern.

portal/server/helpers/map-facility-properties.test.ts
abhi-markan: I have created these unit tests to test both the functions, since I am only exporting `mapFacilityProperties` function and rest are not. I have also sub-described which mapping function is being tested.

portal/server/helpers/map-facility-properties.ts
abhi-markan: Have updated.

libs/common/src/types/mongo-db-models/facility.ts
abhi-markan: Good spot, have moved to `types/portal`.

libs/common/src/types/mongo-db-models/facility.ts
abhi-markan: Good spot, have introduced `PortalFacilityStage`.

portal/server/helpers/map-facility-properties.ts
abhi-markan: Have updated the docstrings.

portal/server/helpers/map-facility-properties.test.ts
abhi-markan: Apologiest this was missed out, have rectified.

portal/server/helpers/map-facility-properties.test.ts
abhi-markan: These test cases asserts scenarios where facilities have either been partially or fully created on Portal but are not risk expired.

Only function `export const mapFacilityProperties` being exported. `mapFacilityStage` is not exported and I am testing the function indirectly this has been defined with sub describe context block.

portal/server/helpers/map-facility-properties.test.ts
abhi-markan: Good spot, have implemented with `cloneDeep`.

portal/server/helpers/map-facility-properties.ts
MarRobSoftwire: It's probably safer to not include any type in the jsdoc so that we have one source of truth & don't override the typescript but will resolve anyway

portal/server/helpers/map-facility-properties.ts
MarRobSoftwire: I see the argument but I still think I'd rather keep it simple unless multiple mappers are needed. Not a big deal though so happy to resolve.

The cloning is necessary though as currently the `facility` object is being mutated I've just checked in a unit test.

image

portal/server/helpers/map-facility-properties.test.ts
```diff
@@ -0,0 +1,116 @@
...
+  });
+});
```
MarRobSoftwire, line 116: I feel like this file can be massively simplified by passing in one facility at a time

```suggestion
import { FacilityDashboard, BSS_EWCS_FACILITY_TYPE, FACILITY_STAGE } from '@ukef/dtfs2-common';
import { mapFacilityProperties } from './map-facility-properties';

const aMockFacilityDashboard = (overrides: Partial<FacilityDashboard>): FacilityDashboard => ({
  _id: '67a9e986dd0c179cf70141da',
  dealId: '67a9e95add0c179cf70141d8',
  name: 'ABC',
  currency: null,
  value: '1000.00',
  type: BSS_EWCS_FACILITY_TYPE.LOAN,
  hasBeenIssued: true,
  updatedAt: 1739188642683,
  lowerExporter: '',
  ...overrides,
});

describe('mapFacilityProperties', () => {
  describe('mapFacilityStage', () => {
    it(`should return facility stage ${FACILITY_STAGE.UNISSUED} when facility has not been issued`, () => {
      // Act
      const result = mapFacilityProperties([aMockFacilityDashboard({ hasBeenIssued: false })]);

      // Assert
      expect(result[0].facilityStage).toBe(FACILITY_STAGE.UNISSUED);
    });

    it(`should return facility stage ${FACILITY_STAGE.ISSUED} when facility has been issued`, () => {
      // Act
      const result = mapFacilityProperties([aMockFacilityDashboard({ hasBeenIssued: true })]);

      // Assert
      expect(result[0].facilityStage).toBe(FACILITY_STAGE.ISSUED);
    });

    it(`should return facility stage ${FACILITY_STAGE.RISK_EXPIRED} when facility has been issued but is risk-expired`, () => {
      // Act
      const result = mapFacilityProperties([aMockFacilityDashboard({ hasBeenIssued: true, facilityStage: FACILITY_STAGE.RISK_EXPIRED })]);

      // Assert
      expect(result[0].facilityStage).toBe(FACILITY_STAGE.RISK_EXPIRED);
    });

    it(`should return facility stage ${FACILITY_STAGE.RISK_EXPIRED} when facility has not been issued but is risk-expired`, () => {
      // Act
      const result = mapFacilityProperties([aMockFacilityDashboard({ hasBeenIssued: false, facilityStage: FACILITY_STAGE.RISK_EXPIRED })]);

      // Assert
      expect(result[0].facilityStage).toBe(FACILITY_STAGE.RISK_EXPIRED);
    });
  });
});
```

natdeanlewissoftwire (dismissed):
lgtm pending mark's comments :)

trade-finance-manager-api/src/v1/controllers/acbs.controller.test.js
```diff
@@ -504,3 +506,54 @@ describe('updateAmendedFacilityAcbs', () => {
...
+    expect(updateAcbs).toHaveBeenCalledWith(output);
+
+    // Facilities
+    expect(updateFacilityAcbs).toHaveBeenCalledTimes(1);
+    expect(updateFacilityAcbs).toHaveBeenCalledWith(facilityIdentifier, acbsFacility);
+  });
+
+  /**
+   * This is not a real word scenario as deals are always submitted with a
+   * atleast one minimum facility.
```
natdeanlewissoftwire, line 537: same nit as before

natdeanlewissoftwire (approved):

PR #4177, opened 2025-01-27
feat(FN-3675): send correction received emails

RLavender98:
## Introduction :pencil2:
When the bank submits a correction we need to send them a confirmation email and also notify PDC that the correction has been received.

- Send email to payment officer team for the bank the report belongs to when correction received
- Send email to UKEF GEF reporting emails when the correction is received

- Move email utilisation report email helpers into libs common so can be used in portal-api and dtfs-central-api
- Add api-test checking that the correction gets updated correctly
- Second PR coming which adds api test with similar checks for fee record - requires a fix which is best split into a second PR


dtfs-central-api/src/v1/controllers/utilisation-report-service/fee-record-correction/put-fee-record-correction.controller/helpers.test.ts
```diff
@@ -0,0 +1,154 @@
...
+      const expectedTemplateId = EMAIL_TEMPLATE_IDS.FEE_RECORD_CORRECTION_RECEIVED_UKEF_NOTIFICATION;
+
+      const expectedVariables = {
+        bankName,
+        exporterName: exporter,
+        tfmHomepageUrl,
+      };
+
+      expect(externalApi.sendEmail).toHaveBeenCalledWith(expectedTemplateId, firstGefReportingEmail, expect.objectContaining(expectedVariables));
+      expect(externalApi.sendEmail).toHaveBeenCalledWith(expectedTemplateId, secondGefReportingEmail, expect.objectContaining(expectedVariables));
```
sw-mattw, line 90: Could we standardise the `expect.objectContaining` usage across this and the above describe block please

There doesn't look to be any difference in terms of nesting etc so think these should be the same?

sw-mattw (commented):
Nice, one minor comment thanks

sw-mattw (dismissed):
Nice

docker-compose.yml
```diff
@@ -89,6 +89,8 @@ services:
       - FF_TFM_DEAL_CANCELLATION_ENABLED
       - DEAL_CANCELLATION_SCHEDULE
       - FF_PORTAL_FACILITY_AMENDMENTS_ENABLED
+      - UKEF_GEF_REPORTING_EMAIL_RECIPIENT
+      - TFM_UI_URL=http://localhost:5003
```
Zainzzkk, line 93: I don't think this should be set here?

libs/common/src/helpers/utilisation-report-emails.ts
```diff
@@ -0,0 +1,34 @@
...
+ * @returns The UKEF GEF Reporting email recipients.
+ */
+export const getUkefGefReportingEmailRecipients = () => {
+  const EmailsSchema = z.array(z.string().email());
+
+  try {
+    const emails = EmailsSchema.parse(JSON.parse(asString(process.env.UKEF_GEF_REPORTING_EMAIL_RECIPIENT, 'UKEF_GEF_REPORTING_EMAIL_RECIPIENT')));
+    return emails;
+  } catch (error) {
+    console.error('Failed to parse UKEF_GEF_REPORTING_EMAIL_RECIPIENT ', error);
```
Zainzzkk, line 16: %o

docker-compose.yml
Zainzzkk: ignore me, deleted

Zainzzkk (approved):

sw-mattw (approved):

PR #4183, opened 2025-01-27
feat(DTFS2-6892): add logout functionality

AlexBramhill:
## Introduction :pencil2:
Add logout functionality to SSO

- Add logout functionality
- Hide `Profile` tab when FF is set


trade-finance-manager-ui/templates/_macros/header.njk
```diff
@@ -17,9 +17,11 @@
                 <li class="moj-header__navigation-item">
                   <span data-cy="header-user-link">{{ params.user.firstName + ' ' + params.user.lastName }}</span>
                 </li>
+                {% if not IS_SSO_ENABLED %}}
```
abhi-markan, line 20: Component test please.

trade-finance-manager-ui/templates/_partials/user-logged-out.njk
```diff
@@ -0,0 +1,20 @@
+{% extends "index.njk" %}
```
abhi-markan, line 1: Component test please.

abhi-markan (changes requested):

trade-finance-manager-ui/templates/_macros/header.njk
AlexBramhill: added for both (resolved a conversation as its on a deleted file)

trade-finance-manager-ui/server/controllers/login/login-sso/login.controller.get-logout.test.ts
```diff
@@ -16,15 +16,15 @@ describe('controllers - login (sso)', () => {
       loginController = new LoginController({ loginService, userSessionService });
     });
 
-    it('redirects to /', () => {
+    it('renders the logout page', () => {
```
rkontogianni, line 19: @AlexBramhill it('should render the logout page

AlexBramhill: Updated, but just fyi due to other review changes in branches closer to main, changes like this might cause conflict (depending on my typing skills) -- we should focus on the reviews closer to main :)

trade-finance-manager-ui/templates/_partials/header.njk
```diff
@@ -1,3 +1,3 @@
 {% import '../_macros/header.njk' as header %}
 
-{{header.render({ user: user })}}
\ No newline at end of file
+{{header.render({ user: user, isSsoEnabled: IS_SSO_ENABLED })}}
```
abhi-markan, line 3: Where is `IS_SSO_ENABLED` value being picked from please.

abhi-markan (changes requested):

trade-finance-manager-ui/templates/_partials/header.njk
AlexBramhill: It's declared as a nunjucks global 
```
  nunjucksEnvironment.addGlobal('IS_SSO_ENABLED', isTfmSsoFeatureFlagEnabled());
```
trade-finance-manager-ui/server/nunjucks-configuration/index.js

If we don't use a global we will have to pass this in on every njk template manually

trade-finance-manager-ui/server/controllers/login/login-sso/login.controller.ts
```diff
@@ -89,15 +89,16 @@ export class LoginController {
...
+   * automattically be re-logged-in from.
    *
    * @param req - The HTTP request object.
    * @param res - The HTTP response object.
    */
   public getLogout = (req: Request, res: Response) => {
     console.info('User has been logged out from TFM');
     req.session.destroy(() => {
-      res.redirect('/');
+      res.render('user-logged-out.njk');
```
rkontogianni, line 101: @AlexBramhill is this for both sso enabled or disabled? why do we need this page? Shouldn't the user be redirected back to the login page? I assume that's because there is no login page for sso into dtfs folders ?

trade-finance-manager-ui/templates/_partials/header.njk
abhi-markan: Thanks.

abhi-markan (approved):
LGTM 👍🏼

trade-finance-manager-ui/server/controllers/login/login-sso/login.controller.ts
AlexBramhill: Exactly :) If we redirect to any other page, we automatically log the user back into TFM -- So this is a page to allow the session to be destroyed without relogging in

abhi-markan (approved):

PR #4188, opened 2025-01-29
fix(DTFS2-6829): fix end of month tenor test failures

RLavender98:
## Introduction :pencil2:
The logic we were using to calculate tenor was not taking into account that in January it can not be the end of the month and also the cover end date can be a lower date as it is set to one month from the day the test is running on.

Set the facility tenor e2e-test assertion using logic which should match this: https://github.com/UK-Export-Finance/databases-datateam/blob/163d71b0aef58652db5e6ccf29edce2ce3df30ca/MASTER_DATA/dbo/Functions/fn_DATEDIFF_FULLMONTH.sql


e2e-tests/tfm/cypress/e2e/journeys/case-amendments/user-flow/8. amendment-details-manual-approval.spec.js
```diff
@@ -9,15 +9,27 @@ import { PIM_USER_1, UNDERWRITER_MANAGER_DECISIONS, BANK1_MAKER1, ADMIN, CURRENC
...
+  // This should match up with the logic here:
+  // https://github.com/UK-Export-Finance/databases-datateam/blob/163d71b0aef58652db5e6ccf29edce2ce3df30ca/MASTER_DATA/dbo/Functions/fn_DATEDIFF_FULLMONTH.sql
+  const todayPlusOneMonth = add(new Date(), { months: 1 });
+  let facilityTenor = '26 months';
+  if (today.date.getDate() > todayPlusOneMonth.getDate()) {
+    facilityTenor = '25 months';
+  } else if (today.date.getDate() === todayPlusOneMonth.getDate()) {
+    const todayIsEndOfMonth = add(new Date(), { days: 1 }).getDate() === 1;
+    const aMonthFromNowIsEndOfMonth = add(new Date(), { months: 1, days: 1 }).getDate() === 1;
+    facilityTenor = todayIsEndOfMonth && !aMonthFromNowIsEndOfMonth ? '25 months' : '26 months';
```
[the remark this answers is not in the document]
MarRobSoftwire: No I think I'm wrong - if today is 30th September then a month from now is 30th October which isnt the end of the month :(

I feel there is some way to simplify this but I can't see this. Happy for this comment to be ignored :)

MarRobSoftwire (dismissed):

e2e-tests/tfm/cypress/e2e/journeys/case-amendments/user-flow/8. amendment-details-manual-approval.spec.js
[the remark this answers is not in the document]
RLavender98: What about if today is Nov 30? Then today is end of month and today plus one month is Dec 30, which is not end of month. Which the code I have written would give 25, but that change would give 26 no?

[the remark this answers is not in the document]
RLavender98: I think `today.date.getDate() > todayPlusOneMonth.getDate() === true` guarantees `todayIsEndOfMonth && !aMonthFromNowIsEndOfMonth === true` but not the inverse?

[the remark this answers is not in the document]
RLavender98: I don't think the SQL depends on the end of the month of the second date at all actually, so I think using the end of the month is a red herring - I might attempt to remove from the code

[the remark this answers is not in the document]
RLavender98: Oh actually it does - if it is it adds one if it is of type BS and it the second date is the end of the month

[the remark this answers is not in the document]
RLavender98: From the comment that existsed from the start I am assuming in this file we are dealing with BS, so I don't think it can be simplified

RLavender98, line 31: I'm worried about the 31st of Jan - I think my change means the 31st of Jan we get '25 months' but actually should get '26 months'....?

BethThomas141: Would this work instead? (Assuming the deal in the test is a BS deal?)

```
 // In the test below we use a BS deal
  // If expiry and commencement dates are the same day then you add one
  // if today is EOM and a month from now is end of month you add one

  //
  let facilityTenor = '25 months';

  if (today.date.getDate() === oneMonth.date.getDate()) {
    facilityTenor = '26 months';
  }

  const isTodayEndOfMonth = add(today.date, { days: 1 }).getDate() === 1;
  const isMonthFromNowEndOfMonth = add(today.date, { months: 1, days: 1 }).getDate() === 1;

  if (isTodayEndOfMonth && isMonthFromNowEndOfMonth) {
    facilityTenor = '26 months';
  }
```

RLavender98: Ooh ok I think mine does work, but yours is so much more readable!

MarRobSoftwire: I think this seems decent but I might give it more of a think tomorrow

e2e-tests/tfm/cypress/e2e/journeys/case-amendments/add-bank-decision/add-bank-decision-proceed.spec.js
```diff
@@ -5,21 +5,37 @@ import facilityPage from '../../../pages/facilityPage';
...
+  }
+
+  const isTodayEndOfMonth = add(today.date, { days: 1 }).getDate() === 1;
+  const isMonthFromNowEndOfMonth = add(today.date, { months: 1, days: 1 }).getDate() === 1;
+
+  if (isTodayEndOfMonth && isMonthFromNowEndOfMonth) {
+    facilityTenor = '26 months';
+  }
+
   const updatedFacilityTenor = '25 months';
```
BethThomas141, line 39: The test using this `updatedFacilityTenor` is used when we change the cover end date from being a month today to being tomorrow. 

I believe this test will fail once every 4 years on the 28th of February on a Leap year as on that day: 

- Tomorrow will be the end of month (29th feb)
- 2 years ago today was also the end of the month 

and so we should add one (making it 26th months)

Not too worried given how rarely this will fail.

BethThomas141: Resolving as not invested in this!

BethThomas141 (approved):

e2e-tests/tfm/cypress/e2e/journeys/case-amendments/add-bank-decision/add-bank-decision-proceed.spec.js
RLavender98: Ah good point! I have changed the test to change the facility cover end date to twoDays instead of tomorrow so that this should never fail.

BethThomas141 (approved):

e2e-tests/tfm/cypress/e2e/journeys/case-amendments/user-flow/8. amendment-details-manual-approval.spec.js
```diff
@@ -9,15 +9,30 @@ import { PIM_USER_1, UNDERWRITER_MANAGER_DECISIONS, BANK1_MAKER1, ADMIN, CURRENC
...
+   * - if today is end of month and a month from now is also end of month you add one
+   */
+  let facilityTenor = '25 months';
+
+  if (today.date.getDate() === oneMonth.date.getDate()) {
+    facilityTenor = '26 months';
+  }
+
+  const isTodayEndOfMonth = add(today.date, { days: 1 }).getDate() === 1;
+  const isMonthFromNowEndOfMonth = add(today.date, { months: 1, days: 1 }).getDate() === 1;
```
MarRobSoftwire, line 31: Whilst we're here we could fix the 4-yearly issue that we're actually comparing 2 years ago to today plus one month


```suggestion
  const isCoverStartDateEndOfMonth = add(twoYearsAgo.date, { days: 1 }).getDate() === 1;
  const isCoverEndDateEndOfMonth = add(today.date, { months: 1, days: 1 }).getDate() === 1;
```

e2e-tests/tfm/cypress/e2e/journeys/case-amendments/add-bank-decision/add-bank-decision-proceed.spec.js
```diff
@@ -99,7 +115,7 @@ context('Amendments underwriting - add banks decision - proceed', () => {
     cy.clickContinueButton();
     cy.url().should('contain', 'cover-end-date');
 
-    cy.completeDateFormFields({ idPrefix: 'amendment--cover-end-date', date: tomorrow.date });
+    cy.completeDateFormFields({ idPrefix: 'amendment--cover-end-date', date: twoDays.date });
```
Zainzzkk, line 118: is there a reason this has changed to two days instead of tomorrow? Ideally we should keep these bits of the test the same as before

RLavender98: Because once every 4 years the test will fail if it is tomorrow and it won't if we change it to the day after tomorrow becuse of how leap years work

BethThomas141: @Zainzzkk I left a comment asking for a change here to `twoDays` instead. Reason is: 
- The test here is just checking that updating the cover end date on the date field forms shows the updated one in the summary. Whether or not it's changed to `tomorrow` or `twoDays` doesn't matter for checking this. 
- If we have `tomorrow` we have some complicated leap year issues that will occur once a year (left in my other comment). Making it 2 days in the future instead of tomorrow means these will never occur, and is the quickest, cleanest fix.

e2e-tests/tfm/cypress/e2e/journeys/case-amendments/add-bank-decision/add-bank-decision-proceed.spec.js
```diff
@@ -5,21 +5,37 @@ import facilityPage from '../../../pages/facilityPage';
...
+    facilityTenor = '26 months';
+  }
+
+  const isCommencementDateEndOfMonth = add(twoYearsAgo.date, { days: 1 }).getDate() === 1;
+  const isExpiryDateEndOFMonth = add(oneMonth.date, { days: 1 }).getDate() === 1;
+
+  if (isCommencementDateEndOfMonth && isExpiryDateEndOFMonth) {
+    facilityTenor = '26 months';
+  }
+
```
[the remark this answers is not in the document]
RLavender98: I don't think it is worth it and would be more effort than time saved here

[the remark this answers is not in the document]
BethThomas141: Agree with @RLavender98 here - priority is getting this in ASAP, I think it's fine having this explicitly written out in each test given these are just test files and it would be fiddly to generalise this to any cover start date / end date

BethThomas141 (dismissed):

e2e-tests/tfm/cypress/e2e/journeys/case-amendments/add-bank-decision/add-bank-decision-proceed.spec.js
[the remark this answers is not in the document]
Zainzzkk: could be in libs/common otherwise and returns facilityTenor

e2e-tests/tfm/cypress/e2e/journeys/case-amendments/add-bank-decision/add-bank-decision-proceed.spec.js
RLavender98: I'll revert if you are keen to leave it though, keen to get this merged ASAP

e2e-tests/tfm/cypress/e2e/journeys/case-amendments/add-bank-decision/add-bank-decision-proceed.spec.js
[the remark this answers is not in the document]
BethThomas141: @Zainzzkk I don't think this is worth it as it's not going to be used outside these tests. Either way, could we merge this in as is to unblock the pipeline, and consider pulling it out in a separate PR if you would be keen to do that?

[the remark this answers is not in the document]
RLavender98: I don't think it should be in libs/common, it is extremely specific to these e2e-tests

e2e-tests/tfm/cypress/e2e/journeys/case-amendments/add-bank-decision/add-bank-decision-proceed.spec.js
RLavender98: I've redone the maths and I think it actually might be safe as usually tomorrow will give 25 months, and then the edge case is on Feb 28th on a leap year the default would be 24 months, but then as both are end of month, we add 1 and get 25 months, so updated tenor should still be constant.

Have reverted.

e2e-tests/tfm/cypress/e2e/journeys/case-amendments/add-bank-decision/add-bank-decision-proceed.spec.js
```diff
@@ -5,21 +5,37 @@ import facilityPage from '../../../pages/facilityPage';
...
-  // In these tests, the mock start date is two years ago today & the mock end date is a month today.
-  // Therefore if today is EOM and a month from now is not EOM we need to NOT add one to the tenor
-  const todayIsEndOfMonth = add(new Date(), { days: 1 }).getDate() === 1;
-  const aMonthFromNowIsEndOfMonth = add(new Date(), { months: 1, days: 1 }).getDate() === 1;
-  const facilityTenor = todayIsEndOfMonth && !aMonthFromNowIsEndOfMonth ? '25 months' : '26 months';
+  /**
+   * The deal in the test is a BS deal.
+   *
+   * The tenor is calculated using the SQL here:
+   * https://github.com/UK-Export-Finance/databases-datateam/blob/163d71b0aef58652db5e6ccf29edce2ce3df30ca/MASTER_DATA/dbo/Functions/fn_DATEDIFF_FULLMONTH.sql
```
Zainzzkk, line 18: this is a private repository.  Can this comment be removed and private repository should not be linked in comments

e2e-tests/tfm/cypress/e2e/journeys/case-amendments/add-bank-decision/add-bank-decision-proceed.spec.js
```diff
@@ -5,21 +5,37 @@ import facilityPage from '../../../pages/facilityPage';
...
-  // difference for BS (but not for EWCS)
-  // BUT... If the commencement date is end of month and the expiry date isn't then we don't add one
-  //
-  // In these tests, the mock start date is two years ago today & the mock end date is a month today.
-  // Therefore if today is EOM and a month from now is not EOM we need to NOT add one to the tenor
-  const todayIsEndOfMonth = add(new Date(), { days: 1 }).getDate() === 1;
-  const aMonthFromNowIsEndOfMonth = add(new Date(), { months: 1, days: 1 }).getDate() === 1;
-  const facilityTenor = todayIsEndOfMonth && !aMonthFromNowIsEndOfMonth ? '25 months' : '26 months';
+  /**
+   * The deal in the test is a BS deal.
```
Zainzzkk, line 15: Missed this BSS

e2e-tests/tfm/cypress/e2e/journeys/case-amendments/add-bank-decision/add-bank-decision-proceed.spec.js
```diff
@@ -5,21 +5,37 @@ import facilityPage from '../../../pages/facilityPage';
...
+   *
+   * The tenor is calculated using the SQL here:
+   * https://github.com/UK-Export-Finance/databases-datateam/blob/163d71b0aef58652db5e6ccf29edce2ce3df30ca/MASTER_DATA/dbo/Functions/fn_DATEDIFF_FULLMONTH.sql
+   *
+   * In these tests, the mock start date is two years ago today & the mock end date is a month today.
```
Zainzzkk, line 20:
```suggestion
   * In these tests, the mock start date is two years ago from today & the mock end date is a month from today.
```

BethThomas141 (approved):

e2e-tests/tfm/cypress/e2e/journeys/case-amendments/add-bank-decision/add-bank-decision-proceed.spec.js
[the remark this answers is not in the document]
Zainzzkk: this can go to tfm/cypress/support/utils then.  We need to ensure that the code is dry and not repeated across tests and for example the previous change needed a comment changed across 3 files.  (no need to test the function so should only take 5mins to move across)

MarRobSoftwire (approved):

e2e-tests/tfm/cypress/e2e/journeys/case-amendments/add-bank-decision/add-bank-decision-proceed.spec.js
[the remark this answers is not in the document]
RLavender98: I will raise another PR later if it is that important - but no one has been able to merge for over 24 hours - so priority needs to be getting this fix in for now please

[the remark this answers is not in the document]
Zainzzkk: can this be done next/today please

PR #4191, opened 2025-01-30
fix(DTFS2-6829): refactor facility tenor test const to be more DRY

RLavender98:
## Introduction :pencil2:
We merged a fix for some tests that where failing and blocking the pipeline. In review some DRY refactors were requested which are happening now in a follow up PR rather than on that PR to avoid blocking the rest of the team for any longer.

- Pull the facility tenor logic that is duplicated in three tests into a fixtures file


e2e-tests/tfm/cypress/fixtures/facility-tenor.js
```diff
@@ -0,0 +1,41 @@
...
+ * The tenor is calculated by MDM using a SQL function.
+ *
+ * The tenor is the difference in months between the expiry date and the commencement date.
+ * But in some cases we add one.
+ *
+ * Specifically in the case of these tests, the tenor is 25 months except for:
+ * - if expiry and commencement dates are the same date of month then you add one
+ * - if commencement is end of month and expiry is also end of month you add one
+ */
+const calculateFacilityTenorForBssFacilityWithCommencementDateTwoYearsAgoAndExpiryDateOneMonthFromToday = () => {
```
Zainzzkk, line 20: this is far too long.  Can this just be calculateBSSFacilityTenorFromCommencementDate or just calculateBSSFacilityTenor

e2e-tests/tfm/cypress/fixtures/facility-tenor.js
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  return facilityTenor;
+};
+
+export const FACILITY_TENOR = {
+  BSS: {
+    COMMENCEMENT_TWO_YEARS_AGO_EXPIRY_IN_ONE_MONTH: calculateFacilityTenorForBssFacilityWithCommencementDateTwoYearsAgoAndExpiryDateOneMonthFromToday(),
+  },
+};
```
Zainzzkk, line 41: I wonder if it is just better exporting the function const instead? `eg const facilityTenor = calculateBSSFacilityTenorFromCommencementDate();`

Zainzzkk: @abhi-markan views?

e2e-tests/tfm/cypress/fixtures/facility-tenor.js
RLavender98: Problem is I then need to pass in the commencement daet, but this logic does not hold up for any commencement date - requires more ifs etc to handle any pair of dates

RLavender98: And it is not a priority to write a function that accepts any pair of dates since is not needed and so would be a waste of time right now

e2e-tests/tfm/cypress/fixtures/facility-tenor.js
RLavender98: I wrote it like this because I do not want to export a const just named 'facilityTenor' because it is very specific to the commencement and expiry dates, and I thought `const facilityTenorForBSSFacilityWithCommencementDateTwoYearsAgoAndExpiryInOneMonth = calculateBSSFacilityTenorFromCommencementDate()` was quite illegible especially when then exported and imported in multiple files

e2e-tests/tfm/cypress/fixtures/facility-tenor.js
Zainzzkk: this is only being used for one purpose right now and there is no plan to add more.  A much shorter name is required here.  This can be updated in the future where needed (but I do not think it will ever be needed anytime soon)

RLavender98: I disagree - this is duplicating logic from an entirely separate codebase. Any new devs trying to maintain this if they ever add in a new test using facility tenor they might think 'nice I can use this existing function to calculate the tenor' and then it be a real headache and confusing when it doesn't work.

If you still disagree, let me know and I will change it, I guess it is very unlikely to be me dealing with any future maintenance.

abhi-markan: * This function deals with both the classic deals `BSS/EWCS` so the function does not represent the true state as it stands, I would highly recommend shortering the function naming any additional function logic is already well written in the DocString.

e2e-tests/tfm/cypress/fixtures/facility-tenor.js
abhi-markan: I would prefer exporting function as the above paradigm is suitable for constanst only.

abhi-markan (changes requested):

e2e-tests/tfm/cypress/fixtures/facility-tenor.js
RLavender98: I don't know so much about this area so correct me if I am wrong, but the SQL that this is mirroring, which I traced through the repos and is where MDM calculates the tenor, has logic where if 'BS' is passed in we add an extra day in some cases where we don't if 'EW' is passed in. I have assumed BS -> BSS and EW -> EWCS but please do correct me if I am wrong.

RLavender98: I will shorten the name however since you are both in favour

e2e-tests/tfm/cypress/e2e/journeys/case-amendments/add-bank-decision/add-bank-decision-proceed.spec.js
```diff
@@ -1,40 +1,17 @@
...
-    facilityTenor = '26 months';
-  }
-
-  const isCommencementDateEndOfMonth = add(twoYearsAgo.date, { days: 1 }).getDate() === 1;
-  const isExpiryDateEndOFMonth = add(oneMonth.date, { days: 1 }).getDate() === 1;
-
-  if (isCommencementDateEndOfMonth && isExpiryDateEndOFMonth) {
-    facilityTenor = '26 months';
-  }
+  const facilityTenor = calculateTestFacilityTenorValue;
```
abhi-markan, line 14:
```suggestion
  const facilityTenor = calculateTestFacilityTenorValue();
```

abhi-markan (changes requested):

e2e-tests/tfm/cypress/e2e/journeys/case-amendments/add-bank-decision/add-bank-decision-proceed.spec.js
RLavender98: Whoops 🤦

abhi-markan (approved):

PR #4201, opened 2025-02-03
chore(FN-3588): remove payment reconciliation FF

sw-mattw:
## Introduction :pencil2:
The FF_TFM_PAYMENT_RECONCILIATION_ENABLED feature flag needs removing.

Any deprecated endpoints should also be removed.

Lots of files but mostly deletions:

- Removed the `FF_TFM_PAYMENT_RECONCILIATION_ENABLED` feature flag, 'FF enabled' validation function and middleware
- Updated state machine docs with the latest event types and transition handlers
- Removed 'put utilisation report status' endpoints (calls the event handlers below), routes (`/utilisation-reports/set-status`), and payload validation in central api/tfm api/tfm ui server
- Removed 'manually set completed' and 'manually set incompleted' event handlers
- Updated utilisation report state machine to remove `MANUALLY_SET_COMPLETED` and `MANUALLY_SET_INCOMPLETE` cases
- Removed manual report reconciliation template - previously shown when FF was disabled
- Removed unused types
- Removed unused helpers for FF disabled case

N/A


trade-finance-manager-ui/templates/utilisation-reports/_macros/report-reconciliation-table.njk
```diff
@@ -21,7 +21,6 @@
 
       <tbody class="govuk-table__body">
       {% for summaryItem in summaryItems %}
-        {% set checkboxId = ['set-status--reportId-', summaryItem.reportId, '-currentStatus-', summaryItem.status] | join %}
```
sw-mattw, line 24: Removed as not used in this template

trade-finance-manager-ui/server/controllers/utilisation-reports/index.ts
```diff
@@ -22,10 +21,6 @@ export const getUtilisationReports = async (req: Request, res: Response) => {
       isPDCReconcileUser: isPDCReconcileUser(user),
     };
 
-    if (!isTfmPaymentReconciliationFeatureFlagEnabled()) {
```
RLavender98, line 25: I think there is a chance there are some fields on the view model that are not needed anymore e.g. download links, which we might then also be able to remove from the api response. Would you mind double-checking?

RLavender98 (commented):
Thanks for this - looks pretty thorough!

libs/common/src/types/utilisation-reports.ts
```diff
@@ -10,11 +10,6 @@ export type ReportPeriod = {
   end: MonthAndYear;
 };
 
-export type ReportWithStatus = {
```
Zainzzkk, line 13: will this still not be required?

trade-finance-manager-ui/server/routes/utilisation-reports/index.ts
```diff
@@ -67,47 +50,31 @@ utilisationReportsRoutes.post(
...
 );
 
-utilisationReportsRoutes.post(
-  '/:reportId/check-keying-data',
-  validateTfmPaymentReconciliationFeatureFlagIsEnabled,
-  validateUserTeam([PDC_TEAM_IDS.PDC_RECONCILE]),
-  validateSqlId('reportId'),
-  postCheckKeyingData,
-);
+utilisationReportsRoutes.post('/:reportId/check-keying-data', validateUserTeam([PDC_TEAM_IDS.PDC_RECONCILE]), validateSqlId('reportId'), postCheckKeyingData);
```
Zainzzkk, line 58: does this need to be behind any feature flag or is it fine to go live on next prod deployment

sw-mattw: Christian mentioned that the FF has been enabled for quite a while on prod so we're going to just merge this PR once approved

Zainzzkk (dismissed):

trade-finance-manager-ui/server/controllers/utilisation-reports/helpers/reconciliation-summary-helper.ts
```diff
@@ -79,28 +52,12 @@ export const getDueDateText = (reportDueDate: Date) => {
...
- * @returns Report period heading with the report periods listed
- */
-const getReportPeriodHeadingWithAllPeriods = (submissionMonth: IsoMonthStamp, reportPeriods: ReportPeriod[]) => {
-  const isCurrentSubmissionMonth = isSameMonth(new Date(submissionMonth), new Date());
-
-  const formattedReportPeriods = reportPeriods.map((reportPeriod) => getFormattedReportPeriodWithShortMonth(reportPeriod, true)).join(' and ');
-
-  return `${
-    isCurrentSubmissionMonth ? BANK_REPORTS_FOR_PERIOD_TABLE_HEADER_PREFIX.CURRENT_PERIOD : BANK_REPORTS_FOR_PERIOD_TABLE_HEADER_PREFIX.OPEN_PERIOD
-  }: ${formattedReportPeriods}`;
```
RLavender98, line 95: I think perhaps this `BANK_REPORTS_FOR_PERIOD_TABLE_HEADER_PREFIX` constant can be simplified if not removed now that there is no variation in the header but can't see any changes to it in the diff (sorry if I have just missed it!)

trade-finance-manager-ui/server/constants/bank-reports-for-period-table-header-prefix.ts
```diff
@@ -9,8 +9,6 @@ const REPORTING_PERIOD_METHOD_OF_DESCRIPTION = {
 } as const;
 
 export const BANK_REPORTS_FOR_PERIOD_TABLE_HEADER_PREFIX = {
-  CURRENT_PERIOD: `${REPORT_PERIOD_DISCRIPTORS.CURRENT} ${REPORTING_PERIOD_METHOD_OF_DESCRIPTION.BY_FULL_PERIOD}`,
```
RLavender98, line 12: Please can we also remove the 'BY_FULL_PERIOD' above as well and can then probably flatten the const it is in to something just like `const REPORTING_PERIOD = 'reporting period'`

RLavender98: Could possibly also simplify the namings of these to just 'CURRENT_PERIOD' and 'OPEN_PERIOD' since no need to distinguish

RLavender98 (approved):

abhi-markan (approved):

PR #4202, opened 2025-02-04
chore(FN-3797): remove unused fixed fee code

RLavender98:
## Introduction :pencil2:
We turned off fixed fee adjustments a few months ago when it was discovered that not all facilities on prod had fields required for the calculation. At the time we commented out and skipped some code we thought would be useful down the line as at the time we were planning to re-enable once the calculation had been reevaluated.

It is now looking like fixed fee adjustments are no longer a requirement - or if they become one again will need significant rework - so we are removing this unused code.

- Remove all unused fixed fee related code

- Remove out of date TO_DO from check the information nunjucks file


e2e-tests/tfm/cypress/e2e/journeys/utilisation-reports/report-reconciliation/generate-keying-sheet.spec.js
```diff
@@ -173,10 +173,8 @@ context('PDC_RECONCILE users can generate keying data', () => {
...
 
     /**
-     * Fixed fee adjustments are currently turned off and so should always be
-     * zero, which is displayed as a dash.
-     *
-     * TODO FN-3639: update with new calculation requirements.
+     * The fixed fee is not being used and so is always zero
+     * which is displayed as a dash
      */
     const expectedFixedFeeAdjustment = '-';
```
Zainzzkk, line 179: is it worth checking with Clara if this should totally be removed since it will not be used

RLavender98: We could ask, but it is out of scope for this ticket and there is a ticket in the backlog to cover redesign of the keying sheet after James Hall proposal decision is made so I'm not sure if it is worth doing anything intermediary?

sw-mattw (approved):

Zainzzkk (approved):

PR #4220, opened 2025-02-12
feat(FN-3813): remove adjustment columns from keying sheet

RLavender98:
## Introduction :pencil2:
The adjustment columns from the keying sheet are not actually being used by users in their current state and there is not a plan to make them work, so we should simply remove them.

- Remove adjustment columns, fixed fee and principal balance, from the keying sheet template.
- Drop facility utilisation data table (table used to track utilisation and fixed fee so that the adjustments can be calculated)
- Remove initialisation of facility utilisation data on report upload (as we no longer need to track this)
- Remove calculation of keying sheet adjustments from generate keying data controller.
- Drop fixed fee adjustment and principal balance adjustment database columns from FeeRecord table.
- Remove fixed fee adjustment and principal balance adjustment from api response when fetching data to populate keying sheet table.
- Remove removed entities from sql db helpers and mock builders.
- Remove facility utilisation data entities from sql-db-seeder
- Fetch facility ids to seed tfm facilities from FeeRecord table instead of removed FacilityUtilisationData table in create-keying-sheet-tfm-facilities script 
- Remove migrate-acbs-utilisation-data-to-tfm script from utils since it is no longer relevant (used to populate FacilityUtilisationData for existing facilities for banks opting in to the service for the first time)

## Screenshots 🖼️ 
Screenshot 2025-02-12 at 15 48 06


abhi-markan (approved):

abhi-markan:
I have changed the PR status to `Ready to review` as it was in Draft.

RLavender98:
Have had confirmation we can merge this when ready!

libs/common/src/sql-db-entities/fee-record/fee-record.entity.ts
```diff
@@ -271,58 +244,11 @@ export class FeeRecordEntity extends AuditableBaseEntity {
...
-   * @param param.status - The status
-   * @param param.requestSource - The request source
-   */
-  public updateWithKeyingData({ fixedFeeAdjustment, principalBalanceAdjustment, status, requestSource }: UpdateWithKeyingDataParams): void {
-    this.status = status;
-    this.fixedFeeAdjustment = fixedFeeAdjustment;
-    this.principalBalanceAdjustment = principalBalanceAdjustment;
-    this.updateLastUpdatedBy(requestSource);
-
-    if (status === FEE_RECORD_STATUS.RECONCILED) {
```
sw-mattw, line 321: Matt comment to self:

We're still handling this case within the updateWithStatus function

sw-mattw (dismissed):

dtfs-central-api/api-tests/v1/utilisation-reports/put-keying-data-mark-as-done.api-test.ts
RLavender98, line 1: Pushed a new commit to fix this test from being flaky. It has been flaky for a while. Until now it has only passed by coincident because we use this bank id in other tests and must not be cleaning up in some other test properly. Will separately try and have a look for the bad clean up test - but will leave for a separate PR to not hold this one up!

abhi-markan (approved):

sw-mattw (approved):

PR #4226, opened 2025-02-13
chore(FN-3849): implement due report initialised event

RLavender98:
## Introduction
When we migrated to SQL we missed a TODO to move due report initialisation logic into a state machine event.

## Resolution
- Implement the body of due report initialised event handler
- Use state machine in the cron job which initialises reports
- Update unit test mocks and assertions


dtfs-central-api/src/cron-scheduler-jobs/create-utilisation-reports/index.ts
```diff
@@ -57,26 +59,30 @@ const createUtilisationReportForBanks = async (): Promise<void> => {
...
+            },
           },
         });
 
-        await UtilisationReportRepo.save(newUtilisationReport);
-        console.info('Successfully inserted report for bank with id %s', id);
+        console.info('Successfully inserted report for bank with id %s', bankId);
       } catch (error) {
-        console.error('Error inserting report for bank with id %s. %o', id, error);
+        console.error('Error inserting report for bank with id %s. %o', bankId, error);
```
abhi-markan, line 82:
```suggestion
        console.error('Error inserting report for bank with id %s %o', bankId, error);
```

abhi-markan (changes requested):

sw-mattw (approved):

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/due-report-initialised/due-report-initialised.event-handler.ts
```diff
@@ -19,7 +20,16 @@ export type UtilisationReportDueReportInitialisedEvent = BaseUtilisationReportEv
  * @param param - The payload
  * @returns The modified report
  */
-export const handleUtilisationReportDueReportInitialisedEvent = (payload: DueReportInitialisedPayload): Promise<UtilisationReportEntity> => {
-  console.error('Utilisation due report error payload %o', payload);
-  throw new NotImplementedError('TODO FN-1860');
+export const handleUtilisationReportDueReportInitialisedEvent = async ({
```
abhi-markan, line 23: Should this be under `try-catch` block?

abhi-markan (changes requested):

dtfs-central-api/src/services/state-machines/utilisation-report/event-handlers/due-report-initialised/due-report-initialised.event-handler.ts
RLavender98: That isn't the pattern we have for the majority of our event handlers - there is handling in the controller that will catch any errors, log them and then return correct response.

abhi-markan (approved):

PR #4236, opened 2025-02-21
fix(DTFS2-7819): fix rich text bug on report upload

Zainzzkk:
## Introduction :pencil2:
This PR fixes a production issue where excels containing rich text for the exporter column would display as [Object object] on uploaded azure csv and in the fee record table in the database as was incorrectly parsed.

* Added check if cell contained rich text value
* mapped rich text value to string
* Added test
* Added e2e test to check exporter name is correctly displayed in error table - if not parsed correctly, table would show [Object object]

* Added missing new lines


portal/server/utils/csv-utils.js
```diff
@@ -63,6 +65,14 @@ const handleFloatingPointRoundingErrors = (number) => {
   return Number(number.toFixed(FLOATING_POINT_ROUNDING_DECIMAL_PLACES));
 };
 
+/**
+ * Checks if the cell value is a rich text value.
+ * returns true if so as needs further processing to convert to a string.
+ * @param {CellValue} cellValue
+ * @returns {boolean} - true if the cell value is a rich text value.
+ */
+const isRichTextValue = (cellValue) => Boolean(cellValue && Array.isArray(cellValue.richText));
```
abhi-markan, line 74:
```suggestion
const isRichTextValue = (cellValue) => cellValue && Array.isArray(cellValue.richText);
```

portal/server/utils/csv-utils.test.js
```diff
@@ -207,6 +208,48 @@ describe('csv-utils', () => {
...
+    it('returns true for a rich text value', () => {
+      const cellValue = { richText: [{ text: 'test' }, { text: 'string' }] };
+
+      const result = isRichTextValue(cellValue);
+
+      expect(result).toBe(true);
+    });
+
+    it('returns false for a non-rich text value', () => {
+      const cellValue = 'test string';
```
abhi-markan, line 221: Add an assertion with cellValue = '' please.

portal/server/utils/csv-utils.js
```diff
@@ -76,13 +86,25 @@ const extractCellValue = (cell) => {
...
+  // sets the value to the cellValue - replaced if it is a richText value
+  let value = cellValue;
+
+  /**
+   * if cellValue is a richText value
+   * then needs to be converted to a string
+   * by joining the text values of the richText array
+   */
+  if (isRichTextValue(cellValue)) {
+    value = cellValue.richText.map(({ text }) => text).join('');
```
abhi-markan, line 98:
```suggestion
    value = cellValue.richText.filter(({ text }) => text).map(({ text }) => text).join(' ');
```

* Have added filter to ensure no empty is being added.
* Was that suppose to be a space rather than an empty string otherwise exporter name could appear as 'ABCLTD' instead of 'ACB LTD'.

abhi-markan (changes requested):

portal/server/utils/csv-utils.js
Zainzzkk: this works as it should.  will add filter.  should not be ' ' otherwise it will become A B C L T D

Zainzzkk: This works as should.  Will add the filter.  If ' ', then it would change ACB LTD to A C B  L T D

Zainzzkk: actually ignore me, I will add the ' '

portal/server/utils/csv-utils.js
Zainzzkk: will keep as is as otherwise does not return false if null (returns null)

abhi-markan (approved):
LGTM 👍🏼

PR #4237, opened 2025-02-24
fix(DTFS2-7741): fixed deal cancellation portal activities

abhi-markan:
## Introduction :pencil2:
When canceling a deal on TFM, the portal activity does not accurately depict the different stages of deal cancellation for effective date in past, present and future.

* Updated the NJK macro to effectively represent the fianl stage of the deal cancellation.
  * If effective date is in past or present it will be set to `Cancelled`.
  * If it is in future it will be set to `Pending cancellation`.
  * Once actioned via CRON job it will be set to `Cancelled`.
* Introduced a new `scheduledCancellation` parameter when storing future deal cancellation portal activity.
* Reflected the updated design where the final stage is now only shown on the portal and activities instead of transition from an old stage, this was updated to avoid an additional API call to fetch the latest current stage when updating the stage to `Pending cancellation`, as when scheduled cancellation is executed the stage transition will show `Pending cancellation > Pending cancellation` (as stages are not stored in portal activities).
* Introduced a new date helper for long date format `getLongDateFormat`.
* 

image

* Used `now` instead of `new Date()` where applicable, this PR does not aim to replace all the instances of `new Date()`.
* Reorganised `.env.sample`
* Minor console message fixes.
* Removed underscore prefixed variables i.e. `_next`.
* Improved documentation.
* Lint fixes.


libs/common/src/helpers/date.test.ts
```diff
@@ -287,4 +287,29 @@ describe('date helpers', () => {
...
+
+      // Assert
+      expect(result).toBe('20 September 1989');
+    });
+
+    it('should return long date format for now', () => {
+      // Arrange
+      const date = now();
+      const month = getMonthName(date.getMonth() + 1);
+      const longDate = `${date.getDate()} ${month} ${date.getFullYear()}`;
```
rkontogianni, line 307: @abhi-markan could you add a line here?

dtfs-central-api/src/services/portal/deal.service.add-gef-deal-cancelled-activity.test.ts
```diff
@@ -34,36 +42,29 @@ describe('PortalDealService - addGefDealCancelledActivity', () => {
...
         dealId,
         dealType: DEAL_TYPE.BSS_EWCS,
         author,
         auditDetails,
+        effectiveFrom,
       });
 
       // Assert
       expect(addPortalActivityMock).toHaveBeenCalledTimes(0);
+      expect(addPortalActivityMock).not.toHaveBeenCalledWith(dealId, expectedActivity, auditDetails);
```
Zainzzkk, line 67: do we need this test if the one above is testing if been called 0 times?

dtfs-central-api/src/services/portal/deal.service.ts
```diff
@@ -43,28 +43,38 @@ export class PortalDealService {
...
+   * Adds a GEF deal cancelled activity to the portal activity repository.
+   * This function is executed for a future deal cancellation
+   * effective date.
+   *
+   * @param {Object} params - The parameters for adding the activity.
+   * @param {ObjectId | string} params.dealId - The ID of the deal.
+   * @param {DealType} params.dealType - The type of the deal.
+   * @param {ActivityAuthor} params.author - The author of the activity.
+   * @param {AuditDetails} params.auditDetails - The audit details for the activity.
+   * @param {Date} params.effectiveFrom - The date from which the cancellation is effective.
```
Zainzzkk, line 55: I don't think we need the types for these params

dtfs-central-api/src/services/tfm/deal-cancellation.service.ts
```diff
@@ -61,6 +61,7 @@ export class DealCancellationService {
       ...cancellation,
     };
 
+    // If the deal cancellation effective date is in Future
```
Zainzzkk, line 64: can this be future with lowercase f

gef-ui/server/controllers/activities-controller/index.test.js
```diff
@@ -11,12 +11,26 @@ const timestamp = 1733311320;
...
     timestamp,
-    author: MOCK_AUTHOR,
+    author: {
+      firstName: MOCK_AUTHOR.firstName,
+      lastName: MOCK_AUTHOR.lastName,
+      _id: MOCK_AUTHOR._id,
+    },
     text: '',
-    label: PORTAL_ACTIVITY_LABEL.AIN_SUBMISSION,
+    label: 'Automatic inclusion notice submitted to UKEF',
```
Zainzzkk, line 33: why has this been changed away from a const?

dtfs-central-api/src/services/portal/deal.service.add-gef-deal-cancelled-activity.test.ts
abhi-markan: Good point have removed called with.

dtfs-central-api/src/services/tfm/deal-cancellation.service.ts
abhi-markan: Updated.

gef-ui/server/controllers/activities-controller/index.js
```diff
@@ -23,39 +24,53 @@ const mapPortalActivities = (portalActivities) =>
...
+ * @returns {Promise<void>} - Renders the application activity page with the relevant data.
+ */
 const getPortalActivities = async (req, res) => {
   const { params, session } = req;
   const { dealId } = params;
   const { userToken } = session;
   const deal = await getApplication({ dealId, userToken });
 
-  // returns objects from IDs stored in gef application
+  // Returns objects from IDs stored in gef application
```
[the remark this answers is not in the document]
abhi-markan: Actually had another look, this comment is frivolous.

gef-ui/server/controllers/activities-controller/index.js
```diff
@@ -23,39 +24,50 @@ const mapPortalActivities = (portalActivities) =>
...
+ * @param {Object} req.session - The session object.
+ * @param {string} req.session.userToken - The user token.
+ * @param {Object} res - The response object.
+ * @returns {Promise<void>} - Renders the application activity page with the relevant data.
+ */
 const getPortalActivities = async (req, res) => {
   const { params, session } = req;
   const { dealId } = params;
   const { userToken } = session;
   const deal = await getApplication({ dealId, userToken });
```
Zainzzkk, line 48: not a blocker, might be worth adding a new line here so we don't have a long list

Zainzzkk (approved):

dtfs-central-api/src/services/portal/deal.service.ts
abhi-markan: As discussed, to be convered in a separate PR for consistency.

Zainzzkk (dismissed):

Zainzzkk (approved):

PR #4242, opened 2025-03-03
fix(DTFS2-7842): graceful exceptional handling for TFM page

abhi-markan:
## Introduction :pencil2:
When searching for an item which does not exists in TFM or when encountering a pagination exception when fetching facilities or deals, user is being re-directed to problem with service page or in case of page specific exception to not found page.

* User should be taken to the `deals` or `facilities` respective page with zero results in case of any pagination, generic or an unexpected exception is thrown.
* Amended test cases.
* Added test for page render with zero results.
* Improved documentation.
* Improved error labels.

* Updated Docker node image to `node:23.9.0-alpine3.20`.
* Updated Docker NGINX image to `nginx:1.27.4`.
* Introduced `libs/common` constants where `deal` and `facility` hardcoded values were being used.


dtfs-central-api/Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Node Alpine Docker Image
-FROM node:23.4.0-alpine3.20
+FROM node:23.9.0-alpine3.20
```
Zainzzkk, line 2: should we update these so close to launch or leave for next release?

trade-finance-manager-ui/server/controllers/case/index.test.js
```diff
@@ -99,7 +99,7 @@ describe('controllers - case', () => {
           deal: mockDeal.dealSnapshot,
           tfm: mockDeal.tfm,
           activePrimaryNavigation: 'manage work',
-          activeSubNavigation: 'deal',
+          activeSubNavigation: 'deals',
```
Zainzzkk, line 102: `MONGO_DB_COLLECTIONS.DEALS` here?

Zainzzkk: and same for FACILITIES

dtfs-central-api/Dockerfile
abhi-markan: We should be okay with minor release.

trade-finance-manager-ui/server/controllers/case/index.js
```diff
@@ -76,7 +76,7 @@ const getCaseDeal = async (req, res) => {
       tfm: deal.tfm,
       successMessage,
       activePrimaryNavigation: 'manage work',
-      activeSubNavigation: 'deal',
+      activeSubNavigation: 'deals',
```
Zainzzkk, line 79: I missed these.  Can be done in the next PR unless you want to do it now?

abhi-markan: Pushed

Zainzzkk (approved):

PR #4253, opened 2025-03-10
feat(DTFS2-7846): acr purge github action

abhi-markan:
## Introduction :pencil2:
In order to keep Azure subscription health and cost low we need to purge old build artifacts from ACR.

* Introduced GHA to automatically purge old artifacts as per organisation retention days for non-production enviornments.
* Introduced two new GHA composite modules to improve modularity and code clarity.
  * Azure - Purge ACR artifacts from specified repository.
  * Notify - To send notification to MS Teams using webhook.

* Deleted `deploy.sh`
* Updated `deployment.yml` and `release.yml` with notify composite module.


.github/actions/azure/action.yml
```diff
@@ -0,0 +1,45 @@
...
+    required: true
+  group:
+    description: 'Azure resource group'
+    required: true
+  acr:
+    description: 'Azure subscription and resource group service principal'
+    required: true
+  days:
+    description: 'Retention days for ACR artifacts purge i.e images older than x days will be removed.'
+    required: true
```
Zainzzkk, line 22: if there's default, do we need required: true?

.github/actions/mssql/action.yml
```diff
@@ -27,7 +27,7 @@ inputs:
   log:
     description: 'Enable MSSQL database logging'
     required: false
-    default: false
+    default: 'false'
```
Zainzzkk, line 30: does it have to be a string?

.github/workflows/deploy.sh
Zainzzkk, line 1: is there a reason for this to be deleted? was very useful for triggering deployments

.github/actions/azure/action.yml
abhi-markan: Good spot, have updated.

.github/actions/mssql/action.yml
abhi-markan: Yes as default values can only be string.
https://docs.github.com/en/actions/sharing-automations/creating-actions/metadata-syntax-for-github-actions#inputsinput_iddefault

.github/workflows/deploy.sh
abhi-markan: This is no longer required in DTFS as automatic deployment to `dev` and release based deployment on `staging` and `production`, whereas `feature` deployment is tied to specific release branch merge.

Zainzzkk (dismissed):

.github/actions/azure/action.yml
```diff
@@ -0,0 +1,45 @@
+# This GitHub Action is designed to execute Azure CLI for infrastructure maintenance
+
+name: 'Deploy Azure WebApp'
```
ChristianMcc, line 3: is this a wrong copy paste?

.github/actions/azure/action.yml
```diff
@@ -0,0 +1,45 @@
...
+      uses: actions/checkout@v4
+
+
+    # Step 2: Log in to Azure.
+    - name: Azure 🔐
+      uses: azure/login@v2
+      with:
+        creds: ${{ inputs.credentials }}
+
+    # Step 3: Purge artifacts
```
ChristianMcc, line 40: do we care about keeping the 'latest' image while purging everything else (arguably this isn't on prod yet so unlikely latest is over 30 days old)

ChristianMcc (commented):
Looking good, few comments on copy pastes and naming

.github/actions/azure/action.yml
```diff
@@ -0,0 +1,45 @@
+# This GitHub Action is designed to execute Azure CLI for infrastructure maintenance
```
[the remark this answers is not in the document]
abhi-markan: Good spot, have updated.

.github/actions/azure/action.yml
abhi-markan: Good spot, have updated.

.github/actions/azure/action.yml
abhi-markan: Good question, have added an optional `keep` attribute which default value of `3`

.github/workflows/deployment.yml
```diff
@@ -95,7 +94,7 @@ jobs:
       ENVIRONMENT: ${{ needs.setup.outputs.environment }}
 
     strategy:
-      max-parallel: 1
+      max-parallel: 2
```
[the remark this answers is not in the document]
abhi-markan: I tried this numerous times today and it seems to have worked without an issue, I think custom role and a new SP might have helped the reduction in flakiness however I am not entirely sure so well might have to update this if the CRON tasks are observed to be failing sporadically.

.github/workflows/purge.yml
```diff
@@ -0,0 +1,84 @@
...
+    runs-on: [self-hosted, DTFS, deployment]
+    env:
+      ENVIRONMENT: ${{ needs.setup.outputs.environment }}
+
+    strategy:
+      # Do not cancel in-progress jobs upon failure
+      fail-fast: false
+      # Single dimension matrix
+      matrix:
+        acr: ['tfsdev', 'tfsstaging']
```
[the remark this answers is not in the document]
abhi-markan: As discussed.

Zainzzkk (approved):

Zainzzkk (approved):

zainUKEF (approved):

ChristianMcc (approved):
LGTM

One note for the future would be we could move more of the formatting/template code for the notify action into the notify/action.yml file and then just pass through the bits of information we want to be able to change from elsewhere in workflows. Would tidy them up and make them more readable.

Not one for this PR though :)

PR #4257, opened 2025-03-12
fix(DTFS2-7825): added all 58 currencies to the currency type literal

abhi-markan:
## Introduction :pencil2:
When a non top four currency is used when creating a deal and then an amendment is processed the amendment facility value is not shown due to `currencySchema` validation which only allowed certain currencies.

* Ensured all 58 supported currencies are allowed under a union of currency literals type.
* Added an E2E tests with `AUD` as a currency for a `Loan` facility.

* Code refactorisation
  * Improve code quality.
  * Improce code readability.
  * Migrated shared schemas to `libs/common`.
  * Improved strict typing for certain functions.
  * Removed few Eslint inline disable rules.
* Improved documentation.
* Updated `feature` deployment CRON schedule to `01:00`.


libs/common/src/interfaces/country.ts
```diff
@@ -0,0 +1,25 @@
+/**
+ * Interface representing a country.
+ */
+export interface CountryInterface {
```
zainUKEF, line 4: The other types/interfaces don't have "Interface" in their names.  Would it be better keeping with the current pattern

libs/common/src/interfaces/currency.ts
```diff
@@ -0,0 +1,25 @@
+/**
+ * Interface representing a currency.
+ */
+export interface CurrencyInterface {
```
zainUKEF, line 4: Same as above

abhi-markan: Only reason I added here as `Currency` is a type and when exporter from `{ Currency } from '@dtfs2/common'` it will import a type, therefore adding explicity suffix.

libs/common/src/interfaces/country.ts
abhi-markan: Good point, have answered below.

external-api/src/helpers/array.test.ts
```diff
@@ -0,0 +1,335 @@
...
+        id: 'D',
+      },
+      {
+        currencyId: 7,
+        text: 'G',
+        id: 'G',
+      },
+    ]);
+  });
+
```
tony-barnes-ukef, line 134: Lots of repetition, could do something like this:

```ts
const currencyA = { ... };
const currencyB = { ... };

const expected = [currencyA, currencyB]
```

Not sure if this would be sensible with the industries

external-api/src/helpers/array.ts
```diff
@@ -1,9 +1,13 @@
-/* eslint-disable @typescript-eslint/unbound-method */
-/* eslint-disable @typescript-eslint/no-unsafe-call */
-/* eslint-disable @typescript-eslint/no-unsafe-member-access */
-/* eslint-disable @typescript-eslint/no-unsafe-assignment */
-/* eslint-disable @typescript-eslint/no-unsafe-argument */
-/* eslint-disable @typescript-eslint/no-explicit-any */
-/* eslint-disable @typescript-eslint/no-unsafe-return */
```
tony-barnes-ukef, line 7: Lovely 😍

libs/common/src/interfaces/industry.ts
```diff
@@ -0,0 +1,34 @@
...
+   * The code of the industry sector.
+   */
+  code: string;
+
+  /**
+   * The name of the industry sector.
+   */
+  name: string;
+
+  /**
```
zainUKEF, line 30: could we instead do IndustrySectorInterface  extends IndustryClassInterface  so we don't repeat

external-api/src/helpers/array.ts
```diff
@@ -1,9 +1,13 @@
...
+
+/**
+ * Sorts an array of objects alphabetically based on a specified field.
+ *
+ * @param {unionInterface[]} arr - The array of objects to be sorted.
+ * @param {string} field - The field name by which to sort the objects.
+ * @returns {unionInterface[]} The sorted array of objects.
+ */
+export const sortArrayAlphabetically = (arr: unionInterface[], field: string): unionInterface[] =>
+  arr.sort((a: unionInterface, b: unionInterface) => (a[field as keyof unionInterface] as string).localeCompare(b[field as keyof unionInterface] as string));
```
tony-barnes-ukef, line 13: Are the `as string`'s necessary?

libs/common/src/constants/currency.ts
zainUKEF, line 1: it would be good one day to have this come from MDM/APIM

libs/common/src/interfaces/currency.ts
tony-barnes-ukef: What's the difference between `Currency` in common compared to here? Worried about potential confusion in the future RE what to use

Same question for IndustryClass/Sector below

external-api/src/helpers/array.ts
```diff
@@ -1,9 +1,13 @@
...
-export const sortArrayAlphabetically = (arr: any, field: any) => arr.sort((a: any, b: any) => a[field].localeCompare(b[field]));
+type unionInterface = CountryInterface | CurrencyInterface | IndustrySectorInterface;
+
+/**
+ * Sorts an array of objects alphabetically based on a specified field.
+ *
+ * @param {unionInterface[]} arr - The array of objects to be sorted.
+ * @param {string} field - The field name by which to sort the objects.
+ * @returns {unionInterface[]} The sorted array of objects.
+ */
```
zainUKEF, line 11: we don't need the types here

external-api/src/helpers/array.test.ts
abhi-markan: Good point, have refactored.

external-api/src/helpers/array.ts
abhi-markan: Covered previously will remove them in one single PR once we have established the standard on TSDocs.

external-api/src/helpers/array.ts
abhi-markan: Yes as `localCompare` is only available to `string` type and not to literal types of `unionInterface`

libs/common/src/constants/currency.ts
abhi-markan: Agree, however in a separate feature ticket as this will also require a either a list of allowed currencies from MDM or APIM (MDM) to return only currencies accepted in DTFS.

libs/common/src/interfaces/industry.ts
abhi-markan: Good point, have extended.

rkontogianni (approved):

libs/common/src/interfaces/currency.ts
tony-barnes-ukef, line 4: What's the difference between Currency in common compared to here? Worried about potential confusion in the future RE what to use

Same question for IndustryClass/Sector below

zainUKEF (approved):

libs/common/src/interfaces/currency.ts
abhi-markan: If you mean `Currency` then that is a type and `CURRENCY` and `ALL_CURRENCIES` are literal types of currencies. I wanted to explicity suffix interfaces with `Interface` so when being imported from DTFS common library then the developer know which one they are exporting.

zainUKEF (dismissed):

trade-finance-manager-ui/server/controllers/case/index.js
```diff
@@ -325,60 +325,65 @@ const formatAmendmentDetails = (allAmendments) => {
...
+      hasAmendmentInProgress,
+      hasAmendmentInProgressButton,
+      allAmendments,
+      amendments,
+      amendmentsInProgress,
+      showFacilityEndDate: facility.facilitySnapshot.isGef,
+    });
+  } catch (error) {
+    console.error('Error getting case facility %o', error);
+    return res.redirect('/not-found');
```
tony-barnes-ukef, line 385: Unit test for this would be great if possible

tony-barnes-ukef (dismissed):

trade-finance-manager-ui/server/controllers/case/index.js
abhi-markan: Good spot, I was meaning to and forgot have added. Thanks

tony-barnes-ukef (approved):

PR #4261, opened 2025-03-13
fix(DTFS2-7846): feature deployment and ACR purge

abhi-markan:
## Introduction :pencil2:
Ensure `feature` enviornment is also incorporated into automated deployment pipelines.

* Added nightly `main` deployment to `feature` at `00:30`.
* Added feature ACR to `purge.yml`.

* Added an exception to cspell.

### DIFF
--- a/.github/workflows/feature.yml
+++ b/.github/workflows/feature.yml
@@ -30,7 +30,7 @@ run-name: 🚀 Deploying to feature
 
 on:
   schedule:
-    - cron: '0 30 * * *'
+    - cron: '30 0 * * *'
 
   push:
     branches:
--- a/.github/workflows/feature.yml
+++ b/.github/workflows/feature.yml
@@ -2,6 +2,7 @@
 # *******************
 #
 # This GitHub Actions workflow automates the deployment process for web apps and Azure Functions to Azure.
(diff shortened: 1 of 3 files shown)


.github/workflows/feature.yml
```diff
@@ -2,6 +2,7 @@
 # *******************
 #
 # This GitHub Actions workflow automates the deployment process for web apps and Azure Functions to Azure.
+# It is triggered when on base default branch i.e. `main` at every day at midnight. The workflow runs after a successful QAT workflow
```
zainUKEF, line 5:
```suggestion
# It is triggered when on base default branch i.e. `main` at every day at midnight. The workflow runs after a successful QAT workflow
```
```suggestion
# It is triggered everyday at midnight when on base default branch i.e `main`. The workflow runs after a successful QAT workflow
```

.github/workflows/feature.yml
```diff
@@ -159,59 +160,45 @@ jobs:
...
+            "@type": "OpenUri",
+            "name": "Deployment",
+            "targets":
+            [
+            {
+            "os": "default",
+            "uri": "https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}",
+            },
+            ],
+            },
```
zainUKEF, line 201: should this be indented?

PR #4266, opened 2025-03-18
chore(deployment): updated deployment cron schedule

abhi-markan:
## Introduction :pencil2:
Updated deployment schedule.

### DIFF
--- a/.github/workflows/deployment.yml
+++ b/.github/workflows/deployment.yml
@@ -30,7 +30,7 @@ run-name: 🚀 Deploying to dev
 
 on:
   schedule:
-    - cron: '0 21 * * *'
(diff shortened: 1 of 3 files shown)


.github/workflows/deployment.yml
```diff
@@ -30,7 +30,7 @@ run-name: 🚀 Deploying to dev
 
 on:
   schedule:
-    - cron: '0 21 * * *'
+    - cron: '0 19 * * *'
```
zainUKEF, line 33: is this not too early? if anyone's QA'ing till late on dev, then it could interupt?

zainUKEF (approved):

PR #4294, opened 2025-04-04
fix(DTFS2-7917): record correction - cancel link

zainUKEF:
## Introduction :pencil2:
This PR fixes the record correction cancel link which was a button but should be a link

* Changed create-record-correction-request and check-the-information templates to have a link instead of a button
* Changed post cancel route to a get (as pressing a link triggers a get request)
* Updated tests
* Updated and added to e2e tests

## Screenshot :camera_flash:
image


abhi-markan (approved):
LGTM 👍🏼

PR #4327, opened 2025-05-23
fix(DTFS2-7953): acbs tfm reconciliation check

abhi-markan:
# Pull Request

## Introduction :pencil2:

When a TFM reconcilliation with ACBS fails for the deal, it allows maker to action facility issuance on TFM however due to a controlloer level validation at `TFM-API` the payload is never constructed and therefore an invalid facility stage is presented to the TFM user stating facility is `Issued` but in reality it is not issued in ACBS due to validation.

* Introduced an additional check `isInAbcs` for facility issuance link appearance on Portal (Maker only).
* Added test cases

* Improved documentation.
* Improved code quality by refactoring helper function.
* Clearer function names.

## Screenshot :camera_flash:
When facility can be issued, the link is not presented to the user since ACBS reconcilliation has not successfully finished.

image


portal-api/src/v1/controllers/tfm.controller.js
```diff
@@ -1,6 +1,6 @@
 const { ALL_TEAM_IDS } = require('@ukef/dtfs2-common');
 const { HttpStatusCode } = require('axios');
-const { getTfmTeam } = require('../api');
+const { getTfmTeam, getTfmDeal } = require('../api');
```
rkontogianni, line 4: @abhi-markan do you have a unit test for this ?

gef-ui/server/services/__tests__/api.test.js
```diff
@@ -578,3 +578,152 @@ describe('getAmendmentsOnDeal()', () => {
     await expect(returned).rejects.toThrow(InvalidDealIdError);
   });
 });
+
+describe.only('getTfmDeal', () => {
```
zainUKEF, line 582: remove .only please

zainUKEF (dismissed):

PR #4344, opened 2025-06-13
fix(DTFS2-7995):  fix getFacilities

rkontogianni:
# Pull Request

## Introduction :pencil2:
This PR is to fix the application-details page when the Portal amendment feature flag is disabled.

- check if isPortalFacilityAmendmentsFeatureFlagEnabled 
-  if is enable then getAcknoledgedAmendments and map the facility values
-  unit tests


portal-api/src/v1/gef/controllers/facilities.controller.js
```diff
@@ -86,9 +87,12 @@ const getAllFacilitiesByDealId = async (dealId) => {
 };
 exports.getAllFacilitiesByDealId = getAllFacilitiesByDealId;
 
+exports.isPortalFacilityAmendmentsFeatureFlagEnabled = isPortalFacilityAmendmentsFeatureFlagEnabled;
```
abhi-markan, line 90:
```suggestion
exports.isPortalFacilityAmendmentsFeatureFlagEnabled = isPortalFacilityAmendmentsFeatureFlagEnabled;
```

Do we need this line, can we not simply just use `isPortalFacilityAmendmentsFeatureFlagEnabled`?

portal-api/src/v1/gef/controllers/facilities.controller.test.js
```diff
@@ -0,0 +1,163 @@
...
+    };
+
+    db.getCollection = mockDatabase.getCollection;
+  });
+
+  afterAll(() => {
+    jest.resetAllMocks();
+  });
+
+  describe('when FF_PORTAL_FACILITY_AMENDMENTS_ENABLED is disabled', () => {
```
abhi-markan, line 80: Need an assertion where `FF_PORTAL_FACILITY_AMENDMENTS_ENABLED` is not specicied.

abhi-markan (changes requested):

portal-api/src/v1/gef/controllers/facilities.controller.js
rkontogianni: @abhi-markan I need to keep this as it is, because I need to mock the result in unit test

zainUKEF (approved):

PR #4350, opened 2025-06-24
fix(DTFS2-8004): unable to access timeZoneConfig schema

abhi-markan:
# Pull Request

## Introduction :pencil2:

`gef-ui` was unable to access `timeZoneConfig` schema's DEFAULT property.

* Simplification of code: Removed `zod` and replaced with direct `TZ` enviornment variable access, in its absence a default constant value will be used `Europe/London`.


audrius-vaitonis (approved):
Looks fine,

some comments:
* if timeZoneConfig access stopped to work, maybe it is better to fix it instead of replacing.
* old name `timeZoneConfig.DEFAULT` better describes that we are about to use static default value. New name `timezone` is very generic.

rkontogianni (approved):

PR #4372, opened 2025-07-31
feat(DTFS2-8016): gef fixed fee calculation as per guarantee fee

abhi-markan:
# Introduction :pencil2:

The current fixed fee calculation formula for `GEF` facilities uses bank's interest margin.
This has been identified as incorrect and therefore should be changed to guarantee fee payable to UKEF.

* Updated to `guaranteeFee` number property.
* Migrated GEF fixed fee formula calculation to `libs/common`.
* Removed `coverEndDateTimestamp` property for GEF fixed fee calculation, as the existing function has been expanded to cater for both date and EPOCH timestamp.
* Updated unit test cases.


libs/common/src/helpers/facility-calculations/index.test.ts
```diff
@@ -235,4 +243,175 @@ describe('GEF drawn amount', () => {
       expect(response).toEqual(FACILITY_UTILISATION_PERCENTAGE.DEFAULT);
     });
   });
+
+  describe('calculateDaysOfCover', () => {
+    it('should return difference in cover start and end date in days for EPOCH start', () => {
```
zainUKEF, line 248: can this be more descriptive for why it should be 0?

libs/common/src/helpers/facility-calculations/index.test.ts
```diff
@@ -235,4 +243,175 @@ describe('GEF drawn amount', () => {
...
+      const drawnAmount = 100;
+      const daysOfCover = 30;
+      const dayCountBasis = 365;
+      const guaranteeFee = 1;
+
+      // Act
+      const response = calculateFeeAmount(drawnAmount, daysOfCover, dayCountBasis, guaranteeFee);
+
+      // Assert
+      expect(response).toBe(0.0821917808219178);
```
zainUKEF, line 297: should this function not return a rounded number?

abhi-markan: No requirement was captured historically, we currently round-off to 2DP elsewhere.

libs/common/src/helpers/facility-calculations/index.test.ts
```diff
@@ -235,4 +243,175 @@ describe('GEF drawn amount', () => {
...
+      // Act
+      const response = calculateFeeAmount(drawnAmount, daysOfCover, dayCountBasis, guaranteeFee);
+
+      // Assert
+      expect(response).toBe(0);
+    });
+  });
+
+  describe('calculateGefFacilityFeeRecord', () => {
+    const facility = {
```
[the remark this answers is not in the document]
abhi-markan: Not in libs/common.

zainUKEF (dismissed):

PR #4377, opened 2025-08-04
fix(DTFS2-8016): revert fixed fee formula to interest percentage

abhi-markan:
# Introduction :pencil2:

Revert `guaranteeFee` to `interestPercentage` in fixed fee formula calculation.


zainUKEF (approved):
can you use the PR format for the description please

PR #4400, opened 2025-08-26
feat(DTFS2-8078): add an additional day for contingent facility

abhi-markan:
# Introduction :pencil2:

When calculating the difference in days for a GEF facility premium schedule, the difference in days calculation is out by a day for the contigent facility.

* Updated the `calculateDaysOfCover` method logic, if the facility type is `Contingent` then an additional day is added as the current formula is exclusive difference (how many days are needed to reach the final date). This has been updated to inclusive difference for `Contingent` facilities only.

* Ellaborated unit test cases

### DIFF
--- a/trade-finance-manager-api/src/v1/helpers/calculate-gef-facility-fee-record.api-test.js
+++ b/trade-finance-manager-api/src/v1/helpers/calculate-gef-facility-fee-record.api-test.js
@@ -20,9 +20,9 @@ describe('calculate-gef-facility-fee-record', () => {
 
   describe('calculateDaysOfCover', () => {
     it('should return the amount of days between start and end cover dates for a cash facility', () => {
-      const result = calculateDaysOfCover(mockCoverStartDate, mockCoverEndDate);
+      const result = calculateDaysOfCover(FACILITY_TYPE.CASH, mockCoverStartDate, mockCoverEndDate);
 
-      const expected = differenceInDays(FACILITY_TYPE.CASH, new Date(Number(mockCoverEndDate)), new Date(Number(mockCoverStartDate)));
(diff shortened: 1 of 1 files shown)


libs/common/src/helpers/facility-calculations/index.test.ts
```diff
@@ -309,9 +405,10 @@ describe('GEF drawn amount', () => {
       // Arrange
       const coverStartDate = null;
       const coverEndDate = null;
+      const type: GefFacilityType = GEF_FACILITY_TYPE.CASH;
```
rkontogianni, line 408: @abhi-markan in this test return 0 but the description says null

rkontogianni (approved):

trade-finance-manager-api/src/v1/helpers/calculate-gef-facility-fee-record.api-test.js
```diff
@@ -19,10 +19,10 @@ describe('calculate-gef-facility-fee-record', () => {
   const mockDayBasis = 365;
 
   describe('calculateDaysOfCover', () => {
-    it('should return the amount of days between start and end cover dates', () => {
+    it('should return the amount of days between start and end cover dates for a cash facility', () => {
       const result = calculateDaysOfCover(mockCoverStartDate, mockCoverEndDate);
 
-      const expected = differenceInDays(new Date(Number(mockCoverEndDate)), new Date(Number(mockCoverStartDate)));
+      const expected = differenceInDays(FACILITY_TYPE.CASH, new Date(Number(mockCoverEndDate)), new Date(Number(mockCoverStartDate)));
```
rkontogianni, line 25: @abhi-markan differenceInDays is not from date-fns? can you pass a thrird parameter type cash here ?

PR #4401, opened 2025-08-28
feat(DTFS2-7812): missing probability of default for BSS/EWCS deals

abhi-markan:
# Introduction :pencil2:

During a `BSS/EWCS` deal submission to TFM, party URN for all the required parties are not being added or created. Upon investigation `/customers` MDM endpoint is not being invoked.

* Since PoD is not provided for `BSS/EWCS`, `tfm.probabilityOfDefault` is null. Have added nullish coalescing with `14.1` value if not available. 

* Added an API test case.
* Added try-catch block to `submitDealAfterUkefIds` for graceful error handling.

### DIFF
--- a/trade-finance-manager-api/src/v1/controllers/deal.submit.controller.js
+++ b/trade-finance-manager-api/src/v1/controllers/deal.submit.controller.js
@@ -1,3 +1,4 @@
+const { HttpStatusCode } = require('axios');
 const { ObjectId } = require('mongodb');
 const { generatePortalAuditDetails } = require('@ukef/dtfs2-common/change-stream');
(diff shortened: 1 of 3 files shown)


trade-finance-manager-api/src/v1/controllers/deal.submit.controller.js
```diff
@@ -48,156 +49,162 @@ const getPortalDeal = async (dealId, dealType) => {
  * Azure function
  */
 const submitDealAfterUkefIds = async (dealId, dealType, checker, auditDetails) => {
-  const deal = await getPortalDeal(dealId, dealType);
+  try {
```
rkontogianni, line 52: @abhi-markan could you add @param in the DocString?

rkontogianni (approved):

PR #4470, opened 2025-10-07
feat(DTFS2-6718): removed post user creation endpoint

abhi-markan:
# Introduction :pencil2:

The POST /user (note that this is different from POST /users) endpoint in Portal API presents a security risk. It allows any user (i.e., including a not-logged-in user) to create a new user with any privileges. Although it is API key protected, the potential impact of a malicious actor obtaining this API key and misusing the endpoint is quite high.

The endpoint appears to only be used in the data-migration and mock-data-loader code to create an initial admin user (which subsequently creates the users, banks etc. needed in local dev/testing).

* Removed the POST `/user` endpoint and replaced it with direct DB insertion.
* To accomplish above a key password hash generation function for both `portal-api` and `trade-finance-manager-api` has been centralised to `libs/common` under `generatePasswordHash`.
* Improved logging.

* Updated test cases.

### DIFF
--- a/utils/mock-data-loader/database/user-repository.ts
+++ b/utils/mock-data-loader/database/user-repository.ts
@@ -5,6 +5,17 @@ import { mongoDbClient } from '../../drivers/db-client';
 
 const PRIV_KEY = Buffer.from(process.env.JWT_SIGNING_KEY ?? '', 'base64').toString('ascii');
 
+/**
+ * Creates a mock user in the database for testing or development purposes.
+ *
+ * This function takes a `PortalUser` object with an additional `password` property,
+ * generates a password hash and salt, and inserts the user into the 'users' collection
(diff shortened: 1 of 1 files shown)


libs/common/src/helpers/crypto/password.test.ts
```diff
@@ -0,0 +1,58 @@
...
+    (crypto.randomBytes as jest.Mock).mockReturnValue(mockSaltBuffer);
+    (crypto.pbkdf2Sync as jest.Mock).mockReturnValue(mockHashBuffer);
+  });
+
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
+
+  it('should crypto functions when generating salt and hash', () => {
+    // Arrange
```
zainUKEF, line 24: remove please

PR #4720, opened 2026-04-17
feat(DTFS2-8378): external-api, trade-finance-manager-api - apim tfs gift endpoint setup

tony-barnes-ukef:
# Introduction :pencil2:

This PR sets up some new endpoints (currently not in use), so that TFM API will be able to call APIM TFS's "Create GIFT facility" endpoint.

Also introduces some minor technical improvements after trying to run, load and test things locally.

- `external-api` - create new route and controller, `giftFacility.create`
- `trade-finance-manager-api`  - Add `createGiftFacility` method to `api.js`, ready for use.

- Add `@types/jsonwebtoken` to `libs/common` production dependencies. This was causing some issues when attempting to run the codebase locally in production mode.
- Add an example constant - APIM GIFT payload to `lib/common`.
- Minor typo fixes.
- Minor readability improvements.
- Update `mock-data-loader` `dotenv` paths. It was not working as expected when running `npm run load` from the root directory.
- Update `mock-data-loader` `createMockDataUser` function to have the correct documentation. Remove unnecessary `false` return.
- Update API swagger route/definitions to include an API key auth security scheme - allowing developers to test endpoints in swagger.
  - `external-api`
  - `portal-api`
  - `trade-finance-manager-api`

## Request / response :eyes:

<details>
  <summary>Show/hide</summary>

```json

```

</details>

<details>
  <summary>Show/hide</summary>

Add screenshots here.

</details>


external-api/api-tests/v1/gift-facility.api-test.ts
```diff
@@ -0,0 +1,77 @@
...
+      data: {},
+      status: HttpStatusCode.Created,
+    };
+
+    it(`should return a ${HttpStatusCode.Created} status with data received from APIM TFS`, async () => {
+      // Arrange
+      nock.abortPendingRequests();
+      nock.cleanAll();
+
+      nock(baseUrl).persist().post(url).reply(HttpStatusCode.Created, mockApimTfsResponse);
```
tony-barnes-ukef: @zainUKEF  @sungharsh-ukef 

I have opted to use `nock` here, the for External API tests to APIM TFS, which allows us to stub the APIM TFS endpoint and test that External API services responds accordingly.

The alternative to this would be to not stub the APIM TFS endpoint, but this could introduce flakiness in DTFS CI/CD for various reasons. Including that, a random generation or check of facility IDs would have to be introduced, so that such DTFS API tests do not fail, if the facility ID is already in use.

Even with a solution to the facility ID issue, there would still be potential instability, as without mocking, this endpoint ultimately calls 2x other services - their endpoints could change. The order of calls without mocking is:

1) DTFS - TFM API
2) DTFS - External API
3) APIM TFS - Multiple calls to GIFT
4) GIFT

For DTFS, It feels best and safest to stub the APIM TFS.

In terms of testing the full flow, we would like to introduce contract tests for the whole flow, outside of DTFS.

In the meantime, manual integration testing will be made outside of API tests, i.e by QA in dev/test environments.

external-api/package.json
```diff
@@ -50,6 +50,7 @@
     "joi": "^17.13.3",
     "lodash": "^4.18.1",
     "mongodb": "^4.17.2",
+    "nock": "^14.0.12",
```
sungharsh-ukef: can we move nock to devDependencies as this should not be shipped to production.

libs/common/package.json
```diff
@@ -56,6 +56,7 @@
     "@types/big.js": "^6.2.2",
     "@types/connect-redis": "^0.0.23",
     "@types/express-session": "^1.18.2",
+    "@types/jsonwebtoken": "^9.0.10",
```
sungharsh-ukef, line 59: same, can we move `@types/jsonwebtoken` to devDependencies as this should not be shipped to production.

tony-barnes-ukef: This should be shipped to production like the other types. I spotted some issues when attempting to run the codebase locally in production mode.

external-api/package.json
tony-barnes-ukef: Good spot, updated

external-api/server/v1/controllers/gift-facility.controller.ts
```diff
@@ -0,0 +1,81 @@
...
+ * Create a GIFT facility.
+ * @param req request object
+ * @param res response object
+ * @returns response with HTTP status `code` and `data`
+ */
+export const create = async (req: Request, res: Response) => {
+  try {
+    console.info('⚡️ Invoking APIM TFS GIFT facility endpoint');
+
+    const url = `${APIM_TFS_URL}v2/gift/facility`;
```
sungharsh-ukef, line 51: I think the `APIM_TFS_URL` need to be normalized before appending to `v2/gift/facility` to remove any unnecessary slashes and remove possible undefined as below
```diff
- const url = `${APIM_TFS_URL}v2/gift/facility`;
+ const baseUrl = String(APIM_TFS_URL || '').replace(/\/+$/u, '');
+ const url = `${baseUrl}/v2/gift/facility`;

Comment (tony-barnes-ukef) [reply external-api/server/v1/controllers/gift-facility.controller.ts:51]: `APIM_TFS_URL` is an environment variable, we shouldn't be replacing or defaulting it to an empty string.

It follows the same pattern as other APIM URL env vars in DTFS:
- APIM_MDM_URL
- APIM_ESTORE_URL

File: external-api/server/v1/controllers/gift-facility.controller.ts [hunk 5]
```diff
@@ -0,0 +1,81 @@
...
+      method: 'post',
+      url,
+      headers,
+      data: req.body,
+    }).catch((error: any) => {
+      console.error('Error calling APIM TFS GIFT facility endpoint %o', error);
+      return {
+        status: error.response?.status,
+      };
+    });
```
sungharsh-ukef, line 63: Can we avoid .catch() on axios(), I think we can still catch errors in the below catch() block

tony-barnes-ukef: This would be sensible, but it follows the existing pattern in seemingly, all `external-api` controllers.

Changing this would make the controllers inconsistent. Would prefer for them to be consistent.

external-api/api-tests/v1/gift-facility.api-test.ts
```diff
@@ -0,0 +1,59 @@
+/* eslint-disable @typescript-eslint/no-unsafe-member-access */
+/* eslint-disable @typescript-eslint/no-unsafe-assignment */
+
+import { HttpStatusCode } from 'axios';
+import nock from 'nock';
```
zainUKEF, line 5: why does this need nock? can it not use normal jest mocking

tony-barnes-ukef: Tagged you in a comment already: https://github.com/UK-Export-Finance/dtfs2/pull/4720#discussion_r3100490398

zainUKEF (approved):

sungharsh-ukef (approved):

PR #4736, opened 2026-05-05
fix(DTFS2-8427): show suspended account page when suspended user retries login

sungharsh-ukef:
# Introduction :pencil2:

This PR fixes a bug where a suspended user sees the suspended account page on the first login attempt, but on re-login the page is not shown and the user remains on the login page.

### The Issue:
For example, when an already suspended user re-submitted the login form without having contacted support to reactivate their account, the login attempt failed with a 403 Forbidden response before any session token was created. The code then redirected them to the suspended account page at `/login/temporarily-suspended-access-code`. However, that page had a middleware guard called `validatePartialAuthToken` which expected a valid session token to be present, the same token that a normal 2FA login would have created. Because the login had failed before that token was ever issued, the session was empty, the middleware found nothing, and quietly sent the user back to /login with no message. From the user's perspective, clicking login simply did nothing.

### The fix:
Removed `validatePartialAuthToken` from the `/login/temporarily-suspended-access-code` route in `login/index.js`, allowing the page to render as a purely static informational message with contact details. Since the page does not read from the session or perform any 2FA logic, there is no functional need for a session token to access it. As a result, the page can now be accessed without a session token. The `validatePortal2FAEnabled` guard is retained to ensure the page is only reachable when the 2FA feature flag is enabled.

As a result, if a user attempts to log in multiple times after their account has been suspended, they will consistently be shown this page.

- portal-ui\server\routes\login\index.js

- portal-ui/server/controllers/login/post-login.js
- portal-ui/server/controllers/login/post-login.test.js-

- portal-ui/server/controllers/login/get-account-suspended.ts
- portal-ui/server/controllers/login/get-account-suspended.test.ts

- activate-deactivate-user.spec.js
- temporarily-suspended-access-code.page.spec.js
- e2e-tests/support/tasks.js 

<details>
  <summary>Show/hide</summary>

https://github.com/user-attachments/assets/9f5f2dd1-06fb-4de8-922c-6228d33fd062

</details>


portal-ui/server/routes/login/index.js
```diff
@@ -397,7 +397,7 @@ router.post('/login/resend-another-access-code', validatePortal2FAEnabled, valid
  *       403:
  *         description: Forbidden
  */
-router.get('/login/temporarily-suspended-access-code', validatePortal2FAEnabled, validatePartialAuthToken, getAccountSuspendedPage);
+router.get('/login/temporarily-suspended-access-code', validatePortal2FAEnabled, getAccountSuspendedPage);
```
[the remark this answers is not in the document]
sungharsh-ukef: Done, added additional session guard to check if numberOfSignInOtpAttemptsRemaing in `portal-ui/server/controllers/login/post-login.js`

portal-ui/server/controllers/login/get-account-suspended.ts
```diff
@@ -1,26 +1,8 @@
...
-
-  if (attemptsLeft === -1) {
-    return res.render('login/temporarily-suspended-access-code.njk');
-  }
-
-  console.error('Invalid OTP attempts: expected -1 remaining attempts but found %d', attemptsLeft);
-
-  return res.redirect('/not-found');
-};
+export const getAccountSuspendedPage = (_req: Request, res: Response) => res.render('login/temporarily-suspended-access-code.njk');
```
sungharsh-ukef, line 8: Note:  This controller intentionally does not read from the session and the previous `GetAccountSuspendedPageRequest` type has been removed. The page is purely a static informational message with heading, suspension explanation and public support contact details and does not display any user or session specific data.

Previously the controller required `req.session.numberOfSignInOtpAttemptsRemaining` to be `-1` and otherwise redirected to `/not-found`. That broke the re-login flow for an already suspended user: the backend rejects the credentials with a 403 before any session/OTP fields are ever set, so on retry the session was empty, the controller redirected to `/not-found`, and the user could never reach this page again. Removing the session check lets a suspended user consistently land on this informational page on every retry.

portal-ui/server/controllers/login/get-account-suspended.test.ts
```diff
@@ -1,54 +1,23 @@
...
-      session: {
-        numberOfSignInOtpAttemptsRemaining: 2,
-      },
-    } as unknown as GetAccountSuspendedPageRequest;
-
-    getAccountSuspendedPage(req, res);
-
-    expect(redirectMock).toHaveBeenCalledTimes(1);
-    expect(redirectMock).toHaveBeenCalledWith('/not-found');
-  });
```
sungharsh-ukef, line 53: These tests were intentionally removed.

They asserted the previous behaviour where the controller read req.session.numberOfSignInOtpAttemptsRemaining and redirected to /not-found unless it was exactly -1. That session check broke the re-login flow for an already suspended user: on retry the backend rejects the credentials with a 403 before any session/OTP fields are set, so the session was empty and the user was sent to /not-found instead of the suspended account page.

The page is purely a static informational message (heading, suspension explanation and public support contact details) with no user- or session specific data, so the controller now always renders it. There is no longer a /not-found branch to test.

e2e-tests/portal/cypress/e2e/journeys/admin/activate-deactivate-user/activate-deactivate-user.spec.js
zainUKEF, line 1: Why has this been updated? Does this need to be updated now or can it wait?

e2e-tests/portal/cypress/e2e/journeys/feature-flag-enabled/login/2fa/pages/temporarily-suspended-access-code.page.spec.js
```diff
@@ -15,12 +15,12 @@ context('2FA Page - Temporarily suspended account', () => {
...
   });
 
-  it('should redirect to login when visited without partial auth', () => {
+  it('should render account suspended page when visited without partial auth', () => {
     cy.clearCookies();
     cy.clearLocalStorage();
 
     cy.request({ url: '/login/temporarily-suspended-access-code', followRedirect: false }).then((resp) => {
-      expect(resp.headers.location).to.equal('/login');
+      expect(resp.status).to.equal(200);
```
zainUKEF: we should not be testing a status code in e2e test

e2e-tests/support/tasks.js
```diff
@@ -52,6 +52,18 @@ module.exports = {
       return users.findOne({ username: { $eq: username } });
     };
 
+    /**
+     * Updates a portal user document by username for E2E setup/teardown.
+     *
+     * This is used by tests that need to temporarily adjust a seeded user's
+     * role or other simple fields without going through the UI.
+     */
+    const updatePortalUserByUsername = async ({ username, update }) => {
```
zainUKEF, line 61: do we not already do something similar to update expiry etc?

portal-ui/server/controllers/login/post-login.js
```diff
@@ -77,6 +77,7 @@ export const postLogin = async (req, res) => {
 
         if (status === HttpStatusCode.Forbidden) {
           console.error('Access temporarily suspended for user %s', email);
+          req.session.numberOfSignInOtpAttemptsRemaining = -1;
```
zainUKEF, line 80: will forbidden response always lead to temporarily suspended?

e2e-tests/portal/cypress/e2e/journeys/admin/activate-deactivate-user/activate-deactivate-user.spec.js
sungharsh-ukef: This test is failing, because when user is deactivated its no longer valid to assert `/login` with 2fa enabled. we need to redirect the user to `login/temporarily-suspended-access-code`, in this refactor we now using `if else` to switch between tests when feature flag enabled /disabled. 
```js
if (PORTAL_2FA_FF === 'true') {
  cy.url().should('eq', relative('/login/temporarily-suspended-access-code'));
} else {
  cy.url().should('eq', relative('/login'));
}

File: portal-ui/server/controllers/login/post-login.js (hunk 6 above)
Comment (sungharsh-ukef) [reply portal-ui/server/controllers/login/post-login.js:80]: Yes in this `post login` controller context we treat 403 forbidden to be temporarily suspended always. Where as in other contexts such as in `attempt-otp-login.ts`, 401 or 403 is treated as incorrect OTP, and if the error payload says expired, it becomes expired instead of suspended. Similarly also applies to incorrect code.

File: e2e-tests/support/tasks.js (hunk 5 above)
Comment (sungharsh-ukef) [reply e2e-tests/support/tasks.js:61]: For expiry we have different helpers. i.e OTP expiry and attempt count `cy.overridePortalUserSignInOTPSendCount` and `cy.overridePortalUserSignInOTPWithExpiredToken ` 

but for this spec, we need this `cy.updatePortalUserByUsername` and it is just being used for a simple role change (admin and user), this is more targeted helpers. we don't need expiry or attempts state for this.

File: e2e-tests/portal/cypress/e2e/journeys/feature-flag-enabled/login/2fa/pages/temporarily-suspended-access-code.page.spec.js (hunk 4 above)
Comment (sungharsh-ukef) [reply e2e-tests/portal/cypress/e2e/journeys/feature-flag-enabled/login/2fa/pages/temporarily-suspended-access-code.page.spec.js]: totally missed this, now update the test to assertText

File: e2e-tests/portal/cypress/e2e/journeys/admin/activate-deactivate-user/activate-deactivate-user.spec.js [hunk 7]
```diff
@@ -21,11 +25,29 @@ context('Admin user updates an existing user', () => {
...
+      update: {
+        roles: [MAKER],
+      },
+    });
   });
 
   it('Create a user, then edit the user and change their role(s)', () => {
     // login and go to dashboard
-    cy.login(ADMIN);
+    cy.resetPortalUserStatusAndNumberOfSignInOTPs(adminUser.username);
```
zainUKEF: do we need this if FF is off?

sungharsh-ukef: we need this, its not about the 2FA flag itself. It resets the admin user’s `sign-in state` so the test can log in cleanly before doing the role change. If we remove it, the admin user can be left with stale resend/OTP state from earlier test steps and the login setup becomes flaky.

zainUKEF: yes but if the FF is off, there will be no OTP, it uses sign in links

e2e-tests/portal/cypress/e2e/journeys/admin/activate-deactivate-user/activate-deactivate-user.spec.js
```diff
@@ -21,11 +25,29 @@ context('Admin user updates an existing user', () => {
...
+    });
+  });
+
+  afterEach(() => {
+    cy.task('updatePortalUserByUsername', {
+      username: adminUser.username,
+      update: {
+        roles: [MAKER],
+      },
+    });
```
zainUKEF: I also don't understand why we need this.  It wasn't needed before and functionality hasn't changed

e2e-tests/portal/cypress/e2e/journeys/admin/activate-deactivate-user/activate-deactivate-user.spec.js
sungharsh-ukef: I have refactored this test to improve clarity. I have reverted the existing test to remain unchanged; I’ve only added a condition so it runs only when the 2FA feature flag is disabled.

I also created a new `describe` block with its own `it` statement to test activating/deactivating a suspended user. This new test runs only when the 2FA feature flag is enabled.

e2e-tests/portal/cypress/e2e/journeys/admin/activate-deactivate-user/activate-deactivate-user.spec.js
sungharsh-ukef: This setup is needed for the 2FA suspended-page test so the seeded user can be temporarily given admin access, run through the admin flow, and then be restored afterward.

e2e-tests/portal/cypress/e2e/journeys/admin/activate-deactivate-user/activate-deactivate-user.spec.js
```diff
@@ -20,12 +21,15 @@ context('Admin user updates an existing user', () => {
...
 
   beforeEach(() => {
-    cy.removeUserIfPresent(userToUpdate, ADMIN);
+    cy.removeUserIfPresent(userToUpdate, ADMIN_USER);
   });
 
   it('Create a user, then edit the user and change their role(s)', () => {
+    if (PORTAL_2FA_FF === 'true') {
+      return;
+    }
```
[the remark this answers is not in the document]
sungharsh-ukef: addressed, refactored to declare `const is2FAEnabled = PORTAL_2FA_FF === 'true';` at higher scope, and use `if()else()` conditional around, no more returning.

e2e-tests/portal/cypress/e2e/journeys/admin/activate-deactivate-user/activate-deactivate-user.spec.js
```diff
@@ -71,4 +75,76 @@ context('Admin user updates an existing user', () => {
...
+
+  describe('when the 2FA feature flag is enabled', () => {
+    const adminUser = BANK1_MAKER1;
+
+    beforeEach(() => {
+      cy.removeUserIfPresent(userToUpdate, ADMIN_USER);
+
+      // Use a seeded user with a working local 2FA path for the admin UI steps.
+      cy.task('updatePortalUserByUsername', {
+        username: adminUser.username,
```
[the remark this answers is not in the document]
sungharsh-ukef: refactored the `beforeEacb/afterEach` only run in the `else()` to test when 2FA flow is enabled.

e2e-tests/portal/cypress/e2e/journeys/admin/activate-deactivate-user/activate-deactivate-user.spec.js
```diff
@@ -71,4 +75,76 @@ context('Admin user updates an existing user', () => {
...
+        update: {
+          roles: [MAKER],
+        },
+      });
+    });
+
+    it('should show the temporarily suspended access code page when 2FA is enabled', () => {
+      if (PORTAL_2FA_FF !== 'true') {
+        return;
+      }
```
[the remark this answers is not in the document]
sungharsh-ukef: done refactored

portal-ui/server/routes/login/index.js
```diff
@@ -397,7 +397,7 @@ router.post('/login/resend-another-access-code', validatePortal2FAEnabled, valid
  *       403:
  *         description: Forbidden
```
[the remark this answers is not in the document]
sungharsh-ukef: done, updated

e2e-tests/portal/cypress/support/portal/login.js
```diff
@@ -2,10 +2,8 @@ const { signInLink } = require('../../e2e/pages');
 const relative = require('../../e2e/relativeURL');
 const { SIGN_IN_TOKENS } = require('../../fixtures/constants');
 
-const PORTAL_2FA_FF = Cypress.env('FF_PORTAL_2FA_ENABLED');
-
 module.exports = ({ username, password }) => {
-  if (PORTAL_2FA_FF === 'true') {
+  if (Cypress.env('FF_PORTAL_2FA_ENABLED') === 'true') {
```
zainUKEF: why has this been done?

zainUKEF (approved):

audrius-vaitonis (approved):
Hope it's fine

PR #4824, opened 2026-06-19
chore(DTFS2-8496): split dependabot npm into groups to avoid timeout

zainUKEF:
# Introduction :pencil2:

Dependabot was timing out with error "Dependabot timed out during its update" (Version update 1413705797) when attempting to resolve npm dependencies.

* Replaced the single patterns: ['*'] npm group with 6 smaller groups (types, eslint, babel, testing, production, dev-other) to reduce the resolution workload per group
* This prevents Dependabot from trying to resolve compatibility for all 2000+ packages in a single pass, which was causing the timeout
* Added documentation comments explaining the grouping strategy

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,6 +3,12 @@
 # Automatic dependencies updates
 # for the following services:
 # 1. NPM
+#    - Uses a single root directory (/) which covers all workspace packages
+#    - Split into 6 groups (types, eslint, babel, testing, production, dev-other)
+#      to avoid Dependabot timeout caused by resolving all 2000+ packages at once
+#    - Groups 1-4 match by package name pattern (e.g. @types/*, eslint*, jest*)
+#    - "production" = packages in "dependencies" in package.json
+#    - "dev-other" = packages in "devDependencies" not matched by groups 1-4
 # 2. Docker
 # 3. GHA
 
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -17,9 +23,55 @@ updates:
...
+          - '@typescript-eslint/*'
+          - 'eslint-*'
+          - '@babel/*'
+          - 'babel-*'
+          - 'jest'
+          - 'jest-*'
+          - '@jest/*'
+          - 'ts-jest'
+          - 'cypress'
+          - 'supertest'
```
tony-barnes-ukef, line 81: Should we order groups and `exclude-patterns` alphabetically?

zainUKEF: done

.github/dependabot.yml
```diff
@@ -17,9 +23,59 @@ updates:
...
+          - '@jest/*'
+          - '@types/*'
+          - '@typescript-eslint/*'
+          - 'babel-*'
+          - 'cypress'
+          - 'eslint*'
+          - 'jest'
+          - 'jest-*'
+          - 'supertest'
+          - 'ts-jest'
```
sungharsh-ukef, line 48: I would also include these three to extend testing patterns alphabetically 
```yml
- '@cypress/*'
- '@cypress-*'
- 'nock'

File: .github/dependabot.yml [hunk 3]
```diff
@@ -17,9 +23,59 @@ updates:
...
+      npm-dev-dependencies:
         patterns:
           - '*'
+        dependency-type: 'development'
+        exclude-patterns:
```
sungharsh-ukef, line 53: and here too additional test patterns same as above

```yml
- '@cypress/*'
- '@cypress-*'
- 'nock'
```
I would also remove `eslint-*` at 67 as as we already have `eslint*` above it which covers the linting already.

.github/dependabot.yml
```diff
@@ -17,9 +23,59 @@ updates:
...
+          - 'eslint-*'
+          - 'jest'
+          - 'jest-*'
+          - 'supertest'
+          - 'ts-jest'
+      npm-eslint:
+        patterns:
+          - '@typescript-eslint/*'
+          - 'eslint*'
+          - 'eslint-*'
```
sungharsh-ukef: I think one is enough that is `- 'eslint*'` the next one is unessential

.github/dependabot.yml
sungharsh-ukef: if included above then we need to mirror here too 
```yml
- '@cypress/*'
- '@cypress-*'
- 'nock'

tony-barnes-ukef (approved):

PR #4902, opened 2026-07-29
chore(DTFS2-8535): change release deployments to be max-parallel 1 to stop failures

zainUKEF:
# Introduction :pencil2:

This PR changes release.yml to have only 1 container deployment at a time to stop failures.  Otherwise deployments take a long time as they need to keep being rerrun.

* Change max-parallel to 1 in release.yml

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -112,7 +112,7 @@ jobs:
       ENVIRONMENT: ${{ needs.setup.outputs.environment }}
 
     strategy:
-      max-parallel: 2
+      max-parallel: 1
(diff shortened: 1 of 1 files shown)


sungharsh-ukef (approved):

tony-barnes-ukef (approved):

sungharsh-ukef:
its just an openion: The trade-off is only throughput, deployments will take longer, but they will be more serialized and may reduce failure if the jobs are competing for shared deployment resources. If the goal to reduce failures 1 is fine. If not we can always revert this back to 2 to speedup.

PR #4946, opened 2026-08-14
chore(deps): update dependencies

zainUKEF:
# Introduction :pencil2:

Updates dependencies and updates failing tests

* Update dependencies
* Update failing tests
* Fix date issues on new date-fns


portal-ui/jest.common.config.js
```diff
@@ -6,6 +8,7 @@ module.exports = {
   },
   moduleNameMapper: {
     '^.+\\.(css|less|scss)$': 'babel-jest',
+    '^cheerio/slim$': path.join(path.dirname(require.resolve('cheerio/package.json')), 'dist/commonjs/slim.js'),
```
tony-barnes-ukef, line 11: What's the need for this?

portal-ui/package.json
```diff
@@ -20,7 +20,7 @@
   "main": "server/index.ts",
   "scripts": {
     "api-test": "jest --coverage --verbose --forceExit --config=api-test.jest.config.js",
-    "build": "rimraf ./public/js && rimraf ./public/css && webpack --mode production --config webpack.prod.config.js",
+    "build": "rimraf ./public/js && rimraf ./public/css && webpack-cli --mode production --config webpack.prod.config.js",
```
tony-barnes-ukef: Should be able to revert this?

scripts/update-ui-asset-integrity.mjs
```diff
@@ -0,0 +1,227 @@
+import crypto from 'node:crypto';
```
tony-barnes-ukef: Great we're looking at automating this.

This introduces a new way to automatically update the hashes. GIFT has a way of doing this (`asset.middleware.ts`).

Has there been any discussion about this and which approach we should follow?

tony-barnes-ukef (changes requested):

portal-ui/jest.common.config.js
zainUKEF: cheerio 1.2 has some changes.  It does not work with our old require statements so we're using the slim version.  This is needed to overwrite those errors

scripts/update-ui-asset-integrity.mjs
zainUKEF: not yet, I can remove this and we can introduce this later

portal-ui/package.json
zainUKEF: done

tony-barnes-ukef (approved):

PR #4959, opened 2026-08-20
chore(DTFS2-8570): remove passwords from login errors

zainUKEF:
# Introduction :pencil2:

Currently passwords are printed in error messages for logging in as they are part of the req.body.  This PR removes them from error messages

* Use set error format in portal and tfm error to remove password from errors


tony-barnes-ukef (commented):
"Currently passwords are printed in error messages for logging in as they are part of the req.body"

I think this should be addressed instead of adding middleware to workaround the issue.

Can we update the logging to not log `req.body` (and therefore the password)? I don't believe it's a good practice.

PR #4960, opened 2026-08-20
feat(DTFS2-8542): amend a gift facility - multiple amendments

tony-barnes-ukef:
# Introduction :pencil2:

This PR updates TFM API and External API to utilise a new APIM TFS endpoint, for making multiple amendments to a single GIFT facility.

This change means that if there is more than 1x amendment, e.g, change amount and change date - DTFS will now only call APIM TFS once with a single payload. Additionally, this means that APIM TFS will bundle all amendments into a single package.

Previously, DTFS had to make multiple API calls and APIM TFS would create a work package for each amendment.

- External API:
  - Add new endpoint, update routes, controller, tests, swagger docs.
- TFM API:
  - Create new method, `api.multipleGiftFacilityAmendments`
  - Update `submitFacilityAmendmentsToApimGift` to conditionally call the singular, or the multiple APIM/GIFT amendment endpoint.

- Replace all instances of `.toEqual(undefined)` with `.toBeUndefined()`.

## Request / response :eyes:

<details>
  <summary>Show/hide</summary>

```json

```

</details>

<details>
  <summary>Show/hide</summary>

Add screenshots here.

</details>


external-api/server/v1/controllers/gift-facility.controller.ts
```diff
@@ -298,3 +298,63 @@ export const amend = async (req: GiftFacilityRequest, res: Response) => {
...
+        error,
+      );
+
+      return { status };
+    });
+
+    const { status } = response;
+
+    if (status !== HttpStatusCode.Accepted) {
+      return res.sendStatus(status);
```
zainUKEF, line 339: do you want to send a successfalse flag when its not successful

tony-barnes-ukef: I don't think so, it wouldn't be used and I don't believe there's patterns like that in `external-api`

The status code returned will indicate it not being successful.

zainUKEF (approved):

PR #4962, opened 2026-08-21
fix(DTFS2-8576): reset otp send count when user reactivated

zainUKEF:
# Introduction :pencil2:

This PR fixes an issue where a user's sign in OTP send count is not reset when user is reactivated by an admin

* Added to update users controller in portal-api to reset otp send count if FF is set
* Added unit test
* Added e2e test

### DIFF
--- a/portal-api/server/v1/users/controller.js
+++ b/portal-api/server/v1/users/controller.js
@@ -1,4 +1,4 @@
-const { MONGO_DB_COLLECTIONS, DocumentNotDeletedError, TIMEZONE, generatePasswordHash } = require('@ukef/dtfs2-common');
+const { MONGO_DB_COLLECTIONS, DocumentNotDeletedError, TIMEZONE, generatePasswordHash, isPortal2FAFeatureFlagEnabled } = require('@ukef/dtfs2-common');
 const { PORTAL_USER } = require('@ukef/dtfs2-common/schemas');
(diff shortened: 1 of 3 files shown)


portal-api/server/v1/users/controller.js
```diff
@@ -189,11 +189,20 @@ exports.update = async (_id, update, auditDetails, callback) => {
     if (existingUser['user-status'] === USER.STATUS.BLOCKED && userSetUpdate['user-status'] === USER.STATUS.ACTIVE) {
       // User is being re-activated.
       userSetUpdate.loginFailureCount = 0;
-      userUnsetUpdate = {
-        signInLinkSendDate: '',
-        signInLinkSendCount: '',
-        blockedStatusReason: '',
-      };
+
+      if (isPortal2FAFeatureFlagEnabled()) {
```
tony-barnes-ukef, line 200: Can we add a comment about these conditions?

Does this introduce a new data type - each field can be a number or string, I'd expect it to be a `number | null`?

zainUKEF: Will add a comment.  I have copied the implementation of resetStaleSignInData which sets these fields to 0 so keeping it consistent

zainUKEF: signInLinkSendDate and signInLinkSendCount usually are set to empty strings but OTP is set to 0

tony-barnes-ukef (approved):

